PR #13, opened 2015-08-24
Add forceUpdate value in database for integer values

szczukot:
What do you think about this solve the problem of duplicate data?
It is much better than the insertion in each module separately. I've done universally.
Advantages :
- Saved all other data (battery level, last seen etc)
- You can easily be applied to other module saves any integer

In the diff is a lot of differences in the spaces etc, but I set the IDE to me rearranging. Give me some code guideline :)

### DIFF
--- a/hardware/SatelIntegra.cpp
+++ b/hardware/SatelIntegra.cpp
@@ -40,6 +40,8 @@ static Model models[TOT_MODELS] =
 	{72, "Integra 256 Plus", 256, 256},
 };
 
+#define MAX_LENGTH_OF_ANSWER 63 * 2 + 4
+
 const unsigned char partitions[4] = {0xFF, 0xFF, 0xFF, 0xFF};
 
 SatelIntegra::SatelIntegra(const int ID, const std::string &IPAddress, const unsigned short IPPort, const std::string& userCode) :
@@ -868,7 +870,7 @@ int SatelIntegra::SendCommand(const unsigned char* cmd, unsigned int cmdLength, 
 
 	delete cmdPayload.first;
 
-	unsigned char buffer[67];
+	unsigned char buffer[MAX_LENGTH_OF_ANSWER];
 	// Receive answer
 	fd_set rfds;
 	FD_ZERO(&rfds);
@@ -884,7 +886,7 @@ int SatelIntegra::SendCommand(const unsigned char* cmd, unsigned int cmdLength, 
 		return -1;
 	}
 
-	int ret = recv(m_socket, (char*)&buffer, 67, 0);
+	int ret = recv(m_socket, (char*)&buffer, MAX_LENGTH_OF_ANSWER, 0);
 
 	if (ret > 6)
 	{


gizmocuz:
Thank you for the patch, but you can not modify the structures in rfxtrx.h, this is not our file, but from RFXCom, and will be overwritten when there is a new SDK

szczukot:
This is not good news :)
Any other idea for solve this task ?

Maybe add bool column "forcedUpdate" to DeviceStatus ?

gizmocuz:
It seems you have problems with the GPIO hardware class, i suggest to only make a patch for this,
or was it with a json script call ? then i would suggest to make the fix in the script thats calls the json api

szczukot:
I use OWFS (not GPIO).
But for Satel I have this same problem. Often they go queries, but data changes infrequently.
I think, many of module for domoticz work this same - and unnecessarily pump data to database.

gizmocuz:
not really, it is up to the hardware class implementer to not send the same switch command over and over again.
Maybe you can propose a patch

szczukot:
I proposed patch, but RFXcom ... :)
I see two solution :
1) Add bool "forceupdate" column in DeviceStatus and use it
2) Add third parameter to sDecodeRXMessage

Which solution try to implement? First ?

gizmocuz:
some switches are one-way, they do not toggle, and you want them in the log.
Better to fix the hardware class that needs to have this option, then to do it globaly

szczukot:
Not necessarily. Satel have one-way switches or zones. 1Wire have two-way switches, 
Why write over and over again the same state to the database ?
Fix hardware class is not good, because not update fields such as: last seen, battery status, etc.

szczukot:
I have question : is all hardware must report value called sDecodeRXMessage ? 
So I have to use the structure of tRBUF for this method ?
Do I limited to the use structures from  RFXtrx.h ?

I'm using now incorrectly types / subtypes such as 
1) sTypeSecX10 - for zones
2) sTypePercentage - for Integra outputs which can not be switched
3) sTypeSecX10R - for "arm" state
4) ..... still none - for info about alarm.

What types can instead use to it would make more sense?? This is very important for me.

gizmocuz:
pTypeGeneral you can also use

gizmocuz:
Yes, in all hardware you must use sDecodeRXMessage, otherwise it will not be send to remote domoticz systems (if you use this)
4) You could use a pTypeGeneral/sTypeText for messages/info

When you read out the status, do you also have a timestamp of the last action?
If this is the same, you do not have to send the switch

szczukot:
1) How do I may use subtype STypeText or sTypeAlert ? How do send text ?? In structure tGeneralDevice i don't see any field fot text/string.
Can I call UpdateValue directly ?? Or expand structure tGeneralDevice and add decode_xyz etc ?
2) I do not have timestamp, but I can check the last state and send it or not. But then (when i dont send) on the website I can't see when the unit was last seen, and I wonder: working or not?

gizmocuz:
If you create a dummy text/alert sensor you see the parameters in the database
also if you search all file in the source folder for STypeText, you will see how it works

szczukot:
I see, that use directly UpdateValue method. 
But in this case, not working many other method realted with DecodeRXMessage (set heartbeat, logs, send to sharing user, check scene ... etc)

gizmocuz:
Thats correct it will not send it to the sharing users, check scene with a text??? heartbeat??

PR #20, opened 2015-08-26
[Satel Integra] Change subtypes and handle alarm

szczukot:


szczukot:
Many changes but still working :)

PR #24, opened 2015-08-27
Mqtt for switches

ghoben:
Changing EnOcean switch behaviour to show button presses and releases.

Each button will be handled independent as an ON/OFF button. We so have 6 buttonIDs; 4 buttons (A0, AI, B0, B1) (Rocker A UP, Rocker A Down, Rocker B UP, Rocker B Down) and 2 buttons showing the simultaneous press of A0+B0 and A1+B1. The button is ON when it's pressed and released if it's not pressed.


gizmocuz:
The topic of this commit (mqtt....) does not match your changes,
and i see you also changed the lighting2 group command behaviour
could you explain this a bit more?

ghoben:
Sorry for the wrong branch name of commit. It should be titled EnOcean-Button-Handling.

Well, I found that the button functionality for EnOcean Switches was broken, due to the release message just always releasing the same button unit 1. After looking at the specs and the code and experimenting, I found no solution to get the correct unit number from the release message to set the button back to OFF. I found the group_OFF and group_ON in the code. What I do as described in the commit I treat every button as individual. According to specs, the release message is always send after a pressed button get's released. Therefore, the release message sends an group_OFF command to the mainworker.cpp, which identifies it as a group command. Then I set back the currently pressed button from the same switch and update the SQLITE DB. I changed the GroupCmd also to set the timestamp of last update correctly. Since I do that for the group in the other method I can get rid of the old group cmd. I changed the GroupCmd in the way, that only those switches are selected, which are not already on the same nvalue. Make Sense?

gizmocuz:
patches looking good, but do you suspect any problems with the current lighting2 group on/off messages ? (from a kaku remote)

ghoben:
I will close this pull request and create on with the correct title.

PR #25, opened 2015-08-27
Changing EnOcean switch behaviour to show button presses and releases isolated.

ghoben:
Changing EnOcean switch behaviour to show button presses and releases.

Each button will be handled independent as an ON/OFF button. We so have 6 buttonIDs; 4 buttons (A0, AI, B0, B1) (Rocker A UP, Rocker A Down, Rocker B UP, Rocker B Down) and 2 buttons showing the simultaneous press of A0+B0 and A1+B1. The button is ON when it's pressed and released if it's not pressed.

Well, I found that the button functionality for EnOcean Switches was broken, due to the release message just always releasing the same button unit 1. After looking at the specs and the code and experimenting, I found no solution to get the correct unit number from the release message to set the button back to OFF. I found the group_OFF and group_ON in the code. What I do as described in the commit I treat every button as individual. According to specs, the release message is always send after a pressed button get's released. Therefore, the release message sends an group_OFF command to the mainworker.cpp, which identifies it as a group command. Then I set back the currently pressed button from the same switch and update the SQLITE DB. I changed the GroupCmd also to set the timestamp of last update correctly. Since I do that for the group in the other method I can get rid of the old group cmd. I changed the GroupCmd in the way, that only those switches are selected, which are not already on the same nvalue.


ghoben:
@gizmocuz . I hope it's clearer this time.

Thanks. Any feedback welcome. It's my first contribution to domoticz, so tell me if I should do something differently.

gizmocuz:
It all looks good, but i hope the current kaku lighting2/group on/off is still working.
i apply the patch and do some testing, but i expect its OK
Thanks for the patch!

gizmocuz:
Oeps... someone just did a merge, and i can not merge anymore as there is a conflict

ghoben:
My assumption was: Since former an OFF message was sent. I changed it to a group OFF which actually should set all association UNITS of an ID to OFF. We might see problems if that should not be the case, which I don't think so. Other than that, we now have a full run on the particular Switch updating the DB also with time stamps.  I have not used the Kaku as you mentioned. Thanks for applying.

ghoben:
I can rebase my branch does that help?

gizmocuz:
i think it will... i am still to unexperienced, it looks like your branch is out of date

ghoben:
@gizmocuz  Should up-to-date now.

gizmocuz:
Thanks, that worked!

gizmocuz:
hope all is up2date now... strange things where happening when i merged this

ghoben:
@gizmocuz I issued another pull request #27 for a needed bug fix.

gizmocuz:
Thanks, its merged

EddyK69:
There is new topic with a GroupOn related problem, see http://www.domoticz.com/forum/viewtopic.php?f=6&t=7787

PR #33, opened 2015-08-30
Bugfix, logging group_on/group_off

ghoben:
Fixed the problem where a switch update from an AC lighting2 switch was
not logged. In case a group_on/group_off command is received, it will
also update all switches (units) from the same ID and updates their logs
too with a group_on/group_off. Therefore a real received off/on can be
distinguished from a group_off/off.


gizmocuz:
Thanks, merged!

PR #34, opened 2015-08-31
Added json-calls for deleting expired (scene)timers

EddyK69:
Added clearexpiredtimers
Added clearexpiredscenetimers


EddyK69:
I could try. I'm not sure where to put the code, in ReloadSchedules() or in Do_Work()
Or create a new method with the sql-delete query and a ReloadSchedules?

gizmocuz:
best to do it in the scheduler.cpp, when an item is triggered, that is a one-time (fixed date), then delete it from the database

EddyK69:
OK, i'll do that in a new request. Closing this one

PR #36, opened 2015-08-31
Basic Tellstick support

martenbohlin:
To actually get the tellstick support you need to have the telldus-core devel package installed. On systems with apt package manager that can be done like this:

Edit the file /etc/apt/sources.list and add the following line:
deb http://download.telldus.com/debian/ stable main

Run the following commands to install the key and to install the headers and libraries:
wget -q http://download.telldus.se/debian/telldus-public.key -O- | sudo apt-key add -
sudo apt-get update

sudo apt-get install libtelldus-core-dev

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,8 @@
-*.orig
+Makefile
+CMakeFiles
+*.cmake
+*.a
+CMakeCache.txt
+appversion.h*
+domoticz
+stdafx.h.gch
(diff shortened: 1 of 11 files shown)


gizmocuz:
Thanks a lot for the hard work !
Could you make/update our wiki page ?

martenbohlin:
Yes I will take a look at the wiki page to.

PR #37, opened 2015-08-31
Fixed switching scenes/groups from Device-menu

EddyK69:
Second attempt :D now with cached icons.

GitHub merge is giving me headaches ;(


gizmocuz:
Thanks! merged

PR #40, opened 2015-09-01
Rego6XX hardware device bugfixes

mattsjo:
Long overdue fixes for problems that started somewhere around Christmas causing switch logs to be broken (including change notification and timeout) and  month and year graphs for counters to be empty.

The following changes has been done:
1, The follow-through intention of switch clause cases in SQLHelper.cpp didn't work for the Rego6XX type.
2, The Rego6XX switch wasn't recognized as a light switch in Webserver.cpp
3, The MTYPE_COUNTER wasn't implemented for general graphs.

I plan to revise the fixes for the MTYPE_COUNTER as it is unnecessary to use a float to represent an integer but it works for now.

### DIFF
--- a/main/SQLHelper.cpp
+++ b/main/SQLHelper.cpp
@@ -2564,7 +2564,7 @@ unsigned long long CSQLHelper::UpdateValueInt(const int HardwareID, const char* 
             break;
         }
 	case pTypeGeneral:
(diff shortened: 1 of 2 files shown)


gizmocuz:
Thanks for the fixes !!

PR #42, opened 2015-09-02
About-background

ghost:
Better walpaper to background


ghost:
Can you give reason why my PR was rejected ?

gizmocuz:
Because, as we already discussed by email, this page is far from finished, it is now handy that it shows the branch number.
Also i am not going to change this background (as also discussed)

PiterEL:
Ok but add translation with i add to other PR and add to Transifex. But what you plan change/add to this page ABOUT if it is not finished ?

But in brand h2 BUILD DATE is the same as in About page COMPILE DATE. Can we change in all site to BUILD DATE or COMPILE DATE ?Thi is the same info...with other name.

gizmocuz:
it is in the beta, anything can be in there, and being worked on!

PR #46, opened 2015-09-03
Added tag for not auto capitalize first letter of username

perkovich:
Accidently changed the wrong file first. Then the correct file. Then removed the added changes in the first file (unfortunately mig Git-fu isn't very strong.


gizmocuz:
Thanks ;)

PR #47, opened 2015-09-03
Add string to translation

ghost:


gizmocuz:
i like to keep it the way it is.
also, i do not want any user, that can not read/write English, to report any bugs.

this page is also not finished, so no need for translation yet,
it is quickly made, so people can report/read the build hash.

ghost:
Everything is not finished. This is when you can enter translations and finish it?
The introduction of translation in no way interferes ...

Everything I submit to reject the changes :(

PR #50, opened 2015-09-07
Add name for Mi Light

ghost:


ghost:
Can you add this ?

gizmocuz:
How about Easybulb ? and there are more names also

ghost:
But look at forum that user agree with me to add Mi Light. More people search hardware Mi Lisght and can dont know that is the same as Limitless.... Mi Light is also popular as Limitless/AppLamp and should be set in name hardware in Domoticz.
Please add this

gizmocuz:
Thanks!

PR #53, opened 2015-09-07
Applamp/Limitless Disco Mode Command Implementation

sijones:
Can this be merged?


gizmocuz:
Thanks for the patch ! Hope its groovy ;)

mikefranken:
Nice, but how can I use this? What URL?

sijones:
If you go in to the suggestions forum and the rgbw applamp improvement
thread you'll see my comment on the url to use it.

PR #55, opened 2015-09-08
Fix: Keep scene-activatorcommand during migration

EddyK69:
Fix: Keep ListenCmd for scenes while updating to new activators-structure.
Fix: SQL error while trying to add MySensorsChilds-fields that already exists.


gizmocuz:
Thanks

PR #61, opened 2015-09-13
Fix to use always backround from styles in events

ghost:


ghost:
Please Can you merged this...?

PR #63, opened 2015-09-13
Extend for sTypeAlert type

szczukot:
Hi
What do you think about this extension for sTypeAlert ?
I use sDecodeRXMessage instead UpdateValue method


gizmocuz:
i am not going to look at this code change.
Please do not do any formatting what so ever on existing code.
please provide a clean patch 'with' actual code changes

PR #64, opened 2015-09-13
Add support for setting setpoint at remote server

mikefranken:
Forwarding OpenTherm Gateway setpoint
Log when Writing to hardware isn’t supported


gizmocuz:
Thanks!

PR #68, opened 2015-09-14
Extend for sTypeAlert type

szczukot:
Hi again

What do you think about this extension for sTypeAlert ?
I use sDecodeRXMessage instead UpdateValue method

Fantom

### DIFF
--- a/hardware/SatelIntegra.cpp
+++ b/hardware/SatelIntegra.cpp
@@ -407,7 +407,7 @@ bool SatelIntegra::ReadZonesState(const bool firstTime)
 				{
 					ReportZonesViolation(index + 1, violate);
 					//UpdateZoneName(index + 1, &buffer[4], buffer[20]);  	 Security1 does not support units - partition
-					UpdateZoneName(index + 1, &buffer[4], 0);
+					UpdateZoneName(index + 1, &buffer[4], 1);
 				}
 				else
 				{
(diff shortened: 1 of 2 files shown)


gizmocuz:
Thanks, looking good ;)

szczukot:
Hi. I have problem - I think with this change.
But when i running domoticz from server (check updates etc) Satel Integra not working :(
Method sDevodeRXMessage was not called or not woking inside. I calling  ReportZoneViolation, but nothing changes, and "last seen" equal a few days ago.

When i compiling (the newest) code on Ubuntu or Rapsberry PI is working ok.

What is wrong ?

gizmocuz:
Hi,

You made the patch, i can not debug this.

Could you try to debug this?

szczukot:
but how to do this debugging ? 
I downloaded the latest sources, compiled both on Linux and in Raspberry PI and its working properly on both platform. But package from "auto update" is not working -value is not updated :(

How enabled DEBUG LOG on package from "auto update" ? I set verbose = 1 but i dont ssee too many logs.

Maybe i create patch for git with many debug logs ? 

Fantom

gizmocuz:
maybe this patch was not yet available in the automatic beta update ?
if you go to the about page, you see the build hash number, you can then see in the commit log on github if this patch was included

szczukot:
I have version with hash bdf47b0 - 22 sep 2015. My patch is exists in this version.
I will create new patch with more logs today.

Fantom

gizmocuz:
just made another beta

PR #71, opened 2015-09-15
MQTT: add ca file option.

xs4:
By setting the CA file, TLS mode will be enabled.
Use the CA file that was used to sign the MQTT server certificate.

Note that the CA file must reside on the Domoticz server.

### DIFF
--- a/hardware/MQTT.cpp
+++ b/hardware/MQTT.cpp
@@ -16,10 +16,11 @@
 #define TOPIC_IN	"domoticz/in"
 #define QOS         1
 
-MQTT::MQTT(const int ID, const std::string IPAddress, const unsigned short usIPPort, const std::string Username, const std::string Password) :
+MQTT::MQTT(const int ID, const std::string IPAddress, const unsigned short usIPPort, const std::string Username, const std::string Password, const std::string CAfilename) :
 m_szIPAddress(IPAddress),
 m_UserName(Username),
-m_Password(Password)
+m_Password(Password),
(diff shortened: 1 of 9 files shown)


gizmocuz:
Thank you for the patch.

Could you modify it a bit:
- In the harware table, use as field name 'Extra' (so we can use this field later for other hardware, Filename is a bit to direct and only targets MQTT)
- the CA file, can we use the CA file domoticz is already using for HTTPS ?
  If yes, then this would be great, and we could reuse this
- or we can make an additional commandline parameter for the MQTT CA file, so there is no need for the above modification, and only use it if the 'secure' checkbox is checked in the MQTT setup ?
- then this patch could be reduced 90%

xs4:
You can share the CA file between Domoticz HTTPS and the MQTT TLS. But that would also mean that the **must** have the same CA. I am not sure you want that restriction. 

Configure the CA via GUI or command line? Both are possible, but that would mean that we must configure our hardware partly via the GUI and partly via the command line, is that desirable from a usability standpoint?  I would prefer to keep it in the GUI, but it is up to you.

I you agree with this approacher I can rework the patch and rename the column name to Extra and also change the http parameter name from filename to extra.

xs4:
Changed the database column name to Extra, also changed the http parameter name to extra.

I looked at Domoticz https handling, but  at the moment Domoticz only accepts a certificate and a private key (so no CA certificate yet).

gizmocuz:
Thanks!

PR #73, opened 2015-09-16
My first request ;-)

gaudryc:
I send you these patches to solve the following issues :
- no log to know if someone is trying to brute force a user password ;
- consumption graphs never show 0 value even if it exists (eg. when counter value remains the same) ;
- a Energy/CMxxx device that have both "Usage" and "Value" values will have a day graph with "Usage" values instead of "Value" values.


remb0:
Great! I missed also the log. I get sometimes incoming connection from singapore :P but don't know if the are trying to log on..

gaudryc:
Do I miss something ?

gizmocuz:
closing #80, reopening #73, merge conflict... its getting a bit of a mess here ;)

gaudryc:
As you can see, I'm just a newbie with Git ;-)

gizmocuz:
So am i ;)

gizmocuz:
Please update to the latest code and review the changes

gaudryc:
It's done !

gizmocuz:
Thanks

Again questions:
- Why you need the mutex lock, did you ever experienced any problem before ?
  Running openzwave for more then a year, so do other, and no problems.
  Also use the same mutex method we use everywhere

i did not test the 'if (TotalValue >= 0)' behaviour, but again, if this works we dont need the 'if (TotalValue >= 0)' and this check can be removed, because it could never be below zero

gaudryc:
Because domoticz is never stable I cannot answer "no I never experienced
any kind of concurrency issue with Openzwave devices". I only want to avoid
them if the concurrency is possible. Race condition issues are very hard to
solve (eg. localtime_r).

About the 'if' I can remove it if you want.

Ps : Please take a look to my last bug report (Duplicated General/kWh
devices) in the forum, it is so strange. Thanks in advance.

gizmocuz:
So, as you tested and stated above, this does not solve anything?(hope you tested)

Strange things happens with beta versions, thats why we have beta testers
Think the duplicate issue is solved in the latest commit

gaudryc:
Are you agree that m_devices is shared between openzwave and mainworker threads, with read/write access for both of them ?
If yes, you need a mutex before using this shared resource, even if no issue has been occured yet.
If you do not want any mutex you have to rewrite openzwave and mainworker code to remove the dependency and to make the openzwave thread the only one that can read/write to this resource.

And I did the job in EventSystem, I will do the job everywhere else if necessary. But I cannot rewrite the architecture so I need use mutex.

gizmocuz:
for example, razberry, updatedevice you use a mutex, then call
CRazberry::FindDeviceByScale

also using the same mutex, but even this is not needed as this is only called by updatedevice

maybe you could split your patches, instead of one large big patch
For example logging when someone logs in, or the failure, i would like to apply right at once...

gaudryc:
One thread cannot lock itself using reentrant lock.
I'll take a look at this function and submit patches instead of one.

I close this request now to be able to create new ones.

PR #74, opened 2015-09-16
[1Wire] fix reporting when more than one unit is used

szczukot:


gizmocuz:
Thanks for the patch!

PR #76, opened 2015-09-16
Devices with plans

mikefranken:
Implementation of fetching devices with their real PlanID instead of 0
Facilitates clients that want to do something with plan without having to do multiple round trips to the server, fetching all plans and then all devices per plan
Does not change the API, just adds a PlanIDs to the resulting JSON:
image
May have performance impact, but probably limited

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,10 +1,7 @@
-<<<<<<< HEAD
 *.orig
 msbuild
 domoticz.db
 appversion.h
-
-=======
 Makefile
 CMakeFiles
 *.cmake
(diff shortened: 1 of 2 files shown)


gizmocuz:
Thanks for the patch, i also think the performance increase is very limited

mikefranken:
Thanks for the quick merge of pull request!

EddyK69:
This change causes a 'Debug Assertion Failed' on line 6715 in Webserver.cpp (std::string pID = result[0][0];)

See also
http://www.domoticz.com/forum/viewtopic.php?f=6&t=8060

Please fix or roll-back this change...

gizmocuz:
it seems this patch also causes problems when you have users that needs to login and only have a subset of devices, to not work. The users gets more devices, and the devices not in a plan will not be offered.

Maybe you could review this patch ? In the mean time a do a quick small roll-back

PR #83, opened 2015-09-18
Added pushbullet notifications

tofra:


gizmocuz:
Please don't modify the cmakelist.txt other then adding the file
Sqlhelper.cpp you do not have to increase the database version,
And it should not be enabled per defaut

gizmocuz:
Thanks, and many thanks for the patch as well !

tofra:
BTW: don't forget adding it to History :-)

gizmocuz:
i fill push a updated history file soon! You could have added it ;)

PR #84, opened 2015-09-19
MQTT: use floorplan and plan names as topic

xs4:
Outgoing messages will still be publised on topic 'domoticz/out'.
If the device is pressent on a plan in a floorplan the
message will also be published on topic:
    domoticz/out/'flooplan name'/'plan name'

For example:
    domoticz/out/Downstairs/Kitchen


gizmocuz:
This patch looks nice, but dont you think this will have a performance impact?
Now for every sensor received, the database is queried...

xs4:
I haven't measured it, but I don't think the extra database call will be noticeable.

xs4:
I added configuration options so the user can choose between topic publication schemes and avoid the extra database call. Does this address your concern about performance impact?

gizmocuz:
Yes, but quick question, the default topic for MQTT (mode1) will be PT_out for existing users?
or will this be PT_none, so current users will see no output anymore after this patch ?

Then i would propose you increase the database version number in SQLHelper.cpp and set the default output to PT_out for MQTT hardware

xs4:
- added database migration so current users see no change
- changed the default for new MQTT hardware from none to flat (aka topic domoticz/out)

gizmocuz:
Thanks!

PR #89, opened 2015-09-21
Force Energy/CMxxx or General/kWh device to display Value graph even if it has Usage

gaudryc:


gizmocuz:
i reverted this patch, as this caused the shortlog graph to display accumulated values in blocks of 1 hour, instead of 5 minutes

PR #91, opened 2015-09-21
Make read/write operations to be thread-safe on the map of ZWave devices

gaudryc:
In his last comment from PR#73, gizmocuz asked me : 
    "for example, razberry, updatedevice you use a mutex, then call CRazberry::FindDeviceByScale"
    "also using the same mutex, but even this is not needed as this is only called by updatedevice"

I use mutex each time a shared resource is read or written by many thread at the same time.
Both CRazberry::FindDeviceByScale and CRazberry::FindDeviceInstance methods are call by 
    CRazberry::UpdateDevice
        CRazberry::GetUpdates()
            ZWaveBase::Do_Work() (zwave hardware thread)
UpdateDevice does not lock anything before calling FindDeviceByScale and FindDeviceInstance before it does not use m_devices directly. Because m_devices can be written from another thread (mainworker) each iterator call must be protected by mutex.


gizmocuz:
in COpenZWave::OnZWaveNotification the first thing that does happen here, is using a mutex
there is no need for additional mutexes/locks
the mainworker does not write in m_devices (it is also private declared)

ZWaveBase::Do_Work() does not do anything with m_devices

this can happen with Razberry as this needs an http call each time.

as razberry is deprecated, and openzwave should be used, i see no reason to rebuild all just for this, do you agree ?

gaudryc:
Is the attribute ZWaveBase::m_devices called by the function ZWaveBase::FindDevice(..) ?
Is the function ZWaveBase::FindDevice(..) called by the function ZWaveBase::WriteToHardware() ?
Is the function ZWaveBase::WriteToHardware() called by the function MainWorker::PerformRealActionFromDomoticzClient() ?
Is the function MainWorker::PerformRealActionFromDomoticzClient() called by the function MainWorker::DecodeRXMessage() ?
How many threads consume the function MainWorker::DecodeRXMessage() ? 

ZWave thread has read/write access to its private attribute m_devices.
All threads calling the method DecodeRXMessage need read-only access to this attribute (using iterator).

gaudryc:
If you trace the current thread id in *:: WriteToHardware and MainWorker::DecodeRXMessage, you will see these functions are running on the caller thread even if you use signals to cut the compilation dependency.
Almost all hardware classes are not thread-safe at all.

gizmocuz:
there is also a mutex in decoderxmessage

about openzwave, m_devices is only updated when new values are arriving

gaudryc:
The mutex decodeRXMessageMutex locks every threads but one (it is not really efficient).
If this thread is calling ZWaveBase::WriteToHardware() and a zwave notification is coming from COpenZWave::OnZWaveNotification that needs to insert a new device in m_devices, you have a big problem. This insert invalidates the m_devices iterator used by "WriteToHardware".

gaudryc:
This kind of patch is necessary but it is quite complicated because we need to find all shared resource and protect them.

For example :
- the public attributes of the class CDomoticzHardwareBase (eg. m_bOutputLog, m_iLastSendNodeBatteryValue) should atomic operations (eg. boost::atomic<T> instead of T) ;
- CEnOceanESP2::m_id_base not atomic ;
- MySensorsBase::m_nodes is not accessed safetly from mainwarkor and Mysensor serial device thread ;
- ... not enough time to check all code today

At the opposite, C1Wire is a good example : no public attribute, no public method excepting the method WriteToHardware, no internal use of the parent class CDomoticzHardwareBase, uses mutexes when necessary... A right way.

Do you ever look for a solution that could use message queues in order to avoid locking all these threads ?
Do you check if the code is thread safe ?

gizmocuz:
going back on topic:

i looked at all places where m_devices is used. (disregarding razberry)

I think we are safe when 'm_NotificationMutex' is also used in ZWaveBase::WriteToHardware

do you agree? also keep in mind, when this project started (many moons ago) it was only using RFXCom

gizmocuz:
(open)zwave is now threadsafe (even the razberry...)

PR #92, opened 2015-09-22
Correct size to full fit and set the same as button on the screen.

ghost:


gizmocuz:
No added value, only makes the widgets slightly wider, and decreases user experience a bit

PR #93, opened 2015-09-22
Remove the attribute m_Firmware_Buffer from RFXComSerial class

gaudryc:
Remove the attribute m_Firmware_Buffer from RFXComSerial class in order to update RFX firmware using a local variable instead.


gaudryc:
Last commit (e5577e8) was needed because I'd forgotten it the first time (it had been commited from my office instead of my home).

gizmocuz:
just to verify (as i currently can not do this), you tested this a few times?

gaudryc:
Yes, 6 times.

gizmocuz:
Great, thanks!

PR #96, opened 2015-09-23
Update from Domoticz

Haasje:


Haasje:
Sorry. My bad. Pull request the wrong way around :-)

PR #99, opened 2015-09-25
Implemented: Logitech Media Server

EddyK69:
This plug-in adds support for Logitech Media Server.
After adding the hardware-settings (ip/port), all found players will be added as player-devices. Each player can be controlled with a remote (just like the Kodi-players)


gizmocuz:
Great patch! thanks!

EddyK69:
Thanx for merging! Now i have to learn to make less commits in a PR ;) GitHub still has some secrets for me...

PR #104, opened 2015-09-27
Fix kWh day graph parameter

gaudryc:
It fixes a HTML syntax error, i18n field name/value pairs.


gizmocuz:
i'm changing this by hand, as we can not update the i18n files directly (this is done with transifex... )

gaudryc:
I'm sorry for the inconvenient. I asked it to me !

PR #107, opened 2015-09-28
Fix HTML syntax error on "Power day grap" attribute

gaudryc:


gizmocuz:
Sorry, just fixed this a few minutes ago

PR #111, opened 2015-09-29
MQTT: add switchType in output message

xs4:
devices appearing on the 'lights' page have a switch type
ie dimmer, on/off, smoke detector, contact ...


gizmocuz:
I will add the switch type recognition

gizmocuz:
done

PR #117, opened 2015-10-05
Fix: Logitech Media Server: wrong status on other devices...

EddyK69:
...after switching LMS-player
Added support for Controller & Receiver


gizmocuz:
i dont see where the new variable "Status" is being used in the LightsController.js ?
Maybe this should not have been committed ?

EddyK69:
It's no new variable ;)
In fact,  it IS used (in line 1734/1735) but it never was set to "", so everytime a switch was swichted AFTER a media-player update, this switch got the same (=previous) state. By setting the initial state to "", this problem is fixed.

gizmocuz:
Thanks, no i see ! Thanks ;)

EddyK69:
Thank you!

PR #118, opened 2015-10-05
Fix a critical security issue on predictable Session IDs.

gaudryc:


gizmocuz:
Thanks.... 
I think now Cookie/Session ids work as long as domoticz is not restarted.
Could we make a (secure) cookie where the username is in, and use that for authentication if the datetime is still valid from the cookie ?

PR #130, opened 2015-10-08
Fix deadlocks between all threads calling the function MainWorker::DecodeRXMessage

gaudryc:
Fix deadlocks between the thread calling the function MainWorker::DecodeRXMessage and the threads associated with hardwares
which receive notification from another thread (eg. COpenZWave).


gizmocuz:
Looking good.!!

Only it is dangerous to use the RFBuf as a container to store the bytes,
as also General Devices are send, who know what is send to it later.

Another option is to store the bytes (with specified length) in a std::vector<uint8_t> or a std::string

In the new background worker, should we sleep (lets say 10ms) in the while loop to easy the system CPU ?
Instead of 'wait and pop' could we check under the sleep if the size of the queue is 0, then 'continue' else pop the first message and process?
You store a pointer to pHardware, would it be better to store the Hardware_ID, and use 'MainWorker::GetHardware(int HwdId)' later in the worker?
This way, you can be sure that the hardware is still there, and enabled

Should we limit the maximum items in the queue? (to protect from memory overflow?) like 200 messages ? Or you dont expect this to happen?

gaudryc:
I think can replace RFBuf with std::vector and pHardware with HwdId in the pushing method. It was only to keep the same arguments between input and output.

The 10ms sleep is there to easy the system CPU but because of wait_and_pop, it is not really not needed.

With the function wait_and_pop you can pop the next message immediately after push. The worker thread is blocked until the next message.

If we limit the queue size we take the risk of dropping message. If we limit the queue size we have to decide what is the good size, I think it could be difficult, don't you ?

gaudryc:
I cannot use MainWorker::GetHardware(int HwdId) to get pHardware from its id because  m_hardwaredevices does not contain CHardwareMonitor instance.

Refactoring have to be done to get CHardwareMonitor initialized as a normal hardware. If you want to avoid its deletion from m_hardwaredevices you should add some code into methods MainWorker::RemoveDomoticzHardware_(_) instead of WebServer, MainWorker...
A class "HardwareManager" should be created to hold the list m_hardwaredevices and all associated methods, and add the m_hardwareManager to MainWorker to be accessed from other classes.

gizmocuz:
True, the hardware monitor could be rewritten to be a normal hardware class (dont know why there was decided to not do this....), or we could also store the hardware type (and if system, then threat this differently)

the stuff happening in 

void CHardwareMonitor::Init()

could also been have done in SQLHelper.cpp, shall i rewrite it to be implemented as a normal hardware device ?
(Have to take care that with AddHardware it is only allowed to add it once)

gaudryc:
It would be better to make it normal, to be easier to read and maintain.
Less you add special case more your code is simple and safe.

gizmocuz:
working on it...

gizmocuz:
Done.... now we can use the HwdID

gizmocuz:
i would suggest to not wait_and_pop

while(...)
- Sleep 200ms
- Check if messages to process, in not, continue while loop
- Get All messages (pop all) returns a new vector with all messages
- process all messages
  - while (message to send)
    - send message
    - sleep 50ms (see below)
- continue loop

this way, we have a maximum of 200ms delay, and all commands are fired fast

another thing to change:
MainWorker::SwitchScene

this because there is a delay in sending the switch commands:

.. sleep_milliseconds(50);

thats why i suggest above to sleep 50ms before handling the next message
this is needed for rfxcom

gizmocuz:
Disregard my last command about 'sleep_milliseconds(50)'
maybe this is not needed, but better to check

gaudryc:
If you use wait_and_pop but no message is coming, the thread is definitively locked in the function wait_and_pop.
If a message is pushed in the queue, the thread is unlocked, the message is popped immediatly and the function wait_and_pop returns.

If you use timed_wait_and_pop with a 5 seconds timeout but no message is coming, the thread is locked for 5s in the function timed_wait_and_pop and returns nothing.
If a message is pushed in the queue before the 5 seconds timeout, thread is unlocked and the message is popped immediatly.
So if there no message at all, thread still can check m_stopserver every 5 seconds.

The function timed_wait_and_pop is better in our situation.

Any other sleep in the while loop is useless.
It is very fast : awake the thread when needed.

gizmocuz:
please consider my pseudo code above, it has no 5 second locks
you only need to make a new 'pop' function that pops all, and returns it, also it processes multiple messages
(this is how i use it in many projects, only i use a std::deque but that serves my usage better ;)

imagine a user with only switches, it could take 5 seconds to terminate the application

gaudryc:
Your code can be a solution but your code always do something even if there no message in the queue.
If a message is pushed while your solution is sleeping you can loose 200ms before popping the message.

What my solution is doing is simply to wait for the next message without loosing time to loop/sleep. Between each message my code only checks for m_stopserver.

gaudryc:
If you need a sleep before sending command to rfxcom, why don't you put this sleep in the RfxCom\* class ?
Could you please add a comment to the MainWorker::SwitchScene ... sleep_milliseconds(50); to know why it is necessary ?

gaudryc:
I have added some log into my code and I have seen that some messages are dropped (vector is empty).
I'm trying to fix this issue.

gaudryc:
If I use RBUF everything is OK.
If I use a vector<unsigned char> some vectors are popped as empty vectors, even if I use a full copy of the buffer.

gizmocuz:
should not be, but otherwise you can do it with a std::string as well
(bdw this code has a huge problem with shutting down (5 seconds), trying to avoid this, and a sleep is good in a thread ;)

gaudryc:
I'm trying std:string...

A sleep is good if you do something to avoid CPU consumption, but in this case if no message is pushed in the queue, the thread is blocked, it does not do anything but wait (sleeping forever until the timeout or a message is pushed).

gizmocuz:
i suggest to add in the queue class a function getAll() that returns a vector array (or queue array) with all items, and clears the current queue
to avoid a lock for 5 seconds when the application terminates/restarts (for example with an update), the sleep method is the best, and also to process multiple messages with the result from the above function

gizmocuz:
you could also insert the RFBUF directly in the structs variable instead of making a std::vector/string first...
rxMessage.rxCommand.insert()

gaudryc:
To unlock I can push a dummy message to the queue. It will unlock immediately.

gizmocuz:
in a thread? then the sleep method proposed upstairs will be better, nothing wrong with a small sleep?
With your function you also kinda sleep, making the sleep interval smaller (50ms) in the proposed method?

gizmocuz:
In either case, it could take some time before the message (queue) is processed.
In the webserver class where a switch/scene is activated we might have to sleep,
otherwise it will return, and the web gui will not have the correct state ?

gaudryc:
rxMessage.rxCommand = std::string(reinterpret_cast< const char\* >(pRXCommand));
...
It does not work, rxMessage.rxCommand is popped empty sometimes.

gizmocuz:
rxMessage.rxCommand,insert() ? it should not be, using this many times without problems...

gaudryc:
Why Appveyor uses Boost 1.58 instead of 1.59 ?
Why Tavis CI uses Boost 1.48 instead of 1.59 ?

gizmocuz:
domoticz (on windows) is not using 1.59, why use this atomic ?

gaudryc:
atomic is needed when you share attributes between threads, changing/reading attribute uses atomic operation, it is thread-safe.

gaudryc:
std::atomic<T> comes in C++ 11

gizmocuz:
we dont support Cxx11

gaudryc:
You support only C++98 ?

gizmocuz:
dont think that raspberry pi's standard image supports c++11, and therefore please dont use +11

gaudryc:
You are right but I can use boost::atomic<bool> and successfully compile and run under raspberry, while Appveyor and Travis CI can not.

gizmocuz:
cant build here as well... no c++11
and not going to install it only for this usage _g_ ;)

gaudryc:
What is your environment (os, g++, boost version) ?
I'm compiling under raspbian 7.8 with g++ 4.6.3 and boost 1.59.0.

gizmocuz:
if you look in the WindowsLibraries folder of domoticz (if you downloaded the archive) you will see domoticz for windows is using 1.58

gaudryc:
boost::atomic is in 1.58 also. So it does not compile in your environment ?

gizmocuz:
Don't know but our build system has a problem and we should not use boost if we can avoid it
Discussion is getting to big as well 
Please adopt the proposed changes above so we can merge and test so we might be able to put it in the upcoming release

gizmocuz:
But I am still worried that when you activate a switch from the Web gui and the queue is not triggered that the gui displays wrong state

gaudryc:
But why the queue will not be triggered ?

gizmocuz:
there could be a case, where the thread is handling an item from the queue (lets say 2 incoming temp sensors, from different hardware)
Currently you handle one item at the time (no getAll() function)
Now someone switches on the light from the web gui.
SwitchLight is called, and after that the web gui requests the current state of the light switch, and displays this.
it could be that the actual switch state is not yet updates with DecodeRXMessage ?

gaudryc:
Before using my queue you have used mutexes, mutexes serialize all calls of
the method DecodeRxMessage. So it does not change nothing. It is even
better because it does not block every body.

gizmocuz:
not quite, everything was waiting on another

gaudryc:
In my solution, all DecodeRxMessage callers are only blocked when pushing
message if RxMessage is popping a message because the queue is protected by
a mutex. All DecodeRxMessage callers can push a message while a message is
processed by the function ProcessRxMessage.

gaudryc:
I'm adding some code to :
- force the queue to unlock when stopping the server (by pushing an dummy message);
- manage the empty message returned when timeout occurred (because rxMessage.hardwareId = 0 when the function timed_wait_and_pop returns on timeout).

gizmocuz:
then i think you should also make a getAll() and process all messages from this
because this has a larger change of web gui failures (see above)
Or we could do a sleep (100ms) in the worker, getAll() and in the web commands, call sleep 200ms, so changes are better that an update is missed?

gaudryc:
Do you want to make a sleep(200ms) each time the Webserver thread calls the function DecodeRxMessage ? (sleep after calling the function DecodeRxMessage)

gizmocuz:
hmmm we need to investigate this first...

gaudryc:
I mean : Do you need the message to be processed before returning the http response ?

gizmocuz:
Only for SwitchLight/SwitchScene/Setpoint/Thermostat Modes,.... better to investigate which functions need this

gaudryc:
The time needed to pop/process a message is very short (~50ms in debug mode with log on each push and pop) so you can do a short sleep (200ms) after calling the DecodeRxMessage, but there is no guaranty on the time the message is processed because other threads push messages also.

If you really needed a "message response" it will be nice to wait until message is processed. Otherwise, you wait for the next refresh.

gaudryc:
I've committed the following :
- Force the queue to unlock while the server is shutting down ;
- Manage the empty message returned after the timeout occurred.

gaudryc:
I have found a way to wait until the message is processed (with timeout to avoid the calling thread from being definitively blocked).

gaudryc:
But to be efficient it would nice to determine which call of the method DecodeRxMessage really needs to wait for the message to be processed.
All other calls should be asynchronous.

To tell to the function DecodeRxMessage to wait, it would be nice to add an argument  to the function DecodeRxMessage or to define other methods : 
- PushRxMessage(pHardware, pRXCommand) : push message without waiting the message to be processed ;
- PushAndWaitRxMessage(pHardware, pRXCommand) : push message and wait for the message to be processed.

gaudryc:
Sorry to ask you this question again but, could you update boost on the Travis CI environment from 1.48 to 1.58 (the same version as in Appveyor) ? It would be so nice :-)

gaudryc:
Get it.
It works.

gizmocuz:
Nice changes, few comments:

In the queue processing thread, there is no need to check for empty messages a second time, as this is already taken care of when it is inserted in the queue

## this can be removed:
-       if ((rxMessage.hardwareId < 1) || (((const unsigned char *) &rxMessage.rxCommand) == NULL)) {
-           _log.Log(LOG_ERROR, "RxMessage: cannot process invalid message from hardware with id=%d, message is %s",
-                   rxMessage.hardwareId,
-                   (((const unsigned char *) &rxMessage.rxCommand) == NULL) ? "null" : "not null");
-           // cannot process message with invalid id or null message
-           continue;

## \+       }

Why you are adding a crc around it? this should not be needed, and just wastes come cpu

I see you are back on the RBUF variable again, this could go wrong as we also have other hardware that pushes things (like a general struct), and who knows in the feature....

You also do not have todo the memset to 0, you could store the length value (pRXCommand[0] + 1) if you need it somewhere, but i dont think you will

Mainworker functions that needs to wait until its message is handled:

MainWorker::SwitchLightInt
MainWorker::SetSetPoint
MainWorker::SetSetPointInt
MainWorker::UpdateDevice

gaudryc:
My code contains a lot of test to be able to debug. I will add a boolean to
disable some tests to be efficient but secure.
About RBUF, I'm not able to something else at this time. I have tried
everything with std::string or std::vector but nothing works : the allocate
memory is not garanty, why, I don't know.
I will take a look to these "general structs".

Did you take a look to Boost on Travis CI ?

gizmocuz:
i could set travis to boost 1.55, not 1.56, is that OK?

gaudryc:
Yes, I think.
I will see it when I will commit changes.

Thank you.

gaudryc:
How did you compile boost under Windows in your Appveyor environment ?
If I compile it under Windows 7 using VC14 by executing the file boost/bootstrap.bat, the file libboost_atomic-vc140-mt-1_58.lib exists.

gizmocuz:
I dont have build all libraries, as this gets to large otherwise
i think i will need to build this one

gizmocuz:
do you have a libboost_atomic-vc140-mt-s-1_58.lib
(note the mt-s-1)
please send this to me

gaudryc:
Oops, it is at my office ;-)

gizmocuz:
i will update windows to 1.59

gizmocuz:
Done.... build passed !

Instead of the debug boolean, could you use
#ifdef _DEBUG

gizmocuz:
instead of push and trigger, could the normal Decode function return a unique integer (like the crc or so), and could that be checked?
Or a variable that increases with each push to the queue, so we could wait until that value (or higher) is returned?

gaudryc:
I can return the unsigned long _tRxMessage.rxMessageIdx which is unique because I have used the incremented value of a bool::atomic<unsigned long> to define its value.
I can greate a getter method to return the last value of this bool::atomic<unsigned long>.
But if "lastvalue > rxMessageIdx" is true, it does not guaranty that rxMessageIdx has been processed.

To be sure the message is asynchronously processed, I think you have two solutions, after the message has been processed :

A- Save the messageIdx in a list (protected with one mutex) and check (with another thread) if the messageIdx is in this list and then remove it (otherwise it will create a memory leak). This solution asynchronous and makes all threads independent.

B- Make a callback method but it is asynchronous (you can't known when the callback occurs) also, and you have to be sure that the callback is short (execution time), thread-safe, it does not create a deadlock... You can use this callback method to post the rxMessageIdx in a list (Solution A) that is held by the thread that has called the DecodeRXMessage.

But what will you do with this information ?

I think most of the time you should fire and fire as all material devices... May be you can catch the state of the device on which you send a command but I think you should wait the confirmation of its state change.
That's why I think it is a bad idea to wait for the message to be processed. I made it for you but you should think twice before replacing the function DecodeRXMessage() with PushAndWaitRxMessage(). If you are sure the information you need will come back from another way, don't wait, just push, you will earn the execution time and simplicity in the code.

For example, in the function MainWorker::SetSetPoint you have :

if (pHardware->HwdType == HTYPE_EVOHOME_SCRIPT || pHardware->HwdType == HTYPE_EVOHOME_SERIAL) {
...
//the latency on the scripted solution is quite bad so it's good to see the update happening...ideally this would go to an 'updating' status (also useful to update database if we ever use this as a pure virtual device)
PushAndWaitRxMessage(pHardware, (const unsigned char*)&tsen);
}

It seems like you could call DecodeRXMessage when pHardware->HwdType == HTYPE_EVOHOME_SERIAL and call PushAndWaitRxMessage when pHardware->HwdType == HTYPE_EVOHOME_SCRIPT.
Why do you have latency ? Can it be reduced or fixed ?

So many questions, sorry ;-)

gizmocuz:
Added the 'concurrent_queue.h' file in Domoticz, and using it for MySensors at the moment in commit #c73f04b
Currently with std::vector<unsigned char> as queue item, and it is working without issues

next update i will replace it with std::string as this is easier in the MySensors case as it needs to send strings.

I also removed the atomic bool (stoprequested) as this is not needed, i just push an empty queue item, check if its empty, and exit the thread

Great class!, Next i will look at the changes related to this pull request

gizmocuz:
Patch merged, made a few changing i am going to commit soon.
Also works perfectly with a std::vector<unsigned char > for the messages
removed the atomic dependency (some where used in debug mode only), and some replaced with a volatile

gizmocuz:
New commit, in the mainworker there is now a define for DEBUG_RXQUEUE (disabled by default)
(did also some code cleanup in this commit, but the queue changes are in the mainworker.cpp)

gizmocuz:
There seems to be an issue when using remote domoticz instances, i already made a patch, not sure if you are happy with this...

http://www.domoticz.com/forum/viewtopic.php?f=6&t=9075&p=62882#p62882

gizmocuz:
And there is an issue in the base hardware class, that sets the name when a sensor is unknown

i am reverting this patch....

gaudryc:
Sorry for the delay.

Could you explain me what was the problem with remote domoticz ? With the hardware ID was not found ?
Could you explain me what was the problem with the base hardware class ?

gizmocuz:
Remote Domoticz -> hardware_id 8765 not found
base hardware class has a lot of functions to quickly add/send a sensor, if the sensor was unknown (a new sensor), a name was assigned
The default name was never set, because the sensor was still in the RXQueue
i could use pushandwait, but as 80% of the hardware uses these baseclass functions, the whole point of the queue was gone...
i received forum messages about the above... and experienced the last one yesterday....

gizmocuz:
for example
https://www.domoticz.com/forum/viewtopic.php?f=42&t=5027&p=63096#p63096
(last topics)

i forgot to add a mutex at the start of ::DecodeRXMessage
do we need this ? if not, it might also solve the problem (dont need a queue)

gaudryc:
I saw it last night.
You also have a problem in CMakeLists.txt, line 8.

Of course you need it. You are in a multithreaded environment and you have a lot of shared resources. If you have concurrent access on "not thread safe" resources, you will have unpredictable problems. But this mutex is a bad solution because you almost stop all thread each time these thread send a rx message.

I think we should use the rxqueue instead of using this mutex.

gizmocuz:
What is wrong on CMakeList.txt line 8? I accepted some pull requests

I was talking about the DecodeRXMessage function if this needed a mutex, where do you think it could go wrong?

Problem is, there are to many current issues with using the queue

gaudryc:
Using messages queue solves concurrency issues. And you still have concurrency issues.

You tell me you have issues using queue because you have some concept problems.
So why don't we solve these concept issues ?
Why don't we consider other domoticz instances as other devices ?
Why don't we consider to fix hardware name issue ?

2 solutions :
- If you don't want to solve anything then you need to put "    boost::lock_guardboost::mutex l(decodeRXMessageMutex);" at the beginning of the DecodeRXMessage function.
- If you want to solve some concept issues then you can give me all informations I need to find a solution.

PS: Do you really take a look at the CMakeList.txt file (line 8 : it talks about SQLITE in a LUA block) ?

gizmocuz:
I will fix the cmakelist file

I still doubting... i have it running without mutex on my main installation, getting also data from other domoticz instances, and it has a lot of sensors updating from various hardware/locations.
No problem at all

The (almost) only thing DecodeRXMessage does is translate the structure to a sql query
99% of domoticz uses sql queries when it needs data from a sensor, and in the sql-query function is already a mutex

But maybe i'm missing something.... i keep it running for longer

gizmocuz:
About the problems why i removed the queue (for now)... same as above
- Did not handle messages from remote domoticz instances, complained about not finding hardware with ID=8765 (was already patched before i removed the queue)
- Please have a look in the hardware base class, for example at the function that sends a temp sensor
  We provide an optional name, the name was never set (updated), because the previous command was put in the queue, and the queue was not processed

The last we can solve by calling "push and wait", but then 80% will use push-and-wait, and also the internal switching already uses push-and-wait, so if we every where push-and-wait, there is no reason for a queue

gaudryc:
Don't you play with the attribute MainWorker::m_LastDeviceName in the code involved in the call of the function DecodeRXMessage ? 

Do you know what happens if multiple calls are made of the function DecodeRXMessage at the same time using different devices or Web ?

If you didn't see any issue, I think you are just lucky.

gizmocuz:
could you replicate an issue ?

I don't 'play' (not a gamer), and for as i can see, this variable is used in sequence in the decoderx function

i am happy to add the mutex back (i placed it there years ago), but up till now, i see no problem

gizmocuz:
bdw, have you seen
https://www.domoticz.com/forum/viewtopic.php?f=6&t=9092

with basic auth, after logout, you are still logged in

gaudryc:
You should know I can't replicate a concurrency issue in a main installation.
To replicate it you have to wait for something to happen or you have to write some code to force the random.

gaudryc:
I try to make you conscious of the problem for many weeks.

I tell you play with an attribute because this attribute should not exist in a concurrency context but you still use it as if everything is executed sequentially or in only one thread.

Now you go back by adding the mutex but you still have this attribute which is a bad practice.

If you want to solve all your issues, you should think as if all callers of the function DecodeRXMessage were browsers, the RxMessage (or RxItem) was a request and the function DecodeRXMessage was a server to which you send requests.
The function DecodeRXMessage should have its own context. The attribute m_LastDeviceName should be in this context (not in the mainworker context).

You want to solve the "default name" issue, just add this string to the "request" and pass it to the function DecodeRXMessage, then pass this string to the database layer to use it when the entry is inserted into the DeviceStatus table.

gizmocuz:
True, i was thinking of doing this, but i will need some time for this
thats why i took it out now, sure i want it 'in' again... and will do

gizmocuz:
Already placed back the queue code, not used atm

gaudryc:
How do you feel it about the "default name" attribute ?

Do I add an "const std::string& defaultdevname" argument to the DecodeRXMessage or add a "BYTE[128] defaultdevname" to all _tRBUF structures ?

gizmocuz:
I was thinking about a 3rd parameter to the DecodeRXMessage function
some calls will not use 3rd argument, but if we check the m_LastDeviceName (if equals unknown, and the supplied 3d argument !empty(), then we can update the database with the new name
will also save a lot of code in the baseclass, and doing difficult stuff to make the ID...

gaudryc:
Something has been broken because the PushAndWaitRxMessage function does not work properly : pushing and processing are OK but the code is NOT WAITING anymore.
You need to use it to get a correct device status after sending an web command.

gizmocuz:
i am working on a patch (it will be a large one!)

gizmocuz:
patch committed, the queue is back again
(45 changed files  with 279 additions and 1,095 deletions)

i have seen a post here that 'PushAndWaitRxMessage' was not working? (not waiting), but it was deleted again, is it working correctly ?

gaudryc:
It was an false alarm. When I put the zwave stick in normal mode (not
debug) the swith command is sent but the status does not come back. So
strange, I have to let it in debug mode :-(

gaudryc:
Dont forget to make atomic attribute come back. Volatile cannot replace atomic operation in a multihreaded context. It is very important.

gaudryc:
Please dont delete m_decodeRXMessageMutex because you need it at the line 1660, before processing message directly (coming from other instances)

gizmocuz:
Added mutex on line 1660 (code committed)

gizmocuz:
At the moment 'rxMessageIdx' is only used for debugging, volatile not an issue here, m_stopRxMessageThread is only used while closing, so also no issue there
So i see no harm not implementing the atomic, also this requires the boost atomic library to be installed, compiled etc.. which could brake current compilations at existing users, and needs wiki updates etc etc

(and only one thread is using these 2... the mainworker)

gaudryc:
You make me afraid because (when debug is used) rxMessageIdx is used (updated) by all threads that call DecodeRXMessage. The DecodeRXMessage is executed IN OTHER threads, not in the mainworker, that is why we need mutexes before using all mainworker attributes (eg. m_LastDeviceName).

gizmocuz:
shall we remove rxMessageIdx ?

gizmocuz:
added a new patch, battery value is now also passed to the DecodeRX function, and removed the global battery value in the base hardware class

gaudryc:
I'd like to keep debug code, defensive code and thread-safe code.
I don't want to sacrifice quality and security to make things easy.

gizmocuz:
Please take a very close look where/how m_LastDeviceName is used.
There is absolutely no problem with this

gaudryc:
The m_LastDeviceName attribute owns to m_mainworker instance so it can be safely used by de mainworker thread.
But the m_LastDeviceName is now used by :
- the m_stopRxMessageThread thread (read/write operations).
- the tcpserver thread (read/write operations).

So you have concurrent access to this attribute : that is a not thread safe shared resource.

If you don't want to have any problem with this attribute : remove the m_LastDeviceName and put it as a local variable in the ProcessRXMessage function.

However, all this does not change our need to use "boost :: atomic".

gizmocuz:
Wrong, the tcpserver is handled with a mutex in DecodeRXMessage, or we should add the mutex IN ProcessRXMessage, and then it is handled correctly correct?

anyway, lets stop with this discussion about atomic needs, it is NOT needed at the moment

gaudryc:
Someone else than me should explain you what thread usage implies...

gaudryc:
Even if you think I'm wrong, you should remove the m_LastDeviceName and put it as a local variable in the ProcessRXMessage function.
If you don't do this, you can cross your fingers and wait for an issue to occur (coming from a trusted user who uses several domoticz instances).

However you can keep the m_decodeRXMessageMutex mutex before calling ProcessRXMessage in the DecodeRXMessage function in order to serialize messages coming from other domoticz instances.

Anyway, as I am wrong ...

gizmocuz:
OK I ask you now a third time to look where this variable is used and in what context.. If you can not see it... To bad... 
We close this topic now

gaudryc:
You need someone else to answer you

gizmocuz:
yes i know.. sorry last comment was a bit harsh... but i mean that variable is used only in the decode_xxxx functions, which is called from ProcessRXMessage, thats why i asked if the mutex should be in that function, as the only thread that is calling this directly is the TCPServer
Or we should support hardware_id 8765 good, so we do not need to call ProcessRXMessage directly for it

gaudryc:
If you want avoid problem with m_LastDeviceName, you need to :
- support hardware_id 8765 as an normal hardware (to avoid the direct call to ProcessRXMessage and have only one thread that call the ProcessRXMessage function) ;
- then remove the m_decodeRXMessageMutex mutex.

But if you really want to make things clean, you had to make things above and remove the m_LastDeviceName attribute to make the ProcessRXMessage function call stateless.

PR #136, opened 2015-10-11
Add check for Device Type

mikefranken:
to prevent marking a device and scene with same idx as duplicate


gizmocuz:
Thanks for the fix!, will make a new beta and announce this on the forum thread

PR #137, opened 2015-10-11
Fix incorrect string compare

mikefranken:
char\* compare replaced by std::string compare
added logging in Debug mode


mikefranken:
Thanks, C++ is a bit rusty, must be ~20yrs ago. Will dig out my Straustrup ;-)

PR #140, opened 2015-10-13
Corrected lost translation

ghost:
Becuase in polish is other: example Knots=Węzły, Waty, Ampery, Stopnie Celsjusza, Stopnie Faranheita

I added translation code to my domoticz and it work ok but with no translation becuase i dont know how add to domoticz-xx.json.gz and i can not add to Transifex.


gizmocuz:
I think everybody knows km/h , m/s ?
Coz if you want to have this translated, there are a lot more places where to translate this (domoticz.jz for instance)

ghost:
But in domoticz.jz i must add only to polish file or to all language files ? And in Polish add translated but in other add in english ?

PR #148, opened 2015-10-20
Fix long term authentication cookie (DRAFT)

gaudryc:
Allow user authentication based on cookie even if the server is restarted.

### DIFF
--- a/main/SQLHelper.cpp
+++ b/main/SQLHelper.cpp
@@ -562,6 +562,7 @@ const char *sqlCreateUserSessions =
 	" [Username] VARCHAR(100) NOT NULL,"
 	" [AuthToken] VARCHAR(100) UNIQUE NOT NULL,"
 	" [ExpirationDate] DATETIME NOT NULL,"
+	" [RemoteHost] VARCHAR(256) NOT NULL,"
 	" [LastUpdate] DATETIME DEFAULT(datetime('now', 'localtime')),"
 	" PRIMARY KEY([SessionID]));";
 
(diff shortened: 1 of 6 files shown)


gizmocuz:
Thanks a lot again!!

gizmocuz:
i had a crash when i restarted domoticz and go to the web page again (session.expire was not filled in and caused a crash)
fixed it, but please check
Also i made the log messages only to be displayed when compiled in debug mode (and _DEBUG is in te preprocessor settings)
Works great now!

gizmocuz:
Could you change one thing ?
Keep the sessions in a vector/map, so we do not have to query the database for every web request?
Or do you think this will not matter for speed ? (i think this might be...)

gizmocuz:
we have now many logs while in debug mode (hard to see anything)
could we add a remark (exclude) the line with

_log.Log(LOG_STATUS, "CheckAuthToken(%s_%s_%s) : user authenticated", session.id.c_str(), session.auth_token.c_str(), session.username.c_str());

(also in debug mode)

gaudryc:
I'm finishing development, I will test and I will comment these logs ;-)

gaudryc:
When you say you've got a crash, you have :
- logged in ;
- restart the server ;
- clicked some in the page before crashing ?

gizmocuz:
quit domoticz
quit the browser
start domoticz
started the browser and pointed to my domoticz instance

gizmocuz:
it retrieved the session from the database, but the session.expire was not filled in.
Made a patch, and it seems to be working now.... hope i did it right

gaudryc:
I have no crash, the user is logged in automatically on browser start up as expected.
Could you tell me where you think the problem is ?

gizmocuz:
please see my commit changes

gaudryc:
OK, I finally found it, I didn't see you had merge my dev before it is finished.
Whatever, you were right, "session.expires" must be updated before being saved.

PR #150, opened 2015-10-20
Add support forgeneral  sTypeTemperature

szczukot:


szczukot:
Hi
Can You merge this ?

Fantom

gizmocuz:
I am traveling at the moment, and need to check this (almost done)

gizmocuz:
I see we have a sTypeSystemTemp sensor, but i can not see any hardware class that is using this now.

i would like to:
rename sTypeTemperature to sTypeReserved1
rename sTypeSystemTemp to sTypeTemperature

But in your class, please use 'SendTempSensor'

Later we change this to a global sensor (after a release)

(SendTempSensor (and others) can be found in CDomoticzHardwareBase)

szczukot:
I use STypeTemperature in Satel Integra.
I cant find any other class that use this type. 
But many class using wrong type now (for example 1wire) and there also would be useful to change to sTypeTemperature.
I waiting for your changes

Fantom

gizmocuz:
it is not a wrong type, but it is just a place holder name.
In the end, i think everything should be the 'General' type, and we should not use specific hardware reference
All hardware classes (almost all) are using the general SendTempSensor), best to do this also

szczukot:
I don't want use SendTempSensor, because its use another inproperly for me type for device : sTypeTEMP5 (lacrose tx3 with humidity).
This sTypeTemperature is perfect for all devices which do not in RFXNames.cpp.
If you rename sTypeSystemTemp to sTypeTemperature its cool for me too. I may use this type.

Fantom

gizmocuz:
That should not be the case, only when there is also a Humidity sensor for your hardware, and i do not think there is.
So you should NOT get also a humidity (or temp+hum) sensor
did you check this ?

See MainWorker::decode_Temp
line 2790 (check if we already had a humidity for this device, if so, keep it!)

szczukot:
Now i have two systems : Satel Integra and 1wire
Satel Integra have temp sensor ATD100 - without humidity
1Wire have sensor DS18b20 - without humidity. Now its using sTypeTemp10
People using inproperly type devices because dont see any "universal" alternative. Almost all modules is used sTpeTemp5 or sTypeTEMP10 which is not true. Wee see bad subtype name etc.
For 1wire i see device "TFA 30.3133". 
sTypeTemperature is very good solution for any devices, that not exists on list RFXNAmes.cpp

gizmocuz:
Again, please read my above post.
Soon we drop all sub types, and just report 'Temperature'
Dont want to have 1000 sub types, thats soon (read... not now) all will be converted to 'general' types

gizmocuz:
Out of 1000ends of users, we have about 10 that complain about this...
They are true, but this is not top priority

szczukot:
I do not complained too - I just wanted to do well in the new module.
Dropp all subtypes is good idea for me, but maybe other people want see how subtype are

Fantom

gizmocuz:
I know (was not talking about you ;)
I am going to remove the subtypes, i do not want 1000 sub types, and next week 1001, and next week 1002 etc.... its not needed, maybe i keep it for the RFXTrx devices , or only in the devices tab

szczukot:
For maybe on this moment you can merge this code ?

gizmocuz:
No, please use SendTempSensor as almost all hardware classes do this (and need todo this)
Then we are compatible for the feature.
(Dont know if you use other sensor types in this class, but see if they are there in the base class as well)

szczukot:
Please close this pull request

PR #152, opened 2015-10-21
Fix long term authentication cookie (Part 2)

gaudryc:
This patch fixes several issues about long term authentication cookie.

### DIFF
--- a/main/WebServer.cpp
+++ b/main/WebServer.cpp
@@ -14848,7 +14848,7 @@ namespace http {
 		}
 
 		/**
-		 * Retrieve user session from store
+		 * Retrieve user session from store, without expiration date and remote host.
 		 */
 		const WebEmStoredSession CWebServer::GetSession(const std::string & sessionId) {
 			//_log.Log(LOG_STATUS, "SessionStore : get...");
@@ -14873,6 +14873,9 @@ namespace http {
 			return session;
 		}
 
+		/**
+		 * Save user session.
+		 */
 		void CWebServer::StoreSession(const WebEmStoredSession & session) {
 			//_log.Log(LOG_STATUS, "SessionStore : store...");
 			if (session.id.empty()) {
@@ -14907,6 +14910,9 @@ namespace http {
 			}
 		}
 
+		/**
+		 * Remove user session and expired sessions.
+		 */
 		void CWebServer::RemoveSession(const std::string & sessionId) {
 			//_log.Log(LOG_STATUS, "SessionStore : remove...");
 			if (sessionId.empty()) {
--- a/main/WebServer.cpp
+++ b/main/WebServer.cpp
@@ -14848,7 +14848,7 @@ namespace http {
 		}
 
 		/**
-		 * Retrieve user session from store, without expiration date and remote host.
+		 * Retrieve user session from store, without remote host.
 		 */
 		const WebEmStoredSession CWebServer::GetSession(const std::string & sessionId) {
 			//_log.Log(LOG_STATUS, "SessionStore : get...");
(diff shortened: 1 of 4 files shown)


gizmocuz:
thanks !
could you comment out the line
_log.Log(LOG_STATUS, "CheckAuthToken(%s_%s_%s) : user authenticated"

(I don't like to see this in debug at all, as I am running 90% in debug while developing)

gaudryc:
Job done

gizmocuz:
Great Thanks, I will merge it shortly to the master

gizmocuz:
Question, i have one user (.... the second user of domoticz!), that had a problem since 19 or 20 october.
His memory goes up, and after 2 days, the web interface does not respond anymore, but domoticz is still logging, controlling everything.

He has an iphone 24/7 open on the utility page
He runs monit
He also checks via a cloud service if his domoticz is up and running

We think this has to do with the latest changes.... do you know if this is possible ?

Is it possible that the session variable keeps adding new sessions, and eventually results in an out of memory ?
If there a way to avoid this ? Maybe flushing/emptying this 'map' ?

(Just thinking aloud here)

gaudryc:
How many times does he check per hour ?
Does he use the session cookie ?
Does he specify rememberme=true all the time ?
Does he login without logout ?

Yes it can.

We should remove all expired session (may be every 15 minutes ou hour).
We should reduce the one year expiration delay because the session never
expires if the year doesn't logout himself.

2015-10-27 18:22 GMT+01:00 gizmocuz notifications@github.com:

gaudryc:
We should manage :
- the session timeout (the session is deleted from the map but not from the
  database) ;
- the session expiration (the session is deleted from the database when the
  session expired).

If we make a thread to do this map cleanup, the map should be mutex
protected.

2015-10-27 23:19 GMT+01:00 Chris Gaudry chris.m.gaudry@gmail.com:

gizmocuz:
He used an online cloud service to monitor the up-status of domoticz via

http://uptimerobot.com/

Using this already for years, without issues, but since ~19 october the memory goes up, and then stops

The poll interval is 5 minutes, and the url that is checked is the SSL port of domoticz

https://domoticzip

gaudryc:
Do you mean "https://domoticzip" without any parameter nor any
authentication parameter ?

Do you known what he is doing with Monit ? He monitors only system or he
configures anything else that performs HTTP requests ?

2015-10-28 9:19 GMT+01:00 gizmocuz notifications@github.com:

gizmocuz:
Okey i checked the system.

Yes, the pure call, without authentication or parameters.
The cloud server needs to authenticate,but that the connection is working is enough for this cloud monitor. (So it will not authenticate)

Monit is running, but NOT checking domoticz, monit is checking the 'system' only

gizmocuz:
We changed the cloud server settings yesterday evening, the cloud will not check port '22' to see if the system is still alive.
From this moment, the memory is stable !!
So it must be definitely be the polling on https....

gaudryc:
I have created a monitor on Uptimerobot to check my domoticz server using
the URL  https://<domainname>:<port>/ but I don't see anything in the
uptime report.
I don't know how it works but I don't see anything even in the domoticz log
file (no incoming connection).

I will create a bash script to poll my https port using curl to see if
there is a memory leak.

2015-10-28 17:16 GMT+01:00 gizmocuz notifications@github.com:

gaudryc:
> I'm doing the following each minute in a script :

curl -i -k -X HEAD https://localhost:8443/

gizmocuz:
nice... if you watch the motherboard memory sensor in domoticz you should see the increase...(after hours...) at his situation it took 2 days, memory on his cubieboard was >30%, and then the web server stopped, domoticz kept on running (logs from temperature sensors can confirm this)

gaudryc:
I run the following test on my Raspberry Pi 2 (polling 600 times using 1
polling per 2 seconds is equivalent to 576 times using 1 polling per 5
seconds) :

echo "sudo /etc/init.d/domoticz.sh restart..."

sudo /etc/init.d/domoticz.sh restart > /dev/null

echo "sleep 180"

sleep 180

send_command.bash "type=devices&rid=15"

for i in {1..600..1}; do

```
echo "$(date) [$i] https polling"

curl -i -s -k -X HEAD https://localhost:8443/ -o /dev/null

sleep 2
```

done

send_command.bash "type=devices&rid=15"

echo "sleep 1200"

sleep 1200

send_command.bash "type=devices&rid=15"

The general memory usage :
- comes from 10,09% to 10,21%, without any polling and session cleanup ;
- comes from 9,99% to 10,30%, with polling but without session cleanup ;
  after sleeping 1200 seconds, it comes to 10,40%
- comes from 9,93% to 10,22%, with polling and session cleanup ; after
  sleeping 1200 seconds, it comes to 10,30%

I think we cannot say that there is a memory leak while polling the HTTPS
port because :
- the results are totally different from yours while the number of times
  the server is polled is the same ;
- we are measuring the system memory instead of the domoticz process memory.

I think you should ask him :
- to turn off his polling for 2 days to see the difference ;
- to tell you how the memory of the domoticz process evolves (from start
  up) ;
- to tell you what he is doing elsewhere with its server other than polling.

Whatever it is, we must clean up the sessions.

gizmocuz:
Is it possible to look again at this patch ?
There have been a few reports:
- When using a local script that calls the WWW JSON interface there will be hundreds of session entries in the database table (and maybe also in memory)
- There are quite a few reports of users that are unable to login, i experienced this myself when upgrading a user (version #3xxx to #4xxx), i had to restart domoticz from SSH to get it working again
- Other say they can just login locally, not from remote

gaudryc:
Could you give me some details :
- How do the local script call the WWW JSON interface ? Could you give me an example of a command (curl, wget...) that generates the problem ?
- What do you mean with "upgrading a user" ?
- How do they login remotly (which kind of browser) ?

gizmocuz:
Sorry, i am just trying to remember what i see on our forum posts.

#2, a friend of mine had a very old domoticz running, before the session patches, and i upgraded him, and i could not login anymore, had to restart domoticz, clear the browser cache

#1 and #3, i found some topics

Login/Authentication issues:
http://www.domoticz.com/forum/viewtopic.php?f=6&t=9050&p=63808&hilit=login#p63808
http://www.domoticz.com/forum/viewtopic.php?f=6&t=9204&p=64149&hilit=login#p64149
http://www.domoticz.com/forum/viewtopic.php?f=6&t=9321&p=64800&hilit=login#p64800
http://www.domoticz.com/forum/viewtopic.php?f=6&t=9460&p=65859&hilit=login#p65859
http://www.domoticz.com/forum/viewtopic.php?f=6&t=8869&p=61320&hilit=session#p61320

UserSessions (last post, not sure if this can be detected/fixed?)
http://www.domoticz.com/forum/viewtopic.php?f=17&t=4390&p=71104&hilit=UserSessions#p71104

gaudryc:
I'm still looking for an issue in login process but you should know that if you remove cookies with one year validity from the browser, theses cookies will be only deleted from the database one year later.
So, I told you already, we should shorten this validity duration.
Anyway, these cookies are neither kept in memory (cleaned every 15 minutes of inactivity) nor loaded into memory on startup.

gizmocuz:
A month is perfect for validity

tcellerier:
Hello,
Would it be possible to put back 1 year validity for the long term session cookie ?
I use domoticz on my computer, my phone as a web app, and my tablet (attached on my living room wall). 
It is very annoying to have to login on each of these devices every month since this update ...

PR #155, opened 2015-10-22
Fixed NL typo

EddyK69:


gizmocuz:
Unfortunately this can not be changed like this.
You have to translate this on
https://www.transifex.com/domoticz/
(or send me a PM)

EddyK69:
Ah ok. See post http://www.domoticz.com/forum/viewtopic.php?f=7&t=120

gizmocuz:
i also find this a strange translation.... but i will change it into this ;) thanks!

PR #156, opened 2015-10-22
Logitech Media Server: Add support for User Authentication on LMS server

EddyK69:
+Added LUA Favorites support


gizmocuz:
i think you can use

sURL << "http://" << sUsername << ":" << sPassword << "@" << sIP << ":" << iPort << "/jsonrpc.js";

there is no need to modify the http class?

EddyK69:
Ok, i will try that also. Thnx!
(Should i change this PR or create a new one?)

gizmocuz:
if the above is working, maybe a new pull request, and close this one

PR #158, opened 2015-10-24
Logitech Media Server: Add support for User Authentication on LMS server

EddyK69:
+Added LUA Favorites support


gizmocuz:
Thanks for the change!

Why are there 3 constructors of the class? This should only be one (including id/ip/port/username/password)

EddyK69:
They're indeed obsolete, leftovers of Kodi-class. I wil remove them in a new commit in this PR.
(Only 2 are used)

EddyK69:
Cleaned up code: removed unused constructors. Please merge ;)

gizmocuz:
Had to manually commit to resolve a conflict (cache version, my fault)
Thanks for the patch!

PR #159, opened 2015-10-25
Update the build status badges to point to the Appveyor and Travis-CI services

ToniA:
The build badge images are now links to Appveyor and Travis-CI

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,5 +26,5 @@ Some Information
 
 Status | Operating system
 ------------ | -------------
-![image](https://travis-ci.org/domoticz/domoticz.svg?branch=master "Linux Build Status") | Linux x86_64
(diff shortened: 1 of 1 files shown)


gizmocuz:
What was wrong with the previous readme ? I liked it this way much better... and compact as well
Also travis might build for some more platforms, and then we need to change it again
The current small indication is enough i suppose ?

gizmocuz:
(And i dont want to point people to travis/appveyor)

ToniA:
Well, I think it is useful to guide potential contributors also to see how the actual build process works. But it's your call anyway to decide.

gizmocuz:
I 'expect !!' that contributors test their code before committing  _g_

ToniA:
Sure. I just think that having the build badge image linking to the build it results from is more useful than linking to the image itself. But like I said, it's your call, feel free to not accept this if you don't like it.

gizmocuz:
Okey thanks ;) i close this pull (for now)

PR #162, opened 2015-10-25
Update download_update.sh

CyberClwn:
Correcting displayed message when failing on archive-file


gizmocuz:
Thanks for spotting the typo !

PR #165, opened 2015-10-26
Fixed Travis-CI FTPS URL, Appveyor to use an environment variable as FTP password

ToniA:
I hope this fixes the Travis-CI deployment, by adding the missing slash at the end of the FTP upload address.

For Appveyor, I also removed the encrypted password from the appveyor.yml. Please go to https://ci.appveyor.com/project/gizmocuz/domoticz/settings/environment and do 'Add variable' under 'Environment variables'. The variable name should be 'FTP_PASSWORD', and the value should be the FTP password. Click the lock icon 'Toggle variable encryption' to encrypt the password, and finally click 'Save'.

After this both Appveyor and Travis-CI will use FTP passwords which are set in the project, instead of the yml files, so the passwords are easier to change.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -92,7 +92,7 @@ matrix:
 deploy:
   skip_cleanup: true
   provider: script
-  script: curl --ftp-ssl -T domoticz_${TRAVIS_OS_NAME}_x86_64_latest.tgz -k -u "$FTP_USER:$FTP_PASSWORD" "ftp://$FTP_HOST/beta"
+  script: curl --ftp-ssl -T domoticz_${TRAVIS_OS_NAME}_x86_64_latest.tgz -k -u "$FTP_USER:$FTP_PASSWORD" "ftp://$FTP_HOST/beta/"
   on:
     branch: master
     repo: domoticz/domoticz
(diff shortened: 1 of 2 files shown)


gizmocuz:
The travis build system uploads in the beta folder ! Perfect
Only yhe OSX archive does not contain any sub folders, that is why the archive is 2MB instead of 9MB

The AppVeyor does not upload, it exit the build with the following lines:

Packaging artifacts...OK
Deploying using FTP provider
Neither artifacts nor application were uploaded.
Build success

ToniA:
I made a new pull request on the Appveyor. It helps if the artifact name actually matches with the name of the file the build generated (this was changed on your request, and I totally ignored the appveyor.yml artifact definition).

And I figured out the OSX problem... I looked for the 'tar' command from the 'makebeta' script, and thought that '--exclude .svn www/ scripts/ Config/' means that all of those directories need to be excluded, even though this means that '.svn' should be excluded (and it does not even exist any more). I'll make a new pull request on that.

gizmocuz:
there are no .svn files , we could remove the exclude ?

gizmocuz:
Yahooo... Appveyor now correctly uploads to the build folder!

PR #167, opened 2015-10-26
Fix domoticz.tar creation for OSX

ToniA:


gizmocuz:
tar creation for OSX working, thanks!

PR #169, opened 2015-10-26
Fix environment variables

ToniA:
This is getting embarrasing... Appveyor uses $(ENVVAR) and Travis-CI uses ${ENVVAR}, and guess who had those mixed up...


gizmocuz:
Only the big archive is uploaded, not the other files...

PR #171, opened 2015-10-27
serious bug on events ?

szczukot:


tomaszszczukobls:
And what do you think about nValueWording in this place ? 
it should be used function nValueToWording  ?

gizmocuz:
I am not that into the event system code, but you mean a rename of a variable ?
I agree that this sound more logical

tomaszszczukobls:
I see how it value set in this line : https://github.com/szczukot/domoticz/blob/de418ebfa2992f4c1ec9aef128fee67ac4e70488/main/EventSystem.cpp#L252
And  probably this methos should be used in place when i fix bug.

PR #173, opened 2015-10-28
Fix denial of service caused by stalled sessions

gaudryc:
What do you think about this solution ?


gizmocuz:
Looking good!, only one request, could you start member variables with m_

for instance these two:
mySessionStore
nextSessionCleanup

gaudryc:
I will change it from public nextSessionCleanup to private myNextSessionCleanup.

Are you agree ?

gizmocuz:
Thanks, merged

Issue #174, opened 2015-10-28
HEAD request must not return a message-body


gaudryc:
The HTTP/1.1 specifies :

"The HEAD method is identical to GET except that the server MUST NOT return a message-body in the response. The metainformation contained in the HTTP headers in response to a HEAD request SHOULD be identical to the information sent in response to a GET request. This method can be used for obtaining metainformation about the entity implied by the request without transferring the entity-body itself. This method is often used for testing hypertext links for validity, accessibility, and recent modification."

For example, Uptimerobot sends HEAD requests while checking http(s) requests...

If we fix it, we save some resources to serve these requests.

gizmocuz:
Great find!

rkluwen:
See my pull request.

gizmocuz:
thanks for the fix

PR #175, opened 2015-10-29
HEAD requests do not return response body anymore.

rkluwen:
In response of issue https://github.com/domoticz/domoticz/issues/174


gaudryc:
It seems to be OK for me.

gizmocuz:
Another nice addition :) thanks

PR #179, opened 2015-10-31
Fix for reloading totals of S0PCM5

StefanPuntNL:
The SQL query in ReloadLastTotals handled the device id as integer. The
energy counters are registered with a device id that contains leading
zero's so the query didn't return any device status. I also added an
extra filter on hardware type because of the fact that water and gas
counters are added as extra devices with the same hardware/device
combination (only without leading zero's).


gizmocuz:
Thanks, hope all is working now!

PR #180, opened 2015-11-01
Fix segmentation fault while cleaning timed out sessions

gaudryc:
This evening I've got a core dump while erasing session from the session map.
I don't know why I never had this before but I had to change the erasing method to avoid CPU overload and then a segmentation fault.

### DIFF
--- a/webserver/cWebem.cpp
+++ b/webserver/cWebem.cpp
@@ -941,14 +941,18 @@ void cWebem::CleanTimedOutSessions() {
 	time_t now = mytime(NULL);
 	if (myNextSessionCleanup >= now) {
 		myNextSessionCleanup = now + 15 * 60; // in 15 minutes
-
+		std::vector<std::string> ssids;
 		std::map<std::string, WebEmSession>::iterator itt;
 		for (itt=m_sessions.begin(); itt!=m_sessions.end(); ++itt) {
 			if (itt->second.timeout < now) {
-				m_sessions.erase(itt);
+				ssids.push_back(itt->second.id);
 			}
 		}
-
+		std::vector<std::string>::iterator ssitt;
+		for (ssitt=ssids.begin(); ssitt!=ssids.end(); ++ssitt) {
+			std::string ssid = *ssitt;
+			m_sessions.erase(ssid);
+		}
 	}
 }
 


gizmocuz:
One of our devs found a possible problem with the session cookies, and proposes the following patch:

diff --git a/webserver/cWebem.cpp b/webserver/cWebem.cpp
index 39ab431..86b0fc2 100644
--- a/webserver/cWebem.cpp
+++ b/webserver/cWebem.cpp
@@ -1317,8 +1317,8 @@ bool cWebemRequestHandler::CheckAuthentication(WebEmSession & session, const req
        // Parse session id and its expiration date
        std::string scookie = cookie_header;
        int fpos = scookie.find("SID=");
-       int upos = scookie.find("_");
-       int ppos = scookie.find(".");
-       int upos = scookie.find("_", fpos);
-       int ppos = scookie.find(".", upos);
      time_t now = mytime(NULL);
      if ((fpos != std::string::npos) && (upos != std::string::npos) && (ppos != std::string::npos))
      {
  @@ -1541,7 +1541,7 @@ void cWebemRequestHandler::handle_request(const request& req, reply& rep)
      {
          std::string scookie = cookie;
          int fpos = scookie.find("SID=");
-           int upos = scookie.find("_");
-           int upos = scookie.find("_", fpos);
          if ((fpos != std::string::npos) && (upos != std::string::npos))
          {
              std::string sSID = scookie.substr(fpos + 4, upos-fpos-4);

Is it OK to apply this ?

I also had some issues where i entered a valid username/password and got a session not found message, and it seems some forum users are also experiencing this.

rkluwen:
This patch is also in the MyDomoticz branch now. So you will get it once that pull request will be committed.
Explanation: If SID is the only cookie in the request, the old code works fine. But as soon as there's more cookies that possibly have an underscore or a period in it, things go wrong.

gizmocuz:
I have another trusted user with this problem, maybe we should add the patch in the master?

gaudryc:
I will take a look to this patch this morning...

gaudryc:
You want to add to the starting position argument ? 
You are absolutely right.
You can patch it as soon as possible.

gizmocuz:
Patch applied in
#define APPVERSION 3700
#define APPHASH "934e896"

PR #182, opened 2015-11-01
Travis-CI cross-compilation for Raspberry Pi

ToniA:
With this change Travis-CI will produce a third build, a cross-compilation for the Raspberry Pi target.

Shortcomings at the moment:
- OpenZWave build does not work yet. Or actually the build works, but it's not built correctly. If the comments are removed, the resulting 'domoticz' binary will complain about missing GLIBC
- WiringPi is not included
- I have done very limited testing so far, but this does start up on both Raspberry PI 1 and Raspberry Pi 2

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ before_install:
   # Domoticz needs the full history to be able to calculate the version string
   - git fetch --unshallow
   # OpenZWave
-  - git clone https://github.com/OpenZWave/open-zwave.git
+  - git clone --depth 1 https://github.com/OpenZWave/open-zwave.git
   - ln -s open-zwave open-zwave-read-only
 
 
@@ -40,7 +40,9 @@ matrix:
   - os: linux
     compiler:
       - gcc
-    env: TARGET_ARCHITECTURE=x86_64
+    env:
+      - "TARGET_ARCHITECTURE=x86_64"
+      - "SHACMD=sha256sum"
     addons:
       apt:
         sources:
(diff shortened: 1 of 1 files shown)


gizmocuz:
Hi Toni,

Thanks for the hard word, but i think i better use my build raspberry pi here for the automatic build process and upload.
Its 99% faster (on minimal changes), and all is working....

Now that we have this build systems, it is time we make use of it ;)

I have already a cloud server running

do you think we should move to a 'development' branch (and set this as default)

When it's time, we merge the development branch in the master branch, and then the file uploads will be made

ToniA:
This also tells you how to cross-compile for Raspberry Pi on your desktop machine. Travis-CI builds for Raspberry are a bit slow, as it will always pull the rather large Docker container. But locally it's really fast to always start a fresh Docker cross-compilation image.

So, are you going to merge this?

gizmocuz:
will this not make travis slow ? (like taking an hour before we can see our pull request builds?)
we also need to static link againt openzwave, with wiring pi, and static boost libraries

ToniA:
Yes, the build time goes up by ~10 minutes, mostly because it spends ~400 seconds in fetching the cross-compilation environment. I do know that OpenZWave and WiringPi are important, I just thought I need to start one bit at a time. The Boost libraries are static.

gizmocuz:
okey, so it is better then to not use travis for rpi builds, but a raspberry pi with a slight modified 'makebeta' script ? (that is checking if it needs to build (if the current revision is not the latest), and started from a cronjob every minute ?
(and upload to the same ftps of course...)

PR #187, opened 2015-11-07
OTGW: immediate update of Setpoint-values after setting them.

RobKuipers:
Special treatment of Room Setpoint, by reading value from 'PR=O'-command if thermostat-SetPoint is overriden (using 'TT'-command). Thus reflecting changes (almost) immediately.


gizmocuz:
Thanks!

mikefranken:
Nice!

PR #188, opened 2015-11-10
Added netatmo thermostat support

sbebrone:
I've added basic hardware support for Netatmo's thermostat. 
The code has been heavily inspired by the weather station and the Nest thermostat.


gizmocuz:
Thanks a lot !
Could you verify that you can use the same m_clientId and m_clientSecret as in the original netatmo class as we already are registrated by Netatmo (and on the support list)

As both classes are almost identical, i think i will merge them into one

Is it possible to send me a PM on the forum?

sbebrone:
I confirm that it works with existing keys, updated the PR with the commit.

Btw, it seems that I cannot send PM on the forum yet: "We are sorry, but you are not authorised to use this feature. You may have just registered here and may need to participate more to be able to use this feature."

gizmocuz:
Send you by accident an empty email..but you have my email now 😀

fredleroy:
Hi, I'm also very interested in this feature. If I can do anything to help...

gizmocuz:
Send me a PM on the forum with the login credentials

fredleroy:
I'd like to... But forum says I cannot send PMs cause I haven't been active enough...

jbfenris:
Hi, Thanks for the implementation !
Why do you have integrated your own Netatmo Client-Id and Client-Secret directly in the code ?
Do you  think it could be possible to put it as a parameters to set when we declare a new hardware ?

gizmocuz:
This is a domoticz wide clientid/secret, and is registrated @ netatmo, we are also in the official support list
So to answer your question, no this will not be allowed

gizmocuz:
Implemented, thanks for the support

PR #189, opened 2015-11-10
Fixed opening graph for a distance sensor from floorplan

djoele:
Added the following condition to function ShowGeneralGraph for sensortype DistanceGeneral:
|| (sensortype == "Distance")


gizmocuz:
Thanks!

PR #190, opened 2015-11-11
Implemented invoke scenes on variable change in FibaroLink

eburud:
This change will invoke LUA script on scene when Fibaro version 4+ is selected and variable changes on Fibaro Link settings. I think this can be invoked on all variable changes because this is the behavior users will expect. You can still avoid LUA execution if you don't define the variable in the header of the LUA code.


gizmocuz:
Thanks for the patch!

PR #194, opened 2015-11-14
Fixed the appearing of horizontal scrollbar when viewing a graph from foorplan device icon

djoele:
… floorplan device icon


gizmocuz:
Thanks! will be added in a next beta version!

PR #198, opened 2015-11-16
Update EnOceanESP3.cpp

matjoker:
The Enocean sensors type TempHum.01 and TempHum.02 were not recognised. szST is never equal to TempHum but to TempHum.01 or 02. I propose to replace "if szST == "TempHum" " by " if szST.find("TempHum")==0"


gizmocuz:
Thank you for the patch!

Issue #199, opened 2015-11-17
Apostrophe in zone name prevents adding device


dwmw2:
I'm using an Evohome system, and one of the zone names is "Foo's Bedroom".

In the device list, this ends up being rendered unescaped, thus:

```
<img onclick="SetUnused(7)" title="Set Unused" src="images/remove.png">
<img onclick="RenameDevice(7,'Heating','Foo's room')" title="Rename Device" src="images/rename.png">
```

I was able to add the zone by using Firebug and just adding the missing \ to make it

```
<img onclick="RenameDevice(7,'Heating','Foo\'s room')" …
```

We should do that correctly in all appropriate places in DevicesController.js

gizmocuz:
please look at the lights/utility controller how this is handled... maybe you could make a patch for the devicescontroller

dwmw2:
That didn't actually _fix_ it until I worked out how to manually clear my browser's app cache. Is that expected?

gizmocuz:
Yes, you need to increase the version number in the app-cache file as well
but i already did that, so in a next commit/beta this will work

PR #200, opened 2015-11-17
Escape item names in devices tab

dwmw2:
Fixes: #199


gizmocuz:
Thank you for the patch!

Issue #201, opened 2015-11-17
RFE: Dynamically link libstdc++, libgcc, libsqlite, liblua


dwmw2:
I'm building domoticz for OpenWRT. On a space-constrained system it's quite important that we use dynamically linked libraries rather than statically linking our own copy.  I may do Fedora packages, and we have similar packaging requirements there for different reasons.

Obviously I can fix this with out-of-tree patches (which is easier in some ways because it can be unconditional), but it would be nicer to fix it properly so the upstream build system allows building either way. Should I work on doing it 'properly', or is that not something that would be welcomed?

gizmocuz:
Is it then still possible that for example raspberry pi users can make a new install (sdcard image) and issue:

 mkdir domoticz
 cd domoticz
 wget http://domoticz.sourceforge.net/domoticz_linux_armv7l.tgz
 tar xvfz domoticz_linux_armv7l.tgz
 rm domoticz_linux_armv7l.tgz

to install domoticz?

Or do they now need to install dependency's ?

How about existing users with the above setup, trying to update to a new beta/stable version from the web interface, i think domoticz will just stop ?

Maybe it is possible to tweak the CMakeList.txt file to implement support for building with or without static build ? or make a #define for openWRT?

dwmw2:
> Maybe it is possible to tweak the CMakeList.txt file to implement support for building with or without static build ? 

That's absolutely possible, and is my preferred approach. I certainly wouldn't want to _break_ the static install that you describe, and it should probably remain the default.

The reason I asked first is because there was a possibility that you _(unwisely, IMHO)_ might have said _"no, I **always** want static builds and I don't even want to **allow** people to link dynamically"_. In that case I could have just done a hackish local patch to make it link dynamically.

You don't seem to be saying that, so I'll take the time to do it properly and submit a PR which allows for building with -DUSE_STATIC_SQLITE=NO, etc. Thanks for the prompt feedback.

(Note that the pull request I've just submitted doesn't cover this; it just does some other basic fixes which are prerequisites.)

dwmw2:
The build breakage for PR#203 with the MSVC build is noted; I'll fix it shortly (in case you get to looking at this before I fix it...)

dwmw2:
OpenWRT covers a lot of devices, many of them _extremely_ limited. Each library might only be a few hundred KiB but that's a few hundred KiB of flash _and_ RAM, for _every_ application which statically links just a few 'small' things...

[...]

PR #203, opened 2015-11-18
Dynamic libstdc++/libsqlite

dwmw2:
As discussed in Issue #201 here is a further pull request. It contains the basic fixes I already submitted which should be uncontroversial, and also it makes it **possible** _(but not mandatory, and not the default)_ to link dynamically against libstdc++/libgcc_s/libsqlite.

If this is acceptable, I'll make it work for lua and possible MQTT too.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -105,10 +105,10 @@ MACRO(ADD_PRECOMPILED_HEADER _targetName _input)
     LIST(APPEND _compiler_FLAGS ${_directory_flags})
 
     SEPARATE_ARGUMENTS(_compiler_FLAGS)
-    MESSAGE("${CMAKE_CXX_COMPILER} -DPCHCOMPILE ${_compiler_FLAGS} -x c++-header -o ${_output} ${_source}")
+    MESSAGE("${CMAKE_CXX_COMPILER} ${CMAKE_CXX_COMPILER_ARG1} -DPCHCOMPILE ${_compiler_FLAGS} -x c++-header -o ${_output} ${_source}")
     ADD_CUSTOM_COMMAND(
       OUTPUT ${_output}
-      COMMAND ${CMAKE_CXX_COMPILER} ${_compiler_FLAGS} -x c++-header -o ${_output} ${_source}
+      COMMAND ${CMAKE_CXX_COMPILER} ${CMAKE_CXX_COMPILER_ARG1} ${_compiler_FLAGS} -x c++-header -o ${_output} ${_source}
       DEPENDS ${_source} )
     ADD_CUSTOM_TARGET(${_targetName}_gch DEPENDS ${_output})
     ADD_DEPENDENCIES(${_targetName} ${_targetName}_gch)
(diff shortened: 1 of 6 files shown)


gizmocuz:
How small is your storage footprint ? i mean lua/sqlite/mqtt these are small..
boost/gcc/stdc i can imagine
I have to review this, as i am not good with cmake and it took me quite a while to build the script

dwmw2:
OpenWRT covers a lot of devices, many of them _extremely_ limited. Each library might only be a few hundred KiB but that's a few hundred KiB of RAM _and_ flash, for _every_ application that statically links just a few 'small' things...

In the case of Fedora packaging, the motivation is different — there, it's not about size. The packaging guidelines require that we do not ship static copies of libraries which are provided by the system, so that we can have more confidence that (especially security) bug-fixes in libraries will actually be effective everywhere.

dwmw2:
Note that the lua one needs lua 5.2 or greater, while OpenWRT has 5.1 _(and would be hard to upgrade)_. Looking at  the things which actually break in CEventSystem::CreateBlocklyLuaState() I'm a little confused about what's going on.

In that function we call luaL_openlibs() which initialises _all_ the standard modules, by _(internally)_ calling luaL_requiref() for them all. But before that, for some reason we call the 5.2-only luaL_requiref() function for just the "os" module. Which was going to be initalised anyway. And we call the init function of various others but _not_ through luaL_requiref(), which does something _different_ and it's not clear it isn't wrong.

Is there a reason that the call to luaL_openlibs() isn't sufficient? If I was going to play with cleaning that up, what specifically should I be testing to ensure I haven't broken stuff?

gizmocuz:
All with all, why wont you consider a raspberry pi 2 ? Its not expensive and can be used for domoticz only ?
And does not use much energy as well !

dwmw2:
Yeah, there's plenty of such hardware lying around already in fact. But there is a Geode box running OpenWRT already sitting in the cupboard under the stairs where the Honeywell HGI80 gets decent coverage of all the evohome traffic, and on a UPS. Adding yet another box to the system is not really my ideal solution. After all, domoticz on OpenWRT is basically already working; I'm just trying to clean it up a bit and package it properly — for others, and for my own sanity because it should Just Work on future upgrades.

gizmocuz:
Okey Fair _g_
I did not implemented Lua, and the original implementer is not active anymore as a developer.

I think you have to test is 'os' calls are still working, i also see you can use a #define 
http://lua-users.org/lists/lua-l/2012-07/msg00111.html

dwmw2:
Thanks. I've now submitted a package for inclusion in OpenWRT: https://github.com/openwrt/packages/pull/2007

I haven't fully investigated the LUA 5.1 compatibility issues so I've stuck with a static 5.2 for now.

I've also disabled the automatic updates in a local patch, since we'd need to make those a little more careful about precisely which system they're running on — the information from uname() isn't sufficient. Different distributions running on the same sysname/machine may be wildly different, including completely different C libraries and ABIs, etc.

gizmocuz:
Great! Hope all works out fine

PR #204, opened 2015-11-18
devices api enhancements

sbouchex:
Some "devices" api enhancements to prepare some integration with grafana. The returned json structure has been enhanced with the following details:
- "Sensor" : This is the sensor name used by the "graph" api to return some data. The possible sensor values are temp, percentage, wind, rain...
- "SensorNames" : it returns a comma separated list of sensor variable couple for a device. For example, a multi-sensor weather device will return "Temperature:te", "Humidity:hu", "Barometer:ba". Because it contains 3 sensors. Each couple is composed on 2 parts: The left part is the name of the sensor, the right part contains the variable name used in the json structure on the "graph" to store the data for each timestamp.

With these 2 changes, grafana will be able to retrieve the list of device and sensors with nice names, query domoticz in order to get sensor data and present the values into a nice dashboard.

Stay tuned !


gizmocuz:
This will make the GetJSonDevices routine more slower (it is already slow because some others wanted more information)
You do a lot of DB requests, while the Type and SubType are already queried, so whay not use this in the DetectSensorType routine ?
You do not have to query the database at all then

in RFXNames.cpp there is already a table with:
RFX_Type_SubType_Values

can't you use this ? Otherwise we have another place that we need to update when there is a new sensor type put in the system

empierre:
I agree on getting more information too, sure there are too much infos today, maybe removing useless could make things quicker ? there are too many special tags depending on device types, either you have a special flag or you have to parse data... not the easiest to interface.

slowness could be leveraged crating indexes:
CREATE INDEX IX_MM_ID_DATE ON MultiMeter(DeviceRowID, Date); 
CREATE INDEX IX_M_ID_DATE ON Meter(DeviceRowID, Date); 
CREATE INDEX IX_Temp_ID_DATE ON Temperature(DeviceRowID, Date); 
CREATE INDEX IX_Rain_ID_DATE ON Rain(DeviceRowID, Date); 
CREATE INDEX IX_Wind_ID_DATE ON Wind(DeviceRowID, Date); 
CREATE INDEX IX_LightingLog_ID_DATE ON LightingLog(DeviceRowID, Date); 
CREATE INDEX IX_Percentage_ID_DATE ON Percentage(DeviceRowID, Date); 
CREATE INDEX IX_CAL_MM_ID_DATE ON MultiMeter_Calendar(DeviceRowID, Date); 
CREATE INDEX IX_CAL_M_ID_DATE ON Meter_Calendar(DeviceRowID, Date); 
CREATE INDEX IX_CAL_Temp_ID_DATE ON Temperature_Calendar(DeviceRowID, Date); 
CREATE INDEX IX_CAL_Rain_ID_DATE ON Rain_Calendar(DeviceRowID, Date); 
CREATE INDEX IX_CAL_Wind_ID_DATE ON Wind_Calendar(DeviceRowID, Date); 
CREATE INDEX IX_CAL_Percentage_ID_DATE ON Percentage_Calendar(DeviceRowID, Date);

gizmocuz:
Currently we have index on those tables (on the DeviceRowID), why would you add one on the Date as well? (this is unique for each sensor, i mean lots of date entries)? (maybe i am not that good with indexes)

Some people would like to see only "Temp Sensor", others (with a RFXCom) might want to see the type
Also internally we have to send out the commands with different types, so types have to stay

But still, your new function does not need to query the database, and maybe you could use the Type/Subtype field, or the other table proposed

sbouchex:
I will modify the routine to rely on the tables in RFXNames.cpp but I will have to create a new one since the type of sensor is not stored somewhere.

empierre:
I already made those (painfull) mappings where I would have to query back
the device to know it is a tmp, a hum, a temp+hum, temp+hum+baro... any
combination that makes it so hard to map (or not). I also need to do a
query just to know information on the device that slows eveything back and
forth...

I am also wondering how to map together a switch and its energy...

sbouchex:
I have made the changes to remove the queries, and made some array static const to avoid to rebuild them at every call

empierre:
thanks a lot Sébastien !

the PI A/B are so slow (and surely the synology to), all will benefit from
that !

sbouchex:
I have seen a lot of possible optimization across the code (like passing const ref instead of the whole object), I will make a PR sometimes

gizmocuz:
Thanks !

gizmocuz:
As i want to move soon to also a general description, i thought making use of this table as well
but it is not accurate:

{ pTypeTEMP_HUM, sTypeTH1, "temp" },

these types are Temp+Hum

{ pTypeTEMP_HUM_BARO, sTypeTHB1, "temp" },

this is a temp+hum+baro

a lot of sensors does not seem right, as for instance some rain sensors have temp as well

This is all for the influxdb right ?

on the standard sdcard image, nodeJS and influx are installed and there is an example that converts data from domoticz into influxdb,
i think you should make the conversion here correct?

i will adopt your 'static' changes

ps. you don't need
#include <boost/algorithm/string/split.hpp>

empierre:
@gizmocuz do us all a favor: put a nodeJS that works, the raspian one has a
serious issue and there is no official support, so tne nodeJS can work fine
on armv7+ but segfault on armv6 (Rpi A/B+)...

the right solution is this one: to have the one from armv6 nodejs website:

@empierre https://github.com/empierre: ARM is not an officially supported
architecture for v0.10 or v0.12, also if you're using any armhf builds from
standard Debian repos you're unlikely to have much luck on a B/B+ as it's
strictly ARMv6 (which is a primary reason why Raspbian is a thing, Debian
does ARMv7+ only). I suggest you switch to Node.js v4 or v5 where ARMv6+ is
officially supported and instead of grabbing them from deb.nodesource.com
(which ships Debian style armhf for ARMv7+) you should get them from
https://nodejs.org/en/download/ where you can choose ARMv6 binaries.

I've spent 2 days searching for one working... and the one from nodejs has
support and is a 4.2.2

sbouchex:
No, the "temp" word in the returned structured should be a single word. In fact, it is not for influxdb but to another cool project: Give a direct access to Domoticz from Grafana by developing a Grafana  datatsource (Javascript based) that does REST query to Domoticz and I need some enhancements in this "devices" api.

The idea is simple:
- Call N°1 : "devices" api to list the devices and sensors
- Call N°2 : "graph" api to retrieve the data and to call this one, I need to fill the "sensor" parameter used in RType_HandleGraph with the sensor type. For a TEMP+HUMIDITY+BAROMETER, the sensor is correctly "temp" and it will returned 3 json entry with the value.

If you think that pTypeTEMP_HUM / sTypeTH1 returns humidy as well, find just modify RType_HandleGraph to return "h" as well. Same for pTypeTEMP_HUM_BARO / sTypeTHB1

gizmocuz:
@empierre 
maybe we can discuss this over email.
Did not know this, strange, as the domoticz code is armv7 and runs perfectly on old systems
does the armv6 nodejs run on a armv7 ?
And... i dont think you want to run nodejs at all on such a slow device right ?
the rpi2 is 6x faster!

@sbouchex
grafana is nice, had it running
in your java based call that issues the rest call, cant you make the translation there? (from the type/subtype)
Because i need to make a kind of general table, and now for each new sensor i have to add it in a lot of tables
it is to much work

what i want to do is not display that you got type tHumBaro1234 with Subtype=BrandNameXYZ, just TempHumBaro

this new table returns temp... nice... but for your project only, and... it sounds like a cool project!
Only i would for now do the translation in your java code, and later you can maybe use my new description
Hope you understand ?

sbouchex:
OK, let's forget about these changes. I will wait until you have them. I will hardcode the "temp","fan","uv" codes

Issue #197, opened 2015-11-15
SMTP: Unable to send mail


mverleun:
I'm trying to use email for notifications and use a OS X mailserver which is a Postfix server.
The OS X server refuses to accept email with the following error message: 
Nov 15 22:38:10 mac-mini.example.com postfix/smtpd[33394]: NOQUEUE: reject: RCPT from unknown[192.168.15.166]: 504 5.5.2 <domoticz>: Help command rejected: need fully-qualified hostname; from=marco@example.com to=marco@example.com proto=ESMTP helo=<domoticz>

The hostname is set correctly and matches the FQDN.
Is it possible to change the code to use the FQDN?

sbouchex:
Your problem does not seem to be an issue in Domoticz but rather a configuration issue on Postfix. Are you sure that postfix does accept FQDN in the 'helo' command ? Check your config and eventually do a manual test on postfix with telnet on port 25.

mverleun:
Hello Sébastien,

Thanks for your reply.

I’m certain that my mailserver is accepting FQDN’s. (I’ve received your email. ;-) ). Telnet shows the following:

# telnet mac-mini 25

Trying 192.168.15.2...
Connected to mac-mini.marcoach.nl.
Escape character is '^]'.
220 mac-mini.marcoach.nl ESMTP Postfix
HELO client.example.com
250 mac-mini.marcoach.nl
MAIL from: marco@marcoach.nl
250 2.1.0 Ok
RCPT to: marco@marcoach.nl
250 2.1.5 Ok
DATA
354 End data with <CR><LF>.<CR><LF>
From: marco@marcoach.nl
To: marco@marcoach.nl
Subject: Test

This is a test
.
250 2.0.0 Ok: queued as D7EC629FF7BD

Again, this time without a FQDN in the helo command:

# telnet mac-mini 25

Trying 192.168.15.2...
Connected to mac-mini.marcoach.nl.
Escape character is '^]'.
220 mac-mini.marcoach.nl ESMTP Postfix
HELO client
250 mac-mini.marcoach.nl
MAIL from: marco@marcoach.nl
250 2.1.0 Ok
RCPT to: marco@marcoach.nl
504 5.5.2 <client>: Helo command rejected: need fully-qualified hostname

These are the messages I find in the logs of my mailserver when Domoticz tries to send mail. I can’t change the postfix setup on a OS X system without breaking the system for future updates.

The FQDN problem is something that I noticed with certain HP devices as well.

[...]

PR #205, opened 2015-11-19
Fixed an issue with smtp client (missing FQDN client name)

sbouchex:
Fixed issue with the smtp client name (see issue #197)


gizmocuz:
strange.... 

could you verify if it also works with only this:

sstr << "smtp://" << m_Server << ":" << m_Port

(nothing behind it)

sbouchex:
You are right, according to the documentation of curl, it should work. I'm committing the change soon

gizmocuz:
Thanks!

PR #207, opened 2015-11-19
Add support for Comm5 MA-5XXX Relay and Sensors hardware

correa:


gizmocuz:
Excellent patch !
Do you have some more information about this hardware? (link for example)
The only thing i do not see is a reconnect if the TCP connection drops?

correa:
Hi gizmocuz,

Here is the link: http://www.comm5.com.br/novo/produtos-categoria/modulo-de-acionamento-i-o-rede-tcp-ip-72/ma-5000-2-255
Unfortunaly the link is in portuguese, but I think there is some documentation available in English.

Thanks for taking a look at it so fast. Having some issues with the linux CI not supporting C++11, working around it :)
From what I understand of AsyncTCP it handles the reconnection. Some empirical tests I did with the hardware also support that it does.

gizmocuz:
there is a problem with the build system (see above)

gizmocuz:
Okey, what is the difference between the
ETH8020

and the 

KMTronic

we already support ?

As you said, the documentation is in Portuguese , it does not look like the majority is using this hardware?

We have to consider standard hardware.... not hardware that a few users will use

correa:
They sure are similar in purpose, but this one is available in the South America, it does a big difference in costs when the absurd third world import taxes don't apply.

This hardware is used in many Digifort [1] and D-Guard[2] deployments in Brazil. I don't know of domoticz deployments outside my own, but I could argue that most users are not C++ developers, therefore they only use what is available.

This hardware is standard in our local market, ETH8020 and KMTronic on the other hand isn't.

Anyway, you can always demand me to keep this up to date.  :)

[1] http://www.digifort.com/
[2] http://www.seventh.com.br/en/

gizmocuz:
Yes i will demand this ;)
going to merge this soon, thanks for the great work!

PR #209, opened 2015-11-20
Netatmo credentials cannot be set anymore

sbouchex:
Renaming has not been done everywhere...


gizmocuz:
The fix should be in the hardware setup controller, not in the c++ code
i will have a look

sbouchex:
The fix can be at different places depending on the name of the hardware device you want to see. My fix displays the name "Netatmo Weather Station" instead of "Netatmo" but the controller (HardwareController.js) is detecting "Netatmo Weather Station"

gizmocuz:
True, but as we now also support the Netatmo Thermostat (users dont have to have a weather station) 'Netatmo' was a good name

gizmocuz:
fixed in #3619

PR #210, opened 2015-11-21
Floorplan undefined message

kerkenit:
When the floorplan wasn't completed, there was an error message in the console.


gizmocuz:
Thanks for the patch!

PR #211, opened 2015-11-21
CDomoticzHardwareBase cleanup

correa:
Simplify a little bit adding new hardware, so it doesn't have to change DomoticzHardware.h by setting the appropriate access specification for the class.


gizmocuz:
Nice, thanks!

PR #215, opened 2015-11-24
Update PhilipsHue.cpp to support OSRAM Lightify lights

arjanwaard:
Added  light type "Color temperature light" (OSRAM Lightify) as a simple dimmable light. Setting the Color Temperature is not supported yet.


empierre:
hello,

  have we there the full HSE also available in the JSON ?

gizmocuz:
you have RGB values that are translated to hue, please do not ask questions that are not related to this topic

PR #218, opened 2015-11-24
Change NaN value to NULL for JSON

PA4WD:
Database can give NaN values
JSON does support NaN only NULL.
With a NaN in JSON, highcharts gives no graph(JSON can not be parsed).
With a NULL only a blank value in the graph.


gizmocuz:
how can this happens, who has placed a nan value in the database?

PR #219, opened 2015-11-24
Added parameters to log line, in case a script is executed (for examp…

djoele:
…le when switching a light)

### DIFF
--- a/main/SQLHelper.cpp
+++ b/main/SQLHelper.cpp
@@ -2074,7 +2074,7 @@ void CSQLHelper::Do_Work()
 			else if (itt->_ItemType == TITEM_EXECUTE_SCRIPT)
 			{
 				//start script
-				_log.Log(LOG_STATUS, "Executing script: %s", itt->_ID.c_str());
+				_log.Log(LOG_STATUS, "Executing script: %s %s", itt->_ID.c_str(),itt->_sValue.c_str());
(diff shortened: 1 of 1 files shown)


gizmocuz:
That does not seem a good idea as sValue is used for various things.
(and can also be private)

the script ID (a string/name) is already displayed

djoele:
Alright, then it is best to close this pull request.

gizmocuz:
if something is not clear at all with the 'ID' that is logged, maybe best to change it on those places.
But from what i see, this is already done very good

djoele:
Agree to keep the log the way it is.
Already was able to see that the scripts that I want to execute with On/Off action are executed the right way.

---

Perhaps it is better if I describe what I wanted to achieve:
1. I have configured a switch with an On and Off action
2. The On action for example is: script:///home/pi/domoticz/swith.sh 1234
3. The Off action for example is: script:///home/pi/domoticz/swith.sh 4321
4. This shell script takes one argument: the code to send (so in this case, 1234 or 4321)

Now the following appears in the log file for either action: 
'Executing script: /home/pi/domoticz/swith.sh'

To be able to figure out exactly what was trigger, it would be good to have the following log lines:
'Executing script: /home/pi/domoticz/swith.sh 1234'
'Executing script: /home/pi/domoticz/swith.sh 4321'

PR #224, opened 2015-11-26
Update EnOceanESP3.cpp

matjoker:
For the enocean Temperature and temperature/humidity sensors, there's a confusion between min and max values for Scale and Range datas
For the temperature sensor, all the values are inverted, so the result was ok, but for the temp/hum sensor, only the scale values were inverted, so the given temperature was false.

=> Temperature: Range min: 255 / Range max: 0 - Scale min : lowest value / Scale max: highest value
=> Temp / Hum: Range min: 0 / Range max: 250 - Scale min : lowest value / Scale max: highest value

(From EnOCean Equipment profiles (EEP) V2.6.3)


gizmocuz:
Thanks

PR #226, opened 2015-11-26
Include internal zlib as build option and use standard Find_Package for OpenSSL.

correa:


gizmocuz:
question, why you have now

+if(NOT OPENSSL_SSL_LIBRARY)

OPENSSL_SSL ? _LIBRARY

correa:
Old code was only checking ssl library, not crypto library.
Will change to OPENSSL_FOUND which means both libraries, since even in windows we provide both libs, can't find one and not the other I guess.

gizmocuz:
Question, do you need

include_directories(${Boost_INCLUDE_DIRS})

On my raspberry/cubieboard/ubuntu this is all found from the standard include folders

correa:
You are supposed to [1]. On linux and mac it probably end up in one of the standard include folders but suppose you want to build it against a different boost with -DBOOST_ROOT=~/myboost
Or if boost isn't installed in a standard location, then it would be necessary.

[1] https://cmake.org/cmake/help/v3.0/module/FindBoost.html

gizmocuz:
Okey makes sence... merged , thanks !

PR #229, opened 2015-11-27
Fix dologout command response to return {"status":"OK", "title":"Logout"} instead of {"status":"ERR"}.

gaudryc:


gizmocuz:
Thanks!

PR #230, opened 2015-11-27
Applamp bug fixes

sijones:
Hi,

Some bug fixes and experimental code that needs wider testing, I've enabled the brightness and disco controls etc to be used on either RGB and RGBW devices as they seem to be supported.

I have introduced a bug that it displays the disco controls on individual channels of the controller which won't work at moment but looking in to enabling it.

Will keep adding some more fixes.


gizmocuz:
Thanks, let us know when you solved the introduced bug

sijones:
With this am struggling a little, I can see that in the light controller.js i know the type of lamp, so white, rgb,rgbw, but I can't see how i can see the group. I need to be able to hide certain controls on the group/zone 1 to 4, but on the All group allow operations. How can I do this? is the data passed in the js page?

sijones:
Now fixed the display bug I introduced, and fixed a few other issues. Can be committed.

gizmocuz:
Thanks for the patch !

PR #236, opened 2015-11-30
cygwin_fixes

JohnAdders:
Fixes to enable compilation under cygwin


gizmocuz:
Thank you for this patch.
But.... why would you like to compile with/under cygwin ?
You can compile with visual studio 2015 (libraries need to be downloaded separate, see install.txt file)

JohnAdders:
Understood, cygwin is very useful as it's installable on machines without admin rights as well as being an easy way to use the standard unix toolset on windows.
It would help with my workflow for adding support for a new meter and changes to the pinger if I can use cygwin.

correa:
About your changeset "quieten cmake when building in cygwin", perhaps we could start requiring cmake 2.8.11?

That's a quite old version already. Current CMake is at 4.x
Would there be any objections to a cmake version bump?

gizmocuz:
My raspberry (raspbian) has version 2.8.9
My ubuntu has version 2.8.12.2

not sure if we can bump the version

then all users need to uninstall the apt-get version and compile cmake from source

dont know why cmake is so old on raspbian, maybe this can be updated?

correa:
Just checked raspian apt-get repository on their site. 2.8.9 is the latest they offer there. Newer than that only if the user downloads from cmake.org.

2.8.9 should be new enough to not require the change, according to comment in the patch, it is required if lower than 2.8.4

Could be our baseline, and it would be nice to keep a comment in there mentioning the pi :)

gizmocuz:
Okey, then we should bump the requirement to 2.8.9

gizmocuz:
@JohnAdders, can you do this ?

JohnAdders:
On the basis that it makes sense to use the lowest version possible I've just upped it to 2.8.4, also I had to move  the project line to make the warning go away completely.  Also my env only has 2.8.7 so would prefer to keep required version low.

PR #238, opened 2015-11-30
Added support for iPhone 6 and iPhone 6 Plus

kerkenit:
In this changes i've tried to add a splash screen for the iPhone 6. I've found a lot info on GitHub of media queries.
Apparently it isn't working on iOS 9 because there a bug. It's possible that it's gonna work.
I've tested in in the simulator of iOS 8.4 of an iPhone 6 and an iPhone 6 Plus and it just work.

Feel free to change the resources, because i've put this together in Photoshop all by myself.


gizmocuz:
thanks for the patch, it seems www/images/startup/apple-touch-startup-image-750x1294.png has a problem loading if i go to 'files changes', not sure if this is actually true

another few megabytes to the appcache....

do you know if it is possible to bring all down to 1 image, and have that scaled on all ios devices?

kerkenit:
The images are in there, i hope. I'm new to Git(Hub). I'm familiar wit SVN. My git has problems. I've a lot of versions to manage. As I can see the images are in there.

In response to your question, the answer is no. Apple made for every screen size a specific splash screen. If the pixels are not matching, it wont show.
Android is much flexibel. You can see. Apple has a few devices. Android little.

gizmocuz:
could it be you introduced one typo ?

meta name="apple-mobile-web-app-status-bar-style" content="black"

missing the slash at the end (it seems i can not add < or > in this message)

kerkenit:
@gizmocuz Could to check this out?
https://github.com/kerkenit/domoticz/commit/d1744fa8c8d4b702e937f6c774d114dea8413752

gizmocuz:
i think you have the wrong image sizes for the 6/6+

http://docs.feedhenry.com/v2/icons_and_splashscreens.html

there is no splash screen showing on my 6s+

kerkenit:
I guess you got iOS 9.x
As I mentioned in the first commit, it's a bug in iOS 9.
https://forums.developer.apple.com/thread/23924
I downloaded iOS 9.2 beta 4 and it's not working eighter. 
In iOS 8.4 it's working in.
On my iPhone 6 it isn't working. Right now. I hope it will work in a future release.

Please feel free to change the image size (Mb's)

gizmocuz:
posted a reply on the forum as well... hope they going to take a look at it soon

Issue #241, opened 2015-12-01
TCPServer crash or TCPServer stopping loop  while changing RemoteSharedPort from Settings


gaudryc:
I'm using domoticz r3735. I have only one domoticz instance.

I have a crash (SIG_ABORT) each time I try to change the remote shared port from Settings Tab  : 

Core was generated by `/home/pi/domoticz/domoticz -daemon -www 8081 -sslwww 8443 -sslcert /home/pi/ssl'.
Program terminated with signal 6, Aborted.
#0  0x76b7f8dc in __GI_raise (sig=6) at ../nptl/sysdeps/unix/sysv/linux/raise.c:67

67  ../nptl/sysdeps/unix/sysv/linux/raise.c: Aucun fichier ou dossier de ce type.
(gdb) bt full
#0  0x76b7f8dc in __GI_raise (sig=6) at ../nptl/sysdeps/unix/sysv/linux/raise.c:67

```
    _sys_result = <optimized out>
    pd = <optimized out>
    pid = <optimized out>
    selftid = 2898
```
#1  0x76b8365c in __GI_abort () at abort.c:92

```
    save_stage = 2
    act = {__sigaction_handler = {sa_handler = 0x143194, sa_sigaction = 0x143194}, sa_mask = {__val = {1937176924, 1996051664, 1937176856, 1996167860, 0, 1996132080, 5, 0, 1, 1996167512, 0, 7680432, 0, 1937176916, 1992725916, 1992725900, 22, 0, 1996167952, 16, 1996167512, 
          1992725900, 1937176972, 10131396, 1937176988, 22, 11539032, 9118188, 1937178512, 1937180292, 1991608912, 1996147472}}, sa_flags = 1992815936, sa_restorer = 0x76c10334 <write+52>}
    sigs = {__val = {32, 0 <repeats 31 times>}}
```
#2  0x009c0c9c in __gnu_cxx::__verbose_terminate_handler() ()

No symbol table info available.
#3  0x009bd5e8 in __cxxabiv1::__terminate(void (*)()) ()

No symbol table info available.
#4  0x009bd610 in std::terminate() ()

No symbol table info available.
#5  0x009bdd38 in __cxa_throw ()

No symbol table info available.
#6  0x004258c4 in boost::throw_exceptionboost::lock_error (e=...) at /usr/local/include/boost/throw_exception.hpp:69

No locals.
[...]

PR #244, opened 2015-12-02
Secure TCPServer actions to avoid its usage while stopping or starting it, by using a mutex and read/write locks.

gaudryc:
Related issue #241


gizmocuz:
Already made the patch here internally and will commit soon

PR #251, opened 2015-12-08
Enable support for CM19A usb transmitter via mochad

### LINKED ISSUE #250: Support for CM19A via Mochad
The domoticz mochad plugin appears to support only CM15 pro, meaning that it will only transmit PL commands. Since mochad supports CM19A, it would be nice to include support in domoticz for the CM19A via mochad. Essentially, rf command transmission should also be enabled.

Note: Domoticz can already read all the info from the CM19A via mochad. The problem is that it sends pl commands only.

After looking at domoticz/hardware/MochadTCP.cpp it appears that the pertinent lines of code are 254-257. I believe modifying lines 255 and 257 as follows should do the trick. Please note that mochad repeats all rf commands over PL [1]. Hence, this change should work for both transmitters. 
255:<code>
sprintf (s_buffer,"RF %c%d on\n",(char)(pdata[4]), pdata[5]);</code>
257:<code> sprintf (s_buffer,"RF %c%d off\n",(char)(pdata[4]), pdata[5]);
</code>

[1] https://bfocht.github.io/mochad/mochad_reference.html#RF_to_PL_repeater

bugzy:
Fixes #250
This should enable the mochad plugin for domoticz to support the CM15Pro as well as the CM19A USB transmitter.


gizmocuz:
Thanks !!

PR #252, opened 2015-12-10
OpenZwave functionality

lightbullet:
Added new config files for the Eurotronic valve and added an extra option for using the dynamic OpenZwave library


gizmocuz:
I think you better post the new config file on the openzwave forum, then it will be in domoticz automatically when i update openzwave

PR #254, opened 2015-12-10
Adds support for setting a web root for the web interface

JohnAdders:
Adds support for setting a web root for the web interface which allows the interface to be setup behind a web proxy.
starting domoticz with -webroot domoticz means that the web interface can be accessed as
http://localhost:8080/domoticz/

Fixes issue raised https://www.domoticz.com/forum/viewtopic.php?f=6&t=8973


rkluwen:
Great, this is what I wanted as well as a preparation to use Domoticz-specific URL's in MyDomoticz. So thanks for the patch.
I'm afraid of having to merge the patch though, since it uses a lot of the same files. Need to see.

rkluwen:
Even though for it to be useful in my case, I would like to see to be able to set the path on a per-request basis. Or something like that.
For instance if you run a different webroot for http and for https requests (or in the near future proxy requests).

JohnAdders:
@rkluwen not sure I fully understand what problem you're trying to solve.  Do you need the code to be able to detect what kind of url was used, if so you should already be able to see that in the request object by looking at the host and method entries. What would you do differently based on the calling method?  But it does sound like a different issue, this patch is trying to allow the webroot to be set and then essentially ignored by the rest of the code.

grote-beer:
Is this pull request likely to be merged in the future? I would really like to have this as a feature in domoticz

gizmocuz:
it currently has conflicts and can not be merged

JohnAdders:
conflicts removed

nayrnet:
I really want this functionality as well, see my wishlist: https://www.domoticz.com/forum/viewtopic.php?f=21&t=9799

Basically I want to share a routable x509 authenticating proxy with multiple services, including domoticz.. 

Thanks for working on this.. any chance you can get domoticz to pull a usertoken out of a header? intention is the proxy can auth user accounts on behalf of domo.. I can pass the DN out of a certificate to domo through a header, but its got to be read and associated with a user.

gizmocuz:
I did some small testing and all seems to still work correctly
Thanks for the patch!

PR #255, opened 2015-12-10
Currentcost

JohnAdders:
Adds support for CurrentCost meters which pump out xml over their serial ports as documented here
http://www.currentcost.com/cc128/xml.htm

Also supports older meters which output slightly different xml at 9600 baud.


gizmocuz:
Thanks !!!

PR #256, opened 2015-12-10
Added OZW config file for POPP Wall Plug Switch Schuko (IP44).

hellqvist:
Added configuration file for the POPP Wall Plug Switch Schuko (IP44) with id POPE009105. 

Tested it live with my device and everything seems to work just fine.


gizmocuz:
Please post this on the openzwave forum, then it will be in domoticz when we update the library

hellqvist:
I saw that OpenZWave had the configurations as well so I've added it they as well (https://github.com/OpenZWave/open-zwave/pull/711).

gizmocuz:
Thanks, i updated the openzwave library in domoticz and made a new beta version

PR #259, opened 2015-12-15
Added column filters to the devices data table

wolph:
Just to make everything a bit more usable I've added some column filters to the devices table. Additionally I've changed the code to use the new data table API but the old API is still supported so there shouldn't be any backwards incompatible changes.

screen shot 2015-12-15 at 05 20 03


wolph:
PS: the diff actually looks far worse than it is due to some whitespace changes. I had my editor reformat the HTML because some tags were missing and it was difficult to find the proper opening tags with inconsistent indentation.

The somewhat more concise diff: https://github.com/domoticz/domoticz/pull/259/files?w=1

gizmocuz:
i think the top right search button lets you also search/filter the table based on hardware/id/unit/name etc....

wolph:
That's true, but this is far more convenient. Especially since you have
select boxes to filter on a specific hardware types and such.

Also, there are a few html5 input fields such as the date filter, so mobile
browsers and such should be able to understand how to use it.

And last but not least, the new api makes it easier to enable different
configurations as well. Reordering or hiding columns. Scroll instead of
pagination, etc.

ghost:
It looks good.

gizmocuz:
cross checked with other users, and we can work perfectly with the standard filter.
next to this, the filters are more usable on top of the table then at the bottom, specially when you have 200 sensors, and specified to show all
but i like the idea of the new data API !, then again, we should go for the angularjs implementation in all views/controllers, and there are other (nicer...) tables for angularjs

wolph:
I could easily add a toggle for the filter in case people don't need it so it won't take any extra space for people that don't find it useful :)

Agree about having them at the top, that's a bit nicer. I've just moved it. I've got nearly 400 devices too (as you can see in the screenshot). Perhaps we can just disable pagination with the filters on top.

As for the Angular implementation, good idea but a lot more work I think. Would be good for the future but probably requires a very large rewrite.

gizmocuz:
Are you good with angular ? time is my major problem !!
I use angularjs now for all business related software, its soooo much better

for the devices table this is not much work, but for the lights/temp/utility/weather tabs its more complicated, but once we have this, fun begins

gizmocuz:
could you make all changes to the formatting (spaces now) to use 'tabs' again ?

wolph:
I've replaced the spaces with tabs now, but I should mention that many files appear to be using a mixture of tabs and spaces.

As for Angular, I've used it a bit but I'm far from an expert yet Unfortunately I have the same problem you have, a lack of time. This was a pretty quick fix (took me about 1-2 hours I guess) but that would require weeks of work I'm guessing.

gizmocuz:
The problem i (still...) have is to see what actually has been changed, because you did a reformatting
it will take a long time to check every line character now

wolph:
That's why I added the link to the github diff that ignores the whitespace: https://github.com/domoticz/domoticz/pull/259/files?w=1

That way the diff is easily visible :)

gizmocuz:
Okey, that was also still a lot but... much better indeed
the new table css/js class does not break the tables in the other views?
Could you add an option to not display this in default, and maybe a small toggle to display it if wanted?

wolph:
I can take a look at it this weekend, the css class is a new one so I don't see how it could break anything :)

gizmocuz:
Uhm well... it could... and also the new .js file ;)

wolph:
I thought you were referring to my custom css class. But yes... the other one might be problematic, I haven't noticed any issues yet but I'll test all pages to be sure.

dannybloe:
I agree with gizmocuz that if you show this filter-bar it should be at the top instead of at the bottom. I think that when you use it you may have to adjust the filtering settings until you are satisfied. It isn't that convenient that you have to scroll all the way down each time you want to tune your filter. Filters hardly ever are at the bottom of the tables they filter.

ghost:
+1 and true should be at the top.

wolph:
It's actually been at the top since that comment :) I simply haven't updated the screenshot.

Unfortunately I'm swamped with work at the moment and probably will be for the next month so I have zero time to work on it.

gizmocuz:
The reason why i have not implemented this yet, is because i want to make this one purely based on angularjs
There have also been to many formatting going on so i can not see what actually changes (specially in the style.css)

wolph:
With the link given earlier the diff is easily visible: https://github.com/domoticz/domoticz/pull/259/files?w=1#diff-4fcfd97ace2383992a44cd111842a8d0

Specifically:

```
+tfoot.filters{
+    display: table-header-group;
+}
```

Regarding to the formatting changes, it might benefit the project to decide on a specific formatting style and enforce that across all files.

Right now some files use tabs, others use spaces, some have it mixed. And within that some use 2 spaces indents and others 4 space or even inconsistent/no indents at times. A recursive reformat across all sources is always painful in the beginning but useful in the end.

And conflicts can easily be avoided by telling your difftool to clobber whitespace.

gizmocuz:
not going to add filters in this table, the search option is more then enough (for now)

PR #261, opened 2015-12-15
OTGW read GPIO A/B status (new firmware feature)

jvandenbroek:
With the OTGW firmware 4.2.5 we can read out the actual GPIO status, instead of just reading out configured values. This commit uses the new command PR=I, which will return PR: BV on older firmware, so no things break up. I think this is better and more useful than reading the configured values (GPIOAPulledToGround), because this switch allows us to monitor connected sensors.

### DIFF
--- a/hardware/OTGWBase.cpp
+++ b/hardware/OTGWBase.cpp
@@ -253,15 +253,15 @@ bool OTGWBase::WriteToHardware(const char *pdata, const unsigned char length)
 			svalue = 255;
 		}
 		SwitchLight(nodeID, LCmd, svalue);
-  } 
-  else if ((packettype == pTypeThermostat) && (subtype == sTypeThermSetpoint)) 
+  }
+  else if ((packettype == pTypeThermostat) && (subtype == sTypeThermSetpoint))
   {
       const _tThermostat *pMeter=(const _tThermostat*)pdata;
       float temp = pMeter->temp;
       unsigned char idx = pMeter->id4;
       SetSetpoint(idx, temp);
-   } 
-  else 
+   }
+  else
   {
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thanks, i will update to the new firmware asap and test

gizmocuz:
Where did you download OTGW firmware 4.2.5 ?
I see only version 4.0b5 as latest version here:
http://otgw.tclcode.com/development.html

jvandenbroek:
http://otgw.tclcode.com/download.html

scns:
You can find it on the page image files

http://otgw.tclcode.com/download.html#hexfiles

Met vriendelijke groet,

Maarten Schmeitz

Verstuurd vanaf mijn iPhone

Op 15 dec. 2015 om 18:56 heeft gizmocuz notifications@github.com het
volgende geschreven:

Where did you download OTGW firmware 4.2.5 ?
I see only version 4.0b5 as latest version here:
http://otgw.tclcode.com/development.html

gizmocuz:
Is there a command to check what firmware version is loaded in the OTGW ?
I am not 100% sure if the upgrade went fine

jvandenbroek:
PR=A should give the current version. 

The code works great here btw ;-)

gizmocuz:
Thanks! Upgrade went perfect.... i am going to put the version number in the hardware page

gizmocuz:
Merged, thank you all !

rhannink:
Is it also possible to set the two GPIO signals from domoticz? I have a air fan connected to the GPIO connections and I would like to control the fan from Domiticz (switch)

gizmocuz:
looking at the code at OTGWBase::SwitchLight, this is already supported (GA/GB)

rhannink:
Ok, thanks I will try this...

PR #262, opened 2015-12-15
Feature/selector switch

gaudryc:
Add an new switch type : Selector.
It is a switch that displays a select menu or a button set. Each option or option is associated with a device level.
A selector can have up to 11 levels (including 0, associated to Off status).
Each level is associated with a name.
All level names are stored in a new DeviceStatus field : Options.
User can add/update/remove a level name.
Level names can be used in Blockly and Lua script as the Selector status.


gaudryc:
Dashbord capture on iPhone :
capture d ecran 2015-12-15 a 17 16 17

Dashboard capture on Chrome (PC) :
capture d ecran 2015-12-15 a 17 17 43

LightSwitches capture on Chrome (PC) :
capture d ecran 2015-12-15 a 17 19 23

Selector Switch edition capture :
capture d ecran 2015-12-15 a 17 20 36

gaudryc:
LightLog capture :
capture d ecran 2015-12-15 a 17 52 04

gaudryc:
Add manual device capture :
capture d ecran 2015-12-15 a 17 54 04

gaudryc:
Blockly capture (check Selector status using level name) :
capture d ecran 2015-12-15 a 17 59 58

Blockly capture (set Selector status using level value) :
capture d ecran 2015-12-15 a 18 01 31

Events log capture :
capture d ecran 2015-12-15 a 18 03 59

gizmocuz:
Great patch!, thanks !!
two things are not clear to me
- that is the language translation (did you update the English file on transifex?) because i see above in the screenshot some French translation
- the commands to action, i mean i see changes in RFXNames.cpp where i see words like
  On/off/group on/off/Paused/Pause/playing etc... how does that work ?

gaudryc:
- Translation is ready for french only (english is already in the sources).
- It works like a Dimmer except for On command because no level is setted on On command. It works like a Dimmer but levels can only have values 0, 10, 20, 100. Each of these level has a user defined level name that is used as Status (instead of percentage). In GetLightCommand if the command is not managed by the Selector switch, or the level has a invalid value, the method simply returns false.

gaudryc:
What is missing now :
- translation (using transifex) ;
- planning form must be changed to manage this new Switch. 

I hope it will be ready tomorrow.

gaudryc:
Planning form capture :
capture d ecran 2015-12-16 a 03 00 36

gaudryc:
Could you please tell me how I can update all translation files ? Because I tried to add some string to french translation but it did not work, and I cannot add these strings to default one (en).

gizmocuz:
you have to do this on transifex, or send me the missing english parts of the i18n file, and i will upload this to transifex
(or send me the complete new english file)

in the domoticz source tree is a folder tools, that has a command to pull all translations

gaudryc:
But if I cannot upload a file using the Web interface, I think I cannot do it from a command line tools, isn't it ?
I'll try anyway but without much hope.

gaudryc:
tx is saying :
Pushing translations for resource domoticz.domoticz-enjson:
Pushing source file (..\www\i18n\domoticz-en.json.)
Exception: Only maintainers are allowed to update the source file.

gaudryc:
Please find all new strings into the following source file :
[domoticz-en.json.gz](https://github.com/domoticz/domoticz/files/64640/domoticz-en.json.gz)

gizmocuz:
Thanks, transifex is updated
Do you want me to pull this request?

gaudryc:
Yes please.
It is OK for me and it will be very useful for those who wanted a switch
with multiple states (someone asked this feature one year ago).
It allows me to replace 12 switches by only 3 switches.

Thanks in advance.

gizmocuz:
Thanks for this great addon!

gizmocuz:
Noticed one type that i will fix:

mainworker.cpp(9983): warning C4102: 'std': unreferenced label

std:map<std::string, std::string> statuses;

should be

std::map<std::string, std::string> statuses;

(two double points between std and map

gizmocuz:
I need more changes, you made adjustments in the RFXCom config file, that won't work and will be overwritten on the next API change
I will rebuild it so a pTypeGeneralSwitch

gizmocuz:
committed the adjusted code, tested, but maybe you could triple check if it is still working correctly

gizmocuz:
Sorry, i screwed something up... made a new commit, but still something seems not right, it is switching correct, but the state is not set correct in the gui...

gaudryc:
I'm testing your changes...

gaudryc:
I'm making a patch (deviceid length issue)

gaudryc:
Please follow the following link Fix selector switch feature merge issue #264

PR #263, opened 2015-12-16
Return version with successful auth, fix unauthorized getconfig

mikefranken:
This is a security fix, currently it is possible to read someone's config (Lat/Long for instance) without (login-based) authentication. Furthermore I'd like to be able to use the app version upon successful authentication.


gizmocuz:
Thanks for the fix and addon!

PR #264, opened 2015-12-17
Fix selector switch feature merge issue

gaudryc:
I'm still testing...

### DIFF
--- a/hardware/Dummy.cpp
+++ b/hardware/Dummy.cpp
@@ -323,7 +323,7 @@ namespace http {
 					unsigned char ID2 = (unsigned char)((nid & 0x00FF0000) >> 16);
 					unsigned char ID3 = (unsigned char)((nid & 0x0000FF00) >> 8);
 					unsigned char ID4 = (unsigned char)((nid & 0x000000FF));
-					sprintf(ID, "%X%02X%02X%02X", ID1, ID2, ID3, ID4);
+					sprintf(ID, "%02X%02X%02X%02X", ID1, ID2, ID3, ID4);
 					unsigned long long devidx = m_sql.UpdateValue(HwdID, ID, 1, pTypeGeneralSwitch, sSwitchTypeSelector, 12, 255, 0, "0", devname);
 					if (devidx != -1)
 					{
--- a/main/SQLHelper.cpp
+++ b/main/SQLHelper.cpp
@@ -31,7 +31,7 @@
 	#include "../msbuild/WindowsHelper.h"
 #endif
 
-#define DB_VERSION 88
+#define DB_VERSION 89
 
 extern http::server::CWebServerHelper m_webservers;
 extern std::string szWWWFolder;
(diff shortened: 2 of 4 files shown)


gaudryc:
- DB update is OK ;
- Device creation is OK (both from LightSwitches and Hardware) ;
- Device update is OK ;
- Device commands are OK ;
- Blockly usage is OK ;
- Planning is OK ;
- Notifications are OK ;
- Log is OK ;
- All levels are OK (0, 10, 20, 30, 40, 50, 70, 80, 90 and 100).

gizmocuz:
Thanks!

remb0:
This is what I wanted to control my ventilation fan.  it has 3 speeds.   in domoticz 1 have 3 on/off switches.   Can I replace that with this one?   so every level I can set a command like:
http://192.168.0.60/control?cmd=GPIO,15,1

that is the onn command for fan speed 2.

gizmocuz:
I was thinking about this too... 
@gaudryc, is it possible that each level can have it's own ON-Action ?

remb0:
than additional lua scripts aren't needed. :)

I think also an off command.   because in my situation:   I have 3 buttons.
 that I press for 1 second. so on and after 1 sec I switch is off with:
http://192.168.0.60/control?cmd=GPIO,15,0

that means when I set it on 1  and then Go to 2.  domoticz has to disable the first button (otherwise the remote has 2 buttons pressed at the same time.

I will use this new switch type also for some presents for my heating: 
18©  21     

it really opens new possibilities.  and I think lot's of users were waiting for this!  so thanks for your great work.

gizmocuz:
to make it logical, the 'off' level/button should only send the off command, that is already implemented

remb0:
yes  but my ventilation system  has no off stand.  it's  1 (default) 2 or  3.
and all the three are buttons on a remote that i soldered on an ESP.  I send an on off button in 1 sec.
and after that my fan is on that speed (not off)    my english sucks so I hope I explained it well.

gaudryc:
I think I could add :
- an option to hide the Off button and set the first level On as the default level (easy);
- some options to the switch to be able to send switch command to other switches (more difficult).

gizmocuz:
is it possible for each 'level' to have an on-action ?

remb0:
an lever or action per option would be great.  then a blocky isn't needed anymore..
is this possible?

gaudryc:
I'm working on a new feature to :
- add a LevelActions option to a Selector device ;
- add a editing form to define an action (http:// or script://) to each level.

gaudryc:
I'm working also working on adding a LevelOffHidden option to a Selector device to hide the Off button.

ghost:
Hi @gaudryc  :)

Is any chance to add features as i describe in new issue #301 ? This good option for user to control AV Reciver (example Denon by HTTP)

One next suggestion is ad ToolTip to button image EDIT and DELETE:
Change name level
Delete level

gaudryc:
Hi @PiterEL,

Please see my answer in #301.

PR #266, opened 2015-12-18
Secure RxMessage decoding by removing m_LastDeviceName from m_mainworker instance

gaudryc:
It is neccessary because m_LastDeviceName is accessed directly by :
- rxqueue thread (by calling ProcessRxMessage);
- tcpserver thread (by calling DecodeRxMessage) ;
- webserver thread (by calling SwitchModal and SetInternalSecStatus);
- sqlhelper thread (by calling SetInternalSecStatus).

So m_LastDeviceName could be change between the time it is updated in CSQLHelper::UpdateValueInt function and the time it is read at the end of ProcessRxMessage function.

Using this fix, each caller of a decode_*(...) function has its own instance of the devicename string setted in CSQLHelper::UpdateValueInt.

Please also take a look to both TODO comment I've added at lines 10752 and 11603 (in mainworker.cpp). Is it possible to call PushRxMessage\* function instead of calling decode_\* function directly ?


gizmocuz:
Did not make the SwitchModel but can this use the normal DecodeRXMessage too ?
(Also the SetInterlSecStatus (with a wait like internal switching), then only the tcpserver is left ;)

It looks i miss some files in the commit ? or this is another genius change ;)

gaudryc:
I'm adding the new DomoticzInternal hardware to attach the Domoticz Security Panel device to it and be able to treat this device as a standard device...

PR #268, opened 2015-12-19
Avoid vertical scrollbar when viewing graphs from floorplan controller

djoele:
Retry for pull request #221 (avoid vertical scrollbar when viewing graphs from floorplan controller)


djoele:
Hello dnpwwo,

I think you are right. I guess you also have a better understanding of the structure of Domoticz.
I go with your suggested change and I will do it like that.

After that I will update this pull request.

Thank you for your comment.

djoele:
It has been done. Thank you again for your feedback.

dnpwwo:
I wrote the floorplans stuff so if I don't understand them we are in real
trouble :)

PR #277, opened 2015-12-29
Add Lua script editor to Domoticz web UI

dnpwwo:
Updated DB event master table to show Event type and interpreter.
Updated Event interface to web UI to show additional fields
Added editor to UI


dnpwwo:
Probably should have mentioned it is backwards compatible with existing Blockly and Lua file based scripts

gizmocuz:
Thanks!

PR #282, opened 2015-12-30
Add support for Remeha Calenta 40C

TheLizard:
Added support for the Remeha Calenta 40C

Currently it is just a single device, when there is demand, it is possible to create device recognition and automatically support all the other devices as well (all config files from the remeha recom software have been added)

For now only the temperature and percentage valued devices are added as devices.


TheLizard:
Unfortunately it took a whole lot of commits, because I don't have a working VS win32 build environment. Please let me know if anything else is required

gizmocuz:
Maybe easier to get yourself an opentherm gateway ? (or a thermostat that you can control from domoticz?)

TheLizard:
This didn't cost me anything and it was fun to build, 
also the opentherm gateway has to be placed next to the Remeha as I can't access the cable anywhere else, and still uses the raspberry pi to retrieve the data.

gizmocuz:
I agree, and it was a lot of work !
It is only that maybe one or two people will use this, over the larger and easier available thermostat support.

gizmocuz:
Hardware to specific, not maybe will use this.

PR #284, opened 2016-01-01
Cleaned up connection.cpp.

rkluwen:
Now there's only one handle_read(), etc. instead of different versions like handle_read_secure() and
handle_read_plain(). So if you want to update the code, now you only have to do this on one place instead of two.
I made this for a nifty new Domoticz feature-to-come. But since this is also convenient for others that work in connection.cpp, I made a separate pull request for this.

### DIFF
--- a/msbuild/domoticz.vcxproj
+++ b/msbuild/domoticz.vcxproj
@@ -55,7 +55,7 @@
       <PrecompiledHeader>Use</PrecompiledHeader>
       <WarningLevel>Level3</WarningLevel>
       <Optimization>Disabled</Optimization>
-      <PreprocessorDefinitions>WIN32;_DEBUG;PTW32_STATIC_LIB;WITH_OPENZWAVE;OPENZWAVE_USEDLL;NS_ENABLE_SSL;_CONSOLE;%(PreprocessorDefinitions)</PreprocessorDefinitions>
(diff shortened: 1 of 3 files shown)


dnpwwo:
What sort of changes are you thinking about?  

I have been looking at this code with a view to moving the floorplan updates from the current 10 second poll to 'http long polling' to make them more responsive. The current webserver doesn't support this very well because it seems to be very 'one in one out' oriented. There appears to be no easy way to 'park' a request until a response is ready because the connection details are buried quite deep and there are limited threads.

gizmocuz:
websockets could a an idea ;)

rkluwen:
http long polling should be possible. Just set the keep-alive timeout value to a large value.
But I am working on a better thing; websockets indeed. The protocol part is working already from the app to Domoticz and vice versa. Just I don't have a use for them now, since the app needs a lot of updating for push events to work.
But if you want, I pass you the code. Having someone to work with that knows their way in the app would surely be helpful.

dnpwwo:
With the current webserver I think you would tie up a thread for each long
poll (please correct me if I'm wrong) which could cause the server to stop
responding.

Websockets would be fine.  The key thing is that it runs on the same port
and uses the same authentication/authorisation model so that there aren't
firewall issues.

I won't have time to look at it for a few weeks because I'm tidying some
other things up but let me know when you finish your current development
and I'll have a play ☺

ghost:
BTW
dnpwwo i send you PM on DOmoticz forum... :)

PR #286, opened 2016-01-02
build issue: git should be called on source-tree instead of build-tree

gms1:
fixes out-of-source build


gizmocuz:
Thanks!

gizmocuz:
I receive Unknown option: -C

PR #287, opened 2016-01-02
build issue: use CMAKE_INSTALL_PREFIX  instead of hardcoded '/opt/domoticz'

gms1:
.) initialize CMAKE_INSTALL_PREFIX after CPACK_PACKAGE_NAME has been initialized
.) a cache variable will/can always be overwritten by command-line or cache settings
.) use CMAKE_INSTALL_PREFIX instead of hardcoded '/opt/domiticz'


gizmocuz:
Thanks!

PR #288, opened 2016-01-03
added support for siemens S7 PLC's

wouterdb:
Hi,

I've written a plugin for siemens S7 PLC's based on the Snap 7 project.
I hope this can become part of the mainline branch.

I've included some of the S7 header files in the build. If this is not acceptable, I can change it.
What is the standard practice for including C projects?

Thanks,
Wouter


gizmocuz:
If you include the license file also in the snap "c" folder, and it does not conflict with the domoticz license, we should be fine
for Linux I always make a cmake project (library) see for example the lua/mqtt source files
Is it possible to keep the snap7 source files in the hardware folder ?

does it always use the same tcp port, or can this be changed ?
If it can be changed, you can use a hardware description that ands with 'with LAN interface' and you do not need to modify the hardware js classes

wouterdb:
I added the license file, it is LGPL, so that should be compatible

Is the CMake file required? I'm new to cmake and the source files included are only the wrappers to included the C library into a C++ project. It doesn't really produce any artifact, so I don't really know how to write the cmake file. 

The port is fixed to 102 (RFC1006), but it would still be nice not to modfiy hardware js. I can just ignore the port. What do you advice?

gizmocuz:
Hardware is to specific, and not many will use this.
you could generate a patch for personal use of course

PR #289, opened 2016-01-03
LightwaveRF inline relay

algernon2:
Summary: Additions to allow the LightwaveRF inline relay to be operated from Domoticz via the Venetian Blind subtype. Previously the stop function was not available. Only open or close worked. I have not changed the icons in the LightsController.

The LightwaveRF inline relay device is recognised by Domoticz as a Switch. Automatic set up in Domoticz via its remote control, identifies the Switch as of type Blinds. 

The LightwaveRF inline relay controls a SPDT switch. A  common supply terminal can be switched between two output terminals. Using the available Switch types, it is possible to have Domoticz turn the relay on or off, ie to switch the common supply between two outputs. However, the device can also be sent a stop signal. This removes the common supply from both outputs. This can be used to stop the drive motor before a window or blind is fully open or closed.  I could not find any way of configuring the current version of Domoticz to send this stop signal. (Using the stop button on Venetian Blind results in no signal and an error message).

The modifications in the patches provided allow the LightwaveRF inline relay to be treated as a Venetian Blind in Domoticz. I chose this type because it provides three possible signals (open/close/stop) and three possible states (open/closed/stopped).

I tested these modifications using a LightwaveRF inline relay controlled by an RFXCOM transceiver. I used Domoticz 2.4020 running on a Linux Mint 17 machine for development.  My active Domoticz system runs on a Raspberry Pi B.

I am not familiar with the Domoticz architecture and coding, and could not find any relevant documentation, so my crude attempts to get this fix working may not be optimal. I have tried to write them in a way that should not impact other devices, but please treat them as a suggestion rather than necessarily a suitable long term solution.

### DIFF
--- a/main/RFXNames.cpp
+++ b/main/RFXNames.cpp
@@ -2185,12 +2185,21 @@ bool GetLightCommand(
 		else if (dSubType!=sTypeLightwaveRF)
 		{
 			//Only LightwaveRF devices have a set-level
-			if (switchcmd=="Set Level")
-				switchcmd="On";
+ 			if (switchcmd=="Set Level")
+ 				switchcmd="On";
+ 		}
+			// The LightwaveRF inline relay has to be controlled by Venetian blinds logic as it has a stop setting
+		else if ((dSubType==sTypeLightwaveRF)&&(switchtype==STYPE_VenetianBlindsEU)){
+				if (switchcmd=="On")
+					switchcmd="Close inline relay";
+				else if (switchcmd=="Off")
+					switchcmd="Open inline relay";
+				else if (switchcmd=="Stop")
+					switchcmd="Stop inline relay";
 		}
-
-		if (switchtype==STYPE_Doorbell)
-		{
+ 
+ 		if (switchtype==STYPE_Doorbell)
+ 		{
 			if ((switchcmd=="On")||(switchcmd=="Group On"))
 			{
 				cmd=light5_sGroupOn;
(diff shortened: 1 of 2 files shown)


gizmocuz:
How about modifying the code like

else if ((dSubType==sTypeLightwaveRF)&&(switchtype==STYPE_VenetianBlindsEU))  
{
    if (switchcmd=="On")
          cmd=light5_sClose;
        else if (switchcmd=="Off")  
          cmd=light5_sOpen;
        else  
          cmd=light5_sStop;
        return true;
}  

then all is working fine ? (no special states needed then?)

gizmocuz:
There also seems to be an error in the webserver class, please check below automatic build logs
but with the above change, maybe you do not need to change the webserver class at all

algernon2:
gizmocuz  - Thank you very much for your comments, and for looking at this do promptly. Apologies for the missing bracket in Webserver.cpp. I am using Github on a Windows machine which does not have a C++ compiler and I did a copy and paste. Now I have mounted a Samba drive from the dev machine.. 

I tried out your suggestion, and with the addition of the 'return true' lines it did work - and as you say, no changes to Webserver.cpp needed. However it does not display the device status fully (Stopped is missing) and it did not seem reflect changes of state made with the device's remote control, which my version does. Maybe I missed something simple again! So I have just corrected my Webserver.cpp error for the moment. I will have a look further at the code when I can.

gizmocuz:
It could be you also need to make the changes in the GetLightStatus function

algernon2:
I have spent quite a bit of time trying to get the approach you suggested working, including changing GetLightStatus, but I have failed. Although I can switch the three states, I can't get Domoticz to integrate with the working of the Lightwave remote control correctly, and the display does not reflect the changes of state produced by the remote control. I do not know how RFXCOM triggers those updates, but using the command strings "Open inline relay" etc, which were already present in RFXNames, seems to be important. 

I have come to the conclusion that it is unlikely there is a simpler solution than the one I proposed, which is coded in the revision currently posted. (Current head of my branch and same as the version called 'insert missing brace' two revisions earlier, because I made some changes and reverted them.) That version behaves correctly on my test system in Domoticz/RFXCOM in conjunction with the Lightwave three way remote JSJSLW826. I hope you will be able to incorporate this modification in a future release. For the moment I will work with a local patch.

Thank you for a really useful, good-looking and comprehensive bit of software.

gizmocuz:
Thanks for the patch, hope all works well

PR #302, opened 2016-01-08
Process incoming messages with ack request

fets13:
All MySensors messages from sensor asking for ack are not managed in domoticz : so no status updated in domoticz.

I modifed MySensorsBase.cpp to not ignore incoming message with ack request. Message sent back from sensor answering switch command from domoticz are still ignored.

Please accept this request because my garage door status was never updated, so I never know the actual status

### DIFF
--- a/hardware/MySensorsBase.cpp
+++ b/hardware/MySensorsBase.cpp
@@ -1561,12 +1561,9 @@ void MySensorsBase::ParseLine()
 			m_AckNodeID = m_AckChildID = -1;
 			m_AckSetType = V_UNKNOWN;
 			m_bAckReceived = true;
-		}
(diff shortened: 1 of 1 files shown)


gizmocuz:
Not sure why i explicitly did this... could be i encountered double messages, or an endless loop
I will apply the patch, and see how we go
Thanks!

fets13:
Thanks gizmocuz.
I didn't notice any endless loop.
How can I check for double messages ?

gizmocuz:
Can't remember, do you see any strange behavior in the log ? like while receiving a command with ACK, you see 2 entries ?
If not, maybe all is OK

fets13:
I ust check and didn't notice duplicate entries

PR #305, opened 2016-01-09
Fixed missing usage value in (virtual) P1 Smart Meter device

bobkersten:
When the utility page gets refreshed with new data, the p1 smart meter device only showed the return watts. This is fixed so that it now show both the used and returned energie.


gizmocuz:
Thanks for the fix!!

PR #315, opened 2016-01-11
Fix selector buttons state on device refresh

gaudryc:
Fix selector buttons state on device refresh (due to a jqueryui bug) by unselecting all buttons and removing "active" style (ui-state-active or ui-state-focus class).

### DIFF
--- a/www/app/DashboardController.js
+++ b/www/app/DashboardController.js
@@ -712,8 +712,18 @@ define(['app'], function (app) {
 											var selector$ = $(id + " #selector" + item.idx);
 											if (typeof selector$ !== 'undefined') {
 												if (item.SelectorStyle === 0) {
-													selector$.find('input[value="' + item.LevelInt + '"]').prop("checked", true);
-													selector$.buttonset('refresh');
+													selector$
+														.find('label')
+															.removeClass('ui-state-active')
+															.removeClass('ui-state-focus')
+															.end()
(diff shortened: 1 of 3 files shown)


ghost:
After this update all is crash
selector
. No icon and no level and no action to level

gaudryc:
Please create one issue in "Issues" because you have posted three messages in three different places.

PiterEL:
Ok i create issue #320

ghost:
Ok i clear cache browser and works ok.
But tell me is any chance to add mixed type with slider dimmer and buttons ? Is anyu chance to you create it in next patch ?

gizmocuz:
Welcome to 'PiterEL', i get sometimes 10 to 15 emails a day

gaudryc:
@PiterEL 

No I will keep it as simple as it is. You should know that the UI is simple but the code is not, therefore I don't want to make things more complicated just for one person.
I'm sorry but I already tell you that you need something specialized.

gaudryc:
@gizmocuz 

Do you know if someone is planning to develop a virtual device which acts as a media player (with source control, volume control, mute control) ?

gizmocuz:
@gaudryc 
i think there is already on inside domoticz (kodi player)

ghost:
But many user use AV Reciver like Denon, Onkyo and other and want control it by Domoticz.

PR #316, opened 2016-01-12
Elemental Theme

nayrnet:
Complete Theme for Domoticz, cleaned up and well tested.


gizmocuz:
Could you post a screenshot of the new theme,
and also a screenshot of the new slider width (before/after)

nayrnet:
screen shot 2016-01-12 at 10 10 20 am

nayrnet:
Sliders Before
frame_slider

nayrnet:
Sliders After, it was more apparent with the doubleiconsmalltable that something was not right.. I could not get any padding on the right so I went to look how default theme did it, it had small alignment issues too also as JavaScript sets the width automatically and css cant touch it.

I dont know if there is a better way to render the sliders and allow the theme to determine the size, probably not.. I could not come up with anything better than just tweaking it slightly. I tried giving the slider an ID that I could over-ride with CSS but then he javascript trumped that.

Nothing I tried to do to add padding to the right worked, I would try to resize the tables, tweak margins and padding.. anytime I would get some room over there it would vanish on a refresh when the JS decided to calculate the new width.

The problem is the default theme does not have an edge to edge #name at the top of the devices, that smaller blue bubble sets the width and if you want that to be full width in a theme it makes the sliders too big.

screen shot 2016-01-12 at 10 14 36 am

nayrnet:
Forum Release Thread: https://www.domoticz.com/forum/viewtopic.php?f=8&t=9795&sid=1502c29203d8fbdd58c8c26f560f6991

nayrnet:
This is what the default theme looks like after the changes.
screen shot 2016-01-12 at 11 26 04 am

gizmocuz:
Great, thanks for this !

PR #318, opened 2016-01-12
Fix install path

IanDury:
Fix install path from default /usr/local to /opt/domoticz. Apparently the set() function FORCE option is needed to overwrite existing entries, including the default value of CMAKE_INSTALL_PREFIX which is /usr/local


gizmocuz:
Thanks

PR #326, opened 2016-01-13
custom-icons: if temporary file is not writable try putting it in /tmp

nayrnet:
On my linux system domoticz is started with systemd and whatever the default path is (likely root) is not writable. This results in a broken custom icon uploader.

This small patch tries to fall back to /tmp before tossing an error.

### DIFF
--- a/main/SQLHelper.cpp
+++ b/main/SQLHelper.cpp
@@ -6598,10 +6598,12 @@ bool CSQLHelper::InsertCustomIconFromZip(const std::string &szZip, std::string &
 	outfile.open(outputfile.c_str(), std::ios::out | std::ios::binary | std::ios::trunc);
 	if (!outfile.is_open())
 	{
+#ifndef WIN32
 		//path might not be writable on linux, fall back to good old /tmp
 		outputfile = "/tmp/custom_icons.zip";
 		outfile.clear();
 		outfile.open(outputfile.c_str(), std::ios::out | std::ios::binary | std::ios::trunc);
+#endif
 		if (!outfile.is_open())
 		{
 			ErrorMessage = "Error writing zip to disk";
--- a/main/SQLHelper.cpp
+++ b/main/SQLHelper.cpp
@@ -6597,8 +6597,8 @@ bool CSQLHelper::InsertCustomIconFromZip(const std::string &szZip, std::string &
 	std::ofstream outfile;
 	outfile.open(outputfile.c_str(), std::ios::out | std::ios::binary | std::ios::trunc);
 	if (!outfile.is_open())
-#ifndef WIN32
 	{
+#ifndef WIN32
 		//path might not be writable on linux, fall back to good old /tmp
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thanks, but of course it should not be done on win32 systems (maybe a #ifndef around it)
But the patch where Domoticz is started should be writable by Domoticz, because there are more files written there (openzwave, curl coockie, ....)
What Linux system are you using ? Can't you use our default startup script with some slight modification?

nayrnet:
I am using debian jessie, prefer to use systemd because it'll restart domo if it crashes.

I was never able to get the custom icon uploader to work, always returned Error writing zip to disk. Looked at code, realized it was a temporary file and appended /tmp to the path and rebuilt and everything worked fine. 

everything else seems to work fine, this is the only issue and I presumed it was because I was not using standard init scripts.. seems this file is just stored so it can be extracted, /tmp seems good to me.

I'll add a #ifndef around it

nayrnet:
meh, sorry about that jumble.. jumping between coding languages is turning my brain to jelly. should be kosher now.

direct link to new code: https://github.com/nayrnet/domoticz/blob/6954d73442b9c71a6f6167540ab5943a494a8a1f/main/SQLHelper.cpp#L6601

gizmocuz:
Thanks for this, it seems fine... but....
could the problem not be in your systemd start script?
(just trying to see if we could make a global script that all can use)
(i'm no systemd expert bdw)

nayrnet:
oh trust me this was my last resort, Ive combed over systemd documentation and example startup scripts and have not came up with anything that fixed this.

my startup script is as simple as I can make it, I am starting to understand why everyone hates systemd so much lol.
see: https://www.domoticz.com/forum/viewtopic.php?f=8&t=5024&start=40#p68413

I figured this is because I bootstraped my own minimal os onto a bunch of arm hardware that nobody else seems to use, all my instances have the problem.. ive got 3 beagle bones (slaves) and a cubox, all running same minimal jessie setup with hardware watchdog timer on OS and systemd doing watchdog for all my buttload of io daemons.

gizmocuz:
Okey, i understand this makes total sense....
I am not going to merge this patch, but instead on 4 more locations i will use the /tmp folder as standard
Done, code committed, is it possible you could test
- backup the database from the settings page
- change a device name
- restore the database
- upload a custom icon
- if using a RFXTRX433(e) upload a new firmware

I hope all is working correct now

nayrnet:
good work, that should cut down on flash writes for people using /tmp as a ramdisk.. its not alot but my /tmp is the fastest fs on most of my instances.

I can test all but RFX, I lack any of that hardware because it operates in same band as my HAM :)

nayrnet:
everything works fine, thank you.

yay, I dont have to patch domoticz anymore and all my bugs been squashed.. now to find some more.

gizmocuz:
Thanks for the feedback, i also tested everything (including firmware upgrade).
Hope you won't find more bugs _g_

PR #330, opened 2016-01-13
Fix selector action https support

gaudryc:
Add also https:// to supported URIs list on action editing forms.


ghost:
But http support is added ?

gaudryc:
Previously JavaScript allow the user to only enter the URIs like "http://" and "script://". Now URIs like "https://" are also allowed.

pepeEL:
When you set up a command to execute at a neutral level ... with this level it does not work and is performed by the level off.

Issue #335, opened 2016-01-13
The characters : ' | " break the selector and/or its editing form


gaudryc:
I have to fix it as soon as possible ;-)

It is not a blocking issue. It is just not very clean.

PR #336, opened 2016-01-14
Fix unauthorized characters issue in selector switch

gaudryc:
This should fixes the issue #335.

Only characters :;|<> will be unauthorized from level names.


gizmocuz:
Does this also need to be patched in the scene timer editor ?

PR #341, opened 2016-01-14
JSON: Added PlanID and PlanIDs to output for Groups/Scenes

EddyK69:
Also added HistoryLog method. This shows all switchdata orderd by date. Can be used for making custom log info incl. types & icons etc...


gizmocuz:
Is this history log not the same as the current RType_LightLog ?

EddyK69:
No, that one only shows 1 device of the Light-types, not the sensors, doorbells & smoke-detectors etc. The RType_HistoryLog shows ALL devices.
Alternativly i could change the existing RType_LightLog to make it show all devices.
Whats you're opinion: 1) change existing RType_LightLog or 2) add new RType_HistoryLog?

EddyK69:
I think i'll close this PR to split in 2: 
1) JSON: Added PlanID and PlanIDs to output for Groups/Scenes
2) JSON: Added HistoryLog method

gizmocuz:
Please split it in two, because the lightlog does show the log of PIR's, Doorbells, smokesensors, all lighting devices

EddyK69:
Done ;)

PR #344, opened 2016-01-14
basic support for x509 sso

nayrnet:
I tested this and it works fine with no further implementation, and no interference with other auth mechanisms.

In Nginx I have the following config:

``` nginx
    # Domoticz
        location / {
        proxy_pass          http://localhost:8080/;
        proxy_set_header    Authorization $ssl_client_s_dn;
        proxy_hide_header   Authorization;
    }
```

it passes the Client Certificate DN to Domoticz in the auth header, Domoticz parses this DN and extracts a username (commonName) and password (emailAddress), then checks the user database for a match and signs them in.

At the same time nginx suppresses any Authorization headers coming from Domoticz requesting a password.

Ideally the password would not be verified for x509 users, so they can fall back onto a more secure password if they bypass the proxy.. however I am afraid of introducing security flaws so simple = better.

See Wiki for more details: http://www.domoticz.com/wiki/Secure_Nginx_Proxy_Setup
See Forums for my brain chatter: https://www.domoticz.com/forum/viewtopic.php?f=21&t=9799

### DIFF
--- a/webserver/cWebem.cpp
+++ b/webserver/cWebem.cpp
@@ -993,25 +993,40 @@ void cWebem::CleanTimedOutSessions() {
 // Return 1 on success. Always initializes the ah structure.
 int cWebemRequestHandler::parse_auth_header(const request& req, struct ah *ah)
 {
-	const char *auth_header;
+        const char *auth_header;
 
-	if ((auth_header = request::get_req_header(&req, "Authorization")) == NULL)
-	{
-		return 0;
-	}
+        if ((auth_header = request::get_req_header(&req, "Authorization")) == NULL)
+        {
+                return 0;
+        }
 
-	//Only accept Basic Auth header
-	if (!boost::icontains(auth_header, "Basic "))
-		return 0;
+        // X509 Auth header
+        if (boost::icontains(auth_header, "/CN="))
+        {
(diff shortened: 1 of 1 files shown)


nayrnet:
ignore my initial message, i managed to get this operational through a better nginx configuration, no configuration is nessicary for Domoticz other than setting up users and enabling basic-auth

This method is preferred because you can fall back to basic auth without the x509-proxy, for example allowing json requests to come through on localhost without x509. The -nowwwpwd run flag works fine along with x509 if you lock your self out.

gizmocuz:
Okey, so we do not this patch/pull request?

nayrnet:
this is ready to go

gizmocuz:
Your nginx implementation or this patch ?

nayrnet:
This patch is good to be merged, it adds support via Nginx.. 

I had an issue originally and was looking for a bit of help, but I found the fix on the proxy side.. edited the original post.. you'll see a PM from me thats nonsencial.

nayrnet:
I had an issue originally with the password prompt being displayed when Auth-Basic was configured, was looking for some help making a config option to bypass it..

in the end I found I could just block the header with nginx and the prompt would not display, the original submission was perfectly fine with a bit of additional configuration on the auth-proxy side.

PR #345, opened 2016-01-15
fix logout button for mobile devices

nayrnet:
Logout button text needs to be hidden like the rest of the buttons.

Before:
screen shot 2016-01-14 at 10 05 57 pm

After:
screen shot 2016-01-14 at 10 04 23 pm


gizmocuz:
Thanks for the patch !

PR #347, opened 2016-01-15
Extended "udevices" API call by adding JSON data injection with JsonPath

### LINKED ISSUE #343: Extend the "udevices" API call with json parsing
Current implementation of "udevices" is doing parsing with lua scripts but json parsing is not very easy with lua since there is not such json library for lua and embedding it in domoticz may not be easy.

Proposal: Add some lua functions to parse json

Example:
local temperature = domoticz_parseJson(buffer,"Monitor.S.S1.item1.value");
domoticz_updateDevice(15,0,temperature);

To do so, I would use the json library already there

I can do it

sbouchex:
Extended "udevices" API call by adding JSON data injection with JsonPath (see http://jsoncpp.sourceforge.net/class_json_1_1_path.html)

A new LUA function has been added : domoticz_applyJsonPath
This function applies a JsonPath to a Json data and returns the value.
Parameters are: <buffer><json path>[arg1][arg2][arg3][arg4][arg5]
Optional arguments are passed to the jsonpath parser (see http://jsoncpp.sourceforge.net/class_json_1_1_path.html for more details)

this commit resolves #343

### DIFF
--- a/main/WebServer.cpp
+++ b/main/WebServer.cpp
@@ -2336,6 +2336,8 @@ namespace http {
 						_log.Log(LOG_ERROR, "WebServer (applyJsonPath from LUA) : Invalid Json data received");
 						return 0;
 					}
+
+					// Grab optional arguments
 					Json::PathArgument arg1;
 					Json::PathArgument arg2;
 					Json::PathArgument arg3;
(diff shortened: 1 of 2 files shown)


gizmocuz:
Thanks!

PR #348, opened 2016-01-15
Ignore eclipse project in git

sbouchex:
For those using Eclipse (like me)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@ Config/zwscene.xml
 backups/
 domocookie.txt
 domoticz.exe
+.project


sbouchex:
Should be done locally (i.e. on the client side)

gizmocuz:
Maybe you could add the Eclipse project to Domoticz?
Have you looked at 'codelite' ?

sbouchex:
I'm just using Eclipse to manage git and doing a fill Eclipse project would be useless. I don't know codelite, but could be interesting.

gizmocuz:
ahhh okey, bummer... so you could have as well build it from the command line instead of using eclipse?

PR #350, opened 2016-01-16
Renamed ScheduleType FixedDateTime to OnceOnly

EddyK69:
OnceOnly is a better name for FixedDateTime because timers of this type will be deleted after they are fired.


gizmocuz:
I think Fixed Date/Time is more clear, then OnceOnly (or Only Once)
I would like to keep it this way...

EddyK69:
np ;)

PR #351, opened 2016-01-16
Extended "udevices" API call by adding XML data injection with XPath

### LINKED ISSUE #333: Extend the "udevices" API call with xml/xpath parsing
Current implementation of "udevices" is doing parsing with lua scripts but xml/xpath parsing is not very easy with lua since there is not such xpath library for lua and embedding it in domoticz may not be easy.

Proposal: Add some lua functions to parse with xpath.

Example:
local temperature = domoticz_parseXPath(buffer,"/Monitor/S/S1/item1/value/text()");
domoticz_updateDevice(15,0,temperature); 

To do so, I would integrate tinyxpath (http://tinyxpath.sourceforge.net) which requires tinyxml (which is already in domoticz) and declare a new "C" lua function to do the XPath parsing

I can do it

sbouchex:
Added the capability to inject XML data with a lua script (see provided example) and some xpath.

tinyxpath + tinyxml have been integrated and tinyxpath has been removed from the zwave code in order to have a clearer code structure.

Resolves #333


gizmocuz:
Only thing is, why did you change the Alert sensor in the webserver class, this should not be done, there is no 'desc' json value anymore, and nowhere used (in domoticz)
Could you revert that change back ?

sbouchex:
Done, sorry about that

PR #353, opened 2016-01-17
Handle DT_UNKNOWN in directory entries

dwmw2:
Not all file systems (and indeed not all ext4 file systems) return a
useful d_type in getdents(). The man page explicitly tells us that we
should handle DT_UNKNOWN. Do so.


gizmocuz:
Thanks for the patch !

PR #356, opened 2016-01-17
Change icon in menu logout

pp1984:
Looks better. Now icon logout and shutdown is very similiar. My icon is
better loook.


pp1984:
Hi
I also change icon devices, settings and add new icon for update.

pp1984:
Screenshot after changes:
screen

screen2

gizmocuz:
I like  my icons more!

pp1984:
But some menu have no icon. Maybe only this icon you can add ?

After analysis I see that a lot of reports from users about appearance and style do not want to accept. So if the project is closed?

gizmocuz:
Yes

pp1984:
Not good. But OK. 
And why some menu have no icon ?
But i dont understand why is on GitHub if is close ?

nayrnet:
make your own theme if you want your own icons.. changing the default for everyone, simply for your personal aesthetics is unlikely to be approved.

I think the original icons look better than yours, Gizmocuz does too.. so there you go, only you want these changes and thats why they wont be merged.

If you find serious display bugs in the default theme, patches are welcome.. if all you want to do is alter its appearance for aesthetic reasons, well good luck.. those typically get rejected so your going to have to do better than this.

PR #364, opened 2016-01-19
Domoticz/master Panasonic TV support

sijones:
Hi,

Could you bring this in to a separate branch maybe called panasonic, so someone can do the javascript ui part for me?

I've done the server side plugin part, works the same as the kodi remote, so we need the remote altering to be able to send the commands to panasonic plugin when the popup is opened. we have more commands available but i guess a basic remote will do for now.

i've looked at the javascript but don't know how to change it to work the way I'd like it to.

Cheers,


sijones:
I think I might have worked it out on how to do the remote, will comment back when I get it a little clearer!

gizmocuz:
Okey....

Well if you have someone working on the JAVA part, they can also use directly your repository correct?

sijones:
Hi Rob,

This is working now, the only thing i'd ask is if you could look at the code that enable/disables the hardware, I seem to have an issue that once you add the plugin you can't disable it, it won't change state and am not sure why?

gizmocuz:
I looked at the code, and make one small change (the try/catch handler), i can disable it now
If from the light widget i try to send an 'Off' command, i get a notification 'Unknown command: Off'
is it possible to send an On/Off to the Panasonic ?

sijones:
Yes, if you send off the TV does go off (can't turn on as no network once
tv is off). but maybe i need to look at the upper and lower case commands,
will check later today

PR #366, opened 2016-01-21
[Satel Integra] Support for Polish letters

tomaszszczukobls:


gizmocuz:
You have a build problem.

To make std::string's constant you have to declare them as

(const std::string &name)

In your header file you have to remove the SatelIntegra:: in front of SatelIntegra::ISO2UTF8(

szczukot:
fixed

gizmocuz:
Thanks!

PR #368, opened 2016-01-22
if dimmer svalue is 255 that is 100%

nayrnet:
if a dimmer reports to the WebUI with a svalue of 255 it does not render the slider @ 100%

this fixes that problem.


nayrnet:
this is a poor fix, think I found issue in zwave hardware

Issue #324, opened 2016-01-12
Switch level 255%


ghost:
Why for switch is show level 255% but not 100%?
![level](https://cloud.githubusercontent.com/assets/4134018/12278447/82ef78b6-b981-11e5-8c99-424ba41b36d3.jpg)

ghost:
Maybe can you give me some answer. Because i dont know.

PR #369, opened 2016-01-22
If dimmer has value > 100%, set value to 100%

nayrnet:
Z-Wave dimmers @ 100% Brightness should report 100% and not 255%

### DIFF
--- a/hardware/ZWaveBase.cpp
+++ b/hardware/ZWaveBase.cpp
@@ -293,11 +293,13 @@ void ZWaveBase::SendSwitchIfNotExists(const _tZWaveDevice *pDevice)
 		// Now check the values
 		if (level == 0)
 			lcmd.LIGHTING2.cmnd = light2_sOff;
-		else if (level == 255)
+		else if (level > 99)
+			if (pDevice->devType==ZDTYPE_SWITCH_DIMMER)
+				level = 100;
+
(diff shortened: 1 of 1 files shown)


nayrnet:
with this patch when you pull a ZWave slider to 100% it stays there and status eventually changes from **100%** to **On**, the expected behavior.

pp1984:
But why change from 100% to ON ? I Think that should 100% ON.

gizmocuz:
That is correct, when you set a dimmer to 25%, then turn it off, then drag the slider to 255 (it will now send ON), it goes back to 25%

nayrnet:
I think this is a bit cleaner of a fix, hopefully ready to merge.. testing on my end all weekend has been fine, no ill effects.

nayrnet:
screen shot 2016-01-25 at 12 52 05 am

Would this be a problem? Shouldn't that say **On** instead of **Set Level 100%**

nayrnet:
oops, I messed that cleaner fix up.. fixed my mistake
screen shot 2016-01-25 at 1 04 58 am
now everything works as should, bug squashed.

screen shot 2016-01-25 at 1 06 26 am

also found this closed issue related to this bug, going to mention it so its pinged. https://github.com/domoticz/domoticz/issues/324

gizmocuz:
Thanks for the patch !

PR #371, opened 2016-01-22
Added domoticz_updateDeviceName method to rename a device

sbouchex:
A new method to update the device name from the received packet

### DIFF
--- a/main/WebServer.cpp
+++ b/main/WebServer.cpp
@@ -2506,6 +2506,29 @@ namespace http {
 			return 0;
 		}
 
+		static int l_domoticz_updateDeviceName(lua_State* lua_state)
+		{
+			int nargs = lua_gettop(lua_state);
+			if (nargs == 2)
+			{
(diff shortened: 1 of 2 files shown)


gizmocuz:
Why can you not do this from the web gui ?

sbouchex:
In the case of some devices pushing xml data like the TCW240B (http://www.teracom.cc/products/tcw240b-data-acquisition-module/), the name of the device is defined in the pushed packet (see http://212.25.45.120:240/status.xml on the node ./Monitor/S/S1/description for example). The developer will just create the dummy device and the lua script will received the data and update the device name. The name may change with the pushed packet and domoticz will be aware of this.

gizmocuz:
Dont think this is a good idea.
Please use the web gui

Second, when the device name should be 'my device name', but i want to name it 'my device in room xx', it gets changed back to 'my device name'

We have a web interface where you could do all setup, like to keep it that way

PR #377, opened 2016-01-23
Create ZMNHAD1.xml

GravityRZ:
this is my first change so hopefully it goes well

changed index10.
before it did not accept the settings so all on/off could not be changed
changed byte size and deleted min max settings


gizmocuz:
I will update openzwave, but you need to get this file into openzwave

PR #383, opened 2016-01-25
Implemented: Extended schedules for SetPoints

EddyK69:
**Added extended schedules for SetPoints:**
- Odd day numbers
- Even day numbers
- Odd week numbers
- Even week numbers
- Monthly (like: every day 5 of month)
- Monthly (weekday) (like: every second sunday of month)
- Yearly (like: every december 5)
- Yearly (weekday) (like: every third tuesday in september)


gizmocuz:
Question, why another web implementation to get the timerlist of the setpoints ?
This function already exists

json.htm?type=setpointtimers&idx=xxxx

Could this be used?

EddyK69:
No, this ones shows **all** the setpointtimers just like the existing gettimerlist & getscenetimerlist.

gizmocuz:
Where/When is this used?
If it is for debugging, it would be better to extern the RType_Schedules function ?

EddyK69:
It is used in custom templates to show an overview of all the timer-schedules (ligthts/scenes/setpoints).
I will revert the getsetpointtimerlist method in this PR, and will rethink your suggestion on using the RType_Schedules function.

gizmocuz:
I think it is best to use the RType_Schedules for this, if you also made a command like this for lights/scenes, it would be better to use the above command as well

EddyK69:
The only problem is, that the RType_Schedules doesn't show the disabled schedules; in the existing gettimerlist & getscenetimerlist i get an overview of **all** schedules, both _**enabled & disabled**_ :)
So i would like to keep those and would like to add one for the setpoints as well...

gizmocuz:
Hmmm why you want disabled timers ?

PR #384, opened 2016-01-26
Implemented: JSON method getsetpointtimerlist

EddyK69:
In addition to the existing gettimerlist & getscenetimerlist methods, this method shows an overview of **all** schedules, both **enabled** & **disabled**.
The proposed alternative RType_Schedules, only shows the active schedules...

### DIFF
--- a/main/WebServer.cpp
+++ b/main/WebServer.cpp
@@ -3369,6 +3369,55 @@ namespace http {
 					}
 				}
 			}
+			else if (cparam == "getsetpointtimerlist")
+			{
+				root["status"] = "OK";
+				root["title"] = "GetSetpointTimerList";
+				std::vector<std::vector<std::string> > result;
+				result = m_sql.safe_query(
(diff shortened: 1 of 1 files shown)


gizmocuz:
As the RType_Schedules is only used for debugging, maybe you can patch this function so it returns all timers (lights/scenes/setpoints) with states ?
Then we can remove also the gettimerlist and getscenetimerlist 
The most important feature of RType_Schedules, is that it is actually telling you exact when the next moment is when the timer is fired
That it is scheduled every thursday (database information) is nice, but you also want to see that at 28 jan @ 19:00:00 that this timer is fired
The combination would be very nice!

EddyK69:
Agree! If you're right about that RType_Schedules is only used for debugging, i gladly want to patch this to the proposed extra info.
PR can be closed if you agree ;)

gizmocuz:
Would be great if you can combine all ! ;)

PR #385, opened 2016-01-27
match behavior with previous z-wave-dimmer-fix

nayrnet:
nevermind i think this may have issues.


gizmocuz:
You closed the issue without reason... do we need to patch this or not ?

nayrnet:
I dont think so, my patch could allow invalid data as it dont have any value = 100 reset, I retracted the patch until i test it a bit more.

gizmocuz:
Thanks for the feedback.
I tested this function, and it works correct

PR #386, opened 2016-01-27
New Dummy Switch

nayrnet:
Adds device pTypeLighting2, sTypeDummy, STYPE_Dimmer with 100 dimming steps instead of 15.

Wont break existing switches, will only effect new dummy switches created from here on.

does not change z-wave behavior, I reverted that back by hand.

screen shot 2016-01-27 at 3 21 25 am
screen shot 2016-01-27 at 3 21 43 am


gizmocuz:
I dont want pTypeLighting2 with sTypeDummy
pTypeLighting2 with sTypeZWaveSwitch has already 100 levels

And there seems to be an unfinished pTypeGeneral, with a sTypeSwitch, (working on it now)

nayrnet:
ok

PR #387, opened 2016-01-27
Added Nest Heating/Cooling logging in a "switch" device.

Delaroyas:
Simple implementation for Nest Heating/Cooling logging in a "switch" device. 
Supports only basic heating/cooling.
Tested on a minimal linux compilation (no OpenZwave) with a real heating thermostat, and a  heating/cooling virtual thermostat.

### DIFF
--- a/hardware/Nest.cpp
+++ b/hardware/Nest.cpp
@@ -561,14 +561,14 @@ void CNest::GetMeterDetails()
 			}
 
 			// Check if thermostat is currently Heating
-			if (nshared["can_heat"].asBool() and !nshared["hvac_heater_state"].empty())
+			if (nshared["can_heat"].asBool() && !nshared["hvac_heater_state"].empty())
 			{
 				bool bIsHeating = nshared["hvac_heater_state"].asBool();
 				UpdateSwitch(113,bIsHeating,"HeatingOn");		
 			}
 
 			// Check if thermostat is currently Cooling
-			if (nshared["can_cool"].asBool() and !nshared["hvac_ac_state"].empty())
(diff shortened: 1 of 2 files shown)


Delaroyas:
Mmm it compiled on my side... I'll investigate..

gizmocuz:
That is strange, because 'and' is not c++, it should be &&

Delaroyas:
That's what I figured, but it seems my compiler (gcc) didn't mind...

Anyway, let's try with actual c++ this time.

gizmocuz:
Not sure i should trust your coding skills now....

Delaroyas:
I dont blame you. I havent coded in c++ for a long time. 
Between matlab, php, javascript, mysql and lua keywords and syntax gets mixup sometimes.
But I did compile and test the code under 64bit linux before submitting the first time.
Anyway, you can always revise my code before accepting.

Delaroyas:
FYI:  C++ defines keywords to act as aliases for a number of operators: and (&&), bitand (&), and_eq (&=), or (||), bitor (|), or_eq (|=), xor (^), xor_eq (^=), not (!), not_eq (!=), and compl (~). For compatibility with C, C++ provides the header ciso646.  The latter file has no effect, being empty. Notwithstanding some compilers, such as Microsoft Visual C++, do require the header to be included in order to use these identifiers.
https://en.wikipedia.org/wiki/Operators_in_C_and_C%2B%2B#C.2B.2B_operator_synonyms
https://en.wikipedia.org/w/index.php?title=Ciso646

Delaroyas:
Merged changes from domoticz.
Rebuild and tested locally, everything is sill working as expected.

gizmocuz:
Thanks !

PR #388, opened 2016-01-28
Add support for pressure readings from 1Wire-devices by EDS

tuomaz:
This adds support for pressure (barometric) readings from 1Wire devices by EDS. 

http://www.embeddeddatasystems.com/OW-ENV-TP--Temperature-Barometric-Pressure-Sensor_p_171.html


gizmocuz:
Thanks!

PR #392, opened 2016-01-29
Allow domoticz to be compiled without builtin MQTT library

xbeaudouin:
Hello,

On FreeBSD there is a port (package) of mosquitto. 

To avoid issues with duplicate code, here is a proposed change of the CMakeList.txt and MQTT.cpp to allow the utilization of local libraries instead of static linked one.

Regards


gizmocuz:
Missing mosquittopp.h

xbeaudouin:
The mosquittopp.h is system one. Eg on FreeBSD /usr/local/include/mosquittopp.h. That's why on MQTT.h I use include <mosquittopp.h> ....

gizmocuz:
We have a build error, see below (appveyor)

xbeaudouin:
Ok Win32 build is a bit different, so I have added #ifdef for that.
Feel free to comment it :)

gizmocuz:
Thanks for the fix.
What is wrong with using the native MQTT headers?

xbeaudouin:
You mean MQTT headers that are in the MQTT/ directory ?
If this is what you mean, those headers are ok, but when you link on an external library, it is better to use the library's header.
About the fix for Windows, I think the MQTT dll and headers doesn't exist on the system, and I was not aware that the Windows build don't use CMake (I have no windows, and I never compiled a thing on windows except using Cygwin...).
So since it seems that the windows build doen't have DLL/Library for MQTT, the best way is use the embedded code as the traditional build.
Regards,
Xavier

gizmocuz:
But we should not link on 'an' .. 'external' library, we should link on the library that is provided 'with' domoticz
Same for 'Lua', we provide the 'Lua' library in our project, and this one should be used.

xbeaudouin:
If you look at your CMakeList, we can link the LUA with an external library.
Why would you like to maintain a LUA, Mosquitto, SQLite, Zlib? Did you changed anything on the code of those libs?
If not this can help people to run the program, but this can add several issues : 
- Duplicate code on the same machine
- Security issue not fixed because lack of time to monitor all software
  And also large binaries, maybe not optimized on the platform used. In my case I speak for FreeBSD.

To make analog idea, do you will like to add all the boost source code inside domoticz ?

sbouchex:
:+1:  and you get a smaller binary (which is better for ARM devices having small storage capability)

gizmocuz:
Are there still ARM devices with small storage then these days ? _g_
Thanks for the patch!

sbouchex:
yes, openwrt on network devices... I have heard that somebody was porting domoticz on a small device running this shrink-warp linux

Moreover, is it legal to mix GPLv3 (domoticz) and Eclipse (Mosquitto) in the same repository and linked statically ? I'm not sure...

gizmocuz:
small storage will not work, unless you are going to do stuff like hard linking network shares to store domoticz database/log/...
And why not buy a cheap raspberry pi (pine64 later) that uses almost no power ?
I don't get it... home automation is not cheap, sure you can invest in some dedicated hardware ?

About licensing, i think we can dual license without issue

sbouchex:
Agree on the storage point but I saw the request one about openwrt.

It would be nice to have a dual license in order to allow linking with other API (like yoctopuce)

PR #398, opened 2016-01-31
Update EnOceanESP3.cpp

stephaneguillard:
Added handling of 4BS teach-in message without EEP profile data: domoticz now creates a 0/40°C temperature sensor, with correct device id, and prompts the user to manually adjust profile data using domoticz's Setup/Hardware/EnOcean Setup button (which commit will follow)

Also fixed the split-line log of RSSI at the entrypoint of ParseRadioDatagram()


gizmocuz:
Thanks!

jlstanus:
Thanks i ll test thas as soon as possible

PR #400, opened 2016-02-01
Update manufacturer_specific.xml

russdan:
Add reference to new scsc17.xml config file to fix errors in the standard config for the thermostat setpoints.


gizmocuz:
This will not work
First, you forgot to attach the real config file
Second, you have to submit a pull request to the openzwave team, when it is accepted there it will show up in domoticz later

PR #402, opened 2016-02-01
Add translation to button Stop Blinds

ghost:


gizmocuz:
Thats not how this works
- First you translate the text one-on-one, dont try to think of a better translation
- Second, you should request this on transifex first
- Third, you can not modify the language files directly, you have to request this on transifex

ghost:
Hi
I dont understand. Can you explaine me...
1. I must change strings in code to add translation
2. I must do translation on transifex but for my country langauge only or fo all language ? Because on transifex i can only download and upload language for my country...

gizmocuz:
Okey, never mind transifex !
There is already a translation for 'Stop' in the system, and the original code was using this, so just use that !

gizmocuz:
You can send messages to us translaters on transifex, in this case you wanted a change in the 'root' English file
Anyway, translate the original word (stop) and use the current (stop) translation word!

ghost:
Okey lets go from begin...
1. I must change code in file xxx.js  
   Ok ? My change in this PR is ok ?
2. Next i send you on transifex message with info which strings should be add to all language and then when you add this i done translation in my country language ?

gizmocuz:
Lets start from the beginning (since this is your third account...)
You want the translation 'Stop Blinds' in the system, request them on transifex
After this is done, and it is inside the domoticz system, then create a new pull request with the JS code to add the translation

ghost:
Ahh ok. I start from transifex... and then PR in JS. Ok i start...

gizmocuz:
Updated translation system with the new word

ghost:
Ok thanks
I create new PR #404 with added translation :)
Thanks for help :)

PR #405, opened 2016-02-01
Add translation for button on tab scenes

ghost:


ghost:
Hi
What with this PR is no good ? I done this the same as for button Stop Blinds...

gizmocuz:
No, you are translating string that are not yet in the language files
so again first transifex, then the rest
i also asked you a question on transifex

ghost:
yes yes ok i see your message on transifex and reply :) Sorry

PR #407, opened 2016-02-02
Correct view tab backup/restore as in other the same tabs

ghost:


ghost:
line

Move line unter title tab. Not under buttons. Now looks as the other tabs.

gizmocuz:
Stop with design/language patches (Again!)

PR #408, opened 2016-02-02
Adds support for other serial-port Z-Wave devices

eskibars:
For example, Razberry on BananaPi / BananaPro binds to ttyS2 by default.  Other RPi-like devices are bound to other ports.

This does add a bit of UI overhead: users are now greeted with more tty\* devices.  However, I believe the tradeoff is worth it in terms of supporting the additional devices OOTB without weird workarounds like linking ttySx to ttyAMAx

The PR also changes "OpenZWave USB" to simply "OpenZWave" to reflect the fact that the serial ports may not be USB.

### DIFF
--- a/main/Helper.cpp
+++ b/main/Helper.cpp
@@ -188,6 +188,11 @@ std::vector<std::string> GetSerialPorts(bool &bUseDirectPath)
 				bUseDirectPath=true;
 				ret.push_back("/dev/" + fname);
 			}
+			else if (fname.find("ttyS") != std::string::npos)
+			{
+				bUseDirectPath = true;
+				ret.push_back("/dev/" + fname);
+			}
 #ifdef __FreeBSD__            
 			else if (fname.find("ttyU")!=std::string::npos)
 			{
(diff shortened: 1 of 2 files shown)


gizmocuz:
That wont work:
1. the word USB in a hardware name causes the hardware editor to display all usb ports

There is also another reason why linking IS good, because when you swap usb devices (sometimes that also happens after a reboot), it all gets mixed up
So linking is very good !
And because we can easily do this, i see no reason for this patch.

Another reason (2):
image

Look on my ubuntu system !, many ttySxxx devices, and i can tell you, there is no USB device in them

But thank you for the patch !

eskibars:
The function is called "GetSerialPorts," which would imply to me that it should be getting serial ports, not USB ports.  The RaZberry device is _not_ a USB device.  It operates on the internal serial ports and does support Z-Wave.  Linking it to something like /dev/ttyUSB is quite incorrect.

gizmocuz:
the Razberry is linked to ttyAMA0 when inserted (or used with openzwave)

PR #411, opened 2016-02-03
Removed forward slash from before hash in ShowLayout 

JohnAdders:
Removed forward slash from before hash on ShowLayout as it was breaking navigation with webroot.

Change is to line 1684
- window.location = '/#' + layout;
- window.location = '#' + layout;


gizmocuz:
Thanks for the patch

PR #415, opened 2016-02-04
Implemented: CounterType 'Time'

EddyK69:
Added a new CounterType 'Time'


gizmocuz:
What does it do, or where can we use this type for ? (hopefully more then just a media center extension?)
Can we use the current clock implement (not used, but we can extend it)

image

Could you replace our current clock with yours, this saves extra data (or use the current clock)

EddyK69:
The intention is to have a counter to keep track of 'time'. Just like how many liter water has been used per day but now ' how many minutes has device x been online'. See screenshot to illustrate the purpose...
Did not know the existence of the Clock-device, but is doesn't have a log-option. 

**Device**
capture1

**Notifications**
capture2

**Logs**
capture3

gizmocuz:
I think there is a small issue in your commit, you are treading the time as water on once stage:

```
                        case MTYPE_TIME:
 +                          sprintf(szTmp, "%.03f min", fvalue / WaterDivider);
 +                          root["result"][ii]["Data"] = szTmp;
 +                          root["result"][ii]["Counter"] = szTmp;
 +                          break;

```

What happens if the counter is 3 days, what will it display ? 3_24_60*60 = 259200 min ?

Is there a 'global' usage for this new type ? Or could you have used a text sensor/user variable ?

EddyK69:
Issue mentioned above fixed. **Will fix non-mergeable issues when you are convinced & willing to merge this PR :D**

Counter will act the same as a water-counter (coutervalue increasing over days, but just show the increments in bars per day, see previous screenshots)
The main difference between this one and the watercounter is the value suffix (minutes instead of m3)

gizmocuz:
Thanks, can you answer the other 2 questions (3 days/global usage), with global usage i mean, do you think 1000 other users will use this ?

EddyK69:
1) Counter is used for example online time per day, so max minutes = 24*60
2) I know a lot of users use the ping-script (or pinger-hardware) to keep track of online devices. This counter can be used in scripts to record this online-time and generate nice graphs. A simple glance at the graph will show any 'glitches' when a device was offline for a while.

EddyK69:
Alternative could be: ability to choose custom value suffix when you selected type Counter...

gizmocuz:
Okey, i still do not see it, but you might got a point ;)
Could you replace our existing clock image with your orange clock image (from the same maker _g_), and use this in your pull request
Is the domoticz.js.gz file the latest? (Just added the option for Scene Logging in it)

EddyK69:
- Renamed icon according to existing clock-icon
- Resolved merge-issue on domoticz.js.gz (is now latest with Scene Logging in it)

Thanks in advance for merging!!

gizmocuz:
Thanks for the patch !

gizmocuz:
We should add this in the manual/wiki ;)

EddyK69:
Happy with the patch, for now :D 
Did some more thinking on your comment of ' general purpose'

I'm making is more general now: add ability to add customize prefixes and suffixes to the value

capture1

So the counter values can be showed as for example:
- Time: xx minutes
- Length: xx meters
- Weight: xx kilo's
- Drinks: xx beers :D 
  etc.

PR #416, opened 2016-02-04
improved formating to make file better readable

saerdnaer:


gizmocuz:
Thanks for the cleanup! Maybe i could ask you for some more changes?
- Maybe we should just link the compilation of every platform to the wiki pages ? (there is one for raspberry,linux,osx,synology,windows?)
- Or maybe make multiple README files, like
  README.osx
  README.linux
  README.raspberry
  README.synology
  README.windows

saerdnaer:
I think you have to decide is this information should be in this repository or in the wiki pages. Keeping both in sync probably doesn't work. In my case the wiki page at https://www.domoticz.com/wiki/Mac_OS_X was more up to date. 

Regarding your request: My tests of domoticz are finished – and it currently looks like my hardware is not supported. So I'm not really motivated to improve the documentation even more, sorry.

gizmocuz:
Sorry to hear, what hardware is that ? Did you consider a raspberry pi ?

saerdnaer:
I'm not talking about the computer which is running the software. In my cases it was a busware CUL usb stick to communicate via 868 MHz/SlowRF with HMS 100 TF temperature sensors.

gizmocuz:
Ok

PR #417, opened 2016-02-04
Panasonic fix

sijones:
Fix for double events when device is detected on.

Change of logic, don't assume device is on and correct just because we connect to a port, we wait until we have a volume reply.

Was a change to header file but changed it back and has gone through as a commit by mistake but no actual change.

### DIFF
--- a/hardware/PanasonicTV.cpp
+++ b/hardware/PanasonicTV.cpp
@@ -218,23 +218,14 @@ bool CPanasonicNode::handleConnect(boost::asio::ip::tcp::socket& socket, boost::
 			if (!ec)
 			{
 				if (DEBUG_LOGGING) _log.Log(LOG_NORM, "Panasonic Plugin: (%s) Connected to '%s:%s'.", m_Name.c_str(), m_IP.c_str(), (m_Port[0] != '-' ? m_Port.c_str() : m_Port.substr(1).c_str()));
-				if (m_CurrentStatus.Status() != MSTAT_ON)
-				{
-					m_CurrentStatus.Clear();
-					m_CurrentStatus.Status(MSTAT_ON);
-					UpdateStatus();
-				}
 				return true;
 			}
 			else
 			{
-				if ((DEBUG_LOGGING) ||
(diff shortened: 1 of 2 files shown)


gizmocuz:
Thanks!

gizmocuz:
There seems to be a memory leak in the Panasonic class

~line 690

> ```
>               boost::thread* tAsync = new boost::thread(&CPanasonicNode::Do_Work, (*itt));
> ```

~line 880

> ```
>       boost::thread* tAsync = new boost::thread(&CPanasonicNode::Do_Work, (*itt));
> ```

I think you based this class on the Kodi class, it has the same problems.

If you would like to fix it, please wait for after my upcoming commit (code refactoring with cppcheck)

sijones:
Yeah I'll wait for you then correct it.

I'll have to wait for the kodi code as well as am not the best in c++ so
need an example!

Cheers,

gizmocuz:
I committed my changes, but i want to review all of them later on github

going to have a look whats happening here... i will make a post here when done...

gizmocuz:
Made a commit... (sorry ... two commits) is it possible you can check if all is still working correctly ?

sijones:
Compiling now.. Will test on Windows and Linux.

sijones:
Seems fine, tested in Win and Linux

gizmocuz:
Thanks for the feedback!
Question, why have you chosen to use the WOL database table to specify your panasonic tv?
I mean, it could also have been a class for 1 panasonic tv, with as settings the IP address.
If you have more panasonic tv's you just add the panasonic hardware multiple times
Then you also have one main thread...

sijones:
There's no reason as such, it was just carried over from the kodi code.

I just don't have enough understanding at the moment of the code to be able
to really question what and why am doing and how it can be done better, as
i get more time am going to go over the code and try and understand it
better.

I did think the kodi way of doing it seemed ok, but now you mention it, it
would have been better to go single hardware route as the amount of TV's is
limited.

Perhaps when i learn a bit more i'll change it!

Sorry if it's a problem!

Issue #258, opened 2015-12-13
Revision on cmake is dependings of git. 


xbeaudouin:
Hello,

I am trying to make a FreeBSD port of domoticz. I use the freebsd port framework that can download a specific tag / version from github

Unfortunatly the tar.gz from github doesn't have .git directory that makes the CMake macro fails like this :

-- Found OpenSSL: /usr/lib/libssl.so;/usr/lib/libcrypto.so (found version "1.0.1p") 
fatal: Not a git repository (or any parent up to mount point /usr)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
CMake Error at CMakeLists.txt:145 (MATH):
  math cannot parse the expression: "+2107": syntax error, unexpected
  exp_PLUS, expecting exp_OPENPARENT or exp_NUMBER (1)

-- Compiling Revision #
-- Boost version: 1.55.0

I have 2 way to fix this :

1- Haxor the CM around lines 135 to 150.
2- Make a quick and dirty tarball from the github and push it into a server somewhere.

The second way is... "usable" but still very dirty.

Is there any good way to fix this ?

Kind regards,
Xavier

dwmw2:
FWIW in the OpenWRT port (https://github.com/openwrt/packages/pull/2007) I used option 1. Although I wasn't stunningly happy about it. I'd just had enough of CMake hacking that week.

In the OpenConnect VPN client I have a trick which generates the revision from git only when building from git. Proper releases have a pregenerated version.h already present in the tarball, generated by the `make dist` step. There's a makefile rule which updates `version.c` **if** any of the other files have changed. And which will make it `-unknown` if it can't find a git repository. See http://git.infradead.org/users/dwmw2/openconnect.git/blob/v7.06:/version.sh and http://git.infradead.org/users/dwmw2/openconnect.git/blob/v7.06:/Makefile.am#l115

Not sure how useful that is for Domoticz which doesn't seem to have the same focus on tarball releases, but you're welcome to lift anything from that if you find it useful. Just having the same _"rebuild appversion.h only if other files have changed"_ logic would at least allow people building from tarballs to drop an appversion.h into place without having to hack the makefiles, as a first step.

xbeaudouin:
So this is really a shame (have same issue with Open-Zwave lib as well).
As discussed with a FreeBSD commiter, dependcy of git "live" access is clairely _forbidden_ during the package building.
[...]

PR #422, opened 2016-02-05
Fix cmake dependencies on git

ffontaine:
Proposal to fix issue #258
Set 0 default values for ProjectRevision, ProjectHash and ProjectDate if source
code has not be retrieved through git clone. Moreover, this patch will add a
default appversion.h.txt with default values. This file should be updated each
time an official release is made. Finally, add ProjectDirty to detect if a
modification has been made locally, in this case "-modified" will be added to
hash.


gms1:
+1 
additional: because of this dependence to git the out-of-source build is broken

gizmocuz:
Hmmmm yes... but that means we have to update this file by hand each time ?
I agree this is needed, if the source is not retrieved from git (but it should?)
I can't think of on offline build system that needs to 'upload' the packages ?
Then it's online, and why not clone/fetch from git

And if we make a new script (like makebeta/makerelease) that pulls the code from git, generates that version file, and nothing more
Then you can transfer it to the offline/online build system ?

If all this does not make sense we merge the patch of course ;) But who's going to maintain this version file ?

ffontaine:
If you retrieve the source code from git clone, the patch will keep the current behavior. The values will be retrieved from git and the appversion.h.txt will be updated automatically.

You will only have to update this file "manually" when tagging an official release to be sure than the values in this file are up to date.

Thanks to this patch, if the source are retrieved out of tree (for example buildroot use this kind of mechanism), the compilation will not fail and the buildsystem will have the version number and date of your latest official release.

gizmocuz:
Thanks, hope i won't forget when we do a new release ;) (Scheduled for 29th feb)

gizmocuz:
This is not working, output of build:

-- domoticz has been modified locally: add "-modified" to hash

why.. because:
appversion.h.txt

I will remove appversion.h.txt from the repo
If you want to use something like this, it should not be a file that the build system is generating
You could name it for instance appversion.default

PR #423, opened 2016-02-05
Fix bug power display with 'Tempo' contract

Pigloox:
Show the 3 meter of the tempo contract with the correct values


gizmocuz:
Thanks for the patch !

Issue #410, opened 2016-02-02
unable to configure install path: bug has been introduced by #318


gms1:
currently the setting for CMAKE_INSTALL_PREFIX gets ignored (even on first cmake run):

```
gms@sirius:~/work/Domoticz/domoticz$ cmake . -DCMAKE_INSTALL_PREFIX=/usr/local/domoticz &>/dev/null && grep CMAKE_INSTALL_PREFIX CMakeCache.txt

CMAKE_INSTALL_PREFIX:PATH=/opt/domoticz`
```

if you only want a different default value ( e.g /opt/domoticz ), but want to let the user configure his preferred value, please use:

```
IF(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
  set(CMAKE_INSTALL_PREFIX "/opt/${CPACK_PACKAGE_NAME}" CACHE PATH "Install path prefix, prepended onto install directories." FORCE)
ENDIF(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
```

gizmocuz:
Thank you
Could you make pull request ?

gms1:
no problem, I will do it this weekend

PR #425, opened 2016-02-05
fixed issue #410: the install prefix is now configurable again

gms1:
the installation prefix is now configurable again:

```
gms@sirius:~/work/Domoticz/domoticz$ rm CMakeCache.txt
gms@sirius:~/work/Domoticz/domoticz$ cmake -DCMAKE_INSTALL_PREFIX=/usr/local . &>/dev/null && grep CMAKE_INSTALL_PREFIX CMakeCache.txt

CMAKE_INSTALL_PREFIX:PATH=/usr/local
```

and the default installation prefix is '/opt/domoticz':

```
gms@sirius:~/work/Domoticz/domoticz$ rm CMakeCache.txt 
gms@sirius:~/work/Domoticz/domoticz$ cmake . &>/dev/null && grep CMAKE_INSTALL_PREFIX CMakeCache.txt
CMAKE_INSTALL_PREFIX:PATH=/opt/domoticz
```

best regards
Guenter


gizmocuz:
Thanks for the patch !

ffontaine:
With this patch, if install prefix is set by the buildsystem for example to /usr then domoticz will install all its directories and files (domoticz, www, License.txt, History.txt, scripts, Config, updatedomo) directly into /usr. I think it would be better to keep the subdirectory domoticz to keep all the files/directories in a single directory especially for the www or scripts directories which are very commonly used. Do you agree?

gizmocuz:
Yes, we also have command line options for this

PR #426, opened 2016-02-05
Fix random behavior on Piface

badzz:
Clean up the spi structure before sending. This fixes on my system a lot of spurious issues and random piface detection issues


gizmocuz:
Thanks for the patch, yes that would help indeed ;)

PR #427, opened 2016-02-05
Fixed: SQL error when starting with new DB

EddyK69:
After implementation of the extended schedules for SetPoints, this bit of code was forgotten ;)


gizmocuz:
Thanks!

PR #428, opened 2016-02-05
Fix lost domoticz.js changes for Panasonic Remote

sijones:


sijones:
Looks like changes have been lost due to it being a compressed file and a compare can't be done so someone else could have checked in an older version and overwritten my changes. 

Not sure if there is something we can do to stop this from happening?

nayrnet:
yeah I thought the same thing, perhaps we store those files uncompressed here on github and then have the build scripts re-compress them? storing compressed and uncompressed side by side might be a problem though.

sijones:
I think that would be a lot better idea!

gizmocuz:
Your right, that probably wont be an easy job
At the moment when there are multiple PR against the domoticz js file, this goes wrong

We also have to do this on travis/appveyor

gizmocuz:
merged, new js file will be committed soon

PR #429, opened 2016-02-05
Fix for missing type & value while edting Counter Notifications

EddyK69:
See screenshot...
capture


sijones:
This changes the domoticz.js file which could possibly overwrite my changes in my pull request, please check before merge!

gizmocuz:
merged, new js file will be committed soon

Issue #435, opened 2016-02-07
building with USE_BUILTIN_MQTT set to false leads to inconsistency


gms1:
an external "mosquittopp.h" is referenced during the compilation of "hardware/MQTT.cpp"
but instead of linking the corresponding "libmosquittopp.so", the internal "MQTT/mosquittopp.cpp" ( referencing the internal "MQTT/mosquittopp.h" ) gets compiled and linked ( using the external "libmosquitto" )

I will make a pull request... I think the existing use-flag "USE_BUILTIN_MQTT" should be used for both "libmosquitto" and "libmosquittopp". because of their strong dependency; do you agree?

gizmocuz:
yes, thanks! Someone changed this recently

gms1:
fixed by pull request #436

PR #436, opened 2016-02-07
fix for #435 : building with USE_BUILTIN_MQTT set to false leads to inconsistency 

gms1:


gizmocuz:
Thanks for the patch!

PR #438, opened 2016-02-08
Force static build of libraries 

ffontaine:
This patch add the STATIC keyword to cmake ADD_LIBRARY functions. Thanks to this, we're are sure that cmake will build a static version and not a dynamic version of the builtin libraries.


gizmocuz:
Thanks!

ffontaine:
I made an horrible mistake when pushing my pull request, as I have also pushed the "merge remote-tracking branch upstream/master", you should just apply the first patch and revert the second one. Sorry for this ...

gizmocuz:
Okey, maybe check the current master and make a new pull request to correct things ?

gizmocuz:
Thanks!

PR #439, opened 2016-02-08
Add appversion.default file to save default values

ffontaine:
Use a dedicated file appversion.default to save default values for revision, hash and date instead of using appversion.h.txt as this file is also generated by domoticz.


gizmocuz:
Thanks, had a small issue with this last week too ;)

PR #441, opened 2016-02-08
Fixed: Preserve DevicePlan & SharedDevices when transferring device

EddyK69:
- Bugfix: In method TransferDevice -> LightSubDevices did not set ParentID
- Bugfix: In method DeleteDevices -> Percentage & Percentage_Calender were not cleared
- Fixed: Preserve DevicePlan when transferring device
- Fixed: Preserve SharedDevices when transferring device


gizmocuz:
Thanks!

PR #442, opened 2016-02-08
Force static build of libraries

ffontaine:
Add STATIC to cmake add_library function to force the static build of mqtt,
sqlite and lua libraries

### DIFF
--- a/MQTT/CMakeLists.txt
+++ b/MQTT/CMakeLists.txt
@@ -31,4 +31,4 @@ IF(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
 	SET(OperatingSystem "Linux")
 ENDIF(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
 
-ADD_LIBRARY(mqtt ${LIB} ${CORE})
+ADD_LIBRARY(mqtt STATIC ${LIB} ${CORE})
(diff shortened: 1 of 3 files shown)


gizmocuz:
What is the difference with #438 ?

ffontaine:
There is no more reference to "@ffontaine   Merge remote-tracking branch 'upstream/master'  
f5ea6f3".

gms1:
I am very sorry, but I think the failing of AppVeyor has todo with my last change from yesterday, but can not imagine why it has not been recognized already yesterday
"MQTT\mosquittopp.cpp" should not be compiled using the precompiled header ( see domoticz.vcxproj )

unfortunately I cannot do the pull-request from here. Sorry!

gms1:
it seems the pull-request build has been cancelled:
https://ci.appveyor.com/project/gizmocuz/domoticz/build/1.0.4236
... and the merge-commit build has failed afterwards:
https://ci.appveyor.com/project/gizmocuz/domoticz/build/1.0.4237

gms1:
@ffontaine 
basically you have done a rebase by reverting the changes and reapplying them on top of upstream/master; considering that the only advantage of rebasing should be a simpler history, that was not worth the hassle

gizmocuz:
Do we need to merge this ?

gms1:
BUILD_SHARED_LIBS should be false by default, so the STATIC keyword shouldn't be necessary:
https://cmake.org/cmake/help/v3.0/variable/BUILD_SHARED_LIBS.html

But I think it does not hurt to be more specific in this case. Users wanting to have shared libraries should use the libraries provided by the system, the domoticz internal libraries are not meant to be shared by other applications
So yes, please merge this

PR #445, opened 2016-02-08
Optimize login/logout processes and session management

gaudryc:
I tried to make the code more simpler and more efficient and I took the opportunities to make some fixes on header parsing.

Now :
- sessions are cleaned every 15 minutes (both in memory and in database) using a separated thread ;
- the validity duration of long term cookies is 30 days ;
- an unauthorized response implies a cookie removal ;
- only one place where the cookie is parsed ;
- the logout command really do something (like login) ; 
- no basic authentication loop where the user logged out and cancel the authentication form.


gaudryc:
Do you know why AppVeyor had failed without any special error ?

gaudryc:
Job is done for this branch.

gizmocuz:
Thanks, appveyor has a problem with the new MQTT patch (that i will resolve in a few minutes)
Thanks for the patch !

gizmocuz:
I think we should stop with the basic-auth, as most browsers will drop support for this soon....or is this to soon ?

gaudryc:
Do you want me to remove it this week ?

gizmocuz:
Well... maybe we should ask in the developers forum section how others feel about this

gaudryc:
OK, I'm waiting your "GO GO GO" ;-)

PR #447, opened 2016-02-08
Fix detection of system Lua

clandmeter:
Some systems has version suffix to the lua pkg-config. we also need include the library directory when linking.


gizmocuz:
Thanks!

PR #448, opened 2016-02-08
MySensors USB gateway to support 38400 bps serial speed

ToniA:
If the serial gateway is built on SenseBender or Arduino Pro Mini 3.3V,
the serial port speed must be limited to 38400. This is the highest
reliable speed on ATmega processors running on the internal 8 MHz oscillator.

With this, the entire MySensors serial gateway fits inside the Raspberry Pi case (using SenseBender, nRF24 radio and 4 DuPont wires).


gizmocuz:
Thanks for the patch !

PR #452, opened 2016-02-09
Spelling correction: "childs" -> "children"

mbardea:
... on "hardware" page.


gizmocuz:
i think both can be used ? Specially when talking about node's and child id's

EddyK69:
I agree with @gizmocuz

mbardea:
The change was for actual words which seem to show on a web page. "Childs" is broken English.

If the change does not affect the web page directly, please ignore.

gizmocuz:
The (big) problem is, it needs to be changed in the web page, we need to add a new translation word, every translator has to translate this word, then we need to build a new version, and then its visible at the end user

PR #455, opened 2016-02-09
Small patch for setup.html

RafiiB:
Correct position
Change class to click Here
Correct size before button TEST in Custom HTTP
Correct color link in Roomplan
Correct line under button backup and restore-move on top button
Add translation to strings MyDomoticz

### DIFF
--- a/www/html5.appcache
+++ b/www/html5.appcache
@@ -1,5 +1,5 @@
 CACHE MANIFEST
-# ref 1201
+# ref 1202
 
 CACHE:
 # CSS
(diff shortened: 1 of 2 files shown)


gizmocuz:
Thanks for the 'small patch' but look at the files changed log:

966 additions, 978 deletions not shown because the diff is too large. Please use a local Git client to view these changes.

That should not be the case, please do not reformat, only make changes where you need them.
Could you adjust this ?

RafiiB:
Ok thanks i create it in new PR #460

PR #457, opened 2016-02-09
New Lua table 'otherdevices_idx'

ToniA:
Generally the devices are known by their name in Lua event scripts, like:

```
otherdevices['name_of_my_device']
```

However, when communicating the status changes back to Domoticz, in some cases only the ID
values can be used. Like, when updating a 'text sensor' device with ID '123' (there is no easy way to connect the sensor name to idx):

```
commandArray['UpdateDevice'] = "123|0|Your text here"
```

This is very cumbersome, as there's no way to get the IDX values in Lua.

With this change, the IDX values can be extracted from the table called 'otherdevices_idx':

```
idx = otherdevices_idx['text sensor']
commandArray['UpdateDevice'] = idx .. "|0|Your text here"
```

or shortly:

```
commandArray['UpdateDevice'] = otherdevices_idx['text sensor'] .. "|0|Your text here"
```

See also these:
https://www.domoticz.com/forum/viewtopic.php?t=6102
https://www.domoticz.com/forum/viewtopic.php?t=4347


ToniA:
... and I have no idea why this wasn't built automatically by Travis & Appveyor.

ToniA:
... but adding a new merge to domoticz/master kicked off a build ...

PR #458, opened 2016-02-09
Under FreeBSD pkg_config name for LUA is lua-5.2.

xbeaudouin:
Hello,

This trivial patch, allow LUA package to be discovered by the CMake.
Since LUA doesn't provide an official pkg-config configuration file, some os/distribution give their own one.
Regards


gizmocuz:
Thanks!

PR #461, opened 2016-02-10
Secure and optimize webserver code

gaudryc:
The goal of this pull request is the following :
- Remove unnecessary session attribute _tWebEmSession::lastRequestPath (used to check the request URI).
- Optimize the reading method of the database backup file by allocating memory before reading the file and set it as the reply content.
- Optimize the methods webem_action_function, webem_include_function and webem_include_function_x execution, by replacing the return value with an argument passed by reference.
- Remove the unnecessary attributes cWebServer::m_retstr and cWebServer::m_wretstr used to received command and action result (these attributes were virtually shared by all connections on the same listening port).
- Remove unnecessary session attribute _tWebEmSession::outputfilename (used to set content disposition header in the HTTP response).
- Optimize the reading method of the '/ozwcp/cp.html' file by allocating memory before reading the file and set it as the reply content.
- Optimize the reading method of the zwave configuration file by allocating memory before reading the file and set it as the reply content.
- Refactor the http::server::reply structure and add some helper methods.


gaudryc:
Do we really need these functions : cWebem::RegisterIncludeCodeW and cWebem::RegisterPageCodeW ?

gizmocuz:
Great patch, yes i need them

gizmocuz:
Huge patch !, Look alright ! going to merge

PR #467, opened 2016-02-12
Fixed Import of Old Month Data and added some logging to the process

qvjester:


gizmocuz:
Thanks for the patch, next time, please don't change the formatting

qvjester:
Was there a reason that there was a chunk not formatted ? Kinda hard to figure out what was going on with no indents.

gizmocuz:
I was indeed referring to the formatting, that was there under visual studio the last time i checked
But because there where not to many changes it was still doable to see what had changed
(i had worse... someone changed a few lines, and committed a file with 998 changes because of formatting)

Issue #464, opened 2016-02-11
OZW Config path is relative to StartupFolder


clandmeter:
I am packaging domoticz on Alpine Linux. To honor FHS as much as possible, I have used the userdata and the wwwroot switch and moved the domoticz binary to usr/bin. Looking at the code OpenZwave needs two directories, one for static config files, and one to store log and zwave data but the static directory will always be relative to the StartupFolder. If an option would be available to specify OZW config it would be prefererend to hold static files inside /usr/share/domoticz/zwave, and the dynamic zwave data would likely to go into /var/lib/domoticz/zwave.

Is there a way this could be implemented?

gizmocuz:
I think it is best if you contact jumbotrol on our forum as he is the maintainer of the domoticz package

clandmeter:
Im not sure you understand my point (or maybe i dont) but I'm not using his package but maintain it for Alpine Linux https://pkgs.alpinelinux.org/package/testing/x86_64/domoticz

buildfile: http://git.alpinelinux.org/cgit/aports/tree/testing/domoticz/APKBUILD

gms1:
quote: I am packaging domoticz on Alpine Linux. To honor FHS as much as possible...

then the easiest way would be to keep the default "/opt/domoticz"

otherwise wouldn't this do the trick:

> cd /usr/share/domoticz/
> /usr/bin/domoticsz -userdata /var/lib/domoticz/

clandmeter:
/opt i not an option,  because Alpine Linux build system does not allow anything under /opt nor /usr/local

As for current working dir, this is not honored by domoticz OpenZwave implementation as seen here:

https://github.com/domoticz/domoticz/blob/master/hardware/OpenZWave.cpp#L954

I see 3 issues with current implementation:
1. ConfigPath is hard-coded as seen above. (from what i can see its the last part which is hard coded)
2. The current Config directory holds both static and dynamic files, and the FHS provides /var/lib/projectname for files that change regularly, and /usr/share/projectname for static files.
3. A directory named Config in the root of default installation is confusing, its (currently) only used for OpenZwave config.

I could probably work out a patch with would support --ozwconfig switch which would set the ConfigPath and if both ozwconfig and userdata is specified, we could store zwave data in $userdata/zwave. This would also make it backwards compatible for people who never used the ozwconfig switch.

gms1:
ok, I thought the startup folder is the working directory, but it is not
wouldn't this work:
[...]

PR #468, opened 2016-02-12
Add option to specify OpenZwave config directory

clandmeter:
The config directory is one of the last static directories still located the userdata folder.
This patch will make it possible to move the static files from the Config directory to the specified directory. 
I'm not really happy with this approach because I would prefer to create a new directory inside the UserData directory to store the OpenZwave log and network state files, but I find no suitable crossplatform way to create  the directory (except boost, but i dont want to depend on a module just to do that). I guess the proper way would be to define these in the cmake and let cmake take care of the creation if necessary.

This has been discussed at #464


gizmocuz:
It should go directly in the root of the szUserDataFolder indeed

I see the option was already there in our OZW implementation,
this patch was not needed

PR #469, opened 2016-02-12
Fix asyncserial port stop

gaudryc:


gizmocuz:
Is it possible to rename stopIfOpen to Close ?

gaudryc:
If I rename it "Close" I will bring some confusing with the existing ::close method.

I think I have 2 solutions :
1 - rename it "cleanClose"
2 - move the code from the ::stop method to the ::close method and delete the ::stop method.

gaudryc:
After thinking, I should :
- add a protected method "init" to the callback method (and make the setReadCallback method private) ;
- rename "stop" with "dispose" because it removes the callback method pointer, and make the clearReadCallback method private.

gaudryc:
We can do everything but we have to be clear on what is doing what.

gizmocuz:
Well what you want, is just 'close' the serial connection
maybe the real 'close' function should be patched then ?

gizmocuz:
Stop if Opened does not "sound" nice, thats what a close function should do
if open -> return, no need to close
close

gizmocuz:
Sorry, i want to merge, but we got a few conflicts

gaudryc:
Sorry for the delay, I'm refactoring the webserver code to be able to add more SSL parameters from the command line.

gizmocuz:
Thanks!

gizmocuz:
Huston, we seem to have an error. (verfied)

See https://www.domoticz.com/forum/viewtopic.php?f=3&t=10843&start=60

Postby tonadam » Wed Mar 09, 2016 9:26 am 

> Thanks to all who worked on Domoticz, great software.
> 
> One issue I found today with this release, when I disconnect my Arduino connected with USB running Rflink firmware, Domoticz crashes.

I will commit a patch for this, it is one change:

void AsyncSerial::readEnd

calling terminate with no arguments (true)

Issue #430, opened 2016-02-06
1wire: Reading sensors values from OWFS sometimes fails


Erik-NA:
Using owfs 3.0p2 with a onewire hub (https://www.hobby-boards.com/store/pages.php?pageid=8) together with humidity/temp-sensors based on DS2438.

Sometimes Domoticz cannot read from owfs to retrieve sensor values. It works from command line, so it must be a problem related to the onewire implementation in Domoticz? 
A restart of the onewire hardware device in Domoticz usually fixes the problem, until next time it fails to read sensor values. If does not work I must restart my raspberry PI.

Another related issue is that when restarting Domoticz, Domoticz sometimes fails to start until owfs is restarted.

gizmocuz:
Did you try with 3.1p1 ?

Erik-NA:
Same thing with owfs 3.1p1 as with 3.0p2 on Domoticz v2.4439. 
Currently Domoticz cannot read any values from DS2438-based sensors on both versions. (Sensor type 26 in owfs) Temperature devices (type 28) are working.

gizmocuz:
Did domoticz support type 26 in the first place ? (sorry i am not the author of the 1wire class)

szczukot:
I see, that this device is supported by 1wire (DS2438 is Smart Battery Monitor).
Erik-NA: please give me info about : 
ls /mnt/1wire/uncached/26.xxxxxxxxxxx

Erik-NA:
In Domoticz the sensors was last seen 2016-02-03. I do not know which Domoticz version, but I update daily if there is a update available.
what kind of information? For some reason I cannot attach a zip file to this issue :(

Erik-NA:
[DS2438.zip](https://github.com/domoticz/domoticz/files/128158/DS2438.zip)
Yes, here it is!

Erik-NA:
Here is one uncached: 
[uncached.zip](https://github.com/domoticz/domoticz/files/128273/uncached.zip)

szczukot:
Did You try set verbose to 1 ? (./domoticz -verbose 1)
Pls set it and show log.

Erik-NA:
[domolog.txt](https://github.com/domoticz/domoticz/files/129258/domolog.txt)
Here is a log file using Domoticz V2.4552 with verbose set to 1

szczukot:
Do you see on this list your device type 26 ?

Erik-NA:
[...]

PR #472, opened 2016-02-12
More logs for devices

szczukot:
What Do You think about this change ? 
This gives more possibilities for logs (Debug). Not only in mainworker, but in all modules.
I wrote this for testing for example this problem : https://github.com/domoticz/domoticz/issues/430#issuecomment-183398606


gizmocuz:
I dont know why this patch was made, but 1=status

gizmocuz:
I have to check this, because we have also a status level, and i dont see that here
But for the OWFS you can enable debug, that was the idea with verbose

gizmocuz:
I think verbose debug should be sufficient to debug things like this ?

szczukot:
Define DEBUG is not solution for this situation.

Im dont testing this in my system. This problem have one of users.
Maybe i enable DEBUG, compile and send app for him

szczukot:
Ok. I did only optimization for reading sensor (1wire).
Pls merge it (but i dont know, how i can set properly one inlcude - Travis is failed)

gizmocuz:
Well above (or below) you can click on the details of travis and see why there is a problem
i can not merge a pull request when our build system fails, thats why we have some build systems

`In file included from /home/travis/build/domoticz/domoticz/main/mainworker.cpp:36:0:
/home/travis/build/domoticz/domoticz/main/../hardware/1Wire.h:4:34: fatal error: ..\1Wire\1WireCommon.h: No such file or directory
compilation terminated.
make[2]: *** [CMakeFiles/domoticz.dir/main/`

szczukot:
I see this error, but i don;'t know, how path is properly :(
I tested:
1) "..\1Wire\1WireCommon.h"
2) "1Wire\1WireCommon.h"
Both is inproperly

szczukot:
Maybe now is ok :)

gizmocuz:
Thanks, this is working !

PR #474, opened 2016-02-13
Fix missing line in ShowMediaRemote

sijones:
Left this line out of domoticz.js which meant the remote doesn't send command, not sure how I missed this!


gizmocuz:
Thanks for the patch, i will make a new commit soon with the added line

Issue #473, opened 2016-02-13
Selector switch messed up Smoke detector settings


pepijng:
The new (great) selector switch messed up the configuration page for a Smoke detector type.

![schermafbeelding 2016-02-13 om 11 27 49](https://cloud.githubusercontent.com/assets/6169050/13027177/e1fa08c2-d244-11e5-8dc8-dae6bffd9896.png)

gaudryc:
Could you please give me a print screen of your smoke detector device in the Devices Tab ?
I'd like to know the device type and sub type.

pepijng:
Sure :)
![schermafbeelding 2016-02-13 om 22 17 28](https://cloud.githubusercontent.com/assets/6169050/13030106/b6f3ca2c-d29f-11e5-87b4-ef6d65e928c5.png)

PR #480, opened 2016-02-14
Fix Security devices edit form by hiding Selector switch options (#473)

gaudryc:


gizmocuz:
Thanks for the patch
There is no need to update the history file, as there will be only a single line in the new release, Implemented new Selector switch

Issue #459, opened 2016-02-10
expand json api for selector switch names


nayrnet:
this is an enhancement request to extend the API to include level names, for example the MQTT API publishes the following JSON for a selector switch change that includes the switch name and level, but no indication of what friendly name that level has configured.

`{"Battery":255,"RSSI":12,"dtype":"Light/Switch","id":"000140E0","idx":145,"name":"TV","nvalue":2,"stype":"Selector Switch","svalue1":"40","switchType":"Selector","unit":1}`

Request it to be extended to include an element with levelName, as such:

`{"Battery":255,"RSSI":12,"dtype":"Light/Switch","id":"000140E0","idx":145,"name":"TV","levelName":"Security Cameras","nvalue":2,"stype":"Selector Switch","svalue1":"40","switchType":"Selector","unit":1}`

For example with my Pioneer AVR interface I include the level name in an array so I can have a friendly log name.. with the name exposed via JSON I could simply map that string to the input level and people can change the name alone through the UI.

I imagine custom interfaces using the JSON API would like to know what the level names are instead of a numerical value.

gizmocuz:
Your "Name" field is very unfriendly, how about making this "Friendly" ?

gaudryc:
It seems I have missed something while implementing the Selector switch.
Can someone explain me how MQTT works ?

nayrnet:
MQTT uses the JSON API but instead of using HTTP requests for input and event scripts for return calls it uses a bi-directional PubSub broker so you device updates pushed out in realtime. Its really nice because you dont have to poll domoticz or maintain a bunch of events to keep external integrations in sync.

incoming commands look like: **domoticz/in**
_{"command":"switchlight","idx":145, "switchcmd":"Set Level 40"}_

and all updates come unsolicited like this: **domoticz/out** _{"Battery":255,"RSSI":12,"dtype":"Light/Switch","id":"000140E0","idx":145,"name":"TV","nvalue":2,"stype":"Selector Switch","svalue1":"40","switchType":"Selector","unit":1}_

Its missing the Level Names being exposed like Lua has, so I cant do logic like 
 _if (json.LevelName === 'Security Cameras') doSomeThing();_

[...]

PR #482, opened 2016-02-14
Fix missing Selector switch levels in MQTT device info response by returning all device options (#459)

gaudryc:


gaudryc:
Thank you for patching my patch. I sent it too quickly.

PR #483, opened 2016-02-15
Fix hardware/1Wire.h compilation issue

gaudryc:


gizmocuz:
Thanks for the patch!

PR #484, opened 2016-02-15
Fix Win32ish #include

xbeaudouin:
# include ..\dir is Win32, under *nix this is not handled by compiler -> fix


xbeaudouin:
Ignore this pull request, the #483 fix this :)

PR #486, opened 2016-02-15
Fix detection of libtellduscore.so 

xbeaudouin:
On some OS libtellduscore.so is in /usr/local/lib.
This fix this and use CMake to find the right place where it is located. Allowing out of source build under FreeBSD.
Notice that pull request #483 have to be pushed to fix the 1Wire stuff that fails domoticz to be built.


gizmocuz:
Thanks for the patch !

PR #491, opened 2016-02-17
[1WireByOWFS] Fix for reading to much devices on start

szczukot:


hardware/1Wire.cpp
```diff
@@ -188,10 +188,11 @@ void C1Wire::GetDeviceDetails()
 	if ((m_devices.size() == 0) || (m_sql.m_bAcceptNewHardware))
```
gizmocuz, line 188: this can always be true:

m_sql.m_bAcceptNewHardware

It looks like you now are always clearing the m_devices array ?

instead of m_devices.size()==0, could you write m_devices.empty()

Maybe it should be:

`if (m_devices.empty() && m_sql.m_bAcceptNewHardware)`

szczukot, line 188: No.
Before, devices always deleted and added.
Now, devices delete/add only when 
1) I do not have any (most likely the first run etc)
2) I allow to add new devices in settings
When in setting I disable "allow new devices" - this method call only one time (when program starting).
When "Allow new devices" is enabled - this working as before (always delete/add devices to list)

gizmocuz:
Thanks! Beta 4621 made for the raspberry

multinet33:
Thanks :)

PR #494, opened 2016-02-18
Generic handling of all lights and switches registered on Hue bridge

arjanwaard:
All lights and switches should now be correctly identified and added to Domoticz independently of name.


gizmocuz:
Thanks for the patch !

PR #496, opened 2016-02-19
Next refactoring for sDecodeRxMessage

szczukot:


gizmocuz:
It is nice that you rebuild all hardware classes to use the functions from the base hardware class (SendUV...), but did you actually checked that this generates the exact same sensor with the exact same type/subtype/ ID unit 
Because if not, a lot of people will get new devices

gizmocuz:
And why did you now pass a std::string to the sDecodexxx function? It needs a constant char
also the build system is failing now

.\hardware\RFLinkBase.cpp(778): error C2664: 'void CDomoticzHardwareBase::SendPercentageSensor(const int,const int,const int,const float,const char *)': cannot convert argument 5 from 'std::string' to 'const char *' [C:\projects\domoticz\msbuild\domoticz.vcxproj]
  ..\hardware\RFLinkBase.cpp(778): note: No user-defined-conversion operator available that can perform this conversion, or the operator cannot be called

tomaszszczukobls:
Yes. I change only for this same type/subtype and set properly id/intval1 etc.
Please check diff

I pass const char_.
Many method (Send_****Sensor) in DomoticzHardwareBase using std::string. But this is not good. I changed it in refactoring

tomaszszczukobls:
Target, cool to will be sDecoreRxMessage private.
And all modules will call only Send****Sensor.
The easier, less code and safer

gizmocuz:
Again, did you actually checked that this generates the exact same sensor with the exact same type/subtype/ ID unit 
Because if not, a lot of people will get new devices

szczukot:
Ok. I found one diff for pressure (bad id) for OTWG. Corrected.

gizmocuz:
Sorry, you do not understand what i mean, and for sure we are not going to change the default SendPressureSensor function.
If hardware sensors get a new ID because it now uses the internal send methods,
then we should patch the current hardware/sensors in the database

It needs to be 100% checked.
That means that you need to own the hardware or a database where this should change

So for now, i wont take this patch, and as i already started this some time ago, i will do this eventually for the other hardware classes
New hardware classes should use the new methods

szczukot:
OMG, i dont understand :(
Why id now may be different as before ?
I calling method, which is exactly the same as before.
I tested (with debug) this changes (from odule 1wire and SatelIntegra). Operate the same because nothing is changed.
Pls give me a sample, why this is may be wrong ?

gizmocuz:
Did you test with an OTGW for example ? I can, and until i know 1000% sure no id's will change, then i will replace this function

if you have changes to satel integra or 1wire, i am happy to accept, but make another pull request and leave other changes out for the moment

gizmocuz:
And again, we are not going to modify 'SendPressureSensor'

szczukot:
But what i changed in SendPressureSensor ? Only std::string to char*. Not all module set DefaultName. How to pass NULL in old version of SendPressureSensor ?
And please again - give me a sample, why this ID may be inproperly ?? If You show me, i try resolve this problem.

gizmocuz:
First, SendPressureSensor has this as argument (and we are not going to change this)

SendPressureSensor(const int NodeID, const int ChildID, const int BatteryLevel, const float pressure, const std::string &defaultname)

In your commit, it has an additional parameter (something with ID or whatever) (, const int ID)

I mentioned this now 3 times

All hardware should provide a default name, in this case, the minimal name could be "Pressure" ?!?

Also what is this

```
SendPercentageSensor(static_cast<int>(1), 1, 255, float(Efficiency), "Efficiency");

static_cast<int>(1)

```

Anyways, please stay at satel and 1wire, i promise to look at changing the hardware so it makes use of the base commands soon (maybe even today), but i want to do this with real data, and testing, and i am not going to accept a pull request (from anyone) without being tested 100%

gizmocuz:
This is an example what i mean when things also need to be patched in the database

https://github.com/domoticz/domoticz/commit/75a287868d21fde252f7bfbdb096b9e1237a546e

this was one of the easiest, and a lot more are to be done

szczukot:
"i promise to look at changing the hardware so it makes use of the base commands soon (maybe even today), "

Super. This is what I ment :) If I did it badly, cool as you do it. It has to be simplified because it becomes difficult to maintain.

gizmocuz:
The (biggest) problem will be when people use more then 1 domoticz (one master and several slaves), you need to update the slaves as well

PR #499, opened 2016-02-20
Fix a security issue (DOS) in webserver

gaudryc:
Fix a security issue (DOS) in webserver to prevent a remote_endpoint exception to be thrown to the WebServer class causing the webserver to be down (no possible stop and no possible connection). Now, the webserver can be scanned using the nmap tool (for the SSL configuration purpose).


gizmocuz:
Thanks!

PR #501, opened 2016-02-21
Make HardwareMonitors works on FreeBSD

xbeaudouin:
Hi,
This pull request add support for FreeBSD HardwareMonitor.
This depends of a package on FreeBSD libsysinfo that "emulate" the sys/sysinfo.h.
Regards


gizmocuz:
Are you sure you need an extra library ?
Is

`#include <sys/sysinfo.h>`

Not available per default ? (like on (all) linux systems)

xbeaudouin:
<sys/sysinfo.h> is a linux only header. So it is available only on Linux.
Fortunatly on FreeBSD there is a third party package that permit the emulation of this.
We can say same remark about /proc/<files> on Linux that are not available everywhere (working on an adaption of this RSN).

gizmocuz:
Did not know this... hmmmm
Best if we all go to debian then ! _g_
Thanks for the patch!

nayrnet:
maby he is? https://www.debian.org/ports/kfreebsd-gnu/
:smile:

PR #502, opened 2016-02-21
Added support for HTU21D(F)/SI702x Humidity+Temp I2C sensor

jvandenbroek:
Used this datasheet and used the BMP085 code as basis: https://www.adafruit.com/datasheets/1899_HTU21D.pdf
Tested with SI7021 sensor, which is 100% compatible with Adafruit's HTU21D.

You've very active today in the source, so I had the merge a few times and fix manually ;) I need to make a way to cross compile, the RPi 1 is sooo slow every update :(


gizmocuz:
Thanks for the hard work!

Few comments:
- Could you extend the bmp085 class ? Maybe even rename it later, because this class is for i2c
- Please use the send functions in CDomoticzHardwareBase, like SendTempHumSensor

Else, did you had a look at the mysensors.org project ?

gizmocuz:
The reason for mysensors.org is that we do not want 1000 of sensors being supported in an i2c class that only works on specific systems
With mysensors.org we support all systems

jvandenbroek:
To be honest, no, I've the I2C sensor attached directly to my Pi, so haven't considered MySensors for this.. I'll check out the bmp085 class, see if this can be extended easily.

gizmocuz:
Would be great if you can add support for it in the bmp085 class
But i think i can count the amount of users using a raspberry pi with i2c with domoticz on 2 hands
There are far more users using mysensors.org (you can even use your sparkfun breakout with it)

nayrnet:
sweet thanks, I am using bmp085 on i2c on all of my beaglebones, honestly pressure is not all that useful to me, but humidity would be very much so.. I'll check these out.

gizmocuz:
you can also use a dht22 and for pressure a bmp180

jvandenbroek:
I've hacked something together to use just one class for both BMP085 and HTU21D, renamed the old BMP085 class to I2C. I intent to clean up the code when I have more time, but for now I think this works well. Could it be merged in this state? Thanks!

gizmocuz:
Thanks for the patch !
I will clean it up a little bit (also removing the unneeded sTypeHTU21D, and fixing the visual studio project)

PR #503, opened 2016-02-21
Pifix

rkluwen:
This patch solves the signal 6 received crash upon exiting Domoticz on a raspberry pi.
At the same time, CProxyClient is made a shared pointer class, like it should be.


gizmocuz:
Thanks!!

PR #505, opened 2016-02-21
added support for Legrand/Bticino openwebnet : blinds via ip gateway

steplebr:
Hello,
I've created a new hardware for Legrand/Bticino Openwebnet IP gateway. As I own blinds, I've implemented blind control with the gateway.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -263,6 +263,7 @@ hardware/NefitEasy.cpp
 hardware/Nest.cpp
 hardware/Netatmo.cpp
 hardware/OpenWebNet.cpp
+hardware/openwebnet/bt_openwebnet.cpp
 hardware/OpenZWave.cpp
 hardware/openzwave/control_panel/ozwcp.cpp
 hardware/openzwave/control_panel/zwavelib.cpp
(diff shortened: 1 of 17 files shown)


gizmocuz:
Thank you for this great patch!
Could you change the following to a pTypeGeneralSwitch/sSwitchBlindsT1

{ pTypeBlinds, sTypeBlindsT12, "Legrand MyHome" },

and if possible change the source comments to English ?

gizmocuz:
Our build system have some compile issues, please check below

steplebr:
I've corrected the problems you mentioned.

gizmocuz:
Thanks!! Thanks for this great addon!
I read we maybe could also use this for KNX systems? (i got some few gira hardware (dutch)) working with a bus system

steplebr:
The difference between KNX and Openwebnet is that Openwebnet has been designed to be simple while KNX is quite complicated. KNX bus can be accessed via different gateways, IP (using KNXNet protocol, which is KNX tunneling) or USB. However, KNXNet is not publicly documented. There is still a java project that uses KNXNet, it is called Calimero (https://github.com/calimero-project/calimero-core).
Some other home automation projects may use proxies : eibnetmux (http://eibnetmux.sourceforge.net/, it exposes a tcp socket protocol and a C client api) or eibd (exposes a tcp protocol, used by jeedom https://domotiga.nl/projects/domotiga/wiki/KNX and domotiga https://domotiga.nl/projects/domotiga/wiki/KNX).

Issue #507, opened 2016-02-22
std::string::compare method issue


gaudryc:
@gizmocuz 

Could you check all call to the std::string::compare method that you introduced in the commit #f6d661aa314955f4e1859c9ad26ff5b87afcb5e3, in the following files ?

hardware/HarmonyHub.cpp
hardware/SBFSpot.cpp
main/EventSystem.cpp
main/SQLHelper.cpp
main/WebServer.cpp
notifications/NotificationHTTP.cpp

[http://www.domoticz.com/forum/viewtopic.php?f=6&t=10677&p=76078#p76078](url)

gizmocuz:
Fixed, thanks for the patch

PR #508, opened 2016-02-22
Fix std::string::compare method issue (#507).

gaudryc:
@gizmocuz 

Could you check all of these changes before merging ?

Thank you.


gizmocuz:
Actually i did made a test application to test this yes
And it worked

```
    std::string szText = "Version CSV1";

    if (szText.find("Version CSV1") == 0)
    {
        std::cout << "Yep1" << std::endl;
    }
    if (szText.compare("Version CSV1") == 0)
    {
        std::cout << "Yep2" << std::endl;
    }

```

So all these so called fixes in your pull request should not be needed, did you test ?

gizmocuz:
Ahhh i now see when this does not work.. sorry my bad

gizmocuz:
Thanks!
I was in the understanding that .compare compared only the part between the quotes "Version ", but apparently it tested the whole string, in my test application i the test string was the complete string..

PR #518, opened 2016-02-24
Fix exception thrown on domoticz termination

gaudryc:
Voilà !

The problem was due to the lack of use of the various destructor while the applicatyion is shutting down. So there was no stop on the cWebem::m_io_service.


gizmocuz:
... almost merged... not ok ?

Edit: I see this does not solve the problem, the delete (*it) also crashes
This has been introduced since the https changes... but cant see where

Also try to compile with "NOCLOUD" as preprocessor

gizmocuz:
In server_settings.hpp we need to also include

`#include <boost/asio/ssl.hpp>`

gaudryc:
Yep it was not OK at all.
I didn't last night...
I had to disable webserver run/stop and and clean_session_timer to begin to solve the problem.
I found the way to start/stop clean_session_timer and a modified version of the server class.
Now I'm trying to return to the HEAD without breaking anything ;-)

gaudryc:
Step by step almost line by line...

gizmocuz:
Hope you find it

gizmocuz:
Found something out

bool CWebServerHelper::StartServers(..)
{
  return true;
  ....
.....

if you return true right at startservers (so it is not actually started), no crash when pressing control-c

(hope its helpfull ;)

gizmocuz:
not the problem, but why is server_settings doing a new/delete in clone ?
it is just a structure ?

gizmocuz:
found the problem (not the fix yet)

WebServer.cpp

if we do

```
            m_pWebEm->SetDigistRealm("Domoticz.com");

            return true;

            m_pWebEm->SetSessionStore(this);

```

no problem

if we do

```
            m_pWebEm->SetDigistRealm("Domoticz.com");
            m_pWebEm->SetSessionStore(this);
            return true;

```

problem !

problem is in the session store (probably a timer or something)

gizmocuz:
Yep, doing this results in no crash

```
        //m_pWebEm->SetSessionStore(this);
```

gizmocuz:
fixed... but i am going to do some more code refactoring (dont like new/delete malloc/free stuff!)

Fix:

```
        void CWebServer::StopServer()
        {
            m_bDoStop = true;
            try
            {
                if (m_pWebEm == NULL)
                    return;
                m_pWebEm->SetSessionStore(NULL);
                m_pWebEm->Stop();
            }
            catch (...)
            {

            }
        }
```

```
            m_pWebEm->SetSessionStore(NULL);
```

dont know if it's good/wrong yet... but it solves the crash

gizmocuz:
I have really no idea whats going on here:

Webserver.cpp

You make (via clone) a copy (new!?) of the settings
then you do
						settings_copy->listening_address = "::";
						settings_copy->listening_address = "0.0.0.0";

nice..

and then

delete settings_copy;

uhm?

at this moment the copy settings could have changed, but the original not..., so each time we get here, the same will happen

gizmocuz:
settings, its just a struct, why a class with getters/setters... looks like c# here ;)

gizmocuz:
i committed my changes, if possible change the whole server settings to a normal struct with normal members, no getters/setters, its so windows 95 ;) _lol_
and the new/delete thing, actually we do not want any new/delete/malloc/free

Still got a problem here with the tlsv1_1 stuff, made a commit for this, but this needs more attention from me..

gizmocuz:
For https i now get (and maybe with previous compile as well)

2016-02-25 21:00:34.644 Error: Failed to start the web server: std::bad_cast
2016-02-25 21:00:34.645 Error: check privileges for opening ports below 1024

gaudryc:
We work hard today !

If you just make a copy of the server_settings argument (which can server_settings or ssl_server_settings) you will loose the "ssl_" because you use the  server_settings::server_settings() copy constructor instead of ssl_server_settings::ssl_server_settings() copy constructor.

I changed the code to pass both mainworker::m_(ssl_)server_settings by reference (without const) to fix the ugly and dangerous "delete".

gaudryc:
I made some changes in WebServer, WebServerHelper, cWebEm, server to 
- use "boost::shared_ptr" ;
- cleaning servers collection and pointer when stopping ;
- remove virtual pure method in server ;
I'm trying to fix the session_store because it seems it is the origin of the "pure virtual method called" message.

gaudryc:
OK.
Thank you for your help (mySessionStore = NULL; had saved my life).
I gonna sleep now ;-)

gizmocuz:
Had a crash again, does not happen many times, but this is the callstack:
image

gizmocuz:
Please keep our history file clean, no need to add a 'fix' line for something that was implemented wrong before with a pull request, or for example, the new selector switch, no need to add fix lines, until we made a new stable version as it is still a work in progress until a new stable has been release...

gizmocuz:
Github code commit history is nice for that

gaudryc:
I understand but the connection clean was implemented but not actived (no call to the method), it seems to me a bug.

Could you tell me what is the error message ?

gizmocuz:
image

gizmocuz:
Having HTTP and HTTPS enabled, browsing via HTTP to the Lights page, hitting a few times F5 (refresh) and then pressing control-c to terminate the application

gaudryc:
So it happens when the server is stopping, and it seems that the io_service is already unavailable when we try to close all connections.
At this time, when the server is stopping we do :
1- close the acceptor
2- stop all connections
Then I think we should :
1- add a boolean to be able to refuse all new connections when the server is stopping ;
2- stop all connections ;
3- finally close the acceptor that close the io_service.

rkluwen:
After you close the acceptor, no new connections will be accepted anymore already.

gaudryc:
Yes but if you stop the acceptor it seems you stop every dirtily.
So I'm trying to make a graceful stop...

gaudryc:
With the actual version, I got these kind of end messages:
------------- normal
`2016-02-26 12:55:34.698  Mainworker Stopped...`
------------- error

```
2016-02-26 12:56:44.466  Stopping all hardware...
*** Error in `./domoticz_test_2.1': double free or corruption (out): 0x6fe005b0 ***
```

------------- error

```
2016-02-26 12:57:04.928  Stopping all hardware...
2016-02-26 12:57:04.951  Error: [web:8080] exception occurred : 'close: Bad file descriptor' (need to run again)
```

------------- error

```
2016-02-26 12:59:33.827  WebServer(SSL) stopped
*** Error in `./domoticz_test_2.1': double free or corruption (out): 0x6fe005b0 ***
```

------------- error

```
2016-02-26 13:03:17.597  Stopping all hardware...
*** Error in `./domoticz_test_2.1': corrupted double-linked list: 0x71201a60 ***
```

------------- error
`2016-02-26 13:03:49.638  Stopping all hardware...`
------------- error
`2016-02-26 13:03:55.571  TCPServer: shared server stopped`

gaudryc:
If I invert connection "connection_manager_.stop_all();" and "acceptor_.close();", I got :
------------- normal
`2016-02-26 13:14:41.588  Mainworker Stopped...`
------------- error

```
2016-02-26 13:14:56.653  Stopping all hardware...
2016-02-26 13:14:56.661  ::ffff:127.0.0.1 -> handle read timeout
```

------------- error

```
2016-02-26 13:15:09.972  Stopping all hardware...
2016-02-26 13:15:09.991  ::ffff:127.0.0.1 -> handle read timeout
*** Error in `./domoticz_test_2': double free or corruption (out): 0x74b005d8 ***
```

gaudryc:
All messages above have been obtained by interrupting domoticz while stressing it with wget on the HTTP port.

gaudryc:
I'm testing something like this:

```
void connection_manager::stop_all(bool graceful_stop)
{
    if (graceful_stop) {
        std::for_each(connections_.begin(), connections_.end(),
                boost::bind(&connection::stop_gracefully, _1));
        int timeout = 10; // force stop after 10 seconds
        time_t start = mytime(NULL);
        while(true) {
            sleep_milliseconds(50);
            if (connections_.size() < 1) {
                break;
            }
            if ((mytime(NULL) - start) > timeout) {
                // timeout occurred : force stop
                _log.Log(LOG_STATUS,"Graceful stop timeout occurred : all connections will be closed now");
                break;
            }
        }
    }
    std::for_each(connections_.begin(), connections_.end(),
            boost::bind(&connection::stop, _1));
    connections_.clear();
}
```

gaudryc:
I've forgotten that stopping the server is asynchonous so even if we make a graceful stop, domoticz continues to stop without waiting for the web server to stop.
So we cannot do anything but wait to avoid these exceptions.
I'm trying to test what appends if we wait for connections to stop.

gaudryc:
I have created a closed pull request to show you what I plan to do (#520)

gizmocuz:
All these problems where not here before your big patch....
We can also revert, and you can make a branch ?
The idea was to have a code stop for a few days before the new stable release
But as i see it now we have to either postpone the stable release for a week,
or revert these changes

gizmocuz:
Also that patch from me yesterday solved the control-c problem

gaudryc:
You did not really solve anything because the original code of domoticz kill the webserver using "delete" instruction without closing anything properly (you make a asynchronous stop and delete a few milliseconds later). It is like sending a "kill -9".

I think it would be a very bad thing for everyeone to revert now.

You will have the solution tomorrow in the pull request #520. It is ready on my computer but it is too late to make good things. I spent too much time on it for 3 days. I am dead.

gaudryc:
You should be happy you will have a mini Apache in Domoticz ;-)

gizmocuz:
Yes, i started with that mini apache

gizmocuz:
i am not blaming anybody, but maybe we should have tested this more

gaudryc:
Yes you are blaming but I'm doing my best to make things better. Even if my path was not so good, you habe to recognize that the original (the code before mine) was not so good also.
Whatever it is. I think you really want to stress the webserver and even if you revert to old code you will always need to close things cleanly.

I will submit the pull request #520 and you will forget all about theses comments.

gizmocuz:
I am not blaming... or is it the same what you did with my code refactor patch ? *g*

gizmocuz:
I am happy with your patches dont get me wrong... but you have to admit this one needed a little bit more testing (lot of users could not start anymore, crashes on exit), everyone makes mistakes.. we are human, and i dont like blaming, so maybe next time you do not do the same as you did before with my previous commit ?

gizmocuz:
Cleaned up this issue for both our interest

gaudryc:
OK, I'm recompiling with your last commits and I will pull very soon.

PR #520, opened 2016-02-26
Make a clean webserver stop

gaudryc:
Now when you terminate the domoticz instance, you ask from the web server to stop asynchronously but delete the cWebem instance after calling its Stop method.

Here it is what I'm testing :
- Make the cWebem instance to wait for all connections to stop gracefully ;
- Ignore exceptions while canceling timers ;
- Ignore exceptions while closing socket (on connection stop).
- Ignore exceptions while stopping acceptor.


gaudryc:
Please have a nice stress testing ;-)

gaudryc:
Now you can really disable one of the two servers or both.

gaudryc:
I which to add a command line argument to activate cWebem/server/connection debug log.
With (ssl)_server_settings it will be easy now.

What do you think about it ?

gizmocuz:
If you think that other users have to debug the www part, this might be an option.

Else, i would compile with a new preprocessor like -DDEBUG_WWW 
.. thats not a bad idea... i think i will pull and change all to
#ifdef DEBUG_WWW

this way, we can also let others compile with this flag and do debugging, what do you thing ?

gizmocuz:
Sorry to say, it still crashes on control-c

Do you use visual studio for debugging ? If not, could you use it ?

`void connection::stop_gracefully() {`

image

fixed by

```
void server_base::handle_stop() {
    // The server is stopped by cancelling all outstanding asynchronous
    // operations. Once all operations have finished the io_service::run() call
    // will exit.
    is_stopping = true;
    **connection_manager_.stop_all(false);**
```

I am changing the function to:

```
void server_base::handle_stop() {
    is_stopping = true;
    try {
        boost::system::error_code ignored_ec;
        acceptor_.close(ignored_ec);
    } catch (...) {
        _log.Log(LOG_ERROR, "[web:%s] exception occurred while closing acceptor", settings_.listening_port.c_str());
    }
    connection_manager_.stop_all(false);
}
```

Reason: First close the accepter (no more connections are possible), then close all connections
Else we closed all connections, and because the acceptor was still open, a possible new connection could be made
(Not related to the problem above here)

gaudryc:
Visual Studio seems to do more things.
I will try to run domoticz with gdb to see if I see something.
Thank you.

About the above issue, do you try to modify the Mainworker to start without anything and add a debug log in the WebServerHerlper, compile, run and interrupt ?

gizmocuz:
I will try... bdw you do not use gdb with visual studio

Is it possible to look into this one:
https://github.com/domoticz/domoticz/commit/73673a89303f28d72f284c4409790742bf122b84#commitcomment-16375675

Maybe it is always closed after 20 minutes, even if there was activity from browser->domoticz

gizmocuz:
What is the difference between the read_timeout and the abandoned_timeout ?
I think the read_timeout is enough ?

When there is no more read, then the connection is abandoned ?

PR #521, opened 2016-02-26
	Restart hardware after a long period of inactivity

szczukot:
What do You think about this ?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -262,6 +262,8 @@ hardware/MySensorsTCP.cpp
 hardware/NefitEasy.cpp
 hardware/Nest.cpp
 hardware/Netatmo.cpp
+hardware/OpenWebNet.cpp
+hardware/openwebnet/bt_openwebnet.cpp
 hardware/OpenZWave.cpp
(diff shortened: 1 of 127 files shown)


gizmocuz:
Thats why we have implemented a data timeout ?

szczukot:
Ahh, i see.
That maybe m_LastHeartbeat is not needed anymore ? And it is for remove from all modules ? Is it needed ? Is it only for logs ?

gizmocuz:
Nope

szczukot:
Is it neede or not ? :) I wrote I too many questions.

szczukot:
Is it needed ?

szczukot:
??
If it is unnecessary, i do refactoring, and I remove it with all code

gizmocuz:
Why do you think that m_LastHeartbeat is not needed anymore ?
Have a look in the mainworker
Please only do refactoring on your own code

szczukot:
I don't know. 
In mainworker this uses only for logs "warning" when device don't responses for more than 60 sec.
I can not found another use for this property.

PR #525, opened 2016-02-28
[1WireByOWFS] Try fix #430

### LINKED ISSUE #430: 1wire: Reading sensors values from OWFS sometimes fails

szczukot:


gizmocuz:
I dont know what you committed in the pull request, but it contains things that should not be in there
Your only change was a sleep call ?
Why not test with the uncached ?

szczukot:
After send "0xCC" (Skip ROM) bus required about 750 ms to send data. Maybe this sleep resolved problem reading bad data. This is for test. I have about 30 devices on 1wire, and i dont have any problem with bad data.

Issue #523, opened 2016-02-26
Abnormal termination even if nothing is started


gaudryc:
Even if the code does not even execute m_webservers.StartServers() / StopServers(), we got :

```
pi@raspberrypi ~/domoticz_test_2 $ sudo ./domoticz_test_2 -www 0 -sslwww 0
2016-02-26 22:46:58.706  Domoticz V2.4790 (c)2012-2016 GizMoCuz
2016-02-26 22:46:58.707  Build Hash: 25a3fb2-modified, Date: 2016-02-26 09:19:49
2016-02-26 22:46:58.708  System: Raspberry Pi
2016-02-26 22:46:58.708  Startup Path: /home/pi/domoticz_test_2/
2016-02-26 22:46:58.800  Starting shared server on: :::10144
2016-02-26 22:46:58.801  TCPServer: shared server started...
2016-02-26 22:46:58.802  RxQueue: queue worker started...
2016-02-26 22:47:00.803  Domoticz Internal: Started
2016-02-26 22:47:00.804  EventSystem: reset all events...
2016-02-26 22:47:00.806  EventSystem: reset all device statuses...
2016-02-26 22:47:00.807  EventSystem: Started
^C2016-02-26 22:47:04.381  Closing application!...
2016-02-26 22:47:04.381  Stopping worker...
2016-02-26 22:47:04.381  RxQueue: queue worker stopped...
2016-02-26 22:47:04.383  TCPServer: shared server stopped
2016-02-26 22:47:04.383  Stopping all hardware...
2016-02-26 22:47:04.804  Domoticz Internal: Stopped...
2016-02-26 22:47:05.800  Scheduler stopped...
2016-02-26 22:47:05.808  EventSystem: Stopped...
2016-02-26 22:47:06.308  Mainworker Stopped...
terminate called after throwing an instance of 'boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<boost::lock_error> >'
  what():  boost: mutex lock failed in pthread_mutex_lock: Invalid argument
```

gaudryc:
Even if I change MainWorker::Start and MainWorker::Stop to
`bool MainWorker::Start() { return true; }`
`bool MainWorker::Stop() { return true; }`
I got :

```
pi@raspberrypi ~/domoticz_test_2 $ sudo ./domoticz_test_2
2016-02-26 23:53:50.615  Domoticz V2.4790 (c)2012-2016 GizMoCuz
[...]

PR #527, opened 2016-02-29
Fix issue 523

gaudryc:
This patch really fixes the issue because the execution of these destructors needs the global object _log which can already be destroyed, because the destruction order of static objects is not guaranteed.

(That is the answer I want before closing an issue)

### DIFF
--- a/main/WebServerHelper.cpp
+++ b/main/WebServerHelper.cpp
@@ -23,7 +23,6 @@ namespace http {
 
 		CWebServerHelper::~CWebServerHelper()
 		{
-			StopServers();
 		}
 
 		bool CWebServerHelper::StartServers(server_settings & web_settings, ssl_server_settings & secure_web_settings, const std::string &serverpath, const bool bIgnoreUsernamePassword, tcp::server::CTCPServer *sharedServer)
--- a/main/mainworker.cpp
+++ b/main/mainworker.cpp
@@ -185,7 +185,6 @@ MainWorker::MainWorker()
 
 MainWorker::~MainWorker()
 {
-	Stop();
 }
 
 void MainWorker::AddAllDomoticzHardware()


gizmocuz:
Thanks for the patch, but i dont understand it 100% (sure about the log understand), but:

For as i can see there is nothing send to the log (only when there is an exception)

https://github.com/domoticz/domoticz/issues/523#issuecomment-189978227

There is no log function in WebServerHelper::StopServers()
(In the current code)

So calling StopServers() should be safe in ~CWebServerHelper() ?

~MainWorker() -> Stop

It should already been have stopped by the call from domoticz.cpp:

m_mainworker.Stop();

At this point, there is no used m_thread object, so nothing should be done

gaudryc:
I know there is no log in the  CWebServerHelper::StopServers() function but someone could add a log. If _log is destroyed before m_webservers, you will have the exception "terminate called after throwing an instance of 'boost::exception_detail::clone_implboost::exception_detail::error_info_injector<boost::lock_error >'
  what():  boost: mutex lock failed in pthread_mutex_lock: Invalid argument".

In MainWorker::Stop() you have a _log.Log(...) call so this exception can occurred letting you think the problem is somewhere else.

Because the Stop() and StopServers() are not needed in these destructors I think It would better to avoid any confusion.

gizmocuz:
because calling MainWorker::Stop() is already done in the main app (see a message above, around line 861 in domoticz.cpp), a second call to it, will never trigger that log

If stop was not called, the destructor will take care of all

The log object is created in domoticz.cpp and will be alive till an exit or the last return...

... as far as i can see, and expecting, as it is first included and used here

That someone will add a malfunction in the code later can always happen, i also see a lot of 'throw' calls in domoticz... if this is not taken care off the application will also stop

gaudryc:
So you should do it everywhere something should be stopped, just in case...

PR #534, opened 2016-03-03
Webserver enhancement

gaudryc:
This pull request contains :
- Speed up the refresh of the browser cache by adding the Keep-Alive response header to "Not Modified" (HTTP 304) and "No content" (HTTP 204) responses.
- Centralize the Keep-Alive header management ;
- Add -wwwgracefulstop argument to stop the webserver gracefully.
- Stop lying to client by sending the real idle timeout (20s instead of 10s) ;
- Fix graceful stop for "waiting-handshake" and "initializing" connections (can be stopped now)
- Fix the restart of the web server if the acceptor is still up despite of the exception thrown.
- Revert to the previous implementation of server_base::stop to make the handle_stop method called from the same thread as io_service::run().

Can I include abandoned connection timeout ?


gizmocuz:
I am sure we do not want this startup option:

-wwwgracefulstop boolean

did not look any further are the changes yet....

gizmocuz:
We should just stop in the correct way... no need for an extra startup option

gizmocuz:
Dont know what the correct way is, but maybe rene can have a look at this as well

gaudryc:
I prefer to stop softly to avoid problems. I have made some change to the connection::stop_gracefully method to speed up the stop.
I run with this since my first pull request with it and it is just clean.

Now I'm just waiting before going forward...

gizmocuz:
Okey.. But for sure it is not a startup parameter
It should just stop....

rkluwen:
Okay, so I got a quick look at it even though not completely.

Here are my comments:
- The usefulness of the -wwwgracefulstop parameter is unclear to me.
- You moved the keep-alive handling outside of the http handler to the connection class. The reason why it was not there in the first place is that there's now http handling in the socket layer. Which seemed unnatural to me. But I can understand your reasoning. I think your way is better.
- I would prefer to see the status strings be an enum. If needed at all.
- Tiny optimization: In add_header_if_absent you can just return from the function if the header exists, instead of setting a boolean.
- About server_base::handle_stop(). Maybe I'm not a boost expert enough. But all documents advice to post a message in the queue.
- I think we would always want a graceful stop. Why have a setting for it?
- What does your graceful stop do exactly? Call stop()... and then wait... and then call stop() another time? What's the use?
- What is is_stopping in the server_base class for? Once you close the acceptor, handle_accept won't be called anymore implicitly.
- The function connection::is_stopping() has some serious side effects. I assume this is intentionally, but you might want to split it up to avoid confusion with future contributors.
- Finally, I plead you to have a look at the Websockets-refactored branch. It's always a drag to merge your patches with this because the connection class needed to be changed (websockets is not a request/response protocol). It's okay to do this once if there's a good reason. But it has been a couple of times now.

Well, these are just my 2 cents, for what it's worth.

gaudryc:
Thank you for this code review. I have made the following changes :
- Handle abandoned connection to avoid memory leaks (appends if SSL handshake fails).
- Move the connection::status from std::string to private enum.
- Optimize the reply::add_header_if_absent method
- Always do a graceful stop
- Remove the server_base::is_stopping attribute which becomes unnecessary since the acceptor is stopped before connections.
- Remove the confusion with the connection::is_stopping method.

rkluwen:
I'm fine with this patch now.

Just for curiosity: What's the advantage of what you call a "graceful stop"? Why is it needed?

gizmocuz:
After applying this pull request, and closing domoticz with control-c under windows i got a crash again

image

@gaudryc , do you test with visual studio ? I asked it now 3 times, if not, please do so !

fixed by changing:

void server_base::handle_stop() {

connection_manager_.stop_all(true);

to

connection_manager_.stop_all(false);

Ps. i only test with HTTP (no https)

callstack:
image

i got this crash everytime, start domoticz, goto devices page, press control-c

gaudryc:
Thank you !

The graceful stop is not absolutely needed but it gives a chance to an active connection to go to the end of writing process before stopping. If the writing process take more than 10 seconds (value defined in connection_manager::stop(bool) method, then the connection is forcibly stopped.

gaudryc:
I didn't see your last message.

I don't understand how you do that so I will not insist with the graceful stop. I will remove everything and everybody will be happy.

gizmocuz:
Look.... Your patches / changes let us crash... So please fix it

gizmocuz:
And do not be so childish when we post where the crash is 
We do want your patches but do accept our bug reports

gizmocuz:
And I ask you for the fifth time again.... Do you test with visual studio?
If not so providing any patch!

gizmocuz:
This is the sixt? Time I need to provide a patch for your pull requests.. Are you really going to argue about what is right? Come on!

GertSanders:
Looks like you have been more then patient, time to end it gracefully ;-)

gizmocuz:
If you want to make me happy... Stop with these crappie pull requests because it is taking to much of my time....
I can accept 1000 failures... But if the person responsible for it is still complaining...
I better have not a single pull request from that person at all.....

gaudryc:
I did not test is under Windows because I need to prepare my environment to do that.
I will remove it in a hour. I'm compiling under OSX and Debian before commiting.

gizmocuz:
I do not care what you use to test your changes, if you do not accept a crash report, what is the use ?

gizmocuz:
I will accept 1000 patches from anyone, but if you can not accept when a pull request is crashing (for 6 times!), and still be mad about it...
It is time to stop sending pull requests....
Please stop !

gaudryc:
I accept your crash report but I can't do any analyze because I don't have your environment.

gizmocuz:
Sure you can have my environment... 
How are you testing ? 'make' ?
Do you know 'codelite' ? Sure you can install windows....

What i do not like is that you criticize my comments about crashes...  and give me a hard time. it crashes !

gizmocuz:
I don't even know what will happen when the 20mins passes.... Do we get another crash?

gizmocuz:
While we are here..please have a look at today's commit logs where I fixed another one of your problems 
.... It is enough... Criticism I can handle till a certain extend...

gaudryc:
I find you too hard. Each of us makes mistakes. And you forget too quickly that I also fixed some of yours.

gizmocuz:
Needed to fix two of yours today, and with my 3000 commits you don't even come close 
Go cry on someone's else shoulders.... I got enough of your comments and pull requests

gizmocuz:
What a lame response...

gaudryc:
I'm testing under Visual Studio. It seems the problem is coming from the connection::stop_gracefully method that should not call "connection_manager_.stop(shared_from_this());" because it erases the connection so it invalidates the iterator used by the for_each loop used in connection_manager::stop(bool) method.
The connection::stop_gracefully should call "stop()" instead. Then the connection will be erased at the end of the connection_manager::stop(bool) method.

gizmocuz:
It sems when we call

connection_manager_.stop_all(false);

(no graceful stop), that everything is also cleared/free'd, and stopped, is this call necessary ? See question from Rene above

gaudryc:
I think it's time to take stock of my tests in debug mode with Visual Studio Community 2015 Update 1 under Windows 10 Pro (VM using 2 cores) and the lastest WindowsLibraries.7z package.

First of all, to correct the exception thrown when graceful stop is active, I must effectively change the connection::stop_gracefully method, just to make a simple stop (socket shutdown). This avoid to erase the connection from the connection_manager::connections_ set while it is reading.

Except that after thoughtful, I think that the end of the acceptor arriving before stopping connections, the io_service quickly becomes unavailable. This could have an adverse effect on the connections that depend in part on it. So, the sooner we destroyed the connection, the best it is (staying clean of course).
And we are not in a critical production environment therefore the HTTP client can be ignored while stopping.

If you are agree, I'll remove the code attached to the graceful stop, the code will be simpler and its execution less risky. The only thing to keep is the wait loop after closing the server.

Anyway you should keep in mind that running in a debugger can involve different behaviour. The latency can reveal some bugs but it can introduce some others.

gaudryc:
Should I post the fix or the removal ?

gizmocuz:
Sorry was hoping @rkluwen had a peek at the fix before i had too

gaudryc:
I can post the fix now.

gizmocuz:
Yes thanks

rkluwen:
Yes, please post it. I think the graceful stop can be removed completely. Anyway, it doesn't really help a lot for anything. The service is going down! All responses should be invalidated by the client in that case.

A small wait loop does also do the job in your case. And even that (IMHO) is a waist of cycles. But okay.
Your patch does include some useful other things though. So yes, please go ahead.

gizmocuz:
Is it possible to check:
http://www.domoticz.com/forum/viewtopic.php?f=6&t=10968
https://www.domoticz.com/forum/viewtopic.php?f=36&t=5052&p=78575#p78558

seems there is now a HTTPS login issue

gaudryc:
Last night, I asked some questions and provide some information in the forum. For now I have no feedback. In any case, in light of what is reported, I'm really not sure that the problem comes from Domoticz, especially as the Domotix software does not seem to be very good (it crash if no switch is present).

mikefranken:
Thanks for looking into it. It turned out to be a Domoticz issue, breaking the HTTP spec. Gizmocuz fixed it with https://github.com/domoticz/domoticz/commit/7c719fd6f8b58d6dfb7683ea55dbb728c524e6fa

Issue #532, opened 2016-03-03
Multi state switch on/off delays not saving


cherowley:
When save is clicked after entering a number in either the on delay of off delay boxes the values are NOT saved.

gizmocuz:
i do not understand... should this not happen ?

cherowley:
Woops! I mean they are NOT saved. It's a blooming shame as the new selector switch is rather nice and makes for a great tidy interface :)

gaudryc:
Sorry for the inconvenient, it will be fixed very soon.

cherowley:
Brilliant! Cheers :)

PR #537, opened 2016-03-03
Fix on/off delays not saved on the selector switch edit form (#532)

gaudryc:
This patch only fixes the "not saved" side of the issue.


gizmocuz:
We got a merging problem probably by accepting a previous patch
i can also do this manually
Maybe you can add the history in the top part of the history document
Looks like this history line is inside the history for the previous (stable) release ?

gaudryc:
It was a html5.appcache conflict.

gizmocuz:
Thanks

Issue #538, opened 2016-03-03
Off Delay not handled for Selector switch


gaudryc:
While testing the patch #537 I saw that Off delay (30s) was taken into account but was fired after 90s that is to say the value of the On Delay.

cherowley:
Cheers :)

gaudryc:
@gizmocuz ,

I'm modifying MainWorker::SwitchLight and CEventSystem::ScheduleEvent methods to handle the switch selector but it seems that other devices don't handle Off delay (these methods add On delay only if the switch command is "On" or something similar.

Am I right ?

PR #540, opened 2016-03-04
Fix issue #538

gaudryc:
This patch consists of :
- Handle OnDelay and OffDelay the same way for all devices ;
- Fix Selector switch OnDelay and OffDelay management.

**I've tested it with a button (On/Off), a Selector switch, a Scene and a Blockly script to control a button but I don't know if it covers all cases.**

### DIFF
--- a/History.txt
+++ b/History.txt
@@ -3,6 +3,7 @@ Version 3.xxxx (March xx 2016)
 - Fixed: Better handling of white listed commands/urls
 - Fixed: Double checking command rights
 - Fixed: Fix on/off delays not saved on the selector switch edit form (#532).
+- Fixed: Fix Off Delay not handled for Selector switch (#538).
 
 Version 3.4834 (March 2nd 2016)
 - Implemented: Blockly, now possible to use expressions in notifications/email/SMS like "My temperature is {{temperaturedevice[1234]}} degrees" (see http://www.domoticz.com/wiki/Events)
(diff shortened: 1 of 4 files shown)


gizmocuz:
The off delay is handled in SQLHelper.cpp, line 3218

if (AddjValue!=0) //Off Delay
...

When i use the current system, and make a virtual dummy switch, with an on delay of 5 seconds, and an off delay of 10 seconds, and press the switch, in the log i see

after 5 seconds, ON
after 5+10 seconds, OFF

this is the expected behaviour

gaudryc:
Do the OffDelay have to be handled in CSQLHelper::UpdateValueInt only ?
What is the reason to make this difference between OnDelay and OffDelay management (i.e. different places) ?

gizmocuz:
When a switch command is send, then we only need to check for the ON delay, because it might need to be scheduled after xxx seconds
Then when the actual switch command is performed, and the database is updated with the current state (ON), we check if that switch had an OFF delay, and if so, we schedule a switch-off command
(not for motion sensors)

gaudryc:
OK, now I understand that On delay and Off delay usage is very confusing.
For me :
- On delay produces a delay to be switch on ;
- Off delay produces a delay to be switch off ;
  What you are saying is that the device will be switch on after OnDelay and then switch off after OnDelay + OffDelay.

I don't blame you but I think it needs some help label in the edit form.

Anyway, I thank you for these informations.
So I will change my pull request to only handle Selector device in the three places it would be.

gaudryc:
Ok, it works as you want now.

gizmocuz:
Thanks, well yes, might need a bit clearing up... hope it is in the manual.. (that is 1.1 year old now _g_)

PR #547, opened 2016-03-07
Syslog facilities as an added command line option argument

hansrune:
My first attempt on contributing via github ... (noob in other words). Please let me know if I have misunderstood the process for this. 

This pull requests will allow a syslog facility to be added after -syslog  option command line option. To stay backwards compatible, no option argument will default to "user". The optional -syslog argument can be user,daemon,local0 to local7

I log everything at level 0 from my Domoticz instances via rsyslog to a central place. Being able to use a common localN or separate ones comes in very handy. Also, I have other equipment that insist on using LOG_USER, and mixing that with Domoticz is undesireable.

Detailed logging is made further possible by using a UPS protected RPI setup combined with  https://www.domoticz.com/wiki/Setting_up_overlayFS_on_Raspberry_Pi for no SD card wear and tear.


gizmocuz:
Hi Hans,

Very nice !!!! And very well done... looks perfect !

I will wait a little bit until our build systems agree that there are no compile errors before i merge this patch, but all looks good !

hansrune:
Thx. Good to hear that

Issue #539, opened 2016-03-03
CMake files do not check correct .git directory


irimitenkan:
During build CMakeList.txt and getgit.cmake check with `git  rev-list HEAD --count` etc. the recent version of domoticz. If the .git directory is not available the `appversion.default` shall be used.

But in some build environments (e.g. openwrt) it is possible the build process will find a .git directory from a foreign repository.
Hence the complete domoticz version info is wrong (based on foreign repository) since the git call does not use `--git-dir ./.git` option.

gizmocuz:
Please make a pull request

irimitenkan:
Has been fixed with merge #548

PR #548, opened 2016-03-07
ensure correct git repository path #539

irimitenkan:
#539 - now all git calls to build version info use the additional optoin  `--git-dir ./.git`.

With these changes build with openwrt build environment works fine and domoticz bin is created with  correct appversion.default when domoticz's .git directory is not available.


gizmocuz:
Thanks!

PR #550, opened 2016-03-08
Fix webserver graceful stop

gaudryc:
This patch DOES NOT ACTIVATE the graceful stop but just fix it.

The first commit is mandatory even if graceful stop is not activated.
The second commit is just the fix for the graceful stop issue.


gizmocuz:
Thanks

PR #553, opened 2016-03-10
Fix ugly separator at start of Evohome zone status message

gordonb3:
Current status message in Evohome zone objects in the Temperature view starts with ", "


gizmocuz:
Thanks!

Issue #373, opened 2016-01-22
Temporary override not visible in the API


Jaap-Jan:
Setting a temporary override works as expected (if I set an override via the web interface, the controller is also updated). The web interface briefly shows the correct status 'TemporaryOverride' before restoring it to 'Auto'. In the controller you can see that the override is still in effect.

gizmocuz:
Please create a pull request

Jaap-Jan:
You may close this one as well. I haven't fixed the issue and I don't know how to fix it.

gizmocuz:
What hardware device ? It could be that the new status is polled, but that you receive old data
is this also after lets say 15 minutes ?

Jaap-Jan:
I use the Honeywell HGI80. I haven't checked, but I will check that when I'm home.

Jaap-Jan:
I've checked this and even after a few hours the returned status is 'Auto', not 'TemporaryOverride'.

gordonb3:
Likely fixed with revision 3.4944

gizmocuz:
Great, and is it fixed?

Jaap-Jan:
Updated Domoticz to 3.4967 (current beta), but unfortunately this patch doesn't seem to help. Still the same symptoms as before.

gizmocuz:
What hardware device is this? (In Domoticz)
It is still not clear for me... You talk about Honeywell... We do not have a Honeywell hardware device

Jaap-Jan:
I'm sorry, the device in Domoticz is called "Evohome USB (for HGI/S80)". The HGI-80 is an USB device to communicate wirelessly with an Honeywell Evohome controller.

gordonb3:
Just for clarity: what are you setting exactly? The override showing in domoticz web UI is because the code simply assumes a success without polling the result. Changing back to auto means that the central unit did not register the override and thus reports the zone to follow the normal schedule.

There's something of a feature here: the central unit will only register changes to zones. Meaning that if you change the setting of a single radiator controller it will not show on the central unit and consequently the RFG100 and/or HGI80 will not report the override - even if that radiator is the only zone member. So when you say "controller", are you then referring the central unit or the radiator controller (HR92/HR80)?

Jaap-Jan:
1) By "controller" I mean the central unit, or as Honeywell calls it, the "central controller". In this case the ATC928G2000.
[...]

PR #554, opened 2016-03-10
Fix invalid TimeUntil value in evohome SetSetPoint

gordonb3:
Somehow the struct that controls the behaviour of this function is not correctly preloaded with the right values. I attempted various fixes to evohome.h and evohome.cpp, to no avail. This fix is to the SetSetPoint() function in mainworker and enforces the expected value for 'year' that makes the code call the hardware interface with the correct parameters.

I do not own an HGI80 but I suspect this will also solve issue #373 opened on Jan 22 by Jaap-Jan


gizmocuz:
Thanks for the patch!

PR #555, opened 2016-03-10
Compress static web content at packaging time

gaudryc:
The .js, .json and .css files will be compressed as .js.gz, .json.gz and .css.gz before creating .tar or .zip in the following scripts :
- makebeta
- makerelease
- .travis.yml
- appveyor.yml

The compression is made by calling "packagingtasks prepare" or "packagingtasks.cmd prepare".
The compressed files are removed from the www directory after making .tar or .zip by calling "packagingtasks clean" or "packagingtasks.cmd clean".


gaudryc:
The scripts packagingtasks or packagingtasks.cmd can be called manually to test compression or clean process.
At this time the script packagingtasks can be run in the domoticz directory only.
The script packagingtasks.cmd can be run from everywhere.

gaudryc:
Is it normal the Travis ci process didn't performed the deploy phase ?

gizmocuz:
Yes, and i also think appveyor did not do this, because it wont deploy anything outside the master branch ;) (... makes sense, otherwise everybody's pool request will be released)

gizmocuz:
Edit... i hope appveyor does the same... does not look like it from the log...

gizmocuz:
In the Appveyor settings it is all OK, so i think it is just simulating

gaudryc:
I'd like to modify the packagingtasks script before merging in order to make the script callable from everywhere by adding call to "readlink" and "dirname" to compute the script dir at the beginning of the script.

gaudryc:
About travis I was talking about before_deploy. Is possible to stop Travis after the before_deploy phase to allow the packaging process validation ?

gizmocuz:
I have no idea...

gizmocuz:
Are you still testing this ? How is this going to work ?
Making a package/archive, we only archive the .gz files, not the original files ?
Or are we going to archive both compressed and uncompressed making the archive bigger?

gaudryc:
The test is over since a long time. 

The goal is to compress the original in order to have them both in the archive.
The compressed file will have the same timestamp than the original one.

When the archive is finished, all the compressed files are removed to avoid to disturb git with files not being managed.

If you need to run domoticz from the source directory with compressed files, you can run "bash ./packagingtasks prepare" (or packagingtasks.cmd prepare) before running domoticz.
Then, if you need to remove all compressed files (before making git diff or git commit), you can run "bash ./packagingtasks clean" (or packagingtasks.cmd clean).

gaudryc:
Once these changes were merged, you can:
- unzip all files .js.gz and .json.gz;
- commit the files .js and .json into git;
- then remove the files .js.gz and .json.gz from git.

You don't have to make it all at once but one by one if you are afraid of it.

gaudryc:
To avoid making the archive bigger, we have to add excluding filters while archiving to exclude all .js, .json and .css files. But it makes the archiving process more complexe and the size difference is not so big, +334 KB with all files .js, .json, .css, .js.gz, .json.gz and .css.gz :

```
-rw-r--r-- 1 pi   pi   13038517 mars  19 16:51 domoticz_beta.tgz
-rw-r--r-- 1 pi   pi   13380854 mars  19 22:23 domoticz_beta.with_both.tgz
-rw-r--r-- 1 pi   pi   13046053 mars  19 22:26 domoticz_beta.with_gz_only.tgz
```

gizmocuz:
true, but on the 'end-user' site, you now have two files instead of one, and uncompressed the size is bigger
There are people using Domoticz on very small systems like openwrt

probably we can do something like

`tar -zcf ${archive_file} www/js/*.gz www/css/*.gz` ..... (not tested, hope that works)

Also i think we should bump the internal db version number and delete all uncompressed files `(www/js/*.js, www/css/*.css)`

gaudryc:
What do you mean exactly when you say "bump the internal db version number and delete all uncompressed files (www/js/.js, www/css/.css)" ?

gizmocuz:
I mean, when we are going to deliver compressed pages, we want to clean up the end-users www folder
So when we increase the dB version number, we normally do database patching, but we could also clean up all uncompressed files that are not needed anymore, leaving the domoticz folders in a clean/correct state

gaudryc:
If you are talking about the html5.appcache, you don't need to do anything because the browser doesn't known anything about how the server is serving a .js, .json or .css. The browser ask for a file, the server send the content of the file.gz if it exists letting the browser think it is the file compressed by the server on the fly. In all case the uncompressed content is saved in the browser cache because the browser uncompresses it on the fly.

So you just need to do what you've done before, when you change a file referenced in the cache you need to increment the db version in the file html5.appcache.

gizmocuz:
No i am talking about cleaning up folders, making the end-user only have compressed files, not original files

gaudryc:
Did you see my last commits about filtering the .js, .json  and .css ?

2016-03-24 9:57 GMT+01:00 gizmocuz notifications@github.com:

PR #557, opened 2016-03-11
Evohome: adjust UTC times to localtime

gordonb3:


gizmocuz:
Thanks for the patch!

PR #560, opened 2016-03-12
Fixed harmony hub url to use http

jaylik:
Implementation should be fixed to use ssl later - now it uses http, but this one url has https prefix -> Squid proxy won't allow connections trough with invalid request


gizmocuz:
I dont understand... because you have a proxy server that can not work with HTTPS you want us to change all HTTPS calls to HTTP ?

jaylik:
Whole implementation is writen with non ssl/tls requests (`authcsocket.connect(LOGITECH_AUTH_HOSTNAME, 80);`) and HTTP POST request has https:// -prefix in its headers so transparent squid proxy drops invalid request.

Whole implementation should be changed to use TLS, but this fixes request for now.

gizmocuz:
Thanks for spotting!

PR #561, opened 2016-03-12
Fix a wrong starting value issue for ZWave kWh devices

gaudryc:


gizmocuz:
I think we need to divide instead of multiply
see the zwavebase.cpp's usage

`SendKwhMeter(pDevice->nodeID, pDevice->instanceID, BatLevel, pPowerDevice->floatValue, pDevice->floatValue / pDevice->scaleMultiply, "kWh Meter");
`

does this change, also goes for line 1888 (searching for "kwh") ?

gaudryc:
Device value should contain Wh, openzwave sends kWh so we have to multiple.
I've changed it because I saw there was an issue after adding some logs.

gaudryc:
It is invisible for everybody because updatevalue came shortly after addvalue. But because I had some strange values in Kwh logs (in meter table) when I was restarting domoticz, I have decided to add some logs then I saw these bad starting values.
This patch makes things coherent between addvalue and updatevalue.

gizmocuz:
Okey, but it needs to be applied in 2 places then ?

gaudryc:
Did I miss something ?

gizmocuz:
Maybe one place

If you search for "kwh", you will see two occurrences... you patched one, i was wondering if this needed to be patched on two places
About line 1888 is the second occurrence

gaudryc:
Ok I will see it this evening.
Thank you

gaudryc:
You're absolutely right, I had to correct my problem without thinking that it could be elsewhere.

gizmocuz:
Well you could not have known all code ;)

gizmocuz:
Thanks!

PR #562, opened 2016-03-14
Speedup SQL queries

gaudryc:
- Speedup the SQL queries on the DeviceStatus table by creating an index on fields that uniquely identifies a device (many queries use this four fields filter);
- Speedup the SQL queries on "Log" tables by creating indexes with DeviceRowId and Date on "Log" tables (these indexes optimize all queries that use a filter like " WHERE DeviceRowId=? ORDER BY DATE ASC");
- Move old indexes from "<TABLEINITIALS>_idx" to "<TABLEINITIALS>_id_idx".

### DIFF
--- a/main/SQLHelper.cpp
+++ b/main/SQLHelper.cpp
@@ -31,7 +31,7 @@
 	#include "../msbuild/WindowsHelper.h"
 #endif
 
-#define DB_VERSION 101
+#define DB_VERSION 102
 
 extern http::server::CWebServerHelper m_webservers;
 extern std::string szWWWFolder;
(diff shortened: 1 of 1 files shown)


gizmocuz:
Great, do you have some performance results ?

gaudryc:
For the first no, I didn't make measurements, but for the second the query time is divided by 3.

For the first, the fields are almost used everywhere as if it was the primary key to find the device, so the use of the index makes a difference for sure.

gizmocuz:
Nice ! Thanks

gaudryc:
Your welcome !

PR #565, opened 2016-03-16
Add ByronSX door button to RFLink

jlpdiez:
In fact it adds all chimes supported by RFLink. Apart from ByronSX it allows to use:
- Select Plus wireless Doorbell
- Plieger York Doorbell protocol
- Deltronic Doorbell
- RL-02 Digital Doorbell
- Lidl/SilverCrest Doorbell  Z31370-TX


gizmocuz:
Thanks

PR #566, opened 2016-03-16
last modified patch

rkluwen:
Small change at which the Last-Modified http header date was wrong after content has been changed by myWebem->Include().


gizmocuz:
Thanks!

PR #567, opened 2016-03-16
Fix some selector switch misses

gaudryc:
The changes are :
- Fix missing icon for selector switch in Devices tab (it fixes a HTTP 404 error) ;
- Fix Selector switch management in Scene/Group (it limits the number of level that can be selected from the select menu and it fixes the value of the level before executing action).


gizmocuz:
Thanks

PR #575, opened 2016-03-22
Improve Power graph for instant+counter meter

ridingtheflow:
This commit improves the reporting for meter that gives both accurate instant load (in W) and total usage (in kWh).
It provides the graph which allows to view simultaneously the instant usage and hourly usage while avoiding accuracy issues with trying to reconstruct one from another (since database already has exact data for both anyway).

chart 5
chart 6
chart 7

This should affect only "General, kWh" meter.


ridingtheflow:
Note - new graph will only appear if "Power day graph" selected in Settings. If the "Energy day graph" is selected, it will show standard hourly usage bar chart.

gizmocuz:
Thanks for the patch !
The graph for 'Energy Used' looks kind of ugly, is it possible to make this a line chart? (without the small and big dots?)

ridingtheflow:
It is possible, but it would be actually less practical this way, I've tried to use it for a bit yesterday. Since yellow series are hourly kWh usage, big dots actually used as hover points to see exact kWh tooltips - its hard to hunt for them with mouse cursor unless you see where they are.
Also I tried to use whole line as solid line before, and it often obscured small "peaks" in instant usage graph - so making it dotted line made them more visible.

Ideally hourly usage should be a bar graph, but Highcharts have issues with zooming in into combined bars/line chart with different X resolutions (making blue graph clip out), so this was the best compromise I found...

ridingtheflow:
I will try with semi-transparent but solid hourly line, will post how it looks in a bit.

ridingtheflow:
Tried without hourly marks, but it does not work that well.
Also tried to change on hover, but it looks very ugly-blinking semi-randomly.

Something like this maybe? 
chart
chart 1
chart 2

gizmocuz:
That looks nice ! Could it be that now the blue graph looks bigger? like the line size is doubled ?

ridingtheflow:
Yes, I've made blue line thicker (actually it was originally this size), to prevent yellow line from "dominating".

gizmocuz:
Thanks! Patch looks great!, only one thing worries me, that is the removal/disabling of the crosshair tooltip, why is this ? Do we still see both values in the tooltip at the same time ?

ridingtheflow:
You don't see both values in tooltip - shared tooltips pretty much unusable in Highcharts when one set of values have 1 hour resolution and other 5min resolution (at least I was not able to make it work). But this is not a regression, since previously there was no second value on this graph and you never seen two values there. P1 smart meter uses different function for graph and its not affected by this change so its still shows both values.

Croshairs I think were turned off at some point because they conflicted with something... But they seem to work fine now, I will put them back.

xbeaudouin:
Love it :D

gizmocuz:
Thanks for the patch !

gizmocuz:
I am changing  back the areaspline vs spline here...  

before:
image

After:
image

This looks much clearer

gizmocuz:
(hope ya won't mind bdw!)

ridingtheflow:
Try to see how it looks in different scenarios. I've changed to areaspline because spline was often hidden by yellow (bright & thick) hourly line in front of it.

gizmocuz:
yellow is not enabled by default

ridingtheflow:
True, but how it looks if you enable both? I would say my main concern (and main reason for this change) was for when you actually see both graphs :)

gizmocuz:
You could disable the yellow?

gizmocuz:
I mean remove

ridingtheflow:
Well, the whole point I made this is to look at _both_ graphs simultaneously to see both peaks and hourly trends at glance. I don't really care about seeing single graphs that much, it was just a side extra feature which highchart provided so why not. But having to click between graphs to be able to see them well would be pretty inconvenient and I will have to keep my own fork just for that :(
Btw I don't see why you don't want areaspline, P1 meter graph also uses areachart (just line, not spline).

gizmocuz:
I think you know how different the area spline is used in the p1 graph 
Okey so I will remove the yellow graph which is kind of useless now 
If you do not like it... Then yes... Keep your own branch of course!

gizmocuz:
Ohhh and the yellow one was..is disabled by default..so you had to click anyway to see both graphs?

ridingtheflow:
Yes, but just once. Once I don't mind. But then I just wanted to look at both, so I can zoom, scroll etc while still seeing both.
What do you mean "remove yellow graph", are you reverting it completely now?

gizmocuz:
Seems like it.... Only the blue graph is of most interest here 
Tomorrow I will check the difference between the previous system and the current 
(on different installations)
If on the old version only the blue graph was there... For sure I will remove the yellow graph 
You can change (could change) between usage and power from the preferences

ridingtheflow:
Umm, but why, I mean, by default it was just like it was before, you see only blue graph. Only that you now can (optionally) switch on the yellow graph too. What the harm?

gizmocuz:
Clearness... For sure the area effect does not help

ridingtheflow:
Well, I'd prefer other people to get at least some benefit of it, so if you can keep it as is (without area affect) its still lot better then just throwing it away completely.
Don't get me wrong, I don't mind you turning areaspline off (even if I may not use it myself) - but having both charts even as a spline (and as optional functionality) still would help other users IMO.

gizmocuz:
Okey I keep it... It may be better to display the yellow as bar chart for each hour?

ridingtheflow:
Yes bars would be better as I mentioned it before. Unfortunately Highcharts have problems with zoom in this case (it starts cutting off line chart when you zoom in). They recommend using pointRange=0 for this not to happen, but this is not an option because pointRange must be set to hour, otherwise it makes bars only 5mins wide because for whatever reason it auto-sizes them to be minimal sample size from _all_ datasets...

xbeaudouin:
Is this can be added also on powergraphes for ecodevices ?

ridingtheflow:
Which exactly ones are these? There are so many of different elec sensors :/

xbeaudouin:
It is a french one (special device) that is accesseed on http. Have a look in hardware/EcoDevices.\* (still don't know how to hack / fix the .js part of domoticz so I cannot yet propose a patch to it).

ridingtheflow:
What is the equivalent in virtual devices (on Dummy hardware)? I don’t have this exact hardware so won’t be able to test/develop unless I can test on virtual device.

xbeaudouin:
I can give you a private access to my Ecodevice, please contact me in private :)

PR #578, opened 2016-03-23
Fix for unvisible Timers in some browsers (not supporting Math.log2...)

EddyK69:
Changed **Math.log2(x)** to **Math.log(x) / Math.log(2)**


gizmocuz:
Thanks for the patch!... I thing those browsers should be patched as well _g_

PR #579, opened 2016-03-23
Logitech Media Server: Added HeartBeat response to respect timeout-settings

EddyK69:
- Added better handling of missing responses


gizmocuz:
Thanks for the patch!

PR #580, opened 2016-03-24
Evohome: correct handling of decimal device ID's with scripted evohome

gordonb3:
This resolves not being able to change mode on EvoTouch if the device ID
is more than six digits long.


narfight:
ho yes ! thank you

PR #600, opened 2016-04-07
Added EnableSetpointTimer & DisableSetpointTimer

EddyK69:
Added EnableSetpointTimer & DisableSetpointTimer in addition to the existing methods for Timers & SceneTimers


gizmocuz:
Thanks for the patch!

PR #604, opened 2016-04-11
Look for the dynamic WiringPi library if the static one was not found

trishika:
This patch aim to solve the issue raise on the forum at  : 
    https://www.domoticz.com/forum/viewtopic.php?t=7678


gizmocuz:
The reason why Domoticz is compiling against the static version is, so that new installations/updates will have this build in.
The wiring-pi step could also just be a small extra section in the wiki... I think 2% of the rpi users is actually using this library

trishika:
I am not sure what you mean by an extra section on the wiki, you mean to embedded wiring-pi or to link to the dynamic version ? 
Perhaps you would prefer this to be an option, the same way this is done for openzwave ? 
Building with wiring pi as a dynamic library makes sense when you are packaging for a distribution.

gizmocuz:
True, but currently we are not in a distribution, and for this we need to make some more changes to the CMake configuration file.
I just meant to add some extra lines in the wiki explaining to compile wiringpi (if needed)
But you could also make an option like OpenZWave, but please put it default on YES (Static libs)

Maybe we could/should add a option to build for a distribution that sets all correct parameters ?

PR #607, opened 2016-04-12
PiFace counter types

bobkersten:
Added the ability to configure the counter to be of a different type. Besides the default rfxmeter, energy is supported. In the future, support for gas, water or others can be added. A divider setting is added that is used to convert the pulses into the target meter type.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ CMakeCache.txt
 domoticz
 stdafx.h.gch
 .DS_Store
+.AppleDouble
 Config/OZW_Log.txt
 Config/zwscene.xml
 backups/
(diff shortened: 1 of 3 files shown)


gizmocuz:
Thanks for the patch !

bobkersten:
can you perhaps shed some light into what battery_level I should use for permanently powered devices such as these new energy devices for PiFace? The S0counter code uses "9", but that shows a battery level of "high", I'd like it to be completely ignored.

gizmocuz:
It probably is, as it has no battery
If you look in the S0MeterBase.cpp you will see calls like

sDecodeRXMessage(this, (const unsigned char *)&tsen.RFXMETER, NULL, 255);

The last argument (255) is the battery level (0-100, 255=no battery)

It gets overrulled

But on most places we used to use '12', but now it should be specified like above

it is also advised to use the 'send' functions in the DomoticzHardware class

PR #609, opened 2016-04-12
Correction for minus values from analog sensors in Sterbox device

houk001:
Removing plus sign also removing minus :) 
Domoticz reads temp correct with plus sign before value.


gizmocuz:
We got some merging problems, please update the code before making new patches
Could you fix it ?

houk001:
how to update the code only in github? I'm fresh in github too xD

gizmocuz:
You make modifications, and commit/push them, then it should be updated
But you could also delete this pull request, fork again, make changes and then commit

PR #610, opened 2016-04-13
Added basic sensor support for the Tellstick hardware.

MrDadoo:
Added basic sensor support for the Tellstick hardware.
Useful when using the Tellstick Duo


gizmocuz:
Thanks for the Pull request, i just answered your email, and applied the serial file patch ;)

MrDadoo:
One question. Tellstick devices generates wind direction, gust and average as separate events.
I've found the SendWind function, but it accetps dir, speed and gust, all at once.

Is there a way to send a "not specified" value and report one variable at at time?

gizmocuz:
Easiest, just keep a structure internal so you have all values, and when one is changing, use the 'other' recorded values

gizmocuz:
Closing this pull request as it is already in the repo

PR #611, opened 2016-04-13
Corrections for minus value from analog Sterbox sensor

houk001:
Removing + character also removes -


gizmocuz:
Thanks for the patch!

houk001:
No problem :D

Issue #599, opened 2016-04-07
make install: possible incorrect permissions (USE_SOURCE_PERMISSIONS)


niziak:
I'm compiling domoticz from source and installing system-wide with "sudo make install". Domoticz is installed into /opt/domoticz. To run domoticz server, I'm using separate user account "domoticz" and I'm executing /opt/domoticz/domoticz with additional parameters to point database to "domoticz" user home directory, etc. For now everything works well.

When I'm cloning and building source of domoticz I'm using my default user account with umask set to 077.
With this umask, It is not possible to install domoticz system wide because my permissions are used.

I know that it is probably hard to accommodate all scenarios, but if there is no good security reason it will be nice to set world readable permissions for /www and /Config directory?

It can be done by simpy removing USE_SOURCE_PERMISSIONS from CMakeLists.txt for /www and /Config

```
install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/www DESTINATION ${CMAKE_INSTALL_PREFIX})
install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/Config DESTINATION ${CMAKE_INSTALL_PREFIX})

```

gizmocuz:
I see no harm in this ! Could you make a pull request ?

PR #612, opened 2016-04-14
make install: use default permissions for /www and /Config

niziak:
See #599


gizmocuz:
Thanks for the patch !

PR #617, opened 2016-04-18
Fix for LUA-values of counter-devices & added LUA-table...

EddyK69:
..."otherdevices_scenesgroups" for reading Scene- & Group status in LUA


gizmocuz:
Thanks!

PR #618, opened 2016-04-19
Ability to force a counter update when interval changes are larger th…

bobkersten:
…an a certain percentage & ability to reload piface from hardware screen without restarting domoticz


gizmocuz:
Thanks for this patch !
Unfortunately there is a merging problem, is it possible to solve this ?

PR #620, opened 2016-04-20
PTEC with no tempo option shoud be set

bibulle:
Tele info do not work for now, if no tempo options, the current values are not correct


gizmocuz:
Thanks!

PR #627, opened 2016-04-23
Added multi-room zone support for HGI80 to Evohome.cpp

Evsdd:
This resolves the problem with not receiving regular temperature readings from any multi-room zones when using the HGI80.


gizmocuz:
Thanks!

PR #633, opened 2016-05-01
Fixed a typo in the CMakeLists message() function call for Mosquitto external

defl:
Was causing issues while trying to get out-of-source MQTT libs to work


gizmocuz:
Thanks!

PR #649, opened 2016-05-13
Line endings

deece:
Amend line endings of checked in text files to their native format.

Addresses #644

PR #657, opened 2016-05-15
1wire enhancements

deece:
Note: This pull request is dependent on #649 

I am currently implementing a 1Wire based home automation system, with DS2408 chips reading the state of wall switches, and controlling solid state relays.

This pull request creates a high frequency (default of 100ms) thread to poll the switch state, while leaving sensors in a low frequency thread. The polling periods for both threads are now user configurable (under Settings/Other/1Wire).

This pull request also fixes a bug that prevents switch states from being read correctly.

Note that this introduces new translation requirements:
Sensor Poll Period
Switch Poll Period
1Wire Settings


gizmocuz:
Thanks for the addon/changes !

Some questions:
- What are the files (.cproject and .project), maybe add them to the git-ignore file ?
- If you have an eclipse project, would you like sharing this ? Maybe making a folder eclipse with the solution files (like the msbuild folder)

I am still struggling with the line-endings, will post a commit soon

gizmocuz:
The Hardware sqlite table has 6 integer configuration values, and 1 string
Could you rewrite this patch to use them, instead of getting hardware configuration options from the global preferences table?

deece:
I've got an eclipse project added in here that has a bunch more static analysis stuff, I'll drop the files from this one that got caught up by mistake.

Happy to implement your suggestion, Is there an example I can follow to see how you are referencing the table?

gizmocuz:
In the mainworker.cpp class, the mode1/5 are available when creating the class
for example see how the opentherm gateway is created

pHardware = new OTGWSerial(ID,SerialPort, 9600, Mode1, Mode2, Mode3, Mode4, Mode5, Mode6);

or when starting the hardware you could query the info like in

bool CFitbit::LoadRefreshToken()

deece:
Thanks, I've made the changes suggested and it seems to be working well with OWFS. I've also made the OWFS path configurable.

gizmocuz:
Great, thanks !
Are you sure you pulled the latest changes? Are build systems (travis/appveyor) has a problem with

/Users/travis/build/domoticz/domoticz/main/mainworker.cpp:602:7: error: use of undeclared identifier 'HTYPE_S0SmartMeter'

there are now two types (HTYPE_S0SmartMeterUSB/HTYPE_S0SmartMeterTCP)

deece:
Sorry, was a bad merge, should be fixed now.

gizmocuz:
This patch has more changes then just in the 1-wire, but its manageable.
Could you remove the 1-wire configuration stuff like  'OneWireSensorPollPeriod'  from the global settings ?
The Hardware sqlite table has 6 integer configuration values, and 1 string
Could you rewrite this patch to use them, instead of getting hardware configuration options from the global preferences table?

deece:
Whoops, there were a few settings variables left unused after the last rework, I just removed them.

gizmocuz:
Thanks !!

gizmocuz:
I think the default 'sensor' poll interval (1000 ms = 1 second) is way to fast... 
How about we change this to 60000 by default (once a minute) ?

Imagine 20 sensors... 20 new log lines every second...

gizmocuz:
Sorry... i think i was using an older version as the default is now 30 seconds

Maybe we should make a check for this (in the constructor of the class)
we can also enter 0 or 1, also for the switch interval, this should be checked in the web interface?

OlbertP:
Hi,
Since 22 may my OWFS 1-wire senrsors stopped working. When i switch back to the release version everything is working again.
I'm using a DS9490R to connect my 1-wire sensors and OWFS (which is still working on OS level).

Here the error message in the log: 
2016-06-13 08:38:07.202 Error: 1Wire: Device not yet supported in Kernel mode (Please report!) ID:00-e7c100000000, family: 00
2016-06-13 08:38:07.202 Error: 1Wire: Device not yet supported in Kernel mode (Please report!) ID:00-17c100000000, family: 00
2016-06-13 08:38:07.202 Error: 1Wire: Device not yet supported in Kernel mode (Please report!) ID:00-97c100000000, family: 00

In the 1-wire setup is filled in OWFS path: /mnt/1wire , Sensor poll periode: 1000 and Switch poll period: 100

Gizmocuz adviced to change sensor poll periode to 30000 but this did not helped.

gizmocuz:
30000 is now the default (30 seconds) as you can read in the post above yours

deece:
@OlbertP It looks like you have the W1 kernel module loaded, could you unload it and try again?

deece:
@gizmocuz The default should be 300,000ms, ie. 5 minutes, which is what it was originally.

gizmocuz:
That is a lot, maybe we could make it once every minute ?

gizmocuz:
The E1 kernel module is loaded by default i think in our image... but if the owfs path is provided, we should use this ?

deece:
@gizmocuz The original code polled the sensors every 5 minutes, so I preserved that behaviour. We can change it, but it should be discussed/tracked as a separate issue.

deece:
@gizmocuz re the W1 module, as per your comments in the code pointing at http://owfs.org/index.php?page=w1-project , running both systems in problematic, and given that OWFS is more desirable, the W1 module should probably be disabled in any distros that are pushed out.

I'll tweak the code now to use OWFS if the path is set, but that does will not prevent the conflicts described on that page.

OlbertP:
I removed the W1 modules and now it works :-)

gizmocuz:
@Deece, i believe in the code before your modification, we scanned first for the OWFS path (/mnt/1wire), if found, this means, OWFS is installed, in use, and we should use this.
Then the kernel-w1 code was not runned.
If the mount point was not found, we check for the kernel-w1 existanse, and if it is working, if yes, we can use this.
Else... no 1Wire support

nesalkiN:
Something happened between betas. My 1-wire sensors stopped to deliver data on the 15th. Just discovered it today. Using OWFS and had to manually enter the OWFS-path (/mnt/1wire) for it to start working again. I have never changed this. Domoticz does not seem to detect it automatically anymore.

gizmocuz:
@NiklasOlofsson , please use our forum, this has been discussed

PR #658, opened 2016-05-15
Add support for Rainforest RAVEn USB ZigBee Smart Meter Adapter

larber:
Hello,

I've been using Domoticz for the past year and have found it incredibly useful.  When I found out the new smart meter we got had a separate radio for ZigBee connectivity, I couldn't resist hooking it up.  The Rainforest RAVEn is a pretty simple device -- plugs into a USB port and emulates a serial interface that spits out energy usage data in xml.  This pull request implements parsing out the current usage and cumulative usage reported by the meter.  The meter also supports reporting real-time price information but there didn't seem to be a good way to report $/hr usage in domoticz so I've left that out.

I also think there was some change to whitespace in this repo very recently as I had some weird commit diffs that I spent way too long trying to resolve but ultimately couldn't fix.  If you use ?w=1 when doing the compare, you'll see the changes are quite minimal.

Let me know if you have any questions or if you feel that something is missing.  And thanks for creating domoticz in the first place!

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,5 +1,5 @@
 # Set the default behavior, in case people don't have core.autocrlf set.
-* text=auto
+#* text=auto
 
 # Explicitly declare text files you want to always be normalized and converted
 # to native line endings on checkout.
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,5 +1,5 @@
 # Set the default behavior, in case people don't have core.autocrlf set.
-#* text=auto
+* text=auto
 
 # Explicitly declare text files you want to always be normalized and converted
 # to native line endings on checkout.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -307,6 +307,7 @@ hardware/Tellstick.cpp
 hardware/Thermosmart.cpp
 hardware/ToonThermostat.cpp
 hardware/VolcraftCO20.cpp
+hardware/RAVEn.cpp
 hardware/Winddelen.cpp
 hardware/WOL.cpp
 hardware/Wunderground.cpp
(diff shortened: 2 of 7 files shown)


gizmocuz:
Thanks for this great patch !
Yep, we are currently working with line-endings... hope it is solved soon

gizmocuz:
Few questions:
- Can you add a example xml output at the top of the hardware class (as comment)
- Could there be multiple meters? If so, does each meter have a numeric ID ? If yes, we could use this in SendKwhMeter
- Instead of using the wptr_, can we not use a std::string in the readCallback function?
- class members start with m_ in domoticz

larber:
Done and checked into my fork if you want to pull.  As for your other questions:
1.  My understanding is that this device is designed to talk a single meter from the utility providing power to the entire premises so I don't think the notion of multiple meters applies.
2.  The reason wptr_ exists is because the callbacks from boost asio serial module may contain partial xml fragments so the driver needs to buffer them up until is receives a complete piece of xml that can be parsed out.  These fragments are appended to buffer_ until the parser succeeds, at which point the entire buffer is shifted down by the amount of text that was parsed.

gizmocuz:
Thanks, merged the changes....

Its just that i am a bit skeptic about using memmove functions on strings (buffers)
There is no character 0 being send on the end of the message ? Or before that a length ?
If not, sure we leave it like this ;)

Thanks again for this addon !

larber:
No and this is why I dislike streaming text-based protocols.  Figuring out where one message ends and the next begins is a huge pain, especially when messages can be split arbitrarily (as they are in this case).  I couldn't think of a way around the memmove since the data returned by the readCallback can span a message boundary so there's no way around parsing the first one and then moving the remainder to the beginning of the buffer while you wait for the rest of the message to come in.

PR #662, opened 2016-05-16
Moved P1 baudrate setting to main hardware screen as opposed to a sep…

bobkersten:
…arate screen for just one setting. This is more consistent with other hardware, such as MySensors, and it fixes a bug where baudrate would get reset to the default 9600 when altering settings on the main settings screen.

### DIFF
--- a/hardware/P1MeterSerial.cpp
+++ b/hardware/P1MeterSerial.cpp
@@ -120,6 +120,7 @@ bool P1MeterSerial::StopHardware()
 		sleep_milliseconds(10);
 	}
 	m_bIsStarted = false;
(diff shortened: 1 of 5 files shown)


gizmocuz:
Thank you for this... Unfortunately i was working on a new hardware type (S0 Meter TCP) and pushed this before i've seen this pull request...
Is it possible to solve the merging issues ?

bobkersten:
There was a small conflict in the HardwareController.js that git coulnd't auto-merge. The webserver.cpp file was auto-merged here. So I've resolved, added and pushed the conflict. I assume this is the way to go :) please double-check.

gizmocuz:
Looking good!, Thanks for the patch

PR #664, opened 2016-05-18
Code Sanitisation

deece:
This pull request aims to make the code more robust by introducing tools & practices to make it more obvious where potential problems lie.
- Increase warning levels
- Add cppcheck make target for static analysis
- Add Eclipse/Codan project files
- Start fixing some warnings


gizmocuz:
Our build systems have problems with this, see the output above/below

deece:
This now builds with no warnings with GCC 6.1 on Linux, using -Wall. This is probably a good point to merge this pull request, before I start looking at the Visual Studio warnings.

gizmocuz:
Thanks for the hard work !
Some changes i would like first before merging the code (its a lot)

Of course we know cppcheck, but not all of it's output makes sense, nor does it make the code readable, not error prove.
You will see a lot of repeated lines below:

**main/types.h**
Please remove this file (for things in the feature, we have mainstructs.h for this)
Dont use the DEVICEID/UNIT just 'unsigned long long' and 'int' in all code again please

**Blocks of code removed**, from o.a. EnOcean (dont do this, might need this later!), instead of remove, just remark (/\* */) if it bothers ?

**Dont change string.find to string.compare**, , compare might be a little bit faster, but with 'find' it is more readble, and when someone copy/paste the code, produces less errors

MQTT/*
dont change any code here, not from us, and might get replaced later

hardware/Meteostick.cpp
Code removed?
if (OpenSerialDevice())     
-                   bFirstTime = true;
  Now the serial devices is not re-opened after a failure?

void MochadTCP::setSecID(unsigned char *p)
j++ ---> j+=3
Did you test this ? Now it increments with 3 instead of 1

MySensorsBase::SendSensor2Domoticz
please remove all your newly added switch states, they are not meant to be send
this is just a stupid warning under mac, cppcheck
we handle switch states that needs to be handled, 

hardware/Netatmo.cpp
/\* below 3000: very low */
changed to
// below 3000: very low */
should then be
// below 3000: very low

CRFLinkBase::ParseLine
else if (results[ii].find("HSTATUS") != std::string::npos)
dont just remove code, this might be used later by the author!

CEventSystem::GetCurrentMeasurementStates(
Dont remove unused code, we might need this later !
Remark

CEventSystem::EvaluateLua
Dont remove unused code, we might need this later !
Remark

main/HttpPush.cpp
Dont remove unused code, we might need this later !
Remark

main/RFXNames.h
Dont remove STYPE_END, MTYPE_END, TTYPE_END, and HTYPE_END
No reason to remove it, and this way it is doomed for errors when adding/inserting things!

main/SQLHelper.cpp
DEVICE_ID, change it back to unsigned long plz

Webserver.cpp
Dont remove unused sql query items, place them all back.
We might need them later, and it does not do any harm, not that it will make your application 100% faster right ?
Please remark stuff, dont just 'delete' stuff
And the compiler will also optimize it (remove a=10; if a is not used)
For instance, rain rate is still on my todo.. gone now

main/mainworker.cpp, to many changes, can not compare/accept patch, not visible in github
probably also a lot can be reverted

main/mainworker.h
remove the DEVICEID

tinyxpath
dont change any code here, not from us, and might get replaced later

webserver/fastcgi.cpp
Dont remove unused code, we might need this later !
Remark

deece:
Thanks for the review,

Is there a test suite available? I had a look around but couldn't see anything obvious. I'm considering writing some myself if there aren't any yet.

further comments:

main/types.h
There are 2 reasons why having explicit typedefs are a good idea. The first is that if the underlying type ever has to change (for example, I just had to change a pile of them from unsigned to signed as they were set to & compared against -1), there is a single place to change it. The second is that if it is ever used in the wrong context, for example, using hardware ID instead of a device ID, the compiler will flag a warning, even if the underlying types are the same.

I'm Happy to move the defs to mainstructs.h though.

Removed code:
There is little reason to clutter the code with commented out sections when you have a revision control system, the old version can always be bought back from the repository, and the disabled code becomes stale/unmaintained in both situations. It also makes searching for things a bit of pain, as you get false hits in the dead code.

String.compare:
I see your point. it is more error prone. boost::starts_with() will gives us the best of both worlds.

hardware/Meteostick.cpp
bFirstTime was never read, if it was meant to do something different on the first pass, we should flag this as a bug.

void MochadTCP::setSecID(unsigned char *p)
The other 2 increments were done in the logic statements above, but the ordering is not guaranteed. I've changed these to static offsets and folded the increments into the line after the logic.

main/RFXNames.h
They're still there, defined immediately below the enum where they should be seen by whoever is next maintaining the code. Looking at the code now though, I think we are better just walking the matching array of structs and looking for the NULL terminator (that is the only use of the *_END members), eg. Switch_Type_Desc.  

main/SQLHelper.cpp
Not a good idea while we are setting & comparing it to -1. You could refactor to #define INVALID_DEVICE_ID ULLONG_MAX, and use that instead of -1.

Webserver.cpp
The value from the compiler notifying on unused variables is that these are indicative that some logic has been missed (eg. Meteostick bFirstTime). This value is lost if unused variables are intentionally left around. If a block of code is currently unused, but will be required later to support a feature in progress, one could keep it in a branch for that feature, to be merged in once the feature that uses it is ready.

main/mainworker.cpp
If you pull the branch locally, you can view the diffs at your leisure:
git checkout cppcheck
git diff master    # Text mode comparison
git difftool -t kdiff3 # Graphical diff tool, assumes kdiff3 in installed, meld is also a good choice
git difftool -t kdiff3 main/mainworker.cpp # # graphical diff of a single file

gizmocuz:
Okey, but

Removed code:

> > > There is little reason to clutter the code with commented out sections when you have a revision control system, the old version can always be bought back from the repository, and the disabled code becomes stale/unmaintained in both situations. It also makes searching for things a bit of pain, as you get false hits in the dead code.

_Disagree here, now i am forced to look at it... please move it back_

String.compare:

> > I see your point. it is more error prone. boost::starts_with() will gives us the best of both worlds.

_Please revert all compare changes where 'find' was used, keep the original find for various reasons
Find is also not the same as starts_with, so where we compare against 0, ok of course ;)_

main/RFXNames.h

> > They're still there, defined immediately below the enum where they should be seen by whoever is next maintaining the code. Looking at the code now though, I think we are better just walking the matching array of structs and looking for the NULL terminator (that is the only use of the *_END members), eg. Switch_Type_Desc. 

_Yes, we can see that, please revert your change_

main/SQLHelper.cpp

> > Not a good idea while we are setting & comparing it to -1. You could refactor to #define _INVALID_DEVICE_ID ULLONG_MAX, and use that instead of -1.

which should be the same correct?_

Webserver.cpp

> > The value from the compiler notifying on unused variables is that these are indicative that some logic has been missed (eg. Meteostick bFirstTime). This value is lost if unused variables are intentionally left around. If a block of code is currently unused, but will be required later to support a feature in progress, one could keep it in a branch for that feature, to be merged in once the feature that uses it is ready.

_You mean could have been missed. Do not take all cppcheck's output to the letter
Please revert_

gizmocuz:
Working through this pull request, going to do some things manually

Please exclude the MQTT folder as well (third party)

hardware/PiFace.cpp
```diff
@@ -4,29 +4,29 @@
  Date: 01-10-2013
  Written by: Robin Stevenhagen
  License: Public domain
- 
+
```
gizmocuz, line 7: Dont reformat code

hardware/serial/serial.cpp
```diff
@@ -163,8 +163,7 @@ Serial::readline (string &buffer, size_t size, string eol)
 {
   ScopedReadLock lock(this->pimpl_);
   size_t eol_len = eol.length ();
-  uint8_t *buffer_ = static_cast<uint8_t*>
-                              (alloca (size * sizeof (uint8_t)));
+  uint8_t buffer_[size];
```
gizmocuz, line 169: Does not work under VS2015

main/mainworker.cpp
```diff
@@ -804,6 +804,7 @@ bool MainWorker::AddHardwareFromParams(
 	case HTYPE_Sterbox:
 		//LAN
 		pHardware = new CSterbox(ID, Address, Port, Username, Password);
+		break;
```
gizmocuz, line 807: Was already solved

hardware/csocket.cpp
```diff
@@ -24,15 +24,15 @@
...
+csocket::csocket() : m_socketState(CLOSED),
                      m_remotePort(0)
 {
     m_socket = 0;
 }
 
 void csocket::close()
 {
-	 if ( m_socketState != csocket::CLOSED || m_socketState != csocket::ERRORED)
+	 if ( m_socketState != csocket::CLOSED && m_socketState != csocket::ERRORED)
```
gizmocuz, line 35: Thanks, fixed

gizmocuz:
Fixed one code line, closing this patch

PR #669, opened 2016-05-22
Allow 1wire polling threads to be disabled

deece:
This pull request allows the 1wire polling threads to be disabled by setting their polling period to 0.

This is useful if you have sensors but no switches, or vice versa.

### DIFF
--- a/hardware/1Wire.cpp
+++ b/hardware/1Wire.cpp
@@ -100,9 +100,6 @@ void C1Wire::LogSystem()
 
 bool C1Wire::StartHardware()
 {
-	m_threadSensors = NULL;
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thanks!

gizmocuz:
Can this be detected automatically by scanning the mount folder ? That would be better?

deece:
Are you suggesting that we terminate the thread if no devices of that type are found?

That sounds reasonable, but  we will also have to restart the thread if acceptNewHardware is enabled.

gizmocuz:
Or sleep 500ms in the thread.. that can't do much harm

PR #672, opened 2016-05-25
energy device optimizations and correct labelling

bobkersten:


bobkersten:
Not sure how to proceed here; the domoticz.js.gz file has changed and git can't automerge binaries which I think is the cause for the conflict.

gizmocuz:
thats correct, i made some changes
If you can redo your javascript changes, and submit the new code, we can merge this

bobkersten:
I've manually run a diff over my version of domoticz.js and the one on master and I can see only my own additions. Could you verify that your changes from May 22nd "Custom Sensor Month/Year graph now displays AVR" are included in my patch?
Is there a way to avoid these conflicts in the future? Do I need to change the way I make a pull request or stage these compressed files? A procedure I'm currently unaware of, or is this one of those things that just need to be lived with? :)

bobkersten:
Hmm, as it turns out, git does know how to automerge gz files (or it doesn't have to if I'm the only contributor)?? Conflict was in appcache. Pull request is now merge-able, but I would double check to make sure :)

gizmocuz:
Thanks for this !
Did you add the 2 new images in the appcache file ?
Do we need the two images, or can we use the already existing once ?

bobkersten:
I guess not; changed it to use css to vertically flip the images when necessary.

gizmocuz:
Thanks for the patch!
There is something strange now on the graphs, plz have a look at this image:

image

As both Y-axis are for Watt, can't you combine those
It is so strange to see 2.3 Watt plottet at 5 Watt (left axis), as you see this usage is pretty constant, and the yellow line was expected at the top of the blue bars

gizmocuz:
Or make both axis the same on the top-value, in this case around 4 (or take the maximum of one of the graphs for both)

bobkersten:
The left Y-axis shows Wh, the right Y-axis shows Watt, they are _not_ the same. One is energy, the other is power and they cannot be combined. Their top values should match if they're close together, that is something that can be improved. I'll take a look.

gizmocuz:
Or maybe not....  but this one is nice (maybe by accident the same top value on both axis)
image

gizmocuz:
Question:
- Is it an idea to enable the 'crosshars/shared tooltip' ?

```
        tooltip: {
            crosshairs: true,
            shared: true
        },
```

It seems there is a 30 minute offset to the Watt/Hour bars... why is this ? normally you say between whole hours the usage was xxx, now it is between 12:30 and 13:30 ....

gizmocuz:
image

bobkersten:
Power (the yellow line) is on the right, and 38.3 Watt is correcty displayed.

bobkersten:
The 30 mins offset is weird, i'll take a look. The 'other' energy devices (without a power line) don't show this behaviour.

gizmocuz:
Thanks, also the yellow line is not correctly displayed, dont look at the right axis, just the left
Now you have a 'bar' with 38.3 Watt(/hour) and a yellow line (38.3Watt) that sits in the middle of the blue bar, which should be arround 20 Watt

bobkersten:
It looks weird indeed, but the graph is 100% correct. It would look better if the maximum values on both the left and right axis would match, in which case the yellow line will sit right on top of the blue bar like the other graph you posted. I'm going to look into that.
Also, it seems as if WebServer.cpp returns energy values for each half hour, that causes the shift of 30 minutes. This is also something that needs to be fixed.
screen shot 2016-05-27 at 09 22 12

gizmocuz:
.. True, and it does make perfect sence... But as you said, its just that when its possible, it would be nice to have both axis on the same top value, but if you have a power peak of 2.3kWatt, then you dont want this when your Energy is 100 Watt average ;)
I think the 30 minutes was introduced by someone in a previous commit
Great graph bdw!

bobkersten:
Both y-axis have their max set to the max power, since it's impossible for the energy to be higher than the highest power peak :) Also removed the code in WebServer.cpp that shifts the bars with 30 mins. Graphs now look like this:

screen shot 2016-05-27 at 13 19 09
screen shot 2016-05-27 at 14 06 37

What do you think?

bobkersten:
Personally it makes more sense to shift the bars by 30 minutes like before. The bar then "sits" between for example 12:00 and 13:00, and that makes sense as it's the total energy generated between these timestamps. Tooltips should then show this range too, 12:00 - 13:00, instead of a 12:30. And finally, other daily graphs with a column series should behave the same, which they currently don't (mainly other energy devices without a power line).

gizmocuz:
Looking good.
About the shifting, it is correctly displayed, the bar goes from 22:00 - 23:00, the x-axis label is plot in the middle at 22:30, (like your second screenshot)

PR #676, opened 2016-05-28
Added Floorplan page 'fullscreen' query string option

dnpwwo:
Specifying 'fullscreen' causes the page to load without the menu bar being visible.


gizmocuz:
Thanks!

PR #677, opened 2016-05-28
Changed TypeImg to baro for Barometer devices. This should solve the …

djoele:
…issue that barometer devices can not show their graph from the Floorplan.

Also in the list of all devices a baro image should be displayed, instead of a gauge image. Then the image shown should be identical as in the Weather Devices tab.


gizmocuz:
Thanks for the patch, but this is solved on another way.
Please see the forum post if you would like to add a correct fix to this problem.
The General/Barometer is an example

PR #678, opened 2016-05-31
persistent chart series visibility state

bobkersten:
consistent- and visually distinctive appearance of trendlines
consistent legend visibility


bobkersten:
There was a hardcoded exception in WebServer.cpp for Youless energy devices which the javascript front-end wasn't aware of. This is now fixed and these Youless energy devices can now make use of the same combined graphs as other Counter+Instant devices. Together with the stateproviding mechanism  the original power/energy switch in settings can now be fully accomplished by swithing on/off the series in the chart.

gizmocuz:
Thanks!

PR #681, opened 2016-06-02
Add /dev/cuaU* USB serial devices on FreeBSD

xbeaudouin:
Some ZWave devices are detect as modem (thru a kernel module).

Then it is working using /dev/cuaUx instead of /dev/ttyUx.

For example Aeotec Z-Stick S2 and also Enocean stick ESP3 as well.

Using /dev/ttyUx may stuck domoticz. So both ttyUx and cuaUx are usefull be depends of the context.

Regards.


gizmocuz:
Thanks! I guess its depending on the platform/kernel you are using

PR #682, opened 2016-06-02
graph consistency and behaviour tweaks

bobkersten:
The addSeries, setData and update methods of highcharts _all_ redraw the chart, these were all called sequentially for most charts, causing those charts to redraw themselves multiple times for no good reason. This is fixed, charts are now only drawn when they're completely setup. This also fixes numerous animation glitches as a side effect.
screen shot 2016-06-02 at 19 34 28
screen shot 2016-06-02 at 19 28 31
screen shot 2016-06-02 at 19 18 23


gizmocuz:
Thanks !

PR #685, opened 2016-06-04
Small language fixes

taljaards:


gizmocuz:
Thanks for the suggestions. Unfortunately this change will changes in our root language file, and all language files needs to be changes as well.
We keep it like this...

taljaards:
Okay no problem.

I would be happy to update the English-only file if you would like?

I also guess we can go ahead and close this PR?

gizmocuz:
The problem is not only with the 'English/American' language, if you make these kind of changes, all language moderators have to update their language
Thanks for the suggestions

PR #699, opened 2016-06-14
Prefer OWFS over W1

deece:
- Prefer OWFS over W1
  - Push the logging of which module is used down into the hardware specific drivers, as it is impossible for the static LogSystem call to access m_path which is required to determine which class will be loaded.


gizmocuz:
Thanks!

PR #703, opened 2016-06-16
Try OWFS again with no path, if the kernel driver is not available.

deece:
This reinstates backwards compatibility for existing 1wire hardware on OWFS (which does not have a path set in the database).


gizmocuz:
Thanks

PR #711, opened 2016-06-19
Energy computed instead of reading

jotakar:
Some modules reads energy in kwh with precission of 1 digit  (2,1 kwh, 2,2 kwh ...) but sometimes the energy consumed in a day is less than 0.1 kwh. 
Think about a lamp type led or low consum: 8 watts for 10 hours is 80 wh (0,080kwh). but module reports 0.0 kwh for this day. 
With this modification system gets the energy value by software: it reads the usage and the interval of time and make a simple operation:  usage*time (usage: watts, time: minutes). 
I use field Options for Energy Meter devices. If  enabled (1)  the energy is calculated by program if  disabled (0) energy value is read from the device. 
I've found this issue with Qubino relays.


gizmocuz:
Thank you for the patch, but as you might have noticed both our build systems have problems with compile errors.

gizmocuz:
You can not change the language files, please provide here (as text) the needed changes to the language file

jotakar:
The words added and their translation into spanish are:
Energy read: Lectura de energía
Computed: Calculada
From device: Desde el módulo

Thanks

gizmocuz:
Why did you remove the 'HTTPPostContentType' in SQLHelper.cpp ?

main/SQLHelper.cpp
```diff
@@ -3095,8 +3090,9 @@ unsigned long long CSQLHelper::UpdateValueInt(const int HardwareID, const char*
 	bool bDeviceUsed = false;
 	bool bSameDeviceStatusValue = false;
 	std::vector<std::vector<std::string> > result;
-	result = safe_query("SELECT ID,Name, Used, SwitchType, nValue, sValue FROM DeviceStatus WHERE (HardwareID=%d AND DeviceID='%q' AND Unit=%d AND Type=%d AND SubType=%d)",HardwareID, ID, unit, devType, subType);
-	if (result.size()==0)
+	result = safe_query("SELECT ID,Name, Used, SwitchType, nValue, sValue, LastUpdate, Description, Options FROM DeviceStatus WHERE (HardwareID=%d AND DeviceID='%q' AND Unit=%d AND Type=%d AND SubType=%d)",HardwareID, ID, unit, devType, subType);
```
gizmocuz: Do we need to query for Description?

main/SQLHelper.cpp
```diff
@@ -3455,7 +3473,7 @@ unsigned long long CSQLHelper::UpdateValueInt(const int HardwareID, const char*
 		CheckSceneStatusWithDevice(ulID);
 		break;
 	}
-	if (bDeviceUsed)
+if (bDeviceUsed)
```
gizmocuz: missing a tab here... (that was previously here)

main/SQLHelper.cpp
```diff
@@ -4017,8 +4035,11 @@ void CSQLHelper::UpdateRainLog()
 			if (splitresults.size()<2)
 				continue; //impossible
 
-			int rate=atoi(splitresults[0].c_str());
-			float total = static_cast<float>(atof(splitresults[1].c_str()));
+			float total=0;
+			int rate=0;
+
+			rate=atoi(splitresults[0].c_str());
+			total = static_cast<float>(atof(splitresults[1].c_str()));
```
gizmocuz: revert these changes please, they are not needed and only give cppcheck issues
This patch is about energy/power anyway ;)

main/SQLHelper.cpp
```diff
@@ -4126,10 +4147,7 @@ void CSQLHelper::UpdateUVLog()
 	GetPreferencesVar("SensorTimeout", SensorTimeOut);
 
 	std::vector<std::vector<std::string> > result;
-	result=safe_query("SELECT ID,Type,SubType,nValue,sValue,LastUpdate FROM DeviceStatus WHERE (Type=%d) OR (Type=%d AND SubType=%d)", 
-		pTypeUV,
-		pTypeGeneral, sTypeUV
-	);
+	result=safe_query("SELECT ID,Type,SubType,nValue,sValue,LastUpdate FROM DeviceStatus WHERE (Type=%d)", pTypeUV);
```
gizmocuz: Revert this ? Why did you do this ?

main/SQLHelper.cpp
```diff
@@ -4169,7 +4187,7 @@ void CSQLHelper::UpdateUVLog()
 			//insert record
 			safe_query(
 				"INSERT INTO UV (DeviceRowID, Level) "
-				"VALUES ('%llu', '%g')",
+				"VALUES ('%llu', '%.1f')",
```
gizmocuz: revert this, why did you do this ?

main/SQLHelper.cpp
```diff
@@ -5388,7 +5406,7 @@ void CSQLHelper::AddCalendarUpdateUV()
 			//insert into calendar table
 			result=safe_query(
 				"INSERT INTO UV_Calendar (DeviceRowID, Level, Date) "
-				"VALUES ('%llu', '%g', '%q')",
+				"VALUES ('%llu', '%.2f', '%q')",
```
gizmocuz: Revert this, why did you do this ?

main/SQLHelper.cpp
jotakar: No, I had to have deleted Description. I'll delete it.

main/SQLHelper.cpp
jotakar: Sorry, I uploaded my version of SQLHelper.cpp and it had this code, so I reverted some changes made after my versión (5231). I'll revert all code I did'nt modified. You're right: this patch is just for energy/power.
Sorry it's my first pull.

main/SQLHelper.cpp
jotakar: Like the upper comment.

main/SQLHelper.cpp
gizmocuz: No problem... ;)

main/SQLHelper.cpp
```diff
@@ -5388,7 +5411,7 @@ void CSQLHelper::AddCalendarUpdateUV()
 			//insert into calendar table
 			result=safe_query(
 				"INSERT INTO UV_Calendar (DeviceRowID, Level, Date) "
-				"VALUES ('%llu', '%g', '%q')",
+				"VALUES ('%llu', '%.g', '%q')",
```
gizmocuz: I still see a lot of non-needed changes, and for example this one (procent point 'g') is not what it should be
Maybe best to revert this whole file, and only make that change needed for your patch
So we also do not get extra lines, or extra spaces at the end of the lines

main/SQLHelper.cpp
```diff
@@ -3132,17 +3133,40 @@ unsigned long long CSQLHelper::UpdateValueInt(const int HardwareID, const char*
...
+	            ntime.tm_min=atoi(sLastUpdate.substr(14,2).c_str());
+	            ntime.tm_sec=atoi(sLastUpdate.substr(17,2).c_str());
+	            interval = now - mktime(&ntime);
+		    StringSplit(result[0][5].c_str(), ";", parts);            
+	            nEnergy = strtof(parts[0].c_str(),NULL)*interval/3600 + strtof(parts[1].c_str(), NULL);
+	            
+	            StringSplit(sValue, ";", parts);
+	            sprintf(sCompValue, "%s;%.0f", parts[0].c_str(),nEnergy);
+	            sValue = sCompValue;
+            }
```
gizmocuz: How do you know it is an Energy device here ? All devices can have sOption="1"

jotakar:
I can't see why Travis C says there is an error. I see 
`Error: Failed to download resource "boost"`
But I can't see anything about my file ... What must I do?

main/SQLHelper.cpp
jotakar: Well I can use devType and subType , that's what I've done in last modif.

jotakar:
Hi
Which is the status of this pull request?

jotakar:
Hi
I'm using this modification in my net and it's working fine. 
Is it being tested by anyone else?

YcKe:
Seems like this would solve indeed the Qubino metering precision issue. Great work jotakar! Hope it gets merged soon.

jotakar:
Thanks
I suppose it's being tested or so. I'm using it and it works fine in my net.

gizmocuz:
Thanks

jotakar:
I am pleased to participate in this program.

PR #717, opened 2016-06-30
Add new hardware Comm5 MA-4200/MI-1000

fabiopaolini:
It was added a "hardware" to be used via serial port.
The Comm5 has a module (called MI-100) with Domoticz embedded, and within this module there is an internal serial communication with the hardware added in this patch.

### DIFF
--- a/main/RFXNames.cpp
+++ b/main/RFXNames.cpp
@@ -3336,6 +3336,7 @@ bool IsSerialDevice(const _eHardwareTypes htype)
 	case HTYPE_P1SmartMeter:
 	case HTYPE_Rego6XX:
 	case HTYPE_DavisVantage:
+	case HTYPE_S0SmartMeterUSB:
 	case HTYPE_OpenThermGateway:
 	case HTYPE_TeleinfoMeter:
 	case HTYPE_OpenZWave:
@@ -3347,6 +3348,7 @@ bool IsSerialDevice(const _eHardwareTypes htype)
 	case HTYPE_KMTronicUSB:
 	case HTYPE_KMTronic433:
 	case HTYPE_CurrentCostMeter:
+	case HTYPE_RAVEn:
 	case HTYPE_Comm5Serial:
 		return true;
 	default:
--- a/main/RFXNames.cpp
+++ b/main/RFXNames.cpp
@@ -212,7 +212,7 @@ const char *Hardware_Type_Desc(int hType)
 		{ HTYPE_RFXtrx868, "RFXCOM - RFXtrx868 USB 868MHz Transceiver" },
 		{ HTYPE_RFLINKTCP, "RFLink Gateway with LAN interface" },
 		{ HTYPE_Comm5TCP, "Comm5 MA-5XXX with LAN interface" },
-		{ HTYPE_Comm5Serial, "Comm5 MA-4XXX/MI-XXXX Serial interface" },
+		{ HTYPE_Comm5Serial, "Comm5 MA-4XXX/MI-XXXX Serial/USB interface" },
 		{ HTYPE_SolarEdgeAPI , "SolarEdge via Web API" },
 		{ HTYPE_CurrentCostMeter, "CurrentCost Meter USB" },
 		{ HTYPE_CurrentCostMeterLAN, "CurrentCost Meter with LAN interface" },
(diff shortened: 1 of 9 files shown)


main/RFXNames.cpp
```diff
@@ -3328,13 +3329,29 @@ const char *Get_Alert_Desc(const int level)
...
+	case HTYPE_MySensorsUSB:
+	case HTYPE_RFLINKUSB:
+	case HTYPE_KMTronicUSB:
+	case HTYPE_KMTronic433:
+	case HTYPE_CurrentCostMeter:
+	case HTYPE_Comm5Serial:
+		return true;
+	default:
+		return false;
+	}
```
gizmocuz, line 3356: This is not correct, for example you miss the HTYPE_RAVEn
Please double check 'all' your changes before committing
How about the HTYPE_S0SmartMeterUSB ?
What did you do ? retyping everything ? Better to just press some enters and then reformat the code....

main/RFXNames.cpp
```diff
@@ -212,6 +212,7 @@ const char *Hardware_Type_Desc(int hType)
 		{ HTYPE_RFXtrx868, "RFXCOM - RFXtrx868 USB 868MHz Transceiver" },
 		{ HTYPE_RFLINKTCP, "RFLink Gateway with LAN interface" },
 		{ HTYPE_Comm5TCP, "Comm5 MA-5XXX with LAN interface" },
+		{ HTYPE_Comm5Serial, "Comm5 MA-4XXX/MI-XXXX Serial interface" },
```
gizmocuz: Name your device with USB instead of serial, so also the below patch in the controllers are not needed

www/app/HardwareController.js
```diff
@@ -100,7 +100,7 @@ define(['app'], function (app) {
                      }
                 });
             }
-            else if (text.indexOf("USB") >= 0)
+            else if ((text.indexOf("USB") >= 0) || (text.indexOf("Serial") >= 0))
```
gizmocuz: Should be fixed, name the hardware with USB instead of Serial

www/app/HardwareController.js
```diff
@@ -785,7 +785,7 @@ define(['app'], function (app) {
...
```
gizmocuz: Should be fixed, name the hardware with USB instead of Serial

www/app/HardwareController.js
```diff
@@ -4119,7 +4119,7 @@ define(['app'], function (app) {
                             $("#hardwarecontent #hardwareparams1wire #OneWireSensorPollPeriod").val(data["Mode1"]);
                             $("#hardwarecontent #hardwareparams1wire #OneWireSwitchPollPeriod").val(data["Mode2"]);
                         }
-                        else if (data["Type"].indexOf("USB") >= 0) {
+                        else if ((data["Type"].indexOf("USB") >= 0) || (data["Type"].indexOf("Serial") >= 0)) {
```
gizmocuz: Should be fixed, name the hardware with USB instead of Serial

www/app/HardwareController.js
```diff
@@ -4273,7 +4273,7 @@ define(['app'], function (app) {
                 $("#hardwarecontent #divunderground").hide();
                 $("#hardwarecontent #divhttppoller").hide();
             }
-            else if (text.indexOf("USB") >= 0)
+            else if ((text.indexOf("USB") >= 0) || (text.indexOf("Serial") >= 0))
```
gizmocuz: Should be fixed, name the hardware with USB instead of Serial

gizmocuz:
Thanks for the patch ! Looking good, except a few things, see the Files Changed comment section

gizmocuz:
Could it be you based your code not on the latest domoticz repo code ?

main/RFXNames.cpp
fabiopaolini, line 3356: I have added the two missing enum's.

gizmocuz:
Thanks ! , Now only rename your hardware device to include USB instead of Serial, and remove the non-needed changes please

gizmocuz:
Thank you very much !

main/RFXNames.cpp
fabiopaolini: We have added a Serial device because the hardware is indeed a Serial one and we thought that it could bring some conflict.

gizmocuz: Might be true, but people will use an usb->serial cable anyway
But you see that this saves a lot of changes ;)

PR #718, opened 2016-07-01
Fix compilation error on mqtt

ffontaine:
Add a mising include_directories(${OPENSSL_INCLUDE_DIR}) in MQTT/CMakeLists.txt otherwise compilation fails if openssl is not in the standard include paths


gizmocuz:
Thanks !

gizmocuz:
Thanks again ;)

Issue #721, opened 2016-07-03
Add digest access authentication camera


dynasticorpheus:
Some camera's do not allow basic authentication and need digest. If I am correctly assuming that libcurl is currently pulling the images it might even be as simple as adding the **_CURLAUTH_DIGEST_** parameter?

Below works when run from command line
`curl --digest -u username:password -O http://192.168.1.10/cgi-bin/img-0.cgi`

dynasticorpheus:
Created my first pull request ever to try and solve this :)

gizmocuz:
Thanks looking good !

dynasticorpheus:
Confirmed, camera requiring digest auth now working as expected.

PR #722, opened 2016-07-03
Enable digest access authentication

dynasticorpheus:
Use http digest access authentication if basic authentication is not available/allowed. For issue #721


dynasticorpheus:
Some camera's do not allow basic authentication and need digest. If I am correctly assuming that libcurl is currently pulling the images it might even be as simple as adding the CURLAUTH_DIGEST parameter?

Below works when run from command line
curl --digest -u username:password -O http://192.168.1.10/cgi-bin/img-0.cgi

gizmocuz:
Thanks !

PR #724, opened 2016-07-04
Changes in uvc camera module

rkluwen:
Changes in camera module to allow for more usb cams at once on the Raspberry.

This is a quick and dirty patch. You can set the video device (video0, video1, etc.) in the username field of the camera settings. Note: Do not prepend /dev/...

It's quick and dirty because the web handling of the camera module is broken. It doesn't correctly url decode the parameters. I leave this as an excersise to the reader to fix this.

### DIFF
--- a/main/Camera.cpp
+++ b/main/Camera.cpp
@@ -216,13 +216,16 @@ bool CCameraHandler::TakeRaspberrySnapshot(std::vector<unsigned char> &camimage)
 	return false;
 }
 
-bool CCameraHandler::TakeUVCSnapshot(std::vector<unsigned char> &camimage)
+bool CCameraHandler::TakeUVCSnapshot(const std::string &device, std::vector<unsigned char> &camimage)
 {
 	std::string uvcparams="-S80 -B128 -C128 -G80 -x800 -y600 -q100";
 	m_sql.GetPreferencesVar("UVCParams", uvcparams);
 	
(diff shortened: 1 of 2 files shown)


gizmocuz:
Thanks for the patch, can i propose some changes, that also solves the user interface issue ?

Instead of the username, we can change the 'ImageURL', which is default (and still could be !) "uvccapture.cgi"

Now, if we add support like this:
uvccapture.cgi?device=video0
uvccapture.cgi?device=video1
uvccapture.cgi?device=xxx

Then without the "device" parameter it works like it is now, and with the parameter you can use it to specify which video source
I did not know you could add multiple UVC camera's to the pi...

rkluwen:
Sure, that's also possible. You could then also change the other uvw parameters on a per-camera basis. Now, this is only one setting for each camera.

rkluwen:
Anyhow, I think we will face the same problem as well in this case: The question mark won't show up right in the database. That's why I did it this way.

gizmocuz:
I think we should atob/btoa in javascript, or  encodeURIComponent ?

rkluwen:
Besides the javascript part (needs encodeURIComponent), also the parameter handling in Domoticz is broken (not properly url-decoded in the database).
I haven't looked into the matter very deeply. Just it occurred to me when I was making the patch.

gizmocuz:
Okey, let me try to make a patch based on your work... will report back asap

PR #725, opened 2016-07-04
Add Samsung SoC serial support

fabiopaolini:
Add support to Samsung chip, which uses the serial name as /dev/ttySAC


gizmocuz:
Thanks!

Issue #680, opened 2016-06-02
Add Google Cloud Pub/Sub data export


sbouchex:
The idea is, as HTTP and FibaroLink,to give the ability to export the data to Google Cloud Pub/Sub service (see https://cloud.google.com/pubsub/)

I can work on it

gizmocuz:
Looking good !
Maybe we should make a new base class for:
- Domoticz Remote push
- HTTP
- Fibaro Link
- Google Cloud

?

sbouchex:
Yep, I wil try to factorize as much code as possible

gizmocuz:
+1

gizmocuz:
Did you made some progress ?

sbouchex:
No but mostly due to a lack of time but I will have time soon.

I'm planning to use the google C++ API (see http://google.github.io/google-api-cpp-client/latest/) but it is only available on linux and not on Windows.
The library will be detected by domoticz's cmake in a folder "google-api-cpp-client" located outside domoticz folder like openzwave.

gizmocuz:
I think (hope) the library is cross platform, so that would be great!

sbouchex:
I saw some messages on the portability to Raspberry Pi but for sure, it's not working on Windows.

If we have it on linux at least, it will be great

sbouchex:
Well... This library seems to be outdated and in maintenance mode, so let's avoid it

Next plan : Use Google Python PubSub API and plug it into domoticz

gizmocuz:
Not all platforms run python, but... it could be a first step
Or maybe implement IFTTT ?

sbouchex:
Other option : java : the jvm is loaded into domoticz (with jni) and comes
with a jar in charge of sending data to pubsub.

Java is cross platform and may bring new ideas

About ifttt, i think it's already possible to do it by using the 'maker'
from ifttt ( https://ifttt.com/maker)

Seb

gizmocuz:
No, java is even worse... dont want to install a JRE when not needed ;)

sbouchex:
Could be optional...

Anyway, let's try with python and we'll see.

gizmocuz:
Thanks, it's included now!

PR #726, opened 2016-07-07
GooglePubsub export

sbouchex:
First Implementation of Google PubSub export !

This implentation is using the googlepubsub.py python from script/python and requires the google-api-python-client pyhon package.
Please read this documentation in the script before testing the package

Domoticz must be built with the option -DUSE_PYTHON=yes in order to activate python

see #680


gizmocuz:
Thanks for this great patch, all is looking good!! ... except
Could you revert/patch again the following files, because something strange happened there (to many changes)

**msbuild/domoticz.vcxproj**
1,268 additions, 1,266 deletions not shown because the diff is too large. Please use a local Git client to view these changes.

**msbuild/domoticz.vcxproj.filters**
3,611 additions, 3,602 deletions not shown because the diff is too large. Please use a local Git client to view these changes.

gizmocuz:
Still a lot of changes... strange...

sbouchex:
Seems to be related to the carriage return as the only changes are the 2 new files beeing added

gizmocuz:
Strange, in the .gitattributes file, the .vcproj file is declared to always have crlf, that is default for windows
Did you edit on windows ? or used a editor that changes the crlf ?

sbouchex:
I'm editing on Windows but git cloning with Eclipse and I'm wondering if the plugin is not respecting the carriage return... I'm going to change the CR on the windows file

PR #727, opened 2016-07-07
GooglePubsub export

sbouchex:
First Implementation of Google PubSub export !

This implentation is using the googlepubsub.py python from script/python and requires the google-api-python-client pyhon package.
Please read this documentation in the script before testing the package

Domoticz must be built with the option -DUSE_PYTHON=yes in order to activate python

see #680

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -202,6 +202,7 @@ main/domoticz.cpp
 main/EventSystem.cpp
 main/Helper.cpp
 main/HttpPush.cpp
+main/GooglePubSubPush.cpp
 main/localtime_r.cpp
 main/Logger.cpp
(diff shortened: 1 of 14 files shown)


sbouchex:
The Windows project files are incomplete and the build is not working on Windows

jsiegmund:
@sbouchex any documentation for this? I'm looking into controlling domoticz via Cortana and would like to reuse the pubsub stuff if that saves me from building another integration (at least for now) but I'm having trouble finding out how it works in the first place.

sbouchex:
Look at the script scripts/python/googlepubsub.py

jsiegmund:
Thanks. For now I've chosen a different path but I might get back on this :)

PR #728, opened 2016-07-07
Refactored code and fixed issue on switches

sbouchex:
- Refactored some code (removed useless variables and move common code to CPush class)
- Fixed an issue with switches : Send value was "Not supported yet" instead of the status value


gizmocuz:
Thanks!

PR #731, opened 2016-07-09
Allowing multiple Toon agreements

IgorYbema:
Added 'agreement' in a Toon installation to allow another Toon in same account to be used. Default agreement is 0.

When having multiple Toon thermostats in your account the agremeent number shifts to the number of Toon's (with index 0). So the second Toon is '1' etc. 

Adding multiple Toon thermostats into Domoticz hardware, each with different agreement numbers allows Domoticz to pull information from the differents Toon thermostats.

### DIFF
--- a/hardware/ToonThermostat.cpp
+++ b/hardware/ToonThermostat.cpp
@@ -83,9 +83,10 @@ const std::string TOON_SWITCH_ALL = "/toonMobileBackendWeb/client/auth/smartplug
 //	STATE_HOLIDAY	//4
 //};
 
-CToonThermostat::CToonThermostat(const int ID, const std::string &Username, const std::string &Password) :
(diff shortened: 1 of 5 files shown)


gizmocuz:
Thanks !,
Is it possible you do some safety checking, as when you enter a to high ID, it will crash

root["agreements"][m_Agreement]["agreementId"]

check if 'root["agreements"][m_Agreement]' exists

For example when m_Agreement = 12345, this causes an out-of-bound error

IgorYbema:
Ok done. Used the already existing size() function to check if the array is at least large enough to get that m_Agreement from the array. Does not need a if exists after that in my opinion.

gizmocuz:
Thanks !!

PR #732, opened 2016-07-10
Add support of nodon inwall module - 2 channels (Enocean)

qix67:
The patch decodes VLD (D2-04) to detect module channels.
I think it also should work for single channel module.
Beware: during detection of the 2 channels, 2 new devices
are added but domoticz only show one box to add device.
Both devices can be found in setup/devices page.

WriteToHardware was modified to use "Dimming" profile (A5-38-02)
only if a switch is a dimmer. Profile F6-02-01 is used
for basic switch (like nodon inwall module).

I'm not sure it really is necessary to add sTypeNodon as new subtype.

Pairing must be done like with a real switch:
1) in domoticz, use "learn light" to detect the module. switch the enocean module on or off to make it detectable.
2) Create a virtual switch in domoticz
3) put the enocean module into learning mode
4) press the virtual switch in domoticz
5) leave module learning mode
6) attach the light detected during step 1 as subdevice of virtual switch

Note: during pairing, domoticz complains about unhandled rorg (d4)
but it does not disturb learning and I found no explanation inside
Enocean Equipement Profile 2.6.5 PDF (what a mess^Wmind-blowing document)

### DIFF
--- a/hardware/EnOceanESP3.cpp
+++ b/hardware/EnOceanESP3.cpp
@@ -228,6 +228,21 @@ typedef enum
 #define S_RPS_RPC 0x0F
 #define S_RPS_RPC_SHIFT 0
 /*@}*/
+
+#define F60201_R1_MASK 0xE0
+#define F60201_R1_SHIFT 5
+#define F60201_EB_MASK 0x10
+#define F60201_EB_SHIFT 4
+#define F60201_R2_MASK 0x0E
+#define F60201_R2_SHIFT 1
+#define F60201_SA_MASK 0x01
+#define F60201_SA_SHIFT 0
+
+#define F60201_BUTTON_A1 0
+#define F60201_BUTTON_A0 1
+#define F60201_BUTTON_B1 2
+#define F60201_BUTTON_B0 3
+
 /**
  * @defgroup status_rpc Status of telegram (for 1BS, 4BS, HRC or 6DT telegrams)
  * Bitmasks for the status-field, if ORG = 1BS, 4BS, HRC or 6DT.
@@ -486,6 +501,18 @@ void CEnOceanESP3::Do_Work()
 
 void CEnOceanESP3::Add2SendQueue(const char* pData, const size_t length)
 {
+#ifdef ENABLE_LOGGING
+	std::stringstream sstr;
+
+	for (int idx=0;idx<length;idx++)
+	{
+		sstr << std::hex << std::uppercase << std::setw(2) << std::setfill('0') << (((unsigned int)pData[idx]) & 0xFF);
+		if (idx!=length-1)
+			sstr << " ";
+	}
+	_log.Log(LOG_STATUS,"EnOcean Send: %s",sstr.str().c_str());	
+#endif
+
 	std::string sBytes;
 	sBytes.insert(0,pData,length);
 	boost::lock_guard<boost::mutex> l(m_sendMutex);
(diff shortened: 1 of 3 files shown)


gizmocuz:
Thanks, great patch !
Could you make use of the pTypeGeneralSwitch and sSwitchGeneralSwitch
If all dimmer hardware supports the dimming profile, maybe it is best that when the users edits the switch and sets the type = dimmer, that you handle it like that.
Or when you add the switch, you specify directly it is a dimmer switch ?

qix67:
I'll take a look next week-end to pTypeGeneralSwitch/sSwitchGeneralSwitch.

When I added my virtual switch, I used on/off type, not dimmer type.

At first, I wanted to use subtype but it is not possible to specify one with UI. I also tried to use the subtype of the subdevice but it looks a bit too complex and I didn't know how to handle case where multiple subdevices are added to a switch.

gizmocuz:
Thanks!

qix67:
I'd just take a look to pTypeGeneralSwitch and sSwitchGeneralSwitch but code lacks comment to make things clear.

I don't know if other home automation protocols work like enocean but here, the inwall module is only reporting light status, it is not possible to send a command to it, domoticz log contains "EnOcean: Can not switch with this DeviceID, use a switch created with our id_base!...". The virtual switch only shows light status thanks to its slave device. When I toggle virtual switch, it sends its own code "over the air" and only module having registered domoticz virtual switch code really will toggle.

Where should I use pTypeGeneralSwitch/sSwitchGeneralSwitch ? If I set it on the module itself, I won't be able to toggle the light because I won't have any usable deviceID in id_base. On the other side, when I manually create the virtual switch, I see no place to set pTypeGeneralSwitch in the web UI.

I know at least one case possibily not working with my code. If a virtual switch is attached to 2 slaves, one allowing diming and one not, if the virtual switch is not of dimmer type, the device allowing diming should work in the same manner as "normal" light (on=100%, off=0%). If the virtual switch is of dimmer type, the device not allowing diming may not power on at all, depending on enocean profile it supports.

qix67:
I just found another problem, when domoticz starts, it does not query module current status therefore, if light is toggled when domoticz is stopped, it does not display the correct status and then cannot send the correct "button" to change status light because sending an "off" to a powered off light does not make module send it status because it is unchanged (on its side). The light is stucked until it is "manually" put into the state domoticz knows.

gizmocuz:
I don't think this is a part of enocean nor domoticz

jlstanus:
Thanks for your patch. My domoticz is waiting for it :+1:

qix67:
I nearly have completed a new version with support of enocean profile. I just try to find how to retrieve current status of "light" when domoticz starts but I start to have a better knowledge of both domoticz and enocean protocol.

jlstanus:
Nice, if you could write a little tutorial how it works. it should be very interesting because there is few developers about Enoncean in Domoticz. So if we have documentation about it we could implement more and faster the new devices.
https://www.domoticz.com/wiki/EnOcean

gizmocuz:
@qix67, how's it going with the patch ? I dont think you can read the current light state with enocean...

gizmocuz:
Please reopen when you got time.
Do not modify rfxtrx.h, you can use the pTypeGeneral (Switch)

jlstanus:
Oh nooo :'(

qix67:
Hi,

sorry for the delay, I was very busy during my holiday (making my garden).

I made few changes to my first patch and I do not need sTypeNodon anymore. I store enocean device profile in the table enoceansensor (already exists).

I will try to commit my changes during the next week-end. 

About enocean, I think your right, I found no way to read light state. A profile seems to support this but my switch does not seem to understand it.

gizmocuz:
Great ;) welcome back

manu-t:
Hi,
It worked with a Nodon smart plug ASP-2-1-01.
I hope your patch will be soon in the master branch, keep up the good work!
Thank you!

bend94:
Hi,

Sorry for this question i want to include this patch in my domoticz
i put the file into domoticz/hardware/EnOceanESP3.cpp

Now i want to recompile all but how to do this ?

Thanks
Regards

manu-t:
Hi,
Since I am kind of noob concerning git (and vcs in general) I am not very confortable with this type of git usage, so I might have used:
git checkout -b nodon-inwall c875d1def49494f7961d48cf3c150054a0432fcf
then you have to download the libraries and compile it

bend94:
Hi

thanks

2017-01-03 0:38 GMT+01:00 manu-t <notifications@github.com>:

PR #735, opened 2016-07-14
Reduce HTML inside JS, Angularize application : temperatures pages

gaeljw:
Hello,

I've been working on improving the web UI using full Angular rather than mixing HTML inside JS.
This is a first PR for the temperatures pages, other pages will follow and there is still work to do on temperatures pages.

Everything is working/displaying as before for all test cases I could imagine.

I understand this is a big PR, so feel free to ask any question :)


jrich523:
I love that you're doing this :)

gizmocuz:
+100, great ! Super you find the time for doing this... In the end i think all should be faster, more clear, and for sure more generic ;)

gaeljw:
Thank you ! It will take some time but I sure think we can clear and simplify the code :)

gizmocuz:
O no... my temperature tab is empty now...

gizmocuz:
Pfffiew... okey, a browser cache issue...

rkluwen:
Gael, I just reviewed this patch, while merging the Websockets branch. This stuff is great! Exactly everything we want!
You might want to take a look at this document: https://drive.google.com/open?id=1Hn_ng_cGLeOk8sAGiuhXtz5Tc4H53TSgrace3CCRZHU

gaeljw:
Thank you @rkluwen, I will have a look at your document as soon as I can and share my thoughts ;)

gizmocuz:
I am experiencing a problem.....
It looks (sorry I can verify at the moment) two items are skipped(for what ever reason)
But this results in two blank spots... 
Expected result was that everything was shifted

gaeljw:
Oh right, I think I got this once. Did you move the widgets ?
To be sure, you mean some widgets are shifted and they leave a blank spot but they don't disappear, right ?
I'll try to reproduce and find a scenario to understand what's going on..

rkluwen:
@gizmocuz, what are the steps to reproduce?

gizmocuz:
I open the temperature page on an iPAD... and i see two blank spots...
image

gizmocuz:
currently i am unable to post the json response, but maybe it does not support the 'type' or 'subtype' of these two, but the break counter (or position) is still increased

gaeljw:
I think this might just be due to some "invalid" bootstrap HTML/CSS and/or drag&drop JS. 

Below is generated output for a widget : 
`<div ng-repeat="t in ctrl.temperatures track by t.idx" class="ng-scope" style="">
            <dztemperaturewidget class="span4 ng-isolate-scope ui-draggable ui-draggable-handle ui-droppable" ...>    
<section>
...
    </section>
</dztemperaturewidget>
        </div>`

The `ng-repeat` div and `dztemperaturewidget` are what differs from other pages. I'll investigate tomorrow.

gaeljw:
JSON response and HTML output would be great to fix this. I don't reproduce with Chrome emulating an iPad :(

gaeljw:
I found a scenario to reproduce : it happens when I drag any widget into a widget which is in the right column.
I found a quick fix (see PR #745, I "rebuild" all the widgets instead of updating ones who need it) but I still don't understand completely what's going on.

Please confirm it fix your issue too !

PR #741, opened 2016-07-16
Improved zwave topology usability

wolph:
Just a few changes to make the z-wave topology a bit more usable.
1. Disabled paging since 10 items per page is really annoying if you manually have to enter IDs and don't have a visible lookup table
2. Added proper tooltips for item removals
3. Added details about the nodes and groups when removing
4. Added details about the nodes when adding
5. Increased the content size of the frame to fit more items
6. Added the highcharts.js include since domoticz.js generates errors without it


gizmocuz:
Thanks!!

PR #745, opened 2016-07-20
Fix PR #735 : blank spot in some cases instead of temperature widget

gaeljw:
This is a fix for the issue discussed in PR #735.

### DIFF
--- a/www/html5.appcache
+++ b/www/html5.appcache
@@ -1,5 +1,5 @@
 CACHE MANIFEST
-# ref 1269
+# ref 1270
 
 CACHE:
 # CSS
--- a/www/views/temperature.html
+++ b/www/views/temperature.html
@@ -107,9 +107,7 @@
 
 	<div id="tempwidgets" class="row divider">
 		<h2 ng-show="ctrl.temperatures.length == 0" data-i18n="No Temperature sensors found or added in the system...">No Temperature sensors found or added in the system...</h2>
-		<div ng-repeat="t in ctrl.temperatures">
-			<dztemperaturewidget class="span4" id="{{t.idx}}" item="t" tempsign="config.TempSign" windsign="config.WindSign" ordering="config.AllowWidgetOrdering" dragwidget="DragWidget(idx)" dropwidget="DropWidget(idx)"></dztemperaturewidget>
-		</div>
+		<dztemperaturewidget ng-repeat="t in ctrl.temperatures" class="span4" id="{{t.idx}}" item="t" tempsign="config.TempSign" windsign="config.WindSign" ordering="config.AllowWidgetOrdering" dragwidget="DragWidget(idx)" dropwidget="DropWidget(idx)"></dztemperaturewidget>
 	</div>
 
 </div>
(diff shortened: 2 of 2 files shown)


gizmocuz:
Thanks

gizmocuz:
Thanks, i installed the new beta version.
It is so strange, but also consistent....
On my windows system with chrome, no issue
on my ipad with safari and with chrome, both show the same issue

Are you still making use of the 'row' class ? (3 items, than a div-row break)

Also note, i do not miss any sensor, it just displays blank spots

Question:
Do we need the allow-reordering, temp/wind sign ? as these are all available in the root-scope

gaeljw:
Oh right ! I didn't notice this but I guess this is what causes the issue : I now use only one "row" which contains all widgets. I'll make this workaround/fix asap.

About the 3 attributes allowOrdering, temp and wind signs : I agree we should not have to set them on each widget. I'll implement a global "config" service or something like that because an angular widget/directive should not have a direct access to root scope IMHO.

gizmocuz:
Great!
I think you can just get those variables from $scope as they are available in the rootscope 
See app.js

gaeljw:
@gizmocuz, before moving back to the old generated HTML (one row for each 3 widgets), I would like to try a cleaner Bootstrap way.

Could you test by replacing (in _temperature.html_) : 

```
<dztemperaturewidget ng-repeat="t in ctrl.temperatures" class="span4" id="{{t.idx}}" item="t" tempsign="config.TempSign" windsign="config.WindSign" ordering="config.AllowWidgetOrdering" dragwidget="DragWidget(idx)" dropwidget="DropWidget(idx)"></dztemperaturewidget>
```

With : 

```
<div ng-repeat-start="t in ctrl.temperatures" class="clearfix" ng-if="$index % 3 == 0 && !$first"></div>
<dztemperaturewidget ng-repeat-end class="span4" id="{{t.idx}}" item="t" tempsign="config.TempSign" windsign="config.WindSign" ordering="config.AllowWidgetOrdering" dragwidget="DragWidget(idx)" dropwidget="DropWidget(idx)"></dztemperaturewidget>
```

I can make this a PR if you prefer but since I can't reproduce and I'm not sure it fixes the issue.. Just tell me :)

rkluwen:
Just my $0.02: I think, for responsiveness purposes, gaeljw's original solution is a better one. If we can make it show up correctly.
Why would you expect to have 3 widgets horizontally, always?

gaeljw:
We can expect 3 widgets always because of the `span4` class (in Bootstrap, a `row` is divided in 12 columns, see http://getbootstrap.com/css/#grid).
But I agree we would not normally need to do anything after each 3 widgets. There is probably a CSS issue somewhere but I don't really have time to find it :/ , so for now to avoid the issue on iPad I prefer to move back to the old way (or intermediate as I proposed above).

gizmocuz:
Yes, great, this solved the issue !

gaeljw:
Ok, I'll make the PR for the temperature and weather page in few minutes :)

gizmocuz:
Great!

rkluwen:
But this is exactly what boost is for, right?

PR #747, opened 2016-07-21
Cursor pointer for not active subtabs

piokuc01:


gizmocuz:
could you show an image from before/after? i see pointers everywhere...

piokuc01:
Active tab before and aftre changes is the same:
active bef

Not active tabs before
not active bef

Not Active tabs after
not active af

gizmocuz:
i like the not-active-tabs before mouse cursor

PR #749, opened 2016-07-21
Fix webserver to fix silly typo that breaks hardware updates

llagendijk:
Please merge this simple fix for WebServer.cpp that broke hardware updates.


rkluwen:
Great catch!

llagendijk:
It took me yesterday some 2 hours to find out why the hardware update for the Goodwe power inverter  I am writing did not work. It was humiliating to see that I broke it myself when I added the Ec3k driver.....

gizmocuz:
Thanks... Yes this is a classic case that is very easily overlooked, great find!

PR #751, opened 2016-07-23
Reduce HTML inside JS, Angularize application : weather page

gaeljw:
Hello,

Still working on improving the web UI using full Angular rather than mixing HTML inside JS.
This is a first PR for the weather page. There is still place for improvements.

As always, feel free to ask any question :)

Also note, that I may soon refactor some things which are common in weather page and temperatures page.


rkluwen:
Great!

gizmocuz:
Thanks, maybe you can also check here the row class? (See the temperature discussion)

Issue #720, opened 2016-07-03
HTTP Service is hanging from time to time


d00616:
I'm using domoticz commit 711a11889d09d42efeca5b97b21712a5ae5dc8a8 (Fri Jul 1 20:10:34 2016 +0200) with openzwave a379e3a799719fa5d7c9958d2a16d98482e8d015 (Tue Jun 28 23:40:28 2016 +0800).

From time to time the HTTP service doesn’t respond. When i run strace, the running process reports a endless loop with nanosleep:

```
# strace -p 21006
Process 21006 attached
nanosleep({1, 0}, NULL)                 = 0
nanosleep({1, 0}, NULL)                 = 0
nanosleep({1, 0}, NULL)                 = 0
```

A lot of STREAMs are reported by netstat:

```
# netstat -anp | grep 21006 | grep STREAM | wc -l
997
```

Lsof reports 15986 open sockets.

gizmocuz:
No idea what to do with this... i can not reproduce it.
Can you debug the issue and make a possible pull request?

d00616:
Thank you for testing. I try to debug it.

d00616:
I thin this is a problem with a socketpair opened every 30 seconds in MQTT/net_mosq.c When no mqtt service is available this socket pair is not closed.

I think the problem is in _mosquitto_connect_init, but i have no Idea how to fix it.

gizmocuz:
Could you post this on the mosquito forum?

d00616:
I have searched in bugzilla. It looks this bug is already fixed.
https://bugs.eclipse.org/bugs/show_bug.cgi?id=484692

 _mosquitto_connect_init looks different at https://github.com/eclipse/mosquitto/blob/master/lib/mosquitto.c#L392

gizmocuz:
Great find!
Could you make a git pull request for domoticz ?

gizmocuz:
Is this issue solved?

d00616:
Yes it is. I have tested it before i created the pull request.

gizmocuz:
Thanks for the feedback, and issue closing

PR #752, opened 2016-07-24
Update mqtt

d00616:
Updated MQTT library to an actual release.

I have modified the include path to allow copying files without modification from original source. Same for WWW_ENABLE_SSL => WITH_TLS.

Fix for https://github.com/domoticz/domoticz/issues/720

### DIFF
--- a/MQTT/CMakeLists.txt
+++ b/MQTT/CMakeLists.txt
@@ -3,7 +3,6 @@ cmake_minimum_required(VERSION 2.8.4)
 
 project (MQTT C)
 
-include_directories("${CMAKE_CURRENT_SOURCE_DIR}")
 SET(MQTT_INCLUDE_DIRS "${CMAKE_CURRENT_SOURCE_DIR}")
 SET(MQTT_LIBRARIES mqtt)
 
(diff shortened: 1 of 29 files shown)


gizmocuz:
Thanks,  but we got some issues with the build system (windows), can not find the mqtt files

d00616:
Any idea how to fix it?

gizmocuz:
Maybe only patch the needed line to fix the memory issue?
Or add the mqtt as include folder in the Windows project 
Did you tested under Windows?

d00616:
There is more than a line to replace. My first try was replace only the contents of the init function. This results into consuming more sockets. In my opinion this isn't a good idea to fork this code by patching code this way.

I have only Linux systems. I’ve found a hint http://stackoverflow.com/questions/1447423/how-to-include-directories-in-cmake-generated-visual-studio-projects so i have pushed this change to github. When build is ok i merge my changes of CMakeLists.txt into one commit.

gizmocuz:
Windows is using visual Studio 
And this uses project files 
Best to download vs2015 and test

d00616:
Sorry without Windows i cannot test VS2015. I have changed a configuration file manually.

d00616:
Windows build should be acceptable but i see warnings like this: "..\MQTT\mosquitto.c(58): warning C4273: 'mosquitto_lib_version': inconsistent dll linkage [C:\projects\domoticz\msbuild\domoticz.vcxproj]"

gizmocuz:
Thanks, hope all is working correct now

PR #754, opened 2016-07-26
[fix] In some cases, widgets could not be aligned

gaeljw:
The final and correct fix for PR #735 !
See also #745 and #751 .


gizmocuz:
Thanks for the fast action!

rkluwen:
Great work.

PR #757, opened 2016-07-29
Update Kodi.cpp

narfight:
shutdown action if multi possibility possible


gizmocuz:
Please be more clear about your pull request...

narfight:
Sure !

Before my PR, the Priority of action to poweroff is :
- suspend
- hibernate
- Shutdown

On Windows OS, suspend is always actif. In Kodi, I can't disable this option. In lot of cases, the suspend action set Windows in logon window without HDMI sound !

With my PR, the Shutdown action is set to prior.

gizmocuz:
Thanks!

PR #760, opened 2016-07-30
Fixing removal of association

sjoerdmulder:
JSON service expects lowercase argument [L4177](https://github.com/domoticz/domoticz/blob/master/hardware/OpenZWave.cpp#L4711) and [Object%20object] as value will not work


gizmocuz:
Thanks!

unheilig:
Hi,

unfortunately version 3.5406 (build 278c1b9) on Synology NAS is not removing groups from devices. 
did anyone of you succeed in this operation?
thank you

Issue #552, opened 2016-03-09
Compress static web content at buildtime


gaudryc:
I will submit a change request allowing :
- the web server to always deliver the gzipped version of a js or css file ;
- git to manage the textual version of js or css file, so we could have access to the history and we could make diff on these files.

It consists of modifying the building process to always compress the files with extensions .js, .json and .css, in the www/ folder.

The compression process consists in :
- compressing the file to .js.gz, .json.gz or .css.gz ;
- change the timestamp of the .gz file with the timestamp of the uncompressed file.

Under OSX and Linux, the compression is done with "gzip" and the timestamp change is done with "touch".
Under Windows, the compression is done with "7z.exe" and the timestamp change is done with "touch.exe" (coming from http://stevemiller.net/apps/).

With all these changes, we can do the whole compression at buildtime :
- by modifying the CMakeLists.txt for OSX and linux environment, or any build environment with CMake support.
- by adding a PostBuildEvent in the Visual Studio solution.
  In both cases :
- the compression is done only when the uncompressed file is newer than its compressed version, or when the compressed version does not exist.
- any new .css, .js or .json will be compressed before packaging.

When building using cmake, the "make clean" command only removes compressed files that have been generated using cmake.

What do you think about that ?

gizmocuz:
Great idea!
This is only in 'release' mode ? I mean how can will still edit html/js for development ?

actually this is needed when making the archive/installer (for example in makebeta for the raspberry, or appveyor/trivis)

gaudryc:
In Visual Studio you can define the postbuildevent only in release configuration if you want.
With CMake, there is no Debug/Release choice at this time, so the compression is done each time the "cmake+make" is done (if changes have been made).

At this time, with my solution, whatever the type of build, the uncompressed and compressed file are both present in the www/ before packaging domoticz (with tar or the package project using Visual Studio).

gizmocuz:
Correct, would it be not better for windows that the NSIS build script is doing this ?

gaudryc:
If you make it too late I think you won't be able to debug the "distribute gzip content" feature without packaging and installing first. If you compress at buildtime you can run domoticz from the source and test it directly.

gizmocuz:
I think we only need to debug the making of the .gz files once ... but...
But also on compile time, i want to edit the json/html/css/images everything,
but when releasing/packaging, then it would be great that everything was compressed

gaudryc:
[...]

PR #762, opened 2016-07-31
Have textfiles in VCS instead of gzipped versions

sjoerdmulder:
According to #552 all files are compressed at buildtime (and this makes sense). However some files in the repository where still compressed. This change gunzips them to allow better maintainablility


gizmocuz:
217 changed files, can't be right ;)
From my current location i am not even able to look at the changes

The i18n files do NOT have to be unzipped, because we translate them at transifex online

If an HTML developer which to tweak/edit the html/javascript code, it should be not so difficult to unzip them and delete the .gz file right ?

gaeljw:
I don't know the Domoticz build process very well but I agree it's not helping maintainability to have these HTML and JS files compressed because we don't have any history on these files.

gizmocuz:
That will only be needed for the web(gui) part, not for the translations, you can see the versioning at the transifex website, and not for the html/javascript files, if there is a modification for lets say.... highcharts, that means, a new library has been used.
Also there is a very good reason for having these files compressed... we are also running on arm systems

gaeljw:
Indeed, I only know about the web gui part in Domoticz. And as you said it should only be for proper domoticz code, not for external libraries nor for translations which are managed by transifex (by the way, is there any info about that somewhere ? how can we help translating ?).

Actually, this sums up to only 4 files : `www/js/domoticz*.js.gz`, everything else is either not compressed or an external library, right ?

rkluwen:
Sure, info about the Domoticz Transifex project is here: https://www.transifex.com/domoticz/domoticz/dashboard/.

You can just sign up for free and join the project and start translating.

sjoerdmulder:
Didn't know about the transifex, will revert the i18n files.

Still i believe all javascript should be uncompressed in the repository, and compressed during the build. Even better would be to introduce a tool like bower.io to manage the dependencies.

Currently there are a gazzilion dependencies just dropped (and zipped) in the libs folder, without knowing which file is actually used (because IDE's don't search in the binary files). If performance was really a concern focus should be on cleaning up these dependencies and removing unused ones.

But then again it's only a issue when refreshing the 'browser cache' feature (manifest) and then the transfering of the files might be taking a longer time. Otherwise they are all cached by the browsers and not refetched every request.

sjoerdmulder:
Example of the importance: Search for the `copy_csv_xls_pdf` SWF (60KB flash file) in the project. It's not referenced anywhere in the project so save to remove?

Now search for the same file in the PR, you will see that it is used by the `dataTables.tableTools.min.js` file, since it's unzipped your editor will find it.

Actually the `dataTables.tableTools.min` contains the `ZeroClipboard 1.0.4` but there is also a separate `ZeroClipboard.js` so these are duplicate?

gizmocuz:
Sure you can use "total commander" and search also in archives.. ? (or another app)

ldrolez:
Yes at least having domoticz*.js.gz files uncompressed would make patching easier :)

PR #763, opened 2016-08-02
Recommended patch for everyone using MyDomoticz

rkluwen:
Some bugs fixed, but most importantly reconnecting works like expected again.
Also shutting down Domoticz goes more smooth (or it is my imagination?).


rkluwen:
This patch was tested on Windows. I don't expect any major Pi problems. But my Pi currently doesnt have a proper build system.

gizmocuz:
Thanks!

PR #764, opened 2016-08-02
Prohibit Python script causing crash of domoticz

janakelarsson:
As in title plus working Python script template


gizmocuz:
Thanks!

PR #767, opened 2016-08-03
Goodwe inverter web api

llagendijk:
Please consider adding the driver for Goodwe solar inverters. 
Kind regards, Louis


gizmocuz:
Thanks for the great work !

two suggestions:
- when using an array, or any type that uses size(), like in your for loops, make it size_t instead of unsigned int

`unsigned int i = 0; i < root.size();`

becomes

`size_t i = 0; i < root.size();`

maybe if you make the string (unit) a std::string, you can use find, index, or other functions, maybe thats a bit cleaner?

llagendijk:
On Fri, 2016-08-05 at 05:59 -0700, gizmocuz wrote:

> Thanks for the great work !
> 
> two suggestions:
> 
> > when using an array, or any type that uses size(), like in your for
> > loops, make it size_t instead of unsigned int
> 
> unsigned int i = 0; i < root.size();
> 
> becomes
> 
> size_t i = 0; i < root.size();

Thanks for the hint. I am not experienced with C++ (the Ec3k and Goodwe
stuff were my first exposure to C++). Will keep this in mind.

> maybe if you make the string (unit) a std::string, you can use find, index, or other functions, maybe thats a bit cleaner?
> Well because of the fact that std::stof is c++, I had to revert to the
> old c-style functions, returning a char*. And converting char \* units
> back to std::string seemed a bit superfluous when a strcasecmp/_stricmp
> was all that was needed. My first implementation using std::stof used
> std::string all the way, although the case insensitive string
> comparison was more complex.
> I would have preferred to use std::string all the way, but that did not
> seem possible (but again: I am a C++ novice so I may have overlooked
> the obvious solution)
> 
> 
> Kind regards, Louis
> 
> 
> {"api_version":"1.0","publisher":{"api_key":"05dde50f1d1a384dd78767c55493e4bb","name":"GitHub"},"entity":{"external_key":"github/domoticz/domoticz","title":"domoticz/domoticz","subtitle":"GitHub repository","main_image_url":"https://assets-cdn.github.com/images/modules/aws/aws-bg.jpg","avatar_image_url":"https://cloud.githubusercontent.com/assets/143418/15842166/7c72db34-2c0b-11e6-9aed-b52498112777.png","action":{"name":"Open in GitHub","url":"https://github.com/domoticz/domoticz"}},"updates":{"snippets":[{"icon":"PERSON","message":"@gizmocuz in #767: Thanks for the great work !\r\n\r\ntwo suggestions:\r\n- when using an array, or any type that uses size(), like in your for loops, make it size_t instead of unsigned int\r\n\r\n`unsigned int i = 0; i \u003c root.size();`\r\n\r\nbecomes\r\n\r\n`size_t i = 0; i \u003c root.size();`\r\n\r\nmaybe if you make the string (unit) a std::string, you can use find, index, or other functions, maybe thats a bit cleaner?\r\n\r\n"}],"action":{"name":"View Pull Request","url":"https://github.com/domoticz/domoticz/pull/767#issuecomment-237843830"}}}

llagendijk:
done, replaced the unsigned int by ArrayIndex as that is what Json:: likes and rewrote the power/energy parsers with pure C++. I gave up on the idea to make unit parsing case insensitive: it is not required.
Please consider merging. Thanks again for the feedback.

gizmocuz:
Thanks!!

www/views/hardware.html
```diff
@@ -1250,6 +1250,39 @@ <h1 data-i18n="Devices">Devices</h1>
...
+		<div id="divgoodweweb">
+                        <br>
+                        <table class="display" id="hardwareparamsgoodweweb" border="0" cellpadding="0" cellspacing="20">
+                                <tr>
+					<td align="right" style="width:110px"><label><span data-i18n="username">Goodwe-power Username</span>:</label></td>
+					<td><input type="text" id="username" style="width: 170px; padding: .2em;" class="text ui-widget-content ui-corner-all" /></td>
+                                </tr>
+                        </table>
+                </div>
+<div id="divwinddelen">
```
sjoerdmulder, line 1262: Now, winddelen is duplicated? Its also already above this fragment

llagendijk, line 1262: Fishy, I don't know what happened here. Will you issue a pull request or shall I do that?

gizmocuz, line 1262: Please do

llagendijk, line 1262: consider it done. I will submit the pull request today.

PR #768, opened 2016-08-04
Crashuponwebserverstop when an exception occurs

rkluwen:
A little background about this patch: In the webserver helper class, when the connection is stopped the CWebServer instance is deleted. At this time, the Do_Work() thread is still running.
So if -while stopping the server- an exception occurs, the catch handler will access member variables of the meanwhile already deleted object.
Two things are done to solve this:
1. The CWebServer class is made a shared pointer class, so it will only get deleted once no thread access it anymore.
2. In the StopServer() method, a join is introduced on the Do_Work() thread.

### DIFF
--- a/main/WebServer.cpp
+++ b/main/WebServer.cpp
@@ -614,6 +614,9 @@ namespace http {
 				if (m_pWebEm == NULL)
 					return;
 				m_pWebEm->Stop();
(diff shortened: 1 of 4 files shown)


gizmocuz:
Thanks!!

PR #772, opened 2016-08-05
small changes in BleBox, OpenWeatherMap, SatelIntegra

szczukot:


gizmocuz:
Thanks ! I know it is a lot to ask, but its better to make separate pull requests when working on different parts (OpenWeatherMap, Satel, and BleBox)

PR #775, opened 2016-08-06
Remove duplicated code in www/views/hardware.html

llagendijk:
Please commit the fix for the duplicated code in www/views/hardware.html

Thanks
Louis


gizmocuz:
Thank you!

PR #776, opened 2016-08-08
Fix the error  File '/ozwcp/favicon.png': No such file or directory

xbeaudouin:
Fix error: Webserver: File '/ozwcp/favicon.png': No such file or directory (2)
When loading OpenZWave Panel page


gizmocuz:
Thanks!

PR #777, opened 2016-08-08
Fix aligment position button "BACK" and title report to set on all pa…

piokuc01:
…ge the same

On page report


gizmocuz:
Plz post a before and after image

piokuc01:
Before
before

After
after

gizmocuz:
Thanks!

PR #780, opened 2016-08-09
Add to other tabs words to title "Setup"

piokuc01:
All tabs has the same tittle with words setup


gizmocuz:
You can not do it like this, because its based on i18n, and i also like it the way it is

PR #781, opened 2016-08-09
[BleBox] implemented Light Box S

szczukot:


hardware/BleBox.cpp
```diff
@@ -168,7 +200,7 @@ std::string BleBox::IPToHex(const std::string &IPAddress)
 	std::vector<std::string> strarray;
 	boost::split(strarray, IPAddress, boost::is_any_of("."));
 	if (strarray.size() != 4)
-		return "0.0.0.0";
+		return "";
```
gizmocuz, line 203: It seems your not handling this further in the code ? Is that correct? (its inserted empty in the database, should the insert take place?)

szczukot, line 203: This "return" should never happen. 
It is only for safety.

gizmocuz, line 203: Hmmm then my previous fix was better, but you better handle it.
I also would like to ask if you could use the default'StringSplit' method used in domoticz, instead of the boost one
Also, you input an IPAddress like (1.2.3.4), and then in this function ouput 1.2.3.4 ?
Not sure if i can see a difference...

szczukot, line 203: I didn't see StringSplit (i changed to this method)
I input ip adress but result is hex for example "0C1B2E98". I save it on this form in database, because later i must get this ip, when handle WriteToHardware - and decode it in this method std::string BleBox::GetDeviceIP(const tRBUF *id)

gizmocuz:
Thanks

PR #782, opened 2016-08-09
HTU21D: 1 byte should be sent for commands instead of 2.

ldrolez:
A try to fix the error message HTU21D: Error reading temperature! as discussed here: https://www.domoticz.com/forum/viewtopic.php?f=6&t=12916&p=92847#p92847


gizmocuz:
Thanks!

PR #783, opened 2016-08-10
Adding more hardware monitor stuff on freebsd using linprocfs emulation.

xbeaudouin:
FreeBSD have Linux procfs emulation. This allow domoticz using this optional procfs to see some CPU usage and more.


gizmocuz:
Thanks!

Issue #778, opened 2016-08-08
Bug on Level for switch


szczukot:
I don't understand level for switch
Is it may be in the values of 0-15 ?
If so, here it is a bug : 
https://github.com/domoticz/domoticz/blob/master/hardware/DomoticzHardware.cpp#L520
?
should be 
double rlevel = (15.0 / 100.0)*Level;

Now, when i sent level = 100, i see on percentage (blind) 106%

Fantom

szczukot:
Does it a bug or not ?
We wrote about this a few days ago:
https://github.com/domoticz/domoticz/pull/785

gizmocuz:
No, it is an implementation issue where possible some hardware classes have a problem with, and as far as those that i can test (kaku,zwave) i see no problems

PR #785, opened 2016-08-10
[BleBox] fix for shutter

szczukot:


gizmocuz:
Whats the change in CDomoticzHardwareBase::SendSwitch ?
This will affect a log more hardware..., if it is not working out for you, then call this function with a level that is working out?

szczukot:
I asked about this a few days ago: https://github.com/domoticz/domoticz/issues/778

szczukot:
And i see now, that ALL other modules use 15 instead 16.
For example:
https://github.com/domoticz/domoticz/blob/master/hardware/Sterbox.cpp#L144
https://github.com/domoticz/domoticz/blob/master/hardware/RFLinkBase.cpp#L250
https://github.com/domoticz/domoticz/blob/master/hardware/PhilipsHue.cpp#L460
https://github.com/domoticz/domoticz/blob/master/hardware/MySensorsBase.cpp#L1098
etc.

szczukot:
You changed this value a few months ago. But why ?

https://github.com/domoticz/domoticz/commit/bf7a641bfb80a7340a044b1bcefeb843b0744316#diff-4a5024c1ddeb20b684105485c68b8a3bR518

i found only one call this function (SendSwitch) where level <> 0
This call is from KMTronic. But level set only to 0 or 100.

szczukot:
And again :)
When level set in this method to 16 in next step it is reduced to 15 :)
https://github.com/domoticz/domoticz/blob/master/main/mainworker.cpp#L10293

gizmocuz:
Yes, again, this seems only for your hardware, otherwise we had heard this before ?

tomaszszczukobls:
Ok, i change my code but this is bug !
You change this line a few month ago only for one hadware (kmtronic). And other ? https://github.com/domoticz/domoticz/commit/bf7a641bfb80a7340a044b1bcefeb843b0744316#diff-4a5024c1ddeb20b684105485c68b8a3bR518

And even when you set level to 16,in next lines you reduce it to 15.

Take any device that sends level and set up it to max (100 or 255 or something like this). KMTronic too. In this line you calculated this value to 16 or more. 
In the next lines, you check this value and value from database:
https://github.com/domoticz/domoticz/commit/bf7a641bfb80a7340a044b1bcefeb843b0744316#diff-4a5024c1ddeb20b684105485c68b8a3bL546

Nex, you redyce this value to maks 15:
https://github.com/domoticz/domoticz/blob/master/main/mainworker.cpp#L10293

and save it to database.
In database max level is 15

And next step: hardware set level to 100, you calculated it to 16, and compare with last value (from database). But in database is 15. 15<>16 then you again send 16 to process, reduced it to 15, save to database 15 and over and over again

This is bug for ALL hardwares that they use it.

All hardware, that send a level > 0 (I checked in the code and gave some examples above) use own SendSwitch without this bug:
for example https://github.com/domoticz/domoticz/blob/master/hardware/MySensorsBase.cpp#L1098

Only one device uses your method with level > 0 - KMTronic.
Wy do you increased this value for this hardware ?

gizmocuz:
Okey, but, it is better to make separate pull requests
I dont know why the 16 is there, who did it or why... and when....
Your pull request implies a change for the BluBox... but it is not only that
KMTronix does not have a dimmer... it are relais.. i will look into this asap
But for the time being i would suggest to remove this from this pull request and make another (or wait) for the level

gizmocuz:
Thanks!

PR #786, opened 2016-08-10
Added TSL2561 illuminance I2C sensor

ldrolez:
Also removed a small ifndef in I2C.cpp that prevented testing on x86 platforms. 
Not sure with the line 'SendLuxSensor(0, 0, 255, lux, "Lux");' but seems to be ok...


gizmocuz:
Thank you very much for this great work !
Can i ask one massive question/suggestion ? Is it possible you can combine the 3 hardware types to 1 type ?
And on the hardware page a dropdown for the type ?
This because you can connect any i2c sensor you want, and this will end up in 100's of hardware types.

Also, do you know the project mysensors.org ? Its supports already a lot of sensor types, you can also connect it directly, and i think you can even compile it on a raspberry pi

ldrolez:
Yes, it would be better to reduce the sensors list. It will take me
more time since I'm new to Domoticz, but I'll try :)

Yes MySensors is a great project, but I do not see options to use the
PI I2C bus, that's why I wired a few sensors directly to my PI :)

ldrolez:
Or even better: I2C sensors auto detection! But that will be the 3rd step :)

gizmocuz:
;)

PR #789, opened 2016-08-12
Added TSL2561 illuminance I2C sensor

ldrolez:
Please forget my previous pull since it was done on my master branch.

### DIFF
--- a/History.txt
+++ b/History.txt
@@ -19,6 +19,7 @@ Version 3.xxxx (August xx 2016)
 - Implemented: RFLink: Displaying firmware version in the hardware setup
 - Implemented: RFXCom, Blinds T12 (Confexx)
 - Implemented: S0 Meter TCP
+- Changed: Local I2C sensors grouping (ldrolez)
 - Changed: ETH-8020, ignoring return message as it always returns (no more Success message)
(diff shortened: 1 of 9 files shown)


ldrolez:
I've added local I2C grouping also. All sensors with a description begining with "Raspberry I2C sensor" will no longer appear in the main combo box.

gizmocuz:
Thank you for this !!
One small request, i would expect this to run on other hardware as well (Cubieboard,ODroid,Pine64,... etc), as the libraries being used are from the linux kernel, and it is not using raspberry dedicated stuff
If thats correct, Maybe rename 'raspberry' to 'Local' ? Or 'System' ? or another name ?

ldrolez:
Yes, sure, maybe Local is clearer?

gizmocuz:
Sounds perfect ;)

ldrolez:
also how do you handle i18n? :)

gizmocuz:
Thanks!

gizmocuz:
Via the site transifex, english file is the main file, best to PM me changes in there

gizmocuz:
Seems there is a possible compile problem
https://github.com/domoticz/domoticz/issues/791

PR #794, opened 2016-08-15
[BleBox] refactor and new method useful for many hardware

szczukot:


hardware/BleBox.cpp
```diff
@@ -586,14 +569,9 @@ namespace http {
 			}
 
 			std::string hwid = request::findValue(&req, "idx");
-			if (hwid == "")
-				return;
-			int iHardwareID = atoi(hwid.c_str());
-			CDomoticzHardwareBase *pBaseHardware = m_mainworker.GetHardware(iHardwareID);
+			CDomoticzHardwareBase *pBaseHardware = GetHardwareBase(hwid, HTYPE_BleBox);
```
gizmocuz: Plz use the normal mainworker gethardware function (with the ID), then after check if the TYPE is correct
You can then remove the webserver changes (which also gave build issues)

tomaszszczukobls: I added this method also for other devices.
Many of them have this same code like:
            if (hwid == "")
                return;
            int iHardwareID = atoi(hwid.c_str());
            CDomoticzHardwareBase *pHardware = m_mainworker.GetHardware(iHardwareID);
            if (pHardware == NULL)
                return;
            if (pHardware->HwdType != HTYPE_Kodi)
                return;
This schema exists about 80 times in code!
Is it better to do a universal function and use it in all devices?

gizmocuz: Maybe then extend the m_mainworker.GetHardware function with an 'Optional' Type parameter then
This code should not be in the webserver class
And it is also not entirely true, as on many places in the code we check for multiple types
But maybe keep the code refactoring for later, and just use the normal method in the blebox class

tomaszszczukobls: On first version i added this method to mainworker, but later i saw that all occurence exists in webserver class.
But ok, i will add new method to mainworker.

main/mainworker.h
```diff
@@ -40,6 +40,7 @@ class MainWorker
 	int FindDomoticzHardware(int HwdId);
 	int FindDomoticzHardwareByType(const _eHardwareTypes HWType);
 	CDomoticzHardwareBase* GetHardware(int HwdId);
+	CDomoticzHardwareBase* GetHardwareBase(const std::string &HwId, const int Type);
```
gizmocuz: Can you rename the function/parameters please, this is more consistent with what we have

CDomoticzHardwareBase\* GetHardwareByIDType(const int HwdId, const _eHardwareTypes HWType);

gizmocuz:
Thanks!

PR #797, opened 2016-08-16
setup: Tweak Pushover labels, add direct link to clone app

jcs:
(I am the author of Pushover)

Pushover calls an "API key" an "API token", and now that it is listed as a public app on https://pushover.net/apps, users can easily clone it at https://pushover.net/apps/clone/domoticz to get the same Domoticz icon and app name.

### DIFF
--- a/www/views/setup.html
+++ b/www/views/setup.html
@@ -394,22 +394,22 @@
 							</div>
 							<div class="row-fluid">
 								<div class="span6">
-									<h2><span data-i18n="Pushover (Android/iOs)"></span>:</h2>
+									<h2><span data-i18n="Pushover (Android/iOS)"></span>:</h2>
 									<table class="display" id="pushovertable" border="0" cellpadding="0" cellspacing="0">
 									<tr>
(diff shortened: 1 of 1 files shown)


jcs:
I didn't notice that there are i18n files that need to be updated to make label changes because they are gzipped (so grep didn't find them).  How are these files updated?

gizmocuz:
I would not change them...
This because it is in the system for a long time now 
And with 30 translations... Every language has to be changed too
If you agree, please undo these..., else I will add them 
We use transifex for translation

gizmocuz:
When you are not signed in (default) and go to the clone url, you are forced to create an account
I thought it was nicer to go to the main website, so users might read what pushover is
If you are logged in (not default), and go the the clone url, then you are presented with a nice page

jcs:
Well I would have preferred two links, one on "Pushover" to go to the main site, and one on "API Token" to go to the clone page, but I didn't want to change the look of the page too much.

gizmocuz:
I see no problem with added a second url underneath it

PR #798, opened 2016-08-17
Correct handling of corrupted P1 messages

gordonb3:
This patch fixes random crashing of Domoticz due to oversized values in corrupted P1 messages.

Includes some code cleaning and changes to incorrect comments

### DIFF
--- a/hardware/P1MeterBase.cpp
+++ b/hardware/P1MeterBase.cpp
@@ -3,6 +3,8 @@
 #include "hardwaretypes.h"
 #include "../main/localtime_r.h"
 
+#include "Logger.h"
+
 typedef enum { 
 	ID=0, 
 	STD, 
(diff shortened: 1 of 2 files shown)


hardware/P1MeterBase.cpp
```diff
@@ -13,15 +15,15 @@ typedef enum {
 
 #define P1_SMID			"/" // Smart Meter ID. Used to detect start of telegram.
 #define P1PU1			"1-0:1.8.1" // total power usage normal tariff
-#define P1PU2			"1-0:1.8.2" // total power usage low tariff
+#define P1PU2			"1-0:1.8.2" // total power usage day/high tariff
```
gizmocuz: It is not day/high tariff, but actually normal/low or low/high (see the weekend)

The documentation is not very clear about this, its called tariff 1 / tariff 2, later normal/high ... ohh and also low and normal/high

http://files.domoticaforum.eu/uploads/Smartmetering/DSMR%20v4.0%20final%20P1.pdf

maybe we should follow chapter 6.1 Electricity data ?
or the labeling in paragraph 6

6 DATA OBJECTS
Currently two tariffs (Rate 1 low tariff and Rate 2 normal/high tariff)
are defined

Or Rate 1 and Rate 2 ?

gordonb3:
Having looked at some example messages I found that if you are on a single tariff scheme usage is registered on the `1-0:1.8.1` counter. `Normal` as defined in the original code therefore appears the correct choice of name. This does complicate naming of the `1-0:1.8.2` counter, but `Low tariff` is definitely wrong. At 11:02 AM on Wednesday (which is high tariff) my current selected counter is `#2` as you can see in this message snippet 

> 1-0:1.8.1(001660.375_kWh)
> 1-0:1.8.2(001690.779_kWh)
> 1-0:2.8.1(000000.000_kWh)
> 1-0:2.8.2(000000.000_kWh)
> 0-0:96.14.0(0002)

I suppose you could rename `Normal` to `Low/night/single` which would be more descriptive. The labelling you quoted appears to contradict the sample messages I found though and would indicate that single tariff would be on the `1-0:1.8.2` counter. Can that be? Another thing I noticed is that my meter produces different sized fields than defined. Which is not really an issue but does indicate that different generation meters can report numbers differently.

In essence: the standard does not appear to be very consistent. It also would not be unthinkable that Tariff 1 and Tariff 2 might be low/high in one country and high/low in another. If that happens to be the case, then I guess naming them `Tariff 1|2` would be the better choice here.

gizmocuz:
True, Please change it to Tariff 1 / 2, i also use this in the log/graph (details) , like Counter T1, Counter T2

gordonb3:
Done

gizmocuz:
Thanks... one small problem our build system has detected

.\hardware\P1MeterBase.cpp(6): fatal error C1083: Cannot open include file: 'Logger.h': No such file or directory

It should be

#include "../main/Logger.h"

Or also remark if not used... saves a few cpu cycles when compiling

gordonb3:
Strange. It compiles just fine in my environment (Gentoo Linux). I needed the logger because I had inserted additional log lines at various points to see which condition was triggered. I removed those in the final code, so yes the Logger.h include serves no purpose any more.

I've removed the line, so the code header is now exactly the same as it was before.

gizmocuz:
Thanks!

PR #801, opened 2016-08-18
Use relative path for OZWCP URL + I2C sensors position fix

ldrolez:
Better if Domoticz is behind a reverse proxy and has another base URL.


gizmocuz:
Thanks!

PR #810, opened 2016-08-22
Http headers enhancements

sbouchex:
- Added support for HTTP headers when using the POST method (HTTP push + HTTP notification)
- Fixed default value issue for HTTP notification content type


gizmocuz:
Thanks

PR #815, opened 2016-08-24
Goodwe warning fixes

llagendijk:
Fixed the warnings that you asked me to fix in the Goodwe adapter. Had to break deviceID calculation unfortunately.

### DIFF
--- a/hardware/GoodweAPI.cpp
+++ b/hardware/GoodweAPI.cpp
@@ -443,9 +443,10 @@ void GoodweAPI::ParseDevice(Json::Value device, std::string sStationId, std::str
 	int NodeID = hash(sStationId);
 	int ChildID = hash(sDeviceSerial);
 
-	// reserve childid below 10 for the station
-	if (ChildID < 10)
-		ChildID =+ 10;
+	// reserve childIDs  0 - 10 for the station
+	if ((ChildID >= 0) && (ChildID <= 10)) {
+		ChildID = ChildID + 10;
+	}
 
 	SendKwhMeter(NodeID, ChildID, 255, currentPowerW, totalEnergyKWh, sStationName + " " + sDeviceSerial + " Return");
 	SendTextSensor(NodeID, ChildID + 1 , 255, sStatus, sStationName + " " + sDeviceSerial + " status");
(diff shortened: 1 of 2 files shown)


hardware/GoodweAPI.cpp
```diff
@@ -166,11 +166,12 @@ int GoodweAPI::hash(std::string str)
...
 	int i = 0;
-	int c;
+	int c = str[i++];
 	
-	while(c = str[i++])
+	while(c)
 	{
 		hash = ((hash << 5) + hash) + c;
+		c = str[i++];
 	}
```
gizmocuz, line 176: I still find this a bit strange way todo this
You know the length of the string ( str.size() )
You could use

while (i<str.size())

or even

for (i=0; i<str.size(); i++)

Both look a bit better

hardware/GoodweAPI.cpp
```diff
@@ -442,9 +443,10 @@ void GoodweAPI::ParseDevice(Json::Value device, std::string sStationId, std::str
 	int NodeID = hash(sStationId);
 	int ChildID = hash(sDeviceSerial);
 
-	// reserve childid below 10 for the station
-	if (ChildID < 10)
-		ChildID =+ 10;
+	// reserve childIDs  0 - 10 for the station
+	if ((ChildID >= 0) && (ChildID <= 10)) {
+		ChildID = ChildID + 10;
+	}
```
gizmocuz, line 451: Could there be negative child id's ? maybe the hash should return an uint32_t ?

hardware/GoodweAPI.cpp
llagendijk: Good comment. I am still too much C-oriented...
Will include this

hardware/GoodweAPI.cpp
```diff
@@ -168,11 +169,12 @@ int GoodweAPI::hash(std::string str)
 	int i = 0;
```
gizmocuz: could you make it size_t instead of int ;)

llagendijk: I should have known this but missed it :-(

gizmocuz:
Thanks !

PR #819, opened 2016-08-24
More useful to have the average Humidity and Barometer value in history

ldrolez:
I think the same also for Luminance and other sensors, but it would require a database schema change?


gaeljw:
I think the server should send both min/max and average (when it makes sense) and the GUI should do whatever it wants with these data.
In this particular case, for humidity and barometer I agree average seems better than max in history.

gizmocuz:
Thanks!, Maybe in the feature we should store min/max/avr humidity

Tommit1:
+1 to please my tenants

PR #820, opened 2016-08-24
Corrected Prerequisites for Ubuntu

ageir:
Added libssl-dev to prerequisites for Ubuntu. If not installed cmake results in an error not finding openssl.
Tested on Ubuntu 16.04

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -164,7 +164,7 @@ make
 ### Ubuntu / Raspberry Pi (wheezy)
 ```
 sudo apt-get install build-essential -y
-sudo apt-get install cmake libboost-dev libboost-thread-dev libboost-system-dev libsqlite3-dev subversion curl libcurl4-openssl-dev libusb-dev libudev-dev zlib1g-dev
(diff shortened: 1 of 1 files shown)


gizmocuz:
I am not sure that this file is up-to-date....
Is it possible to check against our wiki ? (raspberry pi install and/or ubuntu)

http://www.domoticz.com/wiki/Raspberry_Pi

http://www.domoticz.com/wiki/Linux

ageir:
Not sure what you mean. Since this is usually the first file you check when installing/compiling a program it should be up to date. The libssl-dev package is mentioned in the wiki as a requirement but missing in this file.

gizmocuz:
Thanks, yes, that's true, but the file is not up 2 date, maybe it should point to the wiki pages instead

ageir:
I can update the whole file if you want so that it is up to date with the current wiki.
And refer to the wiki in some places so that the file doesn't become to big. I believe that only referring to the wiki is a bad idea. For example you might have an old version you want to compile and then the wiki has been updated to the latest version and the installation instructions are incorrect, or you might not have internet access at the moment.

gizmocuz:
That would be great !

PR #823, opened 2016-08-25
No more needed for sysinfo lib on FreeBSD

xbeaudouin:
Since I have added linuxprocfs support for FreeBSD, we no longer need the dependency against libsysinfo on FreeBSD.
This pull request clean this unnecessary dependency.


gizmocuz:
Thanks for the cleanup !!

Issue #748, opened 2016-07-21
Filter values for openzwave sensors


sbouchex:
I do have   "Qubino ZMNHAA2 Flush 1" Relay zwave device but the temperature pin is not connected to a sensor. So the returned value is -999.9 °C which does not make sense.
However, this valus is stored anyway.

I have the same "Qubino ZMNHBA2 Flush 2 Relays" where the power consumption is reported as "-214748364.800 kWh" :-) (probably because this relay has never been connected to the power)

If one day, I add a temperature sensor, the average value will be wrong and I will have to delete/add the sensor again.

Why not adding some filtering (hardcoded or not) to avoid emitting dummy values into the rest of the server ?

gizmocuz:
It would have been better if openzwave or the quibino itself did this...
But yes... Your right... Could you make the request?
I wonder what the hex value is of the meter, or the unit... Maybe we can check on this...

sbouchex:
I can work in it. I have contacted Qubino to have more details

sbouchex:
still waiting for Qubino response (Summer period...)

sbouchex:
OK, got an answer from Qubino:
- For the -999.9°C : That's an hardcoded value to say that the temperature sensor is not connected.
- For the 214748364..., it was a device issue.

I'll propose to filter out the values based on the vendor id and eventually, the product id as well, the sensor value will not be emitted if the value is filtered out.

gizmocuz:
Thats OK, could you make a patch that when a temp value of -999.99 is received we do not send it ?
(just return in the update value function)

sbouchex:
yes, that's the plan and I will try to make a generic function in order to filter other potential values in the future

PR #824, opened 2016-08-25
Handle missing temperature sensor on some Qubino devices

sbouchex:
Handle missing temperature sensor on some Qubino devices based on sigma 300 zwave chip

If the sensor value is -999.9, the sensor value is not stored.

Related to #748


gizmocuz:
Thanks!
I changed it a bit with this commit:
https://github.com/domoticz/domoticz/commit/0ac106d3fe82c48ada2361ce86eb5cf85387979a

Because it should be done for all temp sensors (this is the same as being done for other hardware in domoticz)

PR #827, opened 2016-08-26
Also get the average Baro/Humidity for the current day

ldrolez:
The last commit only did it when new data was feeded to the _Calendar table.


gizmocuz:
Thanks!

gizmocuz:
Thanks!

PR #828, opened 2016-08-26
Added new notification "when" operators

sbouchex:
Added "greater or equal", "below or equal" and "equal" operators for the notifcations


gizmocuz:
Where do you use this ? (that you need <= xx, if < x+1 is not the same

sbouchex:
The objective is just to add equality to the current tests.

You may want to trigger an alarm when temp is >= xxx. I agree that you may use > xxx-1 but that's less readable . IMHO

gizmocuz:
I dont think it is more readable/easier if you provide the novice user with a lot of options...?

sbouchex:
ok for >= and <=  but = may be usefull

gizmocuz:
Yes, but for what type of sensor ?

sbouchex:
Temperature, Humidity and Ampere

gizmocuz:
You will never get an exact temperature, not Humidity , nor Ampere, for those you Need to use the above/below for sure

sbouchex:
If we exclude all environmental sensors, we have percentage, but I agree it's not very common.

Other operator idea : "Latched Greater" : The notification is emitted when the value goes above the threshold but the notification will not be emitted again if the value is above again. It will only emitted if the value is below once, then goes above again.

Same for "Latched Below"

gizmocuz:
Nice when the temperature fluctuates between 1 degree ...
You can set a notification interval in the settings
Sorry, closing this issue, thanks for your effort !

PR #831, opened 2016-08-28
Fix ambiguous call to abs

ffontaine:
Add cmath header to be able to call abs with a float otherwise call to
abs(temperature*10.0f) is ambiguous as abs from cstdlib accepts either
an int, a long int or a long long int.

### DIFF
--- a/hardware/1Wire.cpp
+++ b/hardware/1Wire.cpp
@@ -14,6 +14,7 @@
 #include "../main/mainworker.h"
 #include "../main/SQLHelper.h"
 
+#include <cmath>
 #include <fcntl.h>
 #include <sys/stat.h>
 #include <sys/types.h>


gizmocuz:
Thanks!

gizmocuz:
This is done on many places in the code actually (by my doing...)
It seems to have no problems, as we do receive correct values with decimals ?

will std::abs be a replacement ?

ffontaine:
You're right, perhaps std::abs would be the best solution. I'm trying to resolve the issue on ARC target: http://autobuild.buildroot.net/results/0ed/0ed9fa9585ec91db3556d6f5b4ca8d99a5eb2144/build-end.log.
The inclusion of cmath does not seem enough. I will try std::abs and tell you informed. An other solution would be to use fabs.

gizmocuz:
Thanks, both sound good. As long as we support the old c++ compilers ;)

ffontaine:
OK, std::abs fix the issue. You can revert this commit as cmath does not seem to be needed. I will send a new pull request.

gizmocuz:
Could you make a new pull request and change all abs( calls that are using floats ?

PR #833, opened 2016-08-29
Fix ambiguous call to abs

ffontaine:
Prefix all calls to abs by "std::" to be able to call abs with a float
otherwise call to abs(float) is ambiguous as the C abs function that is
commonly available for legacy reasons only handle int.


gizmocuz:
Many thanks !

PR #834, opened 2016-08-29
On FreeBSD backtrace(3) is handled on libexecinfo library.

xbeaudouin:
The backtrace(3) call is handled by the libexecinfo. Added the necessary to add this lib during the linking.
Removed an old variable I forgot to remove when removing libsysinfo dependency
on FreeBSD.


gizmocuz:
Thanks!

PR #835, opened 2016-08-29
Remove a warning about extraneous parentheses

xbeaudouin:
Some parentheses are extraneous => removed them (and keep clang happy)


gizmocuz:
Thanks, aldo i find it strange it is complaining ;)

PR #836, opened 2016-08-29
P1 meter message validation

gordonb3:
Does not yet include the CRC check, but just to let you know what I'm working on. When finished I can combine the individual commits into a single big one if you like (create a new pull request).

The first three commits are about an issue I discovered with my previous commit. With `P1MeterSerial` this is currently not an issue, but it will be in combination with commit  df8ea27. And it may already be an issue with `P1MeterTCP`, which I think does pass full messages into `P1MeterBase`. The trouble is that if a validation failure occurs of the type "value is not a number" it will keep producing this failure with every subsequent  message and it will do this on the message header (`ID`).

Also, I found that the test in fact was wrong as it would only fail if the first character was non-numerical. This is corrected in commit  2e1ec54 and it will now require the full length of the value to be numeric.

The fourth commit prepares for performing the CRC check. Since `m_buffer[1028]` was quite large already I decided to use that variable to assemble the complete message and moved the existing references to a new variable `l_buffer[128]` (we're building a single **l**ine to run through `MatchLine()`, not a **m**essage).

### DIFF
--- a/hardware/P1MeterBase.cpp
+++ b/hardware/P1MeterBase.cpp
@@ -2,6 +2,7 @@
 #include "P1MeterBase.h"
 #include "hardwaretypes.h"
 #include "../main/localtime_r.h"
+#include "../main/Logger.h"
 
 #define CRC16_ARC	0x8005
 #define CRC16_ARC_REFL	0xA001
@@ -205,12 +206,14 @@ bool P1MeterBase::MatchLine()
 			if (ePos==std::string::npos)
 			{
 				// invalid message: value not delimited
+				_log.Log(LOG_STATUS,"P1: dismiss incoming - value is not delimited in line \"%s\"",l_buffer);
 				return false;
 			}
 
 			if (ePos>19)
 			{
 				// invalid message: line too long
+				_log.Log(LOG_STATUS,"P1: dismiss incoming - value in line \"%s\" is oversized",l_buffer);
 				return false;
 			}
 
@@ -270,6 +273,7 @@ bool P1MeterBase::MatchLine()
 
 			if (ePos>0 && ((validate - value) != ePos)) {
 				// invalid message: value is not a number
+				_log.Log(LOG_STATUS,"P1: dismiss incoming - value in line \"%s\" is not a number", l_buffer);
 				return false;
 			}
 
@@ -305,12 +309,12 @@ bool P1MeterBase::CheckCRC()
 	strncpy(crc_str, (const char*)&l_buffer+1, 4);
 	crc_str[4]=0;
 	uint16_t m_crc16=strtol(crc_str,NULL,16);
-	int m_size=m_bufferpos;
 
 	// calculate CRC
 	const unsigned char* c_buffer=m_buffer;
 	uint8_t i;
 	uint16_t crc=0;
+	int m_size=m_bufferpos;
 	while (m_size--) {
 		crc = crc ^ *c_buffer++;
 		for (i=0;i<8;i++){
@@ -320,6 +324,9 @@ bool P1MeterBase::CheckCRC()
 				crc = crc >> 1;
 			}
 		}
+	}
+	if (crc != m_crc16){
+		_log.Log(LOG_STATUS,"P1: dismiss incoming - CRC failed");
 	}
 	return (crc==m_crc16);
 }
--- a/hardware/P1MeterBase.cpp
+++ b/hardware/P1MeterBase.cpp
@@ -82,12 +82,14 @@ void P1MeterBase::Init()
 	m_linecount=0;
 	m_exclmarkfound=0;
 	m_bufferpos=0;
+	l_bufferpos=0;
 	m_lastgasusage=0;
 	m_lastelectrausage=0;
 	m_lastSharedSendElectra=0;
 	m_lastSharedSendGas=0;
 
 	memset(&m_buffer,0,sizeof(m_buffer));
+	memset(&l_buffer,0,sizeof(l_buffer));
 
 	memset(&m_p1power,0,sizeof(m_p1power));
 	memset(&m_p1gas,0,sizeof(m_p1gas));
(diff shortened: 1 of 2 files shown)


gizmocuz:
Looking good ! Except i think you have to check for 'sizeof(buffer)-1' (minus one)

gordonb3:
You mean where I did not put the minus one? The minus one is only required when the char array is going to be used as a string, reserving the last character to add the `\0`. Most of the locations where `sizeof()` is used in this part of the code is to determine when to discard the content. I have also experimented with an l_buffer size of 64, which results in a hit on the oversized line with every message (on field ID 1-0:99.97.0) and this returns the correct response.

As shown in a log snippet on the forum, I do have a running code with the CRC check already. Reason for posting it like this is that I've kind of made things complicated for myself by adding a huge amount of debug lines, so splitting it in multiple commits is mostly for me to keep track of all the individual changes in this stage. The code as committed so far will run correctly within the current code base.

gizmocuz:
Okey, let us know when to merge

gordonb3:
CRC check added. Currently verifying the code as submitted - no additional logging so I'm more or less running blind. Should still be able to see it dropping packages, but won't be able to tell for what reason.

gordonb3:
Okay. Got 74 misses in the last 12 hours. Not completely happy about it being all silent about it though. Any idea what log level I should assign for messages about packages being discarded? I currently have them all set to LOG_NORM.

gizmocuz:
By missed, you detected that they are corrupt or incomplete ?
I would use LOG_ERROR for this.

After we got all working, i suggest to buy a P1 cable from the smartmeter dashboard website (<25 euro, and you should have no problems)

gizmocuz:
Please tell me when it's ready to merge ;)

gordonb3:
Yes that is exactly what I meant by "missed". And frankly I think assigning a state of ERROR would be over-exaggerating the problem of a faulty bit in a single message that is not strictly required for proper operation and monitoring. "Warning" seems more appropriate, so I'll assign the discard messages a level of LOG_STATUS.

I must also note that most fails are in fact trapped by the CRC and from my early logging sessions I would conclude that 99%+ of these messages would in reality not hurt Domoticz. That's no reason to keep them though: a faulty message is a faulty message. It does however mean that on older meters (pre DSMR 4.0) the same amount of errors will likely only lead to about 5-10 misses a day, because in this case I can only trap obvious fails in fields that are actually used in Domoticz.

In any case: even with the seemingly high number of misses, this is only just an average of 1 in every 60 messages where any of 30 will allow a correct entry to end up in the database for creating graphs and summaries. Discarding those messages is not in issue - definitely not worth naming an error.

gizmocuz:
There should be no errors ... i dont have had a single error since years with the official supported p1 serial cable you can by at the smartmeter dashboard... 
I would like to know when there is an error ;) ... then i would create a normal log message with a 'warning' text maybe ?

gordonb3:
That you know of :wink:

The hardware did not cause any issues with me for several months as well. Problems just started happening about a month ago, but that doesn't mean that every message was clean before.

I currently have the warnings logged as LOG_STATUS, so they will show with loglevel=1

> 2016-08-31 06:36:08.842  P1: dismiss incoming - CRC failed
> 2016-08-31 06:44:49.386  P1: dismiss incoming - value in line "1-0:1.8.2(001753�385_kWh)" is not a number
> 2016-08-31 07:00:11.306  P1: dismiss incoming - value in line "1-0:1.7.0(00.30M��U�H�1-0:2.7.0(00.000_kW)" is oversized
> Sorry about the line breaks - there's a maximum width to the items on this page as well.

CRC failures would normally pass unnoticed. They can be one or more faulty bits in fields that are not used, but they could also make an `8` out of a `0`  (0x30 vs. 0x38 --- 00110000 vs. 00111000) which might show up as strange peak values.

In essence the same applies to the "not a number" failure. The failing line shown here is a perfect example of one that you would normally not notice.

The third line is the one that was causing all the mayhem prior to the earlier submitted patch to P1MeterBase. During the course of last night I've had two of those.

gordonb3:
Ready for merge

gizmocuz:
Thanks !!

gizmocuz:
Small compile issue, is it possible to solve this ?

..\hardware\P1MeterBase.cpp(311): warning C4244: 'initializing': conversion from 'long' to 'uint16_t', possible loss of data

gordonb3:
Hmmm... g++ does not complain about it. I've added an explicit cast (and changed `strtol()` to the unsigned version) that should solve it for your (Windows?) compiler.

gordonb3:
Oops: that one doesn't show here any more after the merge

Commit: https://github.com/gordonb3/domoticz/commit/7325608b015113d1b182bf88c594f08ee07018dd

gizmocuz:
Huston... could you maybe check this post ?
http://domoticz.com/forum/viewtopic.php?f=6&t=13167

gordonb3:
Huston?

Not completely unexpected. Didn't foresee though that there might be people who have practically no valid messages at all. Did get a chance to verify some of their messages and with at least one of the two sites the discards proved to be correct (see topic linked). That's not the answer he is looking for though.

In essence: they feel it should be possible to disable the CRC. I guess that should not be too much of an issue, but I could use some assistance if I'm to implement it.

Gordon

gizmocuz:
https://en.wiktionary.org/wiki/Houston,_we_have_a_problem _g_

Okey, let me try it here for a few days, if i do not have this error, then the CRC is working, which is a great thing.
Do they also make use of another cable then bought by the smartmeter dashboard website ?

gizmocuz:
So far, arround 100 messages received from my P1 cable, not a single problem

gordonb3:
From the messages posted by `mvveelen` I'd say it can be concluded that the CRC method is good. When I insert the obvious missing characters in his messages, the CRC passes.

The only real difference is that in his case non printable characters are omitted and I get a `�` character. It's impossible for me to tell on which level this difference is initiated. I also can't really comment on the second committer `Cletus` as the messages he posted all passed the CRC. Both did submit log snippets that shows that they are receiving both good and bad messages, it's just that in their case the majority is bad.

I have no insight into what hardware they are using for reading the smart meter.

PR #840, opened 2016-09-01
Show ranges for Lux sensors in month/year views

ldrolez:
Btw it was really difficult to merge my changes in this file because of the gzip compression.
We should really commit a non compressed version of the file :) !


gizmocuz:
extract the gzip in the same folder, you will end up with a file without .gz (otherwise remove the .gz)
Do your changes, test the code, when it's ok, compress to gzip, remove the original file without the gzip
commit
Very easy... but i agree, but this needs more work, as in the final beta's (binaries) i want them all compressed
Lets see what the weekend can do ;), it's on my todo list for more then a few months now

Thanks for the patch !

gizmocuz:
Hmmm the avg lux was not yet provided by the c++ code, working on this now...
Maybe you missed to commit a file... did you test this ?

gizmocuz:
reverting this patch, making it like the other types (like the percentage)

ldrolez:
Yes I was devolved by the .js.gz merge :) Should I resubmit the complete patch?

Also, I know that I simply have to unzip/zip the file, but the real problem is that GIT cannot merge binary files, so if I branch, and then try to merge your changes, git stops working and an awful manual merge session is needed...

gizmocuz:
Yep, thats why we need to solve this, and thats why we have weekends that hopefully brings some time, as i want the binary archives to contain .gz files
Nope, i already submitted a patch, it is available now

ldrolez:
I've seen a PR for this, it wasn't merged?

gizmocuz:
Yes, it already in the beta

ldrolez:
The code for gzipping js files?

gizmocuz:
No, the lux avr

Issue #813, opened 2016-08-24
Add protocol option to camera


benjR:
Hi,

Would it be possible to add the protocol choice when adding a camera ? It's forced to http by default.
I have one that is using ftp, https, I know that some others are using rmtp and so on...

Thanks

sbouchex:
@gizmocuz ok to do it ? I can work on it

I will add a popup menu with HTTP, HTTPS for now, will see later for other type of camera

gizmocuz:
Dont know if that will work... ftp ? also rtsp/rtmp, i do not think mailing a screenshot will work ? (taking a screenshot)
Most camera's (90+ percent) have a url for a image snapshot, this is what we use at the moment

sbouchex:
I agree with you but the @benjR's point was just to have the ability to use HTTPS but CCameraHandler::GetCameraURL current implementation is forcing HTTP (it's hardcoded) and I was proposing to add a field to specify the protocol.

gizmocuz:
If it is an internal camera, then HTTP is not supported?

sbouchex:
Why not offering both ?

gizmocuz:
Can't your camera offer HTTP ?

sbouchex:
I have one where it's either HTTP or HTTPS... and noweverdays, HTTPS is mostly recommanded

gizmocuz:
not for internal use, so you can handle HTTP, also is less cpu overhead

sbouchex:
Domoticz is not only for internal use (you don't know what users are doing with it) and sometimes SSL is just mandatory (knowing that it's more CPU intensive). My PR was to be more open, if you don't like it (throw it away)

gizmocuz:
I hope you can agree that the camera interface in domoticz, will still connect only internally to your camera, even if you connect to domoticz via https
So there is really no need for HTTPS camera support... there is however support needed for AVC/H264 camera's, that mostly work via rtsp, but thats another PR, and we should find a method of creating a snapshot if these as well

sbouchex:
Do what you want

gizmocuz:
Sure, it's a pull 'request', but as you can access your camera internally by http as well, there is no issue, or reason, to rebuild things, giving another user option.
It's getting more and more complicated each day by all the wishes.

meaninglessvanity:
A workaround for photo hosting that domoticz doesn't support would be to mirror the camera's feed in http on the local server.

PR #841, opened 2016-09-01
Added protocols (HTTP & HTTPS) to cameras management

sbouchex:
Added HTTP & HTTPS protocols to cameras (see #813)


gizmocuz:
As discussed, not needed

PR #842, opened 2016-09-02
P1 meter CRC validation on/off switch

gordonb3:
I can't say I totally agree with this, but here's a switch for the hardware page to disable the CRC validation.


gizmocuz:
Thanks!

PR #845, opened 2016-09-04
Change Level for lighting type

szczukot:
Regarding https://github.com/domoticz/domoticz/issues/778

KMTronic always send 0 as level
        bool bIsOn = false;
        SendSwitchIfNotExists(ii + 1, 1, 255, bIsOn, (bIsOn) ? 100 : 0, sstr.str());
ZWave don't use SendSwitch.

I didn't find hardware that using level <> 0 and type lighting2. I'm first :) And this is bug on this moment.

This is proposal for resolve bug and expand possibilities for futer hardware. Now new hardware may set level to max 255 (no limit to 15).


szczukot:
Hi

Did You check this idea and code ?

regards
Fantom

gizmocuz:
Lighting 2 does not have 255, but 15

gizmocuz:
You should use the type general switch which does have 255

szczukot:
If Lighting 2 have max value 15, why don't You fix this bug ?? https://github.com/domoticz/domoticz/blob/master/hardware/DomoticzHardware.cpp#L519

Fantom
ps. I can't using general switch, because i must code/decode ip adress and store it on deviceid column.

gizmocuz:
Because i dont have a problem with it... only you
Anyway, enough of this, use the general switch ... or not
this is not a bug

szczukot:
You dont't have problem because You don't use it yet - i wrote above about this.
But ok - i changed this in my code.

gizmocuz:
Yep as said, thanks for implementing your own implementation in your class

Issue #276, opened 2015-12-29
Update Mysensor TempHum Error


Spugna85:
Hi,
 I have a Mysensor physical node (Arduino) with 6 sensor DHT22.
So I have 12 values, 6 temperature and 6 humidity, so listed:
-Sensor0 (ID=0501): H0 (IDCHILD = 0), T0 (IDCHILD  = 1)
-Sensor1 (ID=0503): H1 (IDCHILD  = 2), T1 (IDCHILD  = 3)
-Sensor2 (ID=0505): H2 (IDCHILD  = 4), T2 (IDCHILD  = 5)
-Sensor3 (ID=0507): H3 (IDCHILD  = 6), T3 (IDCHILD  = 7)
-Sensor4 (ID=0509): H4 (IDCHILD  = 8), T4 (IDCHILD  = 9)
-Sensor5 (ID=050B): H5 (IDCHILD  = 10), T5 (IDCHILD  = 11)

Domoticz create 6 sensor (Model:Temp + Humidity, subtype:WTGR800).

Each update temperature (T0 to T5) correctly updates the sensor device in the table.
Example:
T0 updates the temperature of the first sensor (ID: 0501)
T1 updates the temperature of the second sensor (ID: 0503)
and so on...

While each update of humidity (H0 to H5) always updates the first sensor for the node.
Example:
H0 refresh the humidity of the first sensor (ID: 0501)
H1 refresh the humidity of the first sensor (ID: 0501)
and so on...

I think that the problem is that Domoticz use "IDCHILD" temperature as the sensor address for Model "Temp + Humidity" and uses the default 0 to update the humidity.

So, if I have more sensors, any humidity variation on one of them, generates an incorrect update of the first sensor and a missing update of another sensor.

What do you think?

gizmocuz:
Could you set the temp+hum both on the same child id, for all sensors?

DTH1: (child id 0 for both temp+hum)
DTH2: (child id 1 for both temp+hum)
...
DTH5: (child id 4 for both temp+hum)

Spugna85:
Now I just tried this:
Sensor0 (ID=0500): H0 (IDCHILD = 0), T0 (IDCHILD = 0)
Sensor1 (ID=0501): H1 (IDCHILD = 1), T1 (IDCHILD = 1)
...
Sensor5 (ID=0505): H5 (IDCHILD = 5), T0 (IDCHILD = 5)

And yet I can see the same behavior: 
[...]

PR #847, opened 2016-09-04
MySensors groupID

BmdOnline:
Issue #276


gizmocuz:
Thanks for this great piece of work !

flopp999:
After I updated to 3.5609 it seems that most of my MySensors nodes with Temp and Humidity sensor is visible as two datas. Before it was combined Temp/Hum.
See more information and print screens here
http://www.domoticz.com/forum/viewtopic.php?t=13231

PR #848, opened 2016-09-05
fix MUSL build issue

ffontaine:
Without <linux/ioctl.h> include compiler cannot find _IOC_SIZEBITS
macro, when using MUSL C library

Fixes:
  http://autobuild.buildroot.net/results/62ec0d348153dff0efd4c1975a9198c17f01f1fa

### DIFF
--- a/hardware/PiFace.cpp
+++ b/hardware/PiFace.cpp
@@ -38,6 +38,7 @@
 #include <string.h>
 #include <string>
 #ifdef __arm__
+    #include <linux/ioctl.h>
     #include <linux/types.h>
     #include <linux/spi/spidev.h>
     #include <unistd.h>


gizmocuz:
Thank you !, A shame it works on some platforms, and on some not... _g_

ffontaine:
Thanks :-)

For my information, do you plan to make a new domoticz release?

Indeed, it is difficult to apply our patches on domoticz as the source files in the domoticz tarballs have DOS line endings, which don't appear when fetching the source code from Git (https://git.buildroot.net/buildroot/commit/package/domoticz?id=397962ce1737ba0f920ffc1d3a3ea2bc08b28270).

gizmocuz:
About the dos line endings, they should be taken care of by the .gitattributes file correct ?
(Native line endings)

There is a new version of domoticz almost every day, but i will try to make one in the first week of october

ffontaine:
You're right, .gitattributes takes care of the issue when domoticz is retrieve through git clone. However, buildroot uses the link provided by github (https://github.com/domoticz/domoticz/archive/3.4834.tar.gz) to download domoticz through a simple wget. In this case, the DOS line endings are kept in the source files and the git patches can't be applied without adding DOS line endings everywhere.

Nevertheless, this is great news for the update in october. domoticz has been officially integrated in the latest buildroot release (2016.08), perhaps you could add a link to this buildsystem at https://www.domoticz.com/wiki/Domoticz_Wiki_Manual?

gizmocuz:
Could you not just get the latest github code like we already do with travis and appveyor ?

gizmocuz:
You can differentiate maybe the stable and latest commit (beta)
So stable tag 3.4834 has commit revision xxxx

PR #849, opened 2016-09-05
P1 meter possible incorrect discards due to middleware

gordonb3:
I found a condition (that should not happen) where messages can be silently dropped by P1MeterBase. If the middleware adds an empty line in front of the message, the counter value for the message buffer is never reset because the procedure expects the first character of a message to be `/`. This is a condition that can only occur with the LAN connector that pushes complete P1 messages into the message parser.

The code provided here adds some additional logging where messages would still be silently dropped and a fix to allow messages with empty lines in front of them to pass.

Gordon


gizmocuz:
Thanks

PR #851, opened 2016-09-07
Fix unsigned expression < 0 (that is always true)

xbeaudouin:
Another issues with unsigned and <= 0 :) as discussed on the forum.


gizmocuz:
Can you change it to

if (result.empty())

xbeaudouin:
Done with result.empty() :D

gizmocuz:
Thanks!

PR #853, opened 2016-09-07
Ignore Serial_ID_Button devices in 1wire

sbouchex:
Ignore Serial_ID_Button devices in 1wire as it does not contain usefull data and generate useless activity

### DIFF
--- a/hardware/1Wire/1WireByOWFS.cpp
+++ b/hardware/1Wire/1WireByOWFS.cpp
@@ -96,6 +96,13 @@ void C1WireByOWFS::GetDevices(const std::string &inDir, /*out*/std::vector<_t1Wi
 			{
 				_log.Log(LOG_STATUS, "1Wire (OWFS): Add device %s", device.filename.c_str());
 			}
+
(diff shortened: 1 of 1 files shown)


gizmocuz:
Shall this also be ignored ByKernel or ForWindows ?

sbouchex:
I would say yes as well (even if I can't test it)

gizmocuz:
then maybe it should be ignored in 1Wire.cpp

void C1Wire::BuildSensorList() {

or

void C1Wire::PollSensors()

or

void C1Wire::BuildSwitchList() {

or

void C1Wire::PollSwitches()

But as far as i can see, only these sensors are supported:

```
    case high_precision_digital_thermometer:
    case Thermachron:
    case Econo_Digital_Thermometer:
    case Temperature_memory:
    case programmable_resolution_digital_thermometer:
    case Temperature_IO:
    case Environmental_Monitors:
    case _4k_ram_with_counter:
    case quad_ad_converter:
    case smart_battery_monitor:
```

sbouchex:
problem looks different, new PR :-)

PR #854, opened 2016-09-08
Missing server_cert.pem install (used for https server).

xbeaudouin:
On make install the server_cert.pem is not installed, avoiding https server to start.


gizmocuz:
Good spot ! Thanks!

PR #855, opened 2016-09-08
Fixed device unicity issue on 1wire device due to 1wire linux bug

sbouchex:
For some seasons, on linux, /mnt/1wire contains _twice_ the same device. So C1WireByOWFS is listing 2 times the same device and it is polled twice as well:

root@domoseb:~# ls -la /mnt/1wire/
drwxrwxrwx 1 root root 4096 Sep  8 15:25 28.D7438B030000
drwxrwxrwx 1 root root 4096 Sep  8 15:25 28.D7438B030000
drwxrwxrwx 1 root root 4096 Sep  8 15:25 81.46B730000000
drwxrwxrwx 1 root root 4096 Sep  8 15:25 81.46B730000000

To solve the issue, devices & switched are now considered as unique.

### DIFF
--- a/hardware/1Wire.cpp
+++ b/hardware/1Wire.cpp
@@ -244,7 +244,7 @@ void C1Wire::BuildSensorList() {
 		case _4k_ram_with_counter:
 		case quad_ad_converter:
 		case smart_battery_monitor:
-			m_sensors.push_back(*device);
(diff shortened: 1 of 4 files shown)


gizmocuz:
Thanks!

PR #858, opened 2016-09-08
Fix for P1 hardware submitting arbitrary partial lines

gordonb3:
I received a P1 Wifi Gateway today from Romix and got a chance to play with it. Turns out that the LAN connector actually does not submit full messages to ParseData(), but it also does not (always) submit full lines like the serial connector. As a result there can be trailing characters after the exclamation mark, causing the CRC calculation to fail.

This patch stops the extra characters from being added to the message buffer.


gizmocuz:
Thanks!

PR #859, opened 2016-09-09
FIX: possible deadlock when closing app

sawik-PL:
deadlock occurs when clossing app:
- MainWorker::StopDomoticzHardware() locks m_devicemutex and waits for devices to stop
- eg. MQTT got message in meantime and waits for lock on m_devicemutex in MQTT::on_message -> MainWorker::UpdateDevice -> MainWorker::GetHardware

added another mutex which is locked on app closing and checked before trying to lock m_devicemutex

### DIFF
--- a/main/mainworker.cpp
+++ b/main/mainworker.cpp
@@ -255,14 +255,17 @@ void MainWorker::StartDomoticzHardware()
 
 void MainWorker::StopDomoticzHardware()
 {
-	boost::lock_guard<boost::mutex> l(m_devicemutex);
-	std::vector<CDomoticzHardwareBase*>::iterator itt;
-	for (itt=m_hardwaredevices.begin(); itt!=m_hardwaredevices.end(); ++itt)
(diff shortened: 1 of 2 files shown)


gizmocuz:
There should be no issue with waiting for MainWorker::GetHardware
If the hardware is not found (it was removed in the mean time), it returns NULL

I am not a big fan of mutexes.... now for instance in MainWorker::HeartbeatCheck, we lock 3 mutexes in a row

I would almost vote for an shared_pointer...

sawik-PL:
> There should be no issue with waiting for MainWorker::GetHardware
> If the hardware is not found (it was removed in the mean time), it returns NULL

but there is, really! The problem is not that the hardware is not found, but that 2 threads wait for each other (MainWorker waits for MQTT to stop, MQTT waits for MainWorker to release m_devicemutex).

gizmocuz:
Can't MQTT be fixed?

PR #875, opened 2016-09-17
Basis support for HEOS by DENON

Sneezydevil:
This adds basic support for HEOS by DENON, a multiroom audio system (https://usa.denon.com/us/heos).

The class works alot like the Logitech Media Server class.

I used the ASyncTCP class as a base since my own code was alot messier and less stable.

With HEOS you talk to one Device and it will show the data of all the devices in the netwerk. So you start by adding one of your devices (Hardware page) by ip or hostname and then all will be detected and added to the Devices page.


gizmocuz:
Thank you a lot for this great piece of work !

PR #878, opened 2016-09-18
Fix fo is_number() method

szczukot:
Sometimes owfs fill files with response by 0x00
But this is still properly number (teminated by white spaces)


gizmocuz:
Thanks !

PR #882, opened 2016-09-21
HEOS: Added remote and event handling

Sneezydevil:
Working commands on the remote: Play, Stop, Pauze, Prev, Next, Mute, Volume Up and Volume Down.

Not sure if the app-cache file needs to be updated.


gizmocuz:
When you make changes to the www part, then you need to update the appcache version as well
Is it possible to use the already existing kodi remote ?

Sneezydevil:
Regarding the appcache file: Sure I can, I wasn't sure it was needed (Just seemed to work on my development environment) that's why I asked in my first comment.

Regarding the remote: I was not sure what to do, before creating this one, I looked at the kodi and the Logitech remote. Both are a lot alike and both have buttons I will never be able to support. So it was the choice between having buttons with no purpose or adding a remote without the buttons. If you would rather have the first option let me know.

gizmocuz:
Can we disable some buttons? If not, i'm happy with a 3rd remote... ;)

Sneezydevil:
I would prefer a 3rd remote for now then I can add some HEOS specific things to it aswell. Maybe if I ever have some time I can merge all 3 (Kodi / Logitech / HEOS).

gizmocuz:
that would be great, as i can see this going to 10 remotes ;)
We need a kind of logitech hub remote ;)

sijones:
Kodi and Logitech are the same remote :)

sijones:
Sorry my mistake, it's Panasonic and Kodi are the same!

PR #885, opened 2016-09-22
Panasonic on from standby

sijones:
If the TV reports 400 reponse to web request then the TV will turn on when in stand by, changed the code to implement this.


gizmocuz:
Thanks!

PR #890, opened 2016-09-26
Fix: Evohome HGI80 fixed zone temperature sensor linking

Evsdd:
Fixed bug in Evohome.cpp which caused the temperature sensor linking to fail when first device was created,


gizmocuz:
Thanks !

PR #896, opened 2016-10-04
rewrite enocean VLD support (D2) and bring partial UTI support (D4)

qix67:
```
* VLD decoding is rewritten to bring a better decoding of D2-01-0C device profile
* UTI decoding added to record detected device profile inside table EnoceanSensors

Usage:
1) put the enocean device into learning mode. Automagically, Domoticz will create this device without any action if the device uses universal teach-in (D4 datagram) and announces itself as using profile D2-01-0F or D2-01-12 (module with 1 or 2 channels). The device can be identified using its hardware ID.
2) create a virtual switch in domoticz
3) press the virtual switch in domoticz, the enocean device will record the virtual switch as an authorized switch
4) leave device learning mode
5) attach the light detected during step 1 as subdevice of virtual switch

Note: if you used previous patch, you must put your enocean device into learning mode for few second and leave it to make it send its profile to domoticz. If you don't do this, the light status won't be updated anymore and you will have a message in domoticz log
```


gizmocuz:
Thanks for this great work...

two questions:
- is it possible to put the (1) and (2) behind the word enocean , or at the back of the log line
- it has been quite some time since i started with this, i remember i build a special rule/option when you manually add a EnOncean type light (from the manual light/switches page), where a teach-in was send when you press the test button
  Maybe this solves the additional virtual switch ? I can't remember to have used a virtual switch... not something you would want, because now you get 2 switches for one actual switch.

qix67:
Hi,

> is it possible to put the (1) and (2) behind the word enocean , or at the back of the log line

not sure I understand your question.

About your second question, last time I tried without creating a virtual switch, when I tried to toggle the light (enocean module) directly, domoticz complains it cannot use the ID because it is not in the range of IDs usable by enocean USB key.

gizmocuz:
currently you have now the following two log calls
(1) EnOcean: blaaaa
(2) EnOcean: blaaaa

i prefer:
EnOcean: (1) blaaaa
or
EnOcean: blaaaa (1)

Dit you tried/debugged the 'test' button when you manually add an EnOcean switch ? (this sends a teach-in)

qix67:
Ah! Ok. In fact, I just forget to remove these. I just wanted to identify which function was called. I will fix this.

> Dit you tried/debugged the 'test' button when you manually add an EnOcean switch ? (this sends a teach-in)

I have not seen it when I first use domoticz. I will try to take a look next week-end.

gizmocuz:
Did you had a change to look into this ?

qix67:
Unfortunately, no but tomorrow, I will received additional enocean modules, this will give me the opportunity to test your method

gizmocuz:
Great !, Thanks for your time!

qix67:
Yesterday, I finally install my new enocean modules. I tried the following method:
Usage:
1) put the enocean device into learning mode. 
2) create a virtual switch in domoticz and press the test button (teach-in) before adding the switch.

It works but I still add to go to setup/device/not used, find the enocean device and add it as slave device of the virtual switch else the switch does not display enocean device status. Thus I still have 2 switchs, the one I can press and the one having light status (but I cannot press it).

I will try to modify my patch tomorrow to fix the log line.

gizmocuz:
Great, thanks ! Plz let us know

qix67:
I just detect a little problem. I use a nodon module with 2 channels. On each one, I can connect a physical switch to control the module locally.

Everything works fine as long as I control both channels remotly through domoticz.

Things go wrong as soon as I try to manipulate any of the 2 channels using the local switch. Physically, the light works as expected but domoticz fails to properly display light status. Worse, when I manipulate one of the 2 physical switches, domoticz modifies the light status of both channels. Even worse, in my case, I connect a push button on my module and when I push it, both channels always switch to "group off" (I suppose the result to be a little different if it is a normal switch instead of a push button and both channel toggling between group on/group off).

I dumped packets sent by the module and I notice I received 1 VLD packet correctly reporting light status followed by 2 RPS packets (F6-02) notifying push button status (press then release). 

I don't known how domoticz react if a slave device is a switch. I will try to fix the problem tomorrow afternoon, I have a little idea on how to fix it.

qix67:
I just commit my latest changes fixing the log messages and supporting device behind simultaneously actuator and rocker. In such case, given the fact VLD packet contains light status, the 2 RPS packets notifying button status are useless and are discarded if the device uses profile D2-01-0F of D2-01-12 (only these 2 profiles are supporting button connected to the actuator).

gizmocuz:
Great, thank you very much !

PR #901, opened 2016-10-06
[MultiFun] A few changes

szczukot:


gizmocuz:
Shall we call this code improvement ? I can not see the two new devices

szczukot:
line 413 and 435 added two new devices (alert type)

PR #906, opened 2016-10-09
This fixes the problem that the 'More Options' closes on click.

RaymondNg2:
This is a problem for touchscreen devices because it is nearly impossible to open the 'More Options' menu when it keeps closing onclick.

```
```


RaymondNg2:
This has been tested with
- Chrome
- Firefox
- Edge
- Edge (Windows 10 Mobile)
- IE

gizmocuz:
i do like inline javascript. You could add this to the index.html or domoticz.js
But this also alters the menu for non touch-screen enabled devices ?

jrich523:
Because more options didnt have an onclick to begin with, the impact on a desktop wont be noticed, no one was clicking on it to begin with :)

RaymondNg2:
I will change the PR so that the code will be in domoticz.js.

Do you want me to only apply this to touch/mobile devices?
Or is it ok to apply it to all devices?

Personally I do agree with @jrich523. No one would click on it as that would close the parent menu. 

Whatever the decision I will do what I can. I am already happy to be able to contribute to this great piece of software :)

gizmocuz:
For all devices this would be ok, thanks in advance

RaymondNg2:
I have chosen not to put the code in domoticz.js
It is just not that type of file for this kind of fix.

gizmocuz:
No Problem....
Could you increase the version number in the appcache file as well ?
When you push a change that includes modifications to a html file, this needs to be done

gizmocuz:
Thanks !

PR #911, opened 2016-10-14
[MultiFun] fix warnings

szczukot:


gizmocuz:
Thanks ! That solved it ;)

PR #914, opened 2016-10-15
Quick fix for Toon Eneco blocking Domoticz UserAgent

IgorYbema:
Toon servers are blocking the default domoticz useragent string recently. This is a quickfix to get the Toon to work again, setting the UserAgent string to 'chrome' in the Toon part of domoticz only.
When I have more time I will try to create a config field in the hardware properties of the Toon in Domoticz so that users can choose their own UserAgent string, as it will only take time for this new 'fake' string to be blocked also.

### DIFF
--- a/hardware/ToonThermostat.cpp
+++ b/hardware/ToonThermostat.cpp
@@ -167,6 +167,7 @@ void CToonThermostat::Init()
 
 
 	m_bDoLogin = true;
+	HTTPClient::SetUserAgent("chrome/1.0");
 }
 
 bool CToonThermostat::StartHardware()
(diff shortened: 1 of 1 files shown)


gizmocuz:
We might have a fix for this next week, as you can read on the forum
A useragent is also something like

Mozilla/5.0 (Windows NT 10.0; WOW64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/53.0.2785.143 Safari/537.36

IgorYbema:
Why wait for Eneco to come with a solution? This patch/quickfix is easy and can easily be removed if Eneco comes with a solution from their side.

gizmocuz:
Because we want to play nice ?

IgorYbema:
Whut? You call this not 'playing nice'? And now you create a random useragent process yourself and publish that?
I agree that my pull requests is now obsolete after your random useragent patch but your comment does not make any sense now.

gizmocuz:
It is not only for toon, more for the ATAG, and luckily it is solved for this one as well (as you could see in the commit)
Next, you made a git pull request after my comment
Your pull request also did not increase the poll interval
And last, it is a pull 'REQUEST', no need to accept anything
But thanks for your request

IgorYbema:
Your commit is 2 hours ago. My pull request 3 hours.. We probably overlapped
I really don't care about my patch and my pull request which is now obsolete but I do care about your comment 'playing nice' as your commit which randomizes the User Agent is even more harder for Eneco to detect.

gizmocuz:
But it is nicer to have a large poll interval don't you agree ? Your pull request might have solved the connection issue (for... i dont know), but could cause their servers to have issues again

gizmocuz:
these were two different commits bdw

IgorYbema:
I do agree 5 minutes poll is nicer for hem but I was discussing the Useragent commit diff between your and mine.
Will discuss your 5 minutes poll commit there.

gizmocuz:
Your issue did not solve their problem.
Anyway, hope it is working now.
There are more things to do in live that having a useless discussion here

PR #918, opened 2016-10-16
Allow compressed HTTP response

dynasticorpheus:
Allow compressed HTTP response to improve transfer speed and bandwidth utilization


gizmocuz:
Thanks !

PR #921, opened 2016-10-17
Added Yeelight

corbinmunce:


gizmocuz:
Thanks! I also added it to the visual studio solution so windows user also can use this hardware class

gizmocuz:
Question, why did you make a new type/subtype and command range , and not use the existing AppLamp  /pTypeLimitlessLights type ? (We could have renamed the applamp as it is now used by multiple hardware)
Otherwise we end up with 1000 types, and this is not really needed

gizmocuz:
I am currently rebuilding it...

gizmocuz:
As device ID we could use the IP Address, this saves the need to use the Options field

xbeaudouin:
Sure but this IP Address can be already in use on the network (ie for example another stuff than a router...).

gizmocuz:
Hmmm That wont work, the light is controlled via IP, the destination is IP, so the IP is unique
I will also add a 'Add Manual Light' option, as my wifi is in another network as my normal wired stuff, and a broadcast can not be received by the access point.
Still adding a manual ip address (and specifying type) is working ;)

corbinmunce:
I was new to c++ (and still am) and copied code that looked the closest to what I needed to get it working.  Have since seen better ways to do things and welcome suggestions on how to do it better.

Are you saying I could add each light as a separate hardware IP?

corbinmunce:
Gizmocuz, it is nice adding the Yeelight controller and have it doing the discovery via UDP, but since this doesn't work on all networks I will wait to see how you add it as a 'Manual Light'.  I don't understand c++ very well but do have updated code that I think works better.  In one case is handling a device changing IP on the network.

gizmocuz:
Your code worked 100%, but i am optimizing it a bit
My first push (in a minute) will be the rebuilded class, but not yet the manual add option, as i need more time for this.
Probably this will be pushed tomorrow
You will get new lights when you update

corbinmunce:
I've just had a look at your code with the IP as the device ID, which is good but I think we still need to store the bulb hardware ID somewhere, so that a change of IP address can be detected.  Maybe use options or another field to store the the bulb ID - what do you think?

gizmocuz:
Thats not such a bad idea... but you really want to have static IP's for these lights
And you get a new switch as well.... not ideal

gizmocuz:
I think in all routers/switches you can set fixed addresses ?

gizmocuz:
First i will add the manual option for IP Addresses, then we could add the original id as option
then we can compare the deviceid with the original id, if this is different, we have to change the deviceid

gizmocuz:
committed the code to add a manual ip address (from the hardware page)
Also tackled a crash when developer mode was not enabled on a bulb

corbinmunce:
Great work!

PR #923, opened 2016-10-18
Added SelectorState (LevelStr) in JSON-calls Schedules & Devices

EddyK69:
Added SelectorState (LevelStr) in JSON-calls Schedules & Devices
- http://<ip>:<port>/json.htm?type=schedules
- http://<ip>:<port>/json.htm?type=devices&filter=all&used=true&order=Name


gizmocuz:
What will this do ?

gizmocuz:
Where is this used ?

EddyK69:
I use both JSON-calls in custom html-templates to show all my devices and their scheduled timers. For Selectors the state in 'devices' and the switchcmd in 'schedules' is showing only the level% (as if it was a dimmer). I added the level as a string in both JSON-calls now with an extra field 'LevelStr'.

For instance a selector in state 2 will show:
**LevelStr: "Level 02"**

gizmocuz:
How does the domoticz user interface knows which one/state is active ?

EddyK69:
It just translates the levelint (used as index to an array) to a string based on the levelnames of the device

gizmocuz:
Okey, but this will make the json output larger, can you translate them in your javascript ?
(for me, if it's not needed by the native domoticz gui, i would rather not make it larger, and also the dbase query not larger)

EddyK69:
Yeah, but that means a json-call for every timer-row to catch the levelnames of a selector :(

gizmocuz:
Okey, maybe i do not understand

If you have an index (A) and you create a small lookup table with the strings (B)

B={ "Level 01", "Level 02", "Level 03", "..." }

then plotting the Label (C) = B[A]

EddyK69:
I know, but if i have 30 schedules planned for a selector, i have to get the levelnames for this selector for every item in de schedules-json. 

An other solutions would be (if you do not want to enlarge the json-output) to put the levelname in the existing **status** field instead of an extra field.

gizmocuz:
I think indeed that would be best... also for people not knowing index values ;)

PR #925, opened 2016-10-19
Improve Tellstick support

JBerts:
Clean up whitespaces.
Add support for logging raw device events. This is useful when detecting sensors and remote controls ID.
Add support for repeating commands. Sometimes the command is lost due to poor wireless environment.

It would be nice to being able to configure the amount of re-transmits from the web gui. I will look into how to do that. Can anyone point out a good starting point for creating that?


gizmocuz:
Thanks!
For the re-transmits, if this hardware needs it, it is best to handle this in the WriteToHardware function in the class.
You can make a 'setup' button in the hardware setup for this hardware (like the dummy/rfxcom/opentherm gateway/...), and store values in the hardware parameters in the database

gizmocuz:
Sorry, i need to revert this patch, please do not introduce us to a new boost library
If possible please submit a new pull request without chrono, without changes to CMakeList.txt

gizmocuz:
reverted

JBerts:
Ok, I will remove the dependency on chrono. Just out of curiosity, why is it not possible to use any other boost libraries?
Is it really ok to delay/sleep in WriteToHardware? I thought it would be a good idea to handle the re-transmits in a separate thread. I do want some delay between the re-transmits.

gizmocuz:
Because the build systems do not have this, other users dont have this, the binary will get larger, and it is not needed
Maybe you can make a queue inside your hardware class, see MySensors, that is a good example of a non-blocking-queue

JBerts:
I can easily simplify the implementation so that it don't use the monotonic clock but only use a delay and always does the re-transmits after a wake up. Although it won't be as good, it will cause spurious transmits, e.g. when a new command is added to the queue while another one is re-transmitting.

Another alternative would to use (the deprecated) boost::get_system_time that is not monotonic and  cause problems if the system time is changed when waiting for the condition variable.

I've looked around the code base trying to find other use of a portable monotonic clock usage without success. As far as I know the 3 main ways to do it: boost::chrono, std::chrono (c++11) or own wrapper that uses clock_gettime/QueryPerformanceCounter.

I guess it wont matter that much in this case, but it hurts to make a less correct solution. Don't take me wrong, i'm only trying to make a good effort here.

The binary size increases with 67 kB when linking with boost:chrono. I find it hard to believe that other users have a problem with missing that library when they apparently already have installed the other used boost libraries.

What about enabling c++11? Is that problematic in Windows? I'm not that up to date with windows c++ programming/compiler support.

gizmocuz:
For (almost) sure we can not enable c++11 (unfortunately)
We do NOT want to force people upgrading their GCC
The same for boost, did you try compiling boost on a raspberry, that will take ages.
Our wiki page is explaining how to build only the necessary parts we need, and this does not include chrono
When a new command is added, while another is transmitting ? i do not see any problem with this ?
You send an 'ON' it is in the queue, maybe even transmitted at the moment, and at the same time you put an 'OFF' in the queue, that gets fired next.
And of course, i think we are talking about milliseconds here?
About the monotonic clock, when using a queue, you do not have to use this. are we talking about milliseconds here as well ? or are you talking about daylight saving times ?
I would go for the queue, or even better, maybe this can be solved in the tellstick ?
I did not have transmit issues with other hardware, if a sensor is to far away, that it could mess a command, then i would solve that on a different level.

PR #927, opened 2016-10-21
more responsive event system

bobkersten:
This change will make the event system more responsive as discussed in this thread on the developer forums. I know that for many an interval of 1 second is more than enough, but my situation requires a much shorter interval. Both are now possible. Existing systems won't require any change to keep on working like they did, and for others it's now possible to trigger events much quicker.

http://www.domoticz.com/forum/viewtopic.php?f=9&t=13509

A few notes:
- Comparing a float with zero doesn't make sense, so on all those situations I've used halve an interval to compare with.
- Accessing size of a list should also be thread safe in SQLHelper.cpp.
- This code doesn't have a visual impact on performance on my Raspberry Pi 2.
- If and when this code is implemented I've got additional code ready that uses this for AFTER, RANDOM and FOR to use MSECONDS, SECONDS, MINUTES or HOURS, a feature that has been requested on the forums in multiple threads.

### DIFF
--- a/main/EventSystem.cpp
+++ b/main/EventSystem.cpp
@@ -1920,17 +1920,17 @@ bool CEventSystem::parseBlocklyActions(const std::string &Actions, const std::st
 			}
 			else if (isVariable)
 			{
-				int afterTimerSeconds = 0;
+				float afterTimerSeconds = 0;
 				size_t aFind = doWhat.find(" AFTER ");
 				if ((aFind > 0) && (aFind != std::string::npos)) {
 					std::string delayString = doWhat.substr(aFind + 7);
 					std::string newAction = doWhat.substr(0, aFind);
-					afterTimerSeconds = atoi(delayString.c_str());
+					afterTimerSeconds = atof(delayString.c_str());
 					doWhat = newAction;
 					StripQuotes(doWhat);
 				}
 				doWhat = ProcessVariableArgument(doWhat);
-				if (afterTimerSeconds == 0)
+				if (afterTimerSeconds < (1./TASK_PROCESSOR_HZ/2))
 				{
 					std::vector<std::vector<std::string> > result;
 					result = m_sql.safe_query("SELECT Name, ValueType FROM UserVariables WHERE (ID == '%q')", variableNo.c_str());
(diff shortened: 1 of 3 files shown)


gizmocuz:
Some questions/suggestions:
I know what you want to do, and i understand why, only could this not be done simpler ?

(why is it called TASK_PROCESSOR_HZ ?, maybe just call it 'schedule_resolution' or timer_resolution)
- If the default was 1 second, and the TASK_PROCESSOR_HZ = 25 now, this will now sleep 40 ms ?
- making the delay time a float is perfect
- If we make the schedule_resolution 500ms (for instance), we sleep 500ms
  subtract 500ms from the delay/counter, when the counter <=0 then... we do something
  (that should also work for a float?)
- i do not understand:
  if (m_iAcceptHardwareTimerCounter <= (1./TASK_PROCESSOR_HZ/2))

bobkersten:
- TASK_PROCESSOR_HZ = 25 means the event system will run 25 times per second and indeed sleeps for 40ms. I thought calling it Hz immediately makes it clear how often it runs (Hz includes the timeframe, a second), but I'd be glad to change it if you prefer so.
- Comparing less then or equal does work for floats, it's just not guaranteed to be exactly zero. Hence the comparison with half an interval (1./TASK_PROCESSOR_HZ/2). This also captures situations where the float would be 0.000000001. Using half an interval puts the turning point exactly in the middle and leaves no room for rounding errors.

gizmocuz:
25 times a second ? dont you think that is a bit to much ?

bobkersten:
That's why I've turned it into a #define, to test various frequencies :). Turns out that for my situation I need at least 10 times a second (10Hz, 100ms interval), but like I said, it makes no difference performance wise to push it a bit more. I've had it running for days at 50Hz (20ms interval) without any problems whatsoever, but that was overkill for sure.

szczukot:
maybe timer_resolution_hz will be read from setup->settings ?

bobkersten:
I've thought of that too, but I came to the conclusion that it wouldn't hurt 'ordinary' users in terms of performance (or anything else for that matter)  if the event system runs more frequently. They wouldn't even notice it. And for those in need of it, it is just there :) so why bother making this a setting, imho.

pj-r:
Nice work!
We already had some discussion about this. Nice to see someone had time to make the implementation!
[Problem: actions take a full second to trigger](https://www.domoticz.com/forum/viewtopic.php?f=6&t=11619&p=92910#p92910)

For me the 10 times/sec was quite good.

gizmocuz:
Sorry, i forgot to merge this ;)... Thanks for the patch

PR #929, opened 2016-10-22
Fix inverted door lock status

david81:


gizmocuz:
Sorry, this is not a fix.
- It might work for you, surely not for me, as mine is correct. You need a new type, or a flag
- I also don't think the icon state will be correct? (Maybe it is ?)

david81:
Ok. There were multiple forum threads regarding this issue with no one stating that it was correct for them, so I assumed it as incorrect for everyone.

david81:
How are the options for each switch type defined (different switch types have different options that can be set)? Perhaps adding a flag to invert status here would be the "proper" fix to allow the current switch type to work for everyone?

gizmocuz:
A code fix has to be made on multiple places in the c++ code and html code

Issue #932, opened 2016-10-24
logrotate compatibility


sbouchex:
Current server implementation does not allow logrotate to rotate de domoticz log.
What needs to be done:
- Provide a log rotate.d script to will rotate the file
- Modify the startup script to have a "reload" action which will do a kkill -HUP to domoticz

I can work on it

gizmocuz:
You can for a log rotate by sending a SIGHUP to an application, also for domoticz (non win32 only)

sbouchex:
Yup but the "reload" action is missing to do a kill, a simple pkill -HUP domoticz will do the job, I plan to provide a logrotate.d as well.

gizmocuz:
Okey ;) perfect

sbouchex:
Any suggestion to locate the logrotate.d file ? under script ? elsewhere ?

gizmocuz:
scripts sound like a nice location for this

PR #933, opened 2016-10-24
Add logrotate support

sbouchex:
Added a logrotage template to weekly rotate/compress the domoticz log file

see  #932


gizmocuz:
Thanks!

PR #937, opened 2016-10-27
add suport for SEAHU PLC COMPUTER based on raspberry PI

OndrejLycka:
PLC computer http://seahu.webnode.cz/sh017/ by Ondrej Lycka e-mail: ondrej.lycka@seznam.cz

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -344,6 +344,7 @@ hardware/ZiBlueSerial.cpp
 hardware/ZiBlueTCP.cpp
 hardware/ZWaveBase.cpp
 hardware/Yeelight.cpp
+hardware/Seahu.cpp
 notifications/NotificationBase.cpp
 notifications/NotificationEmail.cpp
 notifications/NotificationGCM.cpp
--- a/hardware/Seahu.cpp
+++ b/hardware/Seahu.cpp
@@ -1,0 +1,306 @@
+#include "stdafx.h"
+#include "Seahu.h"
+#include "../main/Helper.h"
+#include "../main/mainworker.h"
+#include "../main/localtime_r.h"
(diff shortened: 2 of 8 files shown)


gizmocuz:
Nice PLC, but it seems raspberry specific... we do have a hardware support for I2C (RPi) (Logical I2C sensors), with already 3 types.
Maybe you could make it so, that this will be a forth ? (i think it will run on all arm systems, like a odroid or cubieboard as well ?)

The webpage is in a strange language that some people will understand, other not
Is this the manufacturers website, or a setup/installation instruction by you?
Do you have a website to the manufacturer ? Also where you can buy it ?
Could you add a description on top of your class with this information as well ?
If it is good for more people... sure we can add this ;)

Thanks again

OndrejLycka:
Thanks you for your comment
You have right, I can add suport for PCF8574 chip to /hardvare/I2C.cpp.
Or I can add support throught MQTT, or dummy devices.
Idea this device is easy use. Manualy configuration every pin with generic hardware is not this idea, therefore I would like use own hardware add. 
This devide is created for dimension raspberryPI, I not presume use another platform.
Webpage is in construction.
Now I finishing with software, then I will work on documentation, manuals and website and reapair formalit.

gizmocuz:
It is just to combine all I2C hardware/sensors into one dropdown for the end-user.
When your hardware is true I2C, it should work on other devices as well (also arduino's for example)
When you have everything organized, please let us know here, maybe you could send me a hardware sample ?
But please consider using the I2C dropdown method explained above, you will see that the current 3 supported hardware types are having their own class as well

gizmocuz:
Maybe it's best to create a python/bash script for this hardware as it seems to not be a generic hardware, or not used by many users

PR #940, opened 2016-10-27
Ignore lines beginning with #

codeaholics:
I'm developing some custom firmware for Domoticz/Evohome integration on a Raspberry Pi. It is useful for me to have the firmware write "debug stuff" out to the serial interface, but obviously that needs to not interfere with the operation of Domoticz. 

This patch causes the Evohome integration to ignore all incoming lines from the serial interface which begin with #. Without this patch, it logs error messages about the message being corrupt, and there's no visibility of what the line was.

This is mostly useful for me at this point, so feel free to reject.

### DIFF
--- a/hardware/evohome.cpp
+++ b/hardware/evohome.cpp
@@ -717,6 +717,12 @@ bool CEvohomeMsg::DecodePacket(const char * rawmsg)
 		std::string tkn(tkns[i]);
 		if(i==0)//this is some sort of status or flags but not sure what exactly
 		{
+			if(tkn=="#")
+			{
+				SetFlag(flgign);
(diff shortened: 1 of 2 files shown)


gizmocuz:
Debug messages could be nice as well. Rejecting this for the moment, but thanks for the pull request
Maybe you can use a 'tail' command with a grep or awk ?

codeaholics:
Unfortunately, there's nothing to tail. When Domoticz doesn't understand an incoming message it complains in the logs, but doesn't actually log out the message it received. This patch adds that logging, and suppresses the "error" logging when it recognises a "comment" line.

PR #941, opened 2016-10-28
Add instructions for Philips Hue

jdongelmans:
Added instructions for adding Philips Hue hardware.


gizmocuz:
Thanks, i will also add this to the translation system

PR #945, opened 2016-10-29
[Events] Timed commands

bobkersten:
Next up :) more enhanced timed commands. With this patch you're now able to specify the time in seconds, minutes or hours. In addition, you're able to combine FOR, AFTER and RANDOM and I've added an additional REPEAT keyword. This makes the scheduling of commands extremely flexible. Ofcourse it is completely compatible with the current situation, so there's no need to change scripts.

AFTER 5 SECONDS TURN On FOR 1 SECOND REPEAT 10 2 SECONDS
Turns on the light after 5 seconds, leaving it on for 1 second and repeats this an additional 9 times with 2 seconds in between.

Off AFTER 5
On FOR 1
When time quantifiers are not specified, the default is used as it is right now, so the first turns something off after 5 seconds, the second turns it on for 1 minute.

On REPEAT 3 1 SECOND
For those who use the rfxcom 433MhZ transmitter and sometimes experience commands being missed, this one is a great addition. Ofcourse, the best solution would be to upgrade to z-wave, but until then, this one makes a great addition. It repeats the command 3 times with 1 second intervals.


gizmocuz:
Great patch, for some logical names (in blockly) could you make it
REPEAT 10 DELAY 2

repeating 10 times, with 2 seconds delay in between

It is not a good idea to say (and nice to say...) the repeat is good for 433Mhz, as this is not true, you have the same for zwave, but this will try it more times.
You have this also for 868/MySensors, any sensor with one-way communication
(Zwave is expensive, not all can buy these)

Best is to make sure all sensors are in range, or gain some dB's by using a better antenna (which might not be legal in all countries)

gizmocuz:
Could you also post a 'commandArray' example, as i would like to build a block for blockly with this ;)
And we could add this to the lua wiki as well
Maybe next, when we have a repeat with 1000, we need a 'cancel' command that will directly remove all schedules for that idx (as after a cancel we might want to send 'off'

bobkersten:
This one can be used to indicate an entry delay for alarm systems:
commandArray['Entree Lamp'] = 'On FOR 1 SECOND REPEAT 20 1 SECOND'
The examples mentioned in my first post can all be used for commandArray.

I've only tested it using LUA, I've never used Blockly before, but if the ScheduleEvent is used in the same way it should "just work" because it's fully compatible with the current implementation.

You're right about saying it isn't a "good" thing to have to repeat 433MHz commands to make sure they arrive, but in reality, that's just how it is. I've made a LUA function just for that. I need to make sure my Somfy blinds are closed when it gets dark, and as those are located upstairs and obviously fixed in place :) repeating the commands is the only thing that works for me.

I'll add the REPEAT 10 DELAY 2 change asap, that indeed makes the command even more readable.

bobkersten:
Oh and cancelling is already in place, just not from a button in domoticz. Once you schedule a new event from LUA or Blockly, it removes all events in the future for that same device. That's how it has always worked. It might be a good idea to extend this behaviour to direct switches too.

bobkersten:
I've named it INTERVAL btw, makes more sense than DELAY imho.

gizmocuz:
great, so now it becomes

commandArray['Entree Lamp'] = 'On FOR 1 SECOND REPEAT 20 INTERVAL 1 SECOND'

do we need 'second' ? because now it is

commandArray[3727]="Off AFTER 3"

i might get confused here now... 

i am working on a new release version... might be best to put this in the next beta after a new release ?

gizmocuz:
Going to make a release now... exactly 1000 commits since the last one ;)

bobkersten:
When you're using AFTER there's no need to put SECONDS there because that's the default for AFTER. Hence, "Off AFTER 3" will still work. Same goes for all the scripts that use FOR. "On FOR 1" still turns something on for 1 minute, because minutes are the default for FOR. I agree that it is confusing that the default for AFTER is seconds and for FOR and RANDOM it is minutes, but that's the way it works right now and we can't change that without breaking scripts. So I think the Wiki should be updated to always include SECONDS, MINUTES or HOURS.

As for including this in the next release; that would make the release really cutting edge :) might want to allow more developers to test this a little more :)

gizmocuz:
We got a small problem in the current eventsystem, the SET XX to YY for ZZ Minutes, ZZ is now handled in seconds, same for the random blockly.
I will revert this to minutes, but making some blockly options for seconds is a great idea ;) Hours... that is maybe to much... but why not haha

bobkersten:
Resolved merge conflict. I don't know how I've missed that second/minute bug. I think it has to do with the fact that my own source already had this patch in place, and by splitting it up into two separate merge requests I must have missed it. Anyway, if you look at line 2141 and 2147 of EventSystem.cpp you'll see that the defaults are correcly handled.

gizmocuz:
Do i need to change anything, or was it fixed with my previous patch ?

bobkersten:
No changes necessary. The part you've changed was obsoleted with this new patch anyway.

bobkersten:
Reverted the stripping of set- and play from commands. This was done to also allow "SET On" and the likes, but it caused the "Set Level" commands to fail after being picked up by the event system. "TURN On" is still allowed though. Also made the device timers more efficient by not querying the database from within the loop.

bobkersten:
Set Level and Set Volume will now properly revert back to their old state when using FOR. This also didn't work prior to this patch, so that's fixed now.

gizmocuz:
Could you remove the boost tokenizer usage and use the splitstring function ?
Want to get rid of boost someday

bobkersten:
Done!

gizmocuz:
Thanks
... Another one ;).. the whole event system is case sensitive, no need for case insensitive compare
But else, i would gave made it the c++ way

bobkersten:
Also done! :)

gizmocuz:
Sorry, could you use the c++ std::string compare method ?

to compare the whole string

if ( sToken=="FOR" ) {

or to see if FOR is used

if ( sToken.find("FOR" ) != std::string::npos ) {

bobkersten:
There you go!

bobkersten:
Is there something else you'd like me to change?

gizmocuz:
Nope... there are two pull requests that have a big influence (this is one), i just wanted to make a new stable release to solve the previous notification issue
Tomorrow i will have a look this one ;)

gizmocuz:
Thanks !

Siewert308SW:
Question regarding this repeat commandArray.
Today i saw it was available for a long time.
And tried it with a Lua script on the latest Beta for my rpi3 but somehow the commandArray isn't fully working or i'm missing something.

works:
commandArray['Dummy'] = 'On FOR 5 SECOND REPEAT 5 5 SECOND'

doesn't work and only triggered once after 5
commandArray['Dummy'] = 'On AFTER 5 SECOND REPEAT 5 5 SECOND'

Doesn't even trigger the Dummy
commandArray['Dummy'] = 'On REPEAT 5 5 SECOND'

Siewert308SW:
Nevermind my previous question...
Commands below do work.

commandArray['Dummy'] = 'On REPEAT 5 INTERVAL 5'
commandArray['Dummy'] = 'On AFTER 5 REPEAT 5 INTERVAL 5'
commandArray['Dummy'] = 'On FOR 1 REPEAT 5 INTERVAL 5'

PR #949, opened 2016-10-31
fix for log- and notification click not working

bobkersten:
Clicking the log- and notifications button on temperature widgets often didn't work because a drag was already initiated. When using a Wacom tabled, this problem becomes much more apparent, because pressing often also slightly moves the cursor. This irritated the hell out of me, so I started looking for the cause. Also because all other buttons, including the same buttons for other tabs, did work properly.

The reason it works differently is that these two buttons on the temperature widget are the only buttons that use hrefs (ng-href), so I've changed them to ng-click like the rest of the system, and now clicking works perfectly.


gizmocuz:
Thanks !, I probably will remove the need for the location completely as those two controllers just calls the old code.

PR #950, opened 2016-11-01
[LUA] Ability to disable recurring logging

YcKe:
Add the ability to disable recurring logging when a LUA script is
executed. This can be specified in the LUA script by adding `_donotlog_`
to the script name. E.g.: script_time_donotlog_MyScriptName

### DIFF
--- a/hardware/EnOceanESP3.cpp
+++ b/hardware/EnOceanESP3.cpp
@@ -153,8 +153,7 @@ typedef enum {
 	RORG_SM_LRN_REQ = 0xC6,
 	RORG_SM_LRN_ANS = 0xC7, 
 	RORG_SM_REC = 0xA7,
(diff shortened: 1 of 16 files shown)


gizmocuz:
It would be much nicer to let the user specify this in the script.. is that possible ?
This is a quick hack ;)

gizmocuz:
Pff lot of changes.. how does this work now ? Do you have a small lua example ?

YcKe:
The recurring logging is disabled when the user adds the following statement to the Lua script:
commandArray['DoNotPerformLogging'] = "1"

Incase the value is 0 or the value is not added to the commandarray, the recurrent log is shown as it was originally. This way existing scripts are not impacted.

Thanks for your time!

gizmocuz:
Dont know what kind of patch this is... but something must have gone wrong

PR #954, opened 2016-11-02
[Events] fixed endless loop when sending specific notification

bobkersten:
When the notification contains an exclamation mark, Domoticz would end up in an endless loop. The code uses 'continue' to process the next event without incrementing the event iterator. This causes 100% cpu load ofcourse and a failing Event system.


gizmocuz:
Thanks !, Yes this makes sense ! whoops

PR #956, opened 2016-11-03
[LUA] Ability to disable execution logging in the log (Cleaned version)

YcKe:
Add the ability to disable exection logging when a LUA script is
executed. This can be specified in the LUA script by adding
DoNotLogExecution to the commandArray and setting it to 1.
E.g.: commandArray['DoNotLogExecution'] = "1"

### DIFF
--- a/main/EventSystem.cpp
+++ b/main/EventSystem.cpp
@@ -2863,12 +2863,13 @@ void CEventSystem::luaThread(lua_State *lua_state, const std::string &filename)
 	status = lua_pcall(lua_state, 0, LUA_MULTRET, 0);
 	report_errors(lua_state, status, filename);
 
-	bool scriptTrue = false;
+	std::pair<bool, bool> scriptResult;
+
 	lua_getglobal(lua_state, "commandArray");
 	if (lua_istable(lua_state, -1))
 	{
 		int tIndex = lua_gettop(lua_state);
-		scriptTrue = iterateLuaTable(lua_state, tIndex, filename);
+		scriptResult = iterateLuaTable(lua_state, tIndex, filename);
 	}
 	else
 	{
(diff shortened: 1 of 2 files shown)


gizmocuz:
Maybe we should make a global settings to disable 'info logging' for the event system so only errors are logged?

YcKe:
The advantage I see on this way of working is that it can be specified in the script itself. So in the case that some scripts need supervision and others don't.

E.g.: Execution of the fan controlling script should be displayed but the script to accumulate my consumptions doesn’t.

The advantage on your method is that it’s more user friendly as it can be setup in the screen. But people writing LUA code probably won’t have an issue with that. 

Please let me know what you think.

YcKe:
Hi gizmocu,

You have an update on this pull request? 
Please let me know your doubts if you have any.

gizmocuz:
Well the solution you provided (which is working) is a bit to abstract for me.
I rather have a global settings to turn off logging of event script triggers
On the other hand, you only look at the log when you're debugging, so it does not hurt to much
How about a global option ? Then if you need logging, you just log in the lua script

YcKe:
Hi gizmocu,,

That's fine for me. I'll create a new pull request for these changes.
How would you like to name the setting? Event script trigger logging? (Check box)

gizmocuz:
Maybe something like

> 	nValue = 1;
> 	if (!GetPreferencesVar("LogEventScriptTriggers", nValue))
> 	{
> 		UpdatePreferencesVar("LogEventScriptTriggers, nValue);
> 	}
>         m_bLogEventScriptTrigger=(nValue!=0);
> 

Trigger a log line when an event/script is triggered 

Something like that ?

YcKe:
I'm going to close this Pull Request and create a new one for the global parameter.

PR #959, opened 2016-11-04
Update nas-pd01z.xml

steljwagh:
Added parameter 10, set the LED status, default is enabled.
Changed parameter 1, value = 8 to 255 and default is 12.
Also changed the description of parameter 1.


gizmocuz:
Thank you for the patch
But you have to commit this to openzwave 
Then we will sync again

steljwagh:
No problem, anytime.

I'll move my commit to zwave.

Issue #805, opened 2016-08-20
[issue] Master/slave setup with thermostat device


bdegroot123:
https://www.domoticz.com/forum/viewtopic.php?f=6&t=10030&p=70726#p70726

Hardware: Raspberry Pi 1 model B (master and slave setup, both raspberry), with RFXcom transeiver
OS: Domoticz image SD card, with latest beta software (but problem exists in earlier versions as well
Domoticz branch: Beta
Bug/problem or a question: Bug /problem

Problem: When my smartwares thermostat device is connected to the slave domoticz instance, and i want to manage it thru my master domoticz instance. The values are not set at all.
If I set the setpoint directly from my slave domoticz instance, everything works well, and also the status is correctly updated on the master domoticz instance.

The same issue accurs with a nefit thermostat device. So it is likely that the problem exists when using a themostats in combination with a domoticz master/slave construction.

bdegroot123:
## Feedback from gizmocus:

I think you could be right here...

i think we have to look into the function

MainWorker::ProcessRXMessage

and in

MainWorker::PerformRealActionFromDomoticzClient

it should call 

MainWorker::SetSetPoint(const std::string &idx, const float TempValue)

michsv7:
The issue seems to be as feedback from gizmocus in MainWorker::PerformRealActionFromDomoticzClient.
At line 1710: Unit = 0; could be changed to Unit = pMeter-dunit;

When Unit is hard coded to 0, the "original hardware" ID lookup in the SQL query below in the same function fails.  
This results in that the client cannot map the thermostat change correct. In the client you will get a new thermostat with same ID but unknown hardware. (ID=0)
.
else if (devType == pTypeThermostat)
    {
        const _tThermostat _pMeter = reinterpret_cast<const _tThermostat_>(pResponse);
        sprintf(szTmp, "%X%02X%02X%02X", pMeter->id1, pMeter->id2, pMeter->id3, pMeter->id4);
        ID = szTmp;
        Unit = pMeter->dunit;
    }

// find our original hardware
        // if it is not a domoticz type, perform the actual command

```
    result = m_sql.safe_query(
        "SELECT HardwareID,ID,Name,StrParam1,StrParam2,nValue,sValue FROM DeviceStatus WHERE (DeviceID='%q' AND Unit=%d AND Type=%d AND SubType=%d)",
        ID.c_str(), Unit, devType, subType);
```

bdegroot123:
Hi,
[...]

PR #961, opened 2016-11-05
issue] Master/slave setup with thermostat device #805

michsv7:
When Unit is hard coded to 0, the look up of original hardware id in the
client will fail.

#805


gizmocuz:
Thanks for the patch !

PR #963, opened 2016-11-05
OpenWebNet update

Spugna85:
Update names and comments
Introducing light/switch devices


gizmocuz:
Great! Thanks for the hard work !

PR #965, opened 2016-11-06
OpenWebNet light/switch

Spugna85:
Implementation of the lights and switches management.


gizmocuz:
Thanks for the great work !

PR #971, opened 2016-11-09
Fix time handling when crossing DST jumps

gordonb3:
Oops... sorry. Did something terribly wrong and deleted the branch.

See inline comments for details

### DIFF
--- a/hardware/EcoDevices.cpp
+++ b/hardware/EcoDevices.cpp
@@ -145,8 +145,6 @@ void CEcoDevices::GetMeterDetails()
 	//Send Electra if value changed, or at least every 5 minutes
 	if (
 		(m_p1power.usagecurrent != m_lastelectrausage) ||
-//GB3: Unsafe to assume time_t format is seconds
-//		(atime - m_lastSharedSendElectra >= 300)
 		(difftime(atime,m_lastSharedSendElectra) >= 300)
 		)
 	{
--- a/hardware/EcoDevices.cpp
+++ b/hardware/EcoDevices.cpp
@@ -145,7 +145,9 @@ void CEcoDevices::GetMeterDetails()
 	//Send Electra if value changed, or at least every 5 minutes
 	if (
 		(m_p1power.usagecurrent != m_lastelectrausage) ||
-		(atime - m_lastSharedSendElectra >= 300)
+//GB3: Unsafe to assume time_t format is seconds
+//		(atime - m_lastSharedSendElectra >= 300)
+		(difftime(atime,m_lastSharedSendElectra) >= 300)
 		)
 	{
 		if ((m_p1power.powerusage1 != 0) || (m_p1power.powerusage2 != 0) || (m_p1power.powerdeliv1 != 0) || (m_p1power.powerdeliv2 != 0))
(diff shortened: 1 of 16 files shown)


gordonb3:
Seems the automatic build system is timing out on account of the number of source files that need to be rebuilt?

gizmocuz:
I have no idea.. i checked the file... and it is a good 7z archive
Also nothing has been updated for quite a while there... strange
Guess no windows updates for some time...

gizmocuz:
Ehh time_t is seconds, please revert this
http://www.cplusplus.com/reference/ctime/time_t/

gordonb3:
Did you read that link correctly?

For historical reasons, it is **generally** implemented as an integral value representing the number of seconds elapsed since 00:00 hours, Jan 1, 1970 UTC (i.e., a unix timestamp). **Although libraries may implement this type using alternative time representations**.

**Portable programs should not use values of this type directly, but always rely on calls to elements of the standard library to translate them to portable types**.

gizmocuz:
Thats crazy, for sure till 2038, and still then it will be seconds.
Unless you can proof (with c++ code) that this is otherwise, we can change this.

For now, please change it back, i do not want to see 100 cpp files changed, then it is for sure very hard to merge this pull request in the feature

gordonb3:
There are probably just about 10-20 changes directly related to this change. Most of them replacing timeA - timeB = seconds with the difftime function that is guaranteed to return seconds and is already a double.

Al the other instances are related to fixing incorrect time values being returned because the calculations cross a DST jump
Example 1: it is March 25, 2017 and you like to know how many seconds have passed since midnight. So you do (hour \* 3600 + minutes \* 60 + seconds) but there are in fact 3600 seconds less in that particular day.

Example 2: next you want to get the usage from the previous day. So you load the midnight time in a tm struct along with the current DST state (which is ON) and subtract a day. But because you set DST to on, mktime() will subtract another hour. This results in the date part pointing to March 23, 2017.

BTW as you can see there are only 16 files changed. The reason why this fix targets so many files is because I changed localtime_r.h. As a result, even if I revert # fb80124 it will still require rebuilding the 11 files touched in that commit.

gizmocuz:
Hmm ok your right
But then again, putting the shortlog in UTC would have been better

gordonb3:
I know you think that, but when retrieving the data from the database we can only rely on mktime() which references localtime_r(). Of course you can include many directives to the preprocessor to make Windows use __mkgmtime() and Linux timegm() but there may then still be systems on which Domoticz will not compile. The safe bet therefore would still be to have SQLight return all datetime values relative to localtime even when storing them as UTC.

And there are other complications as well. Currently you can fetch all data from the current day by simply running a query with a datetime value of `%Y-%m-%d` and this is in fact done many times in the current code. If you switch to UTC you will have to specify the exact starting time, meaning you will still need to find what time was midnight and adjust for any DST jump that may lie within the range. We can simply not avoid needing to do DST corrections, regardless of whether the database stores localtime or UTC.

gizmocuz:
Could you remove all comments about time_t ?

gordonb3:
Of course. I added the comments to keep track of what I was doing (which turned out to be a good thing because I did in fact forget to add the replacement code after commenting out the original sections in some occasions) and intended to delete the obsolete lines afterwards, but figured it would be better to show what I did and why.

PS I found a neat trick to determine the unit of time_t on stackoverflow. Dead simple actually. You create a class that holds a public variable for `onesecond` and initialize that by calculating the difference between two time_t values that were created with `mktime()`. As you know, this will currently return 1 on POSIX style systems but it might return something else. You could do the same for `oneminute` and `onehour` and address those values instead of the constants 3600 and 60 when calculating a time offset.

gizmocuz:
Thanks! Lets see how this goes in March ;)

PR #974, opened 2016-11-12
-ssd/sdcard-mode feature to reduce amount of writes

blmpl:
I just don't like when files names contain some kind of time stamp and they are not aligned for easy sorting ;) 

2nd part is a bit more twisted.
My domoticz instance receive data from sensors every few seconds.
In consequence during one day of work sqllite performs ~hundred MB of writes(!) 
After few months my old sd-card were corrupted (and I've got problems with another new one before I started digging what's going on)

So... I moved -dbase to ram (/tmp) to reduce sd card usgae, but if you accidentally pull the plug or forget about making a copy before restarting... yes AutoBackup is quite handy but I wanted to automate it a bit.

this patch create extra "backup_latest.db" file in backup directory
it's pointing to latest hourly backup (on windows it's just a copy)
or - when program is closed gently - this file would be replaced by latest db dump.

if you start program with -autorestore option and -dbfile does not exist + we have a backup_latest.db, the dbfile will be restored before program start.

that's all - hope it could be useful for someone else who use sdcard/ssd for db storage

if you have some other ideas/suggestions here please let me know :) 

...and thank you for this program :)


gizmocuz:
Could you make a separate pull request for the leading zero, or change this one ?
I dont think we need the auto restore option, there are also other options:
- do not use sensors that pushes data every second ;)
- if your on a raspberry pi 3, use the latest image from raspbian, you can boot directly to usb or network
- use iSCSI, with for example 'berryboot' or google for some great tutorials

I would recommend the later

Domoticz should also store some of it's database in ram, and flushes when needed

blmpl:
I created new pool requests just for zeros, so it will be easy to merge.

Solutions you suggested would move writes to other devices, so it's partially +we need other device over the lan which can't go to sleep or lose one usb port for pendrive (+for "non pi" boards you always have to buy sd-card to start booting process)
I saw some topics on the forum abut similar case (I/O errors) and people are using ramlog or - as you mentioned - usb/external drive... but why to do that if we can just reduce amount of writes.  

Maybe we don't need autorestore function... but we can name it differently ;)
Actually the code above realise some part of in ram database, which is flushed to disk from time to time. Advantage of this approach (vs ':memory:') is that even when program gets killed, we have a working copy in other folder (on tmpfs ideally).

so instead of autorestore it could be called ssd or sd-card mode with warning that you can lose up to 59minutes of data in case of hard reboot (but intervals could be adjusted in gui/by param if needed)

so for example:
-ssd-mode 
-ram-storage=/tmp/ 
and -dbfile would point to the latest snapshot location

My marketing skills are rather poor, but I believe it could save some ssd/sd devices ;)
Please let me know if you see some potential here and I would happily adopt it.

gizmocuz:
Will this not always wrote a latest backup for windows users?

`if (!latestLinked && (lDir = opendir(szBackupDir.c_str())) != NULL)`

But i don't think this is needed... have you considered my comments above about usb/nfs/iscsi usage ?

gizmocuz:
Sorry closing pull request, if you cant answer a simple question it stops

blmpl:
sorry, for a late replay, but I wanted to update the code after I saw you comment (it was a bug)
and I hope it is better now.  

yes, I was thinking about your solutions, but I'm not running on raspberry so can't boot from network + network is another point of failure + even over the network some storage would be heavily used. At the moment my board could be cut off from the network and still do the job ;) 
adding pendrive probably could solve a case for a while, but also pendrives has a limited lifetime and the issue is with amount of writes not a storage type.

I don't think we should think about this as an issue with sensors - yes they could report values less often, but what if we just have much more of them? 
If we can't (?) avoid writes, the reasonable solution for me is to move them to the place where they are almost unlimited.

I admit that initial solution wasn't clear but this commit and options should be self explaining. (I hope ;) )
If you see that it should be some how fixed/changed to be more user friendly just let me know.
if you think this shouldn't be merged I will have to maintain it by my self, but I thought that someone else (defiantly not everyone) could use it.

PR #978, opened 2016-11-13
Attempt a reconnect if the proxy connect attempt fails

ToniA:
The 'resolver.resolve(query)' throws as exception if it cannot resolve the network address. If the network connection is lost, the resolve fails, and the exception is not catched untill the execution reaches CProxyManager::StartThread(). This will exit the proxy thread, which again is never restarted.

This is very easy to test:
- Configure Domoticz to connect to MyDomoticz
- Pull the network plug. After a while Domoticz gets disconnected from MyDomoticz
- Reconnect the network. Without this change, Domoticz does not reconnect to MyDomoticz

### DIFF
--- a/webserver/proxyclient.cpp
+++ b/webserver/proxyclient.cpp
@@ -97,20 +97,28 @@ namespace http {
 
 		void CProxyClient::ContinueConnect(const boost::system::error_code& error)
 		{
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thanks!

PR #979, opened 2016-11-13
Leading zero in backup filenames

blmpl:
just it ;)

+ one question:
what do you think about adding dependency to boost-filesystem? (beside it's a bad idea:P)

than instead of:
#ifdef WIN32
      backup_DirH << szUserDataFolder << "backups\\hourly\\";
...

we could use:
boost::filesystem::path::preferred_separator

also I thought about compressing files *before* storing them to drive to reduce amount of writes, but I'm missing portable "mktemp" which could be find there too ;) [temp_directory_path/unique_path]
we have zlib already...

### DIFF
--- a/main/mainworker.cpp
+++ b/main/mainworker.cpp
@@ -1226,7 +1226,7 @@ void MainWorker::HandleAutomaticBackups()
 		if ((lDir = opendir(sbackup_DirH.c_str())) != NULL)
 		{
 			std::stringstream sTmp;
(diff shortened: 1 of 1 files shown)


gizmocuz:
Unfortunately this will introduce another library that has to be build, and is not available on current systems
But you could make your own function for it ? (In the Helper.cpp/h)
Same for the maketemp, but remember this takes a long time on old raspberry systems, a database could be 15mb for example

gizmocuz:
Thanks!

PR #980, opened 2016-11-13
[MySensors] Add MQTT gateway device

xs4free:
Instead of using the default [MQTT device](https://www.domoticz.com/wiki/MQTT) in Domoticz, I've added a specialized hardware device that supports the message coming in  (and going out) to a [MySensors MQTT gateway](https://www.mysensors.org/build/mqtt_gateway).

With the default MQTT device in Domoticz almost the same functionality can be created with a Node-RED installation that transforms each message coming out of Domoticz into a MySensors message the MySensors MQTT gateway understands. The only thing that won't work is generating ID's for the sensor-nodes. This problem is fixed with this new 'hardware device' and Node-RED is not required any more. The MQTT messages coming out of Domoticz can go straight into the MySensors MQTT gateway.

I've re-used all the standard code that was already there for MySensors, and inherited from the existing MQTT code. I've tested my code with a ESP8266 running MySensors gateway 2.0 and 2 MySensors nodes running [Humidity](https://www.mysensors.org/build/humidity) and [Relay](https://www.mysensors.org/build/relay).

All test and builds were done on Windows. I've tried to update the makefile as best I can, but haven't tested a build for a different OS.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -287,6 +287,7 @@ hardware/MultiFun.cpp
 hardware/MySensorsBase.cpp
 hardware/MySensorsSerial.cpp
 hardware/MySensorsTCP.cpp
+hardware/MySensorsMQTT.cpp
 hardware/NefitEasy.cpp
 hardware/Nest.cpp
 hardware/Netatmo.cpp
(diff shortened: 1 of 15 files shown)


gizmocuz:
Thank you for the hard work !, Could you solve the compile error travis is having ? (click below on 'details' next to travis)

xs4free:
@gizmocuz The compile error is fixed. I used an enum value the wrong way (not according to Visual Studio, but according to GCC). I would love to see my PR merged, so if you have any question or comments, let me know!

gizmocuz:
Thanks!

PR #981, opened 2016-11-14
Minor changes

corbinmunce:


corbinmunce:
Not quite ready yet - I've changed to AsyncTCP rather than polling, but have a problem of when new Yeelights are added, the Yeelight Hardware must be restarted to get the new lights to connect.  I can't find the event of when new lights are added (made active).

gizmocuz:
I think this class is for TCP not for UDP... but the current implementation worked?
For working of the TCP class, see the RFXCom via TCP class

corbinmunce:
I am still using UDP for discovery, the added AsyncTCP is so a connection can be maintained to each individual light, and Domoticz is then notifed of any state changes immediately.  Not absolutely necessary but I thought it would work better.  It is based it on the RFXCom TCP class, but different in that the main thread doesn't hold a single TCP connection, instead there is a child thread for each light.

PR #982, opened 2016-11-14
domoticz external automation tool support

akatopaz:
Patch set to allow for an external tool to interact with domoticz through virtual hardware/sensor without the hassle to integrate cpp modifications, yet relying on all domoticz features.

The goal of this patch was to modify a minimal number of lines to allow for a full duplex interaction of an external daemon, using only web interface (lua event script for uplink (domoticz -> exttool) and web requests for downlink (exttool -> domoticz)).


gizmocuz:
Thanks, but could you remove the boost dependency (lexical_cast, we have to do this also in other files, i know)
we do not use %llu for device index but PRIu64

gizmocuz:
I am still a little confused what you are trying todo here.. 
You can already request all parameters of a sensor via json via it's idx value

akatopaz:
Ok for the boost stuff, but couldn't find a way to do it otherwise, if you have any suggestion.

The goal of my pull request is to allow for another tool to retrieve the identifiers of created sensors while creating them to simplify the process for the external tool. The retrieving part is ok using json, but not at creation time.

The extension of the lua engine is meant to allow for easier event forwarding to the external tool when dealing with virtual devices.

You'll see more in the projet I'm about to publish: automeshion. Prototype are ongoing, we hope to reveal it early next year.

gizmocuz:
std::stringstream is an option, or plain old sprintf
Okey, i will review the patch

akatopaz:
Thanks for reviewing, moreover, as I wrote earlier, the patch has been shrinked as much as possible.

hardware/Dummy.cpp
```diff
@@ -394,10 +396,13 @@ namespace http {
...
 				root["status"] = "OK";
 				root["title"] = "CreateVirtualSensor";
+				std::stringstream ss;
+				ss << vs_idx;
+				root["idx"] = ss.str().c_str();
 			}
 			if (DeviceRowIdx != -1)
 			{
-				m_sql.safe_query("UPDATE DeviceStatus SET Name='%q', Used=1 WHERE (ID==%" PRIu64 ")", ssensorname.c_str(), DeviceRowIdx);
+				m_sql.safe_query("UPDATE DeviceStatus SET Name='%q', Options='%q', Used=1 WHERE (ID==%" PRIu64 ")", ssensorname.c_str(), soptions.c_str(), DeviceRowIdx);
```
gizmocuz: Could you explain this, as it seems only the custom sensor above needs the options
Also dangerous, sensors not needing Options, but supplied by this call, will not have Options
What are putting into this variable ? And will it not conflict with the above ?

main/WebServer.cpp
```diff
@@ -51,6 +51,8 @@
 #include "../notifications/NotificationHelper.h"
 #include "../main/LuaHandler.h"
 
+#include <boost/lexical_cast.hpp>
```
gizmocuz: Is it possible to remove this include? (boost lexical cast)

main/WebServer.cpp
```diff
@@ -7588,6 +7610,7 @@ namespace http {
 					}
 					root["result"][ii]["idx"] = sd[0];
 					root["result"][ii]["Protected"] = (iProtected != 0);
+					root["result"][ii]["Options"] = sOptions.c_str(); // OTO: output options for all sensors, overriden afterwards depending on sensor type
```
gizmocuz: Also not quite sure whats going on here, and if this is needed...

hardware/Dummy.cpp
akatopaz: Ha right, 
I was not confident about this part. The thing is that there were no way through JSON to modify the Options.
Override of the default Temperature sensor options was problematic (LA Crosse TX... by default).

I think you would prefer a separated JSON call to set the Options, right ?

You can remove this from the pull request for now. I'll setup a separate way to cope with it.

Many thanks for your review so far.

main/WebServer.cpp
akatopaz: Yeah no problem, turns out to be a residu :(

main/WebServer.cpp
akatopaz: This is the pending of the previous comment you made. This was the way to retrieve the options from the external tool. Again, I bet a separate call would be preferable ?

gizmocuz: Depends on what you want to use the 'options' for, this should not be needed as this is handled by the dummy creation function anyway.
Why is the modification of the options needed? I can not see this ?

gizmocuz: ;)

hardware/Dummy.cpp
```diff
@@ -394,6 +396,9 @@ namespace http {
 			{
 				root["status"] = "OK";
 				root["title"] = "CreateVirtualSensor";
+				std::stringstream ss;
```
gizmocuz, line 399: Looks like the ID is already available in char ID[40] ?
No need for the stringstream and include

akatopaz: Damn, that is probably my limit in C++ and my biggest concern as an embedded programmer. I expected the = operator not to perform extra action. Now that you point it out, I guess "returning" the local variable ID through the = operator won't fail after the function returns. Is that it ?

gizmocuz: Thats true ;)

akatopaz: Arg, after a test, the ID contains +82000, therefore, it is different from the virtual sensor id required during json requests afterwards. sstream could be removed, but I still can't use the ID in the sprintfed buffer as is.

gizmocuz: Hmmm your right ! Okey, i am merging your code changes... keep us posted about your super project ;)

gizmocuz:
Thanks!

PR #985, opened 2016-11-15
wunderground - Missing current_observation option

ayasystems:
When Wunderground report "Clear" domoticz is taking this option....


gizmocuz:
Thanks

PR #993, opened 2016-11-17
Handle Yeelight strip and extra commands

corbinmunce:
Changed default light detection to RGBW to handle Yeelight strip light.  Added extra command, including disco mode test


msbuild/domoticz.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 14
-VisualStudioVersion = 14.0.25123.0
+VisualStudioVersion = 14.0.25420.1
```
gizmocuz: Please do not allow/commit changes in this file

gizmocuz:
Thanks, but do you know what the lighting type the strip has?
I know it's a quick hack to turn around the rgbw and white, but would be good if it was implemented correctly

also please leave this in

```
else if (yeelightModel == "color") {
        yeelightName = "YeeLight LED (Color)";
        sType = sTypeLimitlessRGBW;     
}       

maybe the default later will be RGB .... thats why it would be good to know what the led strip says
You would expect "color", so why did this not work before ? Maybe it's something else then white/color
```

corbinmunce:
The docs say it will report as "stripe" but as I didn't have one to test to be sure.  I've adjusted the code now.

gizmocuz:
Thanks, one request, could you re-add the newline add the end of the file ?

gizmocuz:
Thanks!

PR #994, opened 2016-11-18
Fix missing line in commit 6fdf902e

gordonb3:
Missing line causes the code to schedule for next year never to be executed.

Thx to EddyK69 who pointed out there was a problem with this particular schedule type


gizmocuz:
Thanks for the quick fix!

EddyK69:
Thx! But...

Should this also be fixed for the TTYPE_YEARLY_WD? (Line 689 and further...)

gizmocuz:
Added here:
https://github.com/domoticz/domoticz/commit/d280eef7ed1bae979404a8c8eafe5befa7708106

gordonb3:
All that cut & paste turns me blind...

Thanks guys

gizmocuz:
Yes... maybe a function would be nice ;)

gordonb3:
Currently building...

Lots of files, and I haven't even changed any of them to actually use the new functions.

PR #997, opened 2016-11-19
Add DST aware time constructs

gordonb3:
Big rebuild, sorry

Does in fact not change the resulting object files except for the three named main modules, but because localtime_r.h was changed cmake wants to rebuild all files that reference this file for inclusion. I ran a quick check for `On Time`, `Yearly` and `Monthly` schedules and they all behaved correctly.

Gordon

### DIFF
--- a/hardware/SolarEdgeAPI.cpp
+++ b/hardware/SolarEdgeAPI.cpp
@@ -170,26 +170,12 @@ void SolarEdgeAPI::GetMeterDetails()
 	if (ActHourMin - 60 > sunSet)
 		return;
 
-	struct tm ltime_min5;
- 	time_t atime_min5;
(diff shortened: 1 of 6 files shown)


gordonb3:
Hold on. One of my schedules is triggering at the wrong time

gizmocuz:
Okey, let us know when you want it to be merged

gordonb3:
Sorted it out. It was just a light scene that is triggered around sunset. Don't really know why it fired at 1:00 AM this morning, as the returned ctime must have been -1 (invalid) because I accidentally fed the year 116 to it. I double checked the rest of the code and there were no other occurrences of that mistake.

gizmocuz:
Thanks!

PR #999, opened 2016-11-20
Last DST related changes

gordonb3:
I think I've got them all now.

### DIFF
--- a/hardware/TE923Tool.cpp
+++ b/hardware/TE923Tool.cpp
@@ -432,9 +432,9 @@ int CTE923Tool::get_te923_memdata( Te923DataSet_t *data )
 	} else
 		adr = data->__src;
 
-	time_t tm = mytime( NULL );
+	time_t now = mytime( NULL );
 	struct tm timeinfo;
-	localtime_r(&tm, &timeinfo);
+	localtime_r(&now, &timeinfo);
 
 	int sysyear = timeinfo.tm_year;
 	int sysmon = timeinfo.tm_mon;
@@ -457,6 +457,7 @@ int CTE923Tool::get_te923_memdata( Te923DataSet_t *data )
 	int minute = bcd2int( buf[4] );
 
 	struct tm newtime;
+/*
 	newtime.tm_year = year;
 	newtime.tm_mon  = mon - 1;
 	newtime.tm_mday = day;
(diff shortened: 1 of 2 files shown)


gizmocuz:
Thanks!

gizmocuz:
It's not working anymore...
Could you check this ? I verified it does not work, and a forum user also reported this:
http://www.domoticz.com/forum/viewtopic.php?p=105364#p105364

gordonb3:
Aaaargh.... stupid.

Same mistake of forgetting the year and month offset in the tm structs

Need a sec

gizmocuz:
corrected one, hope i did correctly because of the minute -1, this goes all right if it gets -1 ?
https://github.com/domoticz/domoticz/commit/74d514797a65d3a78a0a46790e86af1d502159ca

gordonb3:
Good catch.

"-1900 and -1" came from the original code directly loading a tm struct. 
Shows how dangerous these kind of substitutions can be.

It's hours minus 1 btw, not minutes. And yes it will be fine if it gets 
-1. The whole point of mktime() is that it corrects every bad parameter 
you throw at it. Of course the weird thing is that the year is base 
1900, but essentially you can feed it tm_year 70 and tm_sec linux 
(posix) time and it will give you the correct time for now if all the 
other values are zero. That's why feeding it a wrong dst flag will cause 
the result to be an hour off.

Gordon

gizmocuz:
Seems you quoted an email in here... cleaned up your reply
That was not true, -3600 was taken of the time_t value, which is an hour in the original code.
-1900 you added. but was not needed, and that was why it stopped working.
Makes sense, and for 99,9% of the time we need to call the function with -1900
But glad it's working again, solved also a forum issue

gordonb3:
Had to look that up. But this was my actual change @ [commit 899c974](899c974):

` 			ntime.tm_hour = atoi(actDateTimeHour.substr(11, 2).c_str());`
` 			ntime.tm_min = 0;`
` 			ntime.tm_sec = 0;`
`+			ntime.tm_hour -= 1;`
` 			atime = mktime(&ntime);`
`-			atime -= 3600; //subtract one hour`
`-			localtime_r(&atime, &ntime);`


The -1900 was already there. I just wrongfully copied it in the call to the DateTime parser that also subtracts 1900 to fill the tm_year value.

gizmocuz:
Could this be related to all the changes ?
https://www.domoticz.com/forum/viewtopic.php?f=6&t=14731

gordonb3:
Don't know. I'll have a look.

gordonb3:
Aha! That is a not well thought through construct here starting at line 523:

https://github.com/domoticz/domoticz/commit/f108c65d79622654e17b72d7b9713492322e6a48#diff-e2afb4f4e6f784356abd209e5f00699aL523

Didn't really think about it and just combined the ++ and +1 sequence into a +2. Which is of course effectively the same. Traced the original addition of these lines leading to requesting a month 13 back to commit 71c9251 from Jan 12.

Easy to fix

gizmocuz:
Great! Great find also !

PR #1000, opened 2016-11-20
Option to use liboping instead of Boost-based ICMP echo

UrbaneChimp:
I was finding the Domoticz Pinger to be unreliable (at least on Raspberry Pi) and it seems from reading the forum that others were too. I tried a few things to fix the Boost-based implementation but couldn't get it to work reliably so I added a build-time option to use liboping instead of Boost for ICMP echo handling. So far, this seems to be rock solid on my Raspberry Pi Domoticz deployment.

Use of liboping is disabled by default, since I didn't want to add a new run-time dependency to Domoticz. To build with liboping, install liboping-dev and use "-DUSE_LIBOPING=YES" when running cmake. Target systems must have liboping0 installed. While this /should/ work fine on other architectures, I've only built and tested this on Raspberry Pi.


gizmocuz:
Thanks for your effort !
Is it possible to use this library on both windows/linux ? And could we not just add the oping/liboping as source to domoticz, so it is compiled ?
This way both windows/linux users can benefit from this, and we could use it as default ?
Or maybe look at the piece of code that is working there, as 80% of the library code is comments and option parsing...
Maybe it is a small thing that we overlooked ?

UrbaneChimp:
Are there any reports of the System Alive Checker being unreliable under Windows? I would be loathe to "fix" something in the Windows build that wasn't broken in the first place. My brief trawl of the forum suggested that it was just RPi users having problems, but I didn't do an exhaustive search. If the Windows build doesn't need this then perhaps it could be selectively enabled by default when building for Linux (or even just for RPi).

If a library is well-supported in distros I tend to prefer dynamic linkage so that the dependent program benefits more quickly from security updates and bug fixes. I don't think a run-time dependency on liboping would be a problem for Linux-based Domoticz deployments, but if it is then it could always be statically linked instead. I usually avoid pulling 3rd-party source into projects because it creates maintenance headaches in my experience.

In terms of testing, I've been running a liboping-based build on my RPi for a few weeks now and the System Alive Checker has been rock solid.

gizmocuz:
I would vote for a uniform system, working on all platforms.
Your liboping might do this, but as said, would be better to have the code natively (or check what is different) then using a library

gizmocuz:
Probably not so needed, please open a new pull request with some generic platform code

PR #1001, opened 2016-11-21
Fix midnight and noon calculations

gordonb3:


gizmocuz:
Thanks for the fast patch !

gordonb3:
Lucky I almost instantly realized what must have been wrong.

Ran a full search over the tree and found no other references to tm_year that was missing the +1900. There is one instance of tm_mon that does not have the +1, but that one is in fact correct because month is supposed to be minus 1 (so officially it should read tm_mon-1+1)

PR #1004, opened 2016-11-23
fix for intermittend crash of domoticz and some refactoring of code

philippe72:
Okay, so here's the deal: I noticed that every few days my domoticz would crash on communicating with the Nefit easy server. After some debugging I found out that the cause was that I received a 'systemPressure' response to a 'uiStatus' request. They both contain a "value" key, but instead of the expected dictionary now it was a float causing NefitEasy.cpp:382 to crash.

Don't know the cause, maybe some delayed reply or a bug on Nefit side, but since the response contains the request type in the "id" key I added a check for this, so that we can 'guarantee' that the response type matches the request.

I also refactored the code to remove duplication and make it easier to read. I build and tested it on ubuntu 14.04.

### DIFF
--- a/hardware/NefitEasy.cpp
+++ b/hardware/NefitEasy.cpp
@@ -6,7 +6,6 @@
 #include "hardwaretypes.h"
 #include "../main/localtime_r.h"
 #include "../httpclient/HTTPClient.h"
-#include "../json/json.h"
 #include "../main/RFXtrx.h"
 #include "../main/mainworker.h"
 #include "../webserver/Base64.h"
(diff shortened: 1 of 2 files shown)


gizmocuz:
Thanks for your work.
Could you make a patch with only the 'id' part ?

philippe72:
I can, but I'd rather not? Are there issues with the refactored code?

gizmocuz:
Yes, i liked my code, and found it very easy, also liked my logging
I would like to merge your 'id' checks

philippe72:
Fair enough, your project, your rules. The main importance for me is getting the fix in! :-)

(I collected a lot of duplicate code in some functions: for me that is a natural thing to do but there is nothing as subjective as coding style preferences)

I will make the changes soon and resubmit it

gizmocuz:
Is this still a problem ?

gizmocuz:
If you have some updates, please open a new pull request

PR #1010, opened 2016-11-27
Changing hardcoded colors

Michmis:
Changing item status colors from hardcoded to css classess. in dashboard Added classess for dashboard item types.

### DIFF
--- a/www/app/DashboardController.js
+++ b/www/app/DashboardController.js
@@ -5,21 +5,21 @@ define(['app'], function (app) {
 		
 		//Evohome...
 		//FIXME move evohome functions to a shared js ...see temperaturecontroller.js and lightscontroller.js
-		
+
 		SwitchModal= function(idx, name, status, refreshfunction)
 		{
 			clearInterval($.myglobals.refreshTimer);
-			
+
 			ShowNotify($.t('Setting Evohome ') + ' ' + $.t(name));
-			
+
 			//FIXME avoid conflicts when setting a new status while reading the status from the web gateway at the same time
 			//(the status can flick back to the previous status after an update)...now implemented with script side lockout
 			$.ajax({
-			url: "json.htm?type=command&param=switchmodal" + 
-						"&idx=" + idx + 
+			url: "json.htm?type=command&param=switchmodal" +
+						"&idx=" + idx +
 						"&status=" + status +
 						"&action=1",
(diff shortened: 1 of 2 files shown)


gizmocuz:
Sorry i can not see what you changed, because you also changed the formatting of other things
could you only change the relevant lines ?

Michmis:
@gizmocuz sorry, thats my first commit ever, I will try to write it again, it seems that I deleted some spaces and it looks messy

Michmis:
@gizmocuz  I tried to edit it again on my PC using GitHub desktop and Atom editor, but it still looks messy, there are blank lines shown as changed etc. I will describe my changes

There are items names in dashboard and for different item states it has different background color. For me hardcoding colors is wrong when there is possibility to use Domoticz themes. On the other hand there is no CSS class for item div to describe its state, so i wanted to change this parts:

`	var nbackcolor="#D4E1EE";
	if (item.HaveTimeout==true) {
		nbackcolor="#DF2D3A";
	}
	else if (item.Protected==true) {
		nbackcolor="#A4B1EE";
	}`

which affect td#name directly to:

`	var itemstatus="ok";
	if (item.HaveTimeout==true) {
		itemstatus="timeout";
	}
	else if (item.Protected==true) {
		itemstatus="protected;
	}`

and move it back before DIV of an item is created like:
`xhtm='\t<div class="span3 movable itemtype_temp itemstatus_' + itemstatus +'" id="temp_' + item.idx +'">\n';`

So whole DIV has now 2 new CSS classes, one for item type and one for item status, like .itemstatus_ok or itemstatus_timeout.

Then in style.css file we can make 
`div.itemstatus_timeout table tr td#name {background: #DF2D3A;}`

I am very sorry that I make it that messy, I believe that would be huge improvement for all theme designers.

gizmocuz:
You are free to make a new pull request for the formatting ;)

gizmocuz:
Try notepad2-mod as editor

Michmis:
Thank you, I will try it tommorow and check formatting again

gizmocuz:
Also note the colours are used in all tabs, not only the dashboard

gizmocuz:
Is this still a work in progress ?

gizmocuz:
Please open a new pull request when you have some updates

PR #1011, opened 2016-11-27
Compare libboost version found by CMake and compare to min. required …

ikkemaniac:
…version. Quit CMake if version is to low.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,7 +4,7 @@ cmake_minimum_required(VERSION 2.8.4)
 project(domoticz)
 
 ## required min. libBoost version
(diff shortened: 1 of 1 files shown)


ikkemaniac:
I see travis-ci fails, the log shows it's building against libBoost 1.55.0. I assumed 1.62 is required because of the wiki. If 1.55.0 is the minimum required version the  DOMO_MIN_LIBBOOST_VERSION variable needs to be adjusted accordingly

gizmocuz:
The wiki is always up to date... But yes please change it to 1.55.0

gizmocuz:
Thanks!

PR #1013, opened 2016-11-28
Update Sterbox.cpp

houk001:
After some changes there is no html tags and <head> in page returned to parsing. Getting data from device to domoticz stop working.
This patch resolve problem
Also changed frequency of getting data from device to 8 seconds.

### DIFF
--- a/hardware/Sterbox.cpp
+++ b/hardware/Sterbox.cpp
@@ -13,7 +13,7 @@
 
 extern CSQLHelper m_sql;
 
-#define STERBOX_POLL_INTERVAL 10
+#define STERBOX_POLL_INTERVAL 8
 
 CSterbox::CSterbox(const int ID, const std::string &IPAddress, const unsigned short usIPPort, const std::string &username, const std::string &password) :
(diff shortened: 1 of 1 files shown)


hardware/Sterbox.cpp
```diff
@@ -13,7 +13,7 @@
 
 extern CSQLHelper m_sql;
 
-#define STERBOX_POLL_INTERVAL 10
+#define STERBOX_POLL_INTERVAL 8
```
gizmocuz, line 16: Why 8?

gizmocuz: There is a conflict somewhere

houk001: I only remove 4 lines and change from 10 to 8. Nothing more. 
Why 8? Because i have movement sensor connected to this device and if 10 seconds then sometimes miss movement signals. With 8 is OK :D

gizmocuz: Okey thanks for the explanation
I can not merge this, because it has conflicts, i might have changed a line to solve a compiler warning

PR #1014, opened 2016-11-30
Handle meter/values for Rain and Wind

wroopy:
For wind meters the RawDataEvent is used. The parsing is based on the same logic as telldus-core is parsing the message.  

For rain meters the value for RAINRATE is used.

I've been running this code for the last week to get values from a Oregon WMR200.


gizmocuz:
Thanks!

PR #1018, opened 2016-12-02
Fixed udp socket port conflict

corbinmunce:
Fixed udp socket port conflict
Discard duplicate udp responses in yeelight discovery
Added Limitless_SetBrightnessLevel to update message so light level is updated


gizmocuz:
Thanks

Issue #1017, opened 2016-12-02
Virt counter graph fails


ikkemaniac:
I'm having trouble with one (of several) virtual counters: the graph won't update.

- domoticz 3.5989 compiled on *nix
- domoticz started with `-nocache` flag
- firefox cache disabled

![tn_screenshot from 2016-12-02 00 12 12](https://cloud.githubusercontent.com/assets/845292/20818643/4733b8de-b82f-11e6-9440-d2ccd07fb13a.png)

There is correct data in the db:
> sqlite> select * from Meter where DeviceRowID=18 limit 1000;
> DeviceRowID|Value|Usage|Date
> 18|15952|0|2016-12-01 00:10:00
> 18|15952|0|2016-12-01 00:15:01
> 18|15952|0|2016-12-01 00:20:00
> 18|15952|0|2016-12-01 00:25:00
> ....
> ....
> 18|15952|0|2016-12-01 23:40:00
> 18|15952|0|2016-12-01 23:45:01
> 18|15952|0|2016-12-01 23:50:00
> 18|15952|0|2016-12-01 23:55:00
> 18|15952|0|2016-12-02
> 18|15952|0|2016-12-02 00:00:01
> 18|15952|0|2016-12-02 00:05:00
> 18|15952|0|2016-12-02 00:10:00
> sqlite> 

I don't understand though why there are two records for 'midnight':
> DeviceRowID|Value|Usage|Date
> 18|15952|0|2016-12-02
> 18|15952|0|2016-12-02 00:00:01

But it is working for my other virtual counters
![tn_screenshot from 2016-12-02 00 18 45](https://cloud.githubusercontent.com/assets/845292/20818657/5a4b1af2-b82f-11e6-8f81-3d5d2c5a2adf.png)

> DeviceRowID|Value|Usage|Date
> 16|412835319|0|2016-12-02
> 16|412837901|0|2016-12-02 00:00:01

I'm still investigating but it looks like not all data is passed 'from the db' to `CWebServer::RType_HandleGraph`.

ps. Any best practices on debugging? It's a realy pita right now because
- `CMAKE_BUILD_TYPE=Debug` doesn't seem to be implemented
- `-loglevel` flag does not provide debugging info (doesn't really seem to do much?)
- `-verbose` flag shows more info about device info received but doesn't seem to do a lot more?

I'm just inserting `_log.Log(LOG_STATUS,....` trying to track stuff....

ikkemaniac:
[...]

Issue #958, opened 2016-11-04
Graph not updated when no data


buxol:
In this example the gas consumption graph is not updated this morning while having no consumption last night. When I consume the graph is updated correctly, but in the screenshot I am 'missing' the 4-5 hours of zero consumption at 7.13u
I've created this issue as requested in this topic: http://www.domoticz.com/forum/viewtopic.php?f=6&t=14054

Kind regards,
Roel

![img_0553](https://cloud.githubusercontent.com/assets/23262090/20008966/e832d29e-a2a2-11e6-986e-162cf3fbbb82.PNG)

gizmocuz:
Thanks for reporting!

gizmocuz:
fixed in #1033

PR #1019, opened 2016-12-02
Remove condition to have delta in data which prevented 'zeros' to pop…

ikkemaniac:
…ulate counter bar graph. Fixes domoticz/domoticz#958 and fixes domoticz/domoticz#1017

### DIFF
--- a/main/WebServer.cpp
+++ b/main/WebServer.cpp
@@ -12693,7 +12693,9 @@ namespace http {
 
 											float TotalValue = float(actValue - ulFirstValue);
 
-											if (TotalValue != 0)
+											//~ BETA: fix bar graph not showing zeros; issues #958, #1017
+											//~ if (TotalValue != 0)
+											if (1 == 1)
 											{
 												switch (metertype)
 												{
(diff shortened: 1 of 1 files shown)


gizmocuz:
Can you make a patch where you just 'remark' the checking for 0 ?

because it was obviously there for a reason, willing to make a new beta without it, but do not delete the code yet

gizmocuz:
with other words just put // infront of the line

gizmocuz:
I am closing this pull request, it gets not readably anymore, this time i see if (1==1) stuff in it, some beta comments, some formatting changes, i don't know what you want anymore
If i understand correctly, you only had to comment the above mentioned 'if' line right ?
If you want this, just make a comment on that line, leave the rest of the code alone, or wait until you have your final patch, and then make a pull request please

ikkemaniac:
You asked me to revert the change and formatting i which i did by means of the 'revert' commit, no formatting changes persist.
Commenting out the condition would leave an unused group in the source, adding `if(1 == 1)` keeps it readable and understandable while it accomplishes the same as commenting out the original line. 
I added a comment stating this is a beta modification to keep the code readable and left the original condition in there.
I don't understand your coding rules or guidelines and find it makes it difficult to go through the code.

But I’ve pushed a commit as you requested nonetheless.

gizmocuz:
My coding rules:
- Only change what needs to be changed.
Now your patch looks much better, except for the extra empty line on 12762
Before merging this pull request, it needs to be tested good, it might work in your situation, but these three checks i added for a reason...

PR #1020, opened 2016-12-02
Fixed: Evohome HGI80 Custom Relay control

Evsdd:
Hi,

Fixed a problem that I introduced when making heat demand switches read-only which prevented custom relays from being controlled.

Dan


gizmocuz:
Thanks for the fix!

PR #1023, opened 2016-12-06
Added Evohome HGI80 HR80 battery info and fixed 0 demand bug 

Evsdd:
Added support for battery info for HR80 device. Standardised battery levels to 0-100 across Evohome devices.  Fixed bug which caused heat demand switch devices demand to display Off - 0 following a battery_level message.


gizmocuz:
Thanks!

PR #1024, opened 2016-12-06
implemented new dummy hardware "regulator"

ghost:
Dear Gizmocuz,

I have implemented a new "dummy" hardware to regulate temperature using an association of 3 devices (temperature, thermostat setpoint and switch)
I wanted this to be an alternative to using LUA scripts
The first attempt was to add a new virtual sensor type (on the Dummy device), but is was simpler for me to create a new hardware

Overview & Usage :
- first, create a new "Regulator" device (entry has been added in the hardware "type" listbox)
- it is added to the hardware list, with a "associate devices" button (this button is displayed the same way as the "create virtual sensors" on Dummy hardware)
- in the next dialog, you can select an existing temperature device, a thermostat setpoint and a switch
- when the regulator hardware starts, it compares the temperature to the setpoint and decides to turn the switch On or Off

Using the "timers" capability of the Thermostat setpoint, you can set a target temperature of 18° the night and 21° the day. The regulator will drive the heater switch automatically. Since the Setpoint value can be set by the mobile application, it is very simple to change the temperature of a room at your convenience (until the next timer)

Note & Limitations : 
- new database table "Regulators" to store the associations between the 3 devices (db version incremented to 107)
- regulation is very simple, using hystereris of 0,5° (hardcoded)
- the Regulator cannot detect changes and does not restart automatically when a device is deleted, or when a new association between 3 devices is added
- polling interval is hardcoded to 30 seconds
- no checks are done when adding new associations
- no translation at this time
- the terms "Regulator", "Device" ... may be poorly chosen ! (english is not my native language)

I hope this pull request will catch your attention

Christian

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,5 @@ domocookie.txt
 domoticz.exe
 /Config/zwcfg_*.xml
 /makebetarob
+/www/images/Regulator.png
+/www/images/Regulator48.png
--- a/.gitignore
+++ b/.gitignore
@@ -29,5 +29,4 @@ domocookie.txt
 domoticz.exe
 /Config/zwcfg_*.xml
 /makebetarob
-/www/images/Regulator.png
-/www/images/Regulator48.png
+
--- a/hardware/RegDevice.cpp
+++ b/hardware/RegDevice.cpp
@@ -1,0 +1,23 @@
+#include "stdafx.h"
+#include "RegDevice.h"
+
+
+CRegDevice::CRegDevice()
+{
+	m_iSetpointDevID = 0;
+	m_iSwitchDevID = 0;
+	m_iTempDevID = 0;
+}
+
+
+CRegDevice::~CRegDevice()
+{
+}
+
+CRegDevice::CRegDevice(int temp, int setpoint, int switchid)
+{
+	m_iSetpointDevID = temp;
+	m_iSwitchDevID = setpoint;
+	m_iTempDevID = switchid;
+}
+
--- a/hardware/RegDevice.h
+++ b/hardware/RegDevice.h
@@ -1,0 +1,14 @@
+#pragma once
+class CRegDevice
+{
+public:
+	int m_iTempDevID;
+	int m_iSetpointDevID;
+	int m_iSwitchDevID;
+
+public:
+	CRegDevice();
+	CRegDevice(int temp, int setpoint, int switchid);
+	~CRegDevice();
+};
+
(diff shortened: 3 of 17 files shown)


.gitignore
```diff
@@ -29,3 +29,4 @@ domocookie.txt
 domoticz.exe
 /Config/zwcfg_*.xml
 /makebetarob
+
```
gizmocuz, line 32: please remove this line

hardware/RegDevice.cpp
```diff
@@ -0,0 +1,23 @@
+#include "stdafx.h"
```
gizmocuz, line 1: Could you make this a (private) struct of the Regulator class ? Saves two extra files, and this could also been a 'RegisterDevice' class looking at the name, i would name it RegulatorDevice

hardware/RegDevice.h
```diff
@@ -0,0 +1,14 @@
...
+class CRegDevice
+{
+public:
+	int m_iTempDevID;
+	int m_iSetpointDevID;
+	int m_iSwitchDevID;
+
+public:
+	CRegDevice();
+	CRegDevice(int temp, int setpoint, int switchid);
```
gizmocuz, line 11: CRegDevice(const int temp, const int setpoint, const int switchid);

main/RFXNames.cpp
```diff
@@ -224,6 +224,7 @@ const char *Hardware_Type_Desc(int hType)
 		{ HTYPE_AtagOne, "Atag One Thermostat" },
 		{ HTYPE_Sterbox, "Sterbox v2-3 PLC with LAN interface" },
 		{ HTYPE_HTTPPOLLER, "HTTP/HTTPS poller" },
+		{ HTYPE_REGULATOR, "Regulator" },
```
gizmocuz, line 227: As far as i know this is in order of HTYPE and should be at the bottom

gizmocuz:
Question, i do not see why you could not use script/blockly for this ? blockly/lua is exactly meant for this

gizmocuz:
(Thanks for the hard work bdw!)

ghost:
Thanks a lot for your comments
I will change the code following your recommendations.

About the reason : I found lot of excellent scripts to do temperature regulation, but they require modifications (change the devices id or name, etc...) and this can be a problem for "basic-users" like me. Just a few mouse clicks and the Regulator class do the job
I am aware that this work is not complete and has limitations, but it may be a good start to implement PID regulation, or handle an outdoor temperature probe.

Whatever you decide about this pull request, it was a pleasure to submit my first one here :) (and a great pain to move from Team Foundation to Github !)

gizmocuz:
I dont think so.
If you make a lua script, and you define the sensor idx's at the top of your lua in a comment, a good explanation/help at the top, it should be very easy
I still would like to see this as a script as it is not real hardware you added to the system but rather a kind script...

ghost:
No problem, it was interesting to discover Domoticz internals :)

gizmocuz:
Thanks for your understanding !

Issue #832, opened 2016-08-29
Notification for errors


gizmocuz:
There is need to be notified when there is a error added to the log.

As there could be multiple errors per minute, i will make a system that collects them, and send them in batches every 5 minutes.
When there is a new error received, and there had been no errors send in the last 5 minutes, it will be send straight away.

First implementation will use the 'email' system only

gizmocuz:
Seems this is done ? (august 29)

PR #1027, opened 2016-12-08
Changed topic of mail which contains error log

gerard33:
Notification for errors #832
Changed topic of mail which contains error log to "Domoticz: Error received". The topic of the mail contained the complete error, just as in the body of the mail. Formatting is now the same as for mails which contains errors received in the last 5 minutes.

### DIFF
--- a/main/mainworker.cpp
+++ b/main/mainworker.cpp
@@ -12076,7 +12076,8 @@ void MainWorker::HandleLogNotifications()
 	else
 	{
 		itt = _loglines.begin();
-		sTopic = "Domoticz: " + itt->logmessage;
+		sTopic = "Domoticz: Error received";
+		sstr << "The following error is received:<br><br>";
 	}
 
 	for (itt = _loglines.begin(); itt != _loglines.end(); ++itt)


main/mainworker.cpp
```diff
@@ -12076,7 +12076,8 @@ void MainWorker::HandleLogNotifications()
 	else
 	{
 		itt = _loglines.begin();
-		sTopic = "Domoticz: " + itt->logmessage;
+		sTopic = "Domoticz: Error received";
```
gizmocuz, line 12079: i only read topics, thats not working now ... everything is Error received ?

gerard33: Yeah if you only read the topics, then this is not a handy change ;-)

No need to change it of course, but the topic is much to long to read in the inbox, that's why I proposed this change.
And when opening the mail (when topic is fully visible), you see the same message twice (topic and body).

And your change #832 is only sending notifications for errors, so indeed everything is Error received.

gizmocuz: they i rather not merge this request
if there are multiple errors in a minute, you will get another header and a body that contains all errors
I dont have a problem with the headers, using a normal 1920x1080 screen, not 2k or 4k

gerard33: no problem

PR #1029, opened 2016-12-09
Fix crash in Monthly weekday schedule

gordonb3:


gordonb3:
Sorry: wrong title on commit

PR #1030, opened 2016-12-09
Fix crash in Monthly weekday schedule

gordonb3:


gizmocuz:
Thanks for the quick fix!

gordonb3:
No worries. When you asked me, I knew I had seen that reference to boost in the crash log before. It didn't take too long to figure out there was something fishy about that particular section.

PR #1031, opened 2016-12-10
Python plugins v1.0

dnpwwo:
Working Python Plugin framework that supports TCP/IP connectivity


gizmocuz:
Great! It's it possible to resolve the conflicts?

dnpwwo:
Done and all checks passed

.travis.yml
```diff
@@ -16,6 +16,7 @@ sources:
 
 # What to install before the build
 before_install:
+  - cmake --version
```
gizmocuz, line 19: I think this debug line should be removed?

main/RFXNames.h
```diff
@@ -173,8 +173,9 @@ enum _eHardwareTypes {
 	HTYPE_ZIBLUETCP,			//90
 	HTYPE_Yeelight,				//91
 	HTYPE_MySensorsMQTT,		//92
-	HTYPE_RaspberryPCF8574,				//93
-	HTYPE_XiaomiGateway,		//94
+	HTYPE_RaspberryPCF8574,		//93
+	HTYPE_PythonPlugin,			//94
+	HTYPE_XiaomiGateway,		//95
 	HTYPE_END
```
gizmocuz, line 179: I think you flipped the two hardware types
The python should be at the bottom correct?

dnpwwo: I flipped them on purpose.

The PythonPlugin number is used in the HardwareControll.js but the XiaomiGateway number was not.  By flipping them I didn't need to update HardwareController.  As far as I could tell the XiaomiGateway only uses the constant so doesn't care what number it is.

.travis.yml
dnpwwo: It certainly could be removed.  It was useful because different version of cmake have different handling for Python3 so I needed to know what was running.

I'll remove it in the next update if you like, not fussed now it builds.

main/RFXNames.h
gizmocuz: Except for the people already using it, and the sensors are linked against the hardware_id
But lets leave it like this because only a very limited amount of people are using the xiaomi gateway as it's very new

PR #1032, opened 2016-12-10
Added { HTYPE_XiaomiGateway, "Xiaomi Gateway" },

corbinmunce:
added Xiaomi Gateway to RFXNames.cpp


corbinmunce:
Thank you, I've updated my main Domoticz installation, and the Xiaomi Gateway is working

PR #1036, opened 2016-12-13
Fixed: Evohome HGI80 zone temperatures corrected to 2 decimal places

Evsdd:
Hi,
The raw zone temperatures captured via the Evohome HGI80 device were being rounded to 1 decimal place before being written to the database.  I've now corrected this to retain the 2 decimal places which matches their original accuracy.  This also maintains consistency with the Evohome script set-up which now captures the information to 2 decimal places following a recent change in the API.


gizmocuz:
Thanks, but one decimal is perfect for a temp sensor ;)

PR #1037, opened 2016-12-14
Added delimiter which is required by the new HEOS Firmware

Sneezydevil:
A recent HEOS update breaks HEOS support in Domotcz, this should fix that.

### DIFF
--- a/hardware/HEOS.cpp
+++ b/hardware/HEOS.cpp
@@ -706,11 +706,18 @@ bool CHEOS::WriteInt(const unsigned char *pData, const unsigned char Len)
 
 bool CHEOS::WriteInt(const std::string &sendStr)
 {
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thanks, rather than this way to fix it, would it be not better that you concat the \n\r at the moment you are sending ? So your messages still be clear, and only when ssMessage is actually send, you add the CRLF ?

Sneezydevil:
You are right, I was so happy I found the problem that I did't really take a nice approach at fixing it.

I will change it and update the pull request

Sneezydevil:
I don't do alot of c++ so if I do silly stuff like that just call me out on it. I think this is cleaner.

gizmocuz:
Thanks!

PR #1040, opened 2016-12-15
Correct overseen old tUntil value from commit 6333ac5

gordonb3:
This patch fixes the Time Until value in the Edit Set Point dialogue to wrongfully display the past midnight time due to a non existing variable to be passed to the function.


gizmocuz:
Thanks

PR #1043, opened 2016-12-16
OpenWebNet support for unofficial gateway (without format changes)

Spugna85:
- Support connection to unofficial gateway via "nonce-hash"
authentication (Tested with MH202 and myhome screen 35 LN4890)
- Fix system crash on StopHardware
- Fix filling of fields in the 'hardware' page and add 'passwords'


gizmocuz:
Thanks!

PR #1044, opened 2016-12-16
Python plugins

dnpwwo:
Added support for multiple versions of Python.
For Linux it now recursively  searchs /usr/lib for compatible library


gizmocuz:
Thanks!

PR #1045, opened 2016-12-16
Update Sterbox

houk001:
Remove code for validation device data for html tag title and change
reading frequency to 8 seconds


gizmocuz:
Thanks!

PR #1046, opened 2016-12-17
Added a new parameter to disable the logging of event script triggers

YcKe:
Added a new parameter to disable the logging of event script triggers
'Script event triggered: ...'


YcKe:
It seems that I'm not able to Resolve the conflict. When I click the Resolve conflicts button it gets into an infinite loop.
Does this has to do with some kind of permission setting?

houk001:
I have similar problem couple days ago. Resolved by deleting fork and create it again, made changes and pull.

YcKe:
Will create new pull request as suggested above to solve the conflict.

PR #1054, opened 2016-12-18
Fix for intermittend crash issue

philippe72:
Took some time to finish... Holidays, work, getting domoticz to build again now that I needed boost 1.55

But this is a minimalistic version of the fix


gizmocuz:
Thanks!

PR #1061, opened 2016-12-20
Update MQTT.cpp

jonferreira:
adds setcolbrightnessvalue support

### DIFF
--- a/hardware/MQTT.cpp
+++ b/hardware/MQTT.cpp
@@ -321,11 +321,6 @@ void MQTT::on_message(const struct mosquitto_message *message)
 		if (!m_mainworker.SwitchLight(idx, switchcmd, level, hue, isWhite, 0) == true)
 		{
 			_log.Log(LOG_ERROR, "MQTT: Error sending switch command!");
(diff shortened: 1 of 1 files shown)


hardware/MQTT.cpp
```diff
@@ -278,6 +278,57 @@ void MQTT::on_message(const struct mosquitto_message *message)
...
+		{
+			if (root["isWhite"].isString())
+				isWhite = atoi(root["isWhite"].asString().c_str());
+			else
+				isWhite = root["isWhite"].asInt();
+		}
+		
+		if (!m_mainworker.SwitchLight(idx, switchcmd, level, hue, isWhite, 0) == true)
+		{
+			_log.Log(LOG_ERROR, "MQTT: Error sending switch command!");
```
gizmocuz, line 323: can you make a single log line for this ? ;)

jonferreira: actually those "extra" log lines where supposed to have been removed, they where only there for debug :-)
done now

gizmocuz:
Thanks!

PR #1062, opened 2016-12-20
Debounce GPIO reading

jvandenbroek:
I had troubles with missing switch updates, fixed by setting a 50ms debounce sleeper. This should be short enough for most, if not all people (and I doubt that Domoticz would handle a higher frequency reliably anyway).


gizmocuz:
Thanks!

PR #1074, opened 2016-12-27
Gateway token support added so devices can be controlled

corbinmunce:
Added password, ip, and port settings when adding the Xiaomi Gateway.
Added encryption based on password, so device can be queried, and smart plugs controlled.
Battery Level and Temp/Humidity device support has been started but is not complete.


gizmocuz:
Thanks!

PR #1086, opened 2016-12-29
Arilux/Flux LED - Fix White/RGB and Brightnesssend command to controller

ducseb:
Store hue value in the class, command is now sent by the Limitless_SetBrightnessLevel.
Fix White/RGB command sending
Fix mistakes in TCP protocol
Class header and comment modifications


gizmocuz:
Thanks!

gizmocuz:
Could you have a look here:
https://github.com/domoticz/domoticz/issues/1253

PR #1088, opened 2016-12-29
Avoid device without sValues

jotakar:
I was programming a script for an event and I saw that some devices had none sValues. 
With this modification I try to allow that event scripts can read sValues for any triggering (or not) devices
SQLHelper is the affected file.


gizmocuz (commented):
Nope.. This is not the way
Try looking in the event clasx

jotakar:
Why? otherdevices need also an svalue...
Well OK, but what is event clasx?

gizmocuz:
No, not every device needs an sValue
EventSystem.cpp

jotakar:
Thanks, I'll study EventSystem.cpp 
But althoug every device doesn't need an sValue, in this case I think thermostate Mode (or Fan mode and others)  needs this sValue for event management, because from script we cannot access nValue.  
And I wonder if this modification (use sValue en this device) has any kind of side effect.
Or why it's not a good solution.
Bye and happy new year :)

PR #1094, opened 2017-01-02
Changed authorization from url to basic authentication header

eburud:
Changed authorization from url to basic authentication header in DataPush FibaroLink. Username and password is no longer supported by Fibaro API. These changes are tested on Fibaro HC2 controller with software version 4.104 Beta (which is the latest version per 2. january 2017).


gizmocuz:
Thanks!

PR #1095, opened 2017-01-02
Python plugins

dnpwwo:
OSX build fix for Python 3.6
Add search directory for Python from source
Add Device.Delete() function


gizmocuz:
Thanks!

PR #1096, opened 2017-01-02
Python plugins

dnpwwo:
Don't merge this change it make take some debugging.
Moving Linus build from Precise to Trusty so that the Python version is 3.4


dnpwwo:
This worked better than I expected 1st time.
Linux 14.04 and OSX builds now work

gizmocuz:
Great! I also witnessed this ;)

PR #1100, opened 2017-01-03
Make action script path handling consistent

gordonb3:


gizmocuz:
Thats not a good patch

1. First check if the script has a larger size then 9
2. some people enter scripts://test.sh
3. some people enter scripts:///home/pi/scripts/test.sh
4. some people enter scripts://C:\Scripts\test.bat

gordonb3:
You said so before, but this is the exact code from SQLHelper. As said: this patch is about consistency so that paths handed by evohome.cpp (evohome temperature devices) come out the same as paths handled by SQLHelper.cpp (including evohome controller).

1. This particular part of the code is inside an ` if ` clause that already enforces that the input starts with ` script:// `
i.e. from this we already know that it is at least 9 characters long
1. In SQLHelper this gets translated to ` <domoticzroot>/scripts/<scriptname> `.
This is the consistency bit - evohome.cpp should not translate this into ` /<scriptname> `
1. That is the correct way to reference a script with an absolute path - again according to existing code in SQLHelper
1. Current code in evohome.cpp translates this as ` /C:\Scripts\test.bat ` (invalid) while in SQLHelper this returns ` C:\Scripts\test.bat `. Both will likely fail btw because the ` \ ` will no doubt be handled as an escape by the file system functions (file_exist, stat, ShellExecute).

gizmocuz:
I am still not sure what this has to do with evohome, this is general ?

gordonb3:
No, this is restricted to evohome as this particular hardware handler contains alternate code for calling an Action script if the device is a heating device. This is due to special parameter substitution that is not part of the generic code in SQLHelper. In other words: for these particular devices the action path is treated different from any other action path for any other type of device. Which can be avoided on Linux by using the one path syntax that returns the same result (`#3`) but not on Windows.

gizmocuz:
Thanks!

PR #1110, opened 2017-01-09
bind to ipv4 and ipv6 sockets if net.inet6.ip6.v6only is set to 1

samm-git:
I am using domoticz on the FreeBSD 11. It was found that with default settings it binds only to ipv6 socket. Netstat shows that domoticz is listening only on ipv6 address. It is possible to force using v4 address, but in this case i will loose ipv6 connectivity. So i looked at the source and found that:

1. In FreeBSD there is a sysctl variable `net.inet6.ip6.v6only` and it is set to 1. This disabling support for the "mapped v4 addresses" globally, so v4 connections will not be accepted on v6 sockets.  
```
root@rpi-b:# netstat -an|grep 8080|grep LIST
tcp6       0      0 *.8080                 *.*                    LISTEN
```
2. It is possible to enable this globally (not recommended, could be security risk) or per socket (recommended). 
3. I tried to so this using boost and it works just fine. After this patch netstat shows:
```
root@rpi-b:# netstat -an|grep 8080|grep LIST
tcp46       0      0 *.8080                 *.*                    LISTEN
```
and domoticz accepting connections on the both ports

### DIFF
--- a/webserver/server.cpp
+++ b/webserver/server.cpp
@@ -41,8 +41,10 @@ void server_base::init(init_connectionhandler_func init_connection_handler, acce
 	boost::asio::ip::tcp::endpoint endpoint = *resolver.resolve(query);
 	acceptor_.open(endpoint.protocol());
 	acceptor_.set_option(boost::asio::ip::tcp::acceptor::reuse_address(true));
-	// bind to both ipv6 and ipv4 sockets
-	acceptor_.set_option(boost::asio::ip::v6_only(false));
+	// bind to both ipv6 and ipv4 sockets for the "::" address only
+	if (settings_.listening_address == "::") {
+		acceptor_.set_option(boost::asio::ip::v6_only(false));
+	}
 	// bind to our port
 	acceptor_.bind(endpoint);
 	// listen for incoming requests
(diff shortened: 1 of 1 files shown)


rkluwen:
Isn't the setting net.inet6.ip6.v6only not there because you will want this behaviour? I.e. not binding to ipv4 sockets?
You can make this patch to Domoticz. But at the same time you would need to patch all other software that runs on your box as well. Or?

samm-git:
Let me explain. I want domoticz to listen on ipv6 and ipv4 at the same time. In the FreeBSD as well as in some other distributions net.inet6.ip6.v6only is set to on, so any program listening on `::` address will also accept v4 connections. When i am turning v6_only to false i can accept connection on both protocols, in other case - v6 only. Because domoticz do not allow to specify > 1 listen address - its only way to achieve that. Some of the programs (e.g. nginx) allows to configure v6_only socket value in the config file, so no, i dont have to patch anything else.

rkluwen:
The option ipv6only is in there for a reason. Even if you don't agree with it. But in that case, you can change it.
But it's a small patch just to override that option. And it will help people that want to install Domoticz and don't know about that option. So +1 from me, in this case.

samm-git:
No, i cant change it and this is a problem. Because sysctl changes this **globally** and not per socket. And it is **not** recommended in this sysctl description. Some software, e.g. nginx allows to configure it in the "listen" option.

samm-git:
and this patch is not overriding global option - it sets one only for the domoticz socket.

gizmocuz:
Thanks both !

PR #1111, opened 2017-01-10
openwebnet : new features and a correction

steplebr:
hardware : add openwebnet serial gateway
switch type : add zwave blinds and lights
correction of a bug in the parser


gizmocuz:
Thanks! Great work !!

gizmocuz:
Is it possible you can rename the previous one and add 'with LAN interface' like all other hardware with a lan interface?

PR #1112, opened 2017-01-12
Allow empty UserVariables for string values

YcKe:
Allow empty UserVariables for string values

### DIFF
--- a/main/WebServer.cpp
+++ b/main/WebServer.cpp
@@ -1691,7 +1691,11 @@ namespace http {
 			std::string variablename = request::findValue(&req, "vname");
 			std::string variablevalue = request::findValue(&req, "vvalue");
 			std::string variabletype = request::findValue(&req, "vtype");
-			if ((variablename == "") || (variabletype == "") || (variablevalue == "" && variabletype != "2"))
+			if (
+				(variablename == "") || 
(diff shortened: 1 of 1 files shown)


main/WebServer.cpp
```diff
@@ -1691,7 +1691,7 @@ namespace http {
 			std::string variablename = request::findValue(&req, "vname");
 			std::string variablevalue = request::findValue(&req, "vvalue");
 			std::string variabletype = request::findValue(&req, "vtype");
-			if ((variablename == "") || (variablevalue == "") || (variabletype == ""))
+			if ((variablename == "") || (variabletype == "") || (variablevalue == "" && variabletype != "2"))
```
gizmocuz: Thanks!

To keep it readable (also the below patch) could you change to code like:

if (
----(variablename == "") || 
----(variabletype == "") || 
----( (variablevalue == "") && (variabletype != "2") )
) {
   .....
}

YcKe:
Sure, you want me also to replace the ".empty"with"== 0" to have it consistent in both functions?

gizmocuz:
No, just to write all checks on different lines
.empty() is not the same as '0', looks ok like it is now ;)

gizmocuz:
Thanks!

PR #1114, opened 2017-01-12
Adding GPIO pins of Raspberry B+, 2, 3 and Zero

ogroot:
Domoticz currently support only GPIO pins of Raspberry Pi A 26-pin header. 
Raspberry B+, 2, 3 and Zero have 40 pin header with 7 extra GPIO in/outputs.

The following 7 pins were added: 5,6,12,13,16,19,26.
Change has been running without problems on on Raspberry Pi2B (> 6 months).

Only concern might be that this causes issues on older Pi A (do they still use these with Domoticz??). 
So maybe a check might be necessary for backward compatibility with 26-pin Pi

### DIFF
--- a/hardware/Gpio.cpp
+++ b/hardware/Gpio.cpp
@@ -125,26 +125,28 @@ void interruptHandler1 (void) { pushInterrupt(1); }
 void interruptHandler2 (void) { pushInterrupt(2); }
 void interruptHandler3 (void) { pushInterrupt(3); }
 void interruptHandler4 (void) { pushInterrupt(4); }
-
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thanks!, Is it correct i only see new interrupt handlers ?
If yes, i do not suspect this to give a problem on the old 'A' model.. (yep, people should upgrade anyway)

ogroot:
Yes, that is correct, only added interrupt handlers. 
Has been working flawless in my installation over the past 6 months. 
So, I think others might benefit from it as well.

gizmocuz:
Great, thanks !!!!

PR #1115, opened 2017-01-12
add "with LAN interface" to OpenWebNet

steplebr:
more precise names : 
add "with LAN interface" to OpenWebNet description
add "Bus" to bus switches description
rename COpenWebNet to COpenWebNetTCP


gizmocuz:
Great, thank you very much!

PR #1117, opened 2017-01-14
Multiple additions and bug fixes

dnpwwo:
Add HTTP Protocol support
Change onMessage interface
Make devices available after initial create
Add device create by TypeName


gizmocuz:
Thanks!

PR #1118, opened 2017-01-14
Set LUA_USE_MACOSX for the builtin LUA on OSX

samm-git:
Currently documentation recommends to modify lua sources (see https://www.domoticz.com/wiki/Mac_OS_X#LUA_Fix), however i think that its better to just enable this define automatically for the OSX build.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,6 +9,9 @@ SET(DOMO_MIN_LIBBOOST_VERSION 105500)
 
 option(USE_BUILTIN_LUA "Use builtin lua library" YES)
 IF(USE_BUILTIN_LUA)
+  IF(CMAKE_SYSTEM_NAME STREQUAL "Darwin")
+    add_definitions(-DLUA_USE_MACOSX)
+  ENDIF()
   add_subdirectory (lua)
   get_directory_property (LUA_LIBRARIES DIRECTORY lua
     DEFINITION LUA_LIBRARIES)


gizmocuz:
Is it possible that you could first check if our build system compiles the right binaries for you?

http://www.domoticz.com/downloads/

Our build system has no problems compiling of OSX ?

samm-git:
Hi, i started with your binaries. They are working correctly, but LUA cant load any external modules. This is why i had to recompile this myself and found this issue. In the feature i am planning to create a homebrew formula, so it will be possible to install it from homebrew directly.

gizmocuz:
why do you need external modules ? nobody can do this with the binary releases, nor is this default
can you work with the new python plugin system ?

samm-git:
Because i am getting temperature from my DIY sensors which are exposed using TCP. It is LUA one liner that fetching data from remotes (using luasocket) and updates devices binded to the virtual hardware. Am i missing something?

samm-git:
Anyway, if you feel that it is expected - lets close this PR and i will create homebrew formula which will create domoticz  linked with external LUA to avoid this issue.

gizmocuz:
If possible, use the new python plugin system (currently in beta) for this

gizmocuz:
Could it be that this patch should be in the Lua CMakeList.txt file ?

samm-git:
Ok, i will try to learn new plugin system and convert my LUA script to the python. Anyway, thanks for the commit.

PR #1125, opened 2017-01-16
Set more specific dim level values

YcKe:
Some led bulbs have a wide range of dimming values. In some cases the
level of 5% was still not low enough for some scenes or situations. For this reason I extended the
combobox so the user can select more different values.


gizmocuz:
Nice idea, but try filling the comboboxes with javascript please
and remember some dimmers have only 16 values, you will not see the difference, so maybe first check what type of dimmer it is, how many dim values (also provided via json), and fill the combo boxes with these values ? (in percentage)

gizmocuz:
Thanks, but this does not yet take the total amount of supported dim levels into account

YcKe:
The system has 3 different return values: 
- 'None': In this case the lightSwitch is not of type dimmer. We will do nothing and hide the level combobox. 
- 'All': In this case the lightswitch is of type STYPE_Dimmer, STYPE_BlindsPercentage or STYPE_BlindsPercentageInverted. In this case we know that all levels are supported so we can generate them. (0 to 100)
- Supported values: '1 2 3 4 5,...'; In this case the lightswitch is of type STYPE_Selector and the systen will return the supported dim values. Only these support dim values will be visible in the levels combobox.

Hope this answers your question.

gizmocuz:
i did not raise a question...
in the json output from a dimmer you will see how many dim steps it support, you should use this to build the level combox

YcKe:
This is the output of the function "getlightswitches" of a dimmer switch which supports all the levels:
      {
         "DimmerLevels" : "all",
         "IsDimmer" : true,
         "Name" : "Living: Dimmer zithoek",
         "SubType" : "Switch",
         "Type" : "Light/Switch",
         "idx" : "43"
      },

As you can see it returns for the parameter DimmerLevels 'all' because it supports all levels so we need to generate the values to populate the combobox ourself . 

Only switches of type STYPE_Selector will return something different like this:
      {
         "DimmerLevels" : "0,10,20,30,40,50,60,70,80,90,100",
         "IsDimmer" : true,
         "Name" : "Living: Dimmer zithoek",
         "SubType" : "Switch",
         "Type" : "Light/Switch",
         "idx" : "43"
      },

In the last case the code will use the DimmerLevels values to populate the combobox instead of generating them ourself.

gizmocuz:
Hmmm thats someone else his/her implementation
Do you also have this when you add a dummy AC Switch ?

YcKe:
This is the output for a dummy AC dimmer:
        "DimmerLevels" : "all",
         "IsDimmer" : true,
         "Name" : "Dummy AC",
         "SubType" : "AC",
         "Type" : "Lighting 2",
         "idx" : "257"
      },

This is the output for a dummy AC on/off:
      {
         "DimmerLevels" : "none",
         "IsDimmer" : false,
         "Name" : "Dummy AC ON/OFF",
         "SubType" : "AC",
         "Type" : "Lighting 2",
         "idx" : "258"
      },

gizmocuz:
This is not correct... thats what i expected...
Here you should find more information:

http://127.0.0.1:8080/json.htm?type=devices&filter=light&used=true&order=Name&plan=0

gizmocuz:
Specifically


HaveDimmer:true

MaxDimLevel:15

YcKe:
Thanks for the clarification. Now I understand what you meant before. 

Something weird I saw was that HaveDimmer is always set to yes even for the On/Off switches. SwitchTypes seems to me more reliable to check if we are talking about a dimmer?

What does MaxDimLevel exactly mean? 

What do you think about the idea to move all this logic regarding the dimmer levels determination into the WebServer.cpp class instead of having it spread in the .JS and Webserver.cpp?

gizmocuz:
it means they hardware support 15 (16 with 0), dim steps, thus (16.0/100.0f)*level_in_percentage
or in your case for i=1 to 16, value = (100/16)*i
Logic in the main application sound better then in the web gui
I think the author of the selector_switch added the other dimvalues variable, but forgot (or ...) to implement this for all switch types

YcKe:
I made some changes to the webservice and it now returns the following values for a dimmer with MaxDimLevel 15:
{
   "result" : [
      {
         "DimmerLevels" : "0,6,13,20,26,33,40,46,53,60,66,73,80,86,93,100",
         "IsDimmer" : true,
         "Name" : "Dummy dimmer",
         "SubType" : "AC",
         "Type" : "Lighting 2",
         "idx" : "1"
      }
   ],
   "status" : "OK",
   "title" : "GetLightSwitches"
}

Are these the value that you expect to be filled in the combobox?

gizmocuz:
Yes, this looks like it ;) great.... did you test it with a dummy (if you put it on 46, it stays on 46)

YcKe:
Yes, I tried 2 types of dummy dimmers (AC and regular in percentage) and both work and save the selected value correctly. 

I'll update my repository so you could have a check.

gizmocuz:
Thanks!

PR #1127, opened 2017-01-16
Cross compile without Libudev for Synology.

Jumbotroll:
cmake -DCMAKE_BUILD_TYPE=Release -DUSE_UDEV_OPENZWAVE=NO CMakeLists.txt
For cross compile without libudev for Synology.
Please delete this pull request, patches to openzwave are required.


gizmocuz:
See you closed it... was this not working ?
Better to ask OZW to remove the UDEV support ;) i tried... maybe you could try this too

Jumbotroll:
It was working if I also patched openzwave. 
I think maybe a CMakeLists.txt fix with "If Synology don't use UDEV" Then the build will work on All platforms. But the best thing is if OpenZwave remove UDEV :-)

gizmocuz:
I agree ;) i will also make a post on their forum

PR #1128, opened 2017-01-17
Fix lighttype code

Spugna85:


gizmocuz:
Sorry, i made some changes, could you check this again ? I also started from 400, and made the names/comments consistent

PR #1130, opened 2017-01-17
Added GPIO timer to rate limit interrupts

jvandenbroek:
I had a problem with sudden interrupt storms when reading out my gas meter, which occasionally kept on going when hanging on the reflective part it reads out (resulting in a massive high no count). By adding this timer my GPIO has become very stable and still is able to process multiple interrupts from different pins at the same time. Maybe the MAX_PERIOD_US could be a configurable setting (e.g. set it to several seconds to prevent doorbell being triggered multiple times), but for now I think this value will be fine for most situations. What do you think?

### DIFF
--- a/hardware/Gpio.cpp
+++ b/hardware/Gpio.cpp
@@ -58,8 +58,10 @@ Connection information:
 
 #define NO_INTERRUPT	-1
 #define MAX_GPIO	31
+#define MAX_PERIOD_US	50000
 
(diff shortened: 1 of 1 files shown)


gizmocuz:
Looking good! Not using it myself, so i leave this pull request open for a few days for other to comment
(If no objections... i will merge this, thanks!)

gizmocuz:
Thanks!

PR #1135, opened 2017-01-17
Fix out of bounds exception on some platforms

Risca:
prevPos was used uninitialized. This caused an out of bounds exception on
raspberry pi.


gizmocuz:
Thanks!

Issue #1068, opened 2016-12-23
Add POST option fields to HTTP/HTTPS poller


dynasticorpheus:
Would be great if we could also add the below option fields to the HTTP/HTTPS poller allowing more devices who need this to be exposed to domoticz. (similar to Custom HTTP/Action notification)

- POST Data
- POST Content-Type
- POST Headers

cc: @sbouchex  :)

sbouchex:
I think you are mixing 2 features:
- HTTP/HTTPS poller : domoticz is doing queries to a remote server
- Data inserting with lua scripts (see https://www.domoticz.com/wiki/Inserting_raw_data_lua_parsers) : domoticz is receiving HTTP query from a remote agent

From your request, I understand that you need to have the ability to get access to the fields with a lua_script, right ?

dynasticorpheus:
Thanks for your quick reply

No really would like to customize those fields when to doing the actual http(s) request to the server. For example adding a token to the header when requesting data from the server storing my gas usage.

sbouchex:
Sounds a good idea, another option will be needed : HTTP method (GET or POST)

dynasticorpheus:
Correct, thanks for considering this because I think a lot more people would benefit from this as well.

sbouchex:
Would you be ready to do some testing if a branch is available ?

dynasticorpheus:
Ofcourse!

sbouchex:
@gizmocuz I would need extra fields in the hardware table (to store "post data", "headers" & "content type"), any recommendation ? extra1, extra2, extra3 as VARCHAR (200) ?

gizmocuz:
We gave an extra field... is that not working ?

sbouchex:
HTTP poller hardware is already using the "extra" field to store the parsing script.

gizmocuz:
Yes, this can be extended with the optional fields ? (split them with a special character(s))
param1[/]param2[/]param3, or a simple semicon (;)

sbouchex:
ok I will check that

sbouchex:
@dynasticorpheus how large would be the data your would like to post ? Could you give us a sample ?
@gizmocuz I have some doubts that all fields will feat into the 200 VAR CHAR and I may need to extend the width

gizmocuz:
you could make it a TEXT field if you like, still 200 characters is a lot, most of the URL will be in the Address field anyway i suppose ?
[...]

PR #1138, opened 2017-01-18
Add POST option fields to HTTP/HTTPS poller

sbouchex:
New fields to support HTTP/HTTPS POST

see #1068

### DIFF
--- a/main/SQLHelper.cpp
+++ b/main/SQLHelper.cpp
@@ -33,7 +33,7 @@
 #define __STDC_FORMAT_MACROS
 #include <inttypes.h>
 
-#define DB_VERSION 109
+#define DB_VERSION 110
 
 extern http::server::CWebServerHelper m_webservers;
 extern std::string szWWWFolder;
@@ -2124,14 +2124,14 @@ bool CSQLHelper::OpenDatabase()
 			query("INSERT INTO TimerPlans (ID, Name) VALUES (0, 'default')");
 			query("INSERT INTO TimerPlans (ID, Name) VALUES (1, 'Holiday')");
 		}
-		if (dbversion < 108)
+		if (dbversion < 110)
 		{
 			query("ALTER TABLE Hardware RENAME TO tmp_Hardware;");
-			query("CREATE TABLE IF NOT EXISTS [Hardware] ([ID] INTEGER PRIMARY KEY, [Name] VARCHAR(200) NOT NULL, [Enabled] INTEGER DEFAULT 1, [Type] INTEGER NOT NULL, [Address] VARCHAR(200), [Port] INTEGER, [SerialPort] VARCHAR(50) DEFAULT (''), [Username] VARCHAR(100), [Password] VARCHAR(100), [Extra] TEXT DEFAULT (''),[Mode1] CHAR DEFAULT 0, [Mode2] CHAR DEFAULT 0, [Mode3] CHAR DEFAULT 0, [Mode4] CHAR DEFAULT 0, [Mode5] CHAR DEFAULT 0, [Mode6] CHAR DEFAULT 0, [DataTimeout] INTEGER DEFAULT 0);");
+			query(sqlCreateHardware);
 			query("INSERT INTO Hardware(ID, Name, Enabled, Type, Address, Port, SerialPort, Username, Password, Extra, Mode1, Mode2, Mode3, Mode4, Mode5, Mode6, DataTimeout) SELECT ID, Name, Enabled, Type, Address, Port, SerialPort, Username, Password, Extra, Mode1, Mode2, Mode3, Mode4, Mode5, Mode6, DataTimeout FROM tmp_Hardware;");
 			query("DROP TABLE tmp_Hardware;");
 
-			result = query("SELECT ID, Extra FROM Hardware WHERE Type=74;");
+			result = safe_query("SELECT ID, Extra FROM Hardware WHERE Type=%d", HTYPE_HTTPPOLLER);
 			if (result.size() > 0)
 			{
 				std::stringstream szQuery2;
--- a/main/SQLHelper.cpp
+++ b/main/SQLHelper.cpp
@@ -33,7 +33,7 @@
 #define __STDC_FORMAT_MACROS
 #include <inttypes.h>
 
-#define DB_VERSION 109
+#define DB_VERSION 110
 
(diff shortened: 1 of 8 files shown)


gizmocuz:
Got some compile errors see below

main/SQLHelper.cpp
```diff
@@ -2124,6 +2124,31 @@ bool CSQLHelper::OpenDatabase()
 			query("INSERT INTO TimerPlans (ID, Name) VALUES (0, 'default')");
 			query("INSERT INTO TimerPlans (ID, Name) VALUES (1, 'Holiday')");
 		}
+		if (dbversion < 110)
+		{
+			query("ALTER TABLE Hardware RENAME TO tmp_Hardware;");
+			query("CREATE TABLE IF NOT EXISTS [Hardware] ([ID] INTEGER PRIMARY KEY, [Name] VARCHAR(200) NOT NULL, [Enabled] INTEGER DEFAULT 1, [Type] INTEGER NOT NULL, [Address] VARCHAR(200), [Port] INTEGER, [SerialPort] VARCHAR(50) DEFAULT (''), [Username] VARCHAR(100), [Password] VARCHAR(100), [Extra] TEXT DEFAULT (''),[Mode1] CHAR DEFAULT 0, [Mode2] CHAR DEFAULT 0, [Mode3] CHAR DEFAULT 0, [Mode4] CHAR DEFAULT 0, [Mode5] CHAR DEFAULT 0, [Mode6] CHAR DEFAULT 0, [DataTimeout] INTEGER DEFAULT 0);");
```
gizmocuz: Instead of this, could query(sqlCreateHardware) not be called?

main/SQLHelper.cpp
```diff
@@ -2124,6 +2124,31 @@ bool CSQLHelper::OpenDatabase()
...
+			query("INSERT INTO Hardware(ID, Name, Enabled, Type, Address, Port, SerialPort, Username, Password, Extra, Mode1, Mode2, Mode3, Mode4, Mode5, Mode6, DataTimeout) SELECT ID, Name, Enabled, Type, Address, Port, SerialPort, Username, Password, Extra, Mode1, Mode2, Mode3, Mode4, Mode5, Mode6, DataTimeout FROM tmp_Hardware;");
+			query("DROP TABLE tmp_Hardware;");
+
+			result = query("SELECT ID, Extra FROM Hardware WHERE Type=74;");
```
gizmocuz: if possible make it
`safe_query("SELECT ID, Extra FROM Hardware WHERE Type=%d", HTYPE_HTTPPOLLER);`

main/WebServer.cpp
```diff
@@ -1280,7 +1280,38 @@ namespace http {
...
-			if (htype == HTYPE_PythonPlugin) {
+			if (htype == HTYPE_HTTPPOLLER) {
+				std::string mode1Str = request::findValue(&req, "Mode1");
+				std::string mode2Str = CURLEncode::URLDecode(request::findValue(&req, "Mode2"));
+				std::string mode3Str = CURLEncode::URLDecode(request::findValue(&req, "Mode3"));
+				std::string mode4Str = CURLEncode::URLDecode(request::findValue(&req, "Mode4"));
+				std::string mode5Str = CURLEncode::URLDecode(request::findValue(&req, "Mode5"));
+				std::string mode6Str = CURLEncode::URLDecode(request::findValue(&req, "Mode6"));
+				std::string szScriptEncoded = base64_encode((const unsigned char *)extra.c_str(), extra.length());
+				std::string szMethodEncoded = base64_encode((const unsigned char *)mode1Str.c_str(), mode1Str.length());
```
gizmocuz: Not sure why we need the new mode1Str strings ? This is where we habe an 'extra' field for, also for all function/calls below

main/WebServer.cpp
```diff
@@ -1323,7 +1354,7 @@ namespace http {
 
 				root["idx"] = sd[0].c_str(); // OTO output the created ID for easier management on the caller side (if automated)
 
-				m_mainworker.AddHardwareFromParams(ID, name, (senabled == "true") ? true : false, htype, address, port, sport, username, password, extra, mode1, mode2, mode3, mode4, mode5, mode6, iDataTimeout, true);
+				m_mainworker.AddHardwareFromParams(ID, name, (senabled == "true") ? true : false, htype, address, port, sport, username, password, extra, mode1, mode2, mode3, mode4, mode5, mode6, mode1Str, mode2Str, mode3Str, mode4Str, mode5Str, mode6Str, iDataTimeout, true);
```
gizmocuz: I think the 'extra' field should be sufficient for all hardware
If you need multiple extra parameters, you could contact them by a divider

gizmocuz (commented):
Thanks!
I am not keen about the newly introduced 'Mode1Str' strings/parameters as thats why we have the 'extra' field
Is it possible to just use this field ?

main/SQLHelper.cpp
sbouchex: sure

main/SQLHelper.cpp
sbouchex: ok

main/WebServer.cpp
sbouchex: I have decided to do all base64+concat into the server instead of in the web page part. So I had to transport all parameters. So I need more params

main/WebServer.cpp
sbouchex: I will refactor that part, you are right

main/WebServer.cpp
gizmocuz: you can still base64 encode your separated command string ?
base64enc("cmd1;cmd2;cmd3;cmd4;cmd5;cmdxx)

sbouchex: base64enc is not available in javascript, does it ?

gizmocuz: Well the above was just pseudo code
But they exists, have a look at the btoa() and atob() functions

sbouchex: ah ok, I'll check them 👍

gizmocuz:
Thanks!!

dynasticorpheus:
@sbouchex Was about to resume testing but already merged, apologies and thanks!

gizmocuz:
Testing is best to do from a new branch, not from code submitted to be pulled in the master repo

PR #1140, opened 2017-01-19
Fix returning uncompressed .js file with header Content-Encoding set to gzip

Red-F:
Only set encoding to gzip when the caller accepts gzip and a .gz was loaded.

Retrieving a javascript file from my (pi based) domoticz with the following request headers:
```
Cache-Control: no-cache
Connection: Keep-Alive
Pragma: no-cache
Accept: */*
Accept-Language: en-GB,en-US;q=0.8,en;q=0.6
```
Returns the following:
```
HTTP/1.1 200 OK
Content-Length: 74830
Content-Type: application/javascript;charset=UTF-8
Access-Control-Allow-Origin: *
Content-Encoding: gzip
Last-Modified: Thu, 19 Jan 2017 21:24:16 GMT
Connection: Keep-Alive
Keep-Alive: max=20, timeout=20

/*
//	(c)  Kendel Boul - 2014
*/

String.prototype.setCharAt = function(idx, chr) {
	if(idx > this.length - 1){
		return this.to
snipped

```
The file contents are uncompressed but the Content-Encoding is set to 'gzip'.

This took me some time to figure out since domoticz was setup to run behind a reverse proxy. In the end the solution (as always) was simple.

Regards,
Fred

### DIFF
--- a/webserver/request_handler.cpp
+++ b/webserver/request_handler.cpp
@@ -399,7 +399,7 @@ void request_handler::handle_request(const request &req, reply &rep, modify_info
   reply::add_header(&rep, "Content-Length", boost::lexical_cast<std::string>(rep.content.size()));
   reply::add_header(&rep, "Content-Type", mime_types::extension_to_type(extension));
   reply::add_header(&rep, "Access-Control-Allow-Origin", "*");
-  if (bHaveLoadedgzip)
+  if (bHaveGZipSupport && bHaveLoadedgzip)
   {
 	reply::add_header(&rep, "Content-Encoding", "gzip");
   }


gizmocuz:
Thanks, nice find !
One question, if gzip is not supported, then 'bHaveLoadedgzip' should not become true?

Red-F:
Yes that was my initial reaction too, but on closer look it turns out the request handler is quite flexible and allows html and javascript files to be present on the filesystem as .gz. The handler will load the .gz file and decompress it to append is as uncompressed response data. That part works fine, it is just that the content-encoding wrongly will show gzip. Browsing my Pi's filesystem I found that indeed all the javascript files were installed as .gz files. The reason that it cost me a rather long time to find the cause was me blaming myself for incorrectly configure the reverse proxy. Only when tracing request/response between proxy and domoticz I discovered the Content-Encoding was already present in the response from domoticz.

Cheers,
Fred

gizmocuz:
Thanks for the research!

PR #1142, opened 2017-01-20
Fix regression on notification extradata due to commit c786427bafa417…

ultratoto14:
…03b3860f1782525886a45a29db.

Do not know if it's still used but the given commit removed the extradata from the notification.
Want to do an additional pull request with the choice of the subsystem within lua. Need to fix that before.
But at the end the format will be something like:
commandArray['SendNotification']='subject#body#0#sound#extradata#gcm'.

Tell me if it's ok.


gizmocuz:
thanks!

PR #1143, opened 2017-01-20
InComfort v1.0.1

etamminga:
2 Minor issues fixed. 
'io' values (pump, tab, burner) states now convert correctly and slow updating values are now forced with an update every 15 minutes


gizmocuz:
Thanks!

PR #1147, opened 2017-01-21
[1wire] a few fixes and improvements

szczukot:


main/WebServer.cpp
```diff
@@ -964,7 +964,7 @@ namespace http {
 				if (ii == HTYPE_RaspberryGPIO)
 					bDoAdd = false;
 #endif
-				if (((ii == HTYPE_1WIRE) && (!C1Wire::Have1WireSystem())) || (ii == HTYPE_PythonPlugin))
+				if (ii == HTYPE_PythonPlugin)
```
gizmocuz, line 967: why is this removed ? (Have1WireSystem)
Now on a (windows) system that does not have 1wire we can add this hardware?

hardware/1Wire.cpp
```diff
@@ -32,57 +32,44 @@ C1Wire::C1Wire(const int ID, const int sensorThreadPeriod, const int switchThrea
...
 }
 
 bool C1Wire::StartHardware()
 {
+	if (m_system == NULL)
+		return false;
+
 	// Start worker thread
-	if (0 != m_sensorThreadPeriod)
+	if (-1 != m_sensorThreadPeriod)
```
gizmocuz: Please leave this at 0
Also if you leave it, we do not have to change the language files
(and you can not submit language file changes this way... send me a PM with the needed changes)
But just leave this at 0, and we do not need this ?

main/WebServer.cpp
szczukot: I have one domoticz on windows, but i using owfs which installed on raspberry, and i shared owfs path.
OWFS is always available on any OS.

hardware/1Wire.cpp
szczukot: Before there was a bug. TheradPeriod never was equal 0.
Now it is possible and makes sense. For example I do not want any interruption between the readings sensors.

gizmocuz: Logical 0=disabled 0=false

szczukot: This is not true.
I c++ and on many other languages, methods return -1 as false.
Many times "0" means something (lenght etc).

gizmocuz: Strange then right ? What is the value of a BOOL/false then ? (if you cast it to an integer yes)

gizmocuz: Maybe visual basic does this ?

hardware/1Wire.cpp
```diff
@@ -32,51 +32,38 @@ C1Wire::C1Wire(const int ID, const int sensorThreadPeriod, const int switchThrea
 {
 	m_HwdID = ID;
 
-	// Defaults for pre-existing 1wire instances
-	if (0 == m_sensorThreadPeriod)
```
gizmocuz, line 36: Maybe check if it's not 0, and < 5 * 60 * 1000 then 5 * 60 * 1000 ? else people will enter 1
Or we should change it in seconds/minutes (also in the web gui) ?

szczukot: Now "1" (milisecond) is good solution for many people, who need quick reading (especially the status of the switches). 
Previous change in code from another developer, already introduced ms instead of sec.

gizmocuz:
Thanks!

szczukot:
Pls add to languages (0 = disabled) for this parameters

gizmocuz:
I have reports that 1wire is not working anymore on a raspberry pi, with the onboard 1wire
Are you able to check this ?
Is the POLL interval still like below:
image

szczukot:
I tested on RPI 2 B (kernel and OWFS) and on Windows 10 (OWFS).
What is exactly dont'work ?
Is exists post on forum about it ?

regards
Fantom

gizmocuz:
Have you seen this ?
https://www.domoticz.com/forum/viewtopic.php?f=6&t=12058&p=115744&hilit=1+wire#p115744

gizmocuz:
I am reverting now to see if this solved the problem.
In the previous pull request was also a typo in the hardware.html (-1 to disable...)
And a patch in a language file that should not be here
Hope you can find the issue, please do test directly connected on the gpio pins

szczukot:
Too bad that these changes were removed.
There was fixed a few bugs and nothing more.
Pls read carefully all posts - the problem does not appear in this version.
Someone must have problems with their own hardware, or failure lies somewhere deeper in domoticz. (for example "still operation of the DS18B20s possible, till approx. 21.05: then read-out stopped (without any clear reason)".
Now in 1wire code exist bugs, for example disable polling for switches or sensor is impossible.

Pls do "revert of revert" and i will add some improvements and log for people, who have problem.

Fantom
ps. I use big 1wire network with about 40. sensors/switches. All version of domoticz work properly.

gizmocuz:
"for example disable polling for switches or sensor is impossible" is not a bug
Let me first test if this solves the "not working" problem first OK ?
If not, then i revert my revert of course
If yes, we have to look over the above changes

gizmocuz:
Okey, sensor are working again:
image

So this patch obviously broke something... and no... i did not touch any wires !

PR #1148, opened 2017-01-21
bug correction in hardware web page and refactoring

steplebr:


gizmocuz:
Thanks!, But because all of the refactoring i can not nee the bug you corrected...
Could you explain this ?

steplebr:
When I modified the hardware description from "MyHome OpenWebNet" to "MyHome OpenWebNet with LAN interface" as you asked me, I didn't realize that the conditions text.indexOf("MyHome OpenWebNet with LAN interface") >= 0 in HardwareController.js where not reached because there was a condition text.indexOf("LAN") >= 0 before. On the forum, I saw that spugna85 complianed about

not seeing password field anymore in the hardware page, making it impossible to use some gateways. Thus, I merged the actions for text.indexOf("MyHome OpenWebNet with LAN interface") >= 0 condition and text.indexOf("LAN") >= 0 && xxx. As the conditions were rather complicated, I grouped them similarly to what is done for textindexOf("USB")>=0, where there is a first condition "USB" then there are more specialized conditions "USB and MySensors", "USB and P1 Smart Meter" and so on.

De : gizmocuz <notifications@github.com>
Envoyé : dimanche 22 janvier 2017 13:19
À : domoticz/domoticz
Cc : stéphane; Author
Objet : Re: [domoticz/domoticz] bug correction in hardware web page and refactoring (#1148)


Thanks!, But because all of the refactoring i can not nee the bug you corrected...
Could you explain this ?

-

gizmocuz:
Thanks!

corbinmunce:
@steplebr, can you check your changes for the Xiaomi Gateway on the hardware page?  It should have options for IP address, port, and password.  I'm no longer seeing these fields after this update.

gizmocuz:
Had to revert your patch as it caused issues.
Is it possible to just fix the openwebnet problem ? (no refactoring of any kind please)
Don't forget to increase the appcache version number

see reverted merge and forum topic:
http://www.domoticz.com/forum/viewtopic.php?f=6&t=15674&p=115984#p115984

PR #1150, opened 2017-01-22
datapush: support 'counter' type

ikkemaniac:


push/Push.cpp
```diff
@@ -226,7 +226,7 @@ std::string CPush::ProcessSendValue(const std::string &rawsendValue, const int d
 	}	
 	else if (vType == "Counter")
 	{
-		sprintf(szData,"Not supported yet");
+		sprintf(szData,"%d", atoi(rawsendValue.c_str()));
```
gizmocuz: Maybe a small tweak, use a stringstream and convert it to an uint64_t ?

uint64_t dvalue;
stringstream(str) >> dvalue;

Ehhh on the otherhand, just use

strcpy(szData,rawsendValue.c_str())

ikkemaniac: Sure np. I'm not that familiar with this but why not `strtoul` or `strtoull`? (`strtol` instead of `atoi`?)

gizmocuz: simpel, you making a string from it...
and it is already a string
so just a 'strcpy' would work (mind the .c_str(), because we need to get it from a std::string)

ikkemaniac:
pushed requested change. Please be aware that this introduces a code inconsistency as this now is the only var that is copied as string, all other 'cases' still use sprintf with atoi, atof etc....

push/Push.cpp
```diff
@@ -226,7 +226,7 @@ std::string CPush::ProcessSendValue(const std::string &rawsendValue, const int d
 	}	
 	else if (vType == "Counter")
 	{
-		sprintf(szData,"Not supported yet");
+		strcpy(szData, rawsendValue.c_str())
```
gizmocuz: I don't think you tested this code ;)
(See our automatic build system)
Please always test...

gizmocuz (commented):
There is a typo in your change

If 'rawsendValue' is already a whole number, there is absolute no reason to use 'printf("%d",atoi(string))'
for floats/doubles, this might be valid, because if you want one xx.y , and the string is xx.yyyyyy then it's ok

ikkemaniac:
working on too much at the same time.... fixed typo

gizmocuz:
;) thanks again !

PR #1154, opened 2017-01-23
Adding choice of notification subsystem in LUA and Blocky. Need UI to…

ultratoto14:
… be changed.

Added the choice of notification subsystem for lua and blocky.
Do not know where to change the blocki UI to give the ability to choose the subsystem.
In LUA, will look as this as we already have multiple possible options like extradatan sound and priority.

commandArray['SendNotification']='subject#body#0#sound#extradata#gcm'.


main/EventSystem.cpp
```diff
@@ -2048,7 +2048,7 @@ bool CEventSystem::parseBlocklyActions(const std::string &Actions, const std::st
 		}
 		else if (deviceName.find("SendNotification") != std::string::npos)
 		{
-			std::string subject, body, priority("0"), sound;
+			std::string subject, body, priority("0"), sound, subsystem("");
```
gizmocuz: No need for (""), not done for the other strings

main/EventSystem.cpp
```diff
@@ -3057,7 +3061,7 @@ bool CEventSystem::processLuaCommand(lua_State *lua_state, const std::string &fi
 	std::string lCommand = std::string(lua_tostring(lua_state, -2));
 	if (lCommand == "SendNotification") {
 		std::string luaString = lua_tostring(lua_state, -1);
-		std::string subject(" "), body(" "), priority("0"), sound;
+		std::string subject(" "), body(" "), priority("0"), sound, subsystem("");
```
gizmocuz: No need for (""), not done for the other strings

gizmocuz (commented):
Thanks for this!
To change the blockly's... it is a small research at the first beginning...
The files you need to change are:
domoticzblocks.js
domoticzblocks_messages_en.js

you need to unpack them (they are gzipped) in the same folder, and you need to remove the gzipped versions
when you are done, you have to do the opposite

ultratoto14:
I removed the unneeded spaces and updated the blocky js to add the subsystem. Seems that email and SMS use a different way so i didn't put them in the list.

gizmocuz:
Okey, tell me when its ready to be merged... you tested all (also old) situations ?

ultratoto14:
Yep, tested, my only concern is that if i use the http API and do a send notification without specifying any specific subsystem, the message is sent to all configured way. If i use send_notification in LUA or Blocky, email is not sent. But it was the same behavior on the master branch.
As i saw that send email and send sms has been separated, i thought it was intended.

ultratoto14:
Just tested again on master branch, if i send through blocky or lua, i've got in logs:
Notification sent (email) => Success but no mail received. Is it the expected behavior (i mean not on my branch but on the domoticz/master  one)

ultratoto14:
Ok, seems to be linked to bFromNotification.
It can be merged.

gizmocuz:
Thanks!

PR #1162, opened 2017-01-24
[1wire] fix for kernel mode

szczukot:


szczukot:
I probably find bug for 1wire kernel mode (racing thread for GetDevices and ThreadFunction).
Pls check

hardware/1Wire/1WireByKernel.cpp
```diff
@@ -233,8 +233,11 @@ void C1WireByKernel::GetDevices(/*out*/std::vector<_t1WireDevice>& devices) cons
 {
    // If devices not initialized (ie : at least one time read), return an empty list
    // to not update Domoticz with inconsistent values
-   if (!m_AllDevicesInitialized)
-      return;
+
+	while (!m_AllDevicesInitialized)
```
gizmocuz: what is 'm_AllDevicesInitialized' never becomes true ? this is a deadlock situation then ?
It seems GetDevices is only called once... 
Maybe this should be done in the poll interval (1Wire.cpp, line 139) until m_AllDevicesInitialized is true ?

hardware/1Wire/1WireByKernel.h
```diff
@@ -1,5 +1,6 @@
 #pragma once
 #include "1WireSystem.h"
+#include <boost/atomic.hpp>
```
gizmocuz: Please do not use boost atomic

hardware/1Wire/1WireByKernel.cpp
szczukot: will not deadlock
GetDevices() called from 1wire.cpp will wait endlessly .... but without deadlock
But m_AllDevicesInitialized set to true always after max few seconds (in ThreadFunction)

szczukot (commented):
What can I use instead? mutex ?

gizmocuz:
There are to many mutexes already in this class
Would just leave it like it was
Calling two muxtexes after each other does not look like an option

szczukot:
Maybe write to author of this class, for help ?
This is not very nice code (many mutex, racing threads etc).

szczukot:
pls check this proposal

Fantom
ps. I don't fully understand this code (for kernel mode). Because it seems that it still reads the data from the sensors - without interruption (occupying hardware). A superclass only reads data from it at a given period of time.

hardware/1Wire/1WireByKernel.cpp
```diff
@@ -23,9 +23,10 @@
 #endif //_DEBUG
 
 
-C1WireByKernel::C1WireByKernel() :
-   m_AllDevicesInitialized(false)
+C1WireByKernel::C1WireByKernel()
 {
+   ThreadBuildDevicesList();
```
gizmocuz: This might take a long time and causes domoticz to start slow when you have a lot of sensors
That is probably why this is done in the ThreadFunction
You could perform this in StartHardware

szczukot:
good point

hardware/1Wire/1WireByKernel.cpp
szczukot: good point

szczukot:
another proposal

gizmocuz:
Looks perfect, but can you test this ? (not with OWFS)

szczukot:
I have problem with testing, because all beta, stable and all fixes what i'm did work on my rpi (both owfs and kernel mode).
The problem is for a few users (a different configuration of hardware, daemons etc).
Currently (on beta) 1wire of them does not work, so I suggest to use this fix, and see if the problem will be solved for them.

gizmocuz:
Okey, but then doing a fork, and applying your patches there, and let the users compile your fork to test would be best
or i can checkout your fork, do the building, and we can provide these users with a 'wget' url (and instructions) how they can update

best if they:
- make a backup of their old executable
- compile domoticz
- replace the old exe with the new exe
- restart domoticz
- in case of problems, they van revert to the original exe

szczukot:
I think that better is second solution: do build and give it for they.

gizmocuz:
Could you fork the master repo and commit your changes in your fork ?

szczukot:
https://github.com/szczukot/domoticz/tree/1wire

gizmocuz:
Thanks.. building at the moment... will take an hour
I will report back here shortly with the (download) instructions

gizmocuz:
i think this should work for installation instructions, maybe you can test and post them to the forum ?

step 1, create a backup of the original domoticz executable (needs to be done only once!!!)
cd domoticz
cp domoticz domoticz.orig

still inside the domoticz folder
stop domoticz with

`sudo service domoticz.sh start`

download and extract the test version:

```
wget https://www.domoticz.com/domoticz.bz2
bunzip2 -f domoticz.bz2
chmod +x domoticz

```
start domoticz again

`sudo service domoticz.sh start`

szczukot:
Probably works :)
http://www.domoticz.com/forum/viewtopic.php?p=116709&sid=f9cc3dca424efc04a99616c08178b77b#p116709

gizmocuz:
Yes, but there is the problem of 'accept new hardware' that has to be enabled in settings...
do you know why ? (and could you fix this?)

szczukot:
I no have idea.
Thi is very strange, because in 1wire code, 'accept new hardware' never uses.

gizmocuz:
That's also what i thought... but hmmmm
I will install this on some raspberry's and also do some debugging

gizmocuz:
Its working!

One request, when the user specified a poll interval of 10 minutes, at startup, could you set the first poll interval to 'now' + 5 seconds, so when this hardware is started, after 5 seconds the sensors will be polled, and then at the specified time interval ?

So in:

void C1Wire::SensorThread()

int iteration = pollIterations-5;

The poll/sleep is totally different then the rest of the system uses, and also blocks at least 1000ms, other hardware sleeps for 500ms

I find the whole polling method 'bursting' , meaning, all at the specified poll interval all sensors are polled and updated in the system.
If you have 40 sensors, CPU usage will be high!

A better method:
user specifies poll interval of 60 seconds
user has 25 devices
poll interval between each sensor is 60/25 = 2.4 seconds, rounded to whole seconds = 2 (or to 500ms incase the poll method is improved)

Most routines to report things should use the default reporting functions

ReportTemperature for instance should not call sDecodeRXMessage directly but SendTempSensor()
Do note that a conversion of the sensor sSubType could be needed in the database when this is fixed

sTypeTEMP10 -> sTypeTEMP5, probably also ID/Unit

gizmocuz:
No reaction, guess not needed

szczukot:
I think, that this fix is needed : https://www.domoticz.com/forum/viewtopic.php?f=6&t=12058

PR #1168, opened 2017-01-26
General Fixes and Uplifts

dnpwwo:
Improve HTTP protocol support
Add delayed message sending
Improved ImportError messages
Only plugin.py files are now scanned for plugin definitions
Prevented plugins adding events to queue once the plugin is stopping
Changed onMessage event to take bytes to characters to allow for full 8 bit coding
Resolved issues with several device types created using the TypeName specifier


gizmocuz:
Thanks!!!

PR #1173, opened 2017-01-28
XiaomiGateway updates

corbinmunce:
fix power plug control message
add variable for gateway sid prefix
add new cube text device


gizmocuz:
Thanks!

PR #1176, opened 2017-01-28
Add ability to email multiple camera snapshots.

ultratoto14:
Adding the ability to put comma (;) separated camera ids to send multiple cameras within the same email.
Current behavior remains the same. Multiple ids  add multiple attachements.
BR.
U.

### DIFF
--- a/main/Camera.cpp
+++ b/main/Camera.cpp
@@ -183,7 +183,7 @@ bool CCameraHandler::TakeRaspberrySnapshot(std::vector<unsigned char> &camimage)
 
 	std::string raspistillcmd="raspistill " + raspparams + " -o " + OutputFileName;
 	std::remove(OutputFileName.c_str());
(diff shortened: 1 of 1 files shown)


gizmocuz:
Could you explain where/how this is used?
In the camera settings, you can specify for each camera which switch/motion sensor should send a snapshot
Specifying the same switch/sensor for multiple camera's should send multiple snapshots
In lua/blockly you can also send as many as you want

ultratoto14:
This is when you use the emailcamerasnapshot http API. 
With Blocky, you can sequentially add multiple emails.
With this feature, we can sen one email with multiple snapshots. You have easily one view of multiple cameras. If you use Domoticz as an alarm, having multiple snapshots within the same email can be very useful.

gizmocuz:
Thanks!

PR #1177, opened 2017-01-28
Fixes in GetV6BridgeID & SendV6Command with timing and unexpected mes…

chrisverhoef:
Add some fixes for people who also uses for example the andriod app. Multiple sources to the ibox let the ibox send some undocumented comments. I skip them for now and ask LimitlessLed what the command means and how to act.

### DIFF
--- a/hardware/Limitless.cpp
+++ b/hardware/Limitless.cpp
@@ -238,14 +238,16 @@ bool CLimitLess::StartHardware()
 				{
 					if (!AddSwitchIfNotExits(3, "AppLamp Group3"))
 					{
-						AddSwitchIfNotExits(4, "AppLamp Group4");
+						if (!AddSwitchIfNotExits(4, "AppLamp Group4"))
+						{
+							if (m_BridgeType == LBTYPE_V6)
+							{
+								AddSwitchIfNotExits(5, "AppLamp Bridge");
+							}
+						}
 					}
 				}
 			}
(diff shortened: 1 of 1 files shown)


hardware/Limitless.cpp
```diff
@@ -238,15 +238,17 @@ bool CLimitLess::StartHardware()
 				{
 					if (!AddSwitchIfNotExits(3, "AppLamp Group3"))
 					{
-						AddSwitchIfNotExits(4, "AppLamp Group4");
+						if (!AddSwitchIfNotExits(4, "AppLamp Group4"))
+						{
+							if (m_BridgeType == LBTYPE_V6)
```
gizmocuz: This was removed for a reason.
If you switch from bridge types and do not change the light type, the 5 lamp groups still exist, but the bridge light never gets created
Leaving in where it was before does not do any harm

gizmocuz (commented):
Thanks, but please undo one change

hardware/Limitless.cpp
chrisverhoef: Will do, this was not a change on purpose but from a behind fork. Will send new pull request later.

gizmocuz: if you (keep) push changes to your fork, they will be automatically here

gizmocuz:
Thanks!

PR #1178, opened 2017-01-28
Added a switch indicating the current cost slot

Pigloox:
Added a switch indicating to the user whether it is currently full fare or low fare.

The data is evaluated directly from the EDF frame by the TELEINFO USB module

If the switch is on, the off-peak hours are active
If the switch is off, full working hours are active.

### DIFF
--- a/hardware/Teleinfo.cpp
+++ b/hardware/Teleinfo.cpp
@@ -293,16 +293,9 @@ void Teleinfo::MatchLine()
 				m_p3power.powerusage1 = ulValue;
 			break;
 		case TELEINFO_TYPE_PTEC:
-			if (vString.substr (0,2) == "HC")
-                        {
-                                SendSwitch(5, 1, 255, true, 0, "Heures Creuses");
-                        }
-                        else if (vString.substr (0,2) == "HP")
-                        {
-                                SendSwitch(5, 1, 255, false, 0, "Heures Creuses");
-                        }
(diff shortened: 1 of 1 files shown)


ultratoto14:
Seems that there's a typo as HC is for "Heures Creuses" and HP for "Heures Pleines" but the code use "Heures Creuses" for both.

Pigloox:
Yes only 2 states are possible. 
When the switch is on : HC
When thé switch is off : HP

ultratoto14:
Ok, didn't get that.

hardware/Teleinfo.cpp
```diff
@@ -292,6 +293,15 @@ void Teleinfo::MatchLine()
 				m_p3power.powerusage1 = ulValue;
 			break;
 		case TELEINFO_TYPE_PTEC:
+			if (vString.substr (0,2) == "HC")
+                        {
+                                SendSwitch(5, 1, 255, true, 0, "Heures Creuses");
```
gizmocuz: Please change the whole block (if/else) to

SendSwitch(5, 1, 255, (vString.substr (0,2) == "HC"), 0, "Heures Creuses");

gizmocuz: If possible, also use an English switch name, yes even TeleInfo is only in France (for now... but you never know...)

Pigloox: It's done

gizmocuz:
Thanks!

PR #1180, opened 2017-01-28
Fix USE_PYTHON_PLUGINS=NO compile

dnpwwo:


gizmocuz:
Thanks, is it possible to make the plugins object based?

PR #1186, opened 2017-01-30
Fixed command for RGBW

chrisverhoef:


gizmocuz:
I dont think this is right, you made it the same command as the RGBWW
I tested this with an RGBW, i will test this first here to see if the light also works with the RGBWW command
but i dont think so, otherwise i had build it like this
For the user who had a problem with this.. then he should set the configuration to RGBWW correct?

chrisverhoef:
Ok, I will wait for your test.

gizmocuz:
Tested, the code as it is now is correct.
For RGBW bulbs, we need to send RGBW commands, and not RGBWW
You can also not Link when sending a RGBWW On or Link command

chrisverhoef:
Ok, thanks for testing.

PR #1190, opened 2017-02-01
Upgrade GCM notification to send all the available data.

ultratoto14:
Using google cloud messaging, only a message is sent. Also the message contains only the subject.
I kept the same behavior for compatibility purpose but added other information in the json data.

The added data are:
subject
body (text)
extradata
deviceid
priority.


gizmocuz:
Thanks!

PR #1191, opened 2017-02-01
P1 Smart Meter reduce update interval

jvandenbroek:
My new DSMR 5.0 meter outputs data every second, resulting in the P1 Power meter also updated every second. I changed this to max every 30 seconds, which reduced the CPU load on my RPi 1 significantly.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,4 +30,3 @@ domoticz.exe
 /Config/zwcfg_*.xml
 /makebetarob
 /plugins/manifest.xml
-server_cert.pem
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,4 @@ domoticz.exe
 /Config/zwcfg_*.xml
 /makebetarob
 /plugins/manifest.xml
+server_cert.pem
--- a/hardware/P1MeterBase.cpp
+++ b/hardware/P1MeterBase.cpp
@@ -171,13 +171,9 @@ bool P1MeterBase::MatchLine()
 
 		if (l_exclmarkfound) {
 			time_t atime=mytime(NULL);
-			bool bSend2Shared=(difftime(atime,m_lastSharedSendElectra)>59);
 			if (
 				(std::abs(double(m_lastelectrausage)-double(m_p1power.usagecurrent))>40)&&
-				(difftime(atime,m_lastSharedSendElectra)>29)
-				)
-					bSend2Shared=true;
-			if (bSend2Shared)
+				(difftime(atime,m_lastSharedSendElectra)>9))
 			{
 				m_lastelectrausage=m_p1power.usagecurrent;
 				m_lastSharedSendElectra=atime;
--- a/hardware/P1MeterBase.cpp
+++ b/hardware/P1MeterBase.cpp
@@ -7,12 +7,12 @@
 #define CRC16_ARC	0x8005
 #define CRC16_ARC_REFL	0xA001
 
-typedef enum { 
-	ID=0, 
-	STD, 
-	LINE17, 
-	LINE18, 
-	EXCLMARK 
+typedef enum {
+	ID=0,
+	STD,
+	LINE17,
+	LINE18,
+	EXCLMARK
 } MatchType;
 
 #define P1_SMID			"/" // Smart Meter ID. Used to detect start of telegram.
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -30,3 +30,4 @@ domoticz.exe
 /Config/zwcfg_*.xml
 /makebetarob
 /plugins/manifest.xml
+server_cert.pem
```
gizmocuz: Could you undo this change ?

hardware/P1MeterBase.cpp
```diff
@@ -159,26 +159,29 @@ bool P1MeterBase::MatchLine()
...
 
 		if (l_exclmarkfound) {
 			time_t atime=mytime(NULL);
-			sDecodeRXMessage(this, (const unsigned char *)&m_p1power, "Power", 255);
 			bool bSend2Shared=(difftime(atime,m_lastSharedSendElectra)>59);
-			if (std::abs(double(m_lastelectrausage)-double(m_p1power.usagecurrent))>40)
-				bSend2Shared=true;
+			if (
+				(std::abs(double(m_lastelectrausage)-double(m_p1power.usagecurrent))>40)&&
+				(difftime(atime,m_lastSharedSendElectra)>29)
```
gizmocuz: I don't think this is the way to solve it.
A line above we check if the last-send was 60 seconds ago, if yes, send it
Maybe this line should be changed

if difference>40 AND lastsend was atleast 60 seconds ago, then send

But instead of 60 seconds, 10 seconds would be OK as other meters sends data every 10 seconds.
If the data is the same, then because of the value '40' check, it is not send

Other then this, you data is still processed every second
The only thing not done is write the values in the database (and some extra logic)

If we talk about CPU usage, how much does this increase ?

gizmocuz (commented):
Some questions

.gitignore
jvandenbroek: Ah yes, will do.

hardware/P1MeterBase.cpp
jvandenbroek: I thought to keep the original logic to always send every 60 seconds, wasn't sure why this is included (some form of heartbeat?).

It could be changed to 10 seconds, but is it really necessary to update that often? I thought that 30 secs (or even 60 secs) would be fine in every case.

I realize the data is processed every second, but that doesn't seem to cause much load. The update routine on the other hand is beating on my little RPi's CPU ;-) Could be worse in my case because I'm running the DB on a NFS share.

The (average) load difference is about 0.3, but especially the peak load has decreased significantly from 1.4 to 0.7 (CPU freq between 700 and 1000MHz).

gizmocuz:
Well it is nice to see updates every 10 seconds (... if there are updates.. thats why the 40)
Because you turn on a switch... you want to see the load difference
Every second is maybe to much, every 10 is OK, that's how it is now.. cpu percentage 0%

And 1.4% ... that is nothing ..... really... is the cpu was 24/7 at 80%... then that is what i call a lot... but still OK

jvandenbroek:
OK, I'll change it to 10 seconds. But the load was in Linux system load values, so it would actually be roughly 30 to 70% CPU usage..

gizmocuz:
Thats way to much indeed
But the whole processing is still done, only the 'sending/handling/storing' in domoticz was not
But 10 seconds should be perfect... thats how 90% of the P1 meters are now also and nobody has a high cpu load

jvandenbroek:
Whoops, I'm screwing up the server_cert thing. Will investigate :-( 
Also it will never become true now, will change the code ;-)
Hmm well I'm a bit confused I guess, it starts with value 0, so the difference matching should be fine. Sorry. Now figuring out how to best restore the server_cert.pem which I modified (that's why it had put it in my git ignore before)

jvandenbroek:
I couldn't figure out how to re-add without actually changing the file. Is it fine to merge? I can also revert (reset) completely, but not sure if that's necessary ?

gizmocuz:
You can also close this pull request and create a new one if you like

jvandenbroek:
Well I think I finally figured it out.. (stupid Git). Sorry for the many comments / updates, think the pull is now fine.

gizmocuz:
Thanks! Looking good

PR #1192, opened 2017-02-01
OpenWebNet: fix device status update with unofficial gateway

Spugna85:
Fix the failure interpretation of 'translation command' (what=1000) for lighting devices.
Sometimes this caused the failure of the state of light upgrade, especially with unofficial gateway.

The problem was introduced with the improvements to the parser of OpenWebNet messages, which now include "What Parameters".


gizmocuz:
Thanks!

PR #1193, opened 2017-02-02
XiaomiGateway added extra switch code from Levynger

corbinmunce:


gizmocuz:
Is it possible to have the extra logging only in debug mode (or if verbose level is Debug) ?

gizmocuz:
Thanks...

PR #1195, opened 2017-02-02
Add -sslkey option to allow separate cert/key files

dwmw2:
In some environments (like OpenWRT/LEDE) we have tools which automatically obtain LetsEncrypt certificates but put the cert and key in separate files. Let's make it possible for Domoticz to use them like that.

The new `-sslkey` option is optional, and it'll just use the `-sslcert` path by default if unset.


gizmocuz:
Thanks!

PR #1196, opened 2017-02-02
P1 Smart meter update logic (again)

jvandenbroek:
Make sure that the meter is updated at least once a minute.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,4 +30,3 @@ domoticz.exe
 /Config/zwcfg_*.xml
 /makebetarob
 /plugins/manifest.xml
-server_cert.pem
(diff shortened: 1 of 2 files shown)


gizmocuz:
Thanks!

adgoudsmit:
Guys, this fails when you have solar panels. Then "usagecurrent" is 0 during the day, hence only every minute an update from my smart meter.

gizmocuz:
Ehhh you are talking about the >40 ? This was already there a long time
this is also not for delivery
if you want this for delivery to, you could make a patch so it will be an OR ||
if ((usage_difference>40)||(delivery_difference>40)))

adgoudsmit:
Ok, this might have been there since some time, but with the latest changes this causes a one minute update for me now. So this is an issue for everybody who has solar panels. I also added some comments on 2f1cb58.

PR #1197, opened 2017-02-03
Save svalue to database for LimitlessLights

corbinmunce:
I hope this simple addition is OK


gizmocuz:
Did you had a chance to test this change ? There is a build error
I think this also might have implications later

corbinmunce:
I didn't compile for this pull request since I the change ready to paste in, and then forgot to remove the original line in my rush.  Will make sure to test properly next time.

Regarding implications later... is updating the sValue in this way potentially going to cause a problem?

gizmocuz:
Thanks, let me check the sValue ...

gizmocuz:
Nope, no problem at all.
But what is the value of sValue ?
I think the limitedled type should be converted to a new General Led type... where we can also have more parameters... but thats for later ;)

corbinmunce:
it is for the brightness, which is in lastlevel - but I don't think last level is is available via Lua, when sValue is.  So it can then be used in a script to control bulb brightness: https://www.youtube.com/watch?v=g5sUVF3R2Pk

gizmocuz:
Nice... in the end i think we need more values like also the RGB color

corbinmunce:
yes would be good to always have the RGB color and brightness states available for the lights

gizmocuz:
This would be a semicolon separated sValue later then

PR #1199, opened 2017-02-03
Prevent deadlock in DST "black hole" time calculations

gordonb3:
When doing calculations on times that are skipped due to a DST state change using the DST state flag to validate that our calculation is correct always returns a negative answer. The solution provided here prevents an endless loop by only allowing either state (on/off) to be checked only once and then falls back to `accept any`.

As the resulting time value turns out to be inconsistent between platforms the fallback method also shifts the time one hour ahead which should match the DST jump. For European times this produces the correct response on the platforms previously found to produce different results (Windows and Linux).


gizmocuz:
Thanks!

gordonb3:
No problem.

Be advised that if the DST shift uses a different interval than one hour the result may still be unexpected. It is best to avoid using schedule times that can be inside the black hole time range.

PR #1200, opened 2017-02-03
Make Kodi.py WOL work with python3

sweetpants:
Initial WOL support did not work because it did not use python3


gizmocuz:
Sorry, got some conflicts here that prevent this to be merged

gizmocuz:
There seems to be a conflict with your own python file here...

PR #1206, opened 2017-02-04
[BleBox] show devices uptime

szczukot:


gizmocuz:
please check the build log for errors.
i guess __int64 should be replaced by int64_t ? try to use the std standard

gizmocuz:
Thanks!

Issue #603, opened 2016-04-11
Javascript JSON API authentication issue


sjorsjes:
Hi,

I'm trying to connect to my domoticz using the JSON API. When using the api url in the browser I do get a json result. Unfortunately I am not able to retrieve data from the api using javascript. First I got a CORS errors which can be resolved with using the 'no-cors' mode, but then I get a 401 Unauthorized.

Needles to say I changed the ip and authentication string.

Code example:

```
fetch('http://xx.xxx.xx.xx:xxxx/json.htm?type=devices&filter=all&used=true&order=Name', {
    mode: 'no-cors',
    headers: {
        'Accept': 'application/json',
        'Content-Type': 'application/json',
        Authorization: 'Basic xxx'
    }
}).then(function (response) {
    return response.json();
}).then(function (data) {
    console.log(data);
}).catch(function () {
    console.log('error');
});
```

Something seem to be wrong with the API or atleast for using it with JS.

gizmocuz:
Please use our forum

sjorsjes:
Pretty sure it's a bug, but I will give it a go.

empierre:
I'm interrested in the output too as this is quite standard approch

sjorsjes:
Link to the post on the forum: https://www.domoticz.com/forum/viewtopic.php?f=6&t=11516

PR #1210, opened 2017-02-07
Respond correctly to CORS preflighted requests for JSON API

davidlb:
Fix issue #603 

More details about preflighted requests : https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS#Preflighted_requests


gizmocuz:
Thanks!

PR #1211, opened 2017-02-07
Make daemonize compatible with Systemd forking type

oohlaf:
When using Systemd service with type set to forking, the parent PID is Systemd itself, thus fails on this check and deamonize will return instead of forking.

### DIFF
--- a/History.txt
+++ b/History.txt
@@ -3,6 +3,7 @@ Version 3.xxxx (January xth 2017)
 - Fixed: Devices could become invisible when a scene/group was added to the 'hidden devices' room plan with the same idx
 - Fixed: JSon float/nan value, solved by upgrading JSonCPP
 - Fixed: OZW, when a value is updated and the sensor did not exists before (for example a kWh sensor), it is created (saves a restart)
+- Fixed: Deamonize compatible with Systemd forking type
 - Changed: Blinds T5 till T13 new DeviceID generation, could cause new sensors
 - Changed: Switch type 'Door Lock' renamed to 'Door Contact'
 - Changed: Minimized On/Off/Script execution time
--- a/main/domoticz.cpp
+++ b/main/domoticz.cpp
@@ -227,13 +227,6 @@ void daemonize(const char *rundir, const char *pidfile)
 	struct sigaction newSigAction;
 	sigset_t newSigSet;
 
-	/* Check if parent process id is set */
-	if (getppid() == 1)
-	{
-		/* PPID exists, therefore we are already a daemon */
-		return;
-	}
-
 	/* Set signal mask - signals we want to block */
 	sigemptyset(&newSigSet);
 	sigaddset(&newSigSet, SIGCHLD);  /* ignore child - i.e. we don't need to wait for it */


gordonb3:
Won't this cause issues with program re-entry? Simply removing that check without knowing why it was there in the first place doesn't seem like a very good idea.

oohlaf:
No it will not and I can explain.

The reason mentioned in the code is a false one. You can't judge that the process is already a daemon by checking the parent PID for being equal to 1.

Looking at the running processes on a system it is very likely that a process launched by PID 1 (init) is a daemon. This is very well explained [here](http://unix.stackexchange.com/questions/159964/how-to-check-whether-a-process-is-daemon-or-not).
> Having ppid=1 is not a reliable indicator of being a daemon. Plenty of processes that aren't daemons have survived their parent.

But if you are the program and upon starting you look at your own process before it even attempts to fork to become a daemon is silly. This worked with init.d style system as the bash shell script was in between the init daemon and the program. Thus the parent process was the shell init.d script.

With Systemd the program is directly launched from PID 1, thus this check prevents Domoticz from forking to become a daemon in the first place and only runs in the foreground. There is actually a very good description of what a program is supposed to do to become a nice behaving daemon and it is described in `man 7 daemon`. In that checklist you can see that checking the parent process for number 1 is not something that is needed.

Many other daemon programs contained the same mistake and have also removed this code. For example the program [dmapd in this commit](https://github.com/wangd/dmapd/commit/b4617ae6fb8cefcb4fdaef27f779adb4ff38c289).

So again, the function `daemonize()` is supposed to make a daemon. Checking on parent process is a bad indicator as explained above. Besides domoticz is not a _real forking daemon_ like apache is (for example that the parent process creates many child processes). The only purpose of `daemonize` is to close file descriptors and execute in an isolated session (after `setsid` call).

With program re-entry you are probably referring to the process of forking (?). A syscall to `fork` duplicates the process at that specific spot in the program (line 261): 1 process remains the caller, the other process becomes the child. Checking which of the two you are happens on lines 263-273, and not in lines 231-235 that I removed. Also the PID file creation on lines 276-291 already prevent that the same process forks multiple times, as the PID file can only be locked once. So if this is what you call program re-entry, than no, it will not happen.

I can't spell it out any more than this to motivate ;-)

oohlaf:
Hmmpf, pressed the wrong github button and accidentally closed the pull request...

gizmocuz:
Thanks!

gizmocuz:
do we also need a wait(NULL) or is this not needed ?

gordonb3:
Okay. I have no in-depth knowledge of that bit of code, but usually there is a reason for such constructs to be included. This may include that the program will misbehave on some other platform (Windows?).

gizmocuz:
Under Windows , the application is not launching as a daemon

oohlaf:
> do we also need a wait(NULL) or is this not needed ?

No not needed, only if you want the parent process to wait for the child to die (to restart a child for example).

gizmocuz:
Thanks!

yfaykya:
I think this may have broken systemd behaviour for me. Using this config  :  

[Unit]
Description=domoticz_service
[Service]
ExecStart=/home/pi/domoticz-dev/domoticz -daemon -loglevel 0 -www 8080 -sslwww 8443 -sslcert /root/sslcert.pem  -log /var/log/domoticz.log -verbose 2
StandardOutput=null
[Install]
WantedBy=multi-user.target

Domoticz fails on start with a timeout error (log shows nothing). I reverted this change and it works again.

oohlaf:
That's because your command line options do not correspond to the proper systemd settings.

If you specify `-daemon` on the `execstart` command line, you start in daemon mode which causes Domoticz to fork, thus you need to use systemd `Type=forking`. The default type in Systemd is `simple` which assumes the program runs in the foreground.

Please note that you need to customize the following with the specific options as to how you want to run Domoticz. Personally I prefer to start it as a daemon in forking mode, run as non-root user (pi), specify a PID file location writable for user pi. Log to dedicated log location writable for user pi. And start Domoticz on port http only and use Nginx proxy to the outside world using Let's encrypt sslcert with certbot.

Thus you can use two ways to start Domoticz, as type forking:
```
[Unit]
Description=Domoticz Home Automation
After=network.target

[Service]
Type=forking
User=pi
Group=pi
WorkingDirectory=/home/pi/domoticz
PIDFile=/var/run/domoticz/domoticz.pid
ExecStart=/home/pi/domoticz/domoticz -daemon -pidfile /var/run/domoticz/domoticz.pid -www 8080 -sslwww 0 -log /var/log/domoticz/domoticz.log
ExecReload=/bin/kill -HUP $MAINPID
KillMode=process
Restart=on-failure
RestartSec=30

[Install]
WantedBy=multi-user.target
Alias=domoticz.sh.service
```

Or in the foreground with default type Simple, by default the standard output will go to the journal, you can override them to null as you did.
```
[Unit]
Description=Domoticz Home Automation
After=network.target

[Service]
User=pi
Group=pi
WorkingDirectory=/home/pi/domoticz
StandardOutput=null
ExecStart=/home/pi/domoticz/domoticz -www 8080 -sslwww 0 -log /var/log/domoticz/domoticz.log
KillMode=process
Restart=on-failure
RestartSec=30

[Install]
WantedBy=multi-user.target
Alias=domoticz.sh.service
```

yfaykya:
Ok thx - I still think domoticz should probably still have some logic in there.
PS. I think I got my start script from :  http://www.domoticz.com/wiki/Preparing_a_Raspberry_PI_(Arch_Linux)

oohlaf:
Domoticz should have which logic where? I don't understand.

About the link, that's not my preferred way of starting, but it should work, as that page shows a minimal service with domoticz running in the foreground.

gizmocuz:
And the information in the link is very old ...

PR #1212, opened 2017-02-08
P1 Smart meter poll/update interval configurable

jvandenbroek:
Some users want to have more updates, despite the wattage difference. I changed the code so you can configure the interval in the hardware screen and left the wattage difference out. See also comments on this commit: https://github.com/domoticz/domoticz/commit/2f1cb584666301c912a6f5befdcd4e76cab65ff5#commitcomment-20797118


gizmocuz:
Thanks for the patch, some requests:
- please change the name/variable 'Poll Interval' to 'Rate Limiter', its not a poll interval, and it will confuse users as they now want to poll every second
- A default rate-limit of '0' is good, this means, do not perform any limitations, this because with 10 seconds, we can miss 1 packet (smart meter transmits a packet for example after 9.5 seconds... or the system is busy and misses a second, or...)

jvandenbroek:
I also renamed the vars to be more sensible, default is now 0 renamed to Rate Limit.

Edit: Oh btw I just rate limit the updates, so when a meter spits out after 9.5 secs, it would get the new values after 19 seconds when set to 10 secs. So not sure if 10 (or maybe 9?) is too bad as default value, but I've changed it to 0.

gizmocuz:
Indeed, it just gets the update after 19 seconds... but looking at the widget for 19 seconds is not nice
'updateinterval' is still not the same as a rate limiter...
i want an update every second... but the meter is not going to give this to me

https://en.wikipedia.org/wiki/Rate_limiting

jvandenbroek:
Alright, I just used them in code, but now have changed them also ;-) Probably best for consistency indeed..

gordonb3:
Sorry, but this is wrong:

	+				m_linecount=0;
	+				l_exclmarkfound=0;
				}

Reset of m_linecount and l_exclmarkfound should not be made conditional of whether an update is done. With this change the message buffer will keep filling up with data from multiple messages until it is full and then discard everything. While it won't cause Domoticz to break you're likely to see many CRC fails and probably never see any data being stored.

www/views/hardware.html
```diff
@@ -1204,11 +1204,20 @@ <h1 data-i18n="Settings">Settings</h1>
...
+						<strong><font class="warning-text"><span data-i18n="Should normally not be disabled!"></span></font></strong>
 					</td>
 				</tr>
 			</table>
 		</div>
+    <div id="divratelimitp1">
+			<br>
+			<table class="display" id="hardwareparamsratelimitp1" border="0" cellpadding="0" cellspacing="20">
+				<tr>
+					<td align="right" style="width:110px"><label for="ratelimitp1"><span data-i18n="Rate Limit"></span>Rate Limit (sec):</label></td>
```
gizmocuz: Not sure what you have done here
Think this should be:

<td align="right" style="width:110px"><label for="ratelimitp1"><span data-i18n="Rate Limit">Rate Limit</span>:</label></td>

then the value

on the right of the value

Underneath/right of the value a little more help

<span data-i18n="(Seconds) 0 = Disabled"></span>

Translation is already available for '(Seconds) 0 = Disabled'

jvandenbroek:
@gordonb3 Thanks, fixed.
@gizmocuz Changed it. Not yet added more help, will do.

gordonb3:
Per second ?  :smile: 

Code seems find to me like this. Just one thing: m_lastelectrausage doesn't appear to be used any more? Better delete it then in P1MeterBase.cpp and P1MeterBase.h. Obsolete code can lead to confusion.

jvandenbroek:
@gordonb3 Ha, well that might not be correct than, just removed it. Probably someone has a better explanation of what it's doing :-P Good catch, removed those lines.

gizmocuz:
Thanks both !

PR #1214, opened 2017-02-08
Fixed: Evohome DHW battery levels

Evsdd:
Hi,

This is a small update which fixes the bug which prevented the Evohome hot water sensor (DHW) battery levels from being updated.

Thanks,
Dan


gizmocuz:
Thanks!

PR #1216, opened 2017-02-09
Add GCM in configurable notification subsystems.

ultratoto14:
Did a cleaner pull request.
As stated in previous one, i added the entry for 
"Google Could Messaging" in the i18n files to be translated.


gizmocuz:
Thanks, It it possible to revert the language file changes, this is not needed

ultratoto14:
Hi gizmocuz, i think it's like a dialogue of the deaf. Is there a way we can have a quick chat together ?
Can you explain me why it's not needed as i added that:
`<span data-i18n="Google Cloud Messaging"></span>`
This need to be present in the i18n files in the other case, it will not be present.
I put it there and internationalize it as some languages uses different characters even if the words are the same. For example;  "Kodi Media Player": "Kodi媒体播放器"  or "Kodi Media Player": "Kodi Медиаплеер"
If really it's not needed then i should not put it in data-i18n and then it will be always "Google Cloud Messaging".

gizmocuz:
Thats why its best to communicate via the forum

Because you should add it like this:

`<span data-i18n="Google Cloud Messaging">Google Cloud Messaging</span>`

gizmocuz:
And later i will add the translation as explained before

ultratoto14:
It's taking time but i understood ;-)
Thanks a lot.

gizmocuz:
Thanks! Hmm just see that my last comment did not show clearly in this issue on github, i edited a bit
Going to add the translation now

PR #1217, opened 2017-02-09
Update CMakeLists.txt

ikkemaniac:
Just looking back at this; 
- fixed naming inconsistancy
- Isn't it better to use 'send_error' as message status? This will throw an error, continue the processing but skips generation?
The user will be forced to use the `-DINCLUDE_LINUX_I2C` or `-DINCLUDE_LINUX_SPI` arguments, to acknowledge support will be missing, after which all will be generated as usual.


CMakeLists.txt
```diff
@@ -133,23 +133,23 @@ option(INCLUDE_LINUX_I2C "Include I2C support" YES)
 IF (INCLUDE_LINUX_I2C)
 	CHECK_INCLUDE_FILES ("sys/types.h;linux/i2c-dev.h;linux/i2c.h" HAVE_LINUX_I2C_H)
 	IF (HAVE_LINUX_I2C_H)
-		message(STATUS "Building with I2C support")
+		message(SEND_ERROR "Building with I2C support")
```
gizmocuz, line 136: This is not an error, i would leave this to status

CMakeLists.txt
```diff
@@ -133,23 +133,23 @@ option(INCLUDE_LINUX_I2C "Include I2C support" YES)
...
 		add_definitions(-DHAVE_LINUX_I2C)
 	ELSE()
-		message(WARNING "I2C support disabled: headers not found!")
+		message(SEND_ERROR "I2C support disabled: headers not found!")
 	ENDIF (HAVE_LINUX_I2C_H)
 ENDIF (INCLUDE_LINUX_I2C)
 
-option(INCLUDE_SPI "Include SPI support" YES)
-IF (INCLUDE_SPI)
+option(INCLUDE_LINUX_SPI "Include SPI support" YES)
```
gizmocuz, line 143: I would strip the work "LINUX" as maybe in the feature we could add support for this as well

https://www.i2ctools.com/

ikkemaniac: I just made an error while copying the existing var `HAVE_LINUX_....` so that was/is already in the file. This update is just to keep variable names consistent. With your train of thought 'linux' should be removed from `HAVE_LINUX_...` as well.

CMakeLists.txt
ikkemaniac: ok

gizmocuz: Please change this back to status, you can see what this does in our build system (it errors out... but it should not)

CMakeLists.txt
```diff
@@ -133,23 +133,23 @@ option(INCLUDE_LINUX_I2C "Include I2C support" YES)
 IF (INCLUDE_LINUX_I2C)
 	CHECK_INCLUDE_FILES ("sys/types.h;linux/i2c-dev.h;linux/i2c.h" HAVE_LINUX_I2C_H)
 	IF (HAVE_LINUX_I2C_H)
-		message(STATUS "Building with I2C support")
+		message(SEND_ERROR "Building with I2C support")
 		add_definitions(-DHAVE_LINUX_I2C)
 	ELSE()
-		message(WARNING "I2C support disabled: headers not found!")
+		message(SEND_ERROR "I2C support disabled: headers not found!")
```
gizmocuz, line 139: This should be a warning as well
Default people do not disable this option, and if you build on a normal system, it will fail because there is no SPI/I2C

gizmocuz (commented):
Actually, the way it was... was OK

PR #1218, opened 2017-02-09
OpenWebNet: implementation of request the status of an area's devices

Spugna85:
When MyHome Bticino sent a AREA command (for example send OFF LIGHT for area 2), no message of update status is sent via OpenWebNet.
So need to request the devices status of this area.
It was then implemented this request every time it receives a  type AREA.command.

It has also been implemented periodic scanning of all devices's state every 5 minutes (only a CEN type command is sent)


hardware/OpenWebNetTCP.cpp
```diff
@@ -540,33 +538,41 @@ void COpenWebNetTCP::UpdateDeviceValue(vector<bt_openwebnet>::iterator iter)
     string value = iter->Extract_value(0);
 	string sInterface = iter->Extract_interface();
     string devname;
-	int app_value;
+	int app_value, int_where;
```
gizmocuz: maybe an idea to use the following coding style

[int iAppValue, iWhere](url)

hardware/OpenWebNetTCP.cpp
```diff
@@ -895,14 +905,14 @@ bool COpenWebNetTCP::sendCommand(bt_openwebnet& command, vector<bt_openwebnet>&
 /**
     automatic scan of automation/lighting device
 **/
-void COpenWebNetTCP::scan_automation_lighting()
+void COpenWebNetTCP::scan_automation_lighting(int cen_area)
```
gizmocuz: i think this can be const
`scan_automation_lighting(const int cen_area)`

Maybe with more functions?

gizmocuz (commented):
Thanks

gizmocuz:
Thanks!

PR #1220, opened 2017-02-11
Fix Seg Fault on Notification

dnpwwo:
If someone has disabled plugins a notification resulted in a seg fault due to entries in the plugins hardware map pointing to NULL plugin structures


gizmocuz:
Thanks!

PR #1225, opened 2017-02-12
Add API to change events status without resending everything.

ultratoto14:
Hi Gizmocuz,
Just added a new option in the event api to enable/disable without resending everything.
It has been requested by galadril for the app:
[https://www.domoticz.com/forum/viewtopic.php?t=12557](url)

param is updatestatus to avoid any confusion as it updates only the status.


gizmocuz:
Thanks!

PR #1226, opened 2017-02-13
Support symlinks for plugin and theme directories

oohlaf:
It bothered me that I could not symlink from plugins/<name> to a git checkout for a specific plugin in a different directory. This changes that.


gizmocuz:
Thanks!

PR #1228, opened 2017-02-13
Differentiate "Add Plan" from Floor plan management for i18n translation

Ben0it:
By changing data-i18n value from "Add Plan" to "Add Timer Plan"

### DIFF
--- a/www/views/timerplan.html
+++ b/www/views/timerplan.html
@@ -13,7 +13,7 @@
 	<table class="bannav" id="bannav" border="0" cellpadding="0" cellspacing="0" width="100%">
 	<tr>
 		<td align="left"></td>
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thats not possible as we added 'Add Plan' in i18n, is this a big issue ?

Ben0it:
I thought i18n strings were automatically extracted to Transiflex from the source files (.html mostly).

For **Floor plans** in French, translation "**Ajouter un plan**" is fine. However for **Timer plans** (**"Profils de gestion"**), the action would rather be "**Ajouter un profil**". However "Add Plan" is common to both timer and floor management in Transiflex.

xztraz:
same goes for swedish
"floor plans" = "planlösningar"
"timer plans" = "tidsscheman"

and the word "plans" = "ritningar" (kind of like drawings) doesn't match timer plan very well.

maybe timer plans should be separated from physical drawings(plans) in menu

PR #1235, opened 2017-02-16
Night and day calculation

gordonb3:
There was an oversight when adding DST change awareness to the code. In EventSystem the current time is corrected for such a change, but the sunrise and sunset values are not. They should either all be corrected or none of them.


gizmocuz:
Thanks!

PR #1241, opened 2017-02-17
SBFspot import data fix

novorum:
Archieved data is saved as Wh while domoticz uses kWh.
kWh = Wh * 1000


gizmocuz:
What is your kWh divider in the settings ?

novorum:
Default setting: 1000

gizmocuz:
Okey, as i suspect a lot of users are using SBFSpot with domoticz, could you create a forum post to ask if users are having issues like you have ?
(Maybe it is a SBFSpot configuration ?)

gizmocuz:
Thanks!

PR #1242, opened 2017-02-18
Python plugins

dnpwwo:
Add therostat support


gizmocuz:
Thanks!

hardware/plugins/PluginMessages.h
```diff
@@ -140,6 +142,21 @@ namespace Plugins {
 		DisconnectMessage(int HwdID) : CPluginMessage(PMT_Disconnect, HwdID) {};
 	};
 
+	class CommandMessage : public CPluginMessage
+	{
+	public:
+		CommandMessage(int HwdID, int Unit, const std::string & Message, const int level, const int hue) : CPluginMessage(PMT_Command, HwdID, Unit, Message, level, hue)
+		{
+			m_fLevel = -273.15;
```
oohlaf, line 150: Ah, this is a bug! I was wondering why my onCommand method was always called with value -273.15! :)

dnpwwo: Damn floating point inaccuracy!
-273.15 got stored as -273.149999999. So much for absolute zero.

oohlaf: Ah, I see what you did, comparison to that value in both places
Here is a fix
https://github.com/domoticz/domoticz/pull/1245

PR #1244, opened 2017-02-21
XiaomiGateway: fixing door sensors

corbinmunce:
Broke this previously sorry.


gizmocuz:
Thanks!

marcotini:
It does not work even with this change, when the log shows a heartbeat occasionally by a false signal on for the doors and for the motion sensor.

corbinmunce:
@marcotini, this was for a different problem than what you are experiencing.  Can you setup a test environment for me to look at? PM me the details on the forum if so.

marcotini:
@corbinmunce I already have solved https://github.com/marcotini/domoticz/commit/254c29b97516da937ae36614e252e3560104fecf

PR #1245, opened 2017-02-21
Check for the nearest floating point representation of -273.15

oohlaf:


dnpwwo (changes requested):
This fix is initialising the variables apart from m_fLevel twice. The base class constructor is already doing the others.

I am going to push a change tonight and can fix ths as well, I would suggest you withdraw this pull request

oohlaf:
Right, I removed those lines from the request

oohlaf:
already merged in dnpwwo his changes

PR #1246, opened 2017-02-21
Fix CommandMessage error

dnpwwo:
Split Python Objects into a separate file


hardware/plugins/DelayedLink.h
```diff
@@ -253,7 +253,7 @@ namespace Plugins {
 						while (!shared_lib_ && (ent = readdir(lDir)) != NULL)
 						{
 							std::string filename = ent->d_name;
-							if (dirent_is_directory(szLibrary, ent) && (filename.length() > 2))
+							if ((ent->d_type == DT_DIR) && (filename.length() > 2))
```
oohlaf, line 256: Please keep dirent_is_directory, it supports symlinks also, not only directories

oohlaf (changes requested):

hardware/plugins/DelayedLink.h
oohlaf: Also, better to compare with "." and ".." because the directory can also be a single or two letter directory other than the dots.

filename.compare(".") != 0 && filename.compare("..") != 0

oohlaf (changes requested):
Please keep symlink support
Would be better to check against dots explicitly not on length

dnpwwo:
I've changed the filename check and it will be in a later pull request.

Sym link support has not been removed.  This code only runs on non-Win32 platforms and so only DT_DIR is checked in dirent_is_directory.  I had real problems compiling the delayed link code on Linux due to dirent_is_directory which is why I took it out.

oohlaf:
Huh? I don't understand.

You get symlink support for free as I added it some time ago to dirent_is_directory and it is specifically for non-Win32 platforms as the Linux platform has symlinks, not Win32.

I actually tried your Python-Plugin branch and added back the dirent_is_directory and it compiles fine under Linux. I'll create a pull request against your branch that you can merge easily and try for yourself. I'll leave the filename check up to you....

oohlaf:
https://github.com/domoticz/domoticz/pull/1254 which you can pull into your own branch and incorporate with your other changes

PR #1247, opened 2017-02-21
Significant enhancement to EcoDevices hardware support 

DebugBill:
Hi,

This is my first contribution to Domoticz, sorry if I don't follow some of the rules.

The current support for EcoDevices hardware only supports one smart meter and one subscription plan (peak hours/low hours) when several other exist and are as popular.
 
This version adds support for any or all 4 counters and several electricity subscription plans. Counters 1 and 2 are for French (Teleinfo) electricity smart meters, counters 3 and 4 are simple pulse counters for gaz, water, fuel or anything similar.  

Further work is to be done:
- add support for EJP and Tempo subscription plans (I would need data samples) 
- suppress dependency on code from P1SmartMeter in order to benefit from additional data available (Tri phases subscriptions, alerts when intensity comes close to subscribed limit, tarif switches to higher prices (EJP...)

It would also be interesting to mutualise the teleinfo protocol decoder that exists in both Teleinfo.cpp and Ecodevices.cpp. They are quite different event though they have the same purpose. I'm not sure my C++ an Domoticz knowledge are enough to undertake such a task though.

Thanks.


gizmocuz:
Thanks

Vikingfr:
I'm using Tempo, what kind of data do you need ?
teleinfo is like this:
```
ADCO 040122120674 52
OPTARIF BBR( 83
ISOUSC 45 63
BBRHCJB 113767123 60
BBRHPJB 226386290 80
BBRHCJW 001967783 91
BBRHPJW 003078654 32
BBRHCJR 000858434 77
BBRHPJR 001151206 74
PTEC HPJB 80
DEMAIN ---- 34
IINST 010 88
IMAX 053 71
PAPP 02420 41
HHPHC Y 68
MOTDETAT 003300 72
```
"DEMAIN" can be "BLEU" "BLAN" "ROUG" or "----" when not defined

(sorry if bad english, can we use french ?)

DebugBill:
Hi,

French is OK with me but might confuse other readers on the thread. You can mail me at "blaise at thauvin.org" (replace at with @)

I would need several examples of the output of http//192.168.xx.xx/api/xdevices.json?cmd=10 (replace with your own EcoDevices' IP of course) in order to see the various values that variable T1_PTEC can take. This variable contains the current price/period. We should be able to guess the all six values from 2 or 3 examples.

Also I am not quite sure how to implement this feature. If I understand well, TEMPO means you have 6 different prices for 6 different periods (peak and off-peak on blue, red and white days). This means I need to keep track of 6 independant counters. There is no standard object to do that in Domoticz. The closest match is P1SmartMeter, but also it has 2 counters for peak/off-peak, it seems that it does not make any real use of it and simply sums them.

I see two choices:
- Create one single counter and sum everything into it
- Create six individual devices, each of them being one counter, and maybe a seventh one for the sum.

I plan to use the second solution for simple peak/off-peak subscriptions, creating 3 devices for one smart meter.

I have no clue yet how I can use the "DEMAIN" variable. I'm not sure it is in the JSON output. This will be for a later evolution I guess. How can DEMAIN be "not defined"? I thought the calendar was defined months in advance.

Tell me what you think is best

Blaise

Vikingfr:
I'm not using a eco devices, but another teleinfo device (wifinfo, from Charles Hallard).

You can get some information here: http://www.touteladomotique.com/index.php?option=com_content&id=985:premiers-pas-avec-leco-devices-sur-la-route-de-la-maitrise-de-lenergie&Itemid=89
```
T1_PTEC
Période tarifaire en cours :TH = Toutes Heures, HC = Heures Creuses, HP = Heures Pleines, HN = Heures Normales, PM = Pointe Mobile, HCJB / HPJB / HCJW : HPJW / HCJR / HPJR = Tempo
```

"DEMAIN" is set at about 8PM and reset at 6AM by the electricity meter . There is no calendar. Energy Provider set it at 12AM each day for the next day on is website. Color depends on weather (cold=red/higher price, hot=blue/cheaper electricity). I use this information to choose best time to start water heater (it's cheaper to heat during blue day than white night)

DebugBill:
Sorry, I'm working on the EcoDevices hardware support, not generic Teleinfo (even though that would probably be a good idea). As I don't have your hardware I would not be able to test anything.

How do you plan to read data from your device?  I suspect it is not compatible with the "Teleinfo EDF USB" hardware in the list either? In that case you have to build everything from scripts and virtual devices.

Or you can consider investing in EcoDevices (I don't work for them :-). I track electricity, gaz and water with it and it works like a charm.

Blaise

Vikingfr:
I'm working on convert Wifinfo output to Eco Device output. There is already a json and xml output on "/json" and /xml" urls. I've already add "raw" output (data without formating, like serial teleinfo) . I will add code to handle "/api/xdevices.json?cmd=10" and add "T1_" prefix, so the wifinfo will be seen as first counter of a eco devices.

DebugBill:
This is a good way to start, but there is a little bit more work to do. The code checks quite carefully that the frame is a genuine EcoDevices frame, so you have to simulate it very precisely. This implies converting the data to the proper format (add dots to complete the fixed size text fields, reformat decimal representation....) and send data for all 4 counters (zeros will do for unused ones...).

I suggest you start with simulating a basic subscription and sum all six counters into one. Once you have this working we can go further. Alternatively, you could use six different ports to report all six counters and have your reporting working with no change to the current code.

Here is a sample outpout of my own device so you can base your code on it. As you can see there is no subscription on Teleinfo2 (T2_PTEC="----") and I have two additional counters C1 and C2. All fields must be sent.

{"product":"Eco-devices","T1_PTEC":"TH..","T1_PAPP":510,"T1_BASE":8191264,"T2_PTEC":"----","T2_PAPP":0,"T2_BASE":0,"INDEX_C1":517757,"INDEX_C2":18432840}

PR #1249, opened 2017-02-21
Added support for the Philips hue motion sensor

JordyThien:
Added Philips Hue sensors class
Motion sensor will be added to the database and updates will be send

Tested on Windows


JordyThien:
Closing because it's not building for linux targets. Need to get that up and running first.

gizmocuz:
You can keep pushing to your fork and it will update the pull request...

JordyThien:
Ok, I'll leave it open in that case :)

gizmocuz:
Sorry, our GCC version does not support 'nullptr', please use NULL

JordyThien:
Yeah, I thought I fixed them all. Update incoming :)

gizmocuz:
Thanks!

gizmocuz:
Could you also check if the 'state' of a sensor is changing ? Currently i have the same switch value every 5 minutes for the ambient light sensor

PhilipsHue.cpp line 882

if (asensor.m_state.m_lastupdated == hsensor.m_state.m_lastupdated)

I seems the last update timestamp has changes, but the value stays the same
i would do this for every switch

gizmocuz:
Maybe for every state, also a last_state, and just before sending the state, compare it with the last_state
if different, send the switch, and set last_state=state
Of course, when the hardware is started for the first time, it should always send it

PR #1250, opened 2017-02-21
Add {deviceid}, {status} and {level} expansion to switch action scripts

dwmw2:
There's no good reason for the first two to be for evohome only. And
without {level} it's distinctly non-trivial to make a script which sets
a dimmer (or volume control, in my case).

### DIFF
--- a/main/SQLHelper.cpp
+++ b/main/SQLHelper.cpp
@@ -3418,7 +3418,7 @@ uint64_t CSQLHelper::UpdateValueInt(const int HardwareID, const char* ID, const 
 		std::string lstatus="";
 
 		result = safe_query(
-			"SELECT Name,SwitchType,AddjValue,StrParam1,StrParam2,Options FROM DeviceStatus WHERE (ID = %" PRIu64 ")",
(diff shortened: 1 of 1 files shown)


main/SQLHelper.cpp
```diff
@@ -3433,6 +3433,7 @@ uint64_t CSQLHelper::UpdateValueInt(const int HardwareID, const char* ID, const
 			GetLightStatus(devType, subType, switchtype,nValue, sValue, lstatus, llevel, bHaveDimmer, maxDimLevel, bHaveGroupCmd);
 
 			bool bIsLightSwitchOn=IsLightSwitchOn(lstatus);
+			std::string slevel = sd[6];
```
gizmocuz, line 3436: is sLevel not an integer ? Maybe directly cast it to an int wit atoi(string.c_str())
and use it below also as an integer ?

dwmw2: No, it's only ever used as a string, in the search/replace. If we do update the level, we then convert that integer into a string for this.

gizmocuz:
Thanks!

PR #1254, opened 2017-02-22
Python-Plugins dirent usage in DelayedLink

oohlaf:
Make use of dirent_is_directory again in DelayedLink

### DIFF
--- a/hardware/plugins/DelayedLink.h
+++ b/hardware/plugins/DelayedLink.h
@@ -13,8 +13,9 @@
 #include <structmember.h> 
 #include <frameobject.h>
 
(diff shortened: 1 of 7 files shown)


oohlaf:
Hmm, no response on this...
Now that you merged your branch into main without this, shall I rebase against main?

dnpwwo:
Go for it.

I don't know why it wouldn't compile for me but if you can make it work then using common code is always a good think.

oohlaf:
Well, that's easy as it already compiled :)
As you can see by the green marks. You needed to rearrange the order of the headers files include.

gizmocuz:
Thanks!

PR #1261, opened 2017-02-26
New icon for Water meter

DebugBill:
Add new icon Water48.png showing a water meter

This makes it more consistent with the gaz and generic meters icons used for function SendCustomMeter()


gizmocuz:
Thanks but:
- you should have changed the Water48_On/Off images
- I like my water icon better... in one quick view you can see 'water', your meter could be anything

DebugBill:
Hi,

I didn't change the ON icon because the Wiki says there should always be 4 icons (On, Off, Default and small) and the blue/white water drop looked fine to me for a solenoid valve (or anything that goes ON/OFF). I created the "default" icon to be used only in conjunction with the SendMeterSensor function which already uses generic and gaz meter icons. The EcoDevice hardware itself is designed to show water, electricity and gaz meters (and also fuel.... which is yet to be added), so it looked nice on my interface.

Anyway, this is your software.... so your decision

Blaise

gizmocuz:
Thats incorrect. my water meter is using a RFXMeter, which you can 'edit' and set to 'water' (you actually should do this)
Then you have a water icon instead of a electric meter appliance

If this is not so for ecodevice, maybe this is using the wrong sensor type and should be changed?

DebugBill:
Right, and this is exactly how I use it. When creating the hardware, the devices first show up with their default icon, and I then use the edit button to swith one to Gaz and the other to Water. And the fact that the units goes to m3/m3 for gaz and l/m3 for water is perfect. This is where I'd like to add fuel as the units would then be l/l (or gl/gl in the States maybe). I only chaged the blue drop to the meter in the edit box.

The RFX meter seems to me as an almost perfect match for my need. The "EcoDevices" provides two pulse counters which can in fact monitor anything, but water, gaz and fuel are the most common uses I guess. 

An even better match would be a mix of RFXMeter and General/kWh:
From kWh I would use the daily and total indexes along with the current flow in Watt, VA, l/mn or m3/mn
and from RFX I would keep the ability to change icons and units depending on what I am really measuring.

The EcoDevices provides the flow information, which I don't display currently.

I'm ready to update my code if you can suggest a better sensor to be used. I tried "SendCustom" and "SendUsage" but they were not as good a fit as "SendMeter".

gizmocuz:
I think you have to use 'SendMeterSensor' as this is the only type that outputs a RFXMeter sensor

DebugBill:
As you will see in the other pull request once I'm done with translating, this is the one I chose when reimplementing using standard helpers. Looks like we get to the same conclusion.

Thanks for your help

gizmocuz:
True ;) ... so then the change for a new water icon/meter is not needed anymore correct?

DebugBill:
I still prefer my meter icon for meters and your drops for valves, but this is a mater of taste. I keep it active on my own private branch

DebugBill:
I've created a Wiki page for the Ecodevices hardware. You can see how the icon looks like on my setup from the screenshop I posted there:
http://www.domoticz.com/wiki/Eco_Devices_Via_LAN_Interface

Blaise

PR #1262, opened 2017-02-26
Minor changes

DebugBill:
Missing space in tinyxpath/CMakeLists.txt
One more line in gitignore so "make install" does not make install_manifest appear in "git status"


gizmocuz:
Thanks!

PR #1263, opened 2017-02-26
[BleBox] fixes for gatebox and switchboxd

szczukot:


hardware/BleBox.cpp
```diff
@@ -17,11 +17,9 @@ const _STR_DEVICE DevicesType[TOT_TYPE] =
 	{ 1, "shutterBox", "Shutter Box", int(pTypeLighting2), int(sTypeAC), int(STYPE_BlindsPercentageInverted), "shutter" },
 	{ 2, "wLightBoxS", "Light Box S", int(pTypeLighting2), int(sTypeAC), int(STYPE_Dimmer), "light" },
 	{ 3, "wLightBox", "Light Box", int(pTypeLimitlessLights), int(sTypeLimitlessRGBW), int(STYPE_Dimmer), "rgbw" },
-	{ 4, "gateBox", "Gate Box", int(pTypeLighting2), int(sTypeAC), int(STYPE_Dimmer), "gate" },
+	{ 4, "gateBox", "Gate Box", int(pTypeGeneral), int(sTypePercentage), 0, "gate" },
```
gizmocuz: could you remove all 'int' casts from this block ? It's not needed and it makes it quite unreadable

hardware/BleBox.cpp
```diff
@@ -273,192 +270,249 @@ bool BleBox::WriteToHardware(const char *pdata, const unsigned char length)
...
+					else
+					{
+						state = "0";
+					}
 
-				Json::Value root = SendCommand(IPAddress, "/s/" + state);
-				if (root == "")
-					return false;
+					Json::Value root = SendCommand(IPAddress, "/s/" + state);
+					if (root == "")
```
gizmocuz: Did you intend to do if (root.empty()) ?

hardware/BleBox.cpp
szczukot: You're right

hardware/BleBox.cpp
szczukot: sure-it is better way

gizmocuz:
Thanks!

PR #1264, opened 2017-02-26
Philips hue sensor lux support

JordyThien:
Requested by the community
Also fixed the temperature (no decimal points were shown, now it is)

### DIFF
--- a/hardware/PhilipsHue/PhilipsHue.cpp
+++ b/hardware/PhilipsHue/PhilipsHue.cpp
@@ -904,18 +904,19 @@ bool CPhilipsHue::GetSensors(const Json::Value &root)
 				}
 				else if (hsensor.m_type == SensorTypeZLLTemperature)
 				{
-					ReportTemperature(sID, float(hsensor.m_state.m_temperature / 100.0f), device_name, hsensor.m_config.m_battery);
+					SendTempSensor(sID, hsensor.m_config.m_battery, float(hsensor.m_state.m_temperature / 100.0f), device_name);
 				}
 				else if (hsensor.m_type == SensorTypeZLLLightLevel)
 				{
 					InsertUpdateSwitch(sID, STYPE_Dusk, hsensor.m_state.m_dark, device_name, hsensor.m_config.m_battery);
+
 					double lux = 0.00001;
 					if (hsensor.m_state.m_lightlevel != 0)
 					{
 						float convertedLightLevel = float((hsensor.m_state.m_lightlevel - 1) / 10000.00f);
 						lux = pow(10, convertedLightLevel);
 					}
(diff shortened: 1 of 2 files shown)


hardware/PhilipsHue/PhilipsHue.cpp
```diff
@@ -936,14 +943,32 @@ void CPhilipsHue::ReportTemperature(const int NodeID, const float temperature, c
 	tsen.TEMP.id1 = (sID & 0xFF00) >> 8;
```
gizmocuz: could you please remove this, and use the default way of reporting a (temperature) sensor ?
Just call

`void SendTempSensor(const int NodeID, const int BatteryLevel, const float temperature, const std::string &defaultname);
`

These functions are specified in DomoticzHardware.cpp/h

hardware/PhilipsHue/PhilipsHue.cpp
```diff
@@ -936,14 +943,32 @@ void CPhilipsHue::ReportTemperature(const int NodeID, const float temperature, c
...
+	int at10 = round(abs(temperature*10.0f));
 	tsen.TEMP.temperatureh = (BYTE)(at10 / 256);
 	at10 -= (tsen.TEMP.temperatureh * 256);
 	tsen.TEMP.temperaturel = (BYTE)(at10);
 
 	sDecodeRXMessage(this, (const unsigned char *)&tsen.TEMP, Name.c_str(), BatteryLevel);
 }
 
+
+void CPhilipsHue::ReportLux(const int NodeID, const float lux, const string &Name, uint8_t BatteryLevel)
```
gizmocuz: could you please remove this, and use the default way of reporting a (lux) sensor ?
Just call

`void SendLuxSensor(const int NodeID, const int ChildID, const int BatteryLevel, const float Lux, const std::string &defaultname);
`

These functions are specified in DomoticzHardware.cpp/h

gizmocuz (commented):
your fast! ;) two requests

hardware/PhilipsHue/PhilipsHue.cpp
JordyThien: Good point! Didn't knew about those functions. Updated the code

gizmocuz:
Great! Thanks!!

roelbeun:
Curious, when will this commit be available in the beta or even the master? Thanks for your work!

JordyThien:
@Ruvetuve it's already in the beta release. You can download the latest beta on http://domoticz.com/downloads/ or run ./updatebeta

roelbeun:
Great, thanks! Hope it comes to the stable release soon, in the mean while ill test it on a separate system.

PR #1267, opened 2017-02-27
Added more functionality for Limitlessled (RGBWW, RGBW) v6 and added direct disco mode and nightlight for v4/5

chrisverhoef:


gizmocuz:
If possible, could you make the commands/status match ? For example

case Limitless_DiscoM1:
lstatus="Set to Disco 1";
break;

if (switchcmd == "Disco Mode 1")
cmd = Limitless_DiscoM1

you see the above don't match, i would change it to "Disco Mode 1"

main/RFXNames.cpp
```diff
@@ -1478,32 +1478,32 @@ void GetLightStatus(
 		case Limitless_NightMode:
 			lstatus="Set to Night";
 			break;
-		case Limitless_DiscoM1:
-			lstatus="Set to Disco 1";
+		case Limitless_DiscoMode_1:
+			lstatus="Set to Disco Mode 1";
```
gizmocuz: could you remove the prefix 'Set to '?

chrisverhoef: I will do.

gizmocuz:
Thanks!

chrisverhoef:
And thanks to deennoo for testing!

PR #1268, opened 2017-02-27
Eco devices, further enhancements and switch to standard helper functions

DebugBill:
- All calls to sDecodeRX suppressed and replaced with standard helpers (SendKwhMeter, SendMeterSensor....)
- All 4 main French electricity rates now supported (Basic, Tempo, EJP and "Heures Creuses")
- New functionality: a message device returns the current rate for multiple rates contracts (EJP, HC and Tempo)
- Pulse counters now rely on SendMeterSensor which does a much better job than its P1SmartMeter based counterpart when tracking water or gaz consumption.

The EcoDevices device and the Teleinfo protocol are specifically French, I coded all messages and default
names in French as it is very unlikely that anybody non French speaking would ever read them. Is that OK?


hardware/EcoDevices.h
```diff
@@ -19,14 +19,96 @@ class CEcoDevices : public CDomoticzHardwareBase
...
+
+		Teleinfo m_teleinfo1, m_teleinfo2;
 
 		void Init();
 		bool StartHardware();
 		bool StopHardware();
 		void Do_Work();
 		void GetMeterDetails();
 };
+
```
gizmocuz, line 86: Could you please translate this to English, or both languages ?

DebugBill: Translation done.

Also tried to fix include issues that may be the cause of tests failures while it compiles fine on my Raspberry. Now waiting for tests to start

gizmocuz:
Thanks!

PR #1269, opened 2017-02-27
Security panel : redirection to login page if status query return "unauthorized"

PicoI2:
Security panel : redirection to login page if status query return unauthorized (probably because session is not valid anymore).
Instead it just show "no connect", as if domoticz is down.


gizmocuz:
Thanks!

PR #1278, opened 2017-02-28
Invalid sessions on password update

PicoI2:
Hello, here is a new pull request.

I have notice that I had changed the password of my GF, but she did not had to log again.

This pull request fix it, and other security issues.
Sometimes it could seem overkill, as when a user is deleted, but you could always imagine use case where you delete a user and then create a user with the same login.

Feel free to rename method or variables if you prefer, or event to reject it, I will not be upset !

You should add in contributing.md your favorites coding rules, I tried to guess it looking at the code, but it's not always that easy.

One more thing, why session are limited in time ? Would you accept a pull request to extend their validity, or to make it configurable ?

### DIFF
--- a/main/WebServer.cpp
+++ b/main/WebServer.cpp
@@ -5962,6 +5962,14 @@ namespace http {
 					idx.c_str()
 					);
 				LoadUsers();
+
+				// Invalid user's sessions if password has changed
+				std::vector<std::vector<std::string> > result;
+				result = m_sql.safe_query("SELECT Password FROM Users WHERE (ID == '%q')", idx.c_str());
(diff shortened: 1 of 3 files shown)


gizmocuz:
Thank you for this great patch !

I've got a pretty standard coding style, but there are other users that also contribute code (c# users that are going c++ way.., or scripters...)

I very much agree about the contributing.md file, unfortunately i'm not that advanced with it, and i need to create some time for it.
Maybe you could help ? (we could PM on the forum/email about it)
this could help:
https://www.npmjs.com/package/generate-contributing

The session timeout should extended after almost any request... that was in the beginning (the initial idea)
And then we have a 'remember me' that should keep the session

PR #1290, opened 2017-03-02
P1 Smart Meter add Voltage meter (DSMRv5)

jvandenbroek:
DSMRv5 supports line voltage read-out, which I thought would be handy to monitor. In case of multiple phases, all are added (L1, L2 and L3). Tested with only L1, used info from DSMRv5 datasheet, so other phases should work fine.

Now that I recheck the code, I think the other phases actually won't work. Will change some logic.
Edit: should be fine now, compiling takes forever on my RPi 1, so not tested yet..


gizmocuz:
Thanks!, now i wish i had this new meter !
Instead of new types and so, when you receive the voltage value, just use

SendVoltageSensor(const int NodeID, const int ChildID, const int BatteryLevel, const float Volt, const std::string &defaultname)

This way, all code stays clean, and you do not need to modify the html files also

jvandenbroek:
Well I acutally started with SendVoltageSensor, however there are several changes I made which needed new types etc: 
- 1 decimal instead of 3
- P1 Smart Meter as device instead of General (just as Power and Gas)
- Same ID and Unit style (instead of 00000x and always 1 for Unit).

The SendVoltageSensor did work, but it just felt wrong IMHO. Hopefully you'll see this is for now the cleaner method, until P1 is completely rewritten ;-)

gizmocuz:
Please use SendVoltageSensor
We should not create another sensor type... we have to many already

jvandenbroek:
Changed the code to use SendVoltageSensor and only displaying 1 decimal in case of P1 Smart Meter hardware (so essentially what I wanted but with much less code :)

gizmocuz:
Thanks!!

PR #1292, opened 2017-03-03
Added Relaynet hardware

hvbommel:
Relay-Net V5.7 relay and input module support. This low cost module can be bought for example at Ali express, Ebay and other sources. Build using VisualStudio2015 & Eclipse-Mars. Tested on Windows, Ubuntu and Raspberry Pi.


hardware/RelayNet.cpp
```diff
@@ -0,0 +1,748 @@
...
+	memset(&Packet, 0, sizeof(RBUF));
+
+	if (m_HwdID > 0xFF)
+	{
+		id3 = (m_HwdID >> 8) && 0xFF;
+	}
+
+	if (m_HwdID > 0xFFFF)
+	{
+		id3 = (m_HwdID >> 16) && 0xFF;
```
oohlaf, line 325: I've been reviewing compiler warnings, and by judging from the code you should use a bitwise operation not a logical and. Change && to & (same for id3 above).
Also, shouldn't this be id2 (hence the >> 16)?
Why not the same logic for id1 also?

hvbommel:
I will fix it asap.

PR #1293, opened 2017-03-03
EcoDevices: switch from JSON API to XML. Provides additional data and functionalities

DebugBill:
JSON API in EcoDevices lacks some variables available in the raw Teleinfo Protocol

This version uses XML API instead of JSON in order to provide additional functionality
- Current (Amps) is now reported, for all 3 phases when available
- Alerts are raised when over-consumption arises
- Current rate is displayed for multiple rate contracts (Tempo, Heures Creuses and EJP)
- Next day rate is displayed, user can set up alerts (Tempo contract)
- Water and Gas flow in l/mn are reported

Code has only been tested with basic electricity subscription and may contain bugs when connected to other, more elaborate contracts. I try to contact usersto get some feedback.


gizmocuz:
Maybe best to wait then ?
Bdw, if you miss functionality in the JSON output, could you not request to add this ?
Because i like the json more, and you now also introduce boost xml parsers (we have tinyxml in domoticz, maybe best to use this?)

DebugBill:
If you prefer tinyXML to Boost no problem, I guess I can rewrite this part.

I've tried to contact users on the French and English forums since my first contribution. Up to now I've had no answer or comment, so I'm affraid waiting to get more feedback may take a while. Also, including the patches in the beta version is probably a good way to get feedback. I doubt people will pull from my repository to test my branch. 
This is also why I wish to mutualise similar code between Teleinfo.cpp and EcoDevices.cpp. It would increase the Teleinfo user and testers base. Work on this is currently underway and visible on my repository.

I prefer JSON to XML too, but clearly the manufacturer decided to output only a subset of the available information that way and the full raw output through XML I could probably ask them to change their code, and potentialy break current users but I have no clue whether they would do it and when. Also it would mean users need to upgrade their firmwares... I chose the fast and easy path :-)

gizmocuz:
Okey ;)
If possible, i would prefer 'tinyxpath' thats included in domoticz, if it's not possible, no problem too, as long as we do not need a new boost library

DebugBill:
Done! 
I must say I had a hard time making tinyxpath work :-). It seems to be only used once in LuaHandler and I was lacking examples.

DebugBill:
Build failed for some strange reason on MacOS. It seems the Python header file is missing?

hardware/EcoDevices.cpp
```diff
@@ -122,365 +119,354 @@ bool CEcoDevices::WriteToHardware(const char *pdata, const unsigned char length)
...
 
-		// Peak/Off-Peak subscription scheme
-		else if ((m_teleinfo2.PTEC == "HC..") || (m_teleinfo2.PTEC == "HP.."))
+	// XML format changes dramatically between firmware versions. This code was developped for version 1.05.12
+        using namespace TinyXPath;
+	m_status.version = S_xpath_string(XMLdoc.RootElement(),"/response/version/text()").c_str();
+
+	#ifdef DEBUG_EcoDevices
+	std::string EncodedXML = sResult;
+        using boost::algorithm::replace_all;
```
gizmocuz: Can you use our standard stdreplace function for this ?
Also maybe, as you seem to replace the same characters as in 'CNotificationBase::MakeHtml'
Could you put this function in the Helper.cpp/h file, and use this one ? (Also remove the CNotificationBase::MakeHtml function then of course ;)

hardware/EcoDevices.cpp
```diff
@@ -122,365 +119,354 @@ bool CEcoDevices::WriteToHardware(const char *pdata, const unsigned char length)
...
+        replace_all(EncodedXML, "&",  "&amp;");
+        replace_all(EncodedXML, "\"", "&quot;");
+        replace_all(EncodedXML, "\'", "&apos;");
+        replace_all(EncodedXML, "<",  "&lt;");
+        replace_all(EncodedXML, ">",  "&gt;");
+        _log.Log(LOG_NORM, "DEBUG: XML output for /status.xml\n%s", EncodedXML.c_str());
+	#endif
+
+	major = boost::lexical_cast<int>(m_status.version.substr(0,m_status.version.find(".")).c_str());
```
gizmocuz: Are you 100% sure we always find a '.' ?
If yes, could you remove the boost:lexical_cast dependency (also the included header), and use
atoi(stdstring.c_str()) for int

hardware/EcoDevices.cpp
```diff
@@ -122,365 +119,354 @@ bool CEcoDevices::WriteToHardware(const char *pdata, const unsigned char length)
...
+			m_status.pflow2 = m_status.flow2;
+			m_status.time2 = atime;
+			SendMeterSensor(m_HwdID, 3, 255, m_status.index2/1000.0f, m_status.hostname + " Counter 2");
+			SendWaterflowSensor(m_HwdID, 4, 255, (float)m_status.flow2, m_status.hostname + " Flow counter 2");
 		}
+	}
+	else
+	{
+		message = "EcoDevices firmware needs to be at least version ";
+		message = message + boost::to_string(min_major) + "." + boost::to_string(min_minor) + "." + boost::to_string(min_release);
```
gizmocuz: Could you remove the boost::to_string dependency
you can do this with a standard std::stringstream, see code elsewhere in domoticz
note, you have to clear the stringstream before using it again in the same string

gizmocuz (commented):
Sorry, some more small requests
Thanks

hardware/EcoDevices.cpp
DebugBill: Done

hardware/EcoDevices.cpp
DebugBill: Made sure we always get "."

Done

gizmocuz:
Thanks!
Could you undo all formatting changes?

DebugBill:
Oups, Sorry. 
I automatically apply C beautifier when closing files in order to check for missing braces and other simple stuff. 
Do you use some formating tool that I should apply instead?

hardware/EcoDevices.cpp
DebugBill: Done

gizmocuz:
Thanks!

PR #1296, opened 2017-03-03
FIX : false positive GPIO thread seems to have ended unexpectedly

PicoI2:
The heart-beat time was set only after 12 seconds without any interrupt. But if the GPIO see 1 interrupt every 12 seconds or less, for a minute (the heart beat timeout), you see in log "thread seems to have ended unexpectedly".
The heart-beat time is now set every 12 seconds, even if there is interrupts.


gizmocuz:
Thanks!

PR #1297, opened 2017-03-04
XiaomiGateway: Better socket code and switch update

corbinmunce:
Using a more reliable method to detect the local IP address, and then use that IP address to listen for gateway messages.

Also only updating switch status in Domoticz when state has changed, or the update message is not a heartbeat.


gizmocuz:
Thanks!

PR #1298, opened 2017-03-04
BTicino / Legrand OpenWebnet Alarm , CEN , Time

gibbo520:
Added Who 5 Alarm device recognized (if exist), and listed on device.
---Now you can check  if alarm is Disengaged (0) Engaged (1) in Alarm (4)

Added WHO 25 CEN Command (21,22,23) short, extended, long press button and listed on device as switch .
---Now you can trigger script pressing CEN Button (remember to reset the switch to 0  from the script !!)

Added WHO 13 date time command , now the Myhome date time is synched with date time from domoticz (make sure your sistem is setted with right time!!)


Spugna85:
The time change can be a problem if the Biticino gateway is set as slave for time synchronization. Other devices on the bus may not be updated, which is why I do not think it is correct to set the time of bticino gateway.
This should be an option that can be activated in the hardware page if necessary.

gibbo520:
yes great idea ! i'll put on Hardware webpage

Spugna85:
You can comment "setTime()" call until the option will be implemented in the hardware page? for my plant would be a problem.
Then I believe that synchronize every 5 minutes is perhaps a little too frequently.

Alternatively you could automatically create a "push on button" device, so that you can force synchronization when you want and manage the automatic synchronization via user-defined periods events.

Spugna85:
Thanks.

PR #1301, opened 2017-03-05
Use generic directory handling for Lua scripts, Python scripts and Python libraries

oohlaf:


oohlaf:
Hmm, the autobuilder is broken. Seems there is no python3 installed.

gizmocuz:
Don't think it is a python header as i can see it has found 3.6
I restart the build, lets see

oohlaf:
Ah, I have spotted the error:
The script in Travis config references path with python3/3.6.0, while the brew install output shows python3/3.6.0_1
Note the _1

> "cmake -DCMAKE_BUILD_TYPE=Release -DPYTHON_INCLUDE_DIR=/usr/local/Cellar/python3/3.6.0/Frameworks/Python.framework/Versions/3.6/Headers -DPYTHON_LIBRARY=/usr/local/Cellar/python3/3.6.0/Frameworks/Python.framework/Versions/3.6/lib/libpython3.6.dylib",

gizmocuz:
Thanks!!

PR #1307, opened 2017-03-06
Add dzVents scripting files.

scottydj2:
dzVents files added to distribution.
Change made to EventSystem.cpp to add a new globalvariable "script_path" which is required for use on a Windoze system.
By default, the dzVents scripts will not be active unless the _demo is removed from the files.


dannybloe:
Hey Scotty, interesting to see you doing this. Giz and me where planning to do this and then do it properly which means that dzVents doesn't have to rely on doing http requests to get the extra sensor information that is not part of the current LUA globals. We had these ideas to just create a data structure that is passed to dzVents with all the relevant info so dzVents can use that to create it's object API.
I haven't had time yet to look at your adaptation so far but I will.

Can you drop me a line via email?

gizmocuz:
Hey Scotty, great ! Maybe you can team-up with Danny ?

scottydj2:
Hi Gizmocus,

I'll drop Danny an email.
Is it worth saving this work back to your dzVents branch as this is the only clean way currently to get it working on a Windows platform.

scottydj2:
Closing just now while I sort out my branches and continue work with Danny.

gizmocuz:
I think DZEvents will be merged after the next stable release

PR #1308, opened 2017-03-07
Plugins to accept binary data

dnpwwo:
plus added DeviceID paramater to Device.Create()


gizmocuz:
Seems our build system can not find python anymore (x86)

oohlaf:
He needs to rebase his branch against master first, as this build is not using my committed fix for the path.
The CI builder does a build from the requesting tree, not from the master tree. In this case branch Python-Plugins which is 43 commits behind on master.

PR #1309, opened 2017-03-07
First step for code reorganisation and factorisation (Teleinfo.cpp & EcoDevices.cpp)

DebugBill:
Before I go any further are you OK with this?
The goal is to suppress code duplication (or at least functionality duplication as both codes have been developped idependently) between two hardware pieces TeleinfoUSB and Ecodevices. 
Teleinfo is only a protocol, both devices do the electrical interface and decoding, but the data processing itself can be common. 
I tried to stick to the naming convention and called "serial" the USB serial emulator and "Base" the futue common processing class. EcoDevices is connected via Ethernet.


hardware/TeleinfoBase.cpp
```diff
@@ -0,0 +1,47 @@
...
+Author : Blaise Thauvin
+Version : 0.1
+Description : This class is used by various Teleinfo hardware decoders to process and display data
+			  It is used by EcoDevices, TeleinfoSerial and more to come
+
+History :
+0.1 2017-03-03 : Creation
+*/
+
+#include "TeleinfoBase.h"
```
gizmocuz: The first include file in a .cpp file should be
`#include "stdafx.h"`

This will also solve the compile errors

gizmocuz (commented):
Do you expect a TeleInfo over TCP ?
If not, i would not make a base/serial class

DebugBill:
Hi,

EcoDevices is basicaly a TeleinfoTCP device if you only consider its two URLs (for 2 meters) used to pull the entire set of variables formated as XML. Additionaly it provides a subset in JSON format and can also push this subset. Finally it manages two pulse counters. Most people only use the Teleinfo part of it though.

Then it exists several TeleinfoUSB devices which provide the raw data, unformated as a virtual serial port. Theoretically they should all be compatible with the current TeleinfoSerial implementation. I've only used one.

Another device I know (and plan to acquire some day) is a zigbee adatper for meters located outside of the premises (http://www.cartelectronic.fr/teleinfo-compteur-erdf/9-teleinfo-2-compteurs-xbee.html#/xbee-xbee_classique). This one manages up to two meters and according to the documentation would need a new hardware to be derived from TeleinfoSerial as it differs slightly (managing 2 meters being the main difference). 

All three devices can still share the "data to devices" code that I plan to put in TeleinfoBase as this part is absolutely identical. More devices may appear in the future. There are already a few others available, but they are proprietary and can not be connected to Domoticz (no documentation).

I will add the include line to the code so it compiles. I tested on both Linux and Pi boxes and it compiled fine whithout it so I pushed it with confidence.

Blaise

hardware/TeleinfoBase.cpp
DebugBill: done

gizmocuz:
Thanks, now it compiles under windows as well

PR #1310, opened 2017-03-07
OpenWebNet: bug fix and burglar alarm battery, network and maintenance status and fix warnings

Spugna85:
Fix wrong automatic creation of auxiliary device and update status of dry contact, introduced with the development of CEN PLUS commands.

'Unit' must be left free for the interface, so the alarm zone is used as the 'where' and not as 'interface' in UpdateAlarm function.
After this update, the alarm device will have a different ID.
Adding devices to battery status, network status and alarm maintenance.

Fix compilation Warnings.


hardware/OpenWebNetTCP.cpp
```diff
@@ -1097,6 +1097,8 @@ bool COpenWebNetTCP:: WriteToHardware(const char *pdata, const unsigned char len
 bool COpenWebNetTCP::sendCommand(bt_openwebnet& command, vector<bt_openwebnet>& response, int waitForResponse, bool silent)
 {
     csocket *commandSocket;
+	bool ret;
```
gizmocuz, line 1100: This is an uninitialized value ?
Should it be set default to false ?

Spugna85: I think is not necessary, the variable is always set before being used.
In one case from ParseData() function if you expect the response to the command, in the other case it is always set to true.

However if you want I can add to the default false value ....

gizmocuz:
Sorry, had to look 3 times ;) your right ;)
Thanks!

PR #1311, opened 2017-03-08
Fix: Hardwarecontroller.js, Init Mode1 to Mode5

hvbommel:
Xiaomi Gateway could not be configured any more after recent code change, fixed.


gizmocuz:
Thanks

PR #1314, opened 2017-03-09
Add -lresolv to make open-zwave Dev link ok

hvbommel:
Latest open-zwave Dev failed to link into domoticz because of the missing -lresolv switch in CMaketLists.txt. After this update regenerate your Makefile.


gizmocuz:
Thanks!

PR #1315, opened 2017-03-09
Fixed temp/humidity graph ranges and added minor ticks

xztraz:
humfix

should not show more than 100% mow.

Added some minor ticks for hours and days. but thats maybe more to my taste? :)

### DIFF
--- a/www/html5.appcache
+++ b/www/html5.appcache
@@ -1,5 +1,5 @@
 CACHE MANIFEST
-# ref 1368
+# ref 1369
 
 CACHE:
 # CSS


xztraz:
i don't have a year of data som can somebody please test the code on a system with more data to see if year graph looks ok too?

xztraz:
Before screenshot

beforehumfix

xztraz:
on a sensor with very little change it looks like this



After2
humafterfix2

Before2
humbefore2

xztraz:
oops. wrong file uploaded...

PR #1316, opened 2017-03-09
Fixed Temp/Humidity graph scaling ish and minor ticks - bugs found reverted

xztraz:
Should not show more than 100% humidity and ranges should not be zoomed in to infinity. numbers should be whole numbers on axis. added minor ticks to timeaxis

Could not test with a year worth of data so please try it out on a system with more uptime.

**Result**
humfix
humafterfix2

**Before**
beforehumfix
humbefore2


xztraz:
god damn.. wrong file again.. its working but i have a later one.. closing.. :)

xztraz:
maybe i comment my code too much but this is what it looks like on 3 day graph (formatting is nicer in file...)
```
		title: {
			text: $.t('Temperature') + ' ' + Get5MinuteHistoryDaysGraphTitle()
		},
		xAxis: {
			minorGridLineWidth: 0,
			minorTickPosition: 'inside',
			minorTickInterval: 3600000,   //minor tick every hour
			minorTickWidth: 1,
			minorTickLength: 1,
			type: 'datetime'
		},
		yAxis: [{ //temp label
			minRange: 2,				//min range for auto zoom
			tickAmount: 5,			//match humidity ticks
			labels: {
				formatter: function () {
					return this.value + '\u00B0 ' + $.myglobals.tempsign;
				}
			},
			title: {
				text: 'degrees ' + tempstr
			}
		}, { //humidity label
			showEmpty: false,
			allowDecimals: false,	//no need to show scale with decimals
			ceiling: 100,					//max limit for auto zoom
			floor: 0,							//min limit for auto zoom
			minRange: 10,					//min range for auto zoom
			tickAmount: 5,				//match temp ticks
			labels: {
				formatter: function () {
					return this.value + '%';
```

xztraz:
It still sometimes show scale up to 120%, but it's alot more behaved now (scales always line up and such). i've tried all the parameters for the graph and can't lock it at max 100. could be a bug in highcharts.

Tested some more and when setting tickamount it messes with the autoscaling. but if i don't set that the scales could missalign quite badly.

It doesnt seem to jump up to 160% as before only 120%. i think that is better atleast until i can resolve why highcharts doesen't do exactly as i want.

tested in highcharts api too and same problem there if setting something other than 4.

i could change it to 4 here too. seem to work at that setting but looks nicer with 5 lines.

gizmocuz:
That's also what i experienced
Maybe you could make a fiddle when some sample data, and post this on the highcharts forum ?
It looks like there is a problem in highcharts

xztraz:
Highcharts support forum thread about the issue.

https://forum.highcharts.com/post132801.html#p132801

xztraz:
It still looks better than before so pull should be ok.  i have to do the timescale minor ticks on other graphs too if it looks ok for you guys.

xztraz:
Fixed all graphs to show minor ticks and some experimintation with temp colors,i think blue is better for humidity at least, green for temp?

also changed wind graph background to more of a gradient

image

image

xztraz:
highcharts guys indirect says its a bug and proposes to use a set scale. maybe a no dynamic scale from 0-100% is better. easier to just give a quick glance and see if humidity is high or low.  What do you think? other than highcharts should fix their code.. :)

xztraz:
This might be a reason to upgrade highcharts.
http://www.highcharts.com/docs/chart-design-and-style/style-by-css
Then the theme css could set all graph styling parameters like color tics and whatnot :)

gizmocuz:
Graphs should be dynami
Don't like the blue color for humidity that much, as also the background for the wind graph
Or the ticks between the hours actually

xztraz:
oh well. maybe its better to upgrade highcharts (or go to graphana) and make styling of graphs in themes/gui instead since everyone have different taste and backgrounds and such could differ with theme. 

Closing for now until stuff resolved

xztraz:
another bug in highcharts makes minor ticks dont scale properly when zooming in brower.. removing that too.

xztraz:
better blue? removed the ticks..
image

xztraz:
ok back to square one ish. only applied the scale fix 

image

gizmocuz:
Yeah... like the title of this issue... how convenient
But this does not solve the problem, you have 120% in your day graph
Until this does, it does not make sense to merge now

xztraz:
before i have 140% so its better. and other graphs showed 160%. its just in rare cases it jumps up to 120% now.

gizmocuz:
Yes, we know... but it does not solve the problem

xztraz:
not completely but makes it a lot less. isn't that better? : / and also min scale is alot better than before. i can't solve the bug in highcharts i can only try to workaround and this is the best i can give you. well other pull for just this change and with better screenshots. continue there.

gizmocuz:
This patch is more than just alone a possible better way for 100%, there are many more changes in this file
Lets wait until a day highcharts can solve this

xztraz:
reverted all other stuff in other pull request.

PR #1317, opened 2017-03-09
Compare previous state before updating

JordyThien:
If a sensor is new, always update


gizmocuz:
Great, thanks !!
... Maybe for the temp/lux to send them 
- when there is a change
- or at least every xx (5) minutes ?

JordyThien:
@gizmocuz I'll put it on the list :)

gizmocuz:
Thanks!

PR #1321, opened 2017-03-10
Create custom.css - I've added a new theme.

flatsiedatsie:
I've added a new theme. Hurray!

Making the theme was quite an adventure. The HTML/CSS for Domoticz is extremely messy and doesn't follow standards..

It would be great to clean and update the HTML front-end sometime. I think it really matters to the succes of Domoticz, which currently has some really great features that others don't have. But they are catching up, and they do look pretty and are more usable.

Usability matters. A lot!

### DIFF
--- a/www/styles/progress/custom.css
+++ b/www/styles/progress/custom.css
@@ -1,0 +1,373 @@
+body table#itemtablesmall tr td:first-child + td {margin:0!important; padding:5px!important; line-height:100%!important}
+
+@font-face {
+	font-family: "OpenSans";
+	font-style: normal;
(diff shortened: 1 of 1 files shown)


flatsiedatsie:
Hmm, I just realised I haven't tested all types of things that exist in Domoticz. I don't have anything with a slider.

gizmocuz:
Some screenshots would also be handy...
You can add a Dummy Hardware, and add virtual devices

flatsiedatsie:
Yes I've done just that. I'll be re-adding the theme(s) soon.

PR #1322, opened 2017-03-10
Send RGBW from web to app

szczukot:


gizmocuz:
This is still on the todo list, and requires a lot more than this.
The whole hue stuff should be dropped

szczukot:
Yes, i know.
This is a small first step - communication between web and app.
In the comming days, i will try use this value with blebox RGB devices.

gizmocuz:
for now use the hue value as all other systems does, until this all gets rebuilded

szczukot:
HUE is fake :)
This is only one number for poor devices. 
Blebox and others use 4 channels (RGBW) separately controlled.
Coolpick from domoticz is ok, but we must add support for RGB instead HUE

gizmocuz:
i know... that is what i said.... we should drop the hue stuff
Only AppLamp is using it
But this has to be done in general, not with a quick hack like this

PR #1323, opened 2017-03-10
Raspberry GPIO capture IO-pin status at startup

hvbommel:
When Domoticz currently starts up, on Raspberry Pi, the actual input state is not reflected correctly. Only after an input state transition Domoticz on Raspberry sees the correct input status. This is a problem when for example door or window sensors are monitored. The state would in that case only be correct after all doors and windows have been opened/closed or vise versa. To fix this problem this modification captures the actual input state on startup and updates the corresponding switches. Also after 30 seconds input state is captured again. The latter is done to support Raspberry Pi being used as a slave device providing its inputs to the master. The master usually quickly connects (assumed to do so within 30 seconds). Then Raspberry input states are then captured and updated one more time. As a result of this input states are correctly reflected on the Domoticz Master also. Note: I have build the code on Windows, Ubuntu and Raspberry Pi model3. I have tested it on Raspberry only as this feature is only applicable on the system.


hvbommel:
Will submit version that also builds on Windows later

gizmocuz:
Also when you first email me about this, it is better to also wait for a response before making changes

PR #1324, opened 2017-03-10
Raspberry-Pi GPIO capture input state on startup

hvbommel:
When Domoticz currently starts up, on Raspberry Pi, the actual input state is not reflected correctly. Only after an input state transition Domoticz on Raspberry sees the correct input status. This is a problem when for example door or window sensors are monitored. The state would in that case only be correct after all doors and windows have been opened/closed or vise versa. To fix this problem this modification captures the actual input state on startup and updates the corresponding switches. Also after 30 seconds input state is captured again. The latter is done to support Raspberry Pi being used as a slave device providing its inputs to the master. The master usually quickly connects (assumed to do so within 30 seconds). Then Raspberry input states are then captured and updated one more time. As a result of this input states are correctly reflected on the Domoticz Master also. Note: I have build the code on Windows, Ubuntu and Raspberry Pi model3. I have tested it on Raspberry only as this feature is only applicable on the system.

### DIFF
--- a/hardware/Gpio.cpp
+++ b/hardware/Gpio.cpp
@@ -222,7 +222,7 @@ bool CGpio::StartHardware()
 #endif
 	m_stoprequested=false;
 
-	//Start worker thread that will be responsible for interrupt handling
+	//  Start worker thread that will be responsible for interrupt handling
 	m_thread = boost::shared_ptr<boost::thread>(new boost::thread(boost::bind(&CGpio::Do_Work, this)));
 
 	m_bIsStarted=true;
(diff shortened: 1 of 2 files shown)


hvbommel:
Updated code to only update the Domoticz database when a new GPIO state mismatches the state recorded in the database. When the related device is not present in the database then initial state for a given GPIO pin is not set either.

hardware/Gpio.cpp
```diff
@@ -63,6 +64,8 @@ Connection information:
 bool m_bIsInitGPIOPins=false;
 bool interruptHigh[MAX_GPIO+1]={ false };
 
+extern 	CSQLHelper m_sql;
```
gizmocuz: This is not needed, see header file of SQLHelper.h

gizmocuz (commented):
Thanks!, One small request...

hvbommel:
Database update is forced after 30 seconds. After this time the master should be connected. By updating the database the master gets updated also. This will allow the master to reflect the correct switch states after slave restart. The unneeded extern and some noisy log messages have been taken out also.

gizmocuz:
Thanks !!
Maybe when you got some time you might want to look at the new filesystem 'sysfs' that can handle GPIO on multiple platforms (if not all...)
I think that this is a more generic approach for GPIO

PR #1330, opened 2017-03-11
Notifications add Last Update for sensors

jvandenbroek:
I wanted to get notifications when no data is received for some of my sensors. After enabling SensorTimeoutNotification in the DB, I got flooded with high priority messages from many old sensors, so that wasn't a very good solution ;) 

I changed the code, so that the behavior remains the same when using the SensorTimeoutNotification field, but when disabled, the notifications table is checked for individual settings. To keep things simple, I kept the logic of greater or below, not sure if that makes sense.

Not sure why I see my commits for the P1 Smart Meter in this pull request, that was already merged last week? No files seems to be affected by that. 

Please review and comment, thanks!


main/WebServer.cpp
```diff
@@ -5520,6 +5520,35 @@ namespace http {
 					root["result"][ii]["ptag"] = Notification_Type_Desc(NTYPE_SWITCH_OFF, 1);
 					ii++;
 				}
+				if
```
gizmocuz: Maybe we can use the function 'IsLightOrSwitch' ?

gizmocuz (commented):
Could be you had to start from a clean fork again ?
I got one request (IsLightOrSwitch)

main/WebServer.cpp
jvandenbroek: That's a nice function, changed it :)

jvandenbroek:
I did some things which might have upset Git, but haven't started with a new fork (at some point that might have possible been the better strategy ;)). Ah well, as long as it doesn't affect any older files..

gizmocuz:
Thanks!

PR #1331, opened 2017-03-11
Better behaved humidity scale

xztraz:
Its not perfect(due to bug in highcharts) but better. Doesn't zoom out to 160%, only to 120% in those cases. 
dynamics works better with smaller ranges in month graph and so. 

**Before:**
image
image

**After:**
image
image


xztraz:
This is how it looks after you have enabled/disabled temp and or hum graph 

Before (scale of humidity is all zoomed out)
image

After (with the fix)
image

xztraz:
actually only change needed is this added con hum label for 5min,month,year graph humidity section for the humidity scale to behave better.

```
			allowDecimals: false,			//no need to show humidity scale with decimals
			ceiling: 100,					//max limit for auto zoom, bug in highcharts makes this sometimes not considered.
			floor: 0,						//min limit for auto zoom
			minRange: 10,					//min range for auto zoom (
```

xztraz:
removed some more of my code that wasn't needed. looks the same on humidity (better than before) temp is left unchanged.

**Before**
image

**After**
image

xztraz:
i'm quite happy with this.

xztraz:
conflict.. bah.

xztraz:
how do i rebase? or do i need to fork again from master?

gizmocuz:
Maybe because of the wind direction patch that was also in there ?
I adapted your changes in the last code commit, but it does not work here for the year graph
It is still 120%. If i disable the temperature, the humidity is limited to 100% correctly
If i then enable the temperature graph, it stays as 100%
Only showing Humidity, and doing some zooming always goes to 120%

xztraz:
Hi!
Should only have contained the scaling patch. i upgraded my domoticz installation and took that file as a base and entered the scaling things only. I think atleast :) but good that you look. github confuses me and code review is good.

remember what did it show before as min and max?  i know its not perfect but should show equal or better range than before.   Highcharts don't seem so willing to fix their code :/

i guees it's ok to close this pull then if you have the code snipplet

xztraz:
This is how "year" looks for me (before it had a range of 0-120%
image

i would guess min range is better on your year graph atleast?

PR #1332, opened 2017-03-11
logger trace level evolution

labelette91:
logger evolution:
a log level as been added : LOG_TRACE
it allows to print functionnal trace during debug without recompile with option _DEBUG_.
( trace is actually activated by _DEBUG_ compile option and using LOG_STATUS level )

In order to reduce nomal CPU usage , 
it is recommended to test TRACE level before sending LOG_TRACE in order to avoid unusefull Log function call in standart log level.

example :
if (_log.isTraceEnable()) 
	_log.Log(LOG_TRACE,"MAIN SendCommand: %s", szMessage);

in order to help the trace filtering,
The trace text shall be prefixed by 4 or 5 letters  that represents the module name :
MAIN : mainworker
WEBH : web helper cWebem
RFXN : RFXName
SQLH : SQLHelper
WEBS : WEBServer
....
the second is the function name : ie SendCommand .

In Web Hmi , Some option could be configured in the Setting tab Other :
The Log Filter    value
The Log File Name
The Log Level 

This parameter could be set on command line but it is easyer in web HMI in real time.
The configuration is stored in database in preferences Table.

The trace level has 8 level
ERROR   LOG_ERROR
STATUS 	LOG_STATUS
NORM   	LOG_NORM
TRACE  	LOG_TRACE
ERROR  verbose      LOG_ERROR  and 	EVBL_ALL = 1, 
STATUS verbose 	LOG_STATUS and 	EVBL_ALL = 1,    	
NORM   verbose 	LOG_NORM   and 	EVBL_ALL = 1, 
TRACE  verbose 	LOG_TRACE  and 	EVBL_ALL = 1, 

The log could be fieltered in the Setting TAB : Log: Filter.

A list of key words shall be entered separeted by semicolon ';': 
all the log containing this word will be filtered/ignored and not displayed..
If the key word is prefixed by plus '+' :  
all the log containing this word will be displayed .
example :
"SQLH;WEBS"              : all log comming from SQLHelper and WEBServer will be filtered.
"SQLH;WEBS;+insert"  : all log comming from SQLHelper and WEBServer will be filtered 
                     but the sql request containing insert will be displayed
" ;+tata"            : all log containig a blanc " " will be filtered 
                          --> IE all log is filtered
			but only the log containing tata will be displayed
" ;+insert"          : only the log containing "insert" will be displayed
												
todo: convert all _DEBUG_ into LOG_TRACE( "module name function name ... "


gizmocuz:
Thanks for your great and hard work! But...
- All these trace/debug logs are stored inside the executable, even if not enabled
- I do not want to give more logging options to end-users, they will enable it and confuse them, and it's not necessary
- i expect from developers that they are able to compile in debug mode, debug their code, and can add debug logs if needed
- When using windows (Visual Studio 2015), they are able to set breakpoints, and can easily debug their work
- If you are going to change c++ code, then you can compile/run/debug/test in debug mode, just use two folders (debug and release)

Best practice for CMake is to build in a folder (debug/release)

i really don't mind the new method.. but these should still be inside a 

```
#ifdef _DEBUG
.... do log
#endif 
```

And it should for sure not be able to enable this via the web interface

But... setting up (using) a good build environment will probably mean this might not be needed after all, as you test every line of new code by stepping though it
Windows - Visual Studio
Linux - CodeLight or other ide...

Hope you agree a bit...

szczukot:
Gizmo, i don't agree with You.
Mny times I wrote code "not for me". I dont'have devices etc. 
And many time i fixed errors for situation which not exists for me, but others users report.
And this improvements is good idea.

labelette91:
hello, thanks for the  answer. 
Domoticz is a very nice software.  I am developping a virtual themostat wich use only switch and temperature sensor and integrating Imperihome IIS api. 
The log as been developped in order to understand the internal behaviour of code and also to debug system installed not "on my desktop"!!!!

gizmocuz 
i agree with to the fact that "more logging options to end-users, they will enable it and confuse".
But i observe many times in my professionnal experince that some time the users views bugs that was not discovered during integration. 
It is why we use this trace level in order to investigate on problems discovered by client: it activates this trace and send us the log output for investigation .
I am agree that this trace level shall not be interpreted by him.
Perhaps could we activate this trace / filter / HMI  only by line argument ( no by default).
The interest is that you have not to recompile the software with debug option in order to investigate on a probleme.
This useful to debug on raspberry pi without any debugger.
For example i have use it to connect an ESPEasy to raspberry and view directly the Json request to see that it works very good!!!!!!

Tells me how i can make evolution in order to include this feature !!! If needed!!!
I will be proud to collaborate.
cdt

labelette91:
a DEBUG line argument could activate TRACE log level and HMI ?

gizmocuz:
Okey, how about a compromise ;) ... only possible to activate it via the command line ?
users stop domoticz, start domoticz by hand with the trace ?

szczukot:
Or maybe add options to settings (default disabled checkbox) ?

labelette91:
So  a proposal is to add argument DEBUG in command line 
The use shall stop and run to enable 

1) this will allow the selection of TRACE LEVEL : the user shall sepecify the trace level using the command line LOG or by HMI wich will be hiden by default.

2) this allow the display of HMI in Setting tab to modify 
    - the filter content 
    - the log level.
    - log file name

szczukot : what the checkbox shall allow ? the debug option ?

szczukot:
Yes, checkbox for trace.
Or combobox for level.

gizmocuz:
no options to enable this with a checkbox in domoticz

labelette91:
understand for checkbox.

So  OK also  to enable the HMI option when DEBUG command line argument is set ?

gizmocuz:
Sure.... enabling log/trace via command line is no problem...
Just that 99,9% of the users can not enable it

labelette91:
[evolutionLog.txt](https://github.com/domoticz/domoticz/files/852566/evolutionLog.txt)

The trace level is now only allowed when command line arguement option -debug is present.

I have add the trace    for SQL query  and  for SQL query  results 
SQLQ : SQLHelper database sqlite3 query        :Q for Query
SQLR : SQLHelper database sqlite3 query results:R for Results

to deactivate the sql trace , put SQL in LogFilter 
see in evolutionLog.txt

gizmocuz:
Thanks!

PR #1334, opened 2017-03-11
Relaynet: Only write db on relay or input state changes

hvbommel:
Before writing the database now the current relay or input state, that is recorded in the database, is checked. Only on a state change the database is written. Note: This change is stored in the relaynet branch to keep it separate from the GPIO pull request.


gizmocuz:
Sorry, this patch also has GPIO files... i think these are also already merged

hvbommel:
Isznogood, I will close this one and create a new pull request for the RelayNet one.

PR #1335, opened 2017-03-11
OpenWebNet: GEN, AREA and GROUP devices

Spugna85:
These new devices are added only when a general, area or group command is received.
For now, there is not the manual addition of these devices.


hardware/openwebnet/bt_openwebnet.cpp
```diff
@@ -517,15 +517,18 @@ bt_openwebnet::bt_openwebnet(const std::string& message)
   CreateMsgOpen(message);
 }
 
-bt_openwebnet::bt_openwebnet(int who, int what, int where)
+bt_openwebnet::bt_openwebnet(int who, int what, int where, int group)
```
gizmocuz: could you make these 'const' ?

Spugna85: Done.

PR #1336, opened 2017-03-12
HTTP Protocol uplift

dnpwwo:


hardware/plugins/PluginProtocols.cpp
```diff
@@ -17,6 +17,7 @@
 #include <queue>
 #include <boost/thread/mutex.hpp>
 #include <boost/thread/lock_guard.hpp>
+#include <boost/algorithm/string.hpp>
```
gizmocuz: Please do not introduce another boost usage, if possible add a helper function so others can use it in the feature to (helper.h/cpp)

```
#include <algorithm>
#include <string>

std::string str = "Hello World";
std::transform(str.begin(), str.end(),str.begin(), ::toupper);

```

PR #1337, opened 2017-03-12
[Logitech] Fix bug in set timeout

szczukot:
Before, Logitech set http timeout to other devices too.


gizmocuz:
Instead of this, in the Logitech class, issue a call to
HTTPClient::SetConnectionTimeout(const long timeout)
And your set

szczukot:
SetConnectionTimeout and SetTimeout set timeouts global - for all other modules, which uses httpclient

gizmocuz:
Yes... but look in which classes it is used

szczukot:
I fixed it, because i want use it in BleBox, but default timeout is too big for me.
And i found this bug in Logitech
This commit only fixing bug in logitech - don't change any logic for other modules.
And when in Logitech, someone set ping less that 1000 ms, ale other modules have infinity timeuot.

gizmocuz:
Then before the call to the HTTP class, set the timeout, in both classes and your find
There is really no need to patch the HTTP class

szczukot:
Ok. i set timeout in my module (blebox).
But now 99% of modules don't set this value, and will may work inproperly.

gizmocuz:
your mixing up things here... logitech, blebox, httpclient ...

Logitech is using csocket ?

szczukot:
i closed this pull request a 2 days ago
As You want

gizmocuz:
Yes, but your last comment wat '99% of modules don;t set this value....'
But if you got it working, great!

szczukot:
I think, that this is bug, or all other modules must set timeout. Now about 20 modules use httpclinet, and none of them sets this value. Only Logitech.

gizmocuz:
None of them needs this... no problem with the timeout
I made a recent patch in the logitech harmony class, there was a error
maybe its not needed anymore

PR #1338, opened 2017-03-12
Relaynet only write database on relay or input state changes

hvbommel:
The database is now checked first. Only write to it when the recorded relay or input state does not match the actual, last detected, state.


gizmocuz:
Thanks!

PR #1341, opened 2017-03-13
Renew session expiration at connexion if you checked "Remember me".

PicoI2:
Problem : when you checked "Remeber me" at login, the session was expiring after a month, even is you used it every day.
Fix : session duration is renew after 15 days (if the session is in use).


gizmocuz:
Thanks!!

PR #1342, opened 2017-03-13
Pushsafer add ability to send image

appzer:
add ability to send an image in a push notification though pushsafer.com service

### DIFF
--- a/notifications/NotificationPushsafer.cpp
+++ b/notifications/NotificationPushsafer.cpp
@@ -3,17 +3,8 @@
 #include "../httpclient/HTTPClient.h"
 #include "../main/Logger.h"
 #include "../httpclient/UrlEncode.h"
-
-#include <stdio.h>
+#include "../httpclient/HTTPClient.h"
 #include "../webserver/Base64.h"
-#include <curl/curl.h>
-
-static size_t WriteCallback(void *contents, size_t size, size_t nmemb, void *userp)
-{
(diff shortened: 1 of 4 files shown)


notifications/NotificationPushsafer.cpp
```diff
@@ -4,10 +4,22 @@
 #include "../main/Logger.h"
 #include "../httpclient/UrlEncode.h"
 
+#include <stdio.h>
+#include "../webserver/Base64.h"
+#include <curl/curl.h>
```
gizmocuz: Please use our HTTPClient class for downloading files instead of creating a new curl method

www/views/setup.html
```diff
@@ -392,7 +392,7 @@ <h1 data-i18n="Notifications">Notifications</h1>
 									<br>
 								</div>
 								<div class="span6">
-									<h2>Pushsafer (iOS/Android/Windows10/Chrome/Firefox):</h2>
+									<h2>Pushsafer (iOS/Android/Windows10) + Image:</h2>
```
gizmocuz: With this change we need to change all language translations as well
I dont think it is really needed correct?

appzer:
@gizmocuz 
Have made your proposed changes
Is there a wiki page for notification system.
I would like to explain the [Pushsafer.com](https://www.pushsafer.com) plugin to the Domoticz users.


Regards
Kevin

gizmocuz:
Thanks for the changes!
You can create a wiki page

PR #1343, opened 2017-03-13
Notification fix DST for first result

jvandenbroek:
Fix for setting an incorrect value for DST on an empty tm struct (only affects the first result).


gizmocuz:
Thanks!

PR #1344, opened 2017-03-14
FreeBSD doesn't have libresolv like Linux. Fixing the code

xbeaudouin:
On FreeBSD there is no libresolv. So move the -lresolv only when Linux (or for some OS that needs it).


gizmocuz:
Does this mean you will never be able to compile the feature OpenZWave library ?
Please check if you can compile the development branch of OZW
If possible, please make the patch also there... they made a new class for DNS lookup, no idea why they wont use the default WINSOCK/SOCKET gethostbyname implementation ?

xbeaudouin:
All of this is handled by libc on Freebsd.

xbeaudouin:
Sorry about close/reopen... :D

xbeaudouin:
The developpement branch of OZW works as well too on my platform (just tested).

gizmocuz:
someone added this patch recently... maybe it was not compiling under ubuntu.... have to check

xbeaudouin:
Without this PR, we cannot compile Domoticz on FreeBSD :/

But all check show we can compile on Linux flavor and Darwin, so I think it is ok.

Is there any possibility to add a FreeBSD check somewhere?

hvbommel:
Question: 
Will the latest Domitcz master branch link ok onFreeBSD after you build and included the latest development (Dev branch) of OpenZWave.lib? 

If not that I think that means new OpenZwave development no longer supports FreeBSD unless they fix it. We can put in a request at OpenZWave after all is clarified.

Note: In my case, on Ubuntu, I could build the OpenZWave static lib ok. I also could compile the latest Domoticz error free. When Domotizc was linked I got the linker errors because of the missing -lresolv switch. After adding the switch it was linking ok and testing proved open-zwave was working fine.

hvbommel:
By the way, it looks like the pull request does work around the problem, doesn't it ? I am saying it because my recent experience tells me getting changes into OpenZWave is a slow process. So a workaround seems to be the better option to me. (correct me if I am wrong)

hvbommel:
FYI: This is the linker error I got on Ubuntu when linking Domoticz:

/git/open-zwave-read-only/libopenzwave.a(DNSImpl.o): undefined reference to symbol 'ns_initparse@@GLIBC_2.9'
//lib/x86_64-linux-gnu/libresolv.so.2: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
CMakeFiles/domoticz.dir/build.make:5238: recipe for target 'domoticz' failed
make[2]: *** [domoticz] Error 1
CMakeFiles/Makefile2:103: recipe for target 'CMakeFiles/domoticz.dir/all' failed
make[1]: *** [CMakeFiles/domoticz.dir/all] Error 2
Makefile:149: recipe for target 'all' failed
make: *** [all] Error 2

xbeaudouin:
@hvbommel the error you have on Ubuntu, is with my patch or usualy ?

hvbommel:
it failed without any patch related to this problem. It was fixed after the little -lresolv change in CMakeLists.txt and it will also work with your update, need to test that

hvbommel:
I checked my CMakelists.txt end see that the "IF(CMAKE_SYSTEM_NAME STREQUAL "FreeBSD") ... " is already in. So is there still a problem ? Ie why is this PR still open is my question ?

xbeaudouin:
My reply is clearly yes : 
```
2 warnings generated.
[100%] Linking CXX executable domoticz
/usr/bin/ld: cannot find -lresolv
c++: error: linker command failed with exit code 1 (use -v to see invocation)
gmake[2]: *** [CMakeFiles/domoticz.dir/build.make:5232: domoticz] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:105: CMakeFiles/domoticz.dir/all] Error 2
gmake: *** [Makefile:150: all] Error 2
#define APPVERSION 6959
#define APPHASH "16cba568-modified"
#define APPDATE 1489576896
[kiwi@portsbuild ~]$ uname -a
FreeBSD portsbuild 11.0-RELEASE-p2 FreeBSD 11.0-RELEASE-p2 #0: Mon Oct 24 06:55:27 UTC 2016     root@amd64-builder.daemonology.net:/usr/obj/usr/src/sys/GENERIC  amd64

```

If you look a CMakeLists.txt on line 602:

```
IF(OpenZWave)
  message(STATUS ${OpenZWave})
  target_link_libraries(domoticz ${OpenZWave} -lresolv)
  include_directories(${CMAKE_SOURCE_DIR}/hardware/openzwave)
  add_definitions(-DWITH_OPENZWAVE)
  # open-zwave needs libudev
  IF(CMAKE_SYSTEM_NAME STREQUAL "Darwin")
    set(IOKIT_LIBRARY "-framework IOKit -framework CoreFoundation" CACHE FILEPATH "IOKit framework" FORCE)
    target_link_libraries(domoticz ${IOKIT_LIBRARY})
  else()
    IF(CMAKE_SYSTEM_NAME STREQUAL "FreeBSD")
```

You clearly assume that target_link_libraries(domoticz ${OpenZWave} -lresolv) handle libresolv and then *after* on line 610 you check for FreeBSD.

So this PR fixes FreeBSD support since it doesn't need SysV libresolv.

Thanks

hvbommel:
I put your change proposals into my CMakeLists.txt on Ubuntu and did a build. It works fine. So I think it is the way to go.

xbeaudouin:
@hvbommel Thanks :)

gizmocuz:
Thanks all for testing !

PR #1345, opened 2017-03-14
More serial ports on Raspberry 3

DebugBill:
Modifiy list of serial inputs on Raspberry to take into account new features on Raspberry 3 and Raspbian Jessie
More info at http://spellfoundry.com/2016/05/29/configuring-gpio-serial-port-raspbian-jessie-including-pi-3/

Only add character devices and symbolic links (directories entry "serial/" would be added along with
/dev/serial1 and /dev/serial2 before this change)

Sorry, some indentation cleaning is included again, but there are really less than 10 lines modified/added


gizmocuz:
Maybe you could undo the formatting changes ? then we can see what has really changed ;)

DebugBill:
Done, but as I have added a large IF clause, the identation is wrong and might be misleading for other contributors. Maybe once you've reveiwed the changes I can apply it?

gizmocuz:
I have several raspberry pi 3's
When i plugin a USB device, i do get /dev/ttyUSBxxx
i for sure do not get /dev/ttySxx
And under my ubuntu system i have already 200 ttySxx devices, which is NOT a usb device
Also i do not have a /dev/serial

How can it be that you have these ? What hardware is this ?

DebugBill:
That's because you only consider "fake" serial devices emulated through USB. Mine here is a genuine serial device connected directly to the GPIO pins on a Raspberry3. I've bought this in order to test the TeleinfoSerial code. Most Teleinfo devices are USB.... but at least one, mine is not. And is is cheap so I guess it could be a success (more info at https://hallard.me/pitinfov12/)

I guess other "real" serial devices could be connected to the Pi, and in that case ttyAMA0 will not be enough. On mine this is the bluethooth device because it is activated, and serial GPIO is on ttyS0. Should I turn off bluetooth, AMA0 would revert back to the GPIO

serial0 and 1 are symbolic links created automatically since the new Rasbian Jessie. They are automatically created AND modified to point to the GPIO whatever hardware and setup you use (Pi2 or Pi3, bluestooth active or not) so that the user can take the SD card from a Pi2, plug it to a Pi2, and if its software is configured to use serialX rather than ttyXXXX it will still work. In my case I use serial0 and can change config and board.... Domoticz still finds the right device.

You are right about having lots of ttySxxx on your Ubuntu system, but this is why I took care of placing this patch inside the "if (bHaveTtyAMAfree)" clause in order to limit this detection to Raspberries where only S0 seems to exist under normal circumstances. Maybe we miss a ___Pi_____ precompiler option just like ___APPLE____ or ___FreeBSD___ exists.

gizmocuz:
Thanks, missed the part about inside the AMA free... merged your code, thanks!

I think you are a special case... and 99,999% will not use this S0 port
You could also have followed our wiki about 'persistent devices' or even symlink the serial port and you are also set correct?

DebugBill:
Yes, I could tweak the system to access the port, but I would probably have broken the bluetooth... It feels easier that way for everybody, even if I like to think I am a "special case" :-)

PR #1346, opened 2017-03-14
Percent meter bug

DebugBill:
Fix bug in SendPercentMeter where NodeID and ChildID where not used consistently with other devices

Side-effect: it will break existing devices when applying the update.


gizmocuz:
Can't accept this, as this will cause a lot of sensors to be recreated
you have 255 id's, that must be enough ?
Also see
MainWorker::decode_General

how the ID is generated

PR #1347, opened 2017-03-15
Add dnsapi.lib, to build Domoticz-Win with OpenZWave

hvbommel:
Add dnsapi.lib This is needed to build Domoticz-Windows with the latest OpenZWave (latest master or Dev branch). Note: This is similar to the -lresolv switch that we already have added in CMakeLists.txt for Linux.

### DIFF
--- a/msbuild/domoticz.vcxproj
+++ b/msbuild/domoticz.vcxproj
@@ -64,7 +64,7 @@
       <SubSystem>Windows</SubSystem>
       <GenerateDebugInformation>true</GenerateDebugInformation>
       <AdditionalLibraryDirectories>./Windows Libraries/Boost/boost_1_59_0/Debug;../libusb;%(AdditionalLibraryDirectories)</AdditionalLibraryDirectories>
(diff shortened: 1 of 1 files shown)


gizmocuz:
We have a problem with freebsd users see:
https://github.com/domoticz/domoticz/pull/1344

Maybe you could ask openzwave why they do not just use gethostbyname ?

hvbommel:
With a backlog of 95 items, and growing, I am not sure if OpenZWave is interested to work on  issues that are not really broken. We may trace back and find the developer who added cpp/src/platform/unix/DNSimpl.cpp. May be he/she would be so kind to revisit the design.

gizmocuz:
Addeded my 2 cents... hope it can get rebuilded, or that there is a good reason for using this

https://github.com/OpenZWave/open-zwave/commit/bab30a20b1c31a3f15b60dcf8eb9aafbfea75154

hvbommel:
I think for now we should merge this PR.  We can take it out, some time in the future, when it is no longer needed. Then at least we can include the latest open-zwave into domoticz/windows also for the time being.

gizmocuz:
Thanks!

PR #1350, opened 2017-03-15
Percent meter bug, now with database updater

DebugBill:
If feel disturbed having a similar API for many sensors but a different behaviour (also it broke my code until I understood why :-))
I agree we have enough IDs available with the current SendPercent version, it just feels more consistent that way.

This time I've included the database update tool in order to avoid breaking existing devices. Hope you like it better.


gizmocuz:
Thanks! I am going to look into this one carefully with some real installations

DebugBill:
Forget about it. I made some more checks within the code and it will break quite a few sensors. Most of them don't use the standard helper, and those who do pass various information that my database patch cannot guess.
So long for interface consistency :-)

gizmocuz:
True, also remember that these global function were created years after those hardware plugins
Thus trying to get to there... it just needs some work

You can see from my comments that i encourage people to use those global functions

PR #1353, opened 2017-03-15
Are default dividers set to the wrong value?

DebugBill:
Is there any reason why the dividers for Gaz and Water are intialised to 100 rather than 1000? It does not work for Ecodevices which sends m3 in order to show liters and m3 on screen.
One user was confused and sent me a bug report. I explained he just had to change the setting in preferences. 
Is there another device that needs 100?

Please ignore files other than SQLHelper in this request. I can't help having them modified because GIT replaces the CRLF inside with LF.


gizmocuz:
This was in the beginning the divider from the homewizard and p1 smart meter
You can configure it to 1000 if you need to/like

DebugBill:
I did change the setting on my system when I faced the issue and that's why I created the Wiki page.... so that users find out why the results are 10 times too big. I hope they go and read it.

PR #1356, opened 2017-03-16
Alert now comes with optionnal explanatory messages

DebugBill:
SendAlertSensor now needs one more text parameter (only two devices are using this call, so it was easy to update)
If a text is provided, it appears on the GUI next to the numeric level. This allows for the 4 alarm levels to be reported with a explanation/description of the issue
If no text is provided, the behaviour is as before

I've tested with notifications, it did not break anything.
Did not have to modify the GUI code as it already relied on the string database field.

This makes the Text sensor even less usefull as one does not need to use both Alert and Text to report a detailed alarm.


gizmocuz:
Looks we have a conflict because of a recent pull request merge on ecodevices (i think it was yours...)

gizmocuz:
Do you have a screenshot of the new sensors usage ?
No need to update all current text sensors ?

DebugBill:
Here is a screen shot (hopes it goes through GitHub)

It needed an update of all calls to SendAlertSensor…. But there were only 3 throughout the code.
For direct calls to sDecodeRX there is no impact (thanks to my latest change a few minutes ago. It was not true one hour ago)

[cid:image001.png@01D29EFC.4F843FA0]

hardware/MultiFun.cpp
```diff
@@ -414,7 +414,7 @@ void MultiFun::GetRegisters(bool firstTime)
 					}
 					if (((m_LastAlarms != 0) != (value != 0)) || firstTime)
 					{
-						SendAlertSensor(0, 255, value ? 4 : 1, "Alarm");
+						SendAlertSensor(0, 255, value ? 4 : 1, value ? "Alarm" : "OK", "Alarm");
```
szczukot: wrong. Send always "alarm" as text

hardware/MultiFun.cpp
```diff
@@ -436,7 +436,7 @@ void MultiFun::GetRegisters(bool firstTime)
 					}
 					if (((m_LastWarnings != 0) != (value != 0)) || firstTime)
 					{
-						SendAlertSensor(1, 255, value ? 3 : 1, "Warning");
+						SendAlertSensor(1, 255, value ? 3 : 1, value ? "Warning" : "OK", "Alarm");
```
szczukot: Wrong
Send always "Warning" as text

hardware/SatelIntegra.cpp
```diff
@@ -771,7 +771,7 @@ void SatelIntegra::ReportZonesViolation(const int Idx, const bool violation)
 
 	m_zonesLastState[Idx - 1] = violation;
 
-	SendAlertSensor(Idx, 255, violation ? 3 : 1, NULL);
+	SendAlertSensor(Idx, 255, violation ? 3 : 1, "", "Alert");
```
szczukot: I don't want set name for this sensor.
Must be NULL instead "Alert"

szczukot (changes requested):

gizmocuz:
Maybe because you are reply by email that the image does not work
Also you copy part of your email into this, which i have to remove

Could you copy/paste it directly in github

Ps. Our automatic build systems have issues, see below (on git)

DebugBill:
Fixed the merge issue and reverted Multifun and SatelIntegra to original behaviour.

Here is a screenshot

Now waiting for successful build

alerts

hardware/MultiFun.cpp
DebugBill: reverted to original behaviour

hardware/MultiFun.cpp
DebugBill: reverted to original behaviour

hardware/SatelIntegra.cpp
DebugBill: reverted to original behaviour

szczukot:
Idea with additional text is good.
Maybe i will try use it in Satel and MultiFun

DebugBill:
Please don't merge yet, I'm not happy with the current behaviour but I will not have time to look into it before the week-end

gizmocuz:
Okey, send me a PM on the forum

DebugBill:
This version fixes most (I think) of the negative feedback on the forum.
Still some issues with passing text on alerts, but it should not be a major problem.

I did not find any explanation to the crashes reported by some users, I will continue to investigate now that I have many testers.

gizmocuz:
Please forget the 'negativity' of the feedback... i (most times) consider them as anomalies ;)

PR #1357, opened 2017-03-17
Update XiaomiGateway.cpp

corbinmunce:
Changing switces back from sSwitchGeneralSwitch to sSwitchCustomSwitch as this was causing double updates


gizmocuz:
Thanks!

PR #1358, opened 2017-03-17
Update XiaomiGateway.cpp

corbinmunce:
changing sSwitchCustomSwitch back to sSwitchGeneralSwitch as it did not fix the double update problem


gizmocuz:
Thanks!

PR #1360, opened 2017-03-18
Python options

oohlaf:
Simplify option handling in Python plugins. No more need to base64 encode/decode, just use Python dictionaries.


gizmocuz:
Thanks!

PR #1361, opened 2017-03-18
Add support for InterTechno GW and two common switch types

alexiade:
This adds support for intertechno RF gateway(that sits on UDP) and dumb switches from both brennenstuhl and intertechno that can be commanded with it. It may be possible to use it to command many more devices, but I just happen to have samples of these two. I set it up to be easily expandable with more switch types.

P.S I have no means of fixing the visual studio build as I have no VS available and I don't seem to be able to replicate the other build bot fail(seemingly related to std c++ 11 not being available, is this deliberate? It seems to have no issue with stoi on mac, or in my own linux machine...) on Linux. It's been over a decade since I did anything in CPP :)

### DIFF
--- a/hardware/ITGWBase.cpp
+++ b/hardware/ITGWBase.cpp
@@ -15,6 +15,20 @@
 #include "itgw/ITGWSwitchIntertechno.h"
 #include "itgw/ITGWSwitchBrennenstuhl.h"
 
+/*Intertechno 433Mhz gateway ITGW-433 support
+ *
+ * ITGW is transmitter only, commonly sold in Bauhous brand stores
+ * in eastern parts of europe. The Wlan in the name is misleading,
+ * device requires a cable ethernet to operate and gets it's IP
+ * from DHCP only. To configure it, easyest way of finding out what it got,
+ * is using your smartphone and "Power Switch" app. It broadcasts it's presence
+ * on the network, that this app detects. Or you can listen for the broadcasts
+ * with whatever you preffer for it. You may also want to tell router
+ * to keep giving it the same ip.
+ *
+ * http://intertechno.at/front/produkte/smartphone/itgw-433/
+ */
+
 #ifdef _DEBUG
 	#define ENABLE_LOGGING
 #endif
--- a/hardware/ITGWBase.cpp
+++ b/hardware/ITGWBase.cpp
@@ -154,8 +154,6 @@ bool CITGWBase::WriteToHardware(const char *pdata, const unsigned char length)
 	if (pSwitch->type == pTypeGeneralSwitch) {
 		std::string switchcmnd = GetGeneralITGWFromInt(rfswitchcommands, pSwitch->cmnd);
 
-		_log.Log(LOG_NORM, "ITGW: switch command: %d", pSwitch->cmnd);
-
 		if (switchcmnd.empty()) {
 			_log.Log(LOG_ERROR, "ITGW: trying to send unknown switch command: %d", pSwitch->cmnd);
 			return false;
@@ -168,12 +166,9 @@ bool CITGWBase::WriteToHardware(const char *pdata, const unsigned char length)
       std::string itcmnd;
       if(pSwitch->cmnd == gswitch_sOn) {
         itcmnd = its->GetOnCommandString(ID.c_str());
-        _log.Log(LOG_NORM, "ITGW: switch on command got: %s", itcmnd);
-
       }
       else if(pSwitch->cmnd == gswitch_sOff) {
         itcmnd =  its->GetOffCommandString(ID.c_str());
-        _log.Log(LOG_NORM, "ITGW: switch off command got: %s", itcmnd);
       }
       sstr << its->GetHeader() << itcmnd << txversion << its->GetSpeed();
     }
@@ -182,21 +177,19 @@ bool CITGWBase::WriteToHardware(const char *pdata, const unsigned char length)
       std::string itcmnd;
       if(pSwitch->cmnd == gswitch_sOn) {
         itcmnd = its->GetOnCommandString(ID.c_str());
-        _log.Log(LOG_NORM, "ITGW: switch on command got: %s", itcmnd);
 
       }
       else if(pSwitch->cmnd == gswitch_sOff) {
         itcmnd =  its->GetOffCommandString(ID.c_str());
-        _log.Log(LOG_NORM, "ITGW: switch off command got: %s", itcmnd);
       }
       sstr << its->GetHeader() << itcmnd << txversion << its->GetSpeed();
     }
 
 		sstr << suffix << "\n";
 
-//#ifdef _DEBUG
+#ifdef _DEBUG
 		_log.Log(LOG_STATUS, "ITGW Sending: %s", sstr.str().c_str());
-//#endif
+#endif
 		m_bTXokay = false; // clear OK flag
 		WriteInt(sstr.str());
 		time_t atime = mytime(NULL);
--- a/hardware/ITGWBase.cpp
+++ b/hardware/ITGWBase.cpp
@@ -1,0 +1,336 @@
+
+#include "stdafx.h"
+#include "ITGWBase.h"
+#include "../main/Logger.h"
+#include "../main/Helper.h"
+#include "../main/RFXtrx.h"
+#include "hardwaretypes.h"
+#include "../main/localtime_r.h"
+#include "../main/SQLHelper.h"
+
+#include "../main/mainworker.h"
+#include "../main/WebServer.h"
+#include "../webserver/cWebem.h"
+#include "../json/json.h"
+#include "itgw/ITGWSwitchIntertechno.h"
+#include "itgw/ITGWSwitchBrennenstuhl.h"
+
+
+#ifdef _DEBUG
+	#define ENABLE_LOGGING
+#endif
+
+
+struct _tITGWStringIntHelper
+{
+	std::string szType;
+	int gType;
+};
+
+const _tITGWStringIntHelper rfswitches[] =
+{
+	{ "Intertechno 2 Dials relay", sSwitchTypeInterTechno },               // p9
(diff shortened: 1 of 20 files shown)


gizmocuz:
Thanks! As you see we have some build errors, also on the raspberry pi, we have no c++ 10/11

hardware/itgw/ITGWSwitchBrennenstuhl.cpp
```diff
@@ -26,7 +26,9 @@ std::string ITGWSwitchBrennenstuhl::GetEncodedAddress(const char* ID){
 	std::string res = "";
   std::string _id = ID;
   std::string dipswitch_hex = _id.substr(1,3);
-  int dipswitch = std::stoi(dipswitch_hex, nullptr, 16);
+
+  char * pEnd;
+  int dipswitch = (int) std::strtol(dipswitch_hex.c_str(), &pEnd, 16);
```
gizmocuz: why not using atoi() ?

main/WebServer.cpp
```diff
@@ -4308,7 +4308,8 @@ namespace http {
               (dipswitchcode == "")
               )
               return;
-						sprintf(szTmp, "%03X", std::stoi(dipswitchcode, nullptr, 2));
+            char * pEnd;
+						sprintf(szTmp, "%03lX", std::strtol(dipswitchcode.c_str(), &pEnd, 2));
```
gizmocuz: atol()

alexiade:
Apparenly now the build on VS is failing due to a VS bug... http://stackoverflow.com/questions/16007178/why-passing-t-from-outer-template-as-default-argument-to-stdfunction-causes-co

Any suggestions how to work around it are welcome... I already replaced stoi with strtol that was sufficient for the buildbot, is this sufficient for rasbery pi too? Google failed to give me an answer.

hardware/itgw/ITGWSwitchBrennenstuhl.cpp
alexiade: atoi only does base 10 afaik, this is base 16...

main/WebServer.cpp
alexiade: This is base 2 number, not base 10. The switches have a row of 10 dipswitches for configuration. most people cant do a base 2 to base 10 or base 16 conversion by heart, so the input takes 10 value binary string, that represents the dipswitch states.

gizmocuz:
Seems to bail out here

```
..\hardware\itgw\ITGWSwitchBrennenstuhl.cpp(32): error C2027: use of undefined type 'std::bitset<10>' [C:\projects\domoticz\msbuild\domoticz.vcxproj]
  ..\hardware\itgw\ITGWSwitchBrennenstuhl.cpp(32): note: see declaration of 'std::bitset<10>'
  ..\hardware\itgw\ITGWSwitchBrennenstuhl.cpp(32): note: see declaration of 'std::bitset<10>'
..\hardware\itgw\ITGWSwitchBrennenstuhl.cpp(32): error C2228: left of '.to_string' must have class/struct/union [C:\projects\domoticz\msbuild\domoticz.vcxproj]
```

Thus... don't use std::bitset

hardware/itgw/ITGWSwitchBrennenstuhl.cpp
gizmocuz: Okey, then you can use the normal ::strtol function...

main/WebServer.cpp
gizmocuz: ::strtol ?

alexiade:
Done, at a cost of a bit more ram. C++ is so strange after several years of C# and Java :D

hardware/ITGWBase.cpp
```diff
@@ -15,7 +15,7 @@
 #include "itgw/ITGWSwitchIntertechno.h"
 #include "itgw/ITGWSwitchBrennenstuhl.h"
 
-
+#define _DEBUG
```
gizmocuz: That is not the way...
Please... install VS2017 and just debug ? It's not that complicated... and windows is not a to bad OS either...

gizmocuz:
Can't compare a scripting language against c++
Yes, everything is easier in c#, thanks to the garbage collector for instance...

hardware/ITGWBase.cpp
alexiade: No, it's not the way, that's why I forgot about cleaning it off, but it works for the purpose i needed it. I do my day job on windows. I prefer my free time on Linux. :) Just because I can. I'm told codeblocks is good for c++ but I'm used to netbeans and I couldn't be bothered to find out why the debugging wasn't working the way it was supposed to just to get debug ifdef-ed printouts :P

alexiade:
Yes it is. but it also feels easier in pure C, you need allocate and free in very straightforward way there - everything you allocate you free, if you fail to allocate something, you get a segfault. C++  in these small nuanced aspects is quite unknown to me. But what actually makes it funny for me is the way different compilers and even versions of the same compiler can treat the same code differently. C is probably too mature to suffer(much) from this and scripting language interpreters consider it a bug when this happens, so I haven't really encountered this before. The build bots are nice, or this would become unmanageable fast.

gizmocuz:
Is it possible you can add a link to the hardware web page of the manufacturer at the top of your class (as comment)

Is this hardware based on RFLink ?

```
if (scommand.substr(0, 14).compare("10;rfdebug=on;") == 0) {
				pRFLINK->m_bRFDebug = true; // enable debug
```

We just merged a patch that support extra debugging like

`if (_log.isTraceEnable()) _log.Log(LOG_TRACE,"SQLH HandleOnOffAction: %d OnAction:%s OffAction:%s",bIsOn, OnAction.c_str(),OffAction.c_str());
`

This can be enabled from the command line, maybe thats an option to debug ?

alexiade:
Intertechno gateway is a rf433 commercial transmitter hardware, very limited compared to rflink in capabillities, but very common everywhere you can find Bauhaus construction store brand. The ITGW is sender unit only and must be configured with device codes and issue commands according to switch brand, the switches however are also switchable with RFlink via taught codes.

alexiade:
I will try to add links later today.

gizmocuz:
Thanks!
Okey, but do we even want to add support for this hardware then as it is very/extreme limited in usage ?
Why not buy a real RFXCom (RFXTRX433e) or RFLink then ?

alexiade:
It's commonly around and people are quite likely to already have it. It's got several nice cellphone apps for switching - both manufacturers and openource ones, and its available over the counter - i didnt even know RFLink existed when i bought mine as it was right there next to the switch relays... so id say it's deffinetly worth supporting - it allows people to start using domoticz ie on their synology with what they already have and then upgrade to something better.

Btw, brennenstuhl brand also has a gateway, very similar to itgw, only the command envelope is different, switch command itself is identical. adding support for it based on this would be very little work for someone who has one available.

gizmocuz:
But an RFLink is very cheap... and can do much more (also the RFXtrx433e can of course do a lot more!)

alexiade:
It is cheap, but neither is really readily available. I ordered the bits and am going to make myself rflink, but is being able to solder up an flash arduino really a good prerequisite to set for getting benefit out of domoticz?

alexiade:
Not wanting to support common piece of hardware is pretty surprising. Oh well :)

gizmocuz:
Sorry about that, but there is a difference between new hardware, and hardware that can do the same as others but then less.
RFLink/RFXCom is not expensive, works, can do way more then this hardware.
I always say, Home Automation is not cheap... you want good hardware

Having said that, if you did not want to solder, by an RFXCom/RFXTrx433e, or pay some extra money to have the RFLink soldered (~5 euro was it that much ?... yep it was a whole 5 euro)

alexiade:
This hardware, the patch is for, is not cheap. It's 70 euro hardware, that people CAN buy over the counter and do, unlike rflink that you cant even buy as a product. I am able to make myself rflink but had i not read domoticz source code, i would have not known it existed. Neither did the prossional electronics guy who sold me the intertechno relays. He did know about the gateway. So while i respect your choises on what you merge, i do not agree to your reasoning for NOT supporting common piece of over the counter hardware.

alexiade:
I also do not agree to the notion, that home automation should require soldering and programming skills and cost a fortune. But that is a whole different topic.

stuntteam:
Yes, you can buy an Arduino Mega for 5 euros and rf reciver set for 2 euros from Aliexpress BUT.. if you can not or do not want to do it yourself, you can actually order a ready made RFlink PCB (soldered and tested) including Arduino Mega for 40 euros... that is a non soldering plug&play solution! 
RFlink currently has over 150 plugins/protocols in various flavors supporting hundreds of devices that will never be supported by the Intertechno device...RFXCom also supports far more devices than the Intertechno although it also costs more. 

If you would submit a fully functional patch it might get accepted. But when you submit things that do not fully build, then you are looking at others to spend time on fixing your errors. In that case these persons have priorities and indeed will come to a very understandable conclusion that they do not want to spend time on adding support for a device they do not have and obviously not many people are using as well. On top of that, it does not bring anything new.

alexiade:
As you can see, i put in every effort to make it build on every platform you test for. I would have not bothered had i known, that supporting hardware, that people are likely to own already is not something you wish to do. This is also why I am mildly annoyed. But only mildly. I belive that this argument shows that domotics has an architectural defficency. Support for hardware should be "plugin" like, not hardcoded. ecosystem out there is very diverse, and that such an issue rises at all is unhelpful limitation. Also, i have, as you know, an rflink now, thanks for the project, but i have no intention leaving my intertechno gateway idle. :)

alexiade:
Btw, brennenstuhl relays arent properly supported by rflink - they need the learn mode, where as intertechno gateway i only need to instruct the correct way that i can derive from switch settings... i intend to have an ecosystem where i am not limited by what any single vendor chooses or even can support, but one where any device can easily become supported.

PR #1362, opened 2017-03-18
Only update device state change if it exists in db

hvbommel:
State changes, for non configured domoticz devices, where reported to the database and therefore where showing up in the devices list. Fixed.


gizmocuz:
Thanks!

PR #1363, opened 2017-03-19
[OpenWeatherMap] use own timeouts for http

szczukot:


hardware/OpenWeatherMap.cpp
```diff
@@ -145,6 +145,8 @@ void COpenWeatherMap::GetMeterDetails()
 	{
 		bool bret;
 		std::string szURL = sURL.str();
+		HTTPClient::SetConnectionTimeout(2);
+		HTTPClient::SetTimeout(3);
```
gizmocuz: I don't think there is a need for 'SetTimeout' here
The 'SetConnectionTimeout' at 2 seconds might give problems for user using a dialup network connection
Why does this needs to be so small for you ?

PR #1364, opened 2017-03-19
[MultiFun] fix for uninitialized variable

szczukot:


gizmocuz:
Thanks!
One more... could you check this line:

`if (memset(databuffer, 0, BUFFER_LENGHT) > 0)`

i think this should just read:

memset(databuffer, 0, BUFFER_LENGHT);

correct ?

Or even faster, check the return length of the read function when >0 and issue
databuffer[ret]=0 ?

PR #1365, opened 2017-03-19
Limitless bug fixing RGB V4 commands, added 5 kelvin modes for WW 

chrisverhoef:


gizmocuz:
Okey... let us forget about disco mode can we ?
If someone wants to have a party... use a real remote ?
This is not about serious home automation... NO disco i do not see as home automation but as an incident
unless you are a freak that is discoing 99% of it's live
I want to suggest to remove ALL disco related Stuff from domoticz
Surely users that want to disco can use their android/ios application to turn on this funky mode ?
I do not want this crap in domoticz please

Please take it ALL out

chrisverhoef:
Ok, if that's what you want I will remove it all. But to be sure what we will keep is:
* On/Off
* Set color
* Set bright
* Set NightLight
* Set White
* Set Kelvin
All the rest will be removed, sure?

deennoo:
Oh no please keep disco mode ! 

Using app or remote isn't userfriendly.

Discoing isn't use only for party :
This can be use for notification : 
Red pulse warner trigger on
Green pulse alarme disarme.
White pulse someone is ringing as the door and baby is sleeping.
Etc etc 

Understand that a huge works

That why 2 selector for each groupe can be the easy solution 
1 for disco mode
1 for nightlight set to white and various speed.

With this we can use them on scene/group/event like lua and blocky.

Remove them are a real step backward vs other home automation

gizmocuz:
Okey, lets keep it.... the alarm thing is indeed something related to home automation ;)

deennoo:
i git some idea to help on scene/group/lua/blocky event

have a look a this pic,

https://drive.google.com/file/d/0B46sxK2orIewZjI5RDZ0VnZfYVU/view?usp=sharing

have to be done for each group and for master one

with this kind of widget, we can use all fonction quite easily, and this doesn't ask lot of of work on lua/block/scen/event intgeration

and for rgbw widget this can be a simple one without any option

gizmocuz:
@deennoo , this is crazy, and outside the scope of domoticz
I believe you can go to edit mode and enable all these options
Maybe we do need to remove the disco mode... yes use the App or just use the remote ?

www/views/lights.html
```diff
@@ -80,7 +80,25 @@
 								<button id="speedUp" onclick="appLampSpeedUp()">Speed Up</button><button id="speedDown" onclick="appLampSpeedDown()">Speed Down</button><br />
 								<button id="discoUp" onclick="appLampDiscoMode()">Disco Mode</button><button id="speedUpLong" onclick="appLampSpeedUpLong()">Exclude Lamp</button><br />
 							</div>
+							<div id="optionsRGBW" style="display:none;">
```
gizmocuz: Will this only show up for LimitedLed lights ? Or also for ZWave RGB(W)(WW) lights ? (hope only limitedled)

deennoo:
Just an exemple for a kid 2-3 years old, can't use à phone with an app, remote is to Michel button for him, clik on widget édit is exclued.

But using a simple Amazon dash button where he put his stickers and which trigger disco mode fading rainbow at the lower speed during 1 minute and then turn the bulb to his last state is fantastic for him, and give hum autonomi

And easy to set when all option are aviable, no need to call button URL on script etc etc.

I understand that might bé crazy to show all widget on switch page but final goal is not to always check à screen but live on a easy home.

www/views/lights.html
chrisverhoef: This will show up for all RGBW/RGBWW lights. I will see if I can change this so it's only for Limitless.

gizmocuz:
Forget about Disco, this is home automation deennoo! If you want to party, or let your 2 year old kit party, go ahead... but this has nothing to do with this project
As it seems now, this crazy stuff has been included already, and you can use scripts to enable them (json api)
So buy that Amazon dash button, make a blockly action that calls the json script to do your stuff

deennoo:
Ok understand this, just approve chrisverhoef great job please.

I promise that i will update his wiki with some full lua and blocky exemple to show how to use it

chrisverhoef:
As asked moved all the options to Limitless only.

chrisverhoef:
As agreed: Removed disco related buttons but keep code so it's controllable with json calls.

deennoo:
OK just thinking to provied secret API please

gizmocuz:
Thanks!

deennoo:
@ gizmocuz do you have some Hue bulb ?

Just got a 2.0 bridge 2 bulbs and a led strip

This create 14 * 3 on/off devices ! 
 1 for each scenes available on the bridge /groupe  + bulb contrôle + groupe contrôle. 

And you are crying about 5 clean switch/group for milight ?   

I Real can't understand this position

gizmocuz:
@deennoo , disco mode is not part of home automation, get yourself a real remote to turn this on

PR #1366, opened 2017-03-19
Milight colorfix

mosjonathan999:
I have put back disapeared code,
and fixed the red issue


gizmocuz:
Thanks!, Also for the reformatting ;)

bizziebis:
Does this solve the problem for only some specific Mi-light devices? I have two RGBW lightbulbs, which after this change have red and green swapped.

gizmocuz:
Are you controlling them with an RFLink or with the MiLight gateway ?

bizziebis:
RFLink.
I've reverted this commit and compiled again and now it's working well again.

gizmocuz:
As far as i can see, only this case has changed:

case Limitless_SetRGBColour

Is it possible you could ask on the forum if more people have problems with this (RFLink MiLight)
If yes, we need to revert this patch...

PR #1367, opened 2017-03-19
RelayNet: Create both relay and input devices disabled

hvbommel:
Minor change: Input devices are now created disabled. This was already the case for the relays. The user will enable and rename them before actual use.


gizmocuz:
Thanks!

PR #1372, opened 2017-03-21
A new theme: Progress.

flatsiedatsie:
A new theme that takes Domoticz about as far as I can take it with just CSS. The result of a couple of days of work. The goal was to improve usability, and to a lesser extent attractiveness.

The theme uses a small sub-theme file that houses the most important choices for visual elements, such as the font, colours, background image, and so forth. This makes it relatively easy for people to create new themes using this one as a basis.

The current CSS is still full of manual fixes for the messy html. That's why I've also created a cleaned-up setup.html file that works optimally with this theme. While doing that I also started reshuffling some of the settings' locations in the menu structure.

Details can be found here: 
https://www.domoticz.com/forum/viewtopic.php?f=8&t=16731


gizmocuz:

PR #1374, opened 2017-03-21
Font for the Progress theme

flatsiedatsie:
This is a freely available font. This theme uses two of it's variants (thin and regular).

I've placed them here, perhaps someone with Github skillz could put them in the right folder.

https://www.pineapplejazz.com/public/domoticz/theme/Lato-Thin.ttf
https://www.pineapplejazz.com/public/domoticz/theme/Lato-Regular.ttf

### DIFF
--- a/www/styles/progress/fonts/Lato-Thin.ttf
+++ b/www/styles/progress/fonts/Lato-Thin.ttf
@@ -1,0 +1,1 @@
+The Lato regular font TTF file should be placed here. But I can't upload through the web-interface. And I don't know how Github works. I tried it, it's beyond me.


gizmocuz:
What we actually need is some c++ patches
When a theme is selected, it shall first check for images/js/css/fonts from the theme folder, and then revert back to the domoticz root www folder files

flatsiedatsie:
Sounds cool.  Unfortunately I don't know C++ :-(

I was thinking it would be very useful if Domoticz could load in not just custom.css but also try and load in a custom.js file.

This would allow for nice animations, but also simpel fixes for current html issues through javascript. The current HTML is clearly made by lot of different people, some of whom don't follow CSS standards :-(

It would also be great if all the current 'styles.css" stuff wasn't loaded in universalyl, but  was placed inside the themes. One solution might be to copy all the CSS there into all the themes (I would be willing to help here). That way a new theme doesn't have to use a lot of "!important" in the CSS to override. Without a theme Domoticz would then output totally clean, style-less html.

Back on the subject of the fonts: I purposefully tried to avoid incorporating Google's fonts by the way, for privacy reasons.

gizmocuz:
I made 99% of the HTML, so thanks ;)
I'm a c++ developer, not so HTML, and when i do now, i use AngularJS (not so much in Domoticz, but we really should do)
You already can do all customization in your theme folder via the custom css file
a .js file needs to be loaded via a html file... so unless we also support html overriding, we can not do this

gizmocuz:
I think i will look into this after the next stable release... first need to get all stable again

flatsiedatsie:
haha, ok :-D Well most of the HTML is fine! And Domoticz definitely rules when it comes to very wide support. I was going through the list of devices today, it's incredible. If you'll let me I'll be happy to help with the HTML a bit. Me, I'm not that great with C++ ;-)

I'm trying to follow you on the .js thing (not that it matters really, it's just a fun wish). In my head it would involve finding the code that currently loads in [foldername/custom.css into the html, and then just copying that line but modify it to load in a '[foldername]/custom.js file as well.

It doesn't really matter if the custom.js file even exists in the theme folder or not, because with html it won't cause an error anyway, just a small 404.

gizmocuz:
404 is very bad... specially when using appcache/offline applications
If you search in all the sources for 'custom.css' (with visual studio/total commander/grep) you see it is in index.html

flatsiedatsie:
Hmm. Maybe we can place an empty .js file in all the themes to solve the 404?

gizmocuz:

PR #1375, opened 2017-03-21
Mayor clean-up of CSS, bugfixes, and.. content reshuffle

flatsiedatsie:
I've spent a few days creating a new theme for Domoticz, and during that time I came across some things.

- Fixed error in HTML (unclosed form tag)
- Cleanup: standardised a lot of things that had gotten messy over time, including placing everything in a span, adding titles where they were missing, and lots more.
- Removed all extraneous non-css styling (br tags mostly, but also very divergent inline styles). Positioning should be done with css. I'll help. And: this saves you time! It's really not necessary anymore to add width manually.
- Quite a big one: Reshuffled the various settings for improved usability.
- Added new 'related options' shortcuts, also for usability.
- (all the latest commits are incorporated)

See also:
https://www.domoticz.com/forum/viewtopic.php?f=8&t=16731

### DIFF
--- a/www/views/setup.html
+++ b/www/views/setup.html
@@ -1,1048 +1,1478 @@
 <div id="dialog-findlatlong" title="Find your Geo location">
-    <form>
-        <table class="display" id="latlongtable" border="0" cellpadding="0" cellspacing="0">
-        <tr>
-          <td align="right" style="width:60px"><label for="name">Address: </label></td>
-          <td>
-			<input type="text" id="address" style="width: 300px; padding: .2em;" class="text ui-widget-content ui-corner-all" /><br>
-			(Example: Streetname Number City Country)
-          </td>
-        </tr>
-        <tr>
-			<td></td>
-			<td><button id="getlatlong">GetLatLong</button></td>
-        </tr>
-        <tr>
-          <td align="right"><label for="lbllat">Latitude: </label></td>
-          <td><input type="text" id="latitude" style="width: 90px; padding: .2em;" class="text ui-widget-content ui-corner-all" /></td>
-		    </tr>
-        <tr>
-          <td align="right"><label for="lbllng">Longitude: </label></td>
-          <td><input type="text" id="longitude" style="width: 90px; padding: .2em;" class="text ui-widget-content ui-corner-all" /></td>
-		    </tr>
-		    </table>
-    </form>
+	<form>
+		<table border="0" cellpadding="0" cellspacing="0" class="display" id="latlongtable">
+			<tr>
+				<td><label for="name">Address:</label></td>
+				<td>
+					<input class="text ui-widget-content ui-corner-all" id="address" type="text">
+					<p>(Example: Streetname Number City Country)</p>
+				</td>
+			</tr>
+			<tr>
+				<td></td>
+				<td><button id="getlatlong">GetLatLong</button></td>
+			</tr>
+			<tr>
+				<td><label for="lbllat">Latitude:</label></td>
+				<td><input class="text ui-widget-content ui-corner-all" id="latitude" type="text"></td>
+			</tr>
+			<tr>
+				<td><label for="lbllng">Longitude:</label></td>
+				<td><input class="text ui-widget-content ui-corner-all" id="longitude" type="text"></td>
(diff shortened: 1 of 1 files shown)


gmaslowski:
+1

gizmocuz:
Could you post a screenshot of before and after?

gizmocuz:
You have to many pull requests... lets do this one by one
- first make sure the original theme/css stays functional of course
- we need a patch as i discussed in another of your pull requests that will look for js/images/css/... in the selected theme folder first, then revert back to domoticz's root www files. We might even do this for html files ? Maybe all files ?

This way we can support full theming

flatsiedatsie:
Loading in all HTML, wow, that would be quite something!

But for now though, I hope my additions are useful and can be integrated into the default. Before I uploaded the new file I checked if everything was still working on the default, and it was fine.

Tonight I'll be able to upload some before and after screenshots for the other pages.

The removal of BR tags shouldn't be a problem, it will just look a bit more condensed. My goal is to also create a new CSS update to the current default theme that adds spacing without needing BR tags.

Another option would be to develop the theme I created into a new default theme. Would you be interested in that?

gizmocuz:
Let me remind you, i want my original theme to look as it is !
So removing the BR tags is NOT an option, unless you going to solve this in CSS, which some people do not like...
I am happy to apply patches that fixes/cleans up the original HTML/CSS/JS code, as long as it does not touch the look/feel, unless something was corrupt
Is that is the case with all your patches, you might as well close them all

flatsiedatsie:
Whoa, ok. 

I hear what you're saying about preserving the usability of the current design.

But Domoticz usability and theme-ability can be improved, and this is something I'm trained in. As a designer I think I can help Domoticz evolve here. 

I have not received any replies to earlier attempts to communicate with the core team via the forum about my intention, so I can understand if this comes out of the blue a bit. I'm just trying to help.

flatsiedatsie:
I'll wait with removing the BR tags until we discussed how best to create the spacing instead:
- By modifying the current theme.
- By creating a new default theme.

The patches I've supplied are mostly about changing the HTML to use divs instead of tables to position things on the page, and to add titles where they were missing.

I'm sorry if some of the proposals were too sudden. I can create the patches but leave in the br tag removal.

gizmocuz:
How are your AngularJS skills ? we could use someone to make everything more angular, this way theming is also easier.
If there are errors in the html/js/css , sure please submit patches for this.

My layout, your layout, it's a matter of taste. My design works good, i also started with the 'html admin template' (like react) and quickly decided not too

I can summarize what i do not like about your theme (and again, this is a 'taste' issue of course, no criticism, it's my taste)

- The background in the first screenshot
- The Light theme (there is a reason why most designs are dark these times (whole studies)
- The fonts, to small and fuzzie (hope it is the screenshot only?)
- the small widgets in height on for example the Lights tab, i wonder how this goes with my big fingers coz 90% of the time i use a mobile device. Look at the 'selector switch', how about when we have 10 (we support up to 10)
- In the settings, the left menu, started with this, took too much room, that's why the tabs, and nothing wrong with tabs, maybe they can look better. (remember people use a tabled)
(bootstrap styled, like https://www.w3schools.com/css/css_navbar.asp)

Likes:
- Table/div background in the edit/setup screens, maybe a little more transparent ?
- Organisation of the Setup page

Sure we can tweak the CSS style to make it more darker, it's already better in the second screenshot with the mountain, but i would not pick a mountain but some kind of texture to start with

I also have some 'ideas', like instead of the top menu/tabs, a docking/floating/auto-hide icon bar on the left of the screen, (50px x xxheight), that we can pull out from the side:
http://seegatesite.com/bootstrap/simple_sidebar_menu.html
https://code.tutsplus.com/tutorials/outside-the-box-navigation-with-jquery--net-5535

Maybe you could also help with this:
As for design, there are two setup pages that do not fit in the current folder structure:
"eventsframe.html"
"zwavetopology.html"

About this patch, maybe start step by step fixed broken html/css/js, you can also directly contact me on the forum

flatsiedatsie:
I've contacted you on the forum.

gizmocuz:

PR #1378, opened 2017-03-21
Notifications implemented per minute Last Update check

jvandenbroek:
Last week I implemented the per hour timeout check, however I wanted something more sophisticated and be able to check every minute. I extended the notification function to store last update values, so it doesn't have to query the DB every minute. First I extended the CheckAndHandleNotification function, this however required so much rewriting, a separate function worked out better. I've now tested this with different kind of values (greater, below combined) and it works very well.

The hourly called CheckDeviceTimeout function is now as before, so disabled by default.


gizmocuz:
Thanks!!

PR #1379, opened 2017-03-21
Replaced table with CSS

flatsiedatsie:
It's better to use CSS to place things on the page, and not tables.


gizmocuz:

PR #1382, opened 2017-03-21
Replaced table with CSS

flatsiedatsie:
CSS should be used for positioning.

Makes theming a lot easier too :-)

All these files had a weird, lone, unclosed p tag in them too. Poor fella.


gizmocuz:

PR #1383, opened 2017-03-21
Add CSS bootstrap wrapper

flatsiedatsie:
And also fixed the title from h3 to h2.
Removed BR tags that were used as spacers too. Why do people keep using those..


gizmocuz:
If you make patches to the default HTML styles/files, please post a before/after screenshot, even if there are the same.

gizmocuz:

PR #1385, opened 2017-03-21
Removed BR tags

flatsiedatsie:
While looking for a way to add a 'bannav' id/class to the scenes overview page,I couldn't find it. Along the way I removed some BR tags instead. CSS should be used for spacing.


gizmocuz:
You can't just remove <br> tags as this will mess with the original layout
And please post before/after pictures using the original theme

PR #1390, opened 2017-03-22
Fix and cleanup for 1-wire

Zuikkis:
Related to https://github.com/domoticz/domoticz/pull/1162

This commit includes szczukot's changes, plus suggestions by gizmocuz.

Now sensors are not updated in bursts, but one at a time on a fixed intervals.

BIG CHANGE: All sensor updates are now done with SendTempSensor() and similar functions.
This will create new temperature and humidity sensors, because the subtype changes..
Perhaps there should be a script that automatically updates the database?

I only have 1-wire temperature sensors, so these changes are not tested on other devices.


Zuikkis:
AppVeyor check fail seems unrelated to this commit, 
".\hardware\openwebnet\bt_openwebnet.cpp(5): fatal error C1083: Cannot open include file: 'stdafx.h': No such file or directory [C:\projects\domoticz\msbuild\domoticz.vcxproj]"

szczukot:
Where do You call PrepareDevices ?

Zuikkis:
First line in "C1Wire::StartHardware()".. That part hasn't changed from your code, I think?

gizmocuz:
Could you increase the database version number, check in SQLHelper.cpp how we do database update patches, query all 1-wire sensors, fix the id's, and put them back ?
This way all users have no problems with new ID's

gizmocuz:
I will fix the bt_openwebnet.cpp error, think i forgot a compiler flag for release mode, checking this now

Zuikkis:
Ok, added the database fix. :) I tried to double check everything, I'm pretty sure only temperature and humidity sensors have changed to new subtype.. Others should be as before.

hardware/1Wire.cpp
```diff
@@ -509,83 +481,33 @@ void C1Wire::ReportTemperatureHumidity(const std::string& deviceId, const float
...
-	sDecodeRXMessage(this, (const unsigned char *)&tsen.RFXMETER, NULL, 255);
+	SendMeterSensor(deviceIdByteArray[0], deviceIdByteArray[1]+unit, 255, (const float)counter, "Counter");
 }
 
 void C1Wire::ReportVoltage(const std::string& deviceId, const int unit, const int voltage)
 {
 	if (voltage == -1000.0)
 		return;
 
+	// There is no matching SendXXX() function for this?
```
szczukot, line 533: i think SendVoltageSensor
but thiis other type (like temeprature and humidity)

gizmocuz:
Great! Thanks !
One small request, these days we have a new query function "safe_query" could you use this ?

szczukot:
What with people who use OWFS ? All temp and humidity devices change type ! But owfs all time reported as different type.

Zuikkis:
Oh, didn't think about owfs! I have to change the types in owfs source, so
they match kernel.

I'll do these changes later this evening.

gizmocuz:
Please do not reply with email, or at least make sure you do not resend the original post (i edited your post)

Zuikkis:
Ah, sorry for the email reply! :)

Now it is using safe_query. 

szczukot: I think this code works fine with OWFS. All sensor updating is done in hardware/1Wire.cpp, which is common code for kernel and OWFS and Windows 1-wire.. So all sensor types are the same, no matter what system is used.

szczukot:
But OWFS keep device id in id1,id2,id3,id4. 
After report by SendXXX - we lost this id ? I think

Zuikkis:
Hmm.. Where is that code? I see no difference really..

If you look at SendTempSensor() function in DomoticzHardware.cpp, it is almost identical to 1Wire.cpp's old version of ReportTemperature. Just different subtype... id1/id2 is handled exactly the same way, it is not changed.

szczukot (commented):
Before:
tsen.TEMP.id1=(BYTE)deviceIdByteArray[0];
tsen.TEMP.id2=(BYTE)deviceIdByteArray[1];
tsen.TEMP.battery_level=9;
using sendtempsensor:
tsen.TEMP.id1 = (NodeID & 0xFF00) >> 8;
tsen.TEMP.id2 = NodeID & 0xFF;
tsen.TEMP.battery_level=255; // not important and easy to fix
is it swapped ?

hardware/1Wire.cpp
Zuikkis: But, in the new 1Wire.cpp / ReportTempSensor, I do this:

        uint16_t lID = (deviceIdByteArray[0] << 8) | deviceIdByteArray[1];

        SendTempSensor(lID, 255, temperature, "Temperature");
}

Then in the SendTempSensor() they are shifted back to bytes. It is the same end result. 

I'm not sure about SendVoltageSensor, maybe we shouldn't touch it this point?

szczukot: You're right. It should work.

I don't help with voltage - i don't have any sensor of this type.
But it look this same as temp and humidity: only type is different.

gizmocuz: We have a merge problem, is it possible to solve this if this patch needs to be merged?

Zuikkis: Bah, how is this possible, few days ago it said it merges cleanly? Something changed in 1-wire in in between?

Zuikkis: Ok, it's the szczukot's 1-wire DS2890 merge, and also MySensors database fixes in SQLHelper.. I fixed those now, but building locally first to see that everything works.

Zuikkis:
Checks still running, but should be fine now.

gizmocuz:
Thanks!

PR #1392, opened 2017-03-22
Fixes for crash in Satel and MultiFun plus some compiler warnings

DebugBill:
Fixes for crash in Satel and MultiFun plus some compiler warnings:
- Warning in TeleinfoBase.cpp trigerred in Windows only
- Warning in SQLHelper.cpp which was not created by me.... but I'm willing to help :-)
- And most important a fix in SendAlertSensor in order to accept a null pointer as "default name" for the sensor (don't ask me why it is needed, but it exists in the code so I had to handle it)

I have one person testing the binary (szczukot). I'm expecting a confirmation from him that it works anytime now.

### DIFF
--- a/hardware/DomoticzHardware.cpp
+++ b/hardware/DomoticzHardware.cpp
@@ -835,7 +835,7 @@ void CDomoticzHardwareBase::SendAlertSensor(const int NodeID, const int BatteryL
 	gDevice.subtype = sTypeAlert;
 	gDevice.id = (unsigned char)NodeID;
 	gDevice.intval1 = alertLevel;
-        sprintf(gDevice.text, "%.199s", message.c_str());
+	sprintf(gDevice.text, "%.63s", message.c_str());
 	sDecodeRXMessage(this, (const unsigned char *)&gDevice, defaultname, BatteryLevel);
 }
 
--- a/hardware/DomoticzHardware.cpp
+++ b/hardware/DomoticzHardware.cpp
@@ -829,14 +829,14 @@ void CDomoticzHardwareBase::SendSoundSensor(const int NodeID, const int BatteryL
 	sDecodeRXMessage(this, (const unsigned char *)&gDevice, defaultname.c_str(), BatteryLevel);
 }
 
-void CDomoticzHardwareBase::SendAlertSensor(const int NodeID, const int BatteryLevel, const int alertLevel, const std::string &message, const std::string &defaultname)
+void CDomoticzHardwareBase::SendAlertSensor(const int NodeID, const int BatteryLevel, const int alertLevel, const std::string &message, const char* defaultname)
 {
 	_tGeneralDevice gDevice;
 	gDevice.subtype = sTypeAlert;
 	gDevice.id = (unsigned char)NodeID;
 	gDevice.intval1 = alertLevel;
         sprintf(gDevice.text, "%.199s", message.c_str());
-	sDecodeRXMessage(this, (const unsigned char *)&gDevice, defaultname.c_str(), BatteryLevel);
+	sDecodeRXMessage(this, (const unsigned char *)&gDevice, defaultname, BatteryLevel);
 }
 
 void CDomoticzHardwareBase::SendGeneralSwitchSensor(const int NodeID, const int BatteryLevel, const int switchState, const char* defaultname, const int unitCode)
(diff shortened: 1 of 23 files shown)


gizmocuz:
Thanks, please let me know when to merge

main/SQLHelper.cpp
```diff
@@ -2967,7 +2967,8 @@ std::vector<std::vector<std::string> > CSQLHelper::query(const std::string &szQu
 
 	if (_log.isTraceEnable()) {
 		_log.Log(LOG_TRACE, "SQLQ query : %s", szQuery.c_str());
-		if (!_log.TestFilter("SQLR"))
+		char tmp[5]="SQLR";
+		if (!_log.TestFilter(tmp))
```
gizmocuz: Why this ?

hardware/DomoticzHardware.h
```diff
@@ -81,7 +81,7 @@ class CDomoticzHardwareBase
 	void SendUVSensor(const int NodeID, const int ChildID, const int BatteryLevel, const float UVI, const std::string &defaultname);
 	void SendBlindSensor(const int NodeID, const int ChildID, const int BatteryLevel, const int Command, const std::string &defaultname);
 	void SendSoundSensor(const int NodeID, const int BatteryLevel, const int sLevel, const std::string &defaultname);
-	void SendAlertSensor(const int NodeID, const int BatteryLevel, const int alertLevel, const std::string &message, const std::string &defaultname);
+	void SendAlertSensor(const int NodeID, const int BatteryLevel, const int alertLevel, const std::string &message, const char * defaultname);
```
gizmocuz: why going back to char * ? Please don't, you see all sensors have "std::string &defaultname", this one should too

hardware/DomoticzHardware.cpp
```diff
@@ -829,14 +829,14 @@ void CDomoticzHardwareBase::SendSoundSensor(const int NodeID, const int BatteryL
 	sDecodeRXMessage(this, (const unsigned char *)&gDevice, defaultname.c_str(), BatteryLevel);
 }
 
-void CDomoticzHardwareBase::SendAlertSensor(const int NodeID, const int BatteryLevel, const int alertLevel, const std::string &message, const std::string &defaultname)
+void CDomoticzHardwareBase::SendAlertSensor(const int NodeID, const int BatteryLevel, const int alertLevel, const std::string &message, const char* defaultname)
```
gizmocuz: why going back to char * ? Please don't, you see all sensors have "std::string &defaultname", this one should too

main/SQLHelper.cpp
DebugBill: Because of this:
[ 22%] Building CXX object CMakeFiles/domoticz.dir/main/SQLHelper.cpp.o
/home/pi/domoticz/main/SQLHelper.cpp: In member function ‘std::vector<std::vector<std::basic_string<char> > > CSQLHelper::query(const string&)’:
/home/pi/domoticz/main/SQLHelper.cpp:2970:30: warning: deprecated conversion from string constant to ‘char*’ [-Wwrite-strings]
   if (!_log.TestFilter("SQLR"))

hardware/DomoticzHardware.h
DebugBill: I couldn't agree more, but the developper of SatelIntegra insisted in pull request 1356 to keep passing a NULL pointer:

szczukot 5 days ago Contributor
I don't want set name for this sensor.
Must be NULL instead "Alert"

But now that it is clear this is the root cause, I'd rather revert to std:string and replace the null pointer with an empty string.

DebugBill:
szczukot confirmed the binary I sent is working. Problem was indeed with passing a null pointer to function SendAlertSensor. 

Tell me what you prefer, I'm ready to revert to std::string and modify SatelIntergra.cpp instead.

DebugBill:
Just moved again to std::string in arguments. 
I sent a binary to szcukot for testing again.

main/SQLHelper.cpp
oohlaf: This is not the proper fix for the warning, I already created a fix in my branch, but intended to fix additional warnings, but I'll just create a pull now

gizmocuz:
Thanks all for the great theme work and the quick fix!

PR #1395, opened 2017-03-22
GPIO configurable debounce and period timers

jvandenbroek:
Make debounce and period timers user configurable. Also changed the new behavior of delayed startup with forced updates to only when active users with remote sharing enabled are configured.


gizmocuz:
Thanks!

PR #1398, opened 2017-03-23
[OpenWeatherMap] Use own timeouts

szczukot:


hardware/OpenWeatherMap.cpp
```diff
@@ -145,6 +145,8 @@ void COpenWeatherMap::GetMeterDetails()
 	{
 		bool bret;
 		std::string szURL = sURL.str();
+		HTTPClient::SetConnectionTimeout(5);
+		HTTPClient::SetTimeout(5);
```
gizmocuz, line 149: Why do you need to set this type of timeout ? You know it's not a connection timeout ?
Whats wrong with the default timeout of 10 seconds ? I suppose nothing, and if you have a fast internet connection it will take a sec ?

szczukot: This timeouts are for safe.
All modules may set this global timeouts (logitech for example), and if other don't set this value, will use values from earlier module.

szczukot: GSM or Modem 33600 is faster, and don't need 90 sec for get a few bytes of json.
I may increase connectiontimeout, but this two lines must stay. I don't known when other modules set this values, and my modules stop work in this situation. This is global value (i think, this is bad idea)

szczukot:
Why do You closed it ??
I did a easy test: add Logitech hardware and set timeout for it to 1000 ms.
Now HttpClient have timeout=1s
I try connect to OpenWeatherMap every few seconds. Many times i get TIMEOUT NOW !

gizmocuz:
Then don't touch it in the logitech 
I know where you want to go to, and as said before... No 
I can close any pull requests.. They are not mandatory, specially when my comments are ignored 

If you need some kind of strange timeout for your class... Which I don't expect.. Then create your own function 
But you do not need to do this for logitech..damn... This even connects locally!

szczukot:
-I can close any pull requests
Nice
-They are not mandatory, specially when my comments are ignored
It is not true. I Replied and I change code (I lengthened timeouts).
-But you do not need to do this for logitech..damn... This even connects locally!
You don't understood. Logitech maybe connect localy, but set timeouts for 20 other modules, whcich they use this same class (HTTPClient). 
I give up. Fortunately I do not use Logitech. I wish good luck to others users, who use Logitech with other modules.

gizmocuz:
Again, Pull requests are "requests" and have NOT be taking into account.
Please do not mess with the HTTPClient class or it's usage

PR #1399, opened 2017-03-24
Added retro-compatible handling of dimmers

LemarinelNet:
Since the last patch broke handling of dimmers (because of 100 steps is done via a different type), I put back the old piece of code, so both kind of devices (old created and newly ones) are handled correctly


gizmocuz:
Thanks, but could you also have a look at increasing the database version and making a patch in SQLHelper.cpp so that old MySensor switch types are converted to the new switch type ? (And test...)
This way current stable users won't have issues

gizmocuz:
Make a patch to fix the wrong subtype for the general switch, and also a conversion from old type to new.
Something similar you can do for converting from the old format to the new, because currently when old users are restarting the node, they will get the new sensor type in domoticz, but at the present (without a patch) also keep their old one
https://github.com/domoticz/domoticz/commit/345f92cd0ab0f0f0c5762646d704905ad088348a

gizmocuz:
Make a patch to fix the wrong subtype for the general switch, and also a conversion from old type to new.
Something similar you can do for converting from the old format to the new, because currently when old users are restarting the node, they will get the new sensor type in domoticz, but at the present (without a patch) also keep their old one
https://github.com/domoticz/domoticz/commit/345f92cd0ab0f0f0c5762646d704905ad088348a

LemarinelNet:
Ok, 
I'll  make a patch this afternoon.

gizmocuz:
If you have an old database backup, please use this as test

PR #1400, opened 2017-03-24
Try to make OSX build independent of specific python3 homebrew version

oohlaf:


oohlaf:
Don't merge yet, want to see the output of travis builder

oohlaf:
Works for as long 3.6 remains the latest version.
I do see a lot of warnings when using clang as compiler

gizmocuz:
I dont know about clang.... but maybe we should look at those warnings
Do you want me to merge this ?

oohlaf:
Yes, this ready for merging as it will make the build indepent of minor versioning of homebrew packages. With this you only need to update the travis file when a new 3.X is released, not on every 3.6.y.z

By default, using clang as compiler, it provides a whole lot more warnings on coding practices and potential errors. That has nothing to do with the changes here.

gizmocuz:
Thanks!

PR #1401, opened 2017-03-24
Notifications delay timeouts when just started

jvandenbroek:
When Domoticz is stopped for a while and restarted, it will directly send timeout notifications when relative small timeouts are set. This won't send notifcations until Domoticz startime > sensor timeout.


gizmocuz:
Thanks!

PR #1402, opened 2017-03-24
[1wire] write wiper value to DS2890

szczukot:


oohlaf:
No need to change <python.h> to "python.h", a proper fix is already in master, please merge it here to get it to compile again on OSX.

gizmocuz:
Thanks, Maybe you want to check the result of GetWiper before reporting it in (1Wire.cpp, ~ line 439)

szczukot:
I have check in ReportWiper:
if ((wiper < 0) || (wiper > 255))
 		return;

gizmocuz:
True, and i think it was already changed with the clang patches

gizmocuz:
Thanks!

PR #1407, opened 2017-03-25
Fix missing negation on PyDict_Check

oohlaf:
Only set default options when no options are provided, or when an invalid options dict is given.


gizmocuz:
Thanks!

PR #1408, opened 2017-03-26
Fixed nasty bug affecting only Triphasé users

DebugBill:
Lots of cleaning now that the code is stable, spelling mistakes fixing, commented lines and dead code removed, more debug information added

Added another CRC calculation algorithm after discover in the official documentation that there is a second one, seldom used currently but being the new standard. We like to make it complex in France! Automatic choice of the right CRC to apply. 

Added errors to the log if the CRC calculation is wrong. This might affect users with poor connections to their meters as it will fill their logs with messages. But it will give them a chance to know about the problem and maybe fix it. Throws one error per line with wrong CRC.

And cherry on the cake, fixed a nasty bug due to string comparaisons being done on substrings too. flags IINST1, IINST2 and IINST3 where all considered as equal to IINST because IINST is included in IINSTx.
The code involved is unecessarily complex and comes from the legacy implementation. I did not touch it but I think I will now. 

We have at least one triphasé user. It now works for him


gizmocuz:
Thanks!

PR #1413, opened 2017-03-27
Resolved crash on Raspberry Pi when added icon zip files via a plugin

dnpwwo:


oohlaf:
👍 That fixes the warning: offset of on non-POD type 'CImage' [-Winvalid-offsetof]

One warning remains btw:
```
warning: delete called on 'Plugins::CPluginTransport' that is abstract but has non-virtual destructor [-Wdelete-non-virtual-dtor]
                                        delete m_pTransport;
```

gizmocuz:
Thanks!

PR #1417, opened 2017-03-29
Teleinfo Serial Adapter: Code simplification, nested lookup tables removed

DebugBill:
- After facing a nasty bug on a legacy part of the code, I removed nested lookup tables which were adding complexity and making the code difficult to understand. They had no obvious benefit on performance or maintenability (in my opinion of course :-))
- Removed dead code
- Updated version numbers


gizmocuz:
Thanks!

gizmocuz:
Is it possible to have a look here
https://www.domoticz.com/forum/viewtopic.php?f=6&t=16919

Maybe there is a problem with this pull request...?

PR #1419, opened 2017-03-29
Add =, !=, <= and >= to notification conditions

DebugBill:
Now that I use Alerts to pass messages associated to various levels, I needed to send a notification based on some precise level. The "=" operator was missing. 
While at it, I also added a few others.....

What do you think? Is that usefull? I am still testing it but it seems to work fine. It was also an opportunity to factorise some code in NotificationHelper while adding functionality

If we keep it, news translations are needed for
Lower
Lower or Equal
Equal
Not Equal
More 
More or Equal


gizmocuz:
Is it an option to change all texts to the 'signs' (=, !=, <= and >=,...) this way we do not need any translation anymore

DebugBill:
Done. No need for translation anymore. I used "<>" rather than "!=" as I think it is less "geeky"... but this is really a question of taste.

gizmocuz:
Thanks, could you also remove the data-i18n fields for the options ?
Could you post a screenshot how the dropdown looks ?

DebugBill:
Done.

Here is a screenshot of the new conditions


notifications

gizmocuz:
Great thank you !
What do you think prefer ... text or symbols ? Maybe you where correct adding the new translations ?
If you also think so, please add the text labels again and i will add the translations asap...

DebugBill:
Really your call. I like both, and it is true that symbols are a way to avoid the translation phase.

gizmocuz:
Okey, then sorry for the double work, but please place back the textual notations... i think it is easier for users without a math background ...

DebugBill:
Done, you can merge

gizmocuz:
Thanks!!

PR #1420, opened 2017-03-29
Allow Evohome USB to specify baud rate

codeaholics:
There is an increasing community of makers who are building custom Evohome interface hardware, not all of which runs (or is capable of running) at 115,200 baud. This patch allows the Evohome USB hardware (as distinct from the Evohome Script hardware) to have its baud rate configurable, whilst allowing hardware created previously to correctly default to the previously hard-coded value of 115,200.


gizmocuz:
Thanks!

PR #1421, opened 2017-03-30
Increment Device Number to allow correct zone numbering. 

brucemiranda:
Increment the Device number for relay devices used as part of an Evohome zone to allow correct zone numbering. This is for relay devices other than the CH and DHW zone valve relays.


brucemiranda:
Do I need to provide anything else for this bug fix to be Pulled?

gizmocuz:
No, just wait until it is reviewed.. which i'm doing now

gizmocuz:
Thanks!

PR #1422, opened 2017-03-30
Evohome sane logging

codeaholics:
Some changes to the logging in the Evohome hardware to make logging a bit more useful, sensible, etc. There is more to come in this space...


gizmocuz:
Thanks! Maybe you could use the new TRACE logging feature ?

codeaholics:
I'll definitely look at that. There's a lot of confusion in the logging code in the Evohome driver, so rationalising it is high on my priority list.

PR #1423, opened 2017-03-31
Timer corrections

gordonb3:
1. Next firing time calculation is wrong for sunset/sunrise timers because it takes today's sun values to set tomorrow's scheduled time. https://github.com/gordonb3/domoticz/commit/67f89ac66d86ea548e57be3b7366e8fec1525b43 fixes this by reloading the schedules when sunset/sunrise values change

2. Next firing time calculation contains a default routine that adds 24 hours to the calculated time, either enforced by the caller or because the time references the past. In time zones that have DST this is wrong twice a year. https://github.com/gordonb3/domoticz/commit/de3db55dc1e2e0bb8ff65435cc46ab9ac44023cd fixes this by using a sane time calculation.

### DIFF
--- a/main/mainworker.cpp
+++ b/main/mainworker.cpp
@@ -212,6 +212,7 @@ MainWorker::MainWorker()
 	m_ScheduleLastMinuteTime = 0;
 	m_ScheduleLastHourTime = 0;
 	m_ScheduleLastDayTime = 0;
+        m_LastSunriseSet = "";
 
 	m_bHaveDownloadedDomoticzUpdate=false;
 	m_bHaveDownloadedDomoticzUpdateSuccessFull=false;
@@ -517,6 +518,7 @@ bool MainWorker::GetSunSettings()
 	sprintf(szRiseSet,"%02d:%02d:00",sresult.SunSetHour,sresult.SunSetMin);
 	sunset=szRiseSet;
 
+	m_scheduler.SetSunRiseSetTimers(sunrise, sunset);
 	std::string riseset=sunrise.substr(0, sunrise.size()-3)+";"+sunset.substr(0, sunrise.size() - 3); //make a short version
 	if (m_LastSunriseSet != riseset)
 	{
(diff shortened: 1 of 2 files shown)


main/mainworker.cpp
```diff
@@ -2055,7 +2058,7 @@ void MainWorker::ProcessRXMessage(const CDomoticzHardwareBase *pHardware, const
 
 	const_cast<CDomoticzHardwareBase *>(pHardware)->SetHeartbeatReceived();
 
-	char szDate[100]; szDate[0] = 0;
+	char szDate[100];
```
gizmocuz: Why are my previous changes removed here ? (and below)

gordonb3: Recent change?

I had this branch uploaded four days ago. The master branch in my fork is the same age. Wanted to run some checks before pushing them as some users reported that the sunrise/sunset based schedules did fire at (more or less) the correct time.

gizmocuz: Yes, your current pull request is reverting some of my changes

gizmocuz: Have a look on github and compare current against your changes
My changes are needed to not log a date/time when this is disabled via the command prompt (for example for syslog users)

gizmocuz:
Thanks, but how come your original code is old, for instance also 
case HTYPE_EVOHOME_SERIAL:
case HTYPE_EVOHOME_SCRIPT:
else if (subType == sTypeKwh)

It looks you might have started with some old code, maybe you could update this and submit the changes ?

gordonb3:
It is based on 3.7153 from March 25 (commit #f89f129)

I'm a bit stunned to still see additional changes even though they don't show up in the local diff

gordonb3:
Moved: request #1429

gizmocuz:
It seems we have an issue now:

https://github.com/domoticz/domoticz/issues/1443

This is not working anymore (01:00am)

After Sunset' (10/04/17) ==> ON action
01:00 (11/04/17) ==> OFF action 

The Off action is not executed

Is it possible you can have a look ?

gordonb3:
Hmm... My `on time` timers are running correct. Could be a race condition on the schedule reload, causing it to move to the next day before it fires.

gizmocuz:
Yes, i think it is caused by the schedule reload.. maybe we should revert this ?

gordonb3:
Probably best to disable that for now. Still think it's a good idea to reload them, but to avoid conflicts like these we should likely filter on schedules that are relative to sun times only.

gizmocuz:
Maybe after the moment of DST ? But currently all these timers won't work anymore
What would you advice? Revert this patch ?

gordonb3:
Pull request #1461

PR #1424, opened 2017-03-31
Correctly refresh Evohome parameters when switching between hardware

codeaholics:
I missed this from my previous Evohome baud rate PR. When switching between hardware drivers on the table view, the Evohome parameters were not correctly loaded into the controls.


gizmocuz:
Thanks!

PR #1425, opened 2017-03-31
Temporary css for possible migration to new default theme.

flatsiedatsie:
The new javascript will add a class to the TR, which will allow for more flexible ways for items to respond to status change (like low battery). This css makes sure the current default theme is not affected.


gizmocuz:
Just send you an email and created a new branch, lets continue there (and on slack) first

PR #1426, opened 2017-03-31
Fixes for XiaomiGateway

corbinmunce:
Fixes for wall switch.  Fix for round wireless switch to always update Domoticz. Totally ignore 'heartbeat' messages.

### DIFF
--- a/hardware/XiaomiGateway.cpp
+++ b/hardware/XiaomiGateway.cpp
@@ -496,7 +496,7 @@ bool XiaomiGateway::StartHardware()
 	m_bIsStarted = true;
 	
 	// update any CustomSwitch Xiaomi devices to GeneralSwitch		
- -	//m_sql.safe_query("UPDATE DeviceStatus SET SubType=73 WHERE(HardwareID == %d) AND (SubType == 72)", m_HwdID);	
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thanks!

marcotini:
@deennoo it doesn't surprise me. In my experience I can tell you that my sensor go on with an heartbeat in middle night and turn on the light lol but @corbinmunce could not replicate my problem.

deennoo:
@marcotini, which kind of sensor please ? (round switch? door switch ? motion ? wireless 1 button or 2 button ? wired 1 button or 2 button ?

on future merge heartbeat will be disable to solve this

marcotini:
@deennoo as you can see here: https://github.com/marcotini/domoticz/commit/b8e10e711cbb3e69696bd69dd89201a0104c5b49 I solved removing completely the heartbeat (just for the color of the Gateway).

Btw all the sensor (door and motion) go ON randomly, sometimes just 1 or 2, sometimes all togheter. I don't know about the switch because it came to me when I already fixed this problem of fake positive.

deennoo:
ok thanks ! 

maybe you can help to another problem ?

PR #1427, opened 2017-03-31
Theming-friendly modifications to dashboard

flatsiedatsie:
-Use CSS classes to display status instead of direct JS manipulation. This allows themes more freedom to use colours.
- Placed Section-wrapper around each category. This will allow for creative new dashboard constructions through CSS.
- Replaced Section-wrapper around items (after checking that nothing was dependent on them), and replaced them with div+class.
- Places span tags around #bigtext content, to enable better theming.
- added classes to places where currently only ID's exists (and where the script uses ID's as if they were classes).  #ID's should be unique.
- Fixed a 'bug' where there was a Span tag wrapped around just the utility tables.

All changes are backwards compatible (checked).
BR tags were not removed.

### DIFF
--- a/www/app/DashboardController.js
+++ b/www/app/DashboardController.js
@@ -73,7 +73,7 @@ define(['app'], function (app) {
 
 		EvohomePopupMenu = function(item,strclass)
 		{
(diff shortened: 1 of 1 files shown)


flatsiedatsie:
Wait, I found a small bug.

PR #1428, opened 2017-04-01
Theming-friendly modifications to dashboard

flatsiedatsie:
- Use CSS classes to display status instead of direct JS manipulation. This allows themes more freedom to use colours.
- Placed Section-wrapper around each category. This will allow for creative new dashboard constructions through CSS.
- Replaced Section-wrapper around items (after checking that nothing was dependent on them), and replaced them with div+class.
- Places span tags around #bigtext content, to enable better theming.
- Added CSS classes.

### DIFF
--- a/www/app/DashboardController.js
+++ b/www/app/DashboardController.js
@@ -73,7 +73,7 @@ define(['app'], function (app) {
 
 		EvohomePopupMenu = function(item,strclass)
 		{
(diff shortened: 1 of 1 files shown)


gizmocuz:
Could you apply this to the Gui branch ?

flatsiedatsie:
Yes, just saw your email :-)

PR #1429, opened 2017-04-01
Timer corrections (retry)

gordonb3:
1. Next firing time calculation is wrong for sunset/sunrise timers because it takes today's sun values to set tomorrow's scheduled time. https://github.com/domoticz/domoticz/pull/1429/commits/f7e131409918a0d647955a0b67d226cfcf872da7 fixes this by reloading the schedules when sunset/sunrise values change

2. Next firing time calculation contains a default routine that adds 24 hours to the calculated time, either enforced by the caller or because the time references the past. In time zones that have DST this is wrong twice a year. https://github.com/domoticz/domoticz/pull/1429/commits/62e37ba3b0c55e672f560d0284028cf3db1d8b5b fixes this by using a sane time calculation.


gizmocuz:
Thanks!

PR #1430, opened 2017-04-01
Fix option handling for custom sensors

oohlaf:
Custom sensors don't use base64 encoded options. Keep them as single string.
In Python the option argument remains a dictionary, but when subtype is
sTypeCustom the value string from dictionay key "Custom" is used as option
string.

### DIFF
--- a/hardware/plugins/DelayedLink.h
+++ b/hardware/plugins/DelayedLink.h
@@ -67,6 +67,7 @@ namespace Plugins {
 		DECLARE_PYTHON_SYMBOL(PyObject*, PyDict_New, );
 		DECLARE_PYTHON_SYMBOL(void, PyDict_Clear, PyObject *);
 		DECLARE_PYTHON_SYMBOL(Py_ssize_t, PyDict_Size, PyObject*);
+		DECLARE_PYTHON_SYMBOL(PyObject *, PyDict_GetItemString, PyObject* COMMA const char*);
 		DECLARE_PYTHON_SYMBOL(int, PyDict_SetItemString, PyObject* COMMA const char* COMMA PyObject*);
 		DECLARE_PYTHON_SYMBOL(int, PyDict_SetItem, PyObject* COMMA PyObject* COMMA PyObject*);
 		DECLARE_PYTHON_SYMBOL(int, PyDict_DelItem, PyObject* COMMA PyObject*);
@@ -152,6 +153,7 @@ namespace Plugins {
 					RESOLVE_PYTHON_SYMBOL(PyDict_New);
 					RESOLVE_PYTHON_SYMBOL(PyDict_Clear);
 					RESOLVE_PYTHON_SYMBOL(PyDict_Size);
+					RESOLVE_PYTHON_SYMBOL(PyDict_GetItemString);
 					RESOLVE_PYTHON_SYMBOL(PyDict_SetItemString);
 					RESOLVE_PYTHON_SYMBOL(PyDict_SetItem);
 					RESOLVE_PYTHON_SYMBOL(PyDict_DelItem);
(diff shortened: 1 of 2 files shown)


oohlaf:
Based on feedback in https://www.domoticz.com/forum/viewtopic.php?f=6&t=16791

I must say I don't really like this workaround, but couldn't think of anything better. Feedback is welcome.

oohlaf:
A different approach might be to allow argument option to be either a Dictionary or an array List.

When Dictionary use the current SetDeviceOptions handling with base64 encoding.
When List concatenate the items into a string separated by ";"

Such overloading is not very consistent, could also introduce a argument named CustomOptions that is the above mentioned List. Any comments?

gizmocuz:
I still vote for a function where you can specify what type of sensor you want, with some parameters, and it will create this
A small work around for this problem, a function to create a custom sensor (with it's parameters)

For example this:

        Domoticz.Device(Name="Usage", Unit=1, Type=243, Subtype=29, Switchtype=0, Image=0, Options="").Create()

is way too complex

Unit/Subtype/Type/switchtype/image/options... hmmmmm

AND we also want 1 type of sensor (we are even going to rebuild domoticz to generic sensors only)

How about

CreateTempSensor("Name")
CreateOnOffSwitch("Name", ondelay,offdelay) or CreateSwitch(Name,type)
CreateLuxSensor
CreatePecentageSensor
CreateKwhSensor("Name")
Create.....

all functions could/should return the database Idx value, or 0/-1 in case of a problem

It might seem a lot of work to build the first time but
- we can force the correct type/subtype internally
- check the parameters
- know what we support

We could keep the current method of creating a device... But rather not...

How about this ?

oohlaf:
Right, that makes sense.
Indeed this needs some refactoring. In that case I would also like to rename the various methods to adhere to the Python naming conventions (using lowercase and underscores as opposed to CamelCase)

DebugBill:
I like the idea too.
If I were to create them, all sensors would be individual objects that could be manipulated after being created. This would allow for further customisation while remaining "standard". For example, when creating a "voltage sensor", I could in a second step and through a method add a "precision" value in order to set the number of decimals or unit (currently the voltage meter is 3 decimals, which does not make much sense when mesuring mains), also Power is always show as Watt, but I would like VA sometimes. Currently it needs creating a separate sensor, standard or not. It should just be enriching a base object.

oohlaf:
Made a first draft for what I would like as an interface in the wiki.
https://www.domoticz.com/wiki/Python_Plugin_Proposal

DebugBill:
Added my 2 cents worth

gizmocuz:
When measuring motherboard voltages/cpu voltages you needs those decimals (2 decimals)
There was a reason why 3 decimals where implemented (and used)... can't remember

DebugBill:
Sure, sometime you will even need 6 décimals, and some other times you'd rather have kV. Still you don't wan't to create two voltage sensors, but you'd rather have one meter, and a method on the object to tailor its precision to your needs for one specific device

gizmocuz:
I don't think kV would be appropriate for 'home automation' ;)
I dont see anything wrong with 3 decimals, it is like that some temperature sensors do not have digits, others does... well... they will always been seen as 18.0 now... but this does not seem to be a big issue
It will require a big code change to add a decimals field in the device table... but possible

DebugBill:
Some sensors return the voltage on mains lines which is about 240V. I have several of those, zwave electrical plugs and Ecodevices RT2. One gives 1 decimal, the other is integer. For a 10^3 range value, having 3 decimals looks strange in my opinion. In this case I would instruct the sensor to only show 1 or 0 decimals, still in Volts. Maybe of a motherboard I would prefer mV. (This is definitely usefull for Amps, where the needs and ranges can varry greatly)
volts

DebugBill:
We could add this feature in the database.... and start using it if the big sensor rewamping becomes real....
Maybe add a few "custom use" fields as Node1 to Node6 in hardware and let programmers use them as they see fit for the various sensors. This is where making the new sensors more "object-oriented" will bring its full benefit.
Another way is to have a separate "custom parameters" table (a bit like the preferences one) and let people feed it with their own specific items linked to the Device/Hardware Id. It is much more flexible but rather messy. I don't really like it but it reduces the amount of work to create it

gizmocuz:
We have already an extra field that is enough for all developers 
You can place a compressed base64 object oriented json structure in here

PR #1431, opened 2017-04-01
GPIO polling option (new), add-hardware fix

hvbommel:
[Change1]
New feature: GPIO polling has been added as a safety net to keep track of input states in case interrupts are missed. This could happen when misconfigured or in error cases. The polling interval is disabled by default by setting the poll interval to zero. This is the recommended setting for less critical environments. When poll interval is set to a integer value larger the zero then GPIO inputs are captured every poll interval. Domoticz is updated when state mismatches are detected. A typical setting is 60 seconds, faster polling intervals are not recommended to prevent unneeded overhead.

[Change2] 
Fix: Since de-bounce settings where introduced, at 22-march-2017, GPIO hardware could not be added anymore. This has been fixed. (existing configurations kept working)

### DIFF
--- a/hardware/Gpio.cpp
+++ b/hardware/Gpio.cpp
@@ -64,6 +64,7 @@ bool m_bIsInitGPIOPins=false;
 bool interruptHigh[MAX_GPIO+1]={ false };
 uint32_t m_debounce;
 uint32_t m_period;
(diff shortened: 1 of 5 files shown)


gizmocuz:
Thanks!

PR #1434, opened 2017-04-03
[BleBox] Fixing shutter closing.

gmaslowski:
With previous simplifying change there was a bug with closing, which actually was opening the shutter :). Now it's fixed. With this fix it's possible either to 'close' and to 'open 100%' to actually close the shutter.


gizmocuz:
Thanks!

PR #1436, opened 2017-04-04
Minimise and compress JS files on install. Windows contributors please help!

DebugBill:
Don't merge for now.

This is just a request for comments. Storing JS files as compressed archive is a problem. I've updated the makefile to be able to store the files as plain, indented files which are "minimised" and compressed before being pushed to the final directory.

It seems to work fine on my machines (Linux, Pi) except for a few files which have syntax errors (a good opportunity to fix some bugs maybe....).

It needs to have yui-compressor installed (package is available for Debian).

I fear it will not work out of the box for Windows.

I am no specialist of CMake, and I don't have access to a windows build platform so I would need some help to finish this.


gizmocuz:
just some remarks
Files could just end with .gz, no need for .min.gz
maybe you can use the default gzip to compress ? no need to install a package
7zip is installed for the windows compiler (if not, we can use this to compress to gz)

DebugBill:
There are two steps in this proposal:
1) minimise
2) compress

Minimise will remove all spaces, indentation, carriage returns, comments..... from the JS files to make them about 20% smaller. I know some tools even rename variables to shorter names, but did not check if this one does too. This will improve performance in the browser. This is performed by yui-compressor (which does not compress at all despite its name). This is how I discovered some syntax errors.

Then as a second step the minimised file is compressed using standard gzip.

The files in *.min are minimised, *.min.gz are minimised-compressed. On install, the later is renamed to js.gz. Keeping all allows for the incremental make process to work as usual (only changed files are reprocessed.

DebugBill:
One more detail, as this process could make debugging of JS more difficult, building with flag "Debug" will disable minimisation and compression.

DebugBill:
Looks OK now on both Linux and Pi with any combination of gzip and yui-compressor being available or not. 
Should compile fine on Git as it adapts to binaries not being installed, still it would be better from a performance point of view to have at least gzip available, and better, yui too. As yui detects some syntax errors in JavaScript, it is not a bad idea to have it too.

I had to make small modifications in 4 JS files to correct mistakes. I did not see any effect but this would be worth reviewing.

Last point app.js is still compressed as it can't be minimised yet. Some of its variables are used within other files and the renaming that occurs during minimisation breaks this. I need to add hints within the file to avoid renaming and test it all. All other JS files can be stored as plain text and will be compresse on install. This should greatly ease collaboration and reviewing of JS files. 

Hope we can merge this soon. I fear it might create a few problems at first though, but I think it is worth the hassle

gizmocuz:
At the moment we are making some big gui changes. Is it possible to wait till this is merged in the master?
Maybe start simple
- all js files are uncompressed by default
- in cmake/make/package script, gzip (.gz) all js files for in the final www folder in the archive

- this way we know the code should still work
- no need for a debug mode, the cmake package option will gzip the files, and/or the visual studio post compile script will do this

DebugBill:
No problem to wait. Anyway, before merging, the JS files must be updated again from their GZ sources as GIT cannot detect the changes between the zipped files and their raw counterpart in my pull request

If you don't install yui-compressor, the files will just be compressed but not minifyed, which is the first step you want to take. The code will work as before, with just the added value of being stored uncompressed in GIT. I still need someone to help for Windows build as I could see in the logs that no compression is applied there.

The debug mode is there to bypass the minify step even if yui-compressor is installed. This would be useful if someone wants to run a debuger on the client side as debuging on a minifyed file is quite horrible (all variables lose their original names) . Maybe I can implement it in a different way? Currently it is enabled when invoking cmake in the form "cmake -**DCMAKE_BUILD_TYPE=Debug** CMakeLists.txt"

DebugBill:
This is a much simpler version of the pull request with very little impact. JS files are left untouched in the repository. I use uglifyjs instead of yui-compressor (debian package node-uglify). It is more current and modern but is basically doing the same job, just better.

For the moment running the new CMakeList.txt will ignore js.gz files and only process *.js. All domoticz created JS files should be uncompressed in the future, while library provided ones must remain zipped (anyway, most of them are already minified). This can be done progressively. Only spaces, newlines and obvious optimisations are applied so there should be no impact on execution, except maybe increased speed and download times.

Hope you like it better. You need to install the debian package to get the full benefit but it also works without it (you will get a warning though)

gizmocuz:
Maybe minify could be a second pull request as this could cause issues ? i know it saves some bytes

gizmocuz:
Could you (more importantly) patch the makebeta/makerelease scripts, so only the compressed files are put in the archive ?

DebugBill:
Sure, I'll do.
For the pull request, I will simply comment out the line "find_program(UGLIFY uglisfyjs)", thus simulating a machine lacking the package. Not finding the binary will prevent minification to occur, but enabling it again for testing will be simpler.

DebugBill:
makebeta and makerelease are patched. uglify is commented out in CMakeFiles.txt so Javascript files will be zipped but not minimised.
This should then have no impact at all. 

When this is merged, I will drop a pull request where some of the currently zipped files are unzipped. This will make it easier when you push the new GUI files and we can see the differences in clear text.

DebugBill:
Do you want me to do the same on CSS files? I've compressed mine and it works fine with my browser (Firefox)

makebeta
```diff
@@ -49,7 +49,23 @@ echo "Success, making beta...";
...
+# Cleaning JavaScript files in directory WWW #
+##############################################
+
+# Remove all source .js files as they should now be compressed
+find www -name *.js -exec rm '{}' \;
+
+# Remove all intermediate files ending in .min
+find www -name *.min -exec rm '{}' \;
+
+# Rename files from *.js.min.gz to *.js.gz before inclusion to archive
```
gizmocuz, line 62: This part i do not fully understand.
We have a lot of .min files... 
how would this work with/in the file "html5.appcache" ? This is of most important...
If a file is not found, we have caching issues

Also i would do a minify in step two...

For instance, when someone want to work on a javascript file (lets say, domoticz.js)
now he/she can unpack it, delete the compressed file, clear the browser/appcache, and start developing
If you minimized the file... that could be difficult
On the other hand... he/she could checkout the beta and build himself, but that could be more complicated
But the html5.appcache ... this should match the files in the final output...

xztraz:
This is a welcome change! Big up!

makebeta
gizmocuz: Sorry, this review was on hold... just noticed

DebugBill: Several questions  here.... :-)

There are lots of files **containing** .min in the WWW directory, but none **ending** with ".min". The files I remove are just intermediate files created during the make process out of the original JS files before they are compressed. They are never used in this format and should never appear in this form in the installation directory. The only files that need to be found in the installation directory are the "*.js.gz" ones. We must also make sure that there are never both the "*.js" and the corresponding "*.js.gz" in the destination directory at the same time or caching will be an issue if one only is modified.

minimfy is disabled in this version as you requested, but the code is still there. One comment to remove and it works again so we can have a few people testing it (I have already found a few issues and their fix)

If someone downloads the pre-packed binary and wants to work on the js files then you are right, he will ultimately decompress a minimifed file and all variables will have meaningless names (this feature can be disabled while still removing useless space, CR and others). But do contributors work on pre-packed binaries? I would think they download the full source tree and install from there. In that case it is different. They can work on the original JS file which is still in the directory. When they "make install", the JS file will be minified and compresed again before being pushed to the installation directory. If they want to debug with the real variable names, they just need to run cmake as "cmake -DCMAKE_BUILD_TYPE=**Debug**" instead of the more usual "cmake -DCMAKE_BUILD_TYPE=**Release**".
When using git, they should of course never commit one of the modified files (*.min or *.js.gz) but only on the original *.js file.
You can consider the js file as being equivalent to a cpp file, the minified one to a .o file, and the zipped one to a binary file.

html5.appcache is not a problem. You only put there files ending in .js. When  trying to get the file, the browser looks for both the compressed and uncompressed versions. My code does not change the name of the file, but only adds a .gz extension (the *.min file is a by-product, like a *.o file for gcc), so the file is still found whatever version is in the www directory, compressed or not.

The same can be done with css files.

gizmocuz: A HTML designer is good with HTML, not with c++, does not even know how to install visual studio, or gcc, let alone compile or knows the Linux or Windows operating system.
That is why minification is not really usefull.  Specially not for the domoticz related stuff, sure for third party libraries, but there are for 90% already minimized

But great stuff !, Definitely going to look into this before a new stable release !

DebugBill:
No problem, when you're ready to enable minification, I will just add a flag so it does not alter the variables names. It will reduce slightly the optimisation but still improve the overall speed. 
Then when debuging (at least with Firefox), you usually have an option to reformat the optimised script so it becomes readable again.

In the screenshots below, before clicking on the button and viewing a minified script, and then, after clicking, seing the same script readable again.

And you still have the option to do the full optimisation on stable releases packages only

before
after

gizmocuz:
Yep, but we (you) are talking like developers... those few enters or extra characters is nothing compared to a few icons/images, and not worth all the hassle i suppose, and not for the domoticz files
The library files, we can just see if there are some libraries i missed that are also available in 'min' version and then leave the domoticz stuff like it is

DebugBill:
In fact I think we are debating on a non existent problem :-)
My script (and only if I reenable uglify) will look for it on the build platform and currently would not find it..... so the JS files will be untouched and simply compressed. People downloading the package will then have "plain" files to work on. 
Only more advanced developpers who are building on their own platform could have the files minified.... but only if they have the package uglify installed. This package is not "default install" on any distribution I know of, so only people who know what they do and have willingly installed it would have their files minified.

In short, with the current pull request, you would only get the added value of leting git users work on plain JS files and share their modifications in diff format rather than binary compressed files where colaboration is difficult.

gizmocuz:
Great!, yes i think we are talking about the same ;)

gizmocuz:
Looks like the only thing left is the windows compile/build

DebugBill:
Absolutely. As I don't have access to a windows build system I can't do the same work on the windows side. This is no big deal, just take all js files and zip them to js.gz before storing them in the distribution archive making sure the uncompressed version still sits in the development directory.

gizmocuz:
In our repo/msbuild/windows installer folder is 7z.exe
You do have access to a windows build system if you want (virtual box for example or a normal real installation, or even better AppVeyor ?)

DebugBill:
I'm just starting with visual studio. I've at least understood one thing. This pull request and its windows counterpart are two independant projects. I think nothing prevents you from merging this one  I will submit another one for windows as soon as I'm up enough on the learning curve.
The short term consequence will be a performance improvement for Linux users when refreshing the cache thanks to added compression, but no change for windows installs.
If we start to uncompress a few JS.GZ files, it will degraded slightly refresh performance for windows users only. But I will try to provide the equivalent patch before that.

gizmocuz:
For windows user can build in two ways:
- CMakeList.txt (but i do not think anyone uses this... ok, i know a few)
- Visual Studio

You can also make the patch in the InnoSetup script for windows

CMakeLists.txt
```diff
@@ -152,6 +152,103 @@ IF (INCLUDE_SPI)
...
+file(GLOB_RECURSE js_in_files RELATIVE "${CMAKE_CURRENT_SOURCE_DIR}/www/" *.js)
+
+option(MINIFY_JS "Minify JavaScript files" YES)
+unset(UGLY CACHE)
+unset(GZIP CACHE)
+unset (PARAM CACHE)
+#find_program(UGLY uglifyjs)
+find_program(GZIP gzip)
+
+IF (${CMAKE_BUILD_TYPE} STREQUAL "Debug")
```
gizmocuz: I don't think we can mangle variables at all. Specially when we are using more AngularJS code.
Variables from other JS files are used in another JS file... this can go terribly wrong.
I think we should remove the whole renaming of variables
It is also not needed, it's an opensource project, and we do not need to protect the code

CMakeLists.txt
```diff
@@ -363,6 +460,7 @@ hardware/PhilipsHue/PhilipsHueSensors.cpp
 hardware/PiFace.cpp
 hardware/Pinger.cpp
 hardware/PVOutput_Input.cpp
+hardware/RAVEn.cpp
```
gizmocuz, line 464: I dont think this should be in the pull request, as also the below (xiamigateway)
Maybe you have to bring your forked repository a little up to date ?

DebugBill:
It is up to date. It's just because I am a bit psycho/maniac sometimes and I was upset by the few files not in alphabetical order when building the hardware directory.... :-). I just reordered a few of them that had been added last in the list.

CMakeLists.txt
DebugBill: Angular is already minified and the necessary "extern" variables are protected. 
In domoticz I noticed app.js needed to be protected too and I started that work. You just need to add "hints" inside the code to tell uglify which variables need to stay untouched.

Anyway, I removed the mangle option in this new release, so only optimisation/compression of the code will happen, and ONLY IF uglify is installed. For developpers, reverting it when debugging is just a click in their favorite IDE/Browser.

gizmocuz:
Not of interest anymore? Currently I am away and am not able to debug this

DebugBill:
Several reasons for closing it, the third being the most valid one I guess
1) It's been there a month already
2) I never managed to get a fully functionnal windows build environment because definitely I hate window's useless complexity that looks to me more like obfuscation than functional richness
3) I read the code for cWebem and discovered it compresses files before sending them anyway, so no need to store them compressed in the first place unless compression time and CPU is critical when serving the JS files (which I don't feel it is, even on a Pi2). Only thing to do is to decompress all Domoticz specific JS files in the repository so collaboration on GitHub is easier

gizmocuz:
Project started on a raspberry pi 1 where compression was cpu intensive, specially when reading all the cache files
After I compressed most of them, everything was much faster 
My other idea was to calculate a hash for the files, and when changed (Recommended) compress them
It's the same as with python files, they are compressed/compiled when first used or when changed
After this implementation we could just have them de compressed in the repo... Time is not  always on my side

DebugBill:
I looked into your idea of optimising cache refresh, but it seems to be incompatible with using app.cache, at least on Firefox. When refreshing it refrehes it all.
My suggestion would be to use standard caching mecanism, and use offline cache only for the "offline" page. There is no need to have the full Domoticz code in offline cache as in the end it only works online.

gizmocuz:
Yes, there is a big need for this.
Do you own an ipad/iphone ? If not, then you do not need this.
If yes, with safari, open the domoticz gui, next add a shortcut to the home screen,
You now have an offline application, no need to refresh any page (but more importantly images)
So now (as i often do), when you make a VPN connection to your home from a far away location with a very very bad internet connection, you only communicate via json calls
I dont know of any other method to make an offline application for iOS then using the appcache, and that is also why it is designed (not for ios, but not to download unnecessary items)

DebugBill:
I tested with my iPhone and reverse-proxy logs... got your point :-) It definitely minimises the data to transfer. Not using appcache but relying on "if-modified-since" requests and HTTP 304 replies would add significant amounts of data. Lets investigate the hash way to optimise cache refreshes

gizmocuz:
Maybe we should make this an issue instead of a pull request or discuss on slack.
Do you know how python does this ? It creates a binary/compiled file, and when a source file is changed, it knows it and compiles again
Maybe the first time a .js file is requested after startup we could build some logic.
It is cached at the user for the next request, so it is only when the appcache file is changing.
But maybe there is a better way.

dwmw2:
I've been looking at this a bit, and testing the web server's behaviour.

The only thing that the web browser ever sees is the `foo.js` files. The web browser may or may not send an `Accept-Encoding: gzip` header with its request.

If both `foo.js` and `foo.js.gz` files exist, then the web server will send the most appropriate one. If only one exists, then it will compress or decompress as required.

So... simply decompressing the files in the git repository, right now, would "work"... for some version of "work" which would suck on RPi, as @gizmocuz points out, because it'd end up compressing them all at runtime again. 

But it's actually relatively simple to add CMakeFiles rules which do the compression at build time, and for 'make install' to install the compressed versions instead of the originals. Even if we only have that for the CMake builds and not for Windows, that's may actually be OK. After all, if people are running on Windows then they probably *can* afford the CPU time it takes to compress dynamically. It'd be nice to fix that too, but perhaps it isn't critical enough to prevent the changes from being merged?

Once we have build rules to generate the `foo.js.gz` files, we can tweak those rules later so that they minimise-and-compress instead of just compressing. And again the Windows version doesn't really need to be in lockstep with the CMake version.

I'm not sure if it makes sense to cache the runtime-generated compressed versions. In a lot of the resource-constrained environments you really want the compressed version to be the only one that ever gets stored, which means you really want it compressed at build time.

emontnemery:
@DebugBill Are there any open issues with your patch? Do you need some help debugging or testing?

DebugBill:
Hi,

This patch is over a year old. I used to be ready but does not even compile anymore. I closed it long ago due to lack of interest, but Gizmocuz chose to reopen it. I don't intend to work on it anymore.

gizmocuz:
The reason why i reopened it, is because i still think this is great option, but the implementation i want to do differently when i got some time for this

xbeaudouin:
@gizmocuz what in the implementation has to be redone?  Can we for a start only doing gzip of .js files ?

gizmocuz:
Currently there are way to many files in this patch.
Maybe only the relevant files should be in here (not the uncompressed files... i will do those by hand later)
Next to the patch in the cmake file, also the makebeta/makerelease scripts should be patched, maybe that is already done, but there are 107 ? files in this PR ;)

gizmocuz:
Closing again... it's made against the master branch and there are to many files in here.
Think the only changed files should be the CMakeList.txt and the makebeta/makerelease scripts
If someone would like to re-create it, please do so.
We do not need uglify, only the gzip part. under windows this can be done by the 7zip compressor which is already available in the installer

PR #1440, opened 2017-04-05
Fix for issue 1441 and other things

DebugBill:
Previous title was: New model in the EcoDevices family and other improvements

Quite a few things in this release:
- A new EcoDevices RT2 model has been available for a few weeks. The API is very different from the previous model. Basic support (Teleinfo and 8 counters, but not the EnOcean devices attached) is included in this release through a new combobox "model" in the hardware definition. Please add the word "model" to the translations files
- Because of the error message added in the CRC calculation, some users have discovered their serial lines were not very reliable. Added a "CRC disabled" check box to Teleinfo serial just like it is in P1SmartMeter
- Added a second CRC calculation algorithm as documentations states that newer meters will use this new method (could not test it though as I don't have access to one of those meters yet). Switch is automatic.
- Added a serial port speed combobox for TeleinfoSerial as those new meters may also speak faster (9600 baud instead of 1200)
- Also inspired from P1SmartMeter is the "rate limit" value. Some users requested almost real time while I know my database had Pi harware can not handle that load. Now this is user settable.
- Both modifications above have their default values set through a database update on update (DB_VERSION 114)
- Teleinfo EDF USB renamed to Teleinfo EDF as some devices are serial but not USB. The term could be confusing.
- Added a 5 seconds grace period for all hardware for inactivity watchdog. I had issues whith some harware having an uptade rate of exactly 5mn just like the watchdog. Sometimes the watchdog would trigger first and reset the harware for no reason. Also included in database version 114 update.
- Fixed width of "when" field in Notifications to handle the new, larger conditions.

Support for the new RT2 model has been tested with the manufacturer and is known to work.


main/SQLHelper.cpp
```diff
@@ -2218,6 +2218,18 @@ bool CSQLHelper::OpenDatabase()
...
 		}
+                if (dbversion < 114)
+                {
+                        //Set default values for new parameters in EcoDevices and Teleinfo EDF
+                        std::stringstream szQuery1, szQuery2, szQuery3;
+			szQuery1 << "UPDATE Hardware SET Mode1 = 0, Mode2 = 60 WHERE Type =" << HTYPE_ECODEVICES ;
+                        query(szQuery1.str());
+                        szQuery2 << "UPDATE Hardware SET Mode1 = 0, Mode2 = 0, Mode3 = 60 WHERE Type =" << HTYPE_TeleinfoMeter ;
+                        query(szQuery2.str());
+			szQuery3 << "UPDATE Hardware SET DataTimeout = DataTimeout + 5 WHERE DataTimeout >0";
```
gizmocuz: This should not be necessary, 0=disabled, no data timeout

DebugBill: I'm not sure I understand your comment. The "where  >0" clause is there to leave zeroes untouched and watchdogs disabled. Without it the update would replace 0 with a 5 sec timeout, thus enabling the watchdog. All others are increased by 5 sec in order to match the value set by the GUI

DebugBill:
I reverted the change you were questionning. It appears the problem I was trying to solve, although theoretically possible, is not the one I am facing (see issue 1441 for more details)

gizmocuz:
Thanks!

pchevali:
Hi,

I commented on issue 1441 yesterday, but maybe I should reopen an issue.
Following the changes in the teleinfo hardware I have noticed weird behaviour.
-It appears that changing the watchdog value changes the way the plot is diplayed, long periods of inchanging power consumption leads to "0" value in the plot.
-Moreover as there is a check line 147 : 
`if ((teleinfo.pAlertPAPP != teleinfo.PAPP) || (difftime(atime, teleinfo.last) >= (m_iDataTimeout -10)))`
The value is inserted only if there is a change in the value (PAPP). Thus there can be value spaced irregularly in time on the plot, but the plot is using the variation of the power consumption index as the source of the values. This results in spikes and gaps in the graph.

What do you think of that ?

Thanks

gizmocuz:
If you have this corrected/fixed please create a new pull request 
Thanks in advance

DebugBill:
I'm working on it, but this is only a side effect of setting P1 to "no decimals" and the watchdog to more than 5mn. It is not a really bug.
What I'm doing is just forcing the conf to a consistent combination of parameters

DebugBill:
Workaround in pull request 1511, along with fix of a recently appeared bug (due to GUI updates I think) preventing EcoDevices to be created

gizmocuz:
Merged, thanks!

PR #1442, opened 2017-04-05
Notification add Recovery message

jvandenbroek:
Added recovery notification, using DB CustomMessage column to store it's value. Wrote a split function, so if CustomMessage is already used, this will be appended / removed when needed. Delimiter  `;;` for this field is arbitrary chosen, so ppl can still use a single `;` in the Custom Message field. Could change that to something else if wanted.

Changed recent additions to store `>= <= != and =` in DB instead of the letters M L (which is much clearer when handling DB values).


gizmocuz:
Thanks! Is it possible for next time to not do reformatting ?

jvandenbroek:
OK won't do that again, but I did it because the previous commit mixed space and tab indents, which  I really don't like (and the indents were all over the place).. So my intention was good I guess ;)

gizmocuz:
Yep it was!

notifications/NotificationHelper.cpp
```diff
@@ -1111,6 +1285,8 @@ void CNotificationHelper::ReloadNotifications()
 		}
 		std::string ttype = Notification_Type_Desc(NTYPE_LASTUPDATE, 1);
 		StringSplit(notification.Params, ";", splitresults);
+		if (splitresults.size() < 3)
```
gizmocuz, line 1288: Incorrect, will fix this in a second (try debugging first and set a simple notification on a switch)
https://github.com/domoticz/domoticz/commit/064f5c337be3a5f60422e61e0d93b6c3009ee894

gizmocuz:
Could you check maybe this issue:
https://github.com/domoticz/domoticz/issues/1447

jvandenbroek:
Sorry, that obviously shouldn't be there. Was some leftover from some 
reverted changes.. Thought no code was changed related to switches, so 
didn't test those. My bad.

PR #1449, opened 2017-04-08
Added Notification support to Plugins

dnpwwo:
Plugins can add themselves as 'Notifiers' and the specified name is added to the list of Notification targets in the Web UI 'Notifications' page.  Plugin framework calls onNotification when a notifiable event occurs.


gizmocuz:
Thanks!

PR #1453, opened 2017-04-09
Make plugin disconnect more graceful

dnpwwo:
Add boost shutdown to socket prior to close.  Should resolve Pi crash on plugin restart.


gizmocuz:
Thanks!

Issue #1276, opened 2017-02-28
Switch delayed time timer is 18.5% slow


RicardP:
Hi, 

I am running on Windows10 platform "run as a service". Domoticz version is beta 3.6788.

When using timers for delays I have issues with accuracy, or more exact no accuracy at all... :) I experience a second in Domoticz to be 18.5 percent longer than reality. So on all timers created compared to target delay set I have to manually cut off the timer "seconds" by 15.61 percent to get expected result.

gizmocuz:
How long does one minute take 
How long does 5 minutes take?

RicardP:
I made a switch 9 hour delayed off = 32400 sec. Checking the log later on shows it to actually switch of 10 hours and 40 minutes and 20 seconds later. Same +/- 10 seconds logged 4 days in a row.

So a minute in Domoticz is 71.1 seconds in life... translated in 5 minutes, times five: 355.5 seconds :)

gizmocuz:
looks like a counter does not get incremented correctly
we made some recent changes (some months ago)

RicardP:
Yes I did comment this before some months ago, I don't know/remember if you made any changes but I never seen any improvements. At that time I was on a Raspberry and now Win10 but the issue still follows me :-)

gizmocuz:
we created a new variable 'timer_resolution_hz' that is being used in SQLHelper.cpp in the function Do_Work and handles for instance the Off/On Delays
there is a variable (float) that has the offdelay duration, and we decrease this at each timer resolution (timer_resolution_hz) (which is 25, so 1000/25 = 40 times a second)
Maybe you could have a look at this, or contact the original author

RicardP:
I am not into coding, sorry. However would it not be a better to calculate and set the delay actual clock date/time the requested delay shall be executed at, instead of having a timer counter process running until it's counted out?

RicardP:
After updated to the last beta including this fix I removed my drift correction to a daily timer on the towel heater. It is meant to be on for 9 hours after it's triggered at 06:00 every day. today checking the log it was shut of exactly 15:00 = as my 32400 seconds said it should :) Thanks a lot!!

jvandenbroek:
I think this one can be closed.

PR #1454, opened 2017-04-10
Fix for delayed switch drifting (issue #1276)

jvandenbroek:
This fixes switch on / off delay drifting issue: #1276


gizmocuz:
Thanks!

jvandenbroek:
Unfortunately there is a problem when using lua and dummy switches, I'll look into it tomorrow. Probably best to revert the commit until I've fixed this.

gizmocuz:
Okey, i have reverted the patch

PR #1456, opened 2017-04-10
Small hack for those managing several Domoticz instances

DebugBill:
Ever got confused in your browser between several Domoticz instances because all tabs titles are set to "Domoticz"?

This pull request adds one parameter "Title" in the global settings. This setting will then be used as the text in the tab so you can name your instances. Default value is of course Domoticz.

@gizmocuz : The label in the settings page needs the word "Title" to be added to the translation files before appearing. I also miss the word "Model" from my previous merge.


gizmocuz:
Please post some screenshot first
If you use the right <span data-i18n="Title">Title</span> tags, then you have default the english translation, regardless if the string is translated
Please provide the complete lines that needs to be added to the english translation file in your pull request, i will add them

gizmocuz:
Maybe it is better to put this under the 'Location' setting ?

DebugBill:
Here is a snapshot under "location". The problem in my opinion is that il leaves a large unused space under "User Interface". 
title

DebugBill:
I know that if no translation is provided then the English default value is shown.... but this is not working here, and after quite a while I couldn't figure out why. I've implemented my label in the exact same way it was already done.

I therefore need "Title" and "Model" to be added to the translation files.

gizmocuz:
Screenshot is not working... are you posting via github ?

If you declare your labels like this:

`<span data-i18n="Outside Temperature Sensor">Outside Temperature Sensor</span>:`

then it works without the translation being added

Please post the lines needed to be added into the English translation file

DebugBill:
The screenshot is updated. Don't know why it did not go through the first time. Too big maybe.

Here are the lines:
    "Title": "Title",
    "Model": "Model",

DebugBill:
I've done what you suggested for the label and indeed it works. I wonder why the other labels are not done that way. They would not work without a translation file.

gizmocuz:
Thanks!, Could you rename 'Title' to 'Name' ? This way it also does not need to be translated

DebugBill:
Changed "Title" to "Name" on display, but kept the variables names unchanged as we are really dealing with the "Title" tag here.
Still need the translation tag for "Model" though

gizmocuz:
Thanks for changing it to an instance 'Name' ;)

gizmocuz:
(Model translation is already done)

gerard33:
Thanks @DebugBill, this is a really helpful feature.

AndyWhittaker:
Just what I've always wished for @DebugBill! Excellent - thank-you

DebugBill:
You're welcome :-)

PR #1458, opened 2017-04-11
Fix for delayed switch drifting (issue #1276) (fixed #1454)

jvandenbroek:
There was an iterator count problem in previous commit, fixed it and optimized some handling. Tested with all kind of scripts, everyone works fine now!


gizmocuz:
Thanks! And also thank you for the quick fix!

PR #1459, opened 2017-04-11
Hue updates

scottydj2:
Add a poll interval box to the Hue hardware options using Mode1 in database for storage.


scottydj2:
Oops... not got the branch I wanted... should only have been 5 changes.

PR #1461, opened 2017-04-11
Partial undo of commit #f7e1314

gordonb3:
Undo the reload of schedules when sun times change - it may cause a race condition on schedules that are set around the same time this change occurs. Need to find a better way.


gizmocuz:
Thanks!

PR #1462, opened 2017-04-12
RelayNet: Theoretical fix sizeof(tRBUF)

hvbommel:
Proactive fix for a theoretical problem where sizeof(RBUF) was used instead of sizeof(tRBUF). This may lead to a crash in case the size of RBUF is larger then the size of tRBUF. The problem has never been reported or seen to my knowledge.

### DIFF
--- a/hardware/RelayNet.cpp
+++ b/hardware/RelayNet.cpp
@@ -311,7 +311,7 @@ void RelayNet::Init()
 	BYTE id4 = m_HwdID & 0xFF;
 
 	/* 	Prepare packet for LIGHTING2 relay status packet  */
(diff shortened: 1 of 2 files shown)


gizmocuz:
Should be the same as sizeof(RBUF) ?

gizmocuz:
maybe rename all class member variables and put m_ infront of it

hvbommel:
Hi, I looked at it a bit better and you are right, the sizes of tRBUF and RBUF are the same, so in the end there was not "theoretical" problem.. I did put the m_ in front of Packet. Thanks.

PR #1471, opened 2017-04-14
Generic sysfs gpio, new

hvbommel:
Generic sysfs gpio for Linux makes exported GPIO inputs and outputs available for use. Before adding the hardware within domoticz you need to export the GPIO's that you are going to use. By doing this you specify the direction of the GPIO, in or out, whether pull-ups or pull-downs are turned on etc. Then when this "Generic sysfs gpio" hardware is added it will detect the exported GPIOs and creates devices for it. After that you can fine tune the inputs and outputs for specific purposes like door-contacts etc. as always. Sysfs does not depend on any external libraries like wiringpi.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -382,6 +382,7 @@ hardware/serial/impl/unix.cpp
 hardware/SolarEdgeAPI.cpp
 hardware/SolarMaxTCP.cpp
 hardware/Sterbox.cpp
+hardware/SysfsGPIO.cpp
 hardware/TCPProxy/tcpproxy_server.cpp
 hardware/TE923.cpp
 hardware/TE923Tool.cpp
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -659,6 +659,14 @@ else()
   message(STATUS "==== (Please follow http://wiringpi.com/download-and-install/ if you want to use GPIO.)")
 ENDIF(WiringPi)
 
+IF(EXISTS /sys/class/gpio)
+	message(STATUS "Sysfs GPIO available")
+	add_definitions(-DWITH_SYSFS_GPIO)
+ELSE()
+	message(STATUS "Sysfs GPIO not available")
+ENDIF()
+
+
 find_path(TELLDUSCORE_INCLUDE NAMES telldus-core.h)
 if (TELLDUSCORE_INCLUDE)
   message(STATUS "Found telldus-core (telldus-core.h) at : ${TELLDUSCORE_INCLUDE}")
--- a/hardware/SysfsGPIO.cpp
+++ b/hardware/SysfsGPIO.cpp
@@ -138,23 +138,6 @@
 #define GPIO_MAX_PATH		64
 #define GPIO_PATH			"/sys/class/gpio/gpio"
 
-static int GPIORead(int pin, const char* param);
-static int GPIOReadFd(int fd);
-static int GPIOWrite(int pin, int value);
-
-struct gpio_info
-{
-	int	pin_number;		// GPIO Pin number
-	int	value;			// GPIO pin Value
-	int	db_state;		// Database Value
-	int	direction;		// GPIO IN or OUT
-	int	active_low;		// GPIO ActiveLow
-	int	edge;			// GPIO int Edge
-	int	read_value_fd;	// Fast read fd
-};
-
-static std::vector<gpio_info> GpioSavedState;
-
 CSysfsGPIO::CSysfsGPIO(const int ID)
 {
 	m_stoprequested = false;
--- a/hardware/SysfsGPIO.cpp
+++ b/hardware/SysfsGPIO.cpp
@@ -1,0 +1,639 @@
+/*
+	USAGE:
+	======
+	To use "Generic sysfs gpio" you need to export the inputs and outputs you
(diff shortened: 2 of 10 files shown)


gizmocuz:
Wow, Great !! Thanks!
Have to wait a little to merge because i'm having issues with getting the openzwave dev branch to work.
I think i know where it crashes but needs some time

gizmocuz:
Thanks!!! Hope some users also want to test this.. then we can remove the wiringpi stuff

LKapie:
I tested this new hardware (on Raspi with version 3.7365) and it works ok for me with one PIR sensor on GPIO.
Only issue I found was that old and new GPIO seem to have a minor problem to coexist: 
When I switch the old GPIO (Raspberry's GPIO port) to OFF and after that to ON again Domoticz freezes and a restart is needed. (did not try this before so I can also be an issue of the "Raspberry's GPIO port" hardware/driver) After restart the old and new hardware seem to function/coexist properly. It think this is not a likely use-case though so for me it's ok.

gizmocuz:
Well that makes sense as you probably can access a GPIO only once...
Great it works!
But when you disable the old GPIO and use the sysFS version, you can correctly switch On/Off the GPIO ?

LKapie:
Just checked it and switching On/Off "Generic sysfs gpio" works correctly at my system.

hvbommel:
Thanks for testing !

Can you also test the following:
1. Remove wiringpi (delete or rename the libraries, to _libwiringpi.a and _libwiringpi.so in /usr/lib and or /usr/local/lib) 2. Rebuild domoticz, now it will be build without “wiringpi” GPIO. 3. In hardware “Generic sysfs gpio” will still be present in the list but “Raspberry’s GPIO PORT” should be gone. 4. Then test “Generic sysfs gpio”.

Note:
Use raspi-gpio (instead of wiringpi gpio command) to configure inputs/outputs and pullups/downs etc. It will be the same way as you would have done, using the gpio commands, but slightly different syntax. If you don’t have it installed then install it using: 1. sudo apt-get install raspi-gpio 2. for help on usage type: “raspi-gpio help”

That should work. In that case your is system is no longer relying on wiringpi and its tools.

But also …
I also saw what you did see, problems during rundown of the “Raspberry’s GPIO PORT”. I don’t think it is failing because of a combination conflict (although a combination will never be supported, it either one or the other). I think Raspberry’s GPIO PORT” is failing on its own when we deselect “enabled” and then click on “update”. I did see it happen but did not debug it so far. I will debug it asap so that we exactly know whats going on and get back here with the results (and maybe with a fix for the old GPIO)

hvbommel:
Question: Should we do a wiki page for "Generic sysfs gpio" similar to the "Raspberry GPIO port" one ? If yes, who would do it normally ? Or is it too early ?

hvbommel:
I have entered a pull request to attempt to fix the "Rasberry's GPIO port" hang issue as described earlier by LenardKaptein in the blog. Note: The problem and fix are not related to the new "Generic sysfs gpio" hardware.

LKapie:
Dear @hvbommel I have a very busy week this week and will be abroad from tomorrow until sunday so I'm sorry to tell I cannot find the time soon to do the tests you suggested...

jvandenbroek:
I really like the idea of not using the wiringPi library and the code seems to be well written. But one thing I prefer about the wiringPi implementation, is that it spawns a thread per pin which uses the Linux poll function to efficiently wait for I/O to happen (this uses zero cpu time and reacts instantly). Maybe we can 'lend' some of it's code to make it independent? I think that's quite feasible.

hvbommel:
I have also looked into implementing a edge based version. The reason I implemented the poller: 1. Its still light weight 2. quiet responsive (250 msec) 3. No interrupt storms. 4. No thread & I/O cancelation complexity on rundown. 5. Robust & simple. But indeed I also did a edge based version with a thread per I/O pin. It works. But for a start I thought; lets keep it simple. And if the max response time of 250 msec happens to be a problem in some use cases then an interrupt based extension can easily be added later on (it more of the same). Question: Are you in your case having a problem with the 250 msec input response time ?  I think for the time being we should keep both "sysfs gpio" and "wiringpi gpio" in the system so the user has the choice. Thanks for your feedback.

jvandenbroek:
Well I agree that max +-250ms delay would be fine probably in most cases, but I'd suggest to make this configurable. It indeed has a negligible impact on performance, but running on a RPi 1 hosting multiple services, I'm always looking for ways to optimize things as much as possible. However I'd like to see how this runs with a configurable interval, so I can play with it's settings and monitor the real CPU time used. It might even be a better choice for me memory wise :) (not sure about how much overhead separate threads have, but sure there is some)

hvbommel:
I hear what you say but when we make it configurable people need guidelines to determine the best settings. My vote would be to leave it as is for some time, 3 months or so, and see how it works. It think simple is better, if it does the job. You can always tweek the (3 related) definitions in top of the .cpp file and build a version you are interested in. When we find out fixed values are no good that is a good reason to put in (documented) flexibility. I know it sounds a bit hard minded but believe it or not ... I like the feedback so that we can improve where there is room/need for it. Thanks.

gizmocuz:
I agree, keep it simple...

PR #1473, opened 2017-04-14
Style, element-dark, fixed slider for double icon

xztraz:
Before Fix (opera browser)
image

After fix (ie)
image


xztraz:
Just lift this in when other gui changes are made so cache number is updated.

xztraz:
Added fix for double icon slider on desktop view too.

Before fix
image


After fix
image

gizmocuz:
Thanks!

PR #1474, opened 2017-04-15
Added authenticated HTTP to EcoDevices

DebugBill:
Polling of EcoDevices can now use login/pass, enabling users to secure their devices.

Teleinfo devices are now smarter with min and max polling rates and DataTimeout (user can not configure a maximum polling rate that is incompatible (longer) with choosen datatimeout
login


gizmocuz:
Thanks!

PR #1483, opened 2017-04-17
Style, Element-Light, Fixed dim sliders and added sun indicator as in element-dark, Please test.

xztraz:
Copied dim slider fix from dark element style. Should be straight forward but this **needs to be tested**. I can't save settings, browser locks up so can't change skin to test in my installation


gizmocuz:
Thanks!

xztraz:
Allright. hope it works :)

PR #1486, opened 2017-04-18
Gpio wiringpi

hvbommel:
Fix possible hang when GPIO-wiringpi enabled checkbox is turned off, and the update button is pressed, 
on the domoticz hardware page.


hvbommel:
Will put a squashed version later.

PR #1487, opened 2017-04-18
GPIO-wiringpi: fix hang on disable hardware

hvbommel:
Fix possible hang when GPIO-wiringpi enabled checkbox is turned off, and the update button is pressed,
 on the domoticz hardware page. The fix was made based on test feedback from Lenard Kaptein (see  #1471).


gizmocuz:
Thanks!

PR #1489, opened 2017-04-19
Generic sysfs gpio: functional updates

hvbommel:
On startup check latest sysfs-gpio settings. In case they have changed update Domoticz devices accordingly. When for example an sysfs gpio input is changed into an output the Domoticz input device will be deleted and a new output device is created and vise versa. Also on startup, if an output device is found in the Domoticz database, the on/off state stored in the database is written to the related output pin. This way the Domoticz view and the actual output pins should always match. Also, when the system is used as a slave, all device states are updated about 30 seconds after device startup. This to hand a copy of the device states to the master after it has connected.

### DIFF
--- a/hardware/SysfsGPIO.cpp
+++ b/hardware/SysfsGPIO.cpp
@@ -123,7 +123,16 @@
 #include "../main/SQLHelper.h"
 #include "../main/localtime_r.h"
 
+/* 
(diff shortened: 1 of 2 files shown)


gizmocuz:
Thanks!

PR #1490, opened 2017-04-20
Allow linking to external JsonCPP library

xbeaudouin:
Hello,

On some OS there is already JsonCPP library in the system.
To avoid duplicate code in some embedded and/or localy patched jsoncpp library I have added an option : USE_BUILTIN_JSONCPP.

CMakeFile use pkg-config to find the correct flags where the library and include files are.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -40,6 +40,28 @@ set (domoticz_VERSION_MAJOR 3)
 set (domoticz_VERSION_MINOR 0)
 set (domoticz_VERSION_PATCH ${ProjectRevision})
 
+option(USE_BUILTIN_JSONCPP "Use builtin JsonCPP" YES)
+IF(USE_BUILTIN_JSONCPP)
+  MESSAGE(STATUS "Using builtin JsonCPP")
+  set (domoticz_jsoncpp_SRCS 
+       json/json_reader.cpp
+       json/json_value.cpp
+       json/json_writer.cpp
+      )
+ELSE(USE_BUILTIN_JSONCPP)
+  MESSAGE(STATUS "Using external JsonCPP")
+  find_package(PkgConfig)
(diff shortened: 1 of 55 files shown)


gizmocuz:
This is no problem as the library is very small, and we know it is working with this library
Thank you for the effort.

xbeaudouin:
I understand your point of vue, but if you look at the CMakeLists.txt I have modified, I let *per default* the included jsoncpp to be compiled per default.
On embedded system sharing the data is the best we can do.
As I previously did on other PR, I let the default embedded files to be used, and we have to specify that we use external libraries uppon compilaton.

Please Publish this PR since it can help FreeBSD user to chose what they want : duplicate code or not.

Regards, 
PS: I am on freenode #domoticz (nick kiwi)

gizmocuz:
Will you please make an example c++ program that generates and decodes json and check the filesize, then make the same program without json, and check the filesize, i did, and it's up to nothing...

xbeaudouin:
I don't understand, you want to test the jsoncpp lib present on freebsd ?

gizmocuz:
No, you can check out the binary additional bytes when including jsoncpp in the code.
Note, this is nothing, we also use a very small part of the library, so the overhead is even less

xbeaudouin:
Hello,

I did the check : 
Domoticz with dynamic jsoncpp : 
-rwxr-xr-x  1 kiwi  kiwi  7588824 Apr 21 09:52 domoticz

Domoticz with static jsoncpp : 
-rwxr-xr-x  1 kiwi  kiwi  7757256 Apr 21 10:17 domoticz

I think the overhead is not small as you said (all binaries has been stripped to check that before).

Regards,

gizmocuz:
Thanks 
Well 20kb is nothing but yep

xbeaudouin:
I see more 150k than 20kb. This is small, I agree but not nothing IMHO.

PR #1491, opened 2017-04-20
Teleinfo: Fix for bug reported in the forum

DebugBill:
There was a typo preventing white days to be shown in the "demain" (tomorrow) alert box.
Could only be detected recently as April 20th is the first white day for about a month.
Hope red is OK as there will not be any until next winter.


gizmocuz:
Thanks!

PR #1492, opened 2017-04-20
Sysfs GPIO: Master/Slave minor update

hvbommel:
Minor change to update state of devices in Master/Slave configuration after hardware startup.


gizmocuz:
Thanks!

PR #1494, opened 2017-04-22
Added Evohome Multi controller functionality to filter out devices from nearby installations

Evsdd:
Hi,

This update adds a new parameter to the Evohome Hardware device to capture the DeviceID of the system controller.  The controller DeviceID is then used to filter out devices from other nearby Evohome installations and fixes many problems with the set-up and control of Evohome systems via Domoticz in these environments.


hardware/evohome.cpp
```diff
@@ -173,15 +178,42 @@ bool CEvohome::StartHardware()
 		m_retrycntr=RETRY_DELAY; //will force reconnect first thing
 		
 		std::vector<std::vector<std::string> > result;
-		result = m_sql.safe_query("SELECT Name,DeviceID,nValue FROM DeviceStatus WHERE (HardwareID==%d) AND (Unit==0)", m_HwdID);
+		result = m_sql.safe_query("SELECT Name,DeviceID,nValue FROM DeviceStatus WHERE (HardwareID==%d) AND (Unit==0) AND (Type==%d)", m_HwdID, (int)pTypeEvohome);
+
```
gizmocuz: Should the type not always be evohome with this hwid?

hardware/evohome.cpp
```diff
@@ -173,15 +178,42 @@ bool CEvohome::StartHardware()
...
-		result = m_sql.safe_query("SELECT Name,DeviceID,nValue FROM DeviceStatus WHERE (HardwareID==%d) AND (Unit==0)", m_HwdID);
+		result = m_sql.safe_query("SELECT Name,DeviceID,nValue FROM DeviceStatus WHERE (HardwareID==%d) AND (Unit==0) AND (Type==%d)", m_HwdID, (int)pTypeEvohome);
+			
 		if (result.size()>0)
 		{
 			std::vector<std::string> sd=result[0];
 			std::stringstream s_strid;
 			s_strid << std::hex << sd[1];
-			s_strid >> m_nDevID;
+			if (s_strid.str().size() != 0)  // Check whether the controller's DeviceID has been set
```
gizmocuz: Instead of size, please use the empty boolean function

gizmocuz (commented):
Thank you for the patch.. To small reviews

hardware/evohome.cpp
Evsdd: I've added the additional Type==pTypeEvohome check here as I'm identifying the Device ID of the controller.  I've discovered through testing that other Evohome devices can be created with Unit==0 (relays and zone temeperature sensors).  Without this additional check, this query can return multiple devices leading to mis-identification of the Controller ID.

hardware/evohome.cpp
Evsdd: Thanks for the feedback.  I'll switch to using .empty() in the code and re-test.

Evsdd:
Hi Gizmocuz,  Thanks for your review and comments.  I'll review the code and switch to using .empty(), as suggested.  I think it uses .size()>0 extensively at the moment.  See my response about your other comment.  The additional check Type==pTypeEvohome is needed here to catch non-controller Evohome devices.

gizmocuz:
Thank you... Could you also increase the appcache version number in the file when you make changes to the Web part?

Evsdd:
Thanks Gizmocuz,  appcache version now updated.

gizmocuz:
Thanks!

PR #1498, opened 2017-04-25
Fix for negative outside temperatures (Fahrenheit) and update frequen…

norrland1:
…cy increased.

Based on discussion in thread:
https://www.domoticz.com/forum/viewtopic.php?f=6&t=10530&p=120176&hilit=davis#p120176

Conversion of two complements form int returned by Davis Weather Link to make it work with negative Fahrenheit outside temps (around -18 degrees celsius). 

Logger only caught 1 out of 10 readings with previous setting (30s) which had an impact on especially wind readings. A new reading is sent every 2.5s and pulling at a 4 sec interval will get most readings, miss some occasionally but works stable. Having a 3s reading is not stable and causes the logger to disconnect and reconnect too often.

### DIFF
--- a/hardware/DavisLoggerSerial.cpp
+++ b/hardware/DavisLoggerSerial.cpp
@@ -19,8 +19,8 @@
 	//#define DEBUG_DAVIS
 #endif
 
-#define RETRY_DELAY 30
-#define DAVIS_READ_INTERVAL 30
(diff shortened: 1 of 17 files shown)


gizmocuz:
Sorry, something went wrong with this pull request
Could you make it again please?

PR #1501, opened 2017-04-26
Add support for Horstmann SSR302

cyclops1982:
First time i add support for a device. So i have no clue if this is all correct or not. 
Manual (with zwave details) can be found here:
http://products.z-wavealliance.org/ProductManual/File?folder=&filename=Manuals/1619/SSR%20302%20User%20Manual.pdf


gizmocuz:
Thank you for this 
You will have to submit this to openzwave, we update regularly their Config files

PR #1503, opened 2017-04-28
Update DavisLoggerSerial.cpp

norrland1:
Added conversion of negative (Fahrenheit) outside temp values and changed update frequency for Davis Weather Link.

### DIFF
--- a/hardware/DavisLoggerSerial.cpp
+++ b/hardware/DavisLoggerSerial.cpp
@@ -19,8 +19,8 @@
 	//#define DEBUG_DAVIS
 #endif
 
-#define RETRY_DELAY 15
-#define DAVIS_READ_INTERVAL 4
+#define RETRY_DELAY 30
+#define DAVIS_READ_INTERVAL 30
 
 #define round(a) ( int ) ( a + .5 )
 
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thanks! Why you need such a short update interval?
We are talking about weather sensors here.....
Could you revert the interval?

norrland1:
Because the weather station is sampling data every 2.5s. Having an interval of 30s means that a lot of samples are missed and that affects the wind meter. Storing only one out of ten samples will, as you understand, have Domoticz miss a lot of values and thus show wrong peak values for the wind. 
One of the reasons to use a Davis weather station is that is reports wind more frequently than the cheaper stations that samples once every 40s or so. 
So I would like to keep the short interval, or even better, make it configurable in the settings, but I don't know how to fix that. Hopefully someone else can pick that up.

gizmocuz:
That's incorrect 
We sample every 5 minutes 
In those 5 minutes we receive 10 values which are taken into account in Domoticz (there is a special wind calculator)

norrland1:
Ok, so what you are saying is that it does not matter how often the Davis logger samples data since the wind calculator only stores 10 values per 5 min? 
I was not aware of that and agree that there is no use to shorten the interval on the Davis logger. 
But best would be, in my opinion, to fix the timing on the entire chain, to make the interval configurable.

gizmocuz:
Thanks! This is not a bug but intended

PR #1504, opened 2017-04-28
Sysfs gpio: only write to hardware outputs

hvbommel:
Only write to hardware outputs, do not allow writes to inputs. Remove some unused database selects.


gizmocuz:
Thanks

PR #1505, opened 2017-04-28
Add support for Kaifa DSMR5 gas meter

gordonb3:
As indicated in forum topic [P1 smart Meter: Gas value readout stays zero](https://www.domoticz.com/forum/viewtopic.php?f=14&t=16800) the gas usage on the new Kaifa meters uses a new ID.

I also changed some table style formats to line up correctly and renamed some defines:
- made them all upcase for correct colour coding in editor
- changed prefix P1GTS for gas usage IDs to P1GU


gordonb3:
Hold on...

This needs a revision as it is actually not correct. According to the DSMR documentation there are four channels on which the gas meter can correspond with the smart meter. Channel 2 is not typical to the Gyr E350 and channel 3 is not typical to the new Kaifa meter. It probably all depends on the guy that installs the meter which channel will be used (and may also be 4, which is currently not defined)

gizmocuz:
Thanks!

PR #1506, opened 2017-04-28
When HHTP client gets an HHTP Error, log it!

DebugBill:
Very useful to diagnose authentication issues between devices, trust me :-)


httpclient/HTTPClient.cpp
```diff
@@ -115,7 +116,34 @@ bool HTTPClient::GETBinary(const std::string &url, const std::vector<std::string
...
+                                case 404:
+                                        _log.Log(LOG_ERROR,"HTTP 404: Not Found");
+                                        break;
+                                case 500:
+                                        _log.Log(LOG_ERROR,"HTTP 500: Internal Server Error");
+                                        break;
+                                case 503:
+                                        _log.Log(LOG_ERROR,"HTTP 503: Service Unavailable");
+                                        break;
+                        }
```
gizmocuz: Thanks!
Could you also add a default case handler that displays the response code?

gizmocuz (commented):
Thanks

httpclient/HTTPClient.cpp
DebugBill: You're right, it added

gizmocuz:
Thanks!

PR #1510, opened 2017-04-30
Implementation of Manual Eco mode for Nest thermostat

dotms:
Added functionality for the Nest thermostat device plugin. It allows switching the operation mode of the thermostat between 'manual-eco' (effectively setting the thermostat to 'do-not-heat-or-cool') and 'schedule' which is the regular schedule mode. 

The difference with 'away' mode is that 'manual-eco' forces the thermostat not to heat or cool, even when someone is home. Away mode disables itself as soon as it detects motion, this one does not.


gizmocuz:
Thanks!

PR #1511, opened 2017-04-30
Fixed new bug preventing EcoDevices to be created and graph issue

DebugBill:
New bug appeared in HardwareController.js after recent GUI changes (I guess). EcoDevices could be updated but not created anymore.

Graphs can be inconsistent when update rate is lower than 5mn. Force update to be at least every 5mn

Teleinfo devices always have decimals. When creating one (TeleinfoMeter or EcoDevices hardware pieces), force global P1 setting to "with decimals" as there is almost no chance anybody runs both a P1 with no decimals (specifically NL) and a Teleinfo meter (specifically FR). Anyway, once created, the user can still revert this setting to "no decimals".


gizmocuz:
Thanks!

PR #1513, opened 2017-05-01
adjust for possible (hasty) clock skew in P1 gas meter

gordonb3:
According to [this forum topic](https://www.domoticz.com/forum/viewtopic.php?f=56&t=17357) the clock in the autonomous gas meter that is linked to the smart meter can be hasty, causing incorrect readings in Domoticz. Since the meter is only expected to change on the hour we can adjust for any hastiness by ignoring any updates during the last four minutes of the hour.

### DIFF
--- a/hardware/P1MeterBase.cpp
+++ b/hardware/P1MeterBase.cpp
@@ -204,9 +204,15 @@ bool P1MeterBase::MatchLine()
 					)
 				{
 					//only update gas when there is a new value, or 5 minutes are passed
-					m_lastSharedSendGas=atime;
-					m_lastgasusage=m_p1gas.gasusage;
-					sDecodeRXMessage(this, (const unsigned char *)&m_p1gas, "Gas", 255);
+					struct tm tma;
+					localtime_r(&atime, &tma)
+					// ...but don't accept new values in the last four minutes of the hour
+					// to correct for possible (hasty) clock skew in the gas meter.
+					if (tma.min<56){
+						m_lastSharedSendGas=atime;
+						m_lastgasusage=m_p1gas.gasusage;
+						sDecodeRXMessage(this, (const unsigned char *)&m_p1gas, "Gas", 255);
+					}
 				}
 			}
 			m_linecount=0;
--- a/hardware/P1MeterBase.cpp
+++ b/hardware/P1MeterBase.cpp
@@ -205,7 +205,7 @@ bool P1MeterBase::MatchLine()
 				{
 					//only update gas when there is a new value, or 5 minutes are passed
 					struct tm tma;
-					localtime_r(&atime, &tma)
+					localtime_r(&atime, &tma);
 					// ...but don't accept new values in the last four minutes of the hour
 					// to correct for possible (hasty) clock skew in the gas meter.
 					if (tma.min<56){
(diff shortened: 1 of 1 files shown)


gizmocuz:
Then I would think it is better to just store the latest gas value, together with the timestamp
At xx:05 check if we received a new timestamp, and if yes, send the gas value to Domoticz

gordonb3:
Thought about that, but that would require something of a full redesign of at least P1MeterBase as DSMR4+ has the gas timestamp and -usage on the same line whereas the current code can only handle one value per line. While probably not a bad idea in the long run, given that the code base seems a bit outdated for the current DSMR standard, that will require more time than I currently can spare.

gizmocuz:
This patch can cause no gas value to be updated in the last four minutes of an hour... Result ing in a complete hour of gas lost
I have a look at this shortly 
We already store the gas value and timestamp so it could not be too difficult

gordonb3:
Actually: we're not doing anything with the timestamps supplied by the meter itself. The timestamps stored are our own local times.

Losing data seems impossible as I don't think the gas meter is supposed to update more than once per hour (on the hour to be exact - every sample message I've seen so far has a gas timestamp in which the last four digits are 0) and the smart meter keeps replicating this info in every P1 message sent.

gizmocuz:
The new meters are actually updating every 10 minutes.. or it was 10 seconds as well
If there is a clock mismatch and the meter is sending an update (old meter, once an hour) at 09:59 , then we miss it now with this patch

gordonb3:
Okay, I verified that. Not true.

What the patch does is completely ignore any gas meter updates during the last four minutes of the hour. There is no discarding of the update though. Once the full hour has passed it will be processed and receive a timestamp that has the correct hour. This should not be different for new meters sending an update every ten minutes. As these are battery operated devices I would not expect them to ever send updates every ten seconds.

As said: the alternative would be to process the timestamps that are in the P1 datagrams. Which the current code cannot support.

gordonb3:
Dug into the standards. Not easy - huge pay walls.

New DSMR5 meters apparently exchange gas values every 5 minutes. I also found that what may look as a gas meter may in fact also be a heat or water meter, so we should verify the meter type before accepting any values from it.

May be wise to set a smaller time frame to adjust for negative clock skew for this particular patch, but it does appear that a full redesign of the P1 code is in order.

gizmocuz:
As suggested by Pim: "I saw the reading changed just seconds before the new hour. Can I maybe adjust the script so it gets the readings a few minutes after the new hour?"
This might work, and is very easy.. we already store the last gas value, we also store the time it was received
if we check at xx:05 if there was a new gas reading in lets say the last 15 minutes, then we send the value to the actual database (routine)
We set the timestamp to 'now'-3600, so it wont be send the next time (but there will be an update)

Another way, send it every 5 minutes ( timesecond%5) except when it's 00 (zero)
This way, the new p1 meter with a frequenter update will also work

gordonb3:
That seems like the wrong way around. 00 is not the issue, 59 is. With the new gas meters updating every 5 minutes the four minute time frame might cause issues if their clock was behind.

Anyway. Since I'm currently stuck with the evohome web implementation I did some testing with a revision of the P1 code. Not complete yet, but it includes verifying that gas meter is in fact a gas meter, and P1 version checking so we can make deliberate choices depending on version. I'm currently unsure whether or not to use the timestamps supplied in the P1 data, but that would eliminate all timing conflicts.

PR #1514, opened 2017-05-01
Replace "Domoticz" with instance name in notification messages

DebugBill:
Tabs in browser can contain a instance name rather than the more generic "Domoticz", I had an issue with error notification coming in with all the same subject line starting with "Domoticz". I could not differentiate easily wich instance it came from.

Now the notification subject starts with the instance name, which default value is Domoticz but can be replaced with something more meaningfull for the user.


gizmocuz:
Please fork a clean master copy of Domoticz, make the changes commit and make a pull request
This pull request is not clean as you can see
Thanks

DebugBill:
What is "unclean" in this pull request? 1 file changed, 4 lines added, 3 removed. Hard to make it any cleaner.

gizmocuz:
Surely you also see all above commit messages
While they might be not merged, I do not like to have another gui merge by accident, that's why a clean pull is highly appreciated 
Not to fond of the "title" preference field name... Think we should change this to "Instance name"

DebugBill:
Yes I see the GUI mistake that made me spend quite some time fixing issues in my code base. I also a fixed 2 days ago a non reported bug I was lucky to find where some of my work got half reverted for no clear reason.

Working on Domoticz is a hobby, so it has to be fun. I'm affraid currently it is not as I spend more time fixing issues I did not create than adding stuff I feel useful. I'm done

gizmocuz:
That is probably why you should start with a clean fork
I think you also merged / checkout the gui branch
Also note that not all people are experienced with 'git' and that this user made a mistake and also said sorry on the forum 
Sorry to hear you feel "done"

PR #1515, opened 2017-05-01
Add HttpOnly flag to cookie. Prevents client-side access to session ID

wez3:
Tested by compiling Domoticz on a Raspberry pi. After compilation the cookie was HttpOnly protected. This should not break things. If Domoticz uses client-side scripts to access the cookie HttpOnly might introduce issues. Searched the Domoticz code, did not find any Javascript client-side scripts to access the cookie (document.cookie).


gizmocuz:
Thanks!

PR #1519, opened 2017-05-02
Python plugins

moroen:
First version of Python-Events into Python-Plugins branch.


moroen:
Build fails on Linux due to a undefined nullptr in PythonObjects.cpp, I think it started after the latest commit to Python-Plugins?

dnpwwo:
I've pushed an update to fix the nullptr error.  I hadn't got as far as compiling the latest changes on Linux.

PR #1520, opened 2017-05-03
[1Wire] OWFS is always available

szczukot:


gizmocuz:
Why is this, even if it is not installed?

szczukot:
One Wire File System may be installed on any hardware. Domoticz can be installed on another computer.
OWFS is only file system - we can set path to it from everywhere. This is not only file system from local machine (domoticz). We can use shared path resource etc.

For example, i run domoticz on computer with Windows, but it get data from OWFS, installed on RaspberryPI.

gizmocuz:
Sorry, this is not what i meant.
When OWFS is not installed, we should not show the hardware (like it is now)
When OWFS is installed and detected, then we should show it

szczukot:
What means "OWFS is not installed" or "installed".
Domoticz don't know that owfs is installed or not.

szczukot:
Domoticz get data from path, where exists info about devices. This path may by for example //SERVER/path_to_1wire
Domoticz dont check that OWFS is installed on local machine or not.

gizmocuz:
Maybe this was indeed the case when this library was build... a long time ago...
There is also an windows owfs version... not sure if this is been in use by anyone, but now we have some duplicate code

szczukot:
Exactly.
Today, domoticz don't working with OWFS on Windows.
After fix - will work.

gizmocuz:
Thanks!

PR #1521, opened 2017-05-03
Prepare for integration of evohome client

gordonb3:
I'm planning to integrate Evohome client ([see project here](https://github.com/gordonb3/evohomeclient)) into Domoticz. This requires separating the HGI80 specific workers from the current all-in-1 code.

Note: this is work in progress. There are no real changes in this commit yet. It only splits the current evohome.cpp and evohome.h in two and corrects the static references in mainworker and RFXNames.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -311,7 +311,8 @@ hardware/EcoDevices.cpp
 hardware/EnOceanESP2.cpp
 hardware/EnOceanESP3.cpp
 hardware/Ec3kMeterTCP.cpp
-hardware/evohome.cpp
+hardware/EvohomeBase.cpp
+hardware/EvohomeSerial.cpp
 hardware/ETH8020.cpp
 hardware/FritzboxTCP.cpp
 hardware/GoodweAPI.cpp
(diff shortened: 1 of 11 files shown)


gordonb3:
Scripted evohome now moved to its own class. Obsolete lines in evohome serial deleted and references in mainworker revised.

gordonb3:
AppVeyor seems to fail because the msbuild files are not synched?

gizmocuz:
Yes, if you add/delete files, you also need to do this in the visual studio project files

gordonb3:
Unsure how to do that. Only reference to the old file appears to be in `msbuild/domoticz.vcxproj.filters` which I did alter and include in this push. AppVeyor still references the old file though, likely because it respects the gitignore on msbuild.

gizmocuz:
No, its not only in the filters, just search the whole domoticz source/msbuild folder for files containing the .cpp file
It should be in 'domoticz.vcxproj' also

gordonb3:
Now how did I miss that?

gordonb3:
Right. That got fubarred because of changes to the deleted file.

Resuming [here](1532)

PR #1522, opened 2017-05-03
Http errors: log the URL on top of error code

DebugBill:
Subject says it all


DebugBill:
Yes, I know, unclean....... but a new fork means I loose all my history? Don't like the idea much

gizmocuz:
Not really... If you do a new fork, you get a second domoticz repo under your account. Consider this as clean.
You make the changes there, commit, make the pull request, it will be merged, and you can delete that fork
If i just 'merge' this, i also get the history, i need to queeze this, which i will do
But maybe in the feature...

gizmocuz:
this can be dangerous, as the url can have a username/password.
we need to filter this out when available, it can also have a secret api key, or.....
maybe you can only display the 'url' part.

Example, original url:

https://user:password@my.domain.com/test.cgi?secret-api=63246294&password=27374&somethingelse=123

then if we just displayed:
my.domain.com

maybe that would be enough...

But... i think most classes calling the HTTP query function also check if it succeeds, and if not, log a message, and in the log message it always starts with the class name

Example:
	if (!HTTPClient::GET(sURL, sResult))
	{
		_log.Log(LOG_ERROR, "AtagOne: Error requesting token!");
		return "";
	}

So if this is not done in 'a' class, then we need (or the original author, but faster if we do this) add this

DebugBill:
I understand your point, but the goal here was to answer the question raised by Siewert Lameijer who wonders why he gets a 404 and apparently does not get any other message from another class. Filtering out the information may remove the part causing the issue. 
The logs contain sensitive information, just like the database does. Maybe they should not be readable by all users as they are today?

You can close the request

gizmocuz:
the database is not readable by any user, only the domoticz user, and most parts are encrypted
But i have a quick look on what calls are not logged, but probably for 'this' user, there was another log line underneath it that explains where it was originated from
Thanks for this patch

gizmocuz:
Found one place (YouLess) where this was not logged correctly.

PR #1524, opened 2017-05-04
Python plugins: Bugfix for Options parameter for "Custom" device

999LV:
fixes SQL crash and failure to register the Options parameter when
creating Custom devices within the Python plugins framework


oohlaf:
👍 this was indeed a typo

gizmocuz:
Thanks!

PR #1526, opened 2017-05-05
GPIO complete rewrite, dropped wiringPi dependency

jvandenbroek:
Inspired by the new sysfs gpio hardware code (credits to @hvbommel ), I completely rewrote the 'main' GPIO function to drop wiringPi dependency and keep the exact same functionality. It's quite a big update, but I think I've tested it thoroughly enough.


jvandenbroek:
Hmm why does AppVeyor fail? Shouldn't WITH_GPIO be undefined for the Windows compiler?

PR #1527, opened 2017-05-05
Sanitize user input

wez3:
Compiled and tested on raspberry pi. Please also whether all functionalities are still working :).

### DIFF
--- a/main/WebServer.cpp
+++ b/main/WebServer.cpp
@@ -7901,7 +7901,7 @@ namespace http {
 					isAlpha = false;
 				}
 			}
-			if ((order == "") || (!isAlpha)) {
+			if (order.empty() || (!isAlpha)) {
 				strcpy(szOrderBy, "A.[Order],A.LastUpdate DESC");
 			}
 			else
@@ -8140,7 +8140,7 @@ namespace http {
 						bAllowDeviceToBeHidden = true;
 					}
 
-					if (order == "" || (!isAlpha))
+					if (order.empty() || (!isAlpha))
 						strcpy(szOrderBy, "A.[Order],A.LastUpdate DESC");
 					else
 					{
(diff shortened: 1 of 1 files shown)


main/WebServer.cpp
```diff
@@ -7893,11 +7893,21 @@ namespace http {
 			}
 
 			char szOrderBy[50];
-			if (order == "")
+			char szQuery[1000];
```
gizmocuz: Could you make this a std::string ? we are not using sprintf

main/WebServer.cpp
```diff
@@ -7893,11 +7893,21 @@ namespace http {
...
+			bool isAlpha = true;
+			const std::string orderBy = order.c_str();
+			for(int i = 0; i < orderBy.size(); i++) {
+				if( !isalpha(orderBy[i])) {
+					isAlpha = false;
+				}
+			}
+			if ((order == "") || (!isAlpha)) {
```
gizmocuz: could you change this to 
`if ((order.empty()) || (!isAlpha)) {`

main/WebServer.cpp
```diff
@@ -8129,15 +8140,16 @@ namespace http {
 						bAllowDeviceToBeHidden = true;
 					}
 
-					if (order == "")
+					if (order == "" || (!isAlpha))
```
gizmocuz: could you change this to order.empty() instead of == ""

gizmocuz: I know some other have been peeking/poking in this function as well, but i prefer checking against .empty()

main/WebServer.cpp
```diff
@@ -8244,14 +8258,15 @@ namespace http {
...
```
gizmocuz: could you change this to order.empty() instead of == ""

gizmocuz (commented):
Thanks!! Just a few requests

main/WebServer.cpp
wez3: I am using strncpy later on to concat this variable. This function takes chars. Is there a equivalent function which takes std::string?

gizmocuz: Yes, += you can just add strings.. Or I missed something here 🤔

wez3:
Changed it to std::string.

gizmocuz:
Thanks!!

PR #1529, opened 2017-05-06
GPIO do not create devices on startup, poller status update fix

jvandenbroek:
Disabled auto creating devices, since these should be added manually. Poller didn't update DB status is changed, fixed.


jvandenbroek:
This also fixes the unintended deletion of already added output pins (since the new method uses the options column to determine whether the device is valid or not).

gizmocuz:
Thanks!

PR #1530, opened 2017-05-07
Sysfs gpio: active_low output handling improved

hvbommel:
Removed logic that was reversing the output value based on active_low settings based on a reported issue on the forum. This will give the user direct control of active_low or not for the "Generic sysfs gpio outputs".


gizmocuz:
Thanks!

PR #1532, opened 2017-05-08
Prepare for integration of evohome client: retry after fubar from main

gordonb3:
Splits the HGI80 specific workers from the current all-in-1 code to allow adding a new worker for the web based evohome controls.


gizmocuz:
Thanks!

PR #1535, opened 2017-05-08
Sysfs gpio: minimum output state changes on system reboot

hvbommel:
The SysfsGPIO outputs states must reflect the actual Domoticz database state after startup. This change makes sure the minimum output state changes, on System & Domoticz startup, take place.


gizmocuz:
Thanks!

PR #1537, opened 2017-05-09
P1 meter revision

gordonb3:
Key changes:
 - auto detect gas meter on M-Bus
 - delay gas readings if the meter's clock is ahead
 - don't update gas if there is no meter

### DIFF
--- a/hardware/P1MeterBase.h
+++ b/hardware/P1MeterBase.h
@@ -11,40 +11,41 @@ public:
 	P1MeterBase(void);
 	~P1MeterBase(void);
 
-	P1Power	m_p1power;
-	P1Gas	m_p1gas;
+	P1Power	m_power;
+	P1Gas	m_gas;
 private:
 	bool m_bDisableCRC;
 	int m_ratelimit;
+
+	unsigned char m_p1version;
+
+	unsigned char m_buffer[1400];
+	int m_bufferpos;
+	unsigned char m_exclmarkfound;
+	unsigned char m_linecount;
+	unsigned char m_CRfound;
+
+	unsigned char l_buffer[128];
+	int l_bufferpos;
+	unsigned char l_exclmarkfound;
 
 	unsigned long m_lastgasusage;
 	time_t m_lastSharedSendGas;
 	time_t m_lastUpdateTime;
 
-	unsigned char m_linecount;
-	unsigned char m_exclmarkfound;
-	unsigned char m_CRfound;
+	float m_voltagel1;
+	float m_voltagel2;
+	float m_voltagel3;
 
-	float m_p1voltagel1;
-	float m_p1voltagel2;
-	float m_p1voltagel3;
+	unsigned char m_gasmbuschannel;
+	std::string m_gasprefix;
+	std::string m_gastimestamp;
+	double m_gasclockskew;
+	time_t m_gasoktime;
 
 	void Init();
 	bool MatchLine();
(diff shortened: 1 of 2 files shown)


gordonb3:
Hmmm... Can you see why AppVeyor fails? Appears to be a problem with compiling 1Wire sources, but I don't think I touched that?

gizmocuz:
Lately appveyor seems to have quite some building issues, this one is not related to your PR

hardware/P1MeterBase.h
```diff
@@ -25,9 +25,9 @@ class P1MeterBase : public CDomoticzHardwareBase
 	unsigned char m_exclmarkfound;
 	unsigned char m_CRfound;
 
-	float m_voltagel1;
-	float m_voltagel2;
-	float m_voltagel3;
+	float m_p1voltagel1;
```
gizmocuz: Maybe we could not rename this variables as we are already in the P1 class... thats double,
but maybe it's an idea to rename these two:

	P1Power	m_p1power;
	P1Gas	m_p1gas;

hardware/P1MeterBase.h
```diff
@@ -39,4 +39,12 @@ class P1MeterBase : public CDomoticzHardwareBase
 	unsigned char l_exclmarkfound;
 
 	bool CheckCRC();
+
+
+	unsigned char m_p1version;
```
gizmocuz: i would not add a 'p1' tag to the P1 class variables

gizmocuz:
Question did you had a gas meter to test this with ?

hardware/P1MeterBase.h
gordonb3: Well, it was there on several other variables. As this file appears to be included by several other source files I figured it might be a good idea to add the `p1` prefix because I don't know what the other files are doing with it - they may be extending the class and add the same variable name themselves.

gordonb3:
I do not have a gas meter in my home. I therefore used sample messages formatted in DSMR 2.2, DSMR 4.2 and DSMR 5.0 to test the various scenarios.

See something funny?

gordonb3:
Oops... I seem to have accidentally removed the gas commits

gordonb3:
Thoughts? Wishes?

gizmocuz:
Nope ;) Only the wish for the removal of the 'p1' in the variable names, but I would like to test this somehow in my live environment... Specially since some critical functions have been changed and you can't test the Gas usage changes

gordonb3:
Okay. Can do that.

The problem of course with this connector is that nobody can test every aspect because nobody can have all the different types and versions of the hardware. As a result you're always working in the dark somehow. The main change here is to make the channel identifier (that's the 'n' in the match parameter) a variable rather than assume that channels are linked to specific brands and/or types of meters. This should also open up the possibility to add support for different kinds of M-Bus devices, such as the water meter or heat (as I have - but this is not linked to the smartmeter).

gordonb3:
PS you know you can get the consolidated diff [here](1537.diff), do you? This will feed directly into `patch -p1`

gizmocuz:
Yes i know, only my P1/433Mhz domoticz is a slave domoticz 
(Uptime: 202 Dagen, 9 Uren, 2 Minuten, 55 Seconden)

I will just merge this patch and see if all runs good.

Will report back shortly ;)

gizmocuz:
Huston, we might have a problem:
https://github.com/domoticz/domoticz/issues/1556

PR #1538, opened 2017-05-09
Update rsroom.xml to support decimal values

techdude200:
Added decimal value descriptors for command class thermostat setpoint. This makes it possible to set the setpoint with decimal values and not only integers.


gizmocuz:
Thank you for this ! But you have to submit this patch to the openzwave repository
https://github.com/OpenZWave/open-zwave

PR #1540, opened 2017-05-13
Sanitize user input #2

wez3:
I added brackets that were missing. This fixes the issue for "Sanitize user input" as described here:
http://domoticz.com/forum/viewtopic.php?f=6&t=17467


gizmocuz:
Not good?

wez3:
It is good. But i have some issues creating a pull request with the previous commits as well. Are you able to do this? :)

wez3:
I will send a new PR soon...

PR #1541, opened 2017-05-13
Sanitize user input #2

wez3:
Added missing brackets.


gizmocuz:
Sorry, we have some merge problems here duo recent changes in the webserver class.
Is it possible to adjust the pull request?

PR #1544, opened 2017-05-18
Openbsd

vuokko:
I got Domoticz working out of box. Still there are linker ~~errors~~ warnings of misuse in the end of compilation, but to remove them would need to change every sprintf() function to snprintf() from code for example. 

Please review.


gizmocuz:
Thanks, but according this page:
http://man.openbsd.org/FreeBSD-11.0/linprocfs.5

You should also have /compat/linux/proc/meminfo

if that's correct, there should be no need for a new memory routine ? (... hope thats correct)

vuokko:
That's FreeBSD manual. https://en.wikipedia.org/wiki/Procfs says:

> 4.4BSD cloned its implementation of /proc from Plan 9.[citation needed] As of February 2011, procfs is gradually becoming phased out in FreeBSD.[1] It was removed from OpenBSD in version 5.7, which was released in May 2015, because it "always suffered from race conditions and is now unused".[2]

So it might be better to look how FreeBSD gives memory information through systat interface and implement that in next ten years.

xbeaudouin:
@vuokko: Agreed, your patch maybe adapted to the FreeBSD port. And getting rid of infamous /compat/linux/proc/meminfo.

gizmocuz:
Yes, but does this take care of the cache as well ? Otherwise a call to free would also reveal 'some' free memory

vuokko:
Good catch. That calculates usage totally wrong.
Sample: 

> Memory: Real: 745M/1509M act/tot Free: 6309M Cache: 380M Swap: 0K/4432M

I think I'll better to dig in to the sysctl interface and try to find out the right variables to query.
I think it can be sysctl()'d out of the system and  will look much like sysinfo way in FreeBSD.

gizmocuz:
Thanks !

PR #1546, opened 2017-05-19
Fix for Sound Level subtype

SkorpionNL:
I added the Sound Level vType, because pushing Sound Level type returns a "Could not determine Data Push value" when data is being pushed via HTTP.


gizmocuz:
Thanks!

PR #1548, opened 2017-05-21
Fix for Sound Level Subtype

SkorpionNL:
nvalue of sound level returned 0. Changed to rawsendvalue.

### DIFF
--- a/push/BasePush.cpp
+++ b/push/BasePush.cpp
@@ -579,7 +579,7 @@ std::string CBasePush::ProcessSendValue(const std::string &rawsendValue, const i
 	}
 	else if (vType == "Sound Level")
 	{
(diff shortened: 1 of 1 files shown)


push/BasePush.cpp
```diff
@@ -579,7 +579,7 @@ std::string CBasePush::ProcessSendValue(const std::string &rawsendValue, const i
 	}
 	else if (vType == "Sound Level")
 	{
-		sprintf(szData,"%d", nValue);
+		sprintf(szData,"%d", atoi(rawsendValue.c_str()));
```
gizmocuz: Why dont you do

strcpy(szData, rawsendValue.c_str());

SkorpionNL: Yes that's better. I will edit it.

gizmocuz:
Thanks!

PR #1549, opened 2017-05-21
Evohome web API connector

gordonb3:
This new hardware connector provides native support for the Evohome web API.


gizmocuz:
Great! Thanks!

Could you check the following compile warnings:

..\hardware\EvohomeWeb.cpp(360): warning C4018: '<': signed/unsigned mismatch [C:\projects\domoticz\msbuild\domoticz.vcxproj]
..\hardware\EvohomeWeb.cpp(366): warning C4018: '<': signed/unsigned mismatch [C:\projects\domoticz\msbuild\domoticz.vcxproj]
..\hardware\EvohomeWeb.cpp(653): warning C4018: '<': signed/unsigned mismatch [C:\projects\domoticz\msbuild\domoticz.vcxproj]
..\hardware\EvohomeWeb.cpp(1011): warning C4018: '<': signed/unsigned mismatch [C:\projects\domoticz\msbuild\domoticz.vcxproj]
..\hardware\EvohomeWeb.cpp(1021): warning C4018: '<': signed/unsigned mismatch [C:\projects\domoticz\msbuild\domoticz.vcxproj]
..\hardware\EvohomeWeb.cpp(1001): warning C4101: 'j_day': unreferenced local variable [C:\projects\domoticz\msbuild\domoticz.vcxproj]
..\hardware\EvohomeWeb.cpp(999): warning C4101: 'i': unreferenced local variable [C:\projects\domoticz\msbuild\domoticz.vcxproj]

gordonb3:
Okay, I think this should do it.

gizmocuz:
Hmmm when you check against a 'size' object, then you should use size_t

gordonb3:
Hmmm... right. Is that a Windows compiler thing? The GNU compiler I use isn't that picky at all.

gordonb3:
There were actually some more instances of those than warnings were generated on. It doesn't seem to be consistent at all.

gordonb3:
Funny.... I don't think I've seen this before in my earlier tests....

> 2017-05-22 12:46:46.320  Error: EvohomeWeb: HTTP client error at get status!
2017-05-22 12:47:46.325  EvohomeWeb: connect to Evohome server
2017-05-22 13:01:13.772  EvohomeWeb: Worker stopped...
2017-05-22 13:01:13.772  EvohomeWeb: Worker started...
2017-05-22 13:01:23.773  EvohomeWeb: connect to Evohome server
2017-05-22 14:01:59.506  Error: EvohomeWeb: HTTP client error at get status!
2017-05-22 14:02:59.515  EvohomeWeb: connect to Evohome server
2017-05-22 15:03:00.627  Error: EvohomeWeb: HTTP client error at get status!
2017-05-22 15:04:00.632  EvohomeWeb: connect to Evohome server
2017-05-22 16:04:46.408  Error: EvohomeWeb: HTTP client error at get status!
2017-05-22 16:05:46.413  EvohomeWeb: connect to Evohome server

Almost looks like a pattern that the session is dropped an hour after it was initialized.

gizmocuz:
It seems like it.... Shall i wait a bit before merging this ?

Ps. In the above case, the new compiler being used in Visual Studio is correct ;)

gordonb3:
It's not a big issue. With the stand-alone client which is not a TSR I do a login with every run. It's because Domoticz runs as a service I figured I could reuse static data that I already loaded during a previous run, including the session parameters. This saves three web calls for every run but somehow I failed to notice that Honeywell drops the session after an hour, regardless off whether its use was a lot shorter ago.

Should be easily solved by adding a timer and discard the session before the server tells me they did.

hardware/EvohomeWeb.cpp
```diff
@@ -77,7 +77,7 @@ void CEvohomeWeb::Init()
 
 bool CEvohomeWeb::StartSession()
 {
-	_log.Log(LOG_NORM, "EvohomeWeb: connect to Evohome server");
+	_log.Log(LOG_NORM, "EvohomeWeb: start new session with Evohome server");
```
gizmocuz: This will result in 24 log messages a day ? Maybe an option to silence this log for a session update ?

gordonb3: 24 **planned** log messages yes. A session restart can also be triggered by connection timeouts and login failures. Of course LOG_NORM is the highest verbose level of logging.

gizmocuz:
Thanks!

gordonb3:
Just upgraded my live system to 3.7502 and transferred the evohome devices. It's really a breeze... Just change the hardware type of the existing `evohome via script` connector to  `evohome via web API` and it registers the data like nothing has changed and all history intact.

PR #1550, opened 2017-05-21
Sysfs gpio: Support to add devices manually

hvbommel:
Added support to add devices manually. There is a new hardware configuration option named "Auto configure devices". The option is checked by default meaning behavior is the same as was before.. When this check is removed devices are no longer created automatically. Instead devices can then be created in  the switches "Manual Light/Switch" menu. To do this select Hardware: "SysfsGPIO" and Type: "SystfsGPIO" plus the SysfsGPIO" in and or outputs. The ID: can be set to any value. Values already used for other Ligthing2 devices can best be avoided. For SysfsGPIO the ID does not matter because gpio pins have a unique unit number already.

### DIFF
--- a/hardware/SysfsGPIO.h
+++ b/hardware/SysfsGPIO.h
@@ -8,19 +8,19 @@
 
 struct gpio_info
 {
-	int	pin_number;		// GPIO Pin number
-	int	value;			// GPIO pin Value
-	int	direction;		// GPIO IN or OUT
-	int	active_low;		// GPIO ActiveLow
-	int	edge;			// GPIO int Edge
+	uint8_t	value;		// GPIO pin Value
+	uint8_t	direction;	// GPIO IN or OUT
+	uint8_t	active_low;	// GPIO ActiveLow
+	uint8_t	edge;		// GPIO int Edge
+	uint8_t	db_state;	// Database Value
+	uint8_t	db_state;	// Database Value id1;			// Device id1
+	uint8_t	db_state;	// Database Value id2;			// Device id2
+	uint8_t	db_state;	// Database Valueid3;			// Device id3
+	uint8_t	db_state;	// Database Valueid4;			// Device id4
+	uint8_t	db_state;	// Database Valueid_valid;		// Device valid
+	uint8_t	db_state;	// Database Valueint request_update; // Request update
 	int	read_value_fd;	// Fast read fd
-	int	db_state;		// Database Value
-	int id1;			// Device id1
-	int id2;			// Device id2
-	int id3;			// Device id3
-	int id4;			// Device id4
-	int id_valid;		// Device valid
-	int request_update; // Request update
+	int	pin_number;		// GPIO Pin number
 };
 
 class CSysfsGPIO : public CDomoticzHardwareBase
--- a/hardware/SysfsGPIO.h
+++ b/hardware/SysfsGPIO.h
@@ -9,27 +9,27 @@
 struct gpio_info
 {
 	int		pin_number;		// GPIO Pin number
-	char	value;			// GPIO pin Value
-	char	direction;		// GPIO IN or OUT
-	char	active_low;		// GPIO ActiveLow
-	char	edge;			// GPIO int Edge
-	int		read_value_fd;	// Fast read fd
-	char	db_state;		// Database Value
-	int		id1;			// Device id1
-	int		id2;			// Device id2
-	int		id3;			// Device id3
-	int		id4;			// Device id4
-	char	id_valid;		// Device valid
-	char	request_update;	// Request update
+	int		read_value_fd;	// File descriptor
+	int8_t	request_update;	// Request update
+	int8_t	db_state;		// Database Value
+	int8_t	value;			// GPIO pin Value
+	int8_t	direction;		// GPIO IN or OUT
+	int8_t	active_low;		// GPIO ActiveLow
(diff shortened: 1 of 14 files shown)


main/WebServer.cpp
```diff
@@ -3600,25 +3600,26 @@ namespace http {
 						int ID = atoi(sd[0].c_str());
 						std::string Name = sd[1];
 						_eHardwareTypes Type = (_eHardwareTypes)atoi(sd[2].c_str());
-						switch (Type)
+
```
gizmocuz, line 3603: What was wrong with the 'switch' ?

www/app/HardwareController.js
```diff
@@ -123,8 +123,8 @@ define(['app'], function (app) {
 				(text.indexOf("Evohome") >= 0 && text.indexOf("script") >= 0) ||
 				(text.indexOf("YeeLight") >= 0) ||
 				(text.indexOf("Arilux AL-LC0x") >= 0) ||
-				(text.indexOf("sysfs gpio") >= 0)
-			) {
+				(text.indexOf("sysfs gpio") >= 0))
```
gizmocuz: Please leave the closing ) on the next line for reading purpose

www/app/HardwareController.js
```diff
@@ -4866,11 +4886,14 @@ define(['app'], function (app) {
 								$("#hardwareparami2caddress #i2caddress").val(data["Port"].substring(4));
 							}
 						}
-						else if (data["Type"].indexOf("GPIO") >= 0) {
+                        else if ((data["Type"].indexOf("GPIO") >= 0) && (data["Type"].indexOf("sysfs gpio") == -1)) {
```
gizmocuz: Please leave the ident correct

www/app/HardwareController.js
```diff
@@ -5113,18 +5136,17 @@ define(['app'], function (app) {
 			$("#hardwarecontent #divpollinterval").hide();
 			$("#hardwarecontent #divpythonplugin").hide();
 			$("#hardwarecontent #divrelaynet").hide();
-			$("#hardwarecontent #divgpio").hide();
+            $("#hardwarecontent #divgpio").hide();
```
gizmocuz: Please do not touch the 'ident', they where correct before

www/app/HardwareController.js
```diff
@@ -5113,18 +5136,17 @@ define(['app'], function (app) {
...
+            $("#hardwarecontent #divsysfsgpio").hide();
 
-			if (
-				(text.indexOf("TE923") >= 0) ||
+			if ((text.indexOf("TE923") >= 0) ||
```
gizmocuz: Please do not do reformatting/ident changes

hardware/SysfsGPIO.h
```diff
@@ -6,24 +6,35 @@
 #include "DomoticzHardware.h"
 #include "../main/RFXtrx.h"
 
+struct gpio_info
```
gizmocuz: I don't know how many items the vector will hold, but instead of 'int' could you use uint8_t (0-255), specially for boolean values like direction,active,edge, id's

gizmocuz (commented):
Great thanks!!

I have some small whishes/tweaks here... specially the reformatting/ident changes

main/WebServer.cpp
hvbommel: For a start a I added the case label indeed. (The switch case label became 102 for Sysfs). It looks like a value larger then 99 makes the switch fail the raspberry compiler (the case never matches). I did see other people mention this problem when I googled it. The "if's" fixed it.

hvbommel:
I think the indents are done by my visual studio editor once I change or "touch" a line. I see it replace tabes by spaces. What are your, or  the project, editor tab settings ? When I set them the same that will help me to format the code better.

hvbommel:
I have set tab to 4 now and disabled replace with spaces (for all languages). It looks like formatting is now as the rest. I will do the reformats with this setting. VisualStudio->Tools->Options->TextEditor->AllLanguages: Tabs=4 IndentSize=4 Indenting=None KeepTabs=selected (was deselected, ie it caused the poor formatting on other editors/viewers)

main/WebServer.cpp
gizmocuz: thanks!

gizmocuz:
I have maybe one giant request.... GPIO with capitals ?

hvbommel:
I have done the renames, both internal and external. I will test first before committing.
Now is the good time to change it, I agree...
PS: GPIO = General Purpose Input/output, that why I kept the uppercase. But I was in doubt indeed a while ago when I created it.

hvbommel:
Code formatting should meet expectations now, SysfsGPIO->SysfsGpio everywhere (.cpp, .h class etc) So I also changed the project files to make the build work (I hope).

hvbommel:
The rename from SysfsGPIO to SysfsGpio broke things into a bad shape. I am going to close this one and will revert. I will do a new pull request when I'm done.

hvbommel:
Will do a new pull request when rename is done. It is becoming indeed a "giant request".

gizmocuz:
My idents are set as follows:
image

gizmocuz:
No i mean in the web interface (but also the class name would be OK to rename)
In the hardware setup we now have GPIO and gpio, the first one is better

PR #1551, opened 2017-05-22
Fixed python-interpreter not terminated properly.

moroen:
The event-python interpreter wasn't being terminated when stopping the event system, fixed.

M


gizmocuz:
Thanks!

PR #1552, opened 2017-05-22
Sysfs manual auto

hvbommel:
See PR #1550. The problem was that windows did not change case of SysfsGPIO.cpp to SysfsGpio.cpp (same for .h) but explorer was showing the mixed case version (its a windows feature). Therefore the file was missing for the case sensitive Linux build causing it to fail. Should be ok now (I hope). I reapplied all changes on the base version that contained the spaces instead of tabs. I hope I did not miss any.


gizmocuz:
Best to write GPIO as GPIO (all capitals)

hvbommel:
Thanks for the feedback. Hardware: "Generic sysfs GPIO" and Manual device add: "Sysfs GPIO" is now displayed to the user. It should be good to go now.

gizmocuz:
Thanks

PR #1553, opened 2017-05-22
Websockets refactored

rkluwen:
So finally, here it is: The long awaited Domoticz interface with push events via websocket connections.
No more polling; widgets are updated as soon as a device status changes.
Only the Temperature and Weather pages work so far. The other pages are waiting on being re-written to modular angular widgets, instead of the ugly html markup in jQuery.

Also this version features a new mydomoticz port. Now, instead of connecting via a seperate port to mydomoticz, the http ssl port is being used for this, also using RFC 6455. In theory now it is possible to connect to mydomoticz via a websockets-enabled http(s) proxy (not tested).

In here, there's also a new notification type: "browser". Domoticz notifications are now also pushed via the websocket link and end up as browser notifications on your task bar, using browsers that support that feature.

I understand this is quite a massive pull-request. So take your time to review everything.
Neither do I claim everything to be 100% bug-free. Use at your own risc and comments and patches for improvement are welcome.


www/app/WeatherController.js
```diff
@@ -85,37 +86,66 @@ define(['app'], function (app) {
...
+                    if (typeof data.ServerTime != 'undefined') {
+                        $rootScope.SetTimeAndSun(data.Sunrise, data.Sunset, data.ServerTime);
+                    }
+
+                    if (typeof data.result != 'undefined') {
+                        if (typeof data.ActTime != 'undefined') {
+                            $.LastUpdateTime = parseInt(data.ActTime);
+                        }
+
+                        // Change updated items in temperatures list
```
gizmocuz: Should this be 'Weather List' ?

rkluwen:
Good catch. I changed it with a new commit. In fact, this is inherited from the master branch. There also it says temperature instead of weather.

gizmocuz:
Big big Thanks !!

PR #1560, opened 2017-05-24
Fix for forum bug report "can't select the GPIO hardware"

hvbommel:
Fix for forum bug report "I can't select the GPIO hardware when adding a manual Light/Switch Device." filed on 24-5-2017. 
Note: Please review this change because I see the break was in for a long time (since it was created in 2014). So it must have been broken for a long time or I am missing something. The "break" caused to only show the first found hardware ID. I reproduced the problem & after the change I see it work correctly on my system. I see now not only the 1st device but also the others, the available ones, listed in the switch statement. (I tested on Raspberry Pi). Why was this not reported before is my question.

### DIFF
--- a/main/WebServer.cpp
+++ b/main/WebServer.cpp
@@ -3627,7 +3627,6 @@ namespace http {
 							root["result"][ii]["idx"] = ID;
 							root["result"][ii]["Name"] = Name;
 							ii++;
-							break;
 						}
 					}
 				}


jvandenbroek:
Actually, this commit apparently broke it: https://github.com/domoticz/domoticz/commit/a8a1085883c3b4c983d87ab7694b9ed14451864c#diff-b178aa1b57f3c23ff1c6fbce4b495e30

Makes sense, I reviewed older version of the code.. Going to sleep now, will look into this tomorrow.

hvbommel:
Yes, now I see it. I replaced the "switch" by the "if". I should have removed the break at that time also, ie I created the bug. Using the switch the break does not harm as we only leave the switch, not the for-loop. Using the "if" & "break;" we do leave the for loop. That causes the problem. So I was wrong saying the code is there for a long time. That was true for the "break;" but I overlooked the change from "switch" to "if". So in the end I think my pull request does fix the problem.

hvbommel:
PS: I changed form "switch" to "if" because the "case" label was never matching for HTYPE_SysyfsGPIO on raspberry. I think it is because its value is 102 (larger then 99). After using the "if" is was working (of course)

jvandenbroek:
Ah well that can happen. I can't see why that wouldn't work with a switch statement, are you really sure you didn't break (literally ;)) it somewhere? Switch is so much nicer, but if it's really not working, this should be fine I guess.

hvbommel:
I tried hard to make to switch work and single stepped the code on raspberry from visual studio. It was never matching on my new label (int value 102).  Then I gave up and turned the "switch" into the "if" and that worked. I did not look at the assembly code. When I googled I found others mentioning the max case label value was 99. (A strange value, 128 or 255 would have made more sense to me). Also the code was working fine on my ubuntu system. Because I had it working I did stop putting more time in it at that point.

jvandenbroek:
Well let's just merge this fix, maybe in the future do some research why > 100 would cause issues (some stupid compiler optimization?).

gizmocuz:
Thanks!

Issue #1556, opened 2017-05-23
Domoticz crashes with P1 cable connected


phoenix1337:
Version: 3.7502
Build Hash: fdc764e
Compile Date: 2017-05-23 18:06:41

Hardware:
May 23 22:52:43 illumia kernel: [ 1658.625468] usb 1-6: new full-speed USB device number 4 using xhci_hcd
May 23 22:52:43 illumia kernel: [ 1658.771650] usb 1-6: New USB device found, idVendor=0403, idProduct=6001
May 23 22:52:43 illumia kernel: [ 1658.771654] usb 1-6: New USB device strings: Mfr=1, Product=2, SerialNumber=3
May 23 22:52:43 illumia kernel: [ 1658.771657] usb 1-6: Product: FT232R USB UART
May 23 22:52:43 illumia kernel: [ 1658.771660] usb 1-6: Manufacturer: FTDI
May 23 22:52:43 illumia kernel: [ 1658.771662] usb 1-6: SerialNumber: ALLDAW7
May 23 22:52:43 illumia kernel: [ 1658.774624] ftdi_sio 1-6:1.0: FTDI USB Serial Device converter detected
May 23 22:52:43 illumia kernel: [ 1658.774694] usb 1-6: Detected FT232RL
May 23 22:52:43 illumia kernel: [ 1658.774925] usb 1-6: FTDI USB Serial Device converter now attached to ttyUSB0

When trying to start with the P1 cable connected or connecting afterwards and activating it, i get the following error:

May 23 22:56:29 illumia domoticz: P1 Smart Meter: Serial Worker stopped...
May 23 22:56:29 illumia domoticz: P1 Smart Meter: Using serial port: /dev/ttyUSB-smart-meter
May 23 22:56:33 illumia domoticz: P1: Found gas meter on M-Bus channel 1
May 23 22:56:33 illumia domoticz: (Slimme meter) P1 Smart Meter (Stroom)
May 23 22:56:33 illumia domoticz: Error: Domoticz received fatal signal 6 !...
May 23 22:56:33 illumia domoticz: Error:   /usr/local/sbin/domoticz/domoticz(+0x25120d) [0x55d2b283820d]
May 23 22:56:33 illumia domoticz: Error:   /usr/local/sbin/domoticz/domoticz(_Z14signal_handleri+0x53) [0x55d2b28382e3]
[...]

PR #1562, opened 2017-05-24
defeat possible memory access error due to oversized timestamp value

gordonb3:
Somewhat running in the dark here, but this seems like the most likely cause for issue #1556


gizmocuz:
Thanks!

PR #1567, opened 2017-05-26
Notification use threads

jvandenbroek:
Spawning threads on notifications makes the interface much more responsive in case an handler is taking some time to process the notification (and reducing overal delay). There doesn't seem any code using return value of SendMessageEx, only for SendMessage. Since var priority is unused for the latter, I used value of -100 (arbitrary chosen) for it to determine it should use a thread of act as before.

### DIFF
--- a/notifications/NotificationBase.cpp
+++ b/notifications/NotificationBase.cpp
@@ -117,6 +117,7 @@ bool CNotificationBase::SendMessageEx(
 		fText = CURLEncode::URLEncode(fText);
 	}
 
+	boost::mutex::scoped_lock SendMessageEx(SendMessageExMutex);
 	bool bRet = SendMessageImplementation(Idx, Name, fSubject, fText, ExtraData, Priority, Sound, bFromNotification);
 	if (bRet) {
 		_log.Log(LOG_NORM, std::string(std::string("Notification sent (") + _subsystemid + std::string(") => Success")).c_str());
--- a/notifications/NotificationBase.h
+++ b/notifications/NotificationBase.h
@@ -56,6 +56,7 @@ protected:
 
 	int m_IsEnabled;
 private:
+	boost::mutex SendMessageExMutex;
 	std::string _subsystemid;
 	std::map<std::string, std::string* > _configValues;
 	std::map<std::string, std::string* > _configValuesBase64;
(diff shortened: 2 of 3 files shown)


notifications/NotificationBase.cpp
```diff
@@ -117,6 +117,7 @@ bool CNotificationBase::SendMessageEx(
 		fText = CURLEncode::URLEncode(fText);
 	}
 
+	boost::mutex::scoped_lock SendMessageEx(SendMessageExMutex);
```
gizmocuz, line 120: Why the mutex as this is the only place it's being used ? (i'm not so keep about mutexes, and try to avoid them where possible)

gizmocuz:
Would it be an option to add a 'concurrent_queue' in the base notification system (like CMainWorker::m_rxMessageQueue) and have all notifications going via a queue/thread ?

notifications/NotificationBase.cpp
jvandenbroek: I wanted to keep the 'thread queue' in order, if multiple calls are made to SendMessageEx, I'm not sure what happens (haven't checked the code of all notifiers, if they're all protected individually, then we're fine indeed). It might not be necessary, but I thought better safe than sorry?

jvandenbroek:
That looks interesting! But would you prefer having a separate thread running all the time? Since spawning / destroying is so cheap and there are already many threads running..

notifications/NotificationBase.cpp
jvandenbroek: To clarify, I mean calls to CNotificationHelper::SendMessageEx. Locking this one up would defeat the purpose of using threads (well, not entirely, it would allow to run the notifiers simultaneously), so best alternative is locking up the spawned threads.

gizmocuz:
You got a point there 👍
I am working on merging the websocket branch and will merge this after.
Thanks

jvandenbroek:
Excellent. The websocket implementation looks very interesting, looking forward to it. Good luck :)

gizmocuz:
Thanks!

rkluwen:
Just an idea:

How about putting messages in a producer/consumer queue and push the 
notifications there.
In in a single separate thread, you can read them from there and handle 
the actual notifications.

Just my 2 cents.

== Rene


------ Origineel bericht ------
Van: "Joost van den Broek" <notifications@github.com>
Aan: "domoticz/domoticz" <domoticz@noreply.github.com>
Cc: "Subscribed" <subscribed@noreply.github.com>
Verzonden: 27-5-2017 11:18:46
Onderwerp: Re: [domoticz/domoticz] Notification use threads (#1567)

PR #1568, opened 2017-05-26
:wqIcon

DebugBill:


gizmocuz:
No idea what this pull request does ? Maybe you can edit the topic, or explain some more ?
but:
a) http logging via trace is OK, no need to actually log it to the error log as all classes using the http class will already tell what is not right. (checking the HTTP return code)
b) Not going to change the water icon, this might be possible later when we support full theming. Until then we keep the current water icon

DebugBill:
Sorry, this is a mistake. The title shows I thought I was typing in a vi console when apparently I was typing in the browser and it created a fake pull request. This is my own "personnal" branch with the modifications I like for myself.

I close it.

PR #1570, opened 2017-05-27
Rtl433: RTL SDR hardware driver using rtl_433

peahonen:
see https://github.com/merbanan/rtl_433

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -376,6 +376,7 @@ hardware/RFLinkTCP.cpp
 hardware/RFXBase.cpp
 hardware/RFXComSerial.cpp
 hardware/RFXComTCP.cpp
+hardware/Rtl433.cpp
 hardware/S0MeterBase.cpp
 hardware/S0MeterSerial.cpp
 hardware/S0MeterTCP.cpp
--- a/hardware/Rtl433.cpp
+++ b/hardware/Rtl433.cpp
@@ -1,0 +1,207 @@
+#include "stdafx.h"
+#include "hardwaretypes.h"
+#include "../main/Logger.h"
+#include "../main/RFXtrx.h"
+#include "../main/Helper.h"
+#include "../main/localtime_r.h"
+#include "../main/mainworker.h"
+#include "../main/SQLHelper.h"
+
+#include <cmath>
+#include <fcntl.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <stdio.h>
+
+#include <boost/lexical_cast.hpp>
+
+#include "Rtl433.h"
+
+Rtl433::Rtl433(const int ID) :
+	m_stoprequested(false)
+{
+	m_HwdID = ID;
+
+}
+
+Rtl433::~Rtl433()
+{
+}
+
+
+bool Rtl433::StartHardware()
+{
+	m_threadReceive = boost::shared_ptr<boost::thread>(new boost::thread(boost::bind(&Rtl433::ReceiveThread, this)));
+	m_bIsStarted=true;
+	sOnConnected(this);
+	StartHeartbeatThread();
+	return true;
+}
+
+bool Rtl433::StopHardware()
+{
+	if (m_threadReceive)
+		{
+			m_stoprequested = true;
+			m_threadReceive->join();
+		}
+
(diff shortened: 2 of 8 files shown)


gizmocuz:
Thank you for this !!

Some questions/suggestions:
- Please try to make it work on windows also (this is possible and then we have it supported on all platforms) (_popen)
- I don't think this is the right way to check if it is a temperature sensor (specially not the lexical cast... don't use that... specially not with a try/catch)
- try to support all 83 protocols... you can use the test suite for this.. https://github.com/merbanan/rtl_433_tests

gizmocuz:
Maybe also use 'json' as output format as this is supported by more protocols then csv

gizmocuz:
-q does not need a parameter
maybe add '-I 2' (only decode known protocols instead of all)

gizmocuz:
If possible (when using json), store the last json. When receiving a new json, compare it with the previous one, and if it is less than 2 seconds, ignore it
This way you do not send the new sensor multiple times (if received multiple times)

I will attach my received sensors (json format)

gizmocuz:
Sample data:

[rtl433.zip](https://github.com/domoticz/domoticz/files/1034112/rtl433.zip)

peahonen:
Thanks for your quick feedback.

Detecting types of sensors
- Current approach is a bit like ‘duck typing’. If it has temperature and humidity, it’s a temp+hum. If it has state or command, it’s a switch.
- I don’t want to couple the type detection to concrete sensor models. Hopefully the set of models will grow in rtl_433 going forward, and this way new ones will be supported automatically without touching Domoticz code.
- Not sure if all rtl_433 supported sensors even make sense in the Domoticz context. But if they do, I’d see adding more as an incremental step later on.

Lexical cast + try/catch.
- This one I’m happy to switch to match your coding style. I originally used c++11 std:sto[if], which also throws exceptions.
- std:sto[if] aren’t any prettier and caused trouble as some of the CI targets are using a C++11 compiler, whereas others seem to be using c++98. Maybe the makefiles could explicitly select either one?
- Maybe strto[if] together with checking for a completely missing value would be better? Invalid values would then be interpreted as 0’s. This would not be the case with the current implementation.

JSON/CSV
- I’d rather consider adding missing CSV support for remaining sensors in rtl_433 than switch to JSON.
- CSV is simpler to process and could be piped through rather generic line-based text filtering tools. 

Ignoring duplicates
- Instead of just ignoring duplicates, I’d rather pipe the input through a generic CSV-based median filter (See above). I was thinking of this a separate enhancement further along the road.
- As the sensors typically have very poor checksums, it’s not uncommon to occasionally receive incorrect values. This results in ugly spikes in the graphs. A simple median filter would eliminate these in a general way.

Windows
- Does this patch have adverse effects on the Windows version? If not, I suggest we consider adding Windows support worth another patch later on?
- Also, I don’t own a Windows PC, which makes it difficult to work on right now.

gizmocuz:
I think you own a pc, that should be able to run windows ;) .. virtual box, or dual boot but OK.
JSon is easier to patch, and you could detect the start/stop by looking at the { and } character.
This way, it IS actually feature proof. What happens if the CSV order is changing ?
Also when you start the rtl_433 application, it will tell you (here it does) that not all protocols support CVS ... when switching to JSON you won't see that error.

Okey, i will merge the pull request... but i would like to change some code (also make it cross platform)

gizmocuz:
Thanks!

peahonen:
The code operates on the CSV fields by names assigned on the header row, so the field ordering should not matter.

One more point I missed. The -q options doesn't really have a parameter, it's a redirection of stderr at the end. This -q is supposed to suppress extra printouts. But since there is still something printed to stderr, I thought I'd discard those using the redirection.

gizmocuz:
Made some changes:
https://github.com/domoticz/domoticz/commit/cdbdf7984802a193400d22f82aac530852cd7c94

Is it possible to verify it still works under Linux ?

I added the rain sensor, are you able to add the wind sensor ?

peahonen:
I tried faaef31. Unfortunately it's not receiving anything.

gizmocuz:
For windows, i had to remove the /dev/null (NUL), maybe for linux as well ?
Are you able to debug ?

PR #1571, opened 2017-05-28
Add Python3 event scripts and Plugin support for multiple connections

dnpwwo:


gizmocuz:
Thanks!

jvandenbroek:
The default script_time_test.py now prints out every minute in log and I get a segfault when exiting Domoticz after 'PluginSystem: Exiting work loop'.

PR #1572, opened 2017-05-28
Rtl433: fix output redirection in popen().

peahonen:
rtl_433 option -q is for quiet, but it still prints out log to stderr.
This extra logging can be suppressed by redirecting stderr to /dev/null.

commit cdbdf79 had accidentally redirected stdout instead of stderr,
resulting in no data being received.


gizmocuz:
Thanks! Doing this under windows resulted in a capture file named '2'

PR #1575, opened 2017-05-29
Opens site in new window/tab

JKetelaar:
As I'm using this function quite a lot, I sometimes got annoyed by the link opening in the same tab. This resolves this issue.


gizmocuz:
Thanks! ;)

PR #1576, opened 2017-05-29
cleanup after myself (P1 meter)

gordonb3:
- remove temporary log line that was never triggered
 - disable logging of every processed datagram (way too loud)
 - sanitize all other logging
 - add log for reporting P1 meter version


gizmocuz:
Thanks!

Issue #1577, opened 2017-05-29
EvohomeWeb.cpp: error: expected ')' before 'PRIu64' - crosscompile for powerpc


trusnak:
```
[ 44%] Building CXX object CMakeFiles/domoticz.dir/hardware/EvohomeWeb.cpp.o
/home/linker/turris-power/build_dir/target-powerpc_8540_uClibc-0.9.33.2/domoticz-3.5878/hardware/EvohomeWeb.cpp: In member function 'void CEvohomeWeb::DecodeZone(CEvohomeWeb::zone*)':
/home/linker/turris-power/build_dir/target-powerpc_8540_uClibc-0.9.33.2/domoticz-3.5878/hardware/EvohomeWeb.cpp:433:72: error: expected ')' before 'PRIu64'
    m_sql.safe_query("UPDATE DeviceStatus SET Name='%q' WHERE (ID == %" PRIu64 ")", ndevname.c_str(), DevRowIdx);
                                                                        ^
/home/linker/turris-power/build_dir/target-powerpc_8540_uClibc-0.9.33.2/domoticz-3.5878/hardware/EvohomeWeb.cpp: In member function 'void CEvohomeWeb::DecodeDHWState(CEvohomeWeb::temperatureControlSystem*)':
/home/linker/turris-power/build_dir/target-powerpc_8540_uClibc-0.9.33.2/domoticz-3.5878/hardware/EvohomeWeb.cpp:475:72: error: expected ')' before 'PRIu64'
    m_sql.safe_query("UPDATE DeviceStatus SET Name='%q' WHERE (ID == %" PRIu64 ")", ndevname.c_str(), DevRowIdx);
                                                                        ^
/home/linker/turris-power/build_dir/target-powerpc_8540_uClibc-0.9.33.2/domoticz-3.5878/hardware/EvohomeWeb.cpp:482:86: error: expected ')' before 'PRIu64'
    m_sql.safe_query("UPDATE DeviceStatus SET DeviceID='%q',Name='%q' WHERE (ID == %" PRIu64 ")", dhwdata["dhwId"].c_str(), ndevname.c_str(), DevRowIdx);
                                                                                      ^
/home/linker/turris-power/build_dir/target-powerpc_8540_uClibc-0.9.33.2/domoticz-3.5878/hardware/EvohomeWeb.cpp: In member function 'uint8_t CEvohomeWeb::GetUnit_by_ID(long unsigned int)':
[...]

PR #1580, opened 2017-05-29
reorganise includes in EvohomeWeb

gordonb3:
Having curl.h in EvohomeWeb.h was what caused the PRIu64 compile errors on certain systems

### DIFF
--- a/hardware/EvohomeWeb.cpp
+++ b/hardware/EvohomeWeb.cpp
@@ -16,13 +16,13 @@
 #include "../main/Logger.h"
 #include "../main/SQLHelper.h"
 #include "../main/mainworker.h"
-#include "../json/json.h"
 #include "hardwaretypes.h"
 #define __STDC_FORMAT_MACROS
 #include <inttypes.h>
 #include "../main/localtime_r.h"
 #include "../httpclient/HTTPClient.h"
 #include <string>
+#include <curl/curl.h>
 
 
 #define LOGONFAILTRESHOLD 3
--- a/hardware/EvohomeWeb.h
+++ b/hardware/EvohomeWeb.h
@@ -1,18 +1,15 @@
 /**
- * Place holder for Integrated Evohome client for Domoticz
+ * Json client for UK/EMEA Evohome API
+ *
+ *  Adapted for integration with Domoticz
  *
  *  Copyright 2017 - gordonb3 https://github.com/gordonb3/evohomeclient
  *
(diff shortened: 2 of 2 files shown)


gordonb3:
Please note the final comments on issue #1577 regarding a possible additional change to mainworker.cpp

gizmocuz:
Thanks! Bdw, there are more includes that can be removed from the header file.
Placing includes in a header files increases compilation time a bit

Cleaned some more in
https://github.com/domoticz/domoticz/commit/11af8034d793190befa91eb3eecb321d4f6f10e7

I hope you don't mind, but i'm going to make some small modifications (for example member variables should start with m_)

gizmocuz:
https://github.com/domoticz/domoticz/commit/e6131f2ed2cd9a635ce651e40ebb6de90642464c

gizmocuz:
Please note that on line ~269

double setpoint;

is uninitialized

gizmocuz:
Question, i see in 'DecodeZone' you are not using the mainworker... in fact in most things you are not.
This does mean we can not use notifications / scripts ?

Why did you do this, for example, it's very easy to just call SendTempSensor, SendSetpointSensor ...
SendSwitch ?

gordonb3:
> Thanks! Bdw, there are more includes that can be removed from the header file.

Possibly some more cleaning possible, even though I did already remove some entry points that are obsolete for this particular use. Thing is I wanted the library to be as versatile as possible and of course it had a different approach for handling json (json-c) and connecting to the internet (curl). At some point you simply lose track of what is there and why, but seeing what you did I know exactly why it was there and that it has no use in this implementation any more.

> Question, i see in 'DecodeZone' you are not using the mainworker...

The intention was to keep as close to the original Evohome code as possible. Which is why I split that thing up earlier into Base, Serial and Script. While linking to mainworker I discovered that the exploded time elements in the REVOBUF structs required a lot of parsing and cutting which didn't seem to make a lot of sense and figured I'd best perform the writes myself as I already had all the data in the correct format.

It's actually just DecodeZone and DecodeDHWState (which is essentially a zone as well). The other writes to the database are - aesthetic - corrections that are based on triggers from the main system. e.g. if you change the system state this will have an effect on all the zones. An additional problem here is that the Honeywell server does not return the new values when you query it directly. Which reminds me that I forgot to implement a thought that I should check and possibly set back the counter in Do_Work when a WriteToHardware is done.

PR #1581, opened 2017-05-31
Evohome + P1 + Logging

gordonb3:
Did a review on the Evohome via Web API and deleted some more unused functions from the original library. Also addressed the double that could be uninitialized in some cases.

I've added a logging selector to the hardware page that can be used by 'loud' hardware interfaces. I had disabled logging of the RXMessage processing for P1, but I figured some people might like to be able to see messages being processed. So now they can switch it on and off.

There is some temporary code in EvohomeWeb that I added for implementing the logging selector as I was running out of `Mode` parameters. I'm now combining all 0|1 state selectors as bits into one single Mode value and will be dropping the previously used separate values later on.

Last: I fixed the double datetime entry in the log.


gizmocuz:
Sorry but can you revert the P1 changes (verbosity log level) and just don't log that much ?
I do not want to present the end-user with 'another' option.. it is getting to complex and the whole intention of this software is 'keep it simple'
Maybe make a separate pull request for evohome/logger

gordonb3:
I can. You do realize I just silenced that P1 log earlier this week? You've had that 10 second logging of every incoming datagram (electricity + gas) on for as long as that hardware existed in Domoticz. It was just after I disabled that continuous stream of output that I realized that it might be a bit too quiet for some people.

Also it was in fact your own suggestion to add such an option to the hardware page. And that was related to logging something that only occurs once an hour.

gizmocuz:
Yes, i just mean, do not add an extra parameter and web setting

gordonb3:
I don't think there's another way if you want to offer such an option? At least the web setting (which is a shared div BTW).

In any case I'm already deviating from the system default that has m_bOutputLog set to true, whereas I let the hardware handler use false as the default.

gordonb3:
Hold on. I just realized I don't actually need the additional parameter (for either class). I can simply change the type of one of the existing booleans to int so I won't have to do any changes to mainworker.

gizmocuz:
Could you make a separate pull request for the evohome.

Please leave the P1 class, specially the logging, this is not the way we log in Domoticz
Also there is no need for a change.
If there is a log line to many, please create a separate pull request with the suggested change

gordonb3:
You're a hard man to talk to.

I'm not doing the logging. ProcessRXMessage in mainworker is. So yes this is the way that you log in Domoticz.

gizmocuz:
Sorry... i'm not... but thats why it is best to split pull requests
how i log in domoticz from a hardware type ABC

"ABC: this is a log message"

When the hardware is added to domoticz under the name apple-pie, this is still the log message:

"ABC: this is a log message"

I just committed a patch to revert these changes in the P1 class that slipped in

gordonb3:
Okay, so how do you distinguish between two instances of the same hardware class if both of them can spit out the exact same message?

gizmocuz:
This is for 'internal' logging, you do not connect to two smart meter from one location
You will get an additional log message when a value is received
(You also have one smart meter)
Otherwise, this is where we have master/slave for, and this is still logged on master/slave

gizmocuz:
And i want to keep Domoticz simple and foremost uniform

gordonb3:
Honest? Sounds inconsistent to me as the main logging feature in Domoticz produces messages in the form
`"(Hardware Name) Device Type (Device Name)"`
Why put `Hardware Name` in a log entry that does not provide a lot of information at all, and is controlled by a boolean value that can be set to false by the hardware class if the person that wrote it thinks that a log entry every second (or ten seconds) is a bit too much, but not allow `Hardware Name` in a log entry that is meant to inform about a problem or report something that is considered to be important?

gizmocuz:
Its not

Issue #1574, opened 2017-05-28
Pushbullet doesn't work


mu3:
Domoticz v3.7468

jvandenbroek:
Working fine here on latest build (v3.7540). Could you please try it out first on the latest beta and report back? Thank you.

mu3:
No Update Available !...
![screen shot 2017-05-29 at 16 31 05](https://cloud.githubusercontent.com/assets/661220/26551707/7d93ce52-448c-11e7-9332-e9c37f7624bd.png)
![screen shot 2017-05-29 at 16 32 42](https://cloud.githubusercontent.com/assets/661220/26551708/7d99b1be-448c-11e7-9b28-697a5c5a5857.png)

I'm on mac

jvandenbroek:
Do you get an error message in the log? Are using Test in Settings and does it say the notification has been sent?

mu3:
![monosnap pushbullet - your devices working better together 2017-05-29 18-39-29](https://cloud.githubusercontent.com/assets/661220/26555666/e3716ab0-449e-11e7-8d64-1f5d142fe585.png)
![screen shot 2017-05-29 at 18 40 53](https://cloud.githubusercontent.com/assets/661220/26555667/e39665e0-449e-11e7-8f5c-5becd8d361a5.png)
![screen shot 2017-05-29 at 18 42 46](https://cloud.githubusercontent.com/assets/661220/26555668/e3a4c6b2-449e-11e7-9059-c878cf8f46d4.png)

## Same on v3.7541

jvandenbroek:
I'd suspect this is platform related since you're on the Mac build. 

Could you set https://www.google.com under Custom HTTP/Action, leave all other fields empty and test? Look in the log for Notification sent (http). If it says Failed, I believe something is wrong with the Curl implementation. Otherwise, if it succeeds, we need to look further into this specific issue.

mu3:
![screen shot 2017-05-30 at 16 15 42](https://cloud.githubusercontent.com/assets/661220/26585021/99e666c4-4553-11e7-9fb8-8ae09439d481.png)

`2017-05-30 16:16:05.408 2017-05-30 16:16:05.408 Notification sent (http) => Success`

Which actually does the same:

![screen shot 2017-05-30 at 16 16 35](https://cloud.githubusercontent.com/assets/661220/26585052/ad30e34e-4553-11e7-88e0-d1d15a8b5298.png)

jvandenbroek:
[...]

PR #1582, opened 2017-05-31
HTTPClient added security options function / Notification set back priority to 0

jvandenbroek:
Added security options function for HTTP client, so verify peer, host and TLSv1 can be set / enforced. I think this will solve https://github.com/domoticz/domoticz/issues/1574, but even if not, this makes communication more secure. I only changed Pushover and Pushbullet, since those are the ones I can test.

### DIFF
--- a/httpclient/HTTPClient.cpp
+++ b/httpclient/HTTPClient.cpp
@@ -89,8 +89,8 @@ void HTTPClient::SetTimeout(const long timeout)
 
 void HTTPClient::SetSecurityOptions(const bool verifypeer, const bool verifyhost, const bool tlsv1)
 {
-	m_iVerifyPeer = peer;
-	m_iVerifyHost = host;
+	m_iVerifyPeer = verifypeer;
+	m_iVerifyHost = verifyhost;
 	m_iEnforceTLSv1 = tlsv1;
 }
 
--- a/httpclient/HTTPClient.cpp
+++ b/httpclient/HTTPClient.cpp
@@ -13,7 +13,6 @@
 extern std::string szUserDataFolder;
 
 bool		HTTPClient::m_bCurlGlobalInitialized = false;
-bool		HTTPClient::m_bEnforceTLSv1 = false;
 bool		HTTPClient::m_bVerifyHost = false;
 bool		HTTPClient::m_bVerifyPeer = false;
 long		HTTPClient::m_iConnectionTimeout = 10;
@@ -62,7 +61,7 @@ void HTTPClient::SetGlobalOptions(void *curlobj)
 	curl_easy_setopt(curl, CURLOPT_HTTPAUTH, CURLAUTH_BASIC | CURLAUTH_DIGEST);
 	curl_easy_setopt(curl, CURLOPT_WRITEFUNCTION, write_curl_data);
 	curl_easy_setopt(curl, CURLOPT_ACCEPT_ENCODING, "");
-	curl_easy_setopt(curl, CURLOPT_FOLLOWLOCATION, (m_bVerifyPeer ? 0 : 1L));
+	curl_easy_setopt(curl, CURLOPT_FOLLOWLOCATION, 1L);
 	curl_easy_setopt(curl, CURLOPT_USERAGENT, m_sUserAgent.c_str());
 	curl_easy_setopt(curl, CURLOPT_CONNECTTIMEOUT, m_iConnectionTimeout);
 	curl_easy_setopt(curl, CURLOPT_TIMEOUT, m_iTimeout);
@@ -72,7 +71,6 @@ void HTTPClient::SetGlobalOptions(void *curlobj)
 	std::string domocookie = szUserDataFolder + "domocookie.txt";
 	curl_easy_setopt(curl, CURLOPT_COOKIEFILE, domocookie.c_str());
 	curl_easy_setopt(curl, CURLOPT_COOKIEJAR, domocookie.c_str());
-	curl_easy_setopt(curl, CURLOPT_SSLVERSION, (m_bEnforceTLSv1 ? CURL_SSLVERSION_TLSv1 : CURL_SSLVERSION_DEFAULT));
 }
 
 //Configuration functions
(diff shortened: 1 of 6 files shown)


gordonb3:
Did you verify those changes with SSL enabled servers?

I had some issues with SSL myself after a curl upgrade some time ago. Took some fiddling but I ended up with the following code snippet:

>    if (strcmp(ca_file,"None") == 0) {
        curl_easy_setopt(curl, CURLOPT_FOLLOWLOCATION, 1L);
        curl_easy_setopt(curl, CURLOPT_SSL_VERIFYPEER, FALSE);
    } else {
        curl_easy_setopt(curl, CURLOPT_CAINFO, ca_file);
    }

In essence, you only set `FOLLOWLOCATION` to `1L` if you do not want to verify the peer. If you do have a valid certificate, which in the quoted case is a self signed certificate for which I have a distributed CA file, you should leave `FOLLOWLOCATION` at its default value.

gordonb3:
Did you verify those changes with SSL enabled servers?

I had some issues with SSL myself after a curl upgrade some time ago. Took some fiddling but I ended up with the following code snippet:

>    if (strcmp(ca_file,"None") == 0) {
        curl_easy_setopt(curl, CURLOPT_FOLLOWLOCATION, 1L);
        curl_easy_setopt(curl, CURLOPT_SSL_VERIFYPEER, FALSE);
    } else {
        curl_easy_setopt(curl, CURLOPT_CAINFO, ca_file);
    }

In essence, you only set `FOLLOWLOCATION` to `1L` if you do not want to verify the peer. If you do have a valid certificate, which in the quoted case is a self signed certificate for which I have a distributed CA file, you should leave `FOLLOWLOCATION` at its default value.

jvandenbroek:
I did not change FOLLOWLOCATION, but I don't think it does disable any verification? (That's at  least not documented). Only when calling the SetSecurityOptions function settings are changed, so nothing should break because of this.  

Would you suggest setting this to 0 when verifying the peer? That might be a good idea.

jvandenbroek:
Oops I think I just broke Github a few moments ago ;)
Now when verify peer is enabled, followlocation is disabled. This works fine with Pushbullet and Pushover, so I guess it's better to use the default value indeed.

gordonb3:
What I found was that verifying the peer would not work with followlocation enabled. I don't really know why and I didn't really care at the time because I needed a quick fix for something that used to work before (with older curl).

jvandenbroek:
I see. Well, since https://github.com/domoticz/domoticz/issues/1574 might actually be using an (much?) older version of Curl since it's for Mac, that could actually help to solve the issue :+1: 

Oh wait you're saying it's the other way around, but anyway, I think there shouldn't be any redirects when using verification, so let's keep it this way.

gizmocuz:
Maybe the mac should update curl ?

jvandenbroek:
Yes, but aren't you the one building the releases for Mac? ;) Anyway, the Pushbullet server does respond strange when trying to enforce it to SSLv3, some undefined error occurs. When testing this on the Prowl server, there is a 'handshake failure', so it looks like that's why Prowl is working for this user but Pushbullet isn't. My theory is that an older version of Curl doesn't handle the 'undefined error' correctly. And when he didn't disable peer verification on the command line, the connection was made successfully.

But besides this issue, I think it's good practice to enforce TLSv1 and verify SSL peers/hosts when possible. At some point that should become default with an option for specific hardware/functions to disable it. For now let's enable it when it's known to work with specific services.

gizmocuz:
No, travis is doing this. We might be able to change the travis script (in the domoticz repo)

gizmocuz:
I like default settings ;)

gordonb3:
Might want to note that old curl did in fact not pose the issues I mentioned. I'm also unsure if the last code version presented here is really the best way to do it. From when I was programming that bit of software I remember that it was best to leave curl options alone if you are not using them. Do not set values that you think are the default, even if they are.

gizmocuz:
this is the first time i hear about a problem here... and this class is used a lot... quite a lot ;)

jvandenbroek:
Yes, it's not sure if it has anything to do with older version. Might be platform related, but the fact is that when he executed -k option (disabling verification), the connection failed and without it, it succeeds. 

The only value now set default which isn't already specified in the current code, is the SSLVERSION. I really doubt that CURL_SSLVERSION_DEFAULT in reality isn't default.. I could remove the TLS enforcement if desired, since I believe enabling peer/host verification is more important.

Why would it be preferred to use insecure options for all calls? I can understand that custom scripts etc needs to disable peer verification, but registered notification systems? Must be missing something here.

jvandenbroek:
@gizmocuz Understood, but there isn't really anything changed when not using the new SetSecurityOptions function, it all defaults to what it currently is. So only when asked it will enable verification (and call it twice to set it back as it was). In the future it should be enabled by default, but not for now of course ;)

gizmocuz:
Yes but once it is called, from that time it will be used for everything
I really never ever experience something like this in it's 3 year usage... so very strange... so very mac related ?

gordonb3:
How did I miss that? You're right, it's a single instance. That complicates things.

jvandenbroek:
That why I call the function twice, before calling POST and after. So it would only be active during processing the notification. That's indeed a small timeframe of (potential) failure for other calling functions. But in practice?

Would you integrate this when I remove the TLSv1 enforcement to make sure that absolutely no settings are changed after setting back the default values?

jvandenbroek:
Ah interesting, while looking further into the changes, I see that 27 Feb the SSL_VERIFYHOST has been added to HTTPClient. According to the Curl docs:

DarwinSSL: If verify value is 0, then SNI is also disabled. SNI is a TLS extension that sends the hostname to the server. The server may use that information to do such things as sending back a specific certificate for the hostname, or forwarding the request to a specific origin server. Some hostnames may be inaccessible if SNI is not sent.

There we go, the cause of the issue. So allowing this to be enabled/disabled is mandatory I think?

gizmocuz:
Great! Thank you !!

PR #1585, opened 2017-06-01
GPIO remove obsolete WiringPi dependency

jvandenbroek:
It seems like my commit from some time ago removing wiringPi from CMakeLists.txt was undone.
Since both Sysfs GPIO and GPIO are using /sys/class/gpio, I just put them on the same definition.

Some small changes in GPIO:
- ~~Added signal ignore for SIGHUP, preventing interrupts being processed under special circumstances.~~
- Put the update status delay in separate thread to prevent hanging during start-up.
- Minors, code cleanup

Edit: signal got ignored in the whole app, removed it.


hvbommel:
There are also a #ifdef 's "WITH_SYSFS_GPIO" in :
- main\mainworker.cpp (see line 946)
- main\WebServer.cpp (see lines #997, #3671 and #4169) 

Can you please also replace these to "WITH_GPIO" ? Otherwise "Genercic sysfs GPIO" will be broken after the merge. Thanks.

jvandenbroek:
Yes, sorry I missed that one. Will do later, please wait before merge!

hvbommel:
It needs to be fixed at 4 places, see my previous (edited) comment.

jvandenbroek:
Done, should be fine now.

hvbommel:
Looks good to me now, thanks.

jvandenbroek:
I removed the signal interception, sighup seemed no longer be triggered app wide this way (thought it was limited to it's specific thread).

gizmocuz:
Thanks!

PR #1586, opened 2017-06-01
Renamed scripts to demo, implemented python scripts from web-editor 

moroen:
Changed the names of the device and time scripts so they no longer are executed by default and added editing and running python-scripts from the web-based events-editor.


gizmocuz:
Thanks!

PR #1587, opened 2017-06-01
Evo web revision 2

gordonb3:
Cleanup + some handling improvements:

- general cleanup
- web portal is terribly slow. If you request data just after a write it will return the old data 
- script based Evohome attaches action scripts to the evohome devices. these need to be deleted to prevent a double action
- use bit oriented operations instead of Math functions for speed
- possible cause for Fitbit problem?


hardware/EvohomeWeb.cpp
```diff
@@ -38,21 +39,45 @@ const std::string CEvohomeWeb::weekdays[7] = { "Sunday", "Monday", "Tuesday", "W
...
+	 *
+	 * 0x1 = m_updatedev (let Honeywell server control the device name)
+	 * 0x2 = m_showschedule (show next scheduled switch point as `until` time)
+	 * 0x4 = m_showlocation (prefix device name with location)
+	 *
+	 */
+
+	m_updatedev = ((UseFlags & 1) == 0); // reverted value: default = true
+
+	if ((UseFlags & 15) < 2) // compatibility - load from old parameters
```
gizmocuz: Maybe it is better to do this in SQLHelper.cpp ? (database version ?)
This will keep this code nice and clean ;)

gordonb3:
Increment database version you mean? Not a bad idea at all. Would also allow deleting entries `WHERE Name LIKE 'Fit%'`

gizmocuz:
True :) or delete all entries with the old Id 
Sure this will cause that current evohome users loose their stuff... But that's the beta effect

hardware/EvohomeWeb.cpp
```diff
@@ -149,6 +149,8 @@ bool CEvohomeWeb::StartSession()
 
 bool CEvohomeWeb::StartHardware()
 {
+	if (this->Name.substr(0, 3)=="Fit") // attempt to block old FitBit hardware
```
gizmocuz: Maybe we could find the root cause of the problem inside the evohomeweb class ?
Maybe use the stable domoticz version, add the fitbit, update to the latest beta ?

gizmocuz:
Fixed the crash in this commit
https://github.com/domoticz/domoticz/commit/749fa536dbdb25f3818021b8a9c8d6d6f3f0bf78

the problem was related to 'm_refreshrate' in the constructor, it was zero.. this is now handled.
No need to check for names starting with 'FIT'

gordonb3:
Ow! Division by zero. Good catch!

gizmocuz:
Thanks!

PR #1589, opened 2017-06-01
Add Onkyo AV receiver

dwmw2:
I keep toying with adding new subdevices (there are plenty there), but these are the important ones. And getting the basic framework merged would mean the endless merge conflicts go away...


gizmocuz:
Thanks !! But could you add this also to the visual studio project file so it compiles ?
You can see our automatic build system has a problem with it.
If possible, use visual studio 2017 to debug/test as well (yes, linux users can run windows too) (and visa versa)

gizmocuz:
Thanks!

PR #1591, opened 2017-06-02
Scheduler fix Blinds percentage Off level

jvandenbroek:
When schedule blinds percentage to off, it would set the level to 100. Fixed to set it to 0 as it would be when manually switching off.


gizmocuz:
Okey, but on OFF should send OFF, regardless of it's value
When you send 'ON' it should use the last level... is that still the case ? (also in the gui)

ON, level = 80%
OFF
ON

Is the last on still at 80% ?

This is maybe more for lights and not for dimmers... but it might be that this patch will cause this problem for lights/dimmers

jvandenbroek:
No when sending On, it sets the level to 0.. I worked around that with a user variable to store the previous level, not ideal. I could change it so only blinds are affected?

I would think it's better to send 0 when Off, also looks better in GUI. Wouldn't that be also the case for dimmers?

Edit: Now I'm thinking about it, what you said makes sense. Should just remember the last 
level, but since that isn't the case, I thought to do it this way. Switching off manually it does set to 0 without this change, guess that's also not correct.

jvandenbroek:
I've tested some more, but it seems there was a reason why I did it this way ;) When changing the blinds level, it will always trigger on Closed in Lua. So I can't just check for the Off / Closed command, that's why combining it with level (both level == 0 and devicechanged == closed) will work as intended, but without this patch only on manually switching off and not with a timer.

Does this logic make sense? Or should it really only trigger on Closed when actually turning off? I could take a look at this logic, but I'd be careful not breaking stuff.. What do you think?

gizmocuz:
For a dimmer (maybe you can test this) it keeps the last dim position (let's say 70%) when switching get off
The icon is off
You know that when you turn it on that it will go to 70%

jvandenbroek:
You're right, the dimmer does keep the previous value in lastlevel field. For some reason this was excluded for blinds percentage, removed this. Not sure why this was, can't think of any reason? If anybody want to fully close their blinds, they should just set level 100, right?

~~Also added otherdevices_lastlevel, since I need this value in my LUA script for obvious reasons. I'm using a virtual device btw.~~

~~I accidentally also pushed an update for allowing to (un)set protection on IDX devices, wanted do that in another PR. If you want me to remove that for this PR, let me know.~~

The lua addition was a little bit to easy, need to dig into it further ;) Will redo this later.

gizmocuz:
Tell me when you want to have something merged

jvandenbroek:
OK, I redid the logic:

- Set level > 0 changes icon to Closed (Open when inverted)
- Lua now only reacts on devicechanged open/closed, not on set level
- Switching off no longer resets lastlevel
- Added lua otherdevices_lastlevel, need to use it instead of svalues when device was switched off

The nice thing about how it's now implemented, is that you now can just set a level and afterwards send an On command (since there are separate on/off icons). That way it's easier to precisely set the wanted value before executing. You could of course also just script a direct On command after set level has changed, so I think this makes it really customizable.

gizmocuz:
Great, thanks !!

RicardP:
This update messed up my Qubino relays on 230VAC Somfy Blinds... it was no problems before (except that I had to invert up/down powering wires on the relay to get percentage scale showing correct, winded up 0%).

By the way.. Open and Closed is bad description and can mean anything, or what is Open? Window to sun or blind rolled out...? Would it not be better saying i.e. Up (always fully winded in)  / Down xx% (percentage value after Down).

Now by latest change percentage scale does not change if pushing Open / Close buttons.  
  
skarmklipp
 
Cheers!

jvandenbroek:
In what way did it mess up? I deliberately changed the way the set level is handled, because I believe the last set level should be used when closing your blinds (it's the same logic as a dimmer uses). Or doesn't that work and will it just fully close? If that's the case, I think we indeed should reconsider this logic. 

If you want the Open / Closed status the other way around, you could use blinds percentage inverted.

RicardP:
The percentage scale seems wont follow if you push Close button, the function, remember last position is nice though. So more than expected did in fact change but I had no time to pinpoint different actions and results :-)

Yes I know about the Blinds Inverted but it did not invert the Percentage scale so it made no sense, I had to instead invert the wires to get it right.

jvandenbroek:
Yes I think we need to send the set level instead of an On command when pushing close button. I'm reworking this part of the code. 

Could you post the device details, so I know which hardware we're working on? Just paste the contents of http://domoticzip:8080/json.htm?type=devices&rid=IDX (IDX = your blinds device)

RicardP:
Device is:
Qubino ZMNHCDx Flush Shutter	Qubino	ID: 0x0052	Type: 0x0003

Could you fix the Percentage scale, in Blinds Inverted, so it gets inverted as well?

And, Up / Down xx% instead of Open / Close terms?

Thanks for your efforts! :)

jvandenbroek:
Open / Close is the standard logic for a long time.. And I believe that's correct for horizontal sun screens up and down wouldn't make much sense. Close and open apply to both horizontal and vertical screens (or whatever they're called :))

The inverted option does not invert any level sent, and it shouldn't. Your device is the one responsible for interpreting the value, so inverting your wiring seems to me the correct way to solve it. 

Could you give me the full contents of http://domoticzip:8080/json.htm?type=devices&rid=IDX (IDX = your blinds device)? It helps me to determine which code exactly needs to be modified. Thanks!

RicardP:
Sure if having horizontal would not be right but probably 99 percent of all screenings are vertical movement? And I believe if anyway having a horizontal moving would still "in mind" be a Up or Down in sun screening function point of view :)

Ok then if the Qubino relay delivers the precentage value, strange that Qubino thinks 100% is fully winded in... will check it and come back to you about it.

Sorry but I do not get the values back, regardless what IDX I write I get the same (wrong) device data. My device Node ID from zwave devices list is: 024 (0x18). Maybe I am missunderstanding something? :)

Thanks!

jvandenbroek:
You can find the correct IDX under Setup -> Devices, first column. 
99%? Don't be so sure about that ;) Here in the Netherlands many people use this kind of (outdoor) sunscreen. Besides if it's better to use Up / Down, all scripting is based on Open / Closed, so that wouldn't be easy to change. Also, I'm not the one who decides what's better, that should be discussed elsewhere.

RicardP:
I get this answer:

ActTime	1496846330
ServerTime	"2017-06-07 16:38:50"
Sunrise	"04:10"
Sunset	"21:48"
result	
0	
AddjMulti	1
AddjMulti2	1
AddjValue	0
AddjValue2	0
BatteryLevel	255
CustomImage	0
Data	"Open"
Description	""
Favorite	1
HardwareID	11
HardwareName	"Z-Stick"
HardwareType	"OpenZWave USB"
HardwareTypeVal	21
HaveDimmer	true
HaveGroupCmd	true
HaveTimeout	false
ID	"00001901"
Image	"Light"
IsSubDevice	false
LastUpdate	"2017-06-07 16:17:06"
Level	1
LevelInt	1
MaxDimLevel	100
Name	"Markis Sovrum N"
Notifications	"false"
PlanID	"4"
PlanIDs	
0	4
Protected	false
ShowNotifications	true
SignalLevel	"-"
Status	"Open"
StrParam1	""
StrParam2	""
SubType	"Switch"
SwitchType	"Blinds Percentage"
SwitchTypeVal	13
Timers	"true"
Type	"Light/Switch"
TypeImg	"blinds"
Unit	1
Used	1
UsedByCamera	false
XOffset	"1453"
YOffset	"54"
idx	"183"
status	"OK"
title	"Devices"

jvandenbroek:
https://github.com/domoticz/domoticz/pull/1603

PR #1593, opened 2017-06-03
Sysfs GPIO: interrupt based input support

hvbommel:
To fully utilize the Sysfs GPIO system interrupts are now also supported. State changes of GPIO pins that have "edge" set to "rising", "falling" or "both" will be handled interrupt based. GPIO pins that have "edge" set to "none" or when "edge" is not present will be polled. There is nothing that needs to be configured within Domoticz to enable polling or interrupt based handling, it is all determined by the external sysfs gpio settings. Only one additional thread is used to handle all gpio pin interrupts to keep overhead as small as possible. When interrupts are not used the thread is not created. Polling or interrupt based handling can be set on a per pin basis and can thus be mixed. Only when at least one input pin requires interrupt based handling the additional interrupt handling thread is started.

Summary:
- Polling mode: gpio pin has "edge" set to "none" (or when "edge" not present).
- Interrupt mode: gpio pin has "edge" set to "rising", "falling" or "both"
- When all pins have "edge" set to "none" Generic sysfs GPIO will run exactly as before (i.e. in polling mode)
- The recommended "edge" setting for sysfs gpio input pins is "none".

### DIFF
--- a/hardware/SysfsGpio.cpp
+++ b/hardware/SysfsGpio.cpp
@@ -104,6 +104,8 @@
 	for "rising" and "falling" edges will follow this
 	setting.
 
+	3-june-2017	HvB Add interrupt support for edge = rising, falling or both.
+
 */
 
 #include "stdafx.h"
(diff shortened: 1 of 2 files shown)


gizmocuz:
Thanks!

PR #1594, opened 2017-06-03
Raspberry's GPIO port: Fix forum reported bug for active low outputs

hvbommel:
A bug for Raspberry's GPIO port was reported on the forum on june-3 by AndyBailey  labeled "Active Low Relay Board- behaviour change after recent linux upgrades?". I have reproduced and fixed the problem in this PR.


gizmocuz:
Thanks!

PR #1595, opened 2017-06-04
Resolve seg fault on exit

dnpwwo:


gizmocuz:
Thanks!

hardware/plugins/Plugins.cpp
```diff
@@ -711,10 +711,13 @@ namespace Plugins {
 			// Check all connections are still valid, vector could be affected by a disconnect on another thread
 			try
 			{
-				for (std::vector<CPluginTransport*>::iterator itt = m_Transports.begin(); itt != m_Transports.end(); itt++)
+				if (m_Transports.size())
```
gizmocuz, line 714: i would personally write

`if (!m_Transports.empty())`

PR #1596, opened 2017-06-04
SQLHelper: lessen rounding errors for computed energy

koenkooi:
The power sensors I have (Ubiquiti mFi) don't have a total Energy
counter, only a monthly one, so I used the 'computed energy' features in
domoticz.
The problem is that for low power devices the %.0f format specifier
introduces big rounding errors:

* For a device that alternates between 14 and 17 Watt the computed
  energy is between 19 and 27 Watthour:
overreported

* For a device that alternates between 10 and 15 Watt the computed
  energy is 0 and 2 Watthour:
underreported

After this patch (about 23:00 on june 3rd in the above screenshots) the first devices gets 14-15Wh and the second device
12-13Wh.


gizmocuz:
Thanks

PR #1599, opened 2017-06-05
Switch action set delay

jvandenbroek:
Simple patch for allowing to set a delay for scripts / http actions on switches.
Eg. setting http://www.someurl.com|2.5| will delay executing with 2.5 seconds. Setting value to 0 will immediately execute it (which is why I actually implemented it). If not specified, the standard delay of 0.2 sec will apply. You could also use http://www.some|2.5|url.com, however that doesn't make much sense ;) But it can be placed anywhere after http(s):// and script://, since it will be stripped before executing.

If anything other than numbers are specified between 2 pipes, it will be processed as before (so scripts using pipes won't be influenced by this).

### DIFF
--- a/main/SQLHelper.cpp
+++ b/main/SQLHelper.cpp
@@ -6606,9 +6606,6 @@ bool CSQLHelper::HandleOnOffAction(const bool bIsOn, const std::string &OnAction
 					std::stringstream ss;
 					ss << szTmp;
 					ss >> delay;
-				}
-				if (delay != -1)
-				{
 					szTmp = Action.substr(0, firstpipe - 1);
 					szTmp.append(Action.substr(secondpipe + 1));
 					Action = szTmp;
--- a/main/SQLHelper.cpp
+++ b/main/SQLHelper.cpp
@@ -6574,58 +6574,64 @@ bool CSQLHelper::HandleOnOffAction(const bool bIsOn, const std::string &OnAction
 			_log.Log(LOG_TRACE, "SQLH HandleOnOffAction: OffAction:%s", OffAction.c_str());
 	}
 
+	std::string Action;
+
 	if (bIsOn)
 	{
 		if (OnAction.empty())
-			return true;
-
-		if ((OnAction.find("http://") != std::string::npos) || (OnAction.find("https://") != std::string::npos))
-		{
-			AddTaskItem(_tTaskItem::GetHTTPPage(0.2f, OnAction, "SwitchActionOn"));
-		}
-		else if (OnAction.find("script://")!=std::string::npos)
-		{
-			//Execute possible script
-			std::string scriptname = OnAction.substr(9);
-#if !defined WIN32
-			if (scriptname.find("/") != 0)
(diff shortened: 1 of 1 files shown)


main/SQLHelper.cpp
```diff
@@ -6574,58 +6574,61 @@ bool CSQLHelper::HandleOnOffAction(const bool bIsOn, const std::string &OnAction
 			_log.Log(LOG_TRACE, "SQLH HandleOnOffAction: OffAction:%s", OffAction.c_str());
 	}
 
+	std::string Action;
+
 	if (bIsOn)
 	{
 		if (OnAction.empty())
-			return true;
+			return false;
```
gizmocuz, line 6582: Why should this return false ?

main/SQLHelper.cpp
```diff
@@ -6574,58 +6574,61 @@ bool CSQLHelper::HandleOnOffAction(const bool bIsOn, const std::string &OnAction
...
+	}
 
-		if ((OnAction.find("http://") != std::string::npos) || (OnAction.find("https://") != std::string::npos))
-		{
-			AddTaskItem(_tTaskItem::GetHTTPPage(0.2f, OnAction, "SwitchActionOn"));
-		}
-		else if (OnAction.find("script://")!=std::string::npos)
+	float delay = -1;
+
+	size_t firstpipe = Action.find("|");
```
gizmocuz: Whats with the pipe stuff ? There is no action with pipe's ? http/script ?

gizmocuz:
Can't you use a script for this ?

main/SQLHelper.cpp
jvandenbroek: Oops, needs to be true, will fix.

gizmocuz:
You could also use blockly for this...

jvandenbroek:
Hmm, well I've considered using a script, but setting it this way is easier.. Using a script will introduce at least 300ms delay on my RPi, for that reason I'd like to get an URL called immediately.

gizmocuz:
Yes but it will add sluggish code... how about blockly ? very simple... if switch is activated.. do something in 2.5 seconds ? thats why we have blockly ;)

jvandenbroek:
Yes, well, I can understand your reasoning ;) I just use 0 sec. With this code I get just a few ms delay when calling my doorbell script ;) Blocky / lua is somehow very slow...

Or could we just remove the extra introduced delay of 0.2 sec? But I thought it was there for a reason..

gizmocuz:
blockly should be fast ? you won't notice 0.2 seconds, or 0.5 seconds... fast enough

jvandenbroek:
Well it's not that important, just thought it was handy to do it this way ;) Let's close it..
I guess you're running Domoticz on a much faster system, aren't you? ;)

gizmocuz:
I use a odroid c2 and a raspberry pi 3.... those are quite fast indeed

jvandenbroek:
Should probably upgrade to one of those some day ;) But going offtopic, close me please :P

gizmocuz:
Thanks ;)

PR #1602, opened 2017-06-08
EvohomeWeb additional code polishing

gordonb3:
- Fix a possible crash of Domoticz on an attempt to convert an empty string from localtime to UTC
- Also some improvements to how zones display their status and respond to system mode switches that are initiated through Domoticz


gizmocuz:
Thanks!

PR #1603, opened 2017-06-08
Blinds percentage use lastlevel when closing

jvandenbroek:
Changed blinds percentage logic again after input from https://github.com/domoticz/domoticz/pull/1591:

- Now sending 'Close' it actually sends the "Set Level" command if last level was > 0 && < 100 
- When changing level to 0 (Off) or 100 (On) the commands Off and On respectively will be send instead of "Set Level"

This means that the status won't show 'Closed' unless closed for 100%, I believe this is correct. Hopefully both https://github.com/domoticz/domoticz/pull/1591 and http://www.domoticz.com/forum/viewtopic.php?f=6&t=17654 will be solved now.

### DIFF
--- a/main/SQLHelper.cpp
+++ b/main/SQLHelper.cpp
@@ -3571,8 +3571,15 @@ uint64_t CSQLHelper::UpdateValueInt(const int HardwareID, const char* ID, const 
 			bool bIsLightSwitchOn=IsLightSwitchOn(lstatus);
 			std::string slevel = sd[6];
 
-			if ((bIsLightSwitchOn) && (llevel != 0) && (llevel != 255))
+			if ((bIsLightSwitchOn) && (llevel != 0) && (llevel != 255) ||
+				(switchtype == STYPE_BlindsPercentage) || (switchtype == STYPE_BlindsPercentageInverted))
 			{
+				if (((switchtype == STYPE_BlindsPercentage) ||
+					(switchtype == STYPE_BlindsPercentageInverted)) &&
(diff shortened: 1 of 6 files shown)


jvandenbroek:
Someone objects to the new behavior and I do think we need to discuss a little about what would be the best behavior before merging. I think it's better and more predictable to always use the last set level. With virtual devices / scripting you can always force to fully close the blinds (set level 100 = send On command) despite the previous level. 

See https://www.domoticz.com/forum/viewtopic.php?p=137617#p137617 for comments.

gizmocuz:
Could you undo all code formatting ? I can not see what has changed now
Should there also not be a change in the dashboard code ?

gizmocuz:
If possible, please make it so Open/Close buttons will always do 0/100%

jvandenbroek:
I've reverted much:
- Undo formatting
- Close will set level 100 and send the On command 
- Open will set level to 0 and send the Off command
- Setting level > 0 < 100 will set the close icon (open when inverted)
- Scheduler sends level + 1, this gets subtracted in function SwitchLightInt (scheduling was always setting 1 percent lower)
- Updated dashboard and lightscontroller, simplified logic

Unfortunately now we can't check for the last level setting, so almost everything is as it was before, with some small additions and fixes. I'll just use a uservariable for setting/getting the last level...

BTW I think the + 1 level should also be used for dimmers, but I only wanted to change the blinds logic with this PR.

gizmocuz:
Thanks!

PR #1604, opened 2017-06-08
Fix timers if location is not set in preferences

blino:
Timers were never scheduled when location was not set in preferences.

At start, timers are loaded from MainWorker::GetSunSettings().
It uses CScheduler::SetSunRiseSetTimers() which calls
CScheduler::ReloadSchedules() after getting sunrise/sunset times from
location.

When the location is invalid, ReloadSchedules() is called as fallback
early in GetSunSettings().
Though, it was never called when location was missing from settings.

### DIFF
--- a/main/mainworker.cpp
+++ b/main/mainworker.cpp
@@ -484,10 +484,9 @@ bool MainWorker::GetSunSettings()
 {
 	int nValue;
 	std::string sValue;
-	if (!m_sql.GetPreferencesVar("Location",nValue,sValue))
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thanks! But i would advice to always specify your location (also needed for the sun set/rise)

blino:
Yes, I will, even if I do not use sun set/rise timers.
I wanted to make sure that no one runs into this issue again.
Thanks for the fast feedback!

PR #1609, opened 2017-06-10
Bug fix and instance name added to mails

DebugBill:
Fix bug where watchdog would trigger when updates on Teleinfo are too slow because of very low consumption.

Add instance name to error messages sent out by mail


gizmocuz:
Would it be better to always start the log with 'Domoticz'
"Domoticz " + sTitle

gizmocuz:
Implemented the "10 second grace period into the Teleinfo class)
How about to always start the email with "Domoticz" (If the instance name is different than the default 'Domoticz' of course ?

gizmocuz:
No reaction , closing pull request

PR #1610, opened 2017-06-10
Xiaomi Gateway - adding smoke and gas devices

corbinmunce:


gizmocuz:
Thanks!

jejedelbarro:
Hi, are you able to trigger the siren of the smoke detector through domoticz ? i am not able to, i have an error message.

gizmocuz:
@jejedelbarro , please define 'error message'

jejedelbarro:
It says : error sending switch command, check device/harware. All my other devices works fine in domoticz, and my fire detector works fine in mi-home.

jejedelbarro:
@gizmocuz, does it works for you ?

gizmocuz:
@jejedelbarro , no as i do not own this hardware. But i requested the error message for possible developers who are able to take a look at this

jejedelbarro:
@gizmocuz : any update about the issue ?

gizmocuz:
@jejedelbarro , nope... what hardware do you own ? I have a zwave sirene that does work
Maybe it is only 1-way ? You get a trigger when there is an alarm, and you can not trigger the alarm yourself ?
Did you try it from the control panel ?

jejedelbarro:
@gizmocuz : i have a raspberry with a xiaomi gateway v2 et the honeywell smoke detector. You are certainly right, it’s maybe a one way device. I can’t realy trigger the alarm through the xiaomi app, only request for a self test of the device that trigger a short « bip ». Same probleme from the control panel. To be more specific i get this in the log when i try to trigger the alarm : « error : xiaomigateway: unable to write command - Unsupported device » then « xiaomigateway: sendmessagetogateway failed on first attempt, will try again » then same error again.

gizmocuz:
The log message "unable to write command - Unsupported device" is received from the gateway itself.
But it would have been nice to be able to do the self test.

jejedelbarro:
i try a self test from the xiaomi app, it does nothing from domoticz point of view ( no log, no device status change). I try to trigger the smoke detector with some real smoke and it works fine from domoticz side ( log ok, device status change ok). I think you are right, it’s a one way device.

PR #1611, opened 2017-06-11
Added missing "Locked / Unlocked" in event block editor 

Eorciu:
The "Door lock" switch type has a value of "Locked / Unlocked" that was missing in the event block editor. 
I added them below value "Closed"
image


gizmocuz:
Thanks! Can you also update the appcache for? (Version number)
This is needed when you make changes to the web fikes

Eorciu:
Im new in git. I do not know if I did it well. On my disk right now i have version # ref 1393 but   here its reverted ?. Next time i will remember about this file.

gizmocuz:
Thanks!

PR #1612, opened 2017-06-12
Add mutex to protect against startup race conditions,

dnpwwo:
Also add initial support for HTTP requests GET & POST


gizmocuz:
Thanks!

PR #1613, opened 2017-06-12
Evo web revision 4

gordonb3:
Needed to round this up because there was a mistake in a previous commit that prevents detection of the evohome installation when the hardware was just added.

I've received a user request to reference a second API (the 'old' US API) for fetching actual temperatures measured by the devices. While I do see the value in that there is also a performance hit related to this and I decided the default should stay like the UK/EMEA API presents it (which is half degree precision).


gizmocuz:
Thanks!!

PR #1616, opened 2017-06-13
Use TLS in update scripts.

ultrara1n:
Not using HTTPS will cause the risk of MITM (download faulty package within a compromised network) attacks.


gizmocuz:
Thanks, but please (re) create two different pull request for this.
This one can also not be merged because of recent changes.
Best to leave the ident changes for what it is until the guiwork branch has been merged

PR #1619, opened 2017-06-13
Correct until times no longer showing for overrides after #5c3417d

gordonb3:
Moving away from the string-string maps didn't work out too well for this property.


gizmocuz:
Thanks!

PR #1622, opened 2017-06-15
DarkSky reduce interval from 10 to 5 min / exclude unused data from response

jvandenbroek:
DarkSky allows 1000 requests / day for free, Weather Underground's on the contrary only 500 / day. Both are using an interval of 10 minutes, so I think it's appropriate to reduce the DarkSky interval to 5 min.

Also excluded unused data to reduce request response.


jvandenbroek:
Fixed UV Index and removed some obsolete (leftover WU) rain code.

gizmocuz:
Thanks!

PR #1626, opened 2017-06-16
Pushover sound in Custom Message field

jvandenbroek:
Allow to set a priority and sound in the Custom Message field, like this:
~~`Here some text#2#pushover`~~
`Here some text#pushover`
Where ~~2 is the priority and~~ pushover is the sound.

Edit: Not sure why I included priority, you can already set this :P Will change it to only sound.

### DIFF
--- a/notifications/NotificationPushover.cpp
+++ b/notifications/NotificationPushover.cpp
@@ -1,7 +1,9 @@
 #include "stdafx.h"
 #include "NotificationPushover.h"
 #include "../httpclient/HTTPClient.h"
+#include "../httpclient/UrlEncode.h"
 #include "../main/Logger.h"
+#include "../main/Helper.h"
 
 CNotificationPushover::CNotificationPushover() : CNotificationBase(std::string("pushover"), OPTIONS_URL_SUBJECT | OPTIONS_URL_BODY | OPTIONS_URL_PARAMS)
 {
@@ -25,12 +27,24 @@ bool CNotificationPushover::SendMessageImplementation(
 	const bool bFromNotification)
 {
 	std::string cSubject = (Subject == Text) ? "Domoticz" : Subject;
+	std::string cText = Text;
+	std::string cSound = Sound;
+	int cPriority = Priority;
 
 	bool bRet;
 	std::string sResult;
 	std::stringstream sPostData;
 
-	sPostData << "token=" << _apikey << "&user=" << _apiuser << "&priority=" << Priority << "&title=" << cSubject << "&message=" << Text;
+	std::vector<std::string> splitresults;
+	StringSplit(CURLEncode::URLDecode(Text), "#", splitresults);
+	if (splitresults.size() > 2)
+	{
+		cText = splitresults[0];
+		cPriority = atoi(splitresults[1].c_str());
+		cSound = splitresults[2];
+	}
+
+	sPostData << "token=" << _apikey << "&user=" << _apiuser << "&priority=" << cPriority << "&title=" << cSubject << "&message=" << cText;
 
 	size_t posDevice = ExtraData.find("|Device=");
 	if (posDevice != std::string::npos) {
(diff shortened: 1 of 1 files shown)


gizmocuz:
Could you make a blockly with this ? This should already work

image

When this was created, this was already done for pushover, so this functionality should already be in Domoticz somewhere

jvandenbroek:
Sometimes I over complicate things I guess ;) This seems easy to use, but I've just tested and checked the code and it introduces an extra delay of 1 second. Do we still need that extra second now that the notification system uses separate threads? Otherwise I'll change that and make do a new PR. Could also make an extra split field for LUA to make this a custom delay setting, but I have the feeling you're not really a fan of that kind of adjustments ;)

jvandenbroek:
Offtopic: before I create a new pull request later, I wanted your opinion first about my changes for the event system: https://github.com/domoticz/domoticz/compare/master...jvandenbroek:lua-thread2

This reduces the latency and improves processing speed considerably on my RPi 1, all my switches now respond much faster (more instantaneously) and LUA scripts are executing in a much more predictable speed. 

If you're using eg. script_device_devicename.lua it will only match on that device named 'devicename' (or 'DeviceName', case insensitive). However, script_device_something_else.lua will match on all devices as it do now (as long as there is no device named 'Something Else').

First I also used a queue worker, which also worked fine, but this seemed simpler and since I used some shared mutex locks, I think the speed improvement will be even more noticeable on multi-core systems. So far it's running fine here for a few days with 8 device lua scripts and 4 time scripts, no races whatsoever, just lower load and faster responses ;)

gizmocuz:
There should be no delay of 1 second... why is that ?
Even so, a second for a third party alert service is nothing (and even that online service takes time)

gizmocuz:
For the mutex in your other pull request, for now, use a boost mutex, and the boost way to lock/unlock, that is use the boost::lock_guard<boost::mutex> l(m_mutex);
This way you never have to worry about lock/unlock
We are going more and more to Python/DzVents instead of Lua... 
Nor sure about the device naming in the scripts, i would stick to the true devicename... not some_deviceame_part in a filename

jvandenbroek:
unique_lock/shared_lock is actually also scope based, with the advantage that with shared_lock multiple readers are allowed (almost everything is reading here). The unlocking is to allow writers (unique_lock) to do their work in between. But no need to unlock after leaving the scope. So far the speed improvement is really good, especially with a fairly complex script I use (which is heavily dependent on correct timing). 

Maybe I didn't explain it correctly, but it's actually matching on script_device_devicename.lua. But spaces in the name are converted to _ and it's all lower cased, so DeviceName / Devicename will both match script_device_devicename.lua. But if you would name it something different, like script_device_thisnamedoesnotexist.lua, it will just behave like it does now, matching on ALL devices. That still gives you the flexibility to use one large script if wanted and most likely won't break any users correct running scripts. The logic inside the scripts hasn't changed and still just match on devicechanged/otherdevices etc. 

I understand dzVents / Python, but I'm learning more about Lua, and actually starting to like it :P I also do believe that Python has (even) more overhead, at least memory wise. 

BTW today I also added AddVariable for LUA. This allows to do something like this (making sharing scripts easier):
```
local variable = uservariables['SomeVariable']
if (variable == nil) then
commandArray['AddVariable:SomeVariable'] = tostring("Float:12.12")}
else
commandArray['Variable:SomeVariable'] = tostring("12.12")}
end
```
You can use all the variable types instead of Float. Also no need to worry about the uservariables mutex in my new code, this all works smoothly without lockups ;)

jvandenbroek:
Yes I'm also not sure why that is. Probably to prevent stalling other queued items? But that's no longer the case now, so I'll just remove the delay. My average delay to Pushover is about 700ms, so this would add 1 additional sec to it. Indeed no real deal breaker, but unnecessary. Especially when I'm on the toilet and can't always hear the doorbell ring, it's an additional second you have to run like crazy to the door before the postman is gone :P

gizmocuz:
Thanks, but if you add a blockly like above, it is generating lua in the database... there already should be something with 'sound' in here... ? Maybe i am overlooking something

jvandenbroek:
No I'm sorry, the Lua part above is just something else, I made this thread too confusing.. Will close it now and make new PR's :)

PR #1627, opened 2017-06-16
Support the transmitting of LWRF Group commands

sycar:
Support the transmitting of LWRF Group commands as discussed here: http://www.domoticz.com/forum/viewtopic.php?f=4&t=17823


gizmocuz:
Thanks!

PR #1629, opened 2017-06-18
add virtual sensor that measures weight

ixarka:
I made a couple small changes to support virtual sensors of type weight.  I plan on using this with a script I am writing to push updates from my digital scale into domoticz.  Creating a virtual sensor seemed like the cleanest way to accomplish this.


gizmocuz:
Thanks!. Since you updated some web files as well, could you increase the version number in the appcache file to ?

ixarka:
Done.

gizmocuz:
Thanks!

PR #1630, opened 2017-06-19
support for setting units on weight

larber:
Adds ability to set units (either kg or lb) for display of weight data (via the Meters/Counters page).  This affects display only -- all raw data is still assumed to be stored in kg.


gizmocuz:
Thanks!!

PR #1631, opened 2017-06-19
EventSystem remove 1 sec delay from SendNotification

jvandenbroek:
Remove unnecessary delay from sending notification with Lua / Blockly.


gizmocuz:
Thanks!

PR #1633, opened 2017-06-19
EventSystem make threaded / match Lua and Python file name on devicename 

jvandenbroek:
- Threading the EvaluateEvent function prevents Domoticz to stall on Lua / Blockly / ~~Python~~ processing, makes multi-core processing very efficient. 

- Matching device name on file name. E.g. script_device_devicename.lua (or .py for Python) will only be executed when DeviceName is processed (case insensitive, DeviceName is lower cased and spaces are replaced with underscores). On the other hand script_device_somenamethatdoesntexist.lua will just be matched on ALL devices, just as the current behavior. The logic inside the scripts (devicechanged / otherdevices etc) hasn't changed. Executing Lua is a costly process, so this saves a lot (system load reduced considerably).

- Added AddVariable for Lua, which allows to do something like this in your scripts:
```
local variable = uservariables['SomeVariable']
if (variable == nil) then
commandArray['AddVariable:SomeVariable'] = tostring("Float:12.12")}
else
commandArray['Variable:SomeVariable'] = tostring("12.12")}
end
```

By implementing the threaded / file match concept with unique_lock / shared_lock mutexes, multiple scripts can run fast simultaneously when reading, without the risk of lockups/races. Timing has become much more predictable now. On my RPi 1, all my switches now respond almost instantaneously, despite the fairly high number of Lua scripts I have. I can only imagine the latency reduction on multi-core systems ;)

### DIFF
--- a/main/EventSystem.cpp
+++ b/main/EventSystem.cpp
@@ -413,8 +413,6 @@ void CEventSystem::GetCurrentMeasurementStates()
 		WaterDivider = float(tValue);
 	}
 
-	boost::shared_lock<boost::shared_mutex> devicestatesMutexLock(m_devicestatesMutex);
-
 	//char szTmp[300];
 
 	typedef std::map<uint64_t, _tDeviceStatus>::iterator it_type;
@@ -1121,8 +1119,9 @@ void CEventSystem::EvaluateEvent(const std::string &reason, const uint64_t Devic
 		return;
 	GetCurrentUserVariables();
 
-	//boost::unique_lock<boost::shared_mutex> uservariablesMutexLock(m_uservariablesMutex);
+	EvaluateBlockly(reason, DeviceID, devname, nValue, sValue, nValueWording, varId);
 
+	boost::shared_lock<boost::shared_mutex> devicestatesMutexLock(m_devicestatesMutex);
 	std::vector<std::string> FileEntries;
 	std::vector<std::string>::const_iterator itt;
 	std::string filename;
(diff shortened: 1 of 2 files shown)


jvandenbroek:
Btw If you don't want these changes to apply for Python, I can just leave that out. But I think both Lua and Python do ~~provide~~ profit from this logic.

dnpwwo:
I think ths would be dangerous with Python because it requires particular handling of the Global InterLock to makes things thread save and I do not believe that has been done in Domoticz

jvandenbroek:
@dnpwwo I already set a separate uservariables lock on the Python part, but I think the devicesstates should also be locked indeed. Are there any other locks Python need? But of course if in doubt, I'll just remove that, however I think it would be nice for Python too.

Edit: Currently only the uservariables mutex is locking Python, that's the (only) lock set on EvaluateEvent right now... So in the current code there is actual a potential conflict regarding the devicesstates I think? (the devicesstates mutex seems commented out in the Python part of the code)

jvandenbroek:
@dnpwwo Set the extra devices states mutex on Python and tested with the demo scripts, these seem to work fine. I think Python is now better protected than before? ;)

jvandenbroek:
I've restored the Python logic, so it will be executed in the same thread as it does now, with the same mutex (unique_lock uservariables) set. Now we know for sure this won't introduce any new issues regarding Python. Since there are some problems right now with segfaults etc, let's first make sure that Python runs without issues before (eventually) make it threaded.

I did keep the devicename filename matching. If wanted I could just restore that completely as well, so all would be exactly as it is now for Python.

dnpwwo:
Sounds like a great idea !

jvandenbroek:
Will close this PR, maybe will implement threaded later (works great here, but I can imagine you don't want too many changes now with Python issues and dzVents just implemented). To match file and device name I've created a simplier PR here: https://github.com/domoticz/domoticz/pull/1655

PR #1634, opened 2017-06-19
Resolve crash when Device Options are updated

dnpwwo:
As reported https://www.domoticz.com/forum/viewtopic.php?f=65&t=16692&p=134250#p134250


gizmocuz:
Thanks!

PR #1637, opened 2017-06-20
Notification fix Pushover / Pushbullet Windows

jvandenbroek:
This should fix http://www.domoticz.com/forum/viewtopic.php?f=28&t=17948


gizmocuz:
Thanks

PR #1638, opened 2017-06-20
Evohome schedule reading updates

gordonb3:
I found an issue in the schedule reader where the current setpoint could remain empty. Having to rethink the whole procedure, I've also added a cache for the retrieved values which significantly speeds up subsequent lookups.

When switching to 'Day Off' the code will now also display the correct schedule times (i.e. it will try and adapt if the chosen day turns out to be incorrect).


gizmocuz:
Thanks

PR #1642, opened 2017-06-22
make clicking on custom weight sensor on dashboard display log graph

larber:
One minor detail -- if a custom weight sensor is added to the dashboard, you can't click on the icon to show the graph like you can with every other sensor type.  This pull request fixes that.


gizmocuz:
Thanks!

PR #1643, opened 2017-06-23
Sysfs debounce setting

hvbommel:
A contact debounce value can now be selected from a drop down box on the hardware configuration screen. This setting is only applicable when interrupt based gpio input handling is used, i.e. when gpio "edge" is set to "rising", "falling" or "both". The value can be selected using the dropdown box ranging from 10 msec up to 750 msec. The default value is set to 50 msec. This value as was previously hardcoded. When gpio "edge" is set to "none" (and the recommend polling mode is selected) this setting is not used.


hvbommel:
Will do a new pull request with just one commit in it for same.

PR #1645, opened 2017-06-24
Rtl433: fix regression on restarting failed subprocess

peahonen:
When rtl_433 subprocess dies, it won't be restarted but remains as &lt;defunct&gt;.

I think cdbdf79 takes the fgets failure case out of the while loop condition. The failed pipe never gets closed and reopened again.


gizmocuz:
Thanks!

PR #1646, opened 2017-06-24
Sysfs GPIO: Added debounce parameter on hardware config screen

hvbommel:
The SysfsGPIO input de-bounce value can now be selected . This setting is only applicable when interrupt based gpio input handling is used, i.e. when the gpio pin "edge" is set to "rising", "falling" or "both". The value can be selected using the dropdown box, ranging from 1 msec up to 500 msec, on the hardware configuration screen. The default value is set to 50 msec. This value as was previously fixed. When gpio pin "edge" is set to "none" (polling mode) this setting is not used. During startup the domoticz logfile will indicate the selected de-bounce value.

### DIFF
--- a/hardware/SysfsGpio.cpp
+++ b/hardware/SysfsGpio.cpp
@@ -335,7 +335,7 @@ void CSysfsGpio::Do_Work()
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


gizmocuz:
Thanks!

PR #1653, opened 2017-06-26
OpenWeatherMap use weather condition forecast

jvandenbroek:
Use the weather condition codes to set the forecast.


jvandenbroek:
Ah well this doesn't seem to update the forecast correctly, will investigate later.

PR #1655, opened 2017-06-29
EventSystem Lua match device and file name / disable dzVents when not used

jvandenbroek:
To simplify things, I left the threaded part out as implemented in this pull: https://github.com/domoticz/domoticz/pull/1633

This pull is easier to read / debug, so let's do the basic implementation on file matching first. Also left out Python part completely, since I don't use Python scripts, I really don't care much (however I do believe that device name matching with Python would improve it's performance also considerably).

The introduction of dzVents made the execution of my Lua scripts really really slow, like 1 sec extra duration on each device update. I've disabled the dzVents code in EvaluateLua when dzVents is disabled, so things now run smoothly again.

### DIFF
--- a/main/EventSystem.cpp
+++ b/main/EventSystem.cpp
@@ -3246,7 +3246,8 @@ void CEventSystem::EvaluateLua(const std::string &reason, const std::string &fil
 
 	ExportDeviceStatesToLua(lua_state);
 
-	ExportDomoticzDataToLua(lua_state, DeviceID, varId);
+	if (!m_sql.m_bDisableDzVentsSystem)
+		ExportDomoticzDataToLua(lua_state, DeviceID, varId);
 
 	lua_createtable(lua_state, (int)m_uservariables.size(), 0);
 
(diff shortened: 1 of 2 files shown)


scottydj2:
It might be useful to leave the script path export still even if dzVents is disabled as that is the only way any windoze users will be able to use LUA scripts.

jvandenbroek:
That wasn't included before the dzVents merge, you're saying it didn't work before for Windows and now it does? 

Edit: Just Looked at it, but it only supplies a path for dzVents.. So I'm not sure how this would help Windows users execute normal Lua scripts?

dannybloe:
Not having seen your code.. But for me with dzVents enabled, my scripts run in less than 30ms.
Between pressing a switch and seeing dzVents kick in it takes less than 100ms on my system (with around 300 devices).

scottydj2:
I made that change originally to allow dzVents 1.x to be used on a Windows platform, which it didn't work on before.
I assume the LUA scripting would still work, but getting hold of the paths did not.

Thinking about it, I suspect the change was re-architected for dzVents 2.x, so probably not an issue :)

jvandenbroek:
@dannybloe Try running it on a RPi 1 with an average of 1 device update per second ;) 30ms indicates your running something (much?) faster, I can't get Lua under about 100ms even with just 1 simple print line :P But also with a dozen of Lua scripts being executed on a RPi 3 it really slows down, not if the dzVents part is left out, which should be anyway when disabled. So I don't mind and just disable dzVents, but it might be something for you to look into.

dannybloe:
Sure. Thing is, if you don't have any script in dzVents/scripts folder it shouldn't give you any delay except the preperation in C++. dzVents.lua won't find any script files to execute.
But let's face it... a Pi1 isn't up to running Domoticz in the first place. There's no financial reason not to upgrade to a Pi3 if you ask me.

scottydj2:
BUT, there's still code run in Domoticz to generate the tables even if the LUA side has nothing to do when it gets them :)

dannybloe:
You mean specifically for dzVents or for Lua eventing in general?

jvandenbroek:
In general, which causes a lot of overhead (unfortunately).

main/EventSystem.cpp
```diff
@@ -3301,54 +3336,59 @@ void CEventSystem::EvaluateLua(const std::string &reason, const std::string &fil
...
 
-	std::stringstream lua_DirT;
-
-#ifdef WIN32
-	lua_DirT << szUserDataFolder << "scripts\\dzVents\\";
-#else
-	lua_DirT << szUserDataFolder << "scripts/dzVents/";
-#endif
-
 	lua_createtable(lua_state, 7, 0);
```
scottydj2: As the table size would now conditional on dzVents being enabled, this should probably be lua_createtable(lua_state, 0, 0);

jvandenbroek: Thanks, that's indeed not very efficient of me ;)

gizmocuz:
Thanks!

main/EventSystem.cpp
```diff
@@ -1235,7 +1254,28 @@ void CEventSystem::EvaluateEvent(const std::string &reason, const uint64_t Devic
 		{
 			if (reason == "device" && filename.find("_device_") != std::string::npos)
 			{
-				EvaluateLua(reason, lua_Dir + filename, "", DeviceID, devname, nValue, sValue, nValueWording, 0);
+				bDeviceFileFound = false;
```
pierrenicolasr, line 1257: This changes the way Domoticz loads scripts and can break existing setups.
For instance, if you have one script dealing with multiple devices, it will stop working as before as soon as the script has the name of a device.
For instance I'm having a script named "script_device_garage" to deal with everything with the garage (sensor to know if it's open, zwave module to open the door, and a generic fake device named 'garage'). This stopped working as the script won't be called for anything but the fake device.

If this is intended - we should modify https://www.domoticz.com/wiki/Events#By_adding_your_own_scripts_to_the_scripts.2Flua.2F_dir_in_your_domoticz_installation_folder to document this new behavior.
Thx !

jvandenbroek: It's intended, just rename the script to some name that doesn't exist. You're right about updating the wiki, will look into it, thanks for noticing.

pierrenicolasr: Thanks for the feedback - yes, I went through my scripts and renamed the ones impacted.
Good for the wiki !

PR #1657, opened 2017-06-29
Sysfs GPIO optimize code

jvandenbroek:
Now that the 'old' GPIO code seems to be deprecated, I've optimized the execution of this code for especially the interrupt based part, so there is as little as possible CPU time used (just like the 'old' behaviour). Unfortunately there is (not yet) a period timer in the new code to limit the amount of interrupts, but I can solve that partially script based. At least there is now a configurable debounce setting, so I have a workable solution ;)

### DIFF
--- a/hardware/SysfsGpio.cpp
+++ b/hardware/SysfsGpio.cpp
@@ -241,42 +241,45 @@ bool CSysfsGpio::WriteToHardware(const char *pdata, const unsigned char length)
 
 void CSysfsGpio::Do_Work()
 {
-	char path[GPIO_MAX_PATH];
+	bool bUpdateMaster = true;
 	int counter = 0;
-	int input_count = 0;
-	int output_count = 0;
 	m_polling_enabled = false;
 	m_interrupts_enabled = false;
 
-	for (int i = 0; i < m_saved_state.size(); i++)
 	{
-		snprintf(path, GPIO_MAX_PATH, "%s%d/value", GPIO_PATH, m_saved_state[i].pin_number);
-		m_saved_state[i].read_value_fd = open(path, O_RDONLY);
-		if ((m_saved_state[i].direction == GPIO_IN) ? input_count++ : output_count++);
-
-		/*	Enable polling if at least one input edge is set to NONE or is INVALID. */
-		if	(!m_polling_enabled && 
-			(m_saved_state[i].direction == GPIO_IN) && 
-			((m_saved_state[i].edge == GPIO_EDGE_NONE) || (m_saved_state[i].edge == GPIO_EDGE_UNKNOWN)))
+		char path[GPIO_MAX_PATH];
+		int input_count = 0;
+		int output_count = 0;
+		for (int i = 0; i < m_saved_state.size(); i++)
 		{
-			m_polling_enabled = true;
+			snprintf(path, GPIO_MAX_PATH, "%s%d/value", GPIO_PATH, m_saved_state[i].pin_number);
+			m_saved_state[i].read_value_fd = open(path, O_RDONLY);
+			if ((m_saved_state[i].direction == GPIO_IN) ? input_count++ : output_count++);
+
+			/*	Enable polling if at least one input edge is set to NONE or is INVALID. */
+			if	(!m_polling_enabled &&
+				(m_saved_state[i].direction == GPIO_IN) &&
+				((m_saved_state[i].edge == GPIO_EDGE_NONE) || (m_saved_state[i].edge == GPIO_EDGE_UNKNOWN)))
+			{
+				m_polling_enabled = true;
+			}
+
+			/*	Enable interrupts if at least one input edge is set to RISING FALLING or BOTH. */
+			if	(!m_interrupts_enabled &&
+				(m_saved_state[i].direction == GPIO_IN) &&
+				((m_saved_state[i].edge == GPIO_EDGE_RISING) ||
+				(m_saved_state[i].edge == GPIO_EDGE_FALLING) ||
+				(m_saved_state[i].edge == GPIO_EDGE_BOTH)))
+			{
+				m_interrupts_enabled = true;
+			}
 		}
 
-		/*	Enable interrupts if at least one input edge is set to RISING FALLING or BOTH. */
-		if	(!m_interrupts_enabled && 
-			(m_saved_state[i].direction == GPIO_IN) && 
-			((m_saved_state[i].edge == GPIO_EDGE_RISING) || 
-			(m_saved_state[i].edge == GPIO_EDGE_FALLING) || 
-			(m_saved_state[i].edge == GPIO_EDGE_BOTH)))
-		{
-			m_interrupts_enabled = true;
-		}
+		UpdateDomoticzInputs(false); /* Make sure database inputs are in sync with actual hardware */
+
+		_log.Log(LOG_STATUS, "Sysfs GPIO: Worker startup, polling:%s interrupts:%s debounce:%d inputs:%d outputs:%d",
+			m_polling_enabled ? "yes":"no", m_interrupts_enabled ? "yes":"no", m_debounce_msec, input_count, output_count);
 	}
-
-	UpdateDomoticzInputs(false); /* Make sure database inputs are in sync with actual hardware */
-
-	_log.Log(LOG_STATUS, "Sysfs GPIO: Worker startup, polling:%s interrupts:%s debounce:%d inputs:%d outputs:%d", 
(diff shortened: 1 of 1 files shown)


hvbommel (commented):
What is the purpose of the "{" on  line 249 and it matching "}" You seem to create a block for some reason. I don't see how that helps performance nor readability of the code. The new bool bUpdateMaster doesn't harm I agree. Not sure if the time change from 0.5 to 2 sec helps because it will make the system less responsive on a rundown request. Even on a slow processor the overhead of reissuing the wait for interrupts every 0.5 seconds instead of the 2 seconds will not cause measurable delays in any way (unless you tell me you have measured it and know I am wrong there). I choose the 0.5 carefully. So I am not sure this change helps the system. Also, you mentioned you are missing a interrupt blocker. In my mind that should not be needed when you have stable inputs. Even when you don't you never should have constant interrupts occurring. (But if you do, use polling instead of interrupts). In that case something else (hardware or its design) is badly broken. But with a higher debounce value you should be able to calm down the interrupt bursts should they occur.  Let me know what you think. 
Summary: I would recommend to leave your "bUpdateMaster check" change in, keep the rest as was.

jvandenbroek:
It's good practice to destroy no longer used variables, it also saves a few bytes of precious stack memory (but no, that's hardly an argument these days, ppl don't seem to care anymore, just wanted to mention it ;)) It would actually be better to move this part to StartHardware instead, but I didn't want to move your code around too much and placing it in a separate block does essentially do the same. If you'd rather have it moved, sure, I can do that, let me know.

It doesn't indeed make much difference if you set it to 0,5 or 2 sec, but that's almost always the case with optimizations. However, you must look at it at another perspective; It does now run 4 times less than before. It's also unfortunately that select seems to be less efficient than poll, because of the need to keep copying the FD set in/on the heap on every timeout. I know it's mandatory for select, don't like it, but accept it as it is ;) 

Stopping hardware delay can't hardly be an issue on the other hand, most other hardware in Domoticz has sleeping functions of at least 1 second, and you should only occasionally do restarts..? The average wait time with 2 seconds timeout is 1 second, not too bad if you ask me. We could set it to 1 sec as some sort of compromise if you like, but it's really not too important (I could even completely revert it, but really recommend against it).

Here an example of why the period timer would helps (that's why threaded pins are so powerful, you won't stall other pins with period timer active):

```
2017-07-01 14:15:06.347  Sysfs GPIO: Pin23 new state: 0
2017-07-01 14:15:06.432  Sysfs GPIO: Pin23 new state: 0
2017-07-01 14:15:06.472  Sysfs GPIO: Pin23 new state: 0
2017-07-01 14:15:06.543  EventSystem: Script event triggered: scripts/lua/script_device_doorbell_button.lua
2017-07-01 14:15:06.589  (Sysfs GPIO) Lighting 2 (Doorbell Button)
2017-07-01 14:15:06.600  Sysfs GPIO: Pin23 new state: 1
2017-07-01 14:15:06.773  (RFXtrx433e) Lighting 2 (Deurbel ding dong)
2017-07-01 14:15:06.874  (Sysfs GPIO) Lighting 2 (Doorbell Button)
```
Because of the multiple interrupts, executing time is delayed (simply because of claiming CPU time). Seems not too bad in this case, but a period timer would filter it out. The ringing is caused by large cabling and can only partially be filtered out by using a large RC value set, which I already do (believe me, I've experimented enough with it..). But it's probably a bad example anyway, because since I'm no longer using a reflective sensor for my gas meter, this has become less of an issue for me than before. You're talking about defective hardware, but no, sometimes you can't really control the input status. With a reflective sensor, many users have the issue that it will be occasionally stall at a point that's it just flickering on the edge of the surface. If you have a period timer as in the other GPIO code, it's much easier to filter this out. We can't solve this with a larger debounce delay, that really has another purpose.

Please, review again and I'd like to hear from you :)

hvbommel:
Ok cool, makes sense. Then I would say:
1. Move the code to StartHardware if you feel those sleeping bytes on the stack matter, I think it doesn't. It would if we would leave a sleeping 1k for no reason, I agree. In this case keep the code close to each other helps code readability.
2. Set the value from 2 to 1 seconds.

About blanketing interrupts ... we have to be careful the correct states are maintained. When we block interrupts we may loose the latest state, unless we do something about it like poll after the burst protection has completed. I understand what you say about the reflective sensor "corner case". Because this is an exceptional case maybe we can put something in like max 5 state changes per second. When exceeded disable interrupts for that pin for 10 seconds, and log the fact. Then when re-enabeling first poll the pin to make sure the latest state is up to date, and then re-enable the interrupts again. I would prefer to not make it configurable, but document this instead, to keep configuring simple enough for the users.

hvbommel:
Also ... in the "old" gpio there is a separate thread for each gpio pin. And each thread context requires memory (plus scheduler time even when there is no state change). This takes much more then the few bytes that where saved in this PR. And that is what I did not like of the old design, it is ok when you have just a few gpio pins, say max 8 or so. We can have many more, like 64 or so. Creating a thread fore each one is no good in that case, unless all your system is doing is handling the pins, nothing else. Because of that I have only 1 thread to handle all the interrupts. The footprint remains to same independent of the # of pins. I am saying it just in case you where planning to add more threads into this design. (I thought I read between the lines you like the many threads in the old design, I don't. Correct me if I am wrong.)

In the end I think you might  consider remove this whole PR because I don't think the change will buy anything, and it is not worth the change. Don't get me wrong, I really like changes but only when they address known problems. And every change requires testing as well.

Feel free to disagree.

jvandenbroek:
You overlooked the most important part I've changed; moving UpdateDomoticzInputs to the if (m_polling_enabled) statement in the while loop and placing it separately under the edge trigger. This saves a lot of cpu time when not using any polling inputs (yes, that I've measured ;)). All others are indeed tiny optimizations, but why shouldn't we do that, doesn't cost you a thing ;) I also mentioned that those few bytes are not really why I did it, but it's just considered good practice in to destroy variables which aren't used anymore.

I'm not planning to make the codebase multi-threaded, don't worry ;) But I think you overestimate the overhead of threads (which is really really small, all action happens in the same virtual memory pool, these are not to be confused with forked processes which wiringPi was using previously). Since these threads are idle most the time, the scheduler has not much to do, why do you think different? I think that in the extreme hypothetical case of 64 pins it might need to do some more work, but you really think reading through all those files every 250ms is very efficient? I dare to bet it will introduce more overhead ;) Of course the single edge thread is in this case better yes, that we can agree on. That's also why I don't mind too much about deprecating the other GPIO code, but it should be noted that both solutions have their advantages/disadvantages.

hvbommel:
Ok, I am with you. 

So then this still applies. After that I agree to merge it.
1. Move the code to StartHardware if you feel those sleeping bytes on the stack matter, I think it doesn't)
2. Set the timer value from 2 to 1 seconds.

Thanks for your drive to get things better and the work you did. 

Ps: I know the interrupt handler itself can be optimized. I am planning to that some time in the future.

hvbommel:
Looks good to me now, **ready to merge** from my point of view ... although see below (2nd review) ...
Thanks.

hvbommel:
regarding line 379 and 380. Do_work is already updating. I did not do this in the interrupt loop to prevent storms. I think it is better to take it out. It saves cycles and the system gets less nervous. Even without polling enabled the updates are done already. We may also run into synchronization problems (race conditions). Take it out is my advise, (I missed this change before). The design has separated the data collection (polling, interrupts) and reporting to the domoticz database on purpose. Because Do_work is always running it serves both polling and reporting.

I looked at it again ... maybe it is indeed not the problem to do the updates in the edge thread but then we need to take the updates out in the do_work thread. Do_work and edge should never both do the (assync) updates. But if we do that we may get interrupts storms and lots of domoticz database toggels. So I would think twice and accept the poll overhead in do_work. On my systems (3x raspi-pi 3's) I see no problems at all eben when I log in multiple times and monitor 24 IO pins at much higher rates then we do now in our sysfs_gpio.

hvbommel:
Above discussions prove it is hard to predict why certain code is like it is. I am not talking about bugs, they need top be fixed. But tweeking designs is something else. We need to make sure we don't push in different directions leading to messy or even worse, broken code. My goal for this design was to have a very reliable and robust system that meets requirements, ready input states, without driving it to the limit (like dealing 100's of interrupts per minute, because of a reflective sensor on the corner, was a non-design goal). Anyway, probably this discussion should go somewhere else, not here. & I don't want to bug you but keep the design sound.

hvbommel:
==>>> it can be merged as is. 
In the end it all looks ok. No race conditions because of the hand in hand change you did in do_work as well. Lets go for it.

gizmocuz:
Thanks!

Issue #1656, opened 2017-06-29
Humidity values not being sent by InfluxDB Datapush


GJRos:
Domoticz Ver: 3.8015 on Raspberry Pi

After a recent Beta update, the InfluxDB datapush is incorrectly sending all Humidity values as zero. It is happening with my RFLink and ESPEasy devices. 

2017-06-29 15:05:30.404 InfluxLink: value Humidity,idx=52,name=Room value=0
2017-06-29 15:05:30.404 InfluxLink: value Barometer,idx=52,name=Room value=846.0
2017-06-29 15:05:40.002 MQTT: Topic: domoticz/in, Message: {"idx":52,"nvalue":0,"svalue":"19.53;23.79;0;846.04;0"}
2017-06-29 15:05:40.045 InfluxLink: value Temperature,idx=52,name=Room value=19.5
2017-06-29 15:05:40.045 InfluxLink: value Humidity,idx=52,name=Room value=0

Tommit1:
Could it be because or before of this request? 
#1648 
Now we have Version 3.8025!
Is Version 3.8015 before or after this correction?

ikkemaniac:
@GJRos, please provide the build hash, you can get this info from the Setup -> About screen of your current build that fails.
Please test if #1648 causes this issue by reverting it (`git revert f24fa115bff3cd6948b2ad306ed14e35f6645483` and rebuild) while on the branch/commit you've built 3.8015 with.

@Tommit1 could be since the change is in the generic file. I've made this change because the true 'Humidity' sensor as a dummy device and the reference in the wiki (https://www.domoticz.com/wiki/Domoticz_API/JSON_URL%27s#Humidity) have levels saved as the nValue, humidity status ie 'comfort' levels are stored as sValue.

btw i  don't understand why the legacy svn version numbering is used. Life would be easier if everything was in commit hashes only.

ikkemaniac:
Just tested on my own system with build hash 37b5133 (Version: 3.8025)
- Http Datapush functions correct. 
- InfluxDB: I can't test InfluxDB itself as i don't have it installed but this is the raw data captured that domoticz sends out when I set the device to a new value of 100:

```
POST /write?db=test HTTP/1.1
User-Agent: Mozilla/5.0 (Windows NT 10.0; WOW64) AppleWebKit/601.37 (KHTML, like Gecko) Chrome/54.0.40370.0 Safari/601.36
Host: localhost:12345
Accept: */*
[...]

PR #1658, opened 2017-06-30
Revert "bugfix: return 'nvalue' iso 'svalue' for humidity levels in d…

ikkemaniac:
…atapush handling"

While this fixes the issue with dummy devices it breaks real devices because of inconsistancies in data storage and data parsing in the datapush module between dummy devices and real devices.
see issue #1656 for details on the issue and a possible fix for both situations.
Please revert this fix as it breaks more than it fixes.

This reverts commit f24fa115bff3cd6948b2ad306ed14e35f6645483.


gizmocuz:
Thanks!

Issue #1533, opened 2017-05-08
Humidity graph shows 'temperature' as label


koenkooi:
I was working on adding info from my Ducobox ventilation unit through the JSON API when I noticed that the resulting graphs for the humidity sensors show 'temperature' and 'degrees':

Version 3.7439 revision a2148229cd8823df4657e9f60a70847b11fb8ccc

![schermafbeelding 2017-05-08 om 09 47 52](https://cloud.githubusercontent.com/assets/259525/25794980/81963d10-33d3-11e7-8511-6fb65c3b748f.png)

Tommit1:
What kind of sensor is this in domoticz ... humidity only or a combined TEMPHUM one?

koenkooi:
Humidity only:
<img width="1208" alt="schermafbeelding 2017-05-08 om 14 45 32" src="https://cloud.githubusercontent.com/assets/259525/25804646/06347608-33fd-11e7-86b4-83046fa1fdf3.png">

ikkemaniac:
2nd that for the virtual humidity sensors as well (HUM only)

szczukot:
I have sensor 
Temp + Humidity + Baro (Type THB1 - BTHR918, BTHGN129) and all graphs is without data :(
I only see current value on sensor.

koenkooi:
Why was this closed? It's still an issue in both master and development.

gizmocuz:
whoops sorry, please contact me on the forum, i would like to have a backup of your database to do some debugging against

koenkooi:
I can't do PMs yet, so here's the topic: https://www.domoticz.com/forum/viewtopic.php?f=6&t=20268

koenkooi:
Fix confirmed:
<img width="1193" alt="schermafbeelding 2017-11-08 om 09 04 05" src="https://user-images.githubusercontent.com/259525/32537691-ccd50e66-c463-11e7-90d0-945c41f5c388.png">

szczukot:
I have 
![image](https://user-images.githubusercontent.com/10205661/32540701-c3937900-c46d-11e7-96e3-891616f21015.png)

![image](https://user-images.githubusercontent.com/10205661/32540842-3ced99c0-c46e-11e7-91c8-3f195a7afc5a.png)

and graph for Weather tab (and last icon - barometr)):
![image](https://user-images.githubusercontent.com/10205661/32540807-1cc2643c-c46e-11e7-97c4-f41e490521bd.png)

[...]

PR #1659, opened 2017-06-30
Fix humidity graphs title

ikkemaniac:
Fix chart title when viewing humidity logs. If chart is 'humidity', update title when data is updated. Original chart title remains 'temperature' but is replaced as even the initial data update/load is done through the 'data update' function. probably fixes domoticz/domoticz#1533


gizmocuz:
Did you test this code with also a temp+hum sensor ?
This just seems to replace the chart title always when there is humidity data, even if it can also contain temperature data
better to check in the end if there is no temp data, but hum data, and then replace the title

gizmocuz:
If you make a pull request, please answer the questions (also on your last one)
Else i have to close them

ikkemaniac:
not tested and will replace temp+hum, so merging this will fix 1 out of 2 bug's.

What do you mean with "also on your last one"? The security issue rone hasn't had response yet.

gizmocuz:
Sorry, but i did give a response in the crc check pull request
Please test with the temp+hum sensor

gizmocuz:
Not working, see above comment

PR #1660, opened 2017-06-30
P1 Smart Meter suppress spikes

jvandenbroek:
I think this should prevent spikes on meters without CRC check enabled/possible: http://www.domoticz.com/forum/viewtopic.php?f=6&t=18099

### DIFF
--- a/hardware/P1MeterBase.cpp
+++ b/hardware/P1MeterBase.cpp
@@ -319,28 +319,28 @@ bool P1MeterBase::MatchLine()
 				break;
 			case P1TYPE_POWERUSAGE1:
 				temp_usage = (unsigned long)(strtod(value,&validate)*1000.0f);
-				if (!m_power.powerusage1)
+				if (!m_power.powerusage1 || m_p1version >= 4)
 					m_power.powerusage1 = temp_usage;
 				else if (temp_usage - m_power.powerusage1 < 10000)
 					m_power.powerusage1 = temp_usage;
 				break;
 			case P1TYPE_POWERUSAGE2:
 				temp_usage = (unsigned long)(strtod(value,&validate)*1000.0f);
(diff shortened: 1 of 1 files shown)


gordonb3:
I see what you attempt to do here, but could you include the meter version in those conditional updates? This should not affect meters that have m_p1version >=4

jvandenbroek:
~~Now also check for CRC is disabled, guess that's even better than just relying on the version right?~~
Edit: Wait, that doesn't work :) Thought disable_crc was global..

Ok, now indeed check the meter version as you suggested..

gordonb3:
My vote would be against that. If one deliberately disables error checking we should not replace it with a different error defeating (or rather: masking) mechanism. Except of course errors that may crash dz

jvandenbroek:
I already updated it, because it didn't work ;)

gordonb3:
Noticed the third commit just as I commented.

Don't really know why the call to ParseData is done this way. `disable_crc` is in fact loaded from the Class variable `m_bDisableCRC`. May want to note that the logic on this variable is reverted, i.e. `true` means that there is no CRC checking.

Pigloox:
I have never had strange values ​​with the CRC verification and I do not know how this patch will behave with a "peak consumption"

jvandenbroek:
@Pigloox I'm sorry, but you should really check my edits and the current code before commenting. I reverted the (non) working disable CRC check and implemented the suggestion of @gordonb3 to check on >= v4

gizmocuz:
Should we merge this ?

jvandenbroek:
I say yes ;) I doesn't harm ppl with version >= 4 and helps others without CRC possibility. I kept the numbers high (way higher than needed) so no worries about missing peak values or something.

gizmocuz:
Thanks!

PR #1663, opened 2017-07-02
EventSystem fix high load with dzVents enabled

jvandenbroek:
This should fix the high load issues users have with dzVents enabled

### DIFF
--- a/main/EventSystem.cpp
+++ b/main/EventSystem.cpp
@@ -172,64 +172,75 @@ void CEventSystem::SetEnabled(const bool bEnabled)
 
 void CEventSystem::LoadEvents()
 {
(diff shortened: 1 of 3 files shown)


jvandenbroek:
Wait before merge, seems not to do (fix) what I intended... Need to review some more code.

scottydj2:
Looks good to merge.

jvandenbroek:
This will completely disable executing dzVents Lua code if dzVents is enabled (default) but no scripts are active / available, so users without dzVents scripts won't notice any slowdown anymore.

dannybloe:
I tested this on various systems and seems good to go.
Merging... 
Thanks guys.. 
One more optimization to go...

Issue #1664, opened 2017-07-03
SECURITY ISSUE: download_update.sh returns true when checksum check fails


ikkemaniac:
When downloading updates `scripts/download_update.sh` gets called from domoticz. Based on the script exit value domoticz determines if (1) the new release `update.tgz` download was successful and (2) if the files sha256 check sum matches a value that is separately downloaded in `update.tgz.sha256sum`.

The script return value is currently 0 (positive) while the sha256 checksum check fails.

example contents of `update.tgz.sha256sum`:
```
$ cat update.tgz.sha256sum
d6940bc2481ffaa41b8cd55850d16cdafd103db2fbbab2c3dc5c413ee868d4c9   domoticz_linux_x86_64.tgz
```

and the matching release `update.tgz`:
```
$ sha256sum update.tgz
d6940bc2481ffaa41b8cd55850d16cdafd103db2fbbab2c3dc5c413ee868d4c9  update.tgz
```
the checksum check in `scripts/download_update.sh` fails but the scripts exit status is still a '0':
```
$ ./scripts/download_update.sh `pwd` /beta
sha256sum: domoticz_linux_x86_64.tgz: No such file or directory
sha256sum: WARNING: 1 listed file could not be read
./scripts/download_update.sh: 45: [: !=: unexpected operator
$ echo $?
0
```

the problem is that the exit status is returned from the last command, in this case from `scripts/download_update.sh`:
```
valid=$(LC_ALL=C sha256sum -c update.tgz.sha256sum | grep update.tgz | cut -d':' -f2 | tr -d ' ')
```
the last command is `tr-d ' '` which itself doesn't fail and thus returns 0.

If look more closely at the command we see the following pipe status':
`$ 1 1 0 0`
the first and second command of the pipe chain fail, the third and forth succeed.
This could be easily fixed to test if `$valid` is empty or `OK`.
`  if [ $valid != "OK" ]`
to
`  if [ -z $valid ] or  [ $valid != "OK" ]`

Doing this though will break the update function for the future if the checksum return file is not changed (see below). I advise against implementing this fix by itself.

Looking at the warning message from sha256sum we see it's expecting to find `domoticz_linux_x86_64.tgz` but it cannot, file does not exist because `wget -q "${archive_file}" -O update.tgz` renamed it to `update.tgz`

`LC_ALL=C sha256sum -c update.tgz.sha256sum | grep update.tgz | cut -d':' -f2 | tr -d ' '` 
can be changed to 
[...]

PR #1665, opened 2017-07-03
SECURITY ISSUE: download_update.sh returns true when checksum check fails

ikkemaniac:
see #1664


scripts/download_update.sh
```diff
@@ -41,10 +41,14 @@ fi
...
+  valid=$(LC_ALL=C cat update.tgz.sha256sum | awk '{print $1"  update.tgz"}' | sha256sum -c -|  cut -d':' -f2 | tr -d ' ')
+  if [ -z $valid ] || [ "$valid" != "OK" ]
   then
         echo "Archive checksum mismatch !";
         exit 1;
   fi
 fi
+
+#should never be here
+exit 1;
```
gizmocuz, line 54: When all succeeds, we should be here ?

PR #1666, opened 2017-07-04
Hardware Monitor set different timers per device (less hogging)

jvandenbroek:
The Hardware Monitor is updating every 30 sec, seems a little bit often. Does cause considerable delays with scripting enabled (multiple devices updated simultaneously).

### DIFF
--- a/hardware/HardwareMonitor.cpp
+++ b/hardware/HardwareMonitor.cpp
@@ -59,7 +59,7 @@
 #endif
 
 #define POLL_INTERVAL_CPU	30
-#define	POLL_INTERVAL_TEMP	70
+#define POLL_INTERVAL_TEMP	70
 #define POLL_INTERVAL_MEM	80
 #define POLL_INTERVAL_DISK	150
 
@@ -144,7 +144,7 @@ void CHardwareMonitor::Do_Work()
 	bProcessDisk = true;
 
 	int msec_counter = 0;
-	int sec_counter = POLL_INTERVAL_CPU - 3;
+	int64_t sec_counter = POLL_INTERVAL_CPU - 5;
 	while (!m_stoprequested)
 	{
 		sleep_milliseconds(500);
--- a/hardware/HardwareMonitor.cpp
+++ b/hardware/HardwareMonitor.cpp
@@ -134,6 +134,7 @@ void CHardwareMonitor::Do_Work()
 {
 	_log.Log(LOG_STATUS, "Hardware Monitor: Started");
 
+	bProcessMemDisk = true;
 	int msec_counter = 0;
 	int sec_counter = POLL_INTERVAL - 3;
 	while (!m_stoprequested)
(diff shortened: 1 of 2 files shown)


gizmocuz:
Maybe make it more relaxter... 10 second poll temp 10 seconds later poll CPU 10 seconds later poll memory 10 seconds later poll disk and loop ?

marcelmah:
Wouldn't this spam the logs if you have a poll every 10 seconds? I think that for CPU it might be useful to regularly poll the value (mem, disk and temp don't change much so fast (usually)). If you can prevent it from spamming the logs that would be great.

dannybloe:
Yeah, stuff like disk usage for me could be checked every 10 minutes even. I don't want to know this by the minute.

jvandenbroek:
I've combined it now, so it sleeps 10 sec in between and waits 300 sec before updating memory and disk (kept CPU usage out of it, since I believe that's a more useful monitor to check more often)

gizmocuz:
okey, then poll the disk usage less (only force at statup of domoticz)
But default there is no log output (to file), you can also filter on status/error

gizmocuz:
5 minutes is to high!... needs to be less then 5 because of the shortlog interval

But its getting complex now...

why not increment a counter....

if (count % POLLINTERVAL_TEMP == 0) ---> Poll Temp
if (count % POLLINTERVAL_DISK == 0) ---> Poll Disk
if (count % POLLINTERVAL_MEM== 0) ---> Poll Mem
if (count % POLLINTERVAL_CPU == 0) ---> Poll CPU

All poll intervals should be whole numbers like 10,20,30,40

This way when the counter starts at zero, all values are polled, start the counter on -5 (int64_t)

CPU should be checked every 30 seconds at least.... this is indeed important
Temp each 60 seconds
And the rest each 3 minutes ?

jvandenbroek:
Right, 3 minutes makes sense than yes. More complex? Only easier to do when I move around code. But I was thinking only to do hard disk every 3 minutes, the rest 60 sec (including memory). You want 30 sec for CPU? Hmm that complicate things. 

Oh I see what you mean, define different timers, that's a good idea.

jvandenbroek:
Something like this? We need to set booleans, otherwise timers could get missed if functions take too long to execute.. I've set the timers so they're not executed at the same time. What do you think?

Edit: Just realized that it still can stall on some called function, so maybe just leave the bools out... How can we prevent the counter to slow down without a separate thread? Hmm..

jvandenbroek:
Now run FetchData in a separate thread, so we know for sure that the counter in do_work is incremented correctly and nothing will be missed this way. I also think the timer stalling (thus preventing heart beat to happen) is also the main cause of issues with this hardware on high system load (when devices are processed very slowly)

Edit: reverted the thread commit, no need for it.. Had to take a small break to realize :P

gizmocuz:
Thanks!!

jvandenbroek:
I've added an ifdef which should fix the compile error on win32 and Mac, just pushed it to my own repo, waiting for Travis and appveyor to complete..

PR #1667, opened 2017-07-04
Sysfs GPIO: Poll at 1st timeout after interrupt received

hvbommel:
After one or more GPIO state change interrupts have been detected a poll is done to make
	sure the domoticz database reflects the actual states of all GPIO pins in all cases. A
	missed interrupt can occur when a GPIO pin changes state twice within m_debounce_msec.


jvandenbroek:
Can't harm to do this :+1:

PR #1668, opened 2017-07-05
dzVents: Improve performance and reduce resource usage.

scottydj2:


main/EventSystem.cpp
```diff
@@ -2442,27 +2440,28 @@ void CEventSystem::ExportDomoticzDataToLua(lua_State *lua_state, uint64_t device
 			_hardwareNames[ID] = tlist;
 		}
 	}
+	
+	Json::Value tempjson;
+	m_webservers.GetJSonDevices(tempjson, "true", "", "dzvID", "", "", "", true, false, false, 0, "");
```
gizmocuz, line 2445: This is a very very expensive call... this takes a lot of CPU... specially with systems with lots of devices (like mine)
It seems you are now calling this function for each lua evaluation ?
Why not use the device states that are stored after each update ? That's why it was created... and it should always be up2date

main/EventSystem.cpp
```diff
@@ -2442,27 +2440,28 @@ void CEventSystem::ExportDomoticzDataToLua(lua_State *lua_state, uint64_t device
...
+	if (tempjson["result"].size() != m_devicestates.size())
+	{
+		_log.Log(LOG_ERROR, "Json and devicestates size mismatch!");
+	}
+	
 	lua_createtable(lua_state, 0, 0);
 
 	// First export all the devices.
 	typedef std::map<uint64_t, _tDeviceStatus>::iterator it_type;
+	int ii2 = 0;
```
gizmocuz, line 2456: after ii, there is jj, kk, ll, mm
But sometimes you can label the variable better then ii,jj,...

main/EventSystem.cpp
scottydj2: Agreed, it's an expensive call. 
The problem is that we want the data fields which are only processed and created at the end of that call and put in the json. They are not available anywhere else (unless we duplicate the code).

main/EventSystem.cpp
scottydj2: Agreed. Will fix.

main/EventSystem.cpp
dannybloe: The json has more (constructed) information (formatted) that is unfortunately not in the database otherwise we would definitely have used it.  In EventSystem.cpp at the top you see the list (_tJsonLuaMap) of items that are only available through the json data structure. There is quite some logic in that json-generation code specifically for this data preparation (previously for the GUI alone but dzVents needs that info too). 
But agreed, it's a very expensive call.

@scottydj2 / @jvandenbroek  correct me if I'm wrong.

dannybloe: I think it would be awesome if this constructed data could land in the database somehow.

scottydj2: Not necessarily in the DB, just in the internal tables. I assume the expense of the call is in the DB query?

gizmocuz: If there are to less fields in the devicestates table, should this table not be enhanced ?

scottydj2: Agreed, that would make sense. So would you then expect this special functionality to move from the JSON function into the device_update function instead?

jvandenbroek: I rather have a separate table which only gets filled when dzvents is actually enabled.. We can easily match id's on both.

dannybloe: Agreed however, some of those fields are derived fields like counter results. I think that when  you change some domoticz settings the results of those fields could change as well. Of course they will be updated in the next device-update cycle with the new (formatted) values so that may not be an issue.
But besides this, this PR makes it a whole lot more efficient than what is currently in master so I think this is still a good iteration. 
Next step would be perhaps to add those fields and perhaps look again at @jvandenbroek (retracted) pull request to make better use of multi-core systems when handling events.

scottydj2: But the special handling is currently only done a part of a JSON request.

dannybloe: In memory instead of db.. maybe even better/faster.

scottydj2: Yes, this change is still an awful lot better (performance wise) than what we have mainline currently.

gizmocuz: Please discuss this on Slack, every developer gets updates when there is a change in a pull request
I would like to have one table that is up2date and can be used for any implementation
Yes, in memory, in a table/structure/eventsystem class member

scottydj2:
Will be superceeded by another PR.

PR #1669, opened 2017-07-05
GUI set correct order column

jvandenbroek:
While working hard at fixing dzVents issues, this 'bug' also appeared after fixing code in WebServer.cpp. The default order is set to column 'Name', which in it's current state is being ignored and always falls back to column 'Order'. But there is no reason to keep this in code, it's confusing and we should also fix this in WebServer.cpp to make the order argument actually work (but I suggest to do that after all dzVents patches are merged).


gizmocuz:
Thanks!

PR #1670, opened 2017-07-05
Fix 'SubType' when device change

DT27:
When change the type between normal switch and selector switch, the SubType won't change.
This can fit it.

### DIFF
--- a/main/WebServer.cpp
+++ b/main/WebServer.cpp
@@ -12106,13 +12106,51 @@ namespace http {
 			}
 			else
 			{
-				if (switchtype == -1)
-					m_sql.safe_query("UPDATE DeviceStatus SET Used=%d, Name='%q', Description='%q' WHERE (ID == '%q')",
+				//Fix 'SubType' when device change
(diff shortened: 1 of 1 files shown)


main/WebServer.cpp
```diff
@@ -12106,13 +12106,51 @@ namespace http {
...
+				{
+					case -1:
+					{
+						m_sql.safe_query("UPDATE DeviceStatus SET Used=%d, Name='%q', Description='%q' WHERE (ID == '%q')",
 						used, name.c_str(), description.c_str(), idx.c_str());
-				else
-					m_sql.safe_query(
+						break;
+					}
+					case 6:
```
gizmocuz, line 12118: Can you use the switch type enum ?

gizmocuz:
Selector switches has to be created from the dummy hardware tap..
Best to create a patch where you 'can not' change to a selector switch from the switch edit options..

hvbommel:
see PR #1680

PR #1671, opened 2017-07-05
Fixed HTTP push issues

sbouchex:
Fixed 2 issues :
- Potential crash related to the delimiter position (no testcase found but better to be protected to avoid crash)
- Avoid logging an error if the remote HTTP server is not retuning an output after a POST


gizmocuz:
Thanks!

PR #1674, opened 2017-07-06
Fix 'SubType' when device change with enum

DT27:
When change the type between normal switch and selector switch, the SubType won't change, and the switch can not edit again.
This can fit it.

### DIFF
--- a/main/WebServer.cpp
+++ b/main/WebServer.cpp
@@ -12106,13 +12106,51 @@ namespace http {
 			}
 			else
 			{
-				if (switchtype == -1)
-					m_sql.safe_query("UPDATE DeviceStatus SET Used=%d, Name='%q', Description='%q' WHERE (ID == '%q')",
+				//Fix 'SubType' when device change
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thanks, but do you agree a patch that stops being able to change a switch in a selector switch ?
You are trying to change a non-selector-switch to a selector-switch ?

DT27:
@gizmocuz Sometimes need to change on both sides, like the BroadlinkRM2 plugin. Please do not cut this function, it's useful. 

_Sent from my Samsung SM-G9550 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

gizmocuz:
No worries ;) Thanks for the patch!

hvbommel:
See PR #1680
DT27;  FYI above PR was requested to revert your change.

DT27:
@hvbommel thanks, I will test and fix the problem.

PR #1676, opened 2017-07-07
Sysfs GPIO: remove logfile trace as requested on forum

hvbommel:
Remove logfile trace as requested on forum, at startup, for interrupt based gpio pins, when no database change has occurred. Also added some error case traces for exceptions at startup & rundown.


gizmocuz:
Thanks!

PR #1680, opened 2017-07-08
Webserver: Revert recent change to fix problem reported on forum

hvbommel:
Workaround for problem reported on forum by several people. See forum bugs and problems "On/Off Switch becoming unusable after modification".

Note: I know the intended subtype change wont work any more (as that code is removed by this PR). If still needed this will be re-addressed and tested by DT27 I assume.

### DIFF
--- a/main/WebServer.cpp
+++ b/main/WebServer.cpp
@@ -12106,56 +12106,8 @@ namespace http {
 			}
 			else
 			{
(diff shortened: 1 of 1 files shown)


hvbommel:
I know the intended subtype change wont work any more as it is reverted. This will be addressed and tested by DT27 I assume.

PR #1684, opened 2017-07-09
HardwareMonitor df exclude unused fstypes

jvandenbroek:
Make sure NFS isn't called unnecessary to prevent lockups, also excluded other unused types (we're only looking for /dev). User with specific issue: http://www.domoticz.com/forum/viewtopic.php?f=6&t=18245


gizmocuz:
When Domoticz is installed on a nfs share...Will this still show the disk usage?

xbeaudouin:
It will still show local disk usage. Even if domoticz is installed on a nfs share (which is not a good idea with the SQLite, but it works).

gizmocuz:
Thanks!

PR #1685, opened 2017-07-10
dzVents perfomance improvements

scottydj2:
Currently dzVents uses a json call to get the additional info fields required on every device call for every device.
The code has been modified to maintain a cache of this information within the devicestates table and is only updated with each device update resulting in better performance and reduced cpu load.

There have also been other additions to the dzVents device handling including support for rgbw type lights and temperature farenheight conversion.


scottydj2:
SUbmitted to wrong branch.

PR #1686, opened 2017-07-11
EventSystem / dzVents rewrote update device logic

jvandenbroek:
We've put a lot of effort in rewriting the slow dzVents implementation, by changing the following (compared to current implementation):

- Added maps for different data types to m_devicesstates, which are filled once with all data from Json on start-up (just like the other fields now from DB)
- On each device update, only it's map entry will be updated with a single Json call (takes about 8-10ms on a RPi 1, so negligible)
- For now only fill / update these specific Json maps when dzVents is enabled, but it can be easily used by other scripting as well if needed in the future
- dzVents fixes and improvements

The overall load has reduced significantly on my slow system, which keeps responding well now (with 118 active devices). We ( @dannybloe , @scottydj2 and me) are running it for a few days now without issues.


dannybloe:
Note for gizmocus: you only have to look at EventSystem.*
The rest is dzVents stuff.

gizmocuz:
Thanks!

PR #1688, opened 2017-07-11
Sysfs GPIO: Log new gpio-state after state-change

hvbommel:
Added log message as requested on forum to indicate the new state of the gpio-pin after state change. Example: Sysfs GPIO: State change gpio20:off

### DIFF
--- a/hardware/SysfsGpio.cpp
+++ b/hardware/SysfsGpio.cpp
@@ -762,6 +762,10 @@ void CSysfsGpio::UpdateDomoticzInputs(bool forceUpdate)
 					m_Packet.LIGHTING2.unitcode = (char)m_saved_state[i].pin_number;
 					m_Packet.LIGHTING2.seqnbr++;
 					sDecodeRXMessage(this, (const unsigned char *)&m_Packet.LIGHTING2, "Input", 255);
+
+					_log.Log(LOG_STATUS, "Sysfs GPIO: State change gpio%d:%s",
+						m_saved_state[i].pin_number,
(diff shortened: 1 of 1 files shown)


gizmocuz:
Would this not also trigger a switch change... And this also a log message?

hvbommel:
**In the new case when Setup/Log "Status" is selected in the new situation we get:**
2017-07-12 23:09:27.026 Sysfs GPIO: State change gpio20:on

**When in Setup/Log "All" is selected then we get logging below.**

**previous logging would have been like this:**
2017-07-12 23:09:27.034 (Sysfs GPIO) Lighting 2 (Door-20)

**new logging:**
2017-07-12 23:09:27.026 Sysfs GPIO: State change gpio20:on
2017-07-12 23:09:27.034 (Sysfs GPIO) Lighting 2 (Door-20)

**other option**
So the other option would be to add the new state to the lighting 2 log message like this:
2017-07-12 23:10:54.547 (Sysfs GPIO) Lighting 2 (Door-20) on
This may be better but in that case other hardware types would also see this change and I am not sure if everyone would like that.

This change is only a logging change, nothing else.

If you don't like the change, just close it. I have no problem with that. I was only trying to address a forum request. See http://www.domoticz.com/forum/viewtopic.php?f=4&t=17688 - 9-juli-2017. But I think it is ok to implement this. It is only changing to log file content for those who rely on logging. The system is not effected.

gizmocuz:
Thanks, and thanks for clearing up!

PR #1692, opened 2017-07-14
Fixed request for switch state from node

4ib3r:
Before the fix, when the node request the state of the switch, wrong id was builded.


gizmocuz:
Thanks!

PR #1695, opened 2017-07-16
Webserver: Prevent logfile flooding for incoming connects

hvbommel:
Changed loglevel to LOG_TRACE to Prevent log flooding while maintaining regular logging. 

Example of repeated log messages:
2017-07-16 11:09:48.327 Incoming Domoticz connection from: 192.168.2.230
2017-07-16 11:10:18.331 Incoming Domoticz connection from: 192.168.2.230
2017-07-16 11:10:48.334 Incoming Domoticz connection from: 192.168.2.230
2017-07-16 11:11:18.338 Incoming Domoticz connection from: 192.168.2.230
2017-07-16 11:11:48.599 Incoming Domoticz connection from: 192.168.2.230
2017-07-16 11:12:18.653 Incoming Domoticz connection from: 192.168.2.230
2017-07-16 11:12:48.657 Incoming Domoticz connection from: 192.168.2.230
2017-07-16 11:13:18.660 Incoming Domoticz connection from: 192.168.2.230

### DIFF
--- a/tcpserver/TCPServer.cpp
+++ b/tcpserver/TCPServer.cpp
@@ -63,7 +63,7 @@ void CTCPServerInt::handleAccept(const boost::system::error_code& error)
 		}
 
 		new_connection_->m_endpoint=s;
-		_log.Log(LOG_STATUS,"Incoming Domoticz connection from: %s", s.c_str());
+		_log.Log(LOG_TRACE,"Incoming Domoticz connection from: %s", s.c_str());
 
 		connections_.insert(new_connection_);
 		new_connection_->start();


gizmocuz (commented):
Can not accept this
Please check if already logged one hour ago.. of not.. log it

hvbommel:
I was indeed thinking of doing that, in fact I started it and then reverted locally. Its because it could be that different IP addresses are connecting. So to do it properly I would have to create a history list of IP address that did connect, and where logged with timestamp, and then bypass those but log the others that we see the first time. I taught that is a bit too much. Have we done conditional logging before in other modules ? Please let me know what you think.

The problem is because I have master domoticz connecting to the slave every 30 seconds.

gizmocuz:
For once in a while connecting we always want this log
Repeated every hour is indeed irritating

hvbommel:
It is repeated even more often, every 30 seconds. That is why it is flooding the log making it unusable.

 But if we need the connection log for security reasons I indeed see no simple & proper solution. Maybe we can have a new log level "security". Those who worry about security can then select that level, and accept the noisy logfile. It would be the level just above loglevel "status". Selecting level "security" would then include error+status+security. I am not sure if I want to fool around with that, it may break code that does not properly use the log level defines but use the values (0,1,2,3) 

Do you see a solution ? Or should we accept the "flooding" and close this PR, ie forget & live with it? 

PS: The noisy logfile lives on the slave. My main logfile lives on the master. So I could indeed forget the slave logging but if everyone does that its useless to log in the first place.

gizmocuz:
Maybe log when the cookie needs to be created and at least once a day from the same ip

hvbommel:
We could do something like have a log filter class that takes a string as input, in this case the ip address string, and returns true of false if we need to log. Return false if the string is on the list, true otherwise. 1st time called after midnight it would clean all lists so new connects would then be logged. Does this sound right ?

hvbommel:
I will close it for now and think of a proper solution. When I have that I will open a new PR. Sounds good ? 

If someone wants to step in and do it ... even better. In the nightly summary report we could say: "the original message" occurred x times"

gizmocuz:
Thanks for you time of course... for now we will close this then

hvbommel:
Thanks ! I will enter a new PR once I have a proper solution.

PR #1697, opened 2017-07-17
Fix elemental theme's custom.css for drop down type selector switches

odelma:
Drop down list of a selector switch will align itself below the block in smaller screens if the width is forced to 200px. There does not seem to be any similar setting in other themes.


gizmocuz:
Thanks!

PR #1700, opened 2017-07-19
TCPServer: Prevent log flooding for same IP-address

hvbommel:
TCPServer logging adjustment:

To prevent logfile flooding, when domoticz reconnects for same IP-address occur, this code change will make sure that:
1. The connect for a given IP address is always logged on first the time connect.
2. After that logging connects, for the same IP-address for next 24 hours, are skipped.
3. The first connect that occurs after 24 hours of silence will be logged again.
4. When the user has debug level is set to LOG_TRACE logging will as before, i.e. unfiltered. 

NOTE:
In the future we may implement this as a logging helper class in case the functionality is needed more often in other modules. In that case that class will need more control, like duration settings (instead of a fixed 24 hours setting), a summary report creation and more. For now I like to keep it simple and just do exactly what we need right now.  

See #PR1695. That PR was closed while working on this one.

### DIFF
--- a/tcpserver/TCPServer.h
+++ b/tcpserver/TCPServer.h
@@ -91,7 +91,7 @@ private:
 
 	CTCPClient_ptr new_connection_;
 
-	bool logFirstTime(const std::string &ip_string);
+	bool IsUserHereFirstTime(const std::string &ip_string);
 	std::vector<log_info> m_incoming_domoticz_history;
 };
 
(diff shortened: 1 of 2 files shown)


tcpserver/TCPServer.h
```diff
@@ -84,6 +90,9 @@ class CTCPServerInt : public CTCPServerIntBase {
 	boost::asio::ip::tcp::acceptor acceptor_;
 
 	CTCPClient_ptr new_connection_;
+
+	bool logFirstTime(const std::string &ip_string);
```
gizmocuz: All functions start with a Uppercase letter (camel system)
Name is not very clear, could you rename it to (for instance) IsUserHereFirstTime(...)

hvbommel:
Thanks for the feedback, done.

PR #1702, opened 2017-07-20
EventSystem update single instances of uservars / scenesgroups

jvandenbroek:
Only update single instances in uservariables / scenesgroups instead of reloading complete tables every device update. Scenesgroups were only updated once a minute, now on switch scene to make sure it's up-to-date.

### DIFF
--- a/main/EventSystem.cpp
+++ b/main/EventSystem.cpp
@@ -150,6 +150,8 @@ void CEventSystem::StartEventSystem()
 
 	LoadEvents();
 	GetCurrentStates();
+	GetCurrentScenesGroups();
+	GetCurrentUserVariables();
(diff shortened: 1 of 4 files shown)


gizmocuz:
Thanks!

gizmocuz:
There seems to be problems with this patch on the forum

gizmocuz:
http://www.domoticz.com/forum/viewtopic.php?f=6&t=18468

jvandenbroek:
Let me see what causes this, probably something with user var being nil at startup (which shouldn't), let me see if I can quickly patch it.

jvandenbroek:
https://github.com/domoticz/domoticz/pull/1706 should fix it

PR #1703, opened 2017-07-21
HardwareMonitor fix for busybox df command

jvandenbroek:
Extended function ExecuteCommandAndReturn with returncode to determine if the command errored. Busybox doesn't seem to support extra parameters, so this should fix issues on busybox based installations.

### DIFF
--- a/main/Helper.cpp
+++ b/main/Helper.cpp
@@ -478,7 +478,7 @@ double ConvertTemperature(const double tValue, const unsigned char tSign)
 	return RoundDouble(ConvertToFahrenheit(tValue),1);
 }
 
-std::vector<std::string> ExecuteCommandAndReturn(const std::string &szCommand, int *returncode)
+std::vector<std::string> ExecuteCommandAndReturn(const std::string &szCommand, int &returncode)
 {
 	std::vector<std::string> ret;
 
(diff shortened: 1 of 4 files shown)


main/Helper.cpp
```diff
@@ -478,7 +478,7 @@ double ConvertTemperature(const double tValue, const unsigned char tSign)
 	return RoundDouble(ConvertToFahrenheit(tValue),1);
 }
 
-std::vector<std::string> ExecuteCommandAndReturn(const std::string &szCommand)
+std::vector<std::string> ExecuteCommandAndReturn(const std::string &szCommand, int *returncode)
```
gizmocuz: please change this to:
std::vector<std::string> ExecuteCommandAndReturn(const std::string &szCommand, int &returncode)

jvandenbroek: Ehm, that will not work? Then it would be better to do it this way:
```
int ExecuteCommandAndReturn(const std::string &szCommand, std::vector<std::string> &szResult)
return returncode
```

??

gizmocuz: Sure it does...

gizmocuz: just parse the variable, not a pointer to the variable

gizmocuz: void test(int &var1)
{
var1=1234;
}

int myint=0;
test(myint)

now myint = 1234

jvandenbroek: Oh lol yes, too easy, sorry. Not sure how I flipped that logic.

gizmocuz:
Thanks!

PR #1706, opened 2017-07-22
EventSystem fix Lua setting uservariable

jvandenbroek:
Fix for EventSystem lockups, setting the global uservariables mutex would lock up changing variables in-script (double lockup). Now only set a shared lock on parts needed. Set the full mutex on Python, to make sure we won't change any logic inside Python execution.

http://www.domoticz.com/forum/viewtopic.php?f=6&t=18468
http://www.domoticz.com/forum/viewtopic.php?f=6&t=18467

### DIFF
--- a/main/EventSystem.cpp
+++ b/main/EventSystem.cpp
@@ -1381,8 +1381,6 @@ void CEventSystem::EvaluateEvent(const std::string &reason, const uint64_t Devic
 	if (!m_bEnabled)
 		return;
 
-	boost::unique_lock<boost::shared_mutex> uservariablesMutexLock(m_uservariablesMutex);
-
 	std::vector<std::string> FileEntries;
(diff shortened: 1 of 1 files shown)


jvandenbroek:
However, Python may also be affected by this.. Only tested with Lua, let me see what Python does.
@gizmocuz is there a Python function to set a user variable? Can't find any documentation about this. If it's only parsing variables, we're good to go.

gizmocuz:
Best to ask dnwpro

jvandenbroek:
According to this post: https://www.domoticz.com/forum/viewtopic.php?f=65&t=16308&p=122025&hilit=uservariables#p122020 it seems this is indeed not possible from Python. Is that correct @dnpwwo ? I think we should merge this fix asap.

gizmocuz:
Thanks

PR #1707, opened 2017-07-22
Sysfs GPIO: Max pin number is now 1023 for Odroid C2 sysfs support

hvbommel:
On the forum it was reported by aaron that "Odroid C2 sysfs gpio"  is using sysfs gpio pin numbers higher then 31 (the raspberry 3B case). Because of this we did not detect these pins for use within domoticz. To fix this problem the number has now been increased to support pin numbers up to 1023. See forum  http://www.domoticz.com/forum/viewtopic.php?f=6&t=18323&p=142289#p142289


gizmocuz:
Thanks! Maybe it is documented somewhere

PR #1708, opened 2017-07-22
Sysfs Gpio: Do not log forced updates without db changes (forum req)

hvbommel:
Do not log when a gpio forced update is done without database change, i.e. a trigger without a state change. (this is done in master slave configuration, 30 seconds after startup). Requested on forum by pvgorp.


gizmocuz:
Thanks

PR #1714, opened 2017-07-24
Blockly memory leak fix

jvandenbroek:
This should fix Blockly memory leak as reported here: https://www.domoticz.com/forum/viewtopic.php?f=6&t=16367

Could reproduce it with Valgrind, with this change there is no longer leakage reported.
Did a little reformatting to make reading easier.

### DIFF
--- a/main/EventSystem.cpp
+++ b/main/EventSystem.cpp
@@ -1544,6 +1544,8 @@ void CEventSystem::EvaluateEvent(const std::string &reason, const uint64_t Devic
 lua_State *CEventSystem::CreateBlocklyLuaState()
 {
 	lua_State *lua_state = luaL_newstate();
+	if (lua_state == NULL)
+		return NULL;
 
 	// load Lua libraries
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thanks.. but could you undo all formatting? Hard to see what now actually changed....

jvandenbroek:
Done, it's not really much.. Just removed a double allocation. And yes, the scripts still work ;)

gizmocuz (commented):
Thanks but we probably still want to check if it was allocated correctly.. hence the checking for NULL

jvandenbroek:
It's already checked in function EvaluateBlockly before being called, thus redundant.
Wait, that's not true. But how would it be NULL after it just got allocated? EvaluateLua function uses the same logic.

jvandenbroek:
Suppose it doesn't harm either, put it back.

gizmocuz (commented):
Thanks

gizmocuz:
Thanks!

PR #1715, opened 2017-07-25
EventSystem fix use absolute instead of relative paths

jvandenbroek:
szUserDataFolder is initiated after EventSystem's global variables, making the paths relative instead of absolute (that seems to work fine on most installations, but not all). See comment here: https://github.com/domoticz/domoticz/commit/74b59052883611b869e17b482c7fb95bbcdafad3#commitcomment-23273867

### DIFF
--- a/main/EventSystem.cpp
+++ b/main/EventSystem.cpp
@@ -36,25 +36,12 @@ extern "C" {
 extern std::string szUserDataFolder;
 extern http::server::CWebServerHelper m_webservers;
 
-std::string m_printprefix;
-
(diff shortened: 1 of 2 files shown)


Jumbotroll:
I think this also solve the Segmentation fault on Linux/ synology platform.
https://www.domoticz.com/forum/viewtopic.php?f=6&t=18341
Edit: tested on Synology DS712+ arch-x86 and domoticz started normally again!

gizmocuz:
Thanks!

PR #1718, opened 2017-07-25
Update manufacturer_specific.xml

tlpeter:
Added product type for new fgcd001 (Fibaro CO sensor)


gizmocuz:
We merge from time to time with the openzwave repo?
Is this already in the ozw repo?

tlpeter:
I did the same there and then realized that it would be taken from there.
So i guess it is not needed.
The same applies for #1719 i think.



Op 25 jul. 2017 17:05 schreef "gizmocuz" <notifications@github.com>:

gizmocuz:
True... Okey let us know when it is merged...

tlpeter:
It is merged and i see you already put it in but it has changed a little bit.
Parameter 3 and for do have a max of 7 instead of 4.

gizmocuz:
you have to contact OZW for this...

tlpeter:
I did and was changed after you pulled it from there i guess.

PR #1720, opened 2017-07-26
Reworked GCM data to notification message

galadril:
Reworked GCM data to notification message

- Data message won't display a notification on your device, should use a notification message for that instead.
- For receiving push notifications on the background on should include content_available = true inside the notification key.

new gcm json message body is going to be:

```

{
  "registration_ids" : ["senderid"],
  "notification" : {
    "subject" : "test subject",
    "body" : "test body",
    "extradata" : "test data",
    "deviceid" : "1",
    "message" : "test message",
    "priority" : "high",
    "content_available" : true,
  },
}

```


gizmocuz:
Thanks!

PR #1725, opened 2017-07-27
Add data and notification info

galadril:
For Android we need the 'data' json object, but iOS only receives a notification when it contains a 'notification' json object. 

You need to send "data" messages to the client and then in your service, you will then always receive the callback, whether the app is in the foreground or background.

Read more here: https://firebase.google.com/docs/notifications/android/console-audience

So im changing the body to:

```

{
  "registration_ids" : ["senderid"],  
  "data" : {
    "subject" : "test subject",
    "body" : "test body",
    "extradata" : "test data",
    "deviceid" : "1",
    "message" : "test message",
    "priority" : "high",
  },
  "notification" : {
    "subject" : "test subject",
    "body" : "test body",
    "extradata" : "test data",
    "deviceid" : "1",
    "message" : "test message",
    "priority" : "high",
    "content_available" : true,
  },
}

```

### DIFF
--- a/notifications/NotificationGCM.cpp
+++ b/notifications/NotificationGCM.cpp
@@ -50,8 +50,10 @@ bool CNotificationGCM::SendMessageImplementation(
 		sstr << "\"" << sd[0] << "\"";
 		ii++;
 	}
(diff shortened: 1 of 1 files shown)


galadril:
Not going to work also....

galadril:
im afraid we need data for android.. and notification for ios :(

PR #1730, opened 2017-07-30
EventSystem queue thread for events

jvandenbroek:
Create a queue thread for events, same code base as found in mainworker (push pop). Running it for a few days without problems.


gizmocuz:
Thanks!

PR #1731, opened 2017-07-31
build: Detect and allow external tinyxpath library usage

mooninite:
Signed-off-by: Michael Cronenworth <mike@cchtml.com>

While packaging domoticz for Fedora I needed to unbundle tinyxpath. This should take care of it in a way to allow users to retain the source you ship by default, but if they have it present their system provided library will take precedent.


mooninite:
I saw the Travis CI build failed... using CMake 3.2. Version 3.2 is over 2 years old now. I don't have any build problems on CMake 3.9.

gizmocuz:
Maybe we could add tinyxpath to the path so we can use everywhere 

#include <tinyxml.h>

Is this possible ?

mooninite:
I'm not sure what you are asking. You wish to add the include to every source file?

I want to bring up that tinyxml and tinyxpath are two different projects. The tinyxml.h found in tinyxpath should not be used in place of tinyxml. Tinyxpath is a library that *extends* TinyXML. The source found in domoticz appears to combine the two projects, but they are not the same project.

gizmocuz:
No sorry, just to add the Tinyxpath include path to the include defines (like we do with other libraries like openzwave)
then we van use

#include <tinyxml.h>

So in the cmake file, try to first find the library/headers... like its done qith SQLite

Ps. Why is it a problem that domoticz its own library... yes they may differ for various reasons..

mooninite:
I am including the path in my patch as it is done for SQLite.

The reason why the build succeeds for other libraries, like SQLite, is that it lives behind a option() check. If your Travis CI build would run with those options enabled it would fail just like my patch is failing. The same pkg_check_modules() call is used for them.

I can place the TinyXML check behind an option() call if you would like, but I would suggest upgrading your build system used in Travis CI. Either way I'll work with it.

gizmocuz:
I'm afraid we can not upgrade the build system
If you look at the travis file, it just installs 'cmake' regardless the version number, and i suspect other platforms might use an older version as well.

But again, why does it matter that domoticz uses its own tinyxml ? the code size difference is quite minimal.

If you look in SQLHelper.cpp, sqlite is included with

`#include <sqlite3.h>`

Without if/then/else statements

Would be great if we can do the same for this library, so if you add the library to the 'path' (-D) i think we should be able to do the same...
If tinyxml is found in the system -> use that path/lib (static), else use the folder from domoticz
... Now... as the folder of domoticz does not have a library.. we need to make a tinyxml cmake file for it
so we can compile it if needed... (like sqlite)

xbeaudouin:
Maybe you can change the content of :  
../tinyxpath/tinyxml.h

```
 #ifdef WITH_EXTERNAL_TINYXPATH
 #	include <tinyxml.h>
 #else
  // Add the content of tinyxml.h
 #endif
```

I tried to do that with jsoncpp on my side but I got some race (sig11).

But If this patch is added to mainstream I will be happy also on my side for my FreeBSD port.

mooninite:
Using angle-brackets (#include <foo.h>) to reference locally sourced files is not good programming style.

https://gcc.gnu.org/onlinedocs/cpp/Include-Syntax.html

I will place the option behind an option() block so that the CI test passes. Expect an updated patch soon.

"But again, why does it matter that domoticz uses its own tinyxml ? the code size difference is quite minimal."

Because it is an external library that may receive updates -- such as security updates -- and as such needs to be able to be linked against the upstream copy rather than a bundled copy. This is a common, good programming practice in the FOSS world.

I'm using domoticz with all external libraries (sqlite3, openzwave, tinyxml, tinyxpath, etc) on my production system without any issues.

gizmocuz:
its an external library that could get incompatible with domoticz

gizmocuz:
I see one solution, create a cmake project for tinyxml, and use the same mechanism as SQLite... not very interested in adding if  then else code for 1kb of extra code...
Maybe also read why angle brackets are used for as well ? It makes perfect sense in this case as indeed its an external library ... but in the source repo of domoticz yes
And thats where 'Include Paths' are for

mooninite:
External libraries typically follow SONAME (ABI) or API versioning to prevent incompatibilities. You should not be afraid of external changes. :)

You can set version minimum and/or maximum on library dependencies in the CMake script to prevent someone from using unsupported versions.

gizmocuz:
So yes, we need to build tinyxml as we do with sqlite if tinyxml is not found on the system... thats what i mean...
If i have time i can look into this myself, but if you have some time, that would be great!

mooninite:
I don't agree with using angle-brackets for even local libraries, but I don't have time to argue further.

I've made the necessary changes to the patch. I hope we can compromise on this for now. I hope to contribute more to this project in the future.

mooninite:
The Windows build is failing to find the includes. I don't have a Windows machine and it has been years since I have touched Visual Studio so I'm not sure how to address the include path there.

gizmocuz:
Thanks for the patch, i will look into this asap
That you do not agree with using angle-brackets at all, sorry that i can't change your mind ;)
As for VStudio, this is a cross platform project and code changes should be tested on both platforms whenever possible
Visual Studio 2017 community you can download for free
You can use virtualbox to install/test with, or when using a mac, parallel desktop for example
Thanks again!

hardware/AnnaThermostat.cpp
```diff
@@ -1,3 +1,5 @@
+#include <tinyxml.h>
```
gizmocuz: Could you move this line to the same old position ? (around line 11)
As you might know, it is needed to have stdafx.h as the first line
It is also a good idea to always include the class header as second (or in projects where no precompiled headers are in use, as first)

gizmocuz: Please do so for all files/changes as i see this in other files as well

main/LuaCommon.cpp
```diff
@@ -1,3 +1,6 @@
+#include <tinyxml.h>
```
gizmocuz: Do we need this include? It was not needed previously....

main/LuaHandler.cpp
```diff
@@ -1,3 +1,6 @@
+#include <tinyxml.h>
```
gizmocuz: Do we need this include? It was not needed previously....

msbuild/domoticz.vcxproj
```diff
@@ -618,6 +618,8 @@
     <ClInclude Include="..\main\WindCalculation.h" />
     <ClInclude Include="..\smtpclient\SMTPClient.h" />
     <ClInclude Include="..\sqlite\sqlite3.h" />
+    <ClInclude Include="..\tinyxpath\tinyxml.h" />
```
gizmocuz, line 626: I think you also need to add the tinyxpath folder to the additional includes folder for both debug/release

hardware/AnnaThermostat.cpp
mooninite: I'll move to retain the old position, but order is not important here. Fixed.

main/LuaCommon.cpp
mooninite: It came from the original version of my work. Not needed now. Removed.

msbuild/domoticz.vcxproj
mooninite: I've added the directory (after finding the correct path specification) and the builds are now succeeding.

gizmocuz:
Is it possible to detect tinyxml in the opposite way ?
First check if it is used/found in the system, if yes, use the external one, else use the buildin one
This way, you do not have to specify a YES/NO option for this in CMake

mooninite:
Yes there is, and that is what my original patch did. However, the version of CMake your CI uses has a bug that prevents using the method I used.

The alternative to using pkg_check_modules() is to use find_package(), however there is no native CMake package configuration file provided for TinyXML or TinyXPath. I would have to create one for each.

gizmocuz:
Could it be that you merged another branch/patch into this one ? (python)
Will this patch work with old cmake versions ?

mooninite:
Github messed up the pull request. I've made PR 1759 to fix it. I'll close this one.

Yes, it worked with old cmake versions. The CI builds passed.

PR #1732, opened 2017-07-31
Fixed error code and carriage return when reloading server

sbouchex:


gizmocuz:
I dont see any reason why we should exit with a code here ?

log_end_msg 3
exit 3

Looking at various other startup scripts, this is also not happening... why would you need/want this here ?

gizmocuz:
Most scripts run 

$0 stop
$0 start

sbouchex:
OK, I will remove the error code

gizmocuz:
Okey, but then nothing will be changed ;)

sbouchex:
yes, the missing carriage return ;-)

gizmocuz:
hehe sorry, okey i wait for you patch then

gizmocuz:
Sorry.. which missing carriage return ?

PR #1734, opened 2017-08-01
Fixed carriage return in "reload" action

sbouchex:


domoticz.sh
```diff
@@ -115,6 +115,9 @@ case "$1" in
         PID=$(pidof_domoticz) || true
         if [ "${PID}" ]; then
                 kill -HUP $PID
+                log_end_msg 0
+        else
+                log_end_msg 3
```
gizmocuz: Thanks, but should this not be 1 instead of 3, why 3 ? Is this documented somewhere?

gizmocuz:
Thanks!

PR #1736, opened 2017-08-02
Notification Last update SystemUptime function / check

jvandenbroek:
When my Pi has been down for a while, it will still send a lot of notifications after startup, because of Domoticz getting the old system time before NTP syncs the system time (so m_StartTime is much farther in the past). I've added a SystemUptime function, which I also intent to use for Lua / dzVents scripting later on, for which reason I placed it in the Helper.cpp.

Now the notification lastupdate function first checks if system uptime is less than sensor timeout and returns if true.

### DIFF
--- a/main/Helper.cpp
+++ b/main/Helper.cpp
@@ -22,6 +22,17 @@
 #include "RFXtrx.h"
 #include "../hardware/hardwaretypes.h"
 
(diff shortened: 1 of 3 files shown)


gizmocuz:
Thanks!

PR #1740, opened 2017-08-02
EventSystem fix for Lua scripts no longer working

jvandenbroek:
Fix a small copy/paste error causing Lua and dzVents scripts no longer to work ;)

### DIFF
--- a/main/EventSystem.cpp
+++ b/main/EventSystem.cpp
@@ -3818,7 +3818,7 @@ bool CEventSystem::processLuaCommand(lua_State *lua_state, const std::string &fi
 		m_sql.AddTaskItem(_tTaskItem::SendSMS(1, luaString));
 		scriptTrue = true;
 	}
-	else if (lCommand == "TriggerIFTTT" != std::string::npos)
+	else if (lCommand == "TriggerIFTTT")
 	{
 		std::string luaString = lua_tostring(lua_state, -1);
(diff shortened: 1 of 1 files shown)


gizmocuz:
Sorry, already fixed it a few minutes ago... also fixed some compile warnings as visual studio complained about them, and also this copy/paste issue

jvandenbroek:
No worries, good you've solved it already :)
I only compile on Linux, which is why I don't get any VB compiler warnings... Somehow other compilers are more forgiving for using unsigned and signed together, probably would like have my Gcc compiler also warn about these! Might be a good idea to set up some Windows VM with VB installed, would help if you've a one click tutorial for me ;)

Btw just discovered that GetTickCount only counts up to about 49 days, seems logical since it's / 1000, didn't realize (just used an example from elsewhere). GetTickCount64 solves this, should be fine if I cast it, but it's for Vista and higher. Do we still support 2000/XP?

gizmocuz:
No we don't support XP anymore for quite some time for this vary reason (MQTT needed that too)

You can download/install virtualbox (from Oracle), install windows 10 and install VStudio 2017 community edition
Or just get yourself a native windows PC (like a Shuttle XH61v, not expensive, and small)
Or dualboot, or even better, activate ubuntu under windows 10 (its native these days)

PR #1741, opened 2017-08-03
Helper use GetTickCount64 instead of GetTickCount

jvandenbroek:
Let's see what AppVeyor has to say about this before merging.. Will get a test environment soon! (I hope)

### DIFF
--- a/hardware/DomoticzTCP.cpp
+++ b/hardware/DomoticzTCP.cpp
@@ -55,32 +55,6 @@ bool DomoticzTCP::IsValidAPIKey(const std::string &IPAddress)
 }
 #endif
 
-#if defined WIN32
-int inet_pton(int af, const char *src, void *dst)
-{
-  struct sockaddr_storage ss;
-  int size = sizeof(ss);
-  char src_copy[INET6_ADDRSTRLEN+1];
-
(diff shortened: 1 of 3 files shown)


jvandenbroek:
OK, annoying, too many things to fix before it's compatible with > Vista. Look into this later when I have a decent test environment.

At least it seems to build, but too many new warnings introduced...

gizmocuz:
Why ? the only thing you have todo is in helper.cpp/h return a uint64_t and replace GetTickCount with GetTickCount64
Its not complaining here...

gizmocuz:
this should do it
https://github.com/domoticz/domoticz/commit/602febb3c4ddfd7f260557df35a6c358eb504434

jvandenbroek:
Hmm how could that work if the minimal version is set to XP , which doesn't support this function? That's what I tried to change, setting Vista as minimal version app wide, but as you can see in the AppVeyor logs, it will complain on many other parts.

gizmocuz:
Hmmm your right ;) sorry, reverted it

gizmocuz:
Fixed in latest commit ;)
https://github.com/domoticz/domoticz/commit/645a16cf3939906777ec28b31d3cf458aa4ce1a3

compiles correct

PR #1742, opened 2017-08-03
Logging enhancement

sbouchex:
Added hardwareId, deviceId and zwave nodeId into logged message to ease troubleshooting

### DIFF
--- a/main/mainworker.cpp
+++ b/main/mainworker.cpp
@@ -14,6 +14,7 @@
 #include "../httpclient/HTTPClient.h"
 #include "../webserver/Base64.h"
 #include <boost/algorithm/string/join.hpp>
+#include <boost/lexical_cast.hpp>
 
 #include <boost/crc.hpp>
 #include <algorithm>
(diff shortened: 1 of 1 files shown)


gizmocuz:
Maybe you should add this in the zwave class (as a debug string/trace)
Because the node id can not (always) be calculated this way
Besides, if you name your sensors, you know it from the devicename
So for 'debug' purposes, we have a new 'trace' level in the log, if trace is enabled, then do a trace log
Zwave nodeid is currently only displayed when you compile in debug mode

sbouchex:
I will move the ID -> ZWaveModeID calculation to to the zwave class part.

I don't see the point about trace level : This log already exists in the product and it should be a functional change to move the TRACE level (which is find if explained), so the ZNodeID is always visible in the interface (AFAIK), so it's the debug only. Am I right ?

gizmocuz:
there is no need for calculation as you get the instance/index id already in the add/update functions in the zwave class
The trace log is also working in release mode...

sbouchex:
Yes but from MainWorker::ProcessRXMessage, I don't have this kind of details and I don't want to point to the ZWave structure to find the object and moving this logging elsewhere just for the ZWave device does not make the thing consistent.

gizmocuz:
but the method you used for your calculation is not correct for all sensor types...
I have no idea what or why you want this, and i dont think it is needed
IF you want to find out from what Node/Index/Instance a device comes from, then all i do is go to the devices tab, in the filter i type the name (or part of ) my zwave hardware name then a space en then for example the word 'light' like 'zwave light', now you have a good overview of what comes from where

Also there are people that move from one platform (software) to domoticz, that have 40+ zwave devices... yes it is a nightmare at first to label them right... nothing is impossible

So what we do want is in the openzwave setup, when you select a node, that you will see all associated devices in Domoticz, if that is where you are going to

sbouchex:
OK in that case, I will remove the ZWave nodeID and that's it ?

gizmocuz:
i would leave the logging in the mainworker as it is...
If you run in debug mode then you will see what i mean with the logging

Issue #1509, opened 2017-04-30
LUA domoticz_applyXPath() inconsistent results


nmaster2042:
Domoticz: V3.3792
Platform: Odroid C2
OS: Debian jessie

Hi,

My last issue has been misunderstood and quickly closed: I wasn't requesting help, but was reporting a bug, so please check before closing it.

There is an issue with the domoticz_applyXPath() function in LUA.

There is actually something wrong regarding results returned by domoticz_applyXPath(), a valid Xpath pattern returns nothing in the case below.

Here is a sample of y XML source:

`<!-- --> <root><node><dateIndice>2017-04-29</dateIndice><agglomerationCodeInsee>97209</agglomerationCodeInsee><agglomeration>FORT-DE-FRANCE</agglomeration><valeurIndice>6</valeurIndice><SousIndiceO3>2</SousIndiceO3><SousIndiceNO2>1</SousIndiceNO2><SousIndicePM10>4</SousIndicePM10><SousIndiceSO2/><Commentaire/><sourceAASQA>http://www.lcsqa.org/aasqa/code/39</sourceAASQA></node><node><dateIndice>2017-04-29</dateIndice><agglomerationCodeInsee>76540</agglomerationCodeInsee><agglomeration>ROUEN</agglomeration><valeurIndice>4</valeurIndice><SousIndiceO3>4</SousIndiceO3><SousIndiceNO2>2</SousIndiceNO2><SousIndicePM10>3</SousIndicePM10><SousIndiceSO2>1</SousIndiceSO2><Commentaire/><sourceAASQA>http://www.lcsqa.org/aasqa/code/25</sourceAASQA></node><node><dateIndice>2017-04-29</dateIndice><agglomerationCodeInsee>27229</agglomerationCodeInsee><agglomeration>EVREUX</agglomeration><valeurIndice>4</valeurIndice><SousIndiceO3>4</SousIndiceO3><SousIndiceNO2>1</SousIndiceNO2><SousIndicePM10>3</SousIndicePM10><SousIndiceSO2>1</SousIndiceSO2><Commentaire/><sourceAASQA>http://www.lcsqa.org/aasqa/code/25</sourceAASQA></node> ... ... </root>`

In order to know how to parse it I did it on command line like this:

First I get the XML data from server:

`$ wget -O test.xml http://www.lcsqa.org/surveillance/indices/prevus/jour/xml/`

Then I use xmllint with xpath to get global indice of air quality in my area:

[...]

PR #1744, opened 2017-08-05
Replace XPATH processor with PugiXML

sbouchex:
In order to fix XPATH issue (see #1509), use pugixml (xml + xpath) parser


gizmocuz:
Thanks, can you delete the old library

sbouchex:
yep but forget about my PR, it includes other commits :-(

PR #1745, opened 2017-08-05
Replaced xml/xpath parser by pugixml

sbouchex:
Replace xml/xpath parser to fix bugs (see #1509)

Note for @gizmocuz  : Removing the old library is not doable as it's part of tinyxml used all accross the product and it may be risky to replace it and openzwave is also using tinyxml.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -466,6 +466,7 @@ tinyxpath/xpath_stream.cpp
 tinyxpath/xpath_stack.cpp
 tinyxpath/xpath_static.cpp
 tinyxpath/xpath_syntax.cpp
+pugixml/pugixml.cpp
 )
 add_executable(domoticz ${domoticz_SRCS})
 
--- a/main/LuaCommon.cpp
+++ b/main/LuaCommon.cpp
@@ -16,6 +16,8 @@ extern "C" {
 #endif
 }
 
+#include "../pugixml/pugixml.hpp"
+
 #include "../tinyxpath/xpath_processor.h"
 #include "../json/json.h"
 #include "SQLHelper.h"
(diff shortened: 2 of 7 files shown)


gizmocuz:
Okey but would it not be best to use one library?

sbouchex:
I will see but in anycase, tinyxml is part of openzwave, so you will have 2 xml parsers anyway (if openzwave is enabled)
I'll try to cleanup things

gizmocuz:
What is the other one? Do we have 3 now with this patch?

sbouchex:
if the tinyxpath is removed, there will be 2 (including openzwave) but I have to migrate the existing code to pugixml.

gizmocuz:
Do you mean two XML parser libraries or two classes that are using XML? I thought we had only one XML parser at the moment

sbouchex:
I managed to replace tinyxml mostly everywhere. However, openzwave/owcp relies on tinyxml and it would be difficult to replace with the new parser.

Option 1 : Keep both parsers like I initially did ==> safer option
Option 2 : Copy tinyxml to owcp folder to keep the existing code and migrate other devices to the new parser ==> more dangerous option as I'm not able to validate anything

gizmocuz:
I think i like pugixml... hope we can phase out the old xml library

gizmocuz:
Sorry, i get compile errors on my arm64 build machine, already tried full build. clearing etc... it keeps stopping.
Going to need to revert this patch

> [ 21%] Building CXX object CMakeFiles/domoticz.dir/pugixml/pugixml.cpp.o
> /home/ubuntu/dev-domoticz/pugixml/pugixml.cpp: In member function ‘void pugi::impl::{anonymous}::xml_buffered_writer::flush(const char_t*, size_t)’:
> /home/ubuntu/dev-domoticz/pugixml/pugixml.cpp:3704:8: internal compiler error: in convert_move, at expr.c:345
>    void flush(const char_t* data, size_t size)
>         ^
> Please submit a full bug report,
> with preprocessed source if appropriate.
> See <file:///usr/share/doc/gcc-4.9/README.Bugs> for instructions.
> Preprocessed source stored into /tmp/ccb6hV6H.out file, please attach this to your bugreport.
> CMakeFiles/domoticz.dir/build.make:4838: recipe for target 'CMakeFiles/domoticz.dir/pugixml/pugixml.cpp.o' failed
> make[2]: *** [CMakeFiles/domoticz.dir/pugixml/pugixml.cpp.o] Error 1
> CMakeFiles/Makefile2:64: recipe for target 'CMakeFiles/domoticz.dir/all' failed
> make[1]: *** [CMakeFiles/domoticz.dir/all] Error 2
> Makefile:137: recipe for target 'all' failed
> make: *** [all] Error 2
>

PR #1748, opened 2017-08-07
Blinds Percentage Inverted invert levels send/received to/from hardware

jvandenbroek:
This is an attempt to fix this: http://www.domoticz.com/forum/viewtopic.php?f=6&t=9883
I've tested it with my screen connected the correct way, which resulted in everything being inverted. This seems to work, if I set level 20 it will send 80 to hardware. Values received from hardware (OZW) are converted back, so it get's level 80, it will set level 20.

The only thing that doesn't work correctly is setting the slider to 0, because this will always trigger an Off command instead of set level 0. Can't catch that without digging too deep in the code. For level 100 this works fine, it will send an Off command instead now.

### DIFF
--- a/hardware/ZWaveBase.cpp
+++ b/hardware/ZWaveBase.cpp
@@ -136,7 +136,7 @@ void ZWaveBase::InsertDevice(_tZWaveDevice device)
 	device.string_id=GenerateDeviceStringID(&device);
 
 	bool bNewDevice=(m_devices.find(device.string_id)==m_devices.end());
-	
+
 	device.lastreceived=mytime(NULL);
 #ifdef _DEBUG
 	if (bNewDevice)
@@ -338,12 +338,12 @@ void ZWaveBase::SendDevice2Domoticz(const _tZWaveDevice *pDevice)
 	ID3=(unsigned char)pDevice->nodeID&0xFF;
 	ID4=pDevice->instanceID;
 
-	char szID[10];
-	sprintf(szID,"%X%02X%02X%02X", ID1, ID2, ID3, ID4);
-
 	unsigned long lID = (ID1 << 24) + (ID2 << 16) + (ID3 << 8) + ID4;
+	char szID[10];
+	sprintf(szID, "%08x", (unsigned int)lID);
 
 	int BatLevel = 255;
+
 	if ((pDevice->hasBattery) && (pDevice->batValue != 0))
 	{
 		BatLevel = pDevice->batValue;
@@ -351,6 +351,14 @@ void ZWaveBase::SendDevice2Domoticz(const _tZWaveDevice *pDevice)
 
 	if ((pDevice->devType==ZDTYPE_SWITCH_NORMAL)||(pDevice->devType==ZDTYPE_SWITCH_DIMMER)||(pDevice->devType== ZDTYPE_CENTRAL_SCENE))
 	{
+		bool bInvertLevel = false;
+		std::vector<std::vector<std::string> > result;
+		result = m_sql.safe_query(
+			"SELECT ID FROM DeviceStatus WHERE (DeviceID == '%q') AND (Type == %d) AND (SubType == %d) AND (SwitchType == %d)",
+			szID, pTypeGeneralSwitch, sSwitchGeneralSwitch, STYPE_BlindsPercentageInverted);
+		if (result.size() == 1)
+			bInvertLevel = true;
+
 		//Send as pTypeGeneralSwitch, sSwitchGeneralSwitch
 		_tGeneralSwitch gswitch;
 		gswitch.subtype = sSwitchGeneralSwitch;
@@ -367,17 +375,24 @@ void ZWaveBase::SendDevice2Domoticz(const _tZWaveDevice *pDevice)
 
 		// Now check the values
 		if (level == 0)
+		{
 			gswitch.cmnd = gswitch_sOff;
+			if (bInvertLevel)
+				level = 100;
+		}
 		else if (level > 99)
 		{
 			if (pDevice->devType==ZDTYPE_SWITCH_DIMMER)
 			{
-				level = 100;
+				(!bInvertLevel) ? level = 100 : level = 0;
 			}
 			gswitch.cmnd = gswitch_sOn;
 		}
 		else
 		{
+			if (bInvertLevel)
+				level = 100 - level; // invert reported level for BlindsPercentageInverted
+
 			gswitch.cmnd = gswitch_sSetLevel;
 		}
 
@@ -849,7 +864,6 @@ bool ZWaveBase::WriteToHardware(const char *pdata, const unsigned char length)
 	if ((packettype==pTypeGeneralSwitch)&&(subtype== sSwitchGeneralSwitch))
 	{
 		//light command
-
 		const _tGeneralSwitch *pSwitch= reinterpret_cast<const _tGeneralSwitch*>(pdata);
 
 		unsigned char ID1 = (unsigned char)((pSwitch->id & 0xFF000000) >> 24);
@@ -905,7 +919,7 @@ bool ZWaveBase::WriteToHardware(const char *pdata, const unsigned char length)
 			{
 				if ((cmnd== gswitch_sOff)||(cmnd== gswitch_sGroupOff))
 					svalue=0;
-				else 
+				else
 					svalue=255;
 				return SwitchLight(nodeID,instanceID,pDevice->commandClassID,svalue);
 			}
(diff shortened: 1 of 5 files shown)


hardware/ZWaveBase.cpp
```diff
@@ -338,19 +338,27 @@ void ZWaveBase::SendDevice2Domoticz(const _tZWaveDevice *pDevice)
 	ID3=(unsigned char)pDevice->nodeID&0xFF;
 	ID4=pDevice->instanceID;
 
-	char szID[10];
-	sprintf(szID,"%X%02X%02X%02X", ID1, ID2, ID3, ID4);
-
 	unsigned long lID = (ID1 << 24) + (ID2 << 16) + (ID3 << 8) + ID4;
+	char szID[10];
+	sprintf(szID, "%08x", (unsigned int)lID);
```
gizmocuz: Please undo this change...

hardware/ZWaveBase.cpp
```diff
@@ -338,19 +338,27 @@ void ZWaveBase::SendDevice2Domoticz(const _tZWaveDevice *pDevice)
...
 	{
 		BatLevel = pDevice->batValue;
 	}
 
 	if ((pDevice->devType==ZDTYPE_SWITCH_NORMAL)||(pDevice->devType==ZDTYPE_SWITCH_DIMMER)||(pDevice->devType== ZDTYPE_CENTRAL_SCENE))
 	{
+		bool bInvertLevel = false;
+		std::vector<std::vector<std::string> > result;
+		result = m_sql.safe_query(
+			"SELECT ID FROM DeviceStatus WHERE (DeviceID == '%q') AND (Type == %d) AND (SubType == %d) AND (SwitchType == %d)",
```
gizmocuz: Looking up values in the database is a very costly operation.
This patch should be handled by the web interface (webserver.cpp, gui html/js and/or mainworker.cpp, switchlight function)

gizmocuz: Probably mainly the switchlight function in the mainworker

gizmocuz (commented):
Thank you for the pull request.
Is it possible to not make any changes in the openzwave code.
I think the patch should be done in the web interface (webserver.cpp, gui html/js) and mainworker::SwitchLight

jvandenbroek:
Hmm my earlier comments above seem to be gone, I accidentally added them as reviews somehow and now they're canceled :/

I've moved the logic from ZWaveBase to SQLHelper, much better to do it this way indeed!

main/EventSystem.cpp
```diff
@@ -4691,8 +4691,8 @@ namespace http {
 							std::string actions = array[index].get("actions", "").asString();
 
 							if (
-								(actions.find("SendNotification") != std::string::npos) || 
-								(actions.find("SendEmail") != std::string::npos) || 
+								(actions.find("SendNotification") != std::string::npos) ||
```
gizmocuz: Could you undo the formatting in the EventSystem class ?

main/SQLHelper.cpp
```diff
@@ -3553,12 +3556,34 @@ uint64_t CSQLHelper::UpdateValueInt(const int HardwareID, const char* ID, const
...
+				ssLevel << level;
+				ssLevel >> sValue2;
+			}
 
 			result = safe_query(
 				"UPDATE DeviceStatus SET SignalLevel=%d, BatteryLevel=%d, nValue=%d, sValue='%q', LastUpdate='%04d-%02d-%02d %02d:%02d:%02d' "
 				"WHERE (ID = %" PRIu64 ")",
 				signallevel, batterylevel,
-				nValue, sValue,
+				nValue2, sValue2.c_str(),
```
gizmocuz, line 3586: This is quite unclear now... (at least... i find this unclear nvalue2/svalue2)
That is why i said it might be better to do this in the mainworker ?
No need to adjust any code in SQLHelper.cpp

gizmocuz (commented):
Thanks, but could you make the needed fixed before you call the SQLHelper.cpp functions ? Then all stays clear in this class

main/SQLHelper.cpp
jvandenbroek: The ZWaveBase pushes directly to the RX queue, so it's not using SwitchLightInt.. Which is the whole reason for me to use the ZWaveBase at first, otherwise all could be done in this function. I did comment before, but these seem to be lost because of my error :( but using the SwitchLighInt function in ZwaveBase would introduce even more overhead as you stated before with the SQL query, so this is IMO the best solution...

gizmocuz:
... Okey, and why would you need a blind percentage "inverted" ? why not wire the wires correctly ? I installed many nodes, also other people i know that did it themselves do not need to use an "inverted" blind type...

jvandenbroek:
Well I don't know, just trying to help solve user's problem and I guess it would be a nice feature to have... Otherwise we could just drop the whole inverted type :)

gizmocuz:
True, and that was why i asked this in the forum as well...

gizmocuz:
About dropping... that sounds like a good idea... because it should not be necessary
For a On/Off switch, that is debatable as you can use NO or NC relays.. then again, the hardware/driver beneath it should take care of it...
What was the forum link of this ? i remember this post.. but have not received an update

jvandenbroek:
http://www.domoticz.com/forum/viewtopic.php?f=6&t=9883 (also in OP above)
Well yes, would be a shame though if you drop it now, would have be nice if that happened before I put effort in it, but it's your child I guess..

gizmocuz:
Forum: "I agree that i could just rewire the wrong ones"
Sorry if you put effort into this, you could have always asked on slack.
But we should discuss this on the forum, but if you use a normal blind type, set the percentage to 80% , it should be 80% shut (some people think the opposite way), next when you press 'Open' it should open,.
If you use the OZW control panel and press open, it should open.
If it is also reversed there, people really should rewire ! I find it a very bad idea to support people who do not want to rewire and go for a lazy/quick sollution

gizmocuz:
I see also other issues when using the slider on the normal blind percentage... i will check 'm both but this will be in the web gui

jvandenbroek:
Not sure what you mean with normal blind percentage? Didn't change anything for that.. Or if you mean that OZW immediately reports back the old level, that's correct, somewhat 'weird' behavior, but I didn't bother too much (after set level is reached, it also set correctly in Domoticz).

~~Just re-looked at the code and I think it's feasible to implement the logic in mainworker decode_GeneralSwitch function instead, if you still want to keep the inverted option at all of course..~~ Hmm no there is a reason I didn't do that before, there is no way to get the switchtype without  a SQL query in this function... :(

gizmocuz:
Regarding the initial op question/command that his mind works in reverse...
In mainworker::switchlightint that should be changed, if inv then level = maxlevel-level
That for all switch types:
pTypeLighting2
pTypeGeneralSwitch

I am unable to test this as i am using RFY now... (unfortunately but who knows in the feature)
Maybe i can debug this on my laptop and someone's place in the feature

gizmocuz:
just made a small patch because when you had a normal "blinds percentage" and dragged the slider to 0 (totally left), the status text became "Off" instead of "Open" and 10 seconds later it was replaced with "Open"

jvandenbroek:
Yes, That's exactly what I did in SwitchLightInt for pTypeGeneralSwitch, not (yet) for pTypeLighting2, but could be done easily. Now I hard set it to 100, which is always the maxlevel for blinds percentage (inverted).

What happens now is as follows: SwitchLightInt reverses the values -> sends it to OZW hardware -> send the same still inverted values to the RX queue -> SQLHelper inverts it back.

The problem is with OZW sending the inverted value directly to the RX queue, which would be nice if it could be inverted back before sending to SQLhelper, but switchtype is only fetched from DB it seems..

gizmocuz:
The main thing is, if someones mind is the opposite of what it should be, even as stated in the manual, or also from the working (in this case) via the OZW control panel, then he/she should adjust, and in case of faulty wiring, change the wiring

gizmocuz:
true... thats whats happening when you need to invert stuff...
Do you own a zwave blind pecentage node that you can test against ?

gizmocuz:
Maybe better to discuss in slack

jvandenbroek:
Ah yes, that was always there, the Off -> Open, good to see it fixed ;)
But OK, then it's better to drop the whole inverted thing, giving it as an option implies it could be used this way...

gizmocuz:
Unless you are able to test it with a physical zwave blind node...
So i cant do this now, and still think it should not be needed at all in the first place

jvandenbroek:
I can test on a zwave percentage, not mine, but it works as expected (did this before even committing the PR, otherwise I wouldn't dare doing it ;)) Send you a message on Slack

PR #1749, opened 2017-08-07
Follow PEP-384 when finding and opening python3 library

mooninite:
Instead of maintaining a list of all Python versions the Python 3
standard defines that there MUST be a stable ABI and we can always
rely on a "python3.dll" or a "libpython3.so" existing.

This also fixes library loading on Linux distributions that use
SO versioned libraries.

### DIFF
--- a/hardware/plugins/DelayedLink.h
+++ b/hardware/plugins/DelayedLink.h
@@ -126,16 +126,10 @@ namespace Plugins {
 				if (!shared_lib_) shared_lib_ = LoadLibrary("python35_d.dll");
 				if (!shared_lib_) shared_lib_ = LoadLibrary("python34_d.dll");
 #	else
-				if (!shared_lib_) shared_lib_ = LoadLibrary("python37.dll");
-				if (!shared_lib_) shared_lib_ = LoadLibrary("python36.dll");
-				if (!shared_lib_) shared_lib_ = LoadLibrary("python35.dll");
-				if (!shared_lib_) shared_lib_ = LoadLibrary("python34.dll");
+				if (!shared_lib_) shared_lib_ = LoadLibrary("python3.dll");
 #	endif
 #else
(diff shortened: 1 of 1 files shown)


dnpwwo:
This sounds okay but you need to allow for the fact that not all Python 3 versions will work, 3.4 is the minimum.

Also, finding Python caused users a lot of grief early on which is why it is manual in the code.  If you make this change you will need to help/support users having issues if they occur.

mooninite:
A version check could be added. Much safer than checking a filename -- someone may copy libpython3.2 over to libpython3.4.

I did not remove any of the directory searching code. It will still look by LD_LIBRARY_PATH then /usr/lib then /usr/local/lib.

Does domoticz not link against python to keep the binary / installed footprint small? Otherwise it should link against it and the two above issues are negated.

dnpwwo:
Gizmocuz did not want to introduce a dependency on Python so I created the late linking layer which just disables Python plugins and scripts if it is not found so that Domoticz can still run. Its a bit painful but works fairly well.

If more people use the plugins or events then perhaps this could be revisited and the late linking removed in the future.

gizmocuz:
The problem will rise when old installations (systems without python) are going to update, the software won't launch anymore.
But this late binding is working great!

gizmocuz:
If i understand correctly, this patch causes all systems to malfunction when python is not installed ?

mooninite:
It does not. The software will behave the same as before if python is not detected.

I am not asking to remove dynamic loading. I would rather have this pull request accepted. Thanks.

gizmocuz:
Thanks! Ok, i will leave it to @dnpwwo to merge this

gizmocuz:
Not sure about the status, but shall this be merged or not ?

gizmocuz:
Closing issue as there is no response. Please re-open when needed

PR #1759, opened 2017-08-11
build: Detect and allow external tinyxpath library usage

mooninite:
Combines all previous work from PR 1731 into one clean commit.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -451,22 +451,6 @@ webserver/proxyclient.cpp
 json/json_reader.cpp
 json/json_value.cpp
 json/json_writer.cpp
-tinyxpath/action_store.cpp
-tinyxpath/htmlutil.cpp
-tinyxpath/lex_util.cpp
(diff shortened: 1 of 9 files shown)


gizmocuz:
Thanks!

gizmocuz:
Unfortunately this is not working with the current cmake version on raspberry and other systems (for example my pine64)

CMake Error at CMakeLists.txt:528 (target_sources):
  Unknown CMake command "target_sources".

Current cmake version 3.0.2

https://packages.debian.org/jessie/cmake

gizmocuz:
I reverted it partly, most is still there, but at least it now compiles on these arm systems

mooninite:
The target_sources command first appeared in CMake 3.1 3 years ago. The most important part of this pull request were the CMake changes.

Version 3.6.2 is available in backports to you. I'd highly recommend upgrading to a newer CMake. It will prevent future development more often the longer you stay on an ancient version.

PR #1760, opened 2017-08-12
Global TimeToString function

jvandenbroek:
While working on a function to get the time with ms resolution in a string for the EventSystem (dzVents), I figured out it would be better to make it a global function which you can supply different format options (time, date, datetime and datetime with ms), considering quite some functions construct a time string.

I saw that there were already a few functions added in Helper.cpp, but these are all C style and somewhat poorly formatted, so decided to use my own function and remove these.

When supplying 0 (time_t) as first argument, it will return the current time, otherwise it converts and returns the supplied time. First I had this written with a NULL pointer as valid argument, which would allow 0 as a time value, but quickly remembered how pointers are being disliked and also not needed for this purpose ;) Don't think that 0 is ever being used as time value, seems to me a good compromise.

The reason for using clock_gettime() when available, is that gettimeofday() is being depreciated, see also http://pubs.opengroup.org/onlinepubs/9699919799/functions/gettimeofday.html

If / when this gets merged, I'll do another PR for EventSystem, in which I have some more changes pending (so it's better not combine it with this PR).

### DIFF
--- a/main/Helper.cpp
+++ b/main/Helper.cpp
@@ -526,12 +526,12 @@ std::vector<std::string> ExecuteCommandAndReturn(const std::string &szCommand, i
 	return ret;
 }
 
-std::string TimeToString(time_t ltime, _eTimeFormat format)
+std::string TimeToString(time_t *ltime, _eTimeFormat format)
 {
 	struct tm timeinfo;
 	struct timeval tv;
 	std::stringstream sstr;
-	if (!ltime) // current time
+	if (ltime == NULL) // current time
 	{
 #ifdef CLOCK_REALTIME
 		struct timespec ts;
@@ -551,7 +551,7 @@ std::string TimeToString(time_t ltime, _eTimeFormat format)
 #endif
 	}
 	else
-		localtime_r(&ltime, &timeinfo);
+		localtime_r(&(*ltime), &timeinfo);
 
 	if (format > TF_Time)
 	{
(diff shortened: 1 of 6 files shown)


main/Helper.cpp
```diff
@@ -526,50 +526,54 @@ std::vector<std::string> ExecuteCommandAndReturn(const std::string &szCommand, i
 	return ret;
 }
 
-//convert date string 10/12/2014 10:45:58 en  struct tm
-void DateAsciiTotmTime (std::string &sTime , struct tm &tmTime  )
+std::string TimeToString(time_t ltime, _eTimeFormat format)
```
gizmocuz: Should ltime not be a pointer ? (time expects a pointer, also localtime)
Also makes it a bit more consistant
You could also have choosen for using the strftime format...

jvandenbroek: No, we're not supplying a time struct, just time_t (int), but I guess the naming should be different to avoid confusing... I started this with a struct, which is why it's still named ltime. Anyway, there is no need for a pointer in this case. Of course I could make it a pointer, for consistency purposes? 

You can now just eg. use the time() (or mytime() for that matter) function to get the time as an integer and do your calculations on that and directly get a time as string. Otherwise you need to construct and supply a the struct every single time (it's part of the idea for this function, do the constructing and converting to string transparently).

jvandenbroek: Well thinking about it I guess for consistency purposes you're right and we should use a pointer.. Will change that!

gizmocuz:
Thanks!

PR #1761, opened 2017-08-12
Add support of command line parameters to Rtl433 hardware

rpeyron:
Hi,
Thanks a lot for domoticz!  I use rtl_433 and need some parameters (currently to choose USB device and frequencies that should be scanned), so I added an extra parameters setting for the Rtl433 hardware. This is very basic, but it works.
Thanks again


hardware/Rtl433.cpp
```diff
@@ -17,9 +17,18 @@
 
 #include "Rtl433.h"
 
-CRtl433::CRtl433(const int ID) :
-	m_stoprequested(false)
+void removeCharsFromString(std::string &str, const char* charsToRemove ) {
```
gizmocuz, line 20: can't you use our standard stdreplace function for this ?

rpeyron: I didn't see it. But if I am not mistaken, stdreplace does only replace one string at once, so it would need 11 lines of strreplace to do the same thing (one for each char to remove). Tell me what you prefer.

gizmocuz: Your correct, sorry missed that part ;)

gizmocuz:
Thanks!

PR #1766, opened 2017-08-16
EventSystem added time vars for dzVents

jvandenbroek:
- Added global time vars for dzVents: domoticz start time, current time with ms resolution (not available in Lua) and system uptime.
- Put the static map in the EventSystem class and use string instead of char.
- Implemented global TimeToString function to significantly reduce needed lines of code.


main/EventSystem.cpp
```diff
@@ -44,25 +45,10 @@ static std::string m_printprefix;
...
-	const char* szOriginal;
-	const char* szNew;
-	_eJsonType eType;
-};
-
 // This table specifies which JSON fields are passed to the LUA scripts.
 // If new return fields are added in CWebServer::GetJSonDevices, they should
 // be added to this table.
-static const _tJsonMap JsonMap[] =
+const CEventSystem::_tJsonMap CEventSystem::JsonMap[] =
```
gizmocuz: Could you undo the structure change and change it back to NULL and const char* , that was way cleaner, also in the code below

jvandenbroek:
Undid restructure JsonMap, removed unneeded / redundant code from UpdateJsonMap (operator = does the same as insert, will add new or replace existing values)

main/EventSystem.cpp
```diff
@@ -45,10 +45,25 @@ static std::string m_printprefix;
 extern PyObject * PDevice_new(PyTypeObject *type, PyObject *args, PyObject *kwds);
 #endif
 
+typedef enum
```
gizmocuz: Is it possible you can follow our ENUM declararions ? If you search in our project for 'typedef enum' you will see what i mean ... not like 'tString' ?
Like JTYPE_STRING, JTYPE_FLOAT
Missed this one before...

gizmocuz:
Thanks!

PR #1767, opened 2017-08-16
repare I2C PCF8574 hardware modelu

OndrejLycka:
I repaired problem who is described in forum: https://www.domoticz.com/forum/viewtopic.php?t=15909
In newest version was  deleted important line with: i2c_addr = Port; this is reason for this is not function.
Somebody in forum complain that very very bad comments in the code, totally not understandable.
Now I simplified code so now is very easy and clear understandable.

### DIFF
--- a/hardware/I2C.cpp
+++ b/hardware/I2C.cpp
@@ -167,35 +167,20 @@ bool I2C::WriteToHardware(const char *pdata, const unsigned char length)
 		_log.Log(LOG_NORM,"GPIO: id2 %d", pCmd->LIGHTING2.id2);
 		_log.Log(LOG_NORM,"GPIO: id3 %d", pCmd->LIGHTING2.id3);
 		_log.Log(LOG_NORM,"GPIO: id4 %d", pCmd->LIGHTING2.id4);
-		_log.Log(LOG_NORM,"GPIO: unitcode %d", pCmd->LIGHTING2.unitcode); // in DB columb "Unit" used for identify number switch on board
-		_log.Log(LOG_NORM,"GPIO: cmnd %d", pCmd->LIGHTING2.cmnd);
-		_log.Log(LOG_NORM,"GPIO: level %d", pCmd->LIGHTING2.level);
-		*/
 		unsigned char id1=pCmd->LIGHTING2.id1;
 		unsigned char id2=pCmd->LIGHTING2.id2;
 		unsigned char id3=pCmd->LIGHTING2.id3;
 		unsigned char id4=pCmd->LIGHTING2.id4;
 		int ID=(id1<<24)|(id2<<16)|(id3<<8)|id4;
-		unsigned char Unit = pCmd->LIGHTING2.unitcode;
+		//_log.Log(LOG_NORM,"GPIO: ID %d", ID);
+		_log.Log(LOG_NORM,"GPIO: unitcode %d", pCmd->LIGHTING2.unitcode); // in DB column "Unit" used for identify pin number
+		_log.Log(LOG_NORM,"GPIO: cmnd %d", pCmd->LIGHTING2.cmnd);
+		_log.Log(LOG_NORM,"GPIO: level %d", pCmd->LIGHTING2.level);
+		*/
+		unsigned char pin_number = pCmd->LIGHTING2.unitcode; // in DB column "Unit" used for identify pin number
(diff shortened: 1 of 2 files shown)


gizmocuz:
about the Device ID generation:
- This should be logical, for example if you have a ID and a Unit it could be
(ID=4, unit=5)

%02X%02X (two hex numbers) with ID and Unit, so it will be 0405

In your case you have a I2C address (this is 1 hex value right?) and a PIN number
For example I2C address = 0x62 and the pin is 15 (that is in hex 0x0f)
ID will be %02X%02X = 620F

the id can be up 2 for hex digits

But you should actually not poke/peek/read/write in the database directly but use the already build in commands 'SendSwitch' for example

OndrejLycka:
This should be logical, for example if you have a ID and a Unit it could be (ID=4, unit=5).
Problem is when I create device by SendSwitch(DeviceID, Unit, 255, value, 0, "")  I have not his unique ID this is will be generated by SendSwitch().
Yes I can use I2C address and ping number for create DeviceID (this I do). But this does not exclude the possibility of someone else use same numbers with another significance (e.g. instead number of pin number tcp port, etc..). Would be good add into DeviceID hradwareType number but some standard way.

gizmocuz:
I see, but you do have the ID and the unit code if you use those commands.
ID: 4 unit 5, when using sendswitch it will be 
ID: 00000004 and unit = 5

	//make device ID
	unsigned char ID1 = (unsigned char)((NodeID & 0xFF000000) >> 24);
	unsigned char ID2 = (unsigned char)((NodeID & 0xFF0000) >> 16);
	unsigned char ID3 = (unsigned char)((NodeID & 0xFF00) >> 8);
	unsigned char ID4 = (unsigned char)NodeID & 0xFF;
sprintf(szIdx, "%X%02X%02X%02X", ID1, ID2, ID3, ID4);

See DomoticzHardware.cpp

Yes, we use HEX, so ID 10 will become 0x0A (0A)

gizmocuz:
That is also why for zwave we show the nodes in HEX

OndrejLycka:
For me not so much importance how this number show, but how this number (identificator) composed so that be realy unique for all devices in device list.

Or maybe I do not understand it. This number is not whole identificator and as full identificator of device must be used combination Hardware, ID and Unit .

OndrejLycka:
Now I look this is not chat, but commits. I add commit with cosmetic changes.

hardware/I2C.cpp
```diff
@@ -167,35 +167,20 @@ bool I2C::WriteToHardware(const char *pdata, const unsigned char length)
...
-		_log.Log(LOG_NORM,"GPIO: ID %d", ID);
-		_log.Log(LOG_NORM,"GPIO: Unit %d", Unit);
-		_log.Log(LOG_NORM,"GPIO: new value %d", value);
-		*/
-		// check my serverd i2c_addr (form Unit I can get i2c_addres of switch)
-		unsigned char i2c_address=PCF8574_get_i2c_addr_from_Unit(Unit);
-		/*
-		_log.Log(LOG_NORM,"GPIO: my served i2c address %d", i2c_addr);
-		_log.Log(LOG_NORM,"GPIO: i2c address from Unit %d", i2c_address);
+		//_log.Log(LOG_NORM,"GPIO: ID %d", ID);
```
gizmocuz: Could you remove this debug log code, or surround with '#IFDEF _DEBUG'

OndrejLycka: OK I do. 

Last compile for apple had problem with connect: The remote end hung up unexpectedly
maybe some problem on git?

Is it
DeviceID = (HTYPE_RaspberryPCF8574 << 24) + (0 << 16) + (i2c_addr << 8) + pin_number
correct?

gizmocuz:
I am working on some apple/travis scripts at the moment...

Is your ID 24 bits? then it could be:

(HTYPE_RaspberryPCF8574 << 16) +  (i2c_addr << 8) + pin_number

If pin_number can not be larger then 255

you also do actually NOT need HTYPE_RaspberryPCF8574 , as the DeviceID belongs to the unique hardware

So, it is 16 bit:

(i2c_addr << 8) + pin_number

OndrejLycka:
Perhaps last commit for I2C PCF8574.

gizmocuz:
Thanks!!

PR #1771, opened 2017-08-18
XiaomiGateway - Using extra unit codes for existing and new wall switches

corbinmunce:
As part of this change, there are new unitcodes for wall switch devices.  Existing wall switches will need to be deleted and re-added in Domoticz.  For dual button switches, press each button separately for detection.


gizmocuz:
Thanks!

gizmocuz:
An auto transition from the old to the new unit id would have great (see sqlhelper.cpp database version)

theo-69:
Just updated still seeing: 2017-08-19 10:27:53.567 XiaomiGateway: unhandled model: ctrl_ln1

corbinmunce:
I've had a look, but don't think it will be possible with a database query and update.  Problem is a few other devices have the same unit, type, subtype, and switchtype.

Ideally I would have used distinct unit codes from the start but to change to that now means more deleting and re-adding devices of devices for users, or additional database checks on each device update in XiaomiGateway.cpp.

Your thoughts?

gizmocuz:
Now you are using better unit codes... hope this will make it easier in the feature ;)

PR #1777, opened 2017-08-21
Elemental theme: Fix selector switches to fit to a smaller area

odelma:
With the latest changes, selector switch levels - when not fitting 100% to the area reserved for them - would skip to next line (inline-block style) and run over to the next block (white-space: nowrap style). The font size was also bit large and the padding as well. With these changes everything should once again fit more or less to the area reserved for selector levels from the actual block.

### DIFF
--- a/www/styles/elemental/custom.css
+++ b/www/styles/elemental/custom.css
@@ -173,6 +173,20 @@ body table#itemtablenostatus tbody tr{
 	margin-top: 10px;
 }
 
+.btn-group {
+       display: block;
+       white-space: normal;
+}
+
+.btn-group > .btn-small {
+    font-size: 10px;
+}
+
+.btn-small {
+    padding: 2px 3px;
+    font-size: 10px;
+}
+
 #timesun {
 	height: 2.60em;
 	position: fixed;


gizmocuz:
Could you give a screenshot of how it looks in the default theme ?
First the default theme has to be solved.
There will be more changes to the default theme, and when this all works correct, 'other' themes could be addressed

odelma:
Sure but as there seems to be also some other problems with the default theme I'm not sure whether it would give the correct impression of the issue at hand with the Elemental theme? Maybe it is best that I wait for the default theme to be completed and then see whether the Elemental theme needs some fixes as well? Until that I can add the needed style to custom.css manually.

gizmocuz:
What problems ? Any way, use the default theme for now...

odelma:
Seems that the current beta shows the default version quite ok. The only issue is the same as I wanted to fix with the Elemental theme that on the Dashboard the selector switch items (following the "inline-block" and "nowrap" styles) are out of the area where they are supposed to be on...
screen shot 2017-08-23 at 0 00 57

gizmocuz:
That indeed does not look very good

PR #1779, opened 2017-08-23
Python Events - Removed misleading log entry on startup

moroen:
Minor change, removed a unnecessary and misleading log message when the python event system starts.


gizmocuz:
Thanks!

PR #1781, opened 2017-08-25
EventSystem/Json implemented trigger logic, changed scenegroup/uservar

jvandenbroek:
Changes (mainly in preparation for the next dzVents version):
- Lua: TRIGGER argument added for UpdateDevice and Variable functions (defaults to no trigger)
- Lua: NOTRIGGER argument added for general device actions (defaults to trigger)
- Json: triggerevent argument added (triggerevent=1 trigger, triggerevent=0 no trigger)
- Lua/Json: Scenes/groups and uservars now also work with event triggers, members of scenes/groups will follow this (defaults to trigger for scenes/groups, no trigger for uservars)
- Lua: Scenes/groups and uservars lastupdate logic now works like devices (update lastupdate after running script)
- Small fix preventing dzVents to evaluate when no dzVents scripts are enabled

TRIGGER (triggerevent=1) triggers the event system, NOTRIGGER (triggerevent=0) makes sure it doesn't. 

We made sure the trigger logic defaults to dz's current behaviour, preventing possible breaking of existing installations/scripts. For that reason we use both TRIGGER and NOTRIGGER as arguments for different functions, instead of just one. Same goes for the Json calls, udevice calls defaults to no trigger, others to trigger. Can be toggled in all cases.


gizmocuz:
Crazy patch (bEventTrigger).... should not be needed
Could you make a different patch for other dzVents related things ?

PR #1785, opened 2017-08-28
Update dark-th3me events editor cursor position

gerard33:
Add a fix for the Event editor wrong cursor position


gizmocuz:
Thanks!, But there are quite some more changes that need to be implemented in the default theme.
Because of this , it's best to wait...

gizmocuz:
Thanks!

PR #1786, opened 2017-08-29
Try to make Travis use C++11 with Clang on OSX

xbeaudouin:
Try to make modification on OSX Travis and use C++11 with clang.


gizmocuz:
Great! Question, it seems there is a debug line active ? (which python3)

xbeaudouin:
You are right, I just removed it. 
Regards

gizmocuz:
Thanks!!

PR #1787, opened 2017-08-29
Neo Coolcam 2 channel wall switch config file

LouiS22:
Submitted to OZW as well but could be added here as well


gizmocuz:
Thank you for this, but it will be merged once its in ozw.

PR #1788, opened 2017-08-30
EventSystem extended UpdateDevice function

jvandenbroek:
Needed a simple way to toggle protection for an item on/off (after I accidentally switched something remotely without noticing :-P), added a field for this to the UpdateDevice function. Argument value 1 will enable protection, 0 disables it.

Also changed the logic, now you can supply empty values to prevent changing current values (eg. 1||123 sets svalue to 123 for ID 1, but doesn't touch nvalue. Setting only a nvalue is as simple as just supplying 1|1, it's no longer needed to set all arguments (but of course still supported). Using -1 value for an invalid / empty nvalue, prevents updating it's value in memory (m_devicestates).

Example of usage to switch protection: 
commandArray['UpdateDevice'] = '1|||1' -> enables protection for ID 1. 
commandArray['UpdateDevice'] = '1|||0' -> disables protection for ID 1.


gizmocuz:
Thanks!

PR #1793, opened 2017-09-03
Fix compile with support for openzwave on macOS

moroen:
On macOS, since the header files are not in the same directory as ozwcp.cpp, clang seems to get confused when compiling with both openzwave and python, and imports node.h from the python-libs and not hardware/openzwave/Node.h as it should. This should fix this, and still compile under linux and win.

### DIFF
--- a/hardware/openzwave/control_panel/ozwcp.cpp
+++ b/hardware/openzwave/control_panel/ozwcp.cpp
@@ -41,7 +41,7 @@
 #include <string.h>
 #include "Options.h"
 #include "Manager.h"
-#include "Node.h"
+#include "../Node.h"
 #include "Group.h"
 #include "Notification.h"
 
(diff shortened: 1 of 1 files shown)


gizmocuz:
I see, but then it should also have an issue with Group and Manager ? They all reside in a higher directory ? could you test against all needed header files ?

moroen:
I'm no expert on clang and have pretty much no idea on how it searches the include-paths for includes, but I believe the reason Node.h fails, while Group.h and Manager.h doesn't, is the fact that there's another node.h in the include-path, the one from python. Group.h and Manger.h doesn't, and somehow clang uses files from the parent-directory... 

Anyway, I have updated ozwcp.cpp  and tested with all needed headers, and it still compiles ok on macOS, linux and win.

gizmocuz:
Thanks!

PR #1797, opened 2017-09-07
Evohome TCP interface

codeaholics:
This PR allows access to the Evohome hardware over a TCP link as well as the original serial port. This is the same as many other hardware devices and relies on the use of something like `socat` or `ser2net` on the remote end to expose the serial port over TCP.


Evsdd:
Code looks good.  I've tested both the Evohome HGI80 TCP and direct serial connection and all works great.  Thanks.

xbeaudouin:
So with this code I can use a RPI to access to HGI80 with socat (because the TI Usb2serial chipset is not supported on my FreeBSD server) ?

codeaholics:
If I've understood you correctly, yes. You can put your HGI80 on a RPi running `socat`, and then run Domoticz on another machine.

codeaholics:
I've re-merged the development branch into my PR branch. Could somebody take a look at this please?

codeaholics:
I've just spotted a couple of problems. I'll get them fixed and come back here when they're done. Sorry for the noise.

codeaholics:
OK. Domoticz wasn't stopping correctly with this new hardware device. I have pushed the fix and I think this PR is ready to go again.

gizmocuz:
Great work! Thanks!

PR #1798, opened 2017-09-08
Adding Feynman theme custom.css

flatsiedatsie:
Using the new CSS still works very well, even though the new default theme and the other changes (settings re-ordering, item merging, dataviz on dashboard, etc) haven't been incorporated. So I'm adding the Feynman theme already.

I will upload the other files too.


gizmocuz:
Thanks, but... first want to make the default theme ok
But can you remove all images, they are already in the main theme...

flatsiedatsie:
Unfortunately the images are not all already in the main theme. While I have updated them all, only a few hand-picked ones have been incorporated.

gizmocuz:
i would then only add the changed/new images as the rest will be loaded from the main image folder

flatsiedatsie:
I'd rather just keep the full iconset intact.
- Easier to update the theme: no need to also check which icons are already in Domoticz.
- Futureproof & seperation of concerns: if the themes that come with Domoticz change or even if icons would be removed from Domoticz, then this theme would still function as intended, as would Domoticz.
- Technically it shouldn't matter, right?

gizmocuz:
Then we keep themes stand alone and not in the default setup.
This because it is a waist of disk space/bandwidth otherwise.
Also now you have no idea what icon you have overwritten

flatsiedatsie:
Alright, I'll try to remove the images that are already in.

flatsiedatsie:
Creating a new branch because deleting files through the Github web interface can only be done one at a time.

PR #1802, opened 2017-09-10
EventSystem disable on initialization

jvandenbroek:
Prevent the Event System being enabled on initialization while not yet started, should fix the rare (though multiple times reported) issue of printing dzVents instead of Lua on script execution.


gizmocuz:
Thanks!

PR #1805, opened 2017-09-13
EventSystem implemented trigger functions for dzVents

jvandenbroek:
Implemented SetEventTrigger and GetEventTrigger functions to set/get whether an event should be triggered or not, preventing the need to set the infamous bEvenTrigger bool throughout the whole project (like the previous rejected patch). This patch doesn't include any json related functionality, only focuses on Lua/dzVents.

Changes:

- TRIGGER argument added to UpdateDevice and 'Variable:' functions (defaults to no trigger)
- NOTRIGGER argument added for general device actions (defaults to trigger)
- Scenes/groups and uservars now also work with event triggers, members of scenes/groups will follow this (defaults to trigger for scenes/groups, no trigger for uservars)
- Scenes/groups and uservars lastupdate logic now works like devices (update lastupdate after running script)
- Small fix preventing dzVents to evaluate when no dzVents scripts are enabled

So TRIGGER triggers the event system, NOTRIGGER makes sure it doesn't.

We made sure the trigger logic defaults to Domoticz' current behaviour, preventing possible breaking of existing installations/scripts. For that reason we use both TRIGGER and NOTRIGGER as arguments for different functions, instead of just one.

### DIFF
--- a/main/EventSystem.cpp
+++ b/main/EventSystem.cpp
@@ -3953,18 +3953,16 @@ void CEventSystem::UpdateDevice(const std::string &DevParams, const bool bEventT
 		protect = strarray[3];
 	//Get device parameters
 	std::vector<std::vector<std::string> > result;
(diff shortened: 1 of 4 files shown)


gizmocuz:
Thanks!

PR #1807, opened 2017-09-13
Added support for FreeBSD for Python Plugins

xbeaudouin:
On FreeBSD, the libpython libs finish with a 'm' (don't ask me why, but this is the fact).
This trivial patch allow the delayedlinker to find the right lib :)


gizmocuz:
Is this perhaps for all libraries ? if yes, maybe we should tweak the FindLibrary function ?

xbeaudouin:
@gizmocuz on my freebsd only python3.x has this ..."extension". Maybe this is specific python3.x branch on freebsd?

xbeaudouin:
Forgot this PR. The issue is somewhere else. Digging.

PR #1811, opened 2017-09-15
Add support for Netamo modules battery level

trandbert37:
I retrieve battery level for Netatmo modules by use the percent value of the API instead of battery_vp that is too fuzzy.


gizmocuz:
Thanks!

PR #1812, opened 2017-09-16
Added possibility to select phones for GCM in LUA

Olifant1990:
You can set extraData param in lua to choose which phones would receive
a GCM push notification.
You have to give the idx value of the phone.
Now visible in the Mobile view under Setup -> More Options -> Mobile
Devices. Multiple devices could be given, seperated by a semicolon.

Example:
commandArray['SendNotification'] = "Subject#Message###1;2#gcm"

notifications\GCM
www\app\mobileNot
www\views\mobile_n


gizmocuz:
Thanks!

PR #1815, opened 2017-09-18
MyHome Thermoregulation

olix1981:
Add setpoint service on MyHome OpenWebNetTCP. Default name "Temperature [ZONE] Setpoint"

Enum moved and defined as type once


gizmocuz:
Thanks!

PR #1819, opened 2017-09-21
EventSystem add BackupDB function Lua/dzVents

jvandenbroek:
Allow backups of the DB from Lua/dzVents. Also fix small memory leak in case the specified path can't be written (or some other error occurs).

Example of usage: commandArray['BackupDB'] = '/mnt/backups/domoticz/backup-' .. currentDate .. '.db'


gizmocuz:
You can create a backup from the settings page and from the json command.
And there is a kind of automatic backup thing (which should be worked on a bit)
No need for an additional backup command....
Thanks!

PR #1821, opened 2017-09-23
Fix for correct graph link for custom devices on the Floorplan

djoele:
This fix makes sure that custom devices that are shown on the Floorplan have a correct link to their graph.
Before this change clicking on a custom device did not show its graph.


gizmocuz:
Thanks!

PR #1825, opened 2017-09-24
dzVents 2.3.0

dannybloe:
[2.3.0]

 - Added `active` attribute to devices (more logical naming than the attribute 'bState'). `myDevice.active` is true or false depending on a set of known state values (like On, Off, Open, Closed etc). Use like `if mySwitch.active then .. end`
 - Added `domoticz.urlEncode` method on the `domoticz` object so you can prepare a string before using it with `openURL()`.
 - Updating devices, user variables, scenes and groups now always trigger the event system for follow-up events.
 - Added support for groups and scenes change-event scripts. Use `on = { scenes = { 'myScene1', 'myScene2' }, groups = {'myGroup1'} }`
 - Added adapter for the new Temperature+Barometer device.
 - Added `domoticz.startTime` giving you the time at which the Domoticz service was started. Returns a Time object.
 - Added `domoticz.systemUptime` (in seconds) indicating the number of seconds the machine is running. Returns a Time object.
 - Added more options to the various commands/methods: e.g. `myDevice.switchOff().silent()`, `.forSec()`, `.forHour()`, `.afterHour()`, `.repeatAfterSec()`, `.repeatAfterMin()`, `.repeatAfterHour()`, `.withinHour()` and `.checkFirst()`. This now works not only for devices but also scenes, groups and user variables. See documentation about command options.
 - Added `.silent()` option to commands like `switchOn().afterSec(4).silent()` causing no follow-up events to be triggered. This also works when updating non-switch-like devices, scenes, groups and user variables. If you do not call `silent()``, a follow-up event is *always* triggered (for devices, scenes, groups and user variables).
 - Added time rule `between xx and yy`. You can now do things like: `between 16:34 and sunset` or `between 10 minutes before sunset and sunrise`. See the doc.
 - Added support for milliseconds in `Time` object. This also give a ms part in the time-stamp for historical persistent data. You can now do `msAgo()` on time stamp when you inspect a data point from an history variable.
 - Added `daysAgo()` to `Time` object.
 - Added `compare(time)` method to `Time` object to calculate the difference between them. Returns a table. See documentation.
 - Added `domoticz.round()` method to `domoticz` object.
 - `active` section is now optional in your dzVents script. If you don't specify an `active = true/false` then true is assumed (script is active). Handy for when you use Domoticz' GUI script editor as it has its own way of activating and deactivating scripts.
 - Added `humidityStatusValue` for humidity devices. This value matches with the values used for setting the humidity status.
 - `Time` object will initialize to current time if nothing is passed: `local current = Time()`.
 - Added the lua Lodash library (http://axmat.github.io/lodash.lua, MIT license).
 - Fixed documentation about levelNames for selector switches and added the missing levelName.
 - Moved dzVents runtime code away from the `/path/to/domoticz/scripts/dzVents` folder as this scripts folder contains user stuff.
 - Added more trigger examples in the documentation.
 - You can now put your own non-dzVents modules in your scripts folder or write them with the Domoticz GUI editor. dzVents is no longer bothered by it. You can require your modules in Lua's standard way.
 - Added `/path/to/domoticz/scripts/dzVents/scripts/modules` and `/path/to/domoticz/scripts/dzVents/modules` to the Lua package path for custom modules. You can now place your modules in there and require them from anywhere in your scripts.
 - Added dzVents-specific boilerplate/templates for internal web editor.
 - Fixed the confusing setting for enabling/disabling dzVents event system in Domoticz settings.
 - Fixed a problem where if you have two scripts for a device and one script uses the name and the other uses the id as trigger, the id-based script wasn't executed.


main/EventSystem.cpp
```diff
@@ -370,6 +384,51 @@ std::string CEventSystem::LowerCase(std::string sResult)
 	return sResult;
 }
 
+// Temporarily function used for removing leftover dzVents files
+int CEventSystem::RemoveDir(const std::string &dirnames, std::string &errorPath)
```
gizmocuz: Could you move this function to Helper.cpp, we might need this later for other purposes and it's better to have this global

main/EventSystem.h
```diff
@@ -18,6 +18,8 @@ extern "C" {
 #include "LuaCommon.h"
 #include "concurrent_queue.h"
 
+#define DZVENTS_VERSION "2.3.0"
```
gizmocuz: Please do not make changes to a header file if you need a version number.
Best to do this in the CPP.
Better to have a dzversion.h, and include this in the class it's cpp file (if there is a dz class, in this class, else in EventSystem.cpp)

gizmocuz: (otherwise at a version change we need to recompile a lot more then one class)

main/EventSystem.cpp
```diff
@@ -184,12 +184,26 @@ void CEventSystem::SetEnabled(const bool bEnabled)
...
 #endif
+	const std::string // remove obsolete dirs
+	dzv_rm_Dir1 = dzv_scripts + "runtime",
+	dzv_rm_Dir2 = dzv_scripts + "documentation";
+
+	if ((file_exist(dzv_rm_Dir1.c_str()) || file_exist(dzv_rm_Dir2.c_str())) &&
+		!szUserDataFolder.empty())
+	{
+		std::string errorPath;
+		if (int returncode = RemoveDir(dzv_rm_Dir1 + "|" + dzv_rm_Dir2, errorPath))
```
gizmocuz: Will this be executed every time 'LoadEvents' is executed ?
Normally we place this kind of 'patches' inside SQLHelper.cpp and increase the database version number

main/EventSystem.h
jvandenbroek: Where should we put this file? Like appversion.h in the root?

main/EventSystem.cpp
jvandenbroek: Better indeed, I thought it was better to keep all in eventsystem, but with moving the removedir function to helper, we should do it this way. Will change.

main/EventSystem.h
jvandenbroek: It's being used by webserver, so including it in eventsystem.cpp doesn't work (there is no dz class). Including in eventsystem.h works, is that ok? For now I've put the file dzversion.h in the root next to appversion.h, let me know if you want it elsewhere (in main?).

gizmocuz: Maybe write a wrapper class for dzVents and in this class make a funciton like

char *GetVersion()

that returns the version that is declared in the .cpp file

Maybe also move some stuff from the base EventSystem to this class ?

cyclops1982:
Scene event support. Just what i was looking for!

Tested the scenes support and it seems to work!

dannybloe:
Everything still works like a charm.. For me it is good to go.

gizmocuz:
Thanks for the great work again !!!

PR #1829, opened 2017-09-26
Add support for send rssi value

trandbert37:
Add ability to send rssi value when we use for exemple "SendTempSensor" or "SendTempHumSensor".
I so add a parameter with a default value of 12 (to avoid problems with existants usages).
I'm not very sure of the quality of this work :)
I do this because I want to do an other PR to add Netatmo RSSI values support if this PR is validate (it already work on my device).

Thank !


gizmocuz:
Thanks!

PR #1843, opened 2017-10-03
Properly handle humidities with a decimal place returned by RTL_433, along with accessing all RTL_433 sensors

nahall:
Some RTL_433 devices return a humidity with a decimal place, which was causing Domoticz to throw a casting error. This change rounds the humidity and casts it properly.

There are also some RTL_433 temperature sensors that ignored with the default rtl_433 command line arguments (such as the Acurite 592TXR weather sensors). This adds the "-G" switch so rtl_433 doesn't ignore those devices.


gizmocuz:
is it possible to remove the boost include/dependency ?
you could use

`#define round(a) ( int ) ( a + .5 )`

hardware/Rtl433.cpp
```diff
@@ -232,11 +233,12 @@ void CRtl433::Do_Work()
 				try {
 					if (!data["humidity"].empty())
 					{
-						humidity = boost::lexical_cast<int>(data["humidity"]);
+						humidity = static_cast<int>(boost::math::round(boost::lexical_cast<float>(data["humidity"])));
```
gizmocuz: Please remove the boost dependency... have a small search for the 'round' function used by various other classes (#define)

nahall: Sure, I'll take care of that, thanks.

nahall:
OK I modified it to eliminate the boost dependency.

gizmocuz:
Thanks! Sorry about this, i thought it was the only boost cast in this class, but apparently there are many more.
I will merge the pull request and do some modifications in a few minutes and push the changes back.

humidity = static_cast<int>(round(boost::lexical_cast<float>(data["humidity"])));

could just be

humidity =round(data["humidity"].c_str());

gizmocuz:
Oeps.. you made the changes against the wrong branch,
Could you do them against the 'development' branch ?

nahall:
OK I opened a new pull request #1879 which performs the fix with the development branch instead.

PR #1845, opened 2017-10-05
Reduced the size of the www/images/startup by 4MB having compressed pngs.

omnima:
This change reduces the size of startup images to make loading the UI faster and take less space on the disk.

Signed-off-by: omnima limited <info@omnima.co.uk>


gizmocuz:
Thanks!!

PR #1859, opened 2017-10-10
EventSystem update entries on rename

jvandenbroek:
Make sure entries in EventSystem memory tables are updated/deleted when devices and scenegroup items are updated/deleted.


gizmocuz:
Thanks!

PR #1860, opened 2017-10-10
Add support to toggle groups On/Off

emontnemery:
Extend command "switchscene" with support for switchcmd "Toggle".


gizmocuz:
Thanks! Question

bool SwitchScene(const std::string &idx, const std::string &switchcmd);

calls the other SwitchScene function that does not have the 'const' switchcmd string.

Should the above function also be modified (removing the const) ?

emontnemery:
No need to modify this function:
`bool SwitchScene(const std::string &idx, const std::string &switchcmd);`

The other function, `bool MainWorker::SwitchScene(const uint64_t idx, std::string switchcmd)`,  is
modified such that a non-const local copy of switchcmd is passed instead of a reference.

gizmocuz:
Thanks!

PR #1874, opened 2017-10-15
Some code cleanups

correa:
Remove unnecessary string conversions and simplify some of the code.

There is also some C++11 opportunities here, like using std::atoi so we don't need c_str() conversions anymore.


gizmocuz:
Thank you for the cleanup, good find!
Could you undo the rtl433 changed as i just merged a pull request causing this pull request to be able to merge.
We have to be careful with c++11, i would love to go this way, but we have to make sure it can be compiled on a raspberry pi for instance.

gizmocuz:
Thanks!

PR #1875, opened 2017-10-15
Reduce symbols in headers, removes unused includes. 

correa:
Some help from Include What You Use clang tool


gizmocuz:
Great work ! Thanks!!

PR #1876, opened 2017-10-16
Fix Telegram notifcation setup layout

chatainsim:
Fix Telegram notifcation setup layout
`<label for="TelegramEnabled"/>` was set at the wrong place.
And there was an extra `</div>` I forgot to remove.


gizmocuz:
Thanks!

PR #1877, opened 2017-10-16
Evohome web updates

gordonb3:
1) fix possible crash in retrieval of DHW schedule
2) trottling of connection retries while offline or during maintenance did not work
3) fix overwriting of zone offline state with system state


gizmocuz:
Thanks!!

PR #1878, opened 2017-10-16
Allow gas meter M-Bus channel override through user variable

gordonb3:
As it turns out some users have badly configured meters that require them to be able to override the automatic channel select for the gas meter on the smart meter M-Bus.

This version checks for the existence of a user variable named `P1GasMeterChannel` with allowed values 1..4 (it will drop back to auto-select with any other value) to skip the verification process and set the gas meter channel directly.


hardware/P1MeterBase.cpp
```diff
@@ -133,6 +133,15 @@ void P1MeterBase::Init()
 	m_gastimestamp="";
 	m_gasclockskew=0;
 	m_gasoktime=0;
+
+	std::vector<std::vector<std::string> > result;
+	result = m_sql.safe_query("SELECT Value FROM UserVariables WHERE (Name == 'P1GasMeterChannel')");
+	if ((result.size() > 0) && (result[0][0].size()==1) && (result[0][0]>"0") && (result[0][0]<"5"))
```
gizmocuz: Just for my reading experience... could you simplify this line

if (!result.empty())
{
 int gasmbuschannel=atoi(result[0][0].c_str());
 if ((gasmbuschannel>0)&&(gasmbuschannel<5))
 {
    m_gasmbuschannel=gasmbuschannel;
    ....etc
 }
}

gordonb3:
Good point on the `empty()`. Did consider the number conversion but reading back in the code I found that m_gasmbuschannel is in fact a char value with possible values 0, 0x31..0x34 (i.e. the ASCII values for 1..4). That makes the number conversion rather pointless as we would to convert it back to a char value again (i.e. `result[0][0][0]`) to run our checks and set the right prefix for the gas values.

Obviously I can span the `if` statement over multiple lines if that would improve readability.

gizmocuz:
If you could span the IF that would be great!

gizmocuz:
Thanks!

PR #1879, opened 2017-10-16
Properly handle humidities with a decimal place returned by RTL_433

nahall:
Some RTL_433 devices return a humidity with a decimal place, which was causing Domoticz to throw a casting error. This change rounds the humidity and casts it properly.


nahall:
This pull request replaces #1843 which was made against the master branch.

gizmocuz:
Sorry, we got a merging issue. Is this patch still needed? This because we do not use the lexical cast anymore.
If yes, could you re-do this ?

nahall:
OK now that you made the change to remove boost's lexical cast it would appear that this patch is no longer needed since I was previously getting the exception thrown by the lexical cast. I'll close this, thanks.

PR #1881, opened 2017-10-17
Fix setup floorplan values

trandbert37:
Fix for display floorplan options that aren't retrieved after changes save.


gizmocuz:
Thanks!

PR #1883, opened 2017-10-21
Enable C++11 build and fix build errors

correa:


hardware/Kodi.h
```diff
@@ -9,8 +9,9 @@
 #include <boost/asio.hpp>
 #include <boost/array.hpp>
 #include <boost/enable_shared_from_this.hpp>
+#include <boost/lexical_cast.hpp>
```
gizmocuz, line 12: I'm not a fan of depending on lexical cast, cant we use the stringstream ? what is wrong with this

hardware/EcoDevices.cpp
```diff
@@ -286,7 +287,7 @@ void CEcoDevices::GetMeterDetails()
 	else
 	{
 		message = "EcoDevices firmware needs to be at least version ";
-		message = message + static_cast<std::ostringstream*>(&(std::ostringstream() << min_major << "." << min_minor << "." << min_release))->str();
+		message = message + boost::lexical_cast<std::string>( min_major ) + "." + boost::lexical_cast<std::string>(min_minor) + "." + boost::lexical_cast<std::string>(min_release);
```
gizmocuz, line 290: Whats wrong with stringstream ? not a fan of depending on lexical cast

CMakeLists.txt
```diff
@@ -1,12 +1,15 @@
 #set to minimum version that supports clean build on cygwin
-cmake_minimum_required(VERSION 2.8.4)
+cmake_minimum_required(VERSION 3.8.0)
```
gizmocuz, line 2: i just setup a fresh new Pi64 system, CMake version is 3.7.2, think that should at least be the minimum version?

CMakeLists.txt
```diff
@@ -1,12 +1,15 @@
...
 
 project(domoticz)
 
 ## required min. libBoost version
 SET(DOMO_MIN_LIBBOOST_VERSION 105500)
 ##
 
+set(CMAKE_CXX_STANDARD 11)
```
gizmocuz, line 10: Does this still build on a RPi ?

gizmocuz (commented):
Some thoughts....

CMakeLists.txt
correa: Will check the documentation for the CXX_STANDARD. I don't have a Pi board. Is there somewhere we can check the package versions available?

hardware/EcoDevices.cpp
correa: After enabling cpp11 LLVM on Mac fails to build with erros on those lines. Didn't save the error message, but I would guess it's about taking the address of a temporary. I guess there is no reason not to use operator . instead of operator ->

We are already depending on lexical_cast in other places btw.

gizmocuz: Yes and i hope to remove the dependencies soon...
I am all for c++11 as long as raspberry pi users for instance are able to build as well

CMakeLists.txt
gizmocuz: I'm not sure 100% , but i think you can find them here

https://www.raspbian.org/RaspbianRepository

but cant seem to get a good listing here

gizmocuz:
Sorry, need to close this until we can be sure it compiles on current hardware platforms like a raspberry pi (cmake version)

emontnemery:
@gizmocuz Was this closed because it didn't build on a RPi, or was it closed because it was unclear if it would build?

Edit: 
CMAKE_CXX_STANDARD was introduced in CMake 3.1.0, so no need to require newer CMake than that. Builds fine with CMAKE_CXX_STANDARD set to 11 on RPi 3 with Raspbian Stretch. I can try some other versions if needed. 
Instead of using boost::lexical_cast, std::to_string can be used.
See #2049 for more details.

correa:
I got Raspbian wheeze to run on qemu and I'm checking the tool versions. We got gc 4.6.3 there. It has most of the C++11 features in it, still have to test std::to_string.
cmake is 2.8.9 there.

Should we support wheeze or require newer version, Jessie perhaps. We would get gcc 4.9 in Jessie and gcc 6.3 in Stretch.

gizmocuz:
Jessie should be supported for now... would love to continue with gcc 6 and up but i think when we compile with 6, and try to run it on a jessie system it will fail... if not... then lets move to 6

PR #1887, opened 2017-10-22
add new hardware support : USBtin CAN Gateway and CAN device support

ghost:
Add new hardware support (USBtin CAN Gateway) working like a USB serial device to communicate with a CAN bus.
Flexible base in USBtin management to add more CAN layer in future.
Add CAN MultiblocV8 Layer support to communicate with SFSP (master or slave) blocks. 
- Up to 8 Master Blocks and 8 slaves blocks in the main CAN network can be connected. ( In multiblocV8 possible management of 8 sub networks with a CAN to CAN gateway, so increasement of max number blocks x8 ).
- In domoticz, for one block : ability to control the 6 PWMs channels, read the 4 wired input, reading power voltage, and for a Master blocks : read each EnOcean (wireless) switch receive !


hardware/USBtin.h
```diff
@@ -0,0 +1,101 @@
...
+
+#pragma once
+
+#include <vector>
+#include "ASyncSerial.h"
+#include "USBtin_MultiblocV8.h"
+#include "DomoticzHardware.h"
+
+
+#define USBTIN_BAUD_RATE         115200
```
gizmocuz: Is it possible to move all defines at the top of the .cpp file ?

hardware/USBtin_MultiblocV8.h
```diff
@@ -0,0 +1,216 @@
+#pragma once
+
+#include <iostream>
+#include <string>
+#include "DomoticzHardware.h"
+#include "hardwaretypes.h"
+
+#define	TIME_1sec				1000000
```
gizmocuz: Is it possible to move the defines to the top of the cpp file ?

main/RFXtrx.h
```diff
@@ -470,6 +470,8 @@ SDK version 4.9
 #define sTypeHEU 0x1
 #define sTypeANSLUT 0x2
 #define sTypeKambrook 0x03
+#define	sTypeSFSP_M 0x04
```
gizmocuz: Sorry you can not make any changes to this file.
Can we not use a normal on/off switch we already have? (a lighting2 ?)

www/html5.appcache
```diff
@@ -1,615 +1,615 @@
...
-json.htm
-ozwcp/ozwcp.html
-ozwcp/cp.html
-*
-
-# Fallback content
-FALLBACK:
-images images/iphone-icon.png
-/ views/offline.html
+CACHE MANIFEST
```
gizmocuz: Something as gone wrong patching the version number as every line seems to have a change, maybe it is the line ending.
Could you recheck this ?

main/RFXtrx.h
ghost: Ok, it was to clearly distinguish the origin of the command, because in multibloc we can have lot of blocks.
Maybe a over way to do that ? if not I still have the option to write the original reference in the default name when created...
image

gizmocuz: You can use the 'SendSwitch' command for this ?

hardware/USBtin_MultiblocV8.h
ghost: ok !

www/html5.appcache
ghost: Line ending !

ghost:
Allways a conflict but I thinks it's normal here, no ? For Appveyor I think I must check the project in Visual, no ?

gizmocuz:
Did you add the new source files also in the visual studio solution ? (Did you test with visual studio also ?)

gizmocuz:
What hardware do you use with the CAN system ?
Could you translate all french comments in this pull request?
(and add it to the visual studio project)

ghost:
Yes I'll translate. I'm not familiar with visual studio, I've install it but have lots of error and did not succeed to compile domoticz (including taking the basic release without my modifications). maybe a question of library, I'm trying to resolve it...

ghost:
Visual told me that he can't find boost/thread.hpp. I've allready past the folder "WindowsLibraries" in the msbuild project, so the complete link of the library is WindowsLibraries/Windows Libraries/all folder lib, is-it ok ?

gizmocuz:
See the Install.txt file. You need to download the libraries from a separate repository

ghost:
Yes, problem was between kayboard and screen (as usual) ;-) work in progress, I compiled my modification and all seems to be ok. I'll change comments soon.

ghost:
for information, the final product we control is this : http://www.scheiber.com/doc_technique/sfsp-2012-a1/  but I don't know where to put this information, maybe on forum ?

gizmocuz:
Could you add the link in the top of the .cpp file somewhere ?

gizmocuz:
Thanks!!

jvandenbroek:
Github complains about the big file msbuild/Debug/domoticz.pdb added with this PR, should probably be added to .gitignore? Also get this warning now, better always save with LF to be compatible with *nix/bsd:
`warning: CRLF will be replaced by LF in appversion.h.
The file will have its original line endings in your working directory.
warning: CRLF will be replaced by LF in hardware/USBtin_MultiblocV8.cpp.
The file will have its original line endings in your working directory.
warning: CRLF will be replaced by LF in hardware/USBtin_MultiblocV8.h.
The file will have its original line endings in your working directory.
`

gizmocuz:
This file should not be in here
Thanks for reporting

ghost:
If it's my fault sorry, don't know exactly what to do to not reproduce the same error. Is there a way to ignore some files with tortoise ? I use it with subversion and I can ignore some files or folder but with github I had errors when committing (error due to change of msbuild folder's property ).

gizmocuz:
No problem ! I removed them

PR #1897, opened 2017-10-28
EventSystem optimizations

jvandenbroek:
Used most of this code for a few weeks now, just optimized it a little bit further today.

- Passing reference to item instead of all values separately, much more efficient
- Use enum instead of string for passing reason, this was already there for the queue
- Added name mapping table / array for device name matching DB with enum
- Removed redundant code (double functions / if else checks)


gizmocuz:
Great optimization ! Thanks!

PR #1900, opened 2017-10-30
Add new Hardware Enphase plugin

pperichon:
Hi,
For my first contributing, I added a new hardware plugin for using API of Envoy Enphase product.
I use the API of Enphase product on local network with the IP address. 
I defined an interval of 30 seconds, and the calls are during the day between the sunrise and sunset with 1 hour of shift.
The informations received is the current production in Watt and the total production in kWatt/hour


gizmocuz (commented):
Because of a previous pull request merge, we got some merge issues here.
Mainly the hardware ID (107) is now being used for the USBtinGateway
i made some adjustments, and after the merge i will add the class files back into the visual studio sollution
If you are (probably) already using this, please change the hardware type from 107 to 108 in your database (or change it via the hardware setup page)

pperichon:
ok, thank you and sorry for the problem on merge.
I changed the hardware type to 108 in database on my raspberry.

gizmocuz:
That was not your problem, but it happened because of another pull request ;) Thanks again!

machielw:
2fb54633-658c-4725-ad21-04e2c002c13b
It adds 4 devices. P1 meter (Total)  and generaal production. But it adds it double, 2 of both.

PR #1901, opened 2017-10-30
Fix P1 day graph (DST)

gordonb3:
See attached image old versus new

### DIFF
--- a/main/WebServer.cpp
+++ b/main/WebServer.cpp
@@ -13137,7 +13137,8 @@ namespace http {
 						root["status"] = "OK";
 						root["title"] = "Graph " + sensor + " " + srange;
 
-						result = m_sql.safe_query("SELECT Value1, Value2, Value3, Value4, Value5, Value6, Date FROM %s WHERE (DeviceRowID==%" PRIu64 ") ORDER BY Date ASC", dbasetable.c_str(), idx);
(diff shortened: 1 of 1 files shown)


pepsi92:
Hi gordonb3,

:+1: 
Many thanks for this fix as I am facing this issue since last Sunday (DST time change :) )
I will test it as soon as the current beta will be compiled for the Synology Platform.

Nicolas

gizmocuz:
Would a call to 'FixDaylightSaving' also solve the problem ?

gordonb3:
That function doesn't seem to touch the table we are referencing here. Deleting one of the ranges will also cause the end of Summer time value (e.g. 3:00 AM continental Europe, 2:00 AM UK) to either show a big spike from usage being multiplied by 13 or a zero, possibly depending on platform (my Linux system appears to have a preference on Summer time for the 2:xx AM times).  I'd also not dare to bet that such a procedure will only keep the Summer time values as the ones that were inserted first.

Last, by keeping the values a user may pass the result from `json.htm?type=graph&sensor=counter&idx=[IDX]&range=day` to a graphing tool that allows a custom definition of the X-axis values.

gizmocuz:
Ok true indeed ;) thanks for the patch!

PR #1903, opened 2017-10-30
calculation of sunrise timers should not correct for DST change

gordonb3:


gordonb3:
Please hold on this one. I think there is a chance on a double trigger if random is used.

[edit] never mind - over complexed thought.

gizmocuz:
Thanks, but is there still something that needs FixMe'ing here ? as i see some comments in this pull request

gordonb3:
Obviously yes. If we want to set a timer relative to sunset or sunrise on a specific date it makes sense to use the sunset values from that date. We don't. We use today's values which is why the timers are actually off by a minute or two on normal days. This becomes painfully obvious on dates when DST switched either on or off because then the difference between the set time and the actual sunset values from that day is an hour.

So in essence this is a quick and dirty fix that will take care of the one hour difference, without solving the 1-2 minute difference. If or when we do, the marked block will have to be removed again.

gizmocuz:
Thanks! Ok, well i can live with the 2 minutes .... most timers also use a randomness interval here anyway ;)

gordonb3:
Most people can as well. It was noticed though at some point and mentioned in the forum.

Issue #1863, opened 2017-10-11
Level for JSON API/MQTT switchcmd "On" and switchcmd "Set Level" is not same


emontnemery:
For switchcmd "Set Level":
Accepted dim level is 1 .. Infinity, but will be decreased by 1:
Passing level 101 will set the dimlevel to 100
Passing level 100 will set the dimlevel to 99
Passing level 1 will set the dimlevel to 0
Passing level 0 is a special case: dimlevel is not updated and the switch will be turned off

For switchcmd "On":
Accepted dim level is 1 .. Infinity:
Passing level 101 will set the dimlevel to 101
Passing level 100 will set the dimlevel to 100
Passing level 1 will set the dimlevel to 1
Passing level 0 is a special case: dimlevel is not updated

It seems a bit strange that the same level is treated differently for "On" and "Set Level", which one can be considered correct?
Also, documentation should be updated to document the special cases: https://www.domoticz.com/wiki/Domoticz_API/JSON_URL%27s#Set_a_dimmable_light_to_a_certain_level

emontnemery:
Closed by #1906, #1921:

For switchcmd "Set Level"
Accepted dim level is 0 .. Infinity:
Passing level 10000 will set the dimlevel to max (100) and turn switch on
Passing level 50 will set the dimlevel to 50
Passing level 1 will set the dimlevel to 1
Passing level 0 will not update stored dimlevel
All values > 0 turns switch on
0 turns switch off

For switchcmd "On":
dim level is ignored, switch is turned on without updating dimlevel

For switchcmd "Off":
dim level is ignored, switch is turned off without updating dimlevel

PR #1906, opened 2017-10-31
Improve handling of dimlevel, fix #1842, #1863

### LINKED ISSUE #1842: Dim level set to 0 when turning switch on or off
When turning switch on or off by sending domoticz/in topic, dim level is set to 0 in domoticz/out MQTT message unless level parameter is included.

MQTT log:
```
2017-10-03 17:23:31 Topic: domoticz/in Qos: 0
{"command":"switchlight","idx":5,"switchcmd":"On"}
2017-10-03 17:23:31 Topic: domoticz/out Qos: 0
{ "Battery" : 255, "RSSI" : 12, "description" : "", "dtype" : "Light/Switch", "id" : "00014055", "idx" : 5, "name" : "Sonoff LED test", "nvalue" : 1, "stype" : "Switch", "svalue1" : "0", "switchType" : "Dimmer", "unit" : 1 }
2017-10-03 17:25:36 Topic: domoticz/in Qos: 0
{"command":"switchlight","idx":5,"switchcmd":"Off","level":"25"}
2017-10-03 17:25:36 Topic: domoticz/out Qos: 0
{ "Battery" : 255, "RSSI" : 12, "description" : "", "dtype" : "Light/Switch", "id" : "00014055", "idx" : 5, "name" : "Sonoff LED test", "nvalue" : 0, "stype" : "Switch", "svalue1" : "25", "switchType" : "Dimmer", "unit" : 1 }
2017-10-03 17:25:46 Topic: domoticz/in Qos: 0
```

emontnemery:
Improve handling of dimlevel,  should fix #1842, #1863.
This for example means Reacticz correctly shows dimlevel when a dimmer has been turned off and then back on again.

Maybe the check to only replace `level` with `LastLevel` if `switchtype != STYPE_Selector` is unnecessary.

### DIFF
--- a/main/mainworker.cpp
+++ b/main/mainworker.cpp
@@ -10805,8 +10805,9 @@ bool MainWorker::SwitchLightInt(const std::vector<std::string> &sd, std::string 
 		switchcmd = (atoi(sd[7].c_str()) == 1 ? "Off" : "On");
 	}
 
-	//when level = 0, set switch command to Off
-	if (switchcmd == "Set Level")
+	//adjust level
+	if (switchcmd=="Set Level" ||
+		switchcmd=="On")
 	{
 		if (
 			(level > 0) &&
@@ -10815,8 +10816,22 @@ bool MainWorker::SwitchLightInt(const std::vector<std::string> &sd, std::string 
 		{
 			level -= 1;
 		}
-		if (level == 0)
-			switchcmd = "Off";
+		//when level = 0 and command is "Set Level", set switch command to Off
+		if (level==0 && switchcmd=="Set Level")
+			switchcmd="Off";
+	}
+
+	//when level = 0 and command is "On" replace level with "LastLevel"
+	if (switchcmd=="On" && level == 0 && switchtype != STYPE_Selector)
+	{
+		//Get LastLevel
+		std::vector<std::vector<std::string> > result;
+		result = m_sql.safe_query(
+		"SELECT LastLevel FROM DeviceStatus WHERE (HardwareID=%d AND DeviceID='%q' AND Unit=%d AND Type=%d AND SubType=%d)", HardwareID, sd[1].c_str(), Unit, int(dType), int(dSubType));
+		if (result.size() == 1)
+		{
+			level = atoi(result[0][0].c_str());
+		}
 	}
 
 	//


main/mainworker.cpp
```diff
@@ -11741,7 +11755,7 @@ bool MainWorker::SwitchLight(const uint64_t idx, const std::string &switchcmd, c
     if (_log.isTraceEnabled()) _log.Log(LOG_TRACE,"MAIN SwitchLight idx:%d cmd:%s lvl:%d " ,(long)idx,switchcmd.c_str(),level );
 	std::vector<std::vector<std::string> > result;
 	result = m_sql.safe_query(
-		"SELECT HardwareID,DeviceID,Unit,Type,SubType,SwitchType,AddjValue2,nValue,sValue,Name,Options FROM DeviceStatus WHERE (ID == %" PRIu64 ")",
+		"SELECT HardwareID,DeviceID,Unit,Type,SubType,SwitchType,AddjValue2,nValue,sValue,Name,Options,LastLevel FROM DeviceStatus WHERE (ID == %" PRIu64 ")",
```
gizmocuz: I dont see where you use the 'LastLevel' from the sql query.. maybe this can be removed here (could save some cpu/disk cycles)

gizmocuz:
Thanks, got a quick question up stairs

main/mainworker.cpp
emontnemery: LastLevel is used here: `level = atoi(sd[11].c_str());`
Should I clean it up to add a comment sd[11] holds LastLevel?

Level will be 0 if it was not passed either by JSON API or MQTT API, then stored LastLevel will be used instead for switchcmd="On". 
A cleaner solution would be to modify the code such that missing value is indicated with an illegal level, for example -1, but it requires much more changes to the code.

Also, is it unnecessary to overwrite level with LastLevel only if  switchtype != STYPE_Selector, or in other words is '0' a valid setting for STYPE_Selector? For dimmer level=0 is a special case: turn off in case of switchcmd="Set Level", no change in case of switchcmd="On".

gizmocuz: Ok, now i understand, this is used in MainWorker::SwitchLightInt
But then, you also have to push something in the WebServer.cpp , line 4627 is calling SwitchLightInt also)

emontnemery:
Here's an updated version where LastLevel is retrieved locally in MainWorker::SwitchLightInt.
Edit: I squashed the commits, sorry about the mess.

gizmocuz:
Thanks!

szczukot:
@emontnemery , maybe You also will resolve this bug:
http://www.domoticz.com/forum/viewtopic.php?f=6&t=19887&p=154065#p154065
?

emontnemery:
@szczukot After merge of this PR, dimlevel is treated like below, how do you suggest it should work instead?

**For switchcmd "Set Level":**
Accepted dim level is 1 .. Infinity, but will be decreased by 1:
Passing level 101 will set the dimlevel to 100
Passing level 100 will set the dimlevel to 99
Passing level 1 will set the dimlevel to 0 which is a special case: dimlevel is not updated and the switch will be turned off
Passing level 0 is a special case: dimlevel is not updated and the switch will be turned off

**For switchcmd "On":**
Accepted dim level is 1 .. Infinity, but will be decreased by 1:
Passing level 101 will turn on the switch and set the dimlevel to 100
Passing level 100 will turn on the switch and set the dimlevel to 99
Passing level 1 will turn on the switch and set the dimlevel to 0
Passing level 0 is a special case: turn on the switch but dimlevel is not updated (previous dimlevel will be used instead)

**For switchcmd "Off":**
dim level is ignored, switch is turned off without updating dimlevel

szczukot:
I don't known, but as it is now it is terrible.
User set dimmer level to 70 and see .... 69 ?? 
Maybe (simple and logical for all):

**For switchcmd "Set Level" - only set level!**
Accepted dim level is 0 .. Infinity:
Passing level 10000 will set the dimlevel to 100
Passing level 50 will set the dimlevel to 50
Passing level 0 will set the dimlevel to 0

**For switchcmd "On":**
dim level is ignored, switch is turned on without updating dimlevel

**For switchcmd "Off":**
dim level is ignored, switch is turned off without updating dimlevel

or second version (not good I think)

**For switchcmd "Set Level":**
Accepted dim level is -1 .. Infinity:
Passing level 1000 will set the dimlevel to 100
Passing level 50 will set the dimlevel to 50
Passing level 0 will set the dimlevel to 0
Passing level -1 is a special case: dimlevel is not updated and the switch will be turned off

**For switchcmd "On":**
Accepted dim level is -1 .. Infinity:
Passing level 1000 will turn on the switch and set the dimlevel to 100
Passing level 50 will turn on the switch and set the dimlevel to 50
Passing level 0 will turn on the switch and set the dimlevel to 0
Passing level -1 is a special case: turn on the switch but dimlevel is not updated (previous dimlevel will be used instead)

**For switchcmd "Off":**
dim level is ignored, switch is turned off without updating dimlevel

emontnemery:
@gizmocuz What's your opinion on decrementing dim level?
It was removed in b4bfd5314204539f6552b5d589358b19a661a3ac, but then reintroduced again in bbab7d1a5674249888d8aa9d5f109fd4b97e996a just a few days later..
I agree with @szczukot decreasing dim level by one seems a bit strange, especially because it's not according to documentation here: https://www.domoticz.com/wiki/Domoticz_API/JSON_URL's#Set_a_dimmable_light_to_a_certain_level.

edit: Domoticz standard web UI and Reacticz pass 1 to set dimlevel to 0/off and max+1 to set dimlevel to max. If the current way is widely used, this is more of a documentation issue?

gizmocuz:
It has been quite some time since i coded that... but it had a reason... 
i think 0 means OFF in domoticz
I agree it looks weird... starting counting at 1 instead of 0

szczukot:
But we have command "off" - and we may switch off device (without changing level)
"Set Level"should only set the level - not interfere in On/Off
We should not change (decrease) value set by the user (user set 50 and we save and show 49).

Maybe: 
**For switchcmd "Set Level"**
Accepted dim level is 0 .. Infinity:
Passing level 10000 will set the dimlevel to 100
Passing level 50 will set the dimlevel to 50
Passing level 1 will set the dimlevel to 1
Passing level 0 will set the dimlevel to 0 **and switch off
And maybe all value > 0 turn switch on**

**For switchcmd "On":**
dim level is ignored, switch is turned on without updating dimlevel

**For switchcmd "Off":**
dim level is ignored, switch is turned off without updating dimlevel

gizmocuz:
The maximum value of accepted dim levels depends per device.
This you can see in the GetLightStatus call
It should also be transparent to the end-user, not giving him a slider in the gui with 100 possible positions when only 15 are possible. He/She would expect a dim level to be set on 27%, but in reality it is 32% because of the xx combinations
Setting a dim level to 0 does not make sense at all, and should be considered to be an OFF command
Any other value should be an ON with level (or a direct SetLevel command if the hardware supports it)
In the GUI at the moment, if the user set it to 17%, it will we 17%, i do not know what hardware you are controlling with the GUI that does not do this, but maybe have a quick look there
I know from my KaKu devices it works as designed
Of course for ZWave sensors (and others) that do support 100 level value it is a lot easier.
Also here, level 0 means OFF (as written in the manuals). level 255 puts it on, with the last remembered dim level (in the device itself)

emontnemery:
@gizmocuz With current implementation it's really confusing:
Dimlevel in incoming "Set Level" json and mqtt message will be decremented by 1
Dimlevel in outgoing message, either json query or mqtt message is not incremented by 1
This means a switch communicating over mqtt should

emontnemery:
@gizmocuz I fully agree that it is perfectly logical that dimlevel 0 is translated to off.
With current implementation it's however *really* confusing due to the decrement of dimlevel by 1:
Dimlevel in **incoming** "Set Level" json and mqtt message will be decremented by 1, except if incoming dimlevel is 0. Hence, incoming dimlevel of 0 **and 1** are both translated to "off" command which makes less sense.
Dimlevel in **outgoing** message, either through json query or mqtt message is *not* incremented by 1 causing the issue @szczukot mentions: user sets one level but sees something else.

Regarding MaxDimLevel:
What is reported for both KaKu switch through RFLink and MQTT switch (through dummy) is 100 through JSON `devices` cmd. However, maximum level is not 100, it's 101 due to the decrement of 1.

So, this all means. 
For KaKu: 
To turn KaKu dimmer off, pass dimlevel of 0..1
To set minimum dim level of KaKu dimmer (0),  pass dimlevel of 2..4
To set maximum dim level of KaKu dimmer (15), pass dimlevel of 98..101

For Tasmota PWM LED dimmer through MQTT: 
To turn Tasmota dimmer off, pass dimlevel of 0..1
To set minimum dim level of Tasmota dimmer (1), pass dimlevel of 2
To set maximum dim level of Tasmota dimmer (100), pass dimlevel of 101

emontnemery:
I propose to remove the decrement by 1, while also update:
- Webserver and mqtt code to pass level=-1 in case the level is not set. Today 0 is used to signal illegal value, which might be why the decremetn by 1 is there in the first place?
- Domoticz web UI

*Hence, for switchcmd "Set Level":*
Valid dim level is 0 .. 100:
Passing level 100 will set the dimlevel to 100 (not same as today, but matches wiki documentation and MaxDimLevel value)
Passing level 50 will set the dimlevel to 50
Passing level 0 is a special case: dimlevel is not updated and the switch will be turned off (same as today)

*For switchcmd "On":*
Valid dim level is 0 .. 100:
Passing level 100 will set the dimlevel to 100 (not same as today, but matches wiki documentation and MaxDimLevel value)
Passing level 50 will set the dimlevel to 50
Passing level 0 is a special case: dimlevel is not updated but the switch will be turned on (same as today)
If level is not set, dimlevel is not updated and the switch will be turned on (same as today)

szczukot:
Yes, this is good solution.
But one question: whats mean "If level is not set,...." ? Is that means level == -1 ?
And
"Passing level 0 is a special case: dimlevel is not updated and the switch will be turned off (same as today)"
should be
"Passing level 0 is a special case: dimlevel is updated and the switch will be turned off (same as today)"
because this is "Set level" command (not "Off" command)

emontnemery:
@szczukot If `level` is not included or malformed in MQTT or JSON message, `level` must either be assigned some default value or MQTT or JSON message should be ignored. Currently, default value is set to 0.

szczukot:
Ok. That proper default value should be -1 (unset)

emontnemery:
@szczukot This is what is implemented in #1921:
For switchcmd "Set Level"
Accepted dim level is 0 .. Infinity:
Passing level 10000 will set the dimlevel to max (100) and turn switch on
Passing level 50 will set the dimlevel to 50
Passing level 1 will set the dimlevel to 1
Passing level 0 will not update stored dimlevel
All values > 0 turns switch on
0 turns switch off

For switchcmd "On":
dim level is ignored, switch is turned on without updating stored dimlevel

For switchcmd "Off":
dim level is ignored, switch is turned off without updating stored dimlevel

szczukot:
Ok. I think that this is better then before.

PR #1919, opened 2017-11-07
Added support for MCP23017: 16bit I2C GPIO expander

gneluka:
Details:
Added support for Microchip's MCP23017: 16 bit I2C GPIO expander:

Only output mode is supported currently.
Hardcoded to use I2C port 1 on RPI 2 and 3.
No dependency on wiringpi library.
Output value is inverted on gpio ouput pin.( switch on == 0, switch off == 1)
Restore previous output pin status after power outage or system reboot.

Thanks

### DIFF
--- a/hardware/I2C.cpp
+++ b/hardware/I2C.cpp
@@ -11,6 +11,7 @@ This code supports:
 - PCF8574, PCF8574A
 - HTU21D
 - TSL2561
+- MCP23017
 
 PS there are better density and QNH formulas.
 
@@ -18,6 +19,7 @@ Authors:
 - GizMoCuz
 - Eric Maasdorp
 - Ondrej Lycka (ondrej.lycka@seznam.cz)
+- Guillermo Estevez (gestevezluka@gmail.com)
 */
 
 #include "stdafx.h"
@@ -38,6 +40,7 @@ Authors:
 #include "../main/RFXtrx.h"
 #include "../main/localtime_r.h"
 #include "../main/mainworker.h"
+#include "../main/SQLHelper.h"
 
 #define round(a) ( int ) ( a + .5 )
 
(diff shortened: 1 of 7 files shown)


gizmocuz:
Thanks for the great work !

gizmocuz:
Question, will this also not invert the PCF8574 ?
Why do you have it inverted ?

gizmocuz:
I made some tiny changes in the latest commit
At the top there is now a //#define INVERT_PCF8574_MCP23017
if you uncomment this, it is inverted
But i would like to know why it is inverted at your place ? On=1 , Off=0, that is the logic used everywhere
If it is different at your place, should you not use something like flip flops ?

gneluka:
Hi Gizmocuz, 

Thanks for merging this. 
The reason why I use inverted logic has nothing to do with the chip itself. It's because I'm using it to control an optoisolated relay module that's active low.  At first I thought of adding some kind of checkbox on the hardware setup screen to give the user the possibility to enable/disable inverted logic.  However when I looked at the code for PCF8574 I saw it was already using inverted, and since that was exactly what I needed I reused some of that code. 
I'm planning to add the functionality to allow the user to enable/disable the inverted logic on the hardware configuration screen.  What's your suggestion for storing this flag? I thought about using one bit of the hardware id field, since that's already used to store the IC2 address and GPIO number.
I agree with your change. It makes sense, as not everybody will want the inverterd logic when using the MCP23017, 

Thanks

gizmocuz:
Thanks, that would be great if you could add a checkbox for it

Vulpecula-nl:
Nice you have made this expander available in Domoticz. But there is an issue with it. When I choose this at hardware I can't set the address of the chip. I have 2 MCP23017 chips and both chips are not on address 0x0 but 0x20 and 0x21 I need to set the address before it can work. Could you develop this aswell?

Sorry forgot to mention the version I'm on 
Version: 3.8742
Build Hash: 6bc9850f
Compile Date: 2017-11-18 08:43:56

[Edit]
I see the beta doesn't have this commit. Is it possible to add this to the beta?

gneluka:
Hi Vulpecula-nl,

This commit is included on latest beta, I've just tested it myself.
What you describe sounds like a browser cache issue
If you first installed the stable version (which does not include this commit) and then upgrade to beta you need to clear your cache so that the new version of "HardwareController.js" is loaded.  I had this same issue and was able to solve it following the step listed here: https://easydomoticz.com/les-caches-des-navigateurs-et-domoticz/
Once that is complete you'll be able so enter the IC2 address for your expander, please note that you must enter the address in decimal format, ie: 0x20=32 and 0x21 = 33.
Guillermo

Vulpecula-nl:
Sorry about my late reply but I have finally working. There is only one thing what is bothering me. When my Raspberry Pi needs to reboot, the initial pinout states of the MCP23017 are High. I want to change this to Low. Using all pins as output for switching lights, when the Raspberry is booted all the lights are on. I have tried the "inverse" option but all the pin states keeps High after booting.

gizmocuz:
@Vulpecula-nl , Domoticz can not change that.
You either have to use an inverter (actual hardware)

gneluka:
@Vulpecula-nl, if you are using latest version from development branch, domoticz should restore previous pin states on MPC23017 after reboot. I've been using it for several months already and can confirm it works as expected, also I can confirm the 'Invert GPIO' function works as expected.

Vulpecula-nl:
@gneluka Mine is not! Everytime when I reboot it, all pins are set to High. What I did in November last year was compiled domoticz my self and used it. But last week I updated to version 3.9578 via the domoticz update service. Could this be causing the problem (maybe something with caching?).

gneluka:
@Vulpecula-nl, could you please provide more details about what you are seeing? You mentioned all pins are set to high after reboot, what happens with the domoticz switches? Are they set to ON? What was the status before the reboot? When you reboot the raspberry PI do you also power cycle the MPC23007?

Vulpecula-nl:
I don't see any errors or warnings. The switches in Domoticz are all Off after reboot, but the actual pin is High. In domoticz I will switch them all On, so the state in Domoticz is the same as the actual pins.  And then I switch them all Off again, because they where Off before the reboot.

gneluka:
I took a look at the code and found what seems to be an issue with the initialization of the MCP23017 when invert GPIO is not selected. I believe the right process would be to open an issue and assign it to me.

gizmocuz:
@gneluka , feel free to create a pull request/issue

gneluka:
@Vulpecula-nl , I've fixed the initialization issue. The change was merged into the development branch. Please retest and let me know if any issue.
Thanks

PR #1921, opened 2017-11-07
Don't decrement dimlevel

emontnemery:
As discussed in #1906:
Changes: 
- Domoticz will not decrement dim level from user by 1
- Webserver and mqtt code will pass level=-1 to mainworker in case switchcmd is "On" or "Off".
 (The -1 value is only used for error handling, and will not be visible in MQTT Domoticz/Out message or HTTP JSON API query response)
- Domoticz web UI will not increment dim level by 1

**For switchcmd "Set Level"**
Accepted dim level is 0 .. Infinity:
Passing level 10000 will set the dimlevel to max (typically 100) and turn switch on
Passing level 50 will set the dimlevel to 50 and turn switch on
Passing level 1 will set the dimlevel to 1 and turn switch on
Passing level 0 will not update stored dimlevel and turn switch off

**For switchcmd "On":**
dim level is ignored, switch is turned on without updating stored dimlevel

**For switchcmd "Off":**
dim level is ignored, switch is turned off without updating stored dimlevel

Edit: Updated summary based on what was finally implemented based on dicussion

### DIFF
--- a/hardware/MQTT.cpp
+++ b/hardware/MQTT.cpp
@@ -262,13 +262,18 @@ void MQTT::on_message(const struct mosquitto_message *message)
 		std::string switchcmd = root["switchcmd"].asString();
 		if ((switchcmd != "On") && (switchcmd != "Off") && (switchcmd != "Toggle") && (switchcmd != "Set Level"))
 			goto mqttinvaliddata;
-		int level = 0;
-		if (!root["level"].empty())
+		int level = -1;
+		if (switchcmd == "Set Level")
 		{
-			if (root["level"].isString())
-				level = atoi(root["level"].asString().c_str());
+			if (root["level"].empty())
+				goto mqttinvaliddata;
 			else
-				level = root["level"].asInt();
+			{
+				if (root["level"].isString())
+					level = atoi(root["level"].asString().c_str());
+				else
+					level = root["level"].asInt();
+			}
 		}
 		if (!m_mainworker.SwitchLight(idx, switchcmd, level, -1, false, 0) == true)
 		{
@@ -288,15 +293,20 @@ void MQTT::on_message(const struct mosquitto_message *message)
 		if ((switchcmd != "On") && (switchcmd != "Off") && (switchcmd != "Toggle") && (switchcmd != "Set Level"))
 			goto mqttinvaliddata;
 			
-		int level = 0;
-		if (!root["level"].empty())
+		int level = -1;
+		if (switchcmd == "Set Level")
 		{
-			if (root["level"].isString())
-				level = atoi(root["level"].asString().c_str());
+			if (root["level"].empty())
+				goto mqttinvaliddata;
 			else
-				level = root["level"].asInt();
+			{
+				if (root["level"].isString())
+					level = atoi(root["level"].asString().c_str());
+				else
+					level = root["level"].asInt();
+			}
 		}
-		
+
 		int hue = 0;
 		if (!root["hue"].empty())
 		{
--- a/hardware/MQTT.cpp
+++ b/hardware/MQTT.cpp
@@ -267,13 +267,10 @@ void MQTT::on_message(const struct mosquitto_message *message)
 		{
 			if (root["level"].empty())
 				goto mqttinvaliddata;
+			if (root["level"].isString())
+				level = atoi(root["level"].asString().c_str());
 			else
-			{
-				if (root["level"].isString())
-					level = atoi(root["level"].asString().c_str());
-				else
-					level = root["level"].asInt();
-			}
+				level = root["level"].asInt();
 		}
 		if (!m_mainworker.SwitchLight(idx, switchcmd, level, -1, false, 0) == true)
 		{
@@ -298,13 +295,10 @@ void MQTT::on_message(const struct mosquitto_message *message)
 		{
 			if (root["level"].empty())
 				goto mqttinvaliddata;
+			if (root["level"].isString())
+				level = atoi(root["level"].asString().c_str());
 			else
-			{
-				if (root["level"].isString())
-					level = atoi(root["level"].asString().c_str());
-				else
-					level = root["level"].asInt();
-			}
+				level = root["level"].asInt();
 		}
 
 		int hue = 0;
--- a/main/WebServer.cpp
+++ b/main/WebServer.cpp
@@ -3311,7 +3311,9 @@ namespace http {
 					(isscene == "")
 					)
 					return;
-				int level = atoi(request::findValue(&req, "level").c_str());
+				int level = -1;
+				if (request::hasValue(&req, "level"))
+					level = atoi(request::findValue(&req, "level").c_str());
 				int hue = atoi(request::findValue(&req, "hue").c_str());
 
 				unsigned char command = 0;
@@ -3428,7 +3430,9 @@ namespace http {
 					std::map<std::string, std::string> options = m_sql.BuildDeviceOptions(result[0][6].c_str());
 					GetLightCommand(dType, sType, switchtype, scommand, command, options);
 				}
-				int level = atoi(request::findValue(&req, "level").c_str());
+				int level = -1;
+				if (request::hasValue(&req, "level"))
+					level = atoi(request::findValue(&req, "level").c_str());
 				int hue = atoi(request::findValue(&req, "hue").c_str());
 				root["status"] = "OK";
 				root["title"] = "UpdateSceneDevice";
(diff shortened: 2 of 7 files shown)


gizmocuz:
Most (If not all) switches should turn ON to the last known dim level.
For this reason i so no reason to specify a dim level for the ON command, we have a 'Set Level' command for this.
Same for OFF, and OFF should turn it OFF and there is no need for a dim level there.

emontnemery:
@gizmocuz Setting of dim level through "On" command is documented here:
https://www.domoticz.com/wiki/MQTT#Sending_a_Switch_Command
That functionality is not changed by this patch, do you suggest to remove that option?

gizmocuz:
Yes.. that would make things more logical

emontnemery:
OK, updated + some cleanup.

gizmocuz:
Please use atoi instead of str2int

emontnemery:
atoi doesn't have any error handling, it simply returns 0 for empty or invalid input.

Hence introducing a new helper function, str2int, to be able to distinguish invalid input from valid input.
The way it works now is that it only updates the passed reference in case of valid input.
Maybe it would be more clear if it is changed such that the converted value is returned, and in case of invalid input, a user defined default value is returned (this would be closer to atoi):
`int str2int(const std::string &s, int base = 0, int default = 0);`

Or, another way: revert back to atoi() but check if "level" value was passed instead of blindly converting.
For example, around line 3314 of main/WebServer.cpp

```
 -				int level = atoi(request::findValue(&req, "level").c_str());
 +				int level = -1;
 +				std::string level_s = request::findValue(&req, "level");
 +				if (level_s != "") level = atoi(request::findValue(&req, "level").c_str());
```

gizmocuz:
We do not need error handling there?
first, check if a value is provided, if not.. bail out
else atoi

gizmocuz:
Everywhere you should check if parameters are provided from a web request....

emontnemery:
Ah, you mean instead of silently accepting a "Set Level" command without a level parameter (as today) bail out and generate an error? For both MQTT and web server?

gizmocuz:
No....
If you use the set level command, a level has to be provided... if not... give an error, bail out
use 'atoi' instead of a new function to convert from string to integer, no need for error check, after ATOI you can do this (level < 0 or >100 or higher then the max dim level per hardware)

emontnemery:
OK, reverted back to atoi and now bailing out if there is no dimlevel for "Set Level", both MQTT and WebServer.

hardware/MQTT.cpp
```diff
@@ -262,13 +262,18 @@ void MQTT::on_message(const struct mosquitto_message *message)
...
+		int level = -1;
+		if (switchcmd == "Set Level")
 		{
-			if (root["level"].isString())
-				level = atoi(root["level"].asString().c_str());
+			if (root["level"].empty())
+				goto mqttinvaliddata;
 			else
-				level = root["level"].asInt();
+			{
```
gizmocuz: this 'else' and block is not needed as the condition will always be true because of the previous check and 'goto'
suggestion, remove the else and the { and }

hardware/MQTT.cpp
```diff
@@ -288,15 +293,20 @@ void MQTT::on_message(const struct mosquitto_message *message)
...
```
gizmocuz: this 'else' and block is not needed as the condition will always be true because of the previous check and 'goto'
suggestion, remove the else and the { and }

main/WebServer.cpp
```diff
@@ -6404,11 +6408,13 @@ namespace http {
...
 				std::string switchcmd = request::findValue(&req, "switchcmd");
-				std::string level = request::findValue(&req, "level");
+				std::string level = "-1";
+				if (switchcmd == "Set Level")
+					level = request::findValue(&req, "level");
 				std::string onlyonchange = request::findValue(&req, "ooc");//No update unless the value changed (check if updated)
 				if (_log.isTraceEnabled()) _log.Log(LOG_TRACE, "WEBS switchlight idx:%s switchcmd:%s level:%s", idx.c_str(), switchcmd.c_str(), level.c_str());
 				std::string passcode = request::findValue(&req, "passcode");
-				if ((idx == "") || (switchcmd == ""))
+				if ((idx == "") || (switchcmd == "") || ((switchcmd == "Set Level") && (level == "")) )
```
gizmocuz: its a matter of taste, but i would recommend to use string.empty() instead of "" these days

main/mainworker.cpp
```diff
@@ -12455,7 +12448,7 @@ bool MainWorker::SwitchScene(const uint64_t idx, std::string switchcmd)
 			_log.Log(LOG_NORM, "Activating Scene/Group Device: %s (%s)", DeviceName.c_str(), lstatus.c_str());
 
 
-			int ilevel = maxDimLevel - 1;
+			int ilevel = maxDimLevel - 1; // Why -1?
```
gizmocuz, line 12451: This had something to do with KAKU dimmers... do you own those ?
Else i have to dig into my boxes to find one...
I believe it will blink the light or something... but best to test

www/app/DashboardController.js
```diff
@@ -3967,16 +3967,16 @@ define(['app'], function (app) {
 			$('#dashcontent .dimslider').slider({
 				//Config
 				range: "min",
-				min: 1,
-				max: 16,
-				value: 5,
+				min: 0,
+				max: 15,
```
gizmocuz, line 3971: this is of course cosmetic as it would be set below in the create function ;)

gizmocuz (commented):
Thanks!... just a few questions/remarks

main/WebServer.cpp
emontnemery: Agreed, but there are 310 occurences of `== ""` in main/WebServer.cpp
A change to string.empty() should perhaps be done as a separate "cleanup" pull request?

main/mainworker.cpp
emontnemery: No, sorry, I don't have any KAKU dimmers.
I have some Proove dimmers which blink if ON is sent twice (to allow dimming with an ON/OFF button), not sure if this is what you mean?

www/app/DashboardController.js
emontnemery: Not purely cosmetic because `min` is not set by create function.

gizmocuz: true ;)

main/mainworker.cpp
gizmocuz: i think those are the same...

main/WebServer.cpp
gizmocuz: yes, and if we keep adding code like that it will be a xxxxx of a job to change that later do we ?
so best to start doing it right when possible

main/mainworker.cpp
emontnemery: In case the switch is a dimmer, ilevel is overwritten by this block a few lines down:
```
if (lstatus == "On")
{
	lstatus = "Set Level";
	float fLevel = (maxDimLevel / 100.0f)*level;
	if (fLevel > 100)
		fLevel = 100;
	ilevel = round(fLevel);
} 
```
Anyhow, I did some test with my Proove dimmer, blinking happens if sending "ON" when already turned on.
Nothing special happens when set to max dimlevel (15).

main/WebServer.cpp
emontnemery: OK, done.

gizmocuz:
Thanks for the great work !

gizmocuz:
this aint working correctly now and i would expect some users comments on the forum
Setting my KaKu to 20% , waiting 10+ seconds, and the switch goes to 26% on the web gui (switches tab)
Maybe 20% was not even possible earlier....

gizmocuz:
I think this should be reproducible with a dummy AC switch

emontnemery:
@gizmocuz Was this behavior introduced by my changes? How to reproduce it?
I guess the issue is caused by setting a percentage in WebUI, which is then rounded to 4/16? Do you use some RF gateway which might be replying back with a rounded value?

My KaKu-compatible dimmer controlled through RFLink is reported as having 100 levels, and I can set it from 0% .. 100% in Web UI. Same if I add a dummy switch, and set it to type 'dimmer'.
http://192.168.0.3:8080/json.htm?type=devices&rid=84:
```
{
   "ActTime" : 1510492958,
   "ServerTime" : "2017-11-12 14:22:38",
   "Sunrise" : "07:38",
   "Sunset" : "16:04",
   "result" : [
      {
        -- snip -- 
        "MaxDimLevel" : 100,
        -- snip --
         "idx" : "84"
      }
   ],
   "status" : "OK",
   "title" : "Devices"
}
```

PR #1922, opened 2017-11-07
Security Panel

DewGew:
I suggest this better looking security panel


gizmocuz:
I have no idea ? I only see here some mere pull requests and some other text... ;)
Could you post a screenshot with your pull request ?

DewGew:
secpanel

I just added  this securitypanel. The old one is just renemed to index.html.old

gizmocuz:
Thanks for the image.
I think this is a matter of taste... I liked my original one.
Would also add some spaced between the title
And i am not sure if this if the image is 100% royalty free, even if people use Domoticz commercial wise
https://www.photohdx.com/pic-900/metal-door-with-gold-security-lock-background

evo11x:
@gizmocuz 
The image is 100% free for commercial use as long as the resolution is the free one which is available on the website and as long as you have posted the link here it's enough :) I am the author of the website. You can use any image which you like from https://www.photohdx.com

PR #1929, opened 2017-11-09
Adding Danalock V3 to Polycontrol

hhward:


gizmocuz:
Thank you for the configuration file.
You have to submit this file to the openzwave team, Domoticz will from time to time poll for changes and update the configuration files received from OZW

hhward:
Okay, thanks for the info. Will do that.

PR #1930, opened 2017-11-09
Add support for UART ports on Armbian

jvandenbroek:
On Armbian (www.armbian.com) /dev/ttyS* is being used for UART (in my case ttyS3). This patch checks for the existence of /etc/armbian-release and only then consider including ttyS* (I suppose there is a reason these aren't included by default?).


gizmocuz:
This has been requested a lot of times.
There are systems where there are 99 Sxx files.
Best to make a link, see our persistent devices wiki

jvandenbroek:
Ah OK, will do, sorry.

PR #1933, opened 2017-11-10
Remove double percentage signs (%%) from notifications

harryd100:
NUC CPU Percentage is 88.8 %**%** [> 80.0 %**%**]


gizmocuz:
Thanks!

PR #1935, opened 2017-11-10
Move Evohome Web installation parameters to where mainworker expects …

gordonb3:
…them to be

Refactoring/code cleanup in mainworker caused the Evohome via Web API hardware class to always be called with installation parameters [0,0,0] regardless of what the user specified.


gizmocuz:
Thanks!

PR #1942, opened 2017-11-11
Alert device text improvements

Olifant1990:
Alert devices can handle newlines (\r\n, \n\r, \n of \r).
Those newlines are translated to <br /> at the Utility tab and on the
Dashboard.
Therefore the devices on the Dashboard can grow in height, as an Alert
device can contain somewhat more text.


gizmocuz:
Thanks!
Is this also possible for the text sensor ?
Is this also working on the floorplan ?
Could you post a screenshot of the min-height change, or how it looks when it is high ?

Olifant1990:
@gizmocuz 
- Also done for the text sensor
- Floorplan I don't know, can you point me where the device content is generated for those devices on the plan?
- See attachements. In the Utility tab is was already expanding on larger text. I have copied this behaviour in the dashboard for favorite devices. See legacy.css, changed height property to min-height
dashboard
utility

gizmocuz:
Thanks m and thanks for the screenshots... that is a lot of text....
You need to add the text and alert sensor to a roomplan, next create a floorplan (you can use the default image), add the room plan to the floor position the devices and press the update button
From the settings, you have to enable to show the floorplan

The floorplan devices are created in www\js\domoticzdevices.js

Olifant1990:
@gizmocuz 

Yes, it is a lot. But I made it a lot to test it with the newlines.
I know how to place devices in the floorplan, I had to be more specific to ask for your last line: I only needed to know what .js file is responsible for the data. Sorry ;)

I will have a look into it soon. Thanks!

Olifant1990:
@gizmocuz 
I have changed it for the floorplan too.
floorplan

gizmocuz:
Thanks for great work!

PR #1947, opened 2017-11-13
EventSystem fix update device on set used/unused

jvandenbroek:
Fixes this: http://www.domoticz.com/forum/viewtopic.php?f=6&t=19250
File USBtin_MultiblocV8.h now with correct line endings.


gizmocuz:
Thanks for the fixes !

PR #1948, opened 2017-11-14
Addding feynman theme v2

flatsiedatsie:
A pragmatic way to have two new themes for Domoticz.
- Feynman (colourful)
- K-Tron (dark)

Changes:
- Only added the images that override the Domoticz base
- Uses JS to add some CSS classes that didn't make it into Domoticz (yet).


flatsiedatsie:
Weird that that check failed, as no Domoticz files are overwritten by this proposal (just 2 new theme folders created).

gizmocuz:
I agree...
I think the best way forward with this is to create a sub project/repo for themes
Users can then use the download button on github to download a theme and place it in the domoticz www folder
At the moment when we merge this, this will cause a much larger binary archive

flatsiedatsie:
That's dissapointing, as my main goal is to let the themes come with Domoticz as standard!

I would propose to just drop the included images and fonts. This will greatly reduce size, making them about as big as the other ones that come as standard.

People can then download an icon-pack if they want other icons.

gizmocuz:
That would partly solve the main problem yes.
But what if there will be 50 themes ?
A few standard themes would be nice, but a theme system would be better in the feature

flatsiedatsie:
In the future that would be better, sure. But I'd love it if for now these two themes could come with Domoticz (leaving the images and fonts out). Is that ok?

gizmocuz:
I see three themes here, feyn, feynman and ktron
You also make use of an old visual studio project file/libraries (boost includes/libries for example)

gizmocuz:
Have some issues with coloring (the red)... that is why a separate repo/system could be handy

image

gizmocuz:
tables:
image

flatsiedatsie:
Feyn and K-tron are the themes I tried to create a new pull request for. But for some reason Feynman is still in there, I don't know why. Feel free to delete that folder, as well as the images and fonts in the Feyn and K-tron folder.

I don't know anything about visual studios. I haven't added a library anywhere that I know of. As far as I know each theme is just a folder full of CSS files and one javascript file.

I will happily change the coloring if you could offer some assurance that you will allow the themes in.

gizmocuz:
I will not assure anything and vote for a separate theme system
I can also not change files in/on your pull request
Maybe you started development with an old code state and not a fresh repo pull.
As far as i am aware we also not have currently a way to store theme settings and that needs to be resolved too
And maybe some 'needed' css/js patches needs to be applied in the main domoticz repo first

ghost:
This is just my opinion, but I think out of the box, Domoticz should only include the standard theme.

Then there should be a plugin for themes, or at least a method of having an interface/page that displays all the themes on the repository, and a checkbox to download/install them to the users individual installation.

sweetpants:
I totally agree with GizMoCuz, keep domoticz itself as lightweight as possible, if users want different themes, they can create them themselves and add them to their installation. Personally I only use the default theme and use Dashticz for my mobile devices.

gizmocuz:
@flatsiedatsie, as written per email, and also the above two comments, i am closing this pull request.
Please feel free to make the needed 'basic' modifications to the default development branch (lets discuss the changes first) and then we could make a repository for themes

PR #1952, opened 2017-11-17
EventSystem fix update uservariable type

### LINKED ISSUE #1946: mqtt set uservariables (type string) don't work in beta versions
In the last stable versions setuservariables by mqtt work well

In the last beta versions setuservariables by mqtt don't work well. The uservariables (type string) returns 0 value in lua scripts when are updated by mqtt with command. Example:

`{"command": "setuservariable", "idx": 2, "value": "Off"}`

In the stable version mqtt works like a charm in this case.

Regards

jvandenbroek:
Should fix https://github.com/domoticz/domoticz/issues/1946


gizmocuz:
Thanks! Was just about to patch this ;)

PR #1953, opened 2017-11-20
OpenWbeNet add termo valves/actuators status

Spugna85:


gizmocuz:
Thanks, but can you review the files you changed in this pull request? it contains access permission changes, changes to startup images etc.
Instead of added 3 new switches to the system you can also use sSwitchGeneralSwitch and work with a different unit value for each switch type

Spugna85:
Unit value is already used how MyHOME interface type.

gizmocuz:
Thanks!

PR #1958, opened 2017-11-24
Stop limiting selectors to ten choices

dwmw2:
My Onkyo amp has about 12 inputs I can choose from. Even when I'm testing with a dummy switch, I can edit it and set values that go far past 100, but it gets silently truncated and I can never actually choose the later ones.


gizmocuz:
You have to make a pull request against the development branch

PR #1960, opened 2017-11-25
OnkyoAVTCP: Add source selection, detection via NRIQSTN

dwmw2:
Now it detects what the names of all the sources are, and lets me choose them. Also has support for 4 zones and automatically enables the ones that exist in the detected hardware.

This contains the patch from PR #1958 too.


gizmocuz:
Sorry, but uou have to commit against domoticz:development which is our native branch

PR #1961, opened 2017-11-26
Possibility to set name when adding device (CreateDevice refactor)

szczukot:


gizmocuz:
Who/What is calling this function that you need this ?

szczukot:
Many of modules call this method for create (for example EvoHome, Satel and MySensors).
Now method get devname, but dont set it (when device is created).
Now i want refactor another module (BleBox) and  and I noticed this inaccuracy.

gizmocuz:
Cant you call the functions in DomoticzHardware.h ? This will set the name and is the preferred method
Because if we change something later in the system we need to patch also all the code in those 'other' classes that directly are calling the other functions

szczukot:
Are You mean "Send***Sensor" ? 
I may use it, but i want send Selector. 
Should I add a new method?
And are You sure, that this methods is set name ? Those methods also call UpdateValue()

gizmocuz:
Hmmm yes then maybe a SendSelectorSwitch... Don't know of how many choices you need, but you can also add separate switches, specially if one is not that often used there is no reason to always give the possibility to switch this in a selector switch

szczukot:
Ok. I will create SendSelectorSwitch.

szczukot:
One question: How I may set SwitchType to STYPE_Selector amd how set selector options ?
Should I call "m_sql.safe_query("UPDATE DeviceStatus SET SwitchType=..." etc after call SendSelectorSensor ?

szczukot:
I'm not sure, is this a good direction :(
All dummy device call only UpdateValue - and this is ok for creating only.
Other modules also call only UpdateValue for create sensor.
SendXXXSensor is good for sending data - not for create.

szczukot:
May provide a method RType_CreateVirtualSensor as normal method for all modules (not only for web) ?
The modules could then create sensors using one common method.

szczukot:
I pushed next proposal.
I extracted code from Dummy to QSLHelper.
Now, Dummy use this same code (without change) and all modules may create devices using this commone method.
Next step: set all UpdateValue method as private (and change in all modules UpdateValue to CreateDevice).
Is the direction of changes is good in your opinion?

gizmocuz:
This code is/was for the dummy hardware only. If this should be make generic then make we should use a string for the 'iSensorType' like 'SoundLevel', 'Baro', 'TempHumBaro'
But i think i still do not see the whole picture of what you are trying to do ...
This function might also be placed in 'DomoticzHardware' and it could/should call the internal functions like 'SendTempSensor'
But why in your class you do not call 'SendTempSensor' it has the option to specify a name
If you miss a 'Send' sensor type, then you can add this of course

szczukot:
Yes, in next step, i want change iSensorType to Type and SubType (in html for dummy also)
I thought to put it in DomoticzHardware, but in this version every hardware would duplicate a lot of code. If You think, that DomoticzHardware is better, i move this code to it.
If we want create simply sensor like Temp etc, we can call SendTempSensor. But when we want create complicated sensor (like Dimmer etc) we call many method and many combination (all modules own):
1) call UpdateValue and next call update sql (for example https://github.com/domoticz/domoticz/blob/development/hardware/OpenWebNetTCP.cpp#L519)
2) call own sql insert (https://github.com/domoticz/domoticz/blob/development/hardware/Kodi.cpp#L1088)
3) send RXMessage and call own sql (https://github.com/domoticz/domoticz/blob/development/hardware/ZWaveBase.cpp#L237)
4) call UpdateValue and next call update sql (https://github.com/domoticz/domoticz/blob/development/hardware/EvohomeWeb.cpp#L678)
and other strange and different combination.
All modules try using own creating, because not exist common helper methods.

szczukot:
In last commit, I divided SensorType fo Type and SubType (without optimalisation code yet)

szczukot:
Ok. Please check this refactor. Now we have this same behavior as before. But now code is more uniwersal, and ready for create device for all modules.

szczukot:
One question: creating dummy device, is use UpdateValue. Why don't use RXMessage ? RXMesage call UpdateValue, and doing many more actons (logs, events, notification, shared server).

gizmocuz:
This is only for the first time it creates it? Then it's ok

szczukot:
Ok. Then this refactor is finished.
Code do exactly this same as before.
But now modules can call this method to create device.

hardware/Dummy.cpp
```diff
@@ -75,8 +75,10 @@ namespace http {
 			if ((idx == "") || (ssensortype.empty()) || (ssensorname.empty()))
 				return;
 
-			bool bCreated = false;
-			int iSensorType = atoi(ssensortype.c_str());
+			unsigned int fullType = std::stoul(ssensortype, nullptr, 16);
```
gizmocuz: i think you can use atoi and cast this to a uint16_t ?
uint16_t var = (uint16_t)atoi(ssensortype.c_str());

szczukot: But in this situation, i must rewrite values for device id type. Now code get HEX xxyy where xx=type and yy=subtype.

szczukot:
This need changes in documentation:
https://www.domoticz.com/wiki/Domoticz_API/JSON_URL's#Create_a_virtual_sensor
Now sensortype=xxYY
xx = device type
YY = device subtype
For example:
sensortype=8001 mean pTypeTEMP + sTypeTEMP1

hardware/Dummy.cpp
gizmocuz: Sorry yes you are right, so a standard std::stringstream could work

uint16_t iSensorType;
std::stringstream si;
si << std::hex << ssensortype;
si >> iSensorType;

szczukot: Done

gizmocuz:
Thanks!

dannybloe:
**Can you please revert this PR immediately???** The API is now totally fckd up.
You cannot do this. Earlier you could do this:
`http://localhost:8080/json.htm?type=createvirtualsensor&idx=2&sensorname=vdSwitch&sensortype=6`

And now this:
`http://localhost:8080/json.htm?type=createvirtualsensor&idx=2&sensorname=vdSwitch&sensortype=0xF449`

It is totally wrong to break the API like this. Please add another parameter if you want to do subtyping but not like this. 

All integration tests for dzVents now fail due to this.

jvandenbroek:
Agree with @dannybloe , you shouldn't break existing functionality if not necessary. Better use an optional subtype param and default back to previous behavior if not set.

szczukot:
@dannybloe  Now correct is 
http://localhost:8080/json.htm?type=createvirtualsensor&idx=2&sensorname=vdSwitch&sensortype=F449
instead 
http://localhost:8080/json.htm?type=createvirtualsensor&idx=2&sensorname=vdSwitch&sensortype=0xF449
Before API is wrong too. Because whats means for developer and users "sensortype=6" ? Nothing. Becuase this call set (for example) sensor type to 15 and subtype to 7.
Maybe better add another parameter to json (subtype), but in this case all dzvents tests is not correctly too.  
I understand that tests are fail. If test is failed, should be corrected.

@jvandenbroek - this is good idea i think. I can add this behavior.

dannybloe:
I do not agree. The integration tests for dzVents create _all_ dummy devices available in Domoticz with many subtypes and they are were created correctly before this 'fix'. I never sent any hex value (as you can see here [local VIRTUAL_DEVICES](https://github.com/domoticz/domoticz/blob/development/dzVents/runtime/integration-tests/testIntegration.lua).
After your PR this behavior breaks. **You just cannot change the API like this!!** If you want to change it then create a new endpoint 'createvituralsensor2' or something like that.

szczukot:
I created fix for old bahavior. I creating pull request now.

szczukot:
https://github.com/domoticz/domoticz/pull/1987

Now old method works (single virtual value mapped to type and subtype)

szczukot:
@gizmocuz please add to documentation new method:
http://localhost:8080/json.htm?type=createvirtualsensor&idx=2&sensorname=vdSwitch **&devicetype=243&devicesubtype=1**

dannybloe:
So no hex values? Just decimals?

dannybloe:
And it's not just changing the documentation. You have to document it as per which version of Domoticz you intend to change it. Not everybody is using the latest beta. That's why it is always better to create a new endpoint if you change the API so you don't break it.

szczukot:
Hmmm. You're right. I will create new API name for new opportunities

szczukot:
Done: https://github.com/domoticz/domoticz/pull/1987

gizmocuz:
@szczukot, you should be able to update wiki pages ?

szczukot:
done

mcmellow:
After updating to v3.8795 I face problems with creating virtual devices. Does not work via the web interface and it does not work with a JSON command like

http://192.168.1.xxx.xxxx/json.htm?type=createvirtualsensor&idx=yy&sensorname=Lux&sensortype=F601

yy is the indx of the dummy hardware

szczukot:
Please check API https://www.domoticz.com/wiki/Domoticz_API/JSON_URL's#Create_a_virtual_sensor

mcmellow:
szczukot, thanks for the fast reply. 

The JSON command using the number 246 for Lux  works OK. The virtual device is created. It works also with number 2 for Percentage. However, doing the same trick for a Custom Sensor not with 1004 or F31F or 0xF31F does not work at all.

Using the web interface and clicking on the button "Create Virtual Sensors" result also in an error. For a Lux as well as a Custom Sensor. 

Any more tricks or work arrounds?

szczukot:
Please show me which JSON doesn't work.
I tested JSON and "Create virtual Sensor" and I don't see problem.

mcmellow:
http://IPADRESS:8080/json.htm?type=createvirtualsensor&idx=5&sensorname=Test&sensortype=1004

This command results in {   "status" : "ERR"  }

szczukot:
I checked.
For "Custom Sensor" You must add one parameter "sensoroptions" - axis label
Try:
http://IPADRESS:8080/json.htm?type=createvirtualsensor&idx=5&sensorname=Test&sensortype=1004&sensoroptions=1;VOLT

mcmellow:
szczukot,
thanks for your support. It works OK with the parameter sensoroptions. 

Last question. Where can I find a list with virtual hardwaretype numbers, virtual sensortype numbers and sensoroption numbers. Search already a lot in the wiki etc, but I haven't found it yet.

McMelloW

szczukot:
Now exist two methods
1) Old https://www.domoticz.com/wiki/Domoticz_API/JSON_URL's#Create_a_virtual_sensor
for a small group of sensor like this 
https://github.com/domoticz/domoticz/blob/development/hardware/Dummy.cpp#L70
2) New https://www.domoticz.com/wiki/Domoticz_API/JSON_URL's#Create_a_device
when You can add probably all possible combination type and subtype

mcmellow:
Thanks very much

PR #1962, opened 2017-11-26
Fix for refresh hardware (LMS) content param - port,   and update (MySensors) hardware  

Eorciu:
1. I commented line which is responsible for no port refresh in hardware content params "Logitech Media Server".

image

2. Fix error when update or refresh some type of hardware (for example "Mysensor Gataway for Lan", becouse "indexof" is not function, correct name is "indexOf"

### DIFF
--- a/www/app/HardwareController.js
+++ b/www/app/HardwareController.js
@@ -5609,7 +5609,7 @@ define(['app'], function (app) {
 				$("#hardwarecontent #divserial").hide();
 				$("#hardwarecontent #divremote").show();
 				$("#hardwarecontent #divlogin").show();
-				$("#hardwarecontent #hardwareparamsremote #tcpport").val(9000);
+//				$("#hardwarecontent #hardwareparamsremote #tcpport").val(9000);
 			}
 			else if (text.indexOf("HEOS by DENON") >= 0) {
 				$("#hardwarecontent #divserial").hide();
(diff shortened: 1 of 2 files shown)


www/app/HardwareController.js
```diff
@@ -5609,7 +5609,7 @@ define(['app'], function (app) {
 				$("#hardwarecontent #divserial").hide();
 				$("#hardwarecontent #divremote").show();
 				$("#hardwarecontent #divlogin").show();
-				$("#hardwarecontent #hardwareparamsremote #tcpport").val(9000);
+//				$("#hardwarecontent #hardwareparamsremote #tcpport").val(9000);
```
gizmocuz, line 5612: I think this line should be

$("#hardwarecontent #divremote #tcpport").val(9000);

could you check this ?

Eorciu: I double checked , unfortunately, Your line does not work.

gizmocuz: Strange as this is how it is used on the other location in this file. It should not crash
Your patch just does not set the value anymore...

gizmocuz: I will merge this patch and have a closer look

gizmocuz:
Thanks!

www/app/HardwareController.js
gizmocuz: Checked and tested, and it works like it should
https://github.com/domoticz/domoticz/commit/1c7411dc561901d9ca430254e8db4c6be731fefb

PR #1963, opened 2017-11-27
Allow scripts to set Evohome setpoints

dwmw2:
... and clean up the polymorphism in MainWorker:SetSetPoint() somewhat. The
three-argument version of SetSetPoint() only worked for Evohome devices,
while the one-argument version worked for everything, by calling through
to the Evohome version where necessary. I'm not convinced I would have
given those functions the same name, but OK. If the intent is that this
interface will be used to handle Domoticz-based scheduling for currently
"dumb" devices at some point in the future, that makes some sense.

Fix up the three-argument version to call through to SetSetPointInt() for
devices that it doesn't support, and then change the script invocations
so that they can handle up to three arguments.

Also sift the string matching on PermanentOverride/TemporaryOverride into
SetSetPoint() from the web server code, to avoid duplicating it.

Now with a little bit of scripting and a virtual switch, I can say
'Alexa, turn on bathtime" and ths kids' radiators all turn up to 23°C
for the next hour.

### DIFF
--- a/main/SQLHelper.h
+++ b/main/SQLHelper.h
@@ -65,6 +65,7 @@ struct _tTaskItem
 	int _nValue;
 	std::string _sValue;
 	std::string _command;
+	std::string _sUntil;
 	unsigned char _level;
 	int _Hue;
 	std::string _relatedEvent;
@@ -226,13 +227,15 @@ struct _tTaskItem
 			getclock(&tItem._DelayTimeBegin);
 		return tItem;
 	}
-	static _tTaskItem SetSetPoint(const float DelayTime, const uint64_t idx, const std::string &varvalue)
+	static _tTaskItem SetSetPoint(const float DelayTime, const uint64_t idx, const std::string &varvalue, const std::string &mode = std::string(), const std::string &until = std::string())
 	{
 		_tTaskItem tItem;
 		tItem._ItemType = TITEM_SET_SETPOINT;
 		tItem._DelayTime = DelayTime;
 		tItem._idx = idx;
 		tItem._sValue = varvalue;
+		tItem._command = mode;
+		tItem._sUntil = until;
 		if (DelayTime)
 			getclock(&tItem._DelayTimeBegin);
 		return tItem;
--- a/main/mainworker.cpp
+++ b/main/mainworker.cpp
@@ -11815,12 +11815,12 @@ bool MainWorker::SwitchLight(const uint64_t idx, const std::string &switchcmd, c
 		return SwitchLightInt(sd, switchcmd, level, hue, false);
 }
 
-bool MainWorker::SetSetPoint(const std::string &idx, const float TempValue, const int newMode, const std::string &until)
+bool MainWorker::SetSetPoint(const std::string &idx, const float TempValue, const std::string &newMode, const std::string &until)
 {
 	//Get Device details
 	std::vector<std::vector<std::string> > result;
 	result = m_sql.safe_query(
-		"SELECT HardwareID, DeviceID,Unit,Type,SubType,SwitchType,StrParam1 FROM DeviceStatus WHERE (ID == '%q')",
+		"SELECT HardwareID, DeviceID,Unit,Type,SubType,SwitchType,StrParam1,ID FROM DeviceStatus WHERE (ID == '%q')",
 		idx.c_str());
 	if (result.size() < 1)
 		return false;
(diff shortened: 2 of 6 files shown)


main/SQLHelper.h
```diff
@@ -226,13 +228,15 @@ struct _tTaskItem
...
 	}
-	static _tTaskItem SetSetPoint(const float DelayTime, const uint64_t idx, const std::string &varvalue)
+	static _tTaskItem SetSetPoint(const float DelayTime, const uint64_t idx, const std::string &varvalue, const std::string &mode = std::string(), const std::string &until = std::string())
 	{
 		_tTaskItem tItem;
 		tItem._ItemType = TITEM_SET_SETPOINT;
 		tItem._DelayTime = DelayTime;
 		tItem._idx = idx;
 		tItem._sValue = varvalue;
+		tItem._sMode = mode;
```
gizmocuz: Your solution is OK. But maybe this could have been set in the sValue delimited with a semicolon ( ; )
This because in the feature i can see 100 new variables being added here...
Is it possible to change this ?

main/mainworker.cpp
```diff
@@ -11815,12 +11815,12 @@ bool MainWorker::SwitchLight(const uint64_t idx, const std::string &switchcmd, c
...
 }
 
-bool MainWorker::SetSetPoint(const std::string &idx, const float TempValue, const int newMode, const std::string &until)
+bool MainWorker::SetSetPoint(const std::string &idx, const float TempValue, const std::string &newMode, const std::string &until)
 {
 	//Get Device details
 	std::vector<std::vector<std::string> > result;
 	result = m_sql.safe_query(
-		"SELECT HardwareID, DeviceID,Unit,Type,SubType,SwitchType,StrParam1 FROM DeviceStatus WHERE (ID == '%q')",
+		"SELECT HardwareID, DeviceID,Unit,Type,SubType,SwitchType,StrParam1,ID FROM DeviceStatus WHERE (ID == '%q')",
```
gizmocuz, line 11823: I think something is duplicate here.. selecting the ID where ID ==

dwmw2: That's done deliberately so that we can pass the result directly to `SetSetPointInt()` without having to go via the one-arg version of `SetSetPoint()`and do the query again.

Now if you look *really* hard you'll probably see that the only time `SetSetPointInt()` actually looks at `sd[7]` it's the code path that calls back into the three-arg Evohome `SetSetPoint()` function, so you could argue that I didn't need to add it. But then you get to go and sit on the naughty step. Ick :)

main/SQLHelper.h
dwmw2: I'm reluctant to break the string apart when it comes in from the script, then put it back together again, then break it apart again. I suppose we could just pass the whole "temp#mode#until" string through and then break it apart in `CSQLHelpers::Do_Work()` but then we lose out on the error handling when the arguments are bogus. 

In the general case though... this is what unions are for ;)

main/mainworker.cpp
dwmw2: Looking back at that, I missed the addition of HTYPE_EVOHOME_TCP when switching from master to development branches. Fixed in the push below.

main/SQLHelper.h
dwmw2: I could also just abuse the `_command` field (like `TITEM_SEND_NOTIFICATION` does) and the `_relatedEvent` field...

dwmw2: Hm, not `_relatedEvent` as that gets interpreted generically, but `_command` seems to be fair game. So now I'm adding only *one* strnig. Is that an acceptable compromise?

main/mainworker.cpp
gizmocuz: Good catch! +1 for the ID

dannybloe:
For which evohome implementation is this relevant? The web version or the HGI80 version?
I'd like to know so I can adjust the adapters for dzVents so peeps can start using this method.

dwmw2:
It should work for for both. Anywhere that this works:
```
curl http:/$DOMOTICZ/json.htm?type=setused&idx=25&setpoint=11&mode=TemporaryOverride&until=2017-11-29T01:29&used=true
```
... this should now also work (in a lua script, for example):

```
if (devicechanged['bathtime_switch'] == 'On') then
  commandArray['SetSetPoint:25'] = "15#TemporaryOverride#2017-11-27T06:03"
end
```

I've actually been testing with the web version on a test box, but once I get around to rebuilding the version that runs in production on my OpenWRT router (and uses HGI80), I see no reason it shouldn't work there.

PR #1965, opened 2017-11-27
Fix date formats

dwmw2:
Use standard ISO date format YYYY-MM-DD everywhere, never the parochial local dialects.

The first commit should be fairly straightforward. The second actual changes the timer results returned on the wire, which is why I've pulled it out into a separate commit for specific consideration.

But as I note in that commit comment, it should be fine. What kind of person would see "dd/mm/yy" or "mm/dd/yy" on the wire and *not* assume that it **is** going to change?

### DIFF
--- a/main/Scheduler.cpp
+++ b/main/Scheduler.cpp
@@ -971,7 +971,7 @@ namespace http {
 						int Year = atoi(sdate.substr(0, 4).c_str());
 						int Month = atoi(sdate.substr(5, 2).c_str());
 						int Day = atoi(sdate.substr(8, 2).c_str());
-						sprintf(szTmp, "%02d-%02d-%04d", Month, Day, Year);
+						sprintf(szTmp, "%04d-%02d-%02d", Year, Month, Day);
 						sdate = szTmp;
 					}
 					else
@@ -1043,7 +1043,7 @@ namespace http {
 						int Year = atoi(sdate.substr(0, 4).c_str());
 						int Month = atoi(sdate.substr(5, 2).c_str());
 						int Day = atoi(sdate.substr(8, 2).c_str());
-						sprintf(szTmp, "%02d-%02d-%04d", Month, Day, Year);
+						sprintf(szTmp, "%04d-%02d-%02d", Year, Month, Day);
 						sdate = szTmp;
 					}
 					else
@@ -1430,7 +1430,7 @@ namespace http {
 						int Year = atoi(sdate.substr(0, 4).c_str());
 						int Month = atoi(sdate.substr(5, 2).c_str());
 						int Day = atoi(sdate.substr(8, 2).c_str());
-						sprintf(szTmp, "%02d-%02d-%04d", Month, Day, Year);
+						sprintf(szTmp, "%04d-%02d-%02d", Year, Month, Day);
 						sdate = szTmp;
 					}
 					else
@@ -1772,7 +1772,7 @@ namespace http {
 						int Year = atoi(sdate.substr(0, 4).c_str());
 						int Month = atoi(sdate.substr(5, 2).c_str());
 						int Day = atoi(sdate.substr(8, 2).c_str());
-						sprintf(szTmp, "%02d-%02d-%04d", Month, Day, Year);
+						sprintf(szTmp, "%04d-%02d-%02d", Year, Month, Day);
 						sdate = szTmp;
 					}
 					else
--- a/main/Scheduler.cpp
+++ b/main/Scheduler.cpp
@@ -1141,9 +1141,16 @@ namespace http {
 			{
 				if (sdate.size() == 10)
 				{
-					Month = atoi(sdate.substr(0, 2).c_str());
-					Day = atoi(sdate.substr(3, 2).c_str());
-					Year = atoi(sdate.substr(6, 4).c_str());
+					if (sdate.c_str()[2] == '/' && sdate.c_str()[5] == '/') {
+					/* Legacy US-local date format */
+						Month = atoi(sdate.substr(0, 2).c_str());
+						Day = atoi(sdate.substr(3, 2).c_str());
+						Year = atoi(sdate.substr(6, 4).c_str());
+					} else {
+						Year = atoi(sdate.substr(0, 4).c_str());
+						Month = atoi(sdate.substr(5, 2).c_str());
+						Day = atoi(sdate.substr(8, 2).c_str());
+					}
 				}
 			}
 			else if (iTimerType == TTYPE_MONTHLY)
@@ -1246,9 +1253,16 @@ namespace http {
 			{
 				if (sdate.size() == 10)
 				{
-					Month = atoi(sdate.substr(0, 2).c_str());
-					Day = atoi(sdate.substr(3, 2).c_str());
-					Year = atoi(sdate.substr(6, 4).c_str());
+					if (sdate.c_str()[2] == '/' && sdate.c_str()[5] == '/') {
+					/* Legacy US-local date format */
+						Month = atoi(sdate.substr(0, 2).c_str());
+						Day = atoi(sdate.substr(3, 2).c_str());
+						Year = atoi(sdate.substr(6, 4).c_str());
+					} else {
+						Year = atoi(sdate.substr(0, 4).c_str());
+						Month = atoi(sdate.substr(5, 2).c_str());
+						Day = atoi(sdate.substr(8, 2).c_str());
+					}
 				}
 			}
 			else if (iTimerType == TTYPE_MONTHLY)
@@ -1478,9 +1492,16 @@ namespace http {
 			{
 				if (sdate.size() == 10)
 				{
-					Month = atoi(sdate.substr(0, 2).c_str());
-					Day = atoi(sdate.substr(3, 2).c_str());
-					Year = atoi(sdate.substr(6, 4).c_str());
+					if (sdate.c_str()[2] == '/' && sdate.c_str()[5] == '/') {
+					/* Legacy US-local date format */
+						Month = atoi(sdate.substr(0, 2).c_str());
+						Day = atoi(sdate.substr(3, 2).c_str());
+						Year = atoi(sdate.substr(6, 4).c_str());
+					} else {
+						Year = atoi(sdate.substr(0, 4).c_str());
+						Month = atoi(sdate.substr(5, 2).c_str());
+						Day = atoi(sdate.substr(8, 2).c_str());
+					}
 				}
 			}
 			else if (iTimerType == TTYPE_MONTHLY)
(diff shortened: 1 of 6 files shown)


gizmocuz:
I would.
If you want to create a patch then please make it so it is in the user's locale always
Here starting with the year makes no sense at all

dwmw2:
I strongly believe that making this stuff depend on the user's locale is just gratuitous complexity. The ISO standard YYYY-MM-DD is simple and works everywhere — and is what Domoticz uses in *most* cases already. **Before** my patch to Scheduler.cpp:
```
$ grep Month.*substr Scheduler.cpp  | grep -c '5, 2'
7
$ grep Month.*substr Scheduler.cpp  | grep -c '0, 2'
6
```
This PR is just about being consistent.

travellingkiwi:
Although I favour YYYY-MM-DD above all else, I dislike arguments like this. It should take the date from the users locale. Too often I see websites trying to force me to read MM/DD/YY or similar. 

H

dwmw2:
At the moment, Domoticz *is* forcing people to read the mm/dd/yyyy format that is only valid in some parts of the world — at least in some parts of Domoticz. The only format that's clear and unambiguous is the ISO standard YYYY-MM-DD which works everywhere in the world. And which Domoticz *does* use in the majority of places, thankfully.

I just wanted Domoticz to consistently do the right thing, instead of doing the right thing *most* of the time but not all of the time.

There isn't really any need for localisation of date formats. Everyone understands the ISO standard.  But if you want to implement that as an additional step, by all means go ahead. *Please* make it default to the standard and then let people change it to local dialects if they really must, though.

dwmw2:
@gizmocuz what is your preferred approach here? Should I resubmit just the first patch making the UI consistent, without the second one changing the protocol it outputs on the wire?

I've been looking into localisation, and browsers don't really offer a full set of locale options. There's an `X-Accept-Language` header in the request, but that doesn't really map to the date format very well. And I still think it's just generally a bad idea when we can use a form that *isn't* locale-specific, and is recognised everywhere in the world. And is what Domoticz uses in the majority of cases already, except for the stragglers that my patch fixes.

gizmocuz:
The "preferred" approach would be to have the User interface (Web Gui) en the log in the users native locale.
I think there is a javascript way to find out the current browser locale.
And of course we can use the locale time functions to have the date/time in the log presented in the locale region.
The API/JSON calls should stay as they are, and are documented.

As you can see here, DMY is the most used. (even in Germany)

https://en.wikipedia.org/wiki/Date_format_by_country

travellingkiwi:
How about…

1. Drop a system-wide default into the settings. (Include static options (YYYYMMDD, DDMMYYYY and MMDDYYYY, plus look at the system settings)
2. Allow over-ride of system-wide by user settings
3. Allow over-ride of user settings at web interface by browser locale if it’s specified

1. allows logs to look correct. Although if you’re using syslog that goes away anyway. Timezone or offset should always be logged if it’s done in domoticz itself (i.e. non  syslog)
2. If a web user doesn’t like the system (backend) settings they can always set one statically for their user
3. Finally the browser settings should be used for displaying to the user allowing it to over-ride everything else.

H

gizmocuz:
Way to complicated, just use the users local/region setting

gizmocuz:
Keep It Simple !

dwmw2:
https://en.wikipedia.org/wiki/Date_format_by_country is confusing.

The ISO standard YYYY-MM-DD is applicable basically everywhere. Some countries have gone even further and published it as part of their set of national standards, as listed at https://en.wikipedia.org/wiki/ISO_8601#Related_standards — but even *those* countries don't appear in the Date_format_by_country table with a 'yes' by the YMD column.

The way to Keep It Simple is just to use YYYY-MM-DD everywhere. That's what we are already doing in *most* of Domoticz. Adding locale stuff really *isn't* keeping it simple.

travellingkiwi:
I admire the concept, but if it were to be simple, it wouldn’t be a monolith… Or include its own web server. Adding that functionality (e.g. web server) generally means adding the functionality that something separate and already existing (e.g. nginx) already has.

dwmw2:
OK... let me stop just repeating myself grumpily, and try to explain why I'm really dubious about doing it based on locales — why I don't even want to see it in the local dialect of *my* country either.

It's very similar to the problem with character sets. Information moves from one part of the system to another *(UI to wire protocol to database, to scripts, to other wire protocols, etc.)*. To get this right in a system of many moving parts, you have to get two things right, *everywhere*:
- You have to label the information with the character set being used.
- You have to convert, potentially at any stage in the system where information is being passed from on place to another which might need a different character set.

With character sets, we have come to the realisation that this doesn't work. People *forget* to convert in the places they need to, and it's worse than that; they can't even do the labelling consistently. The labels fall off, and you're left with a string like `6e 61 ef 76 65` and no idea how to interpret it. You can make a *guess*... but that just means that you have a system which mostly works OK but then screws up unexpectedly *sometimes*. Which is horrid.

This is how we ended up with abominations like the one described at http://david.woodhou.se/draft-woodhouse-cert-best-practice.html#rfc.section.7.1

Seriously go and read §7.1 there and learn about how OpenSSL screwed it up, and then made it worse instead of fixing it. And then you'll realise why I feel so strongly about this kind of thing :)

So... all this horrid mess is why we have settled on Unicode/UTF-8 for data interchange. And then it goes away. We don't have to worry about conversion, we don't have to worry about labelling. In any sane system, all strings are stored as Unicode. IF you have to convert occasionally when you're talking to some external legacy API or system, that's easy enough to handle as a special case — you still don't have to worry about tracking "what charset is this data in?" throughout the whole of your own systems.

Now, date formats have fairly much precisely the same characteristics. They get moved around, and we'd need to convert them consistently to make them work. And much like charsets, we *can* often guess the right interpretation if we have to... except when we can't, and we get hard-to-reproduce bugs (that only happen on the first 12 days of any month, etc.).

That's why I feel very strongly that we should standardise on YYYY-MM-DD absolutely everywhere and not invite the chaos of trying to handle labelling and conversions.

travellingkiwi:
Hi.

I note your grumpiness. And appreciate it. trust me. I deal with it every day (No, not your grumpiness. My own when it comes to exactly this situation. You should try working with oracle databases some time and their arcane DATE and TIMESTAMP formats). 

However, that’s just it. I wasn’t suggesting internal representation of dates. If I had my way we’d use Unix epochs always. Until it’s time to display (Or text log). And then display (Or write) in the appropriate locale format.

Don’t get me wrong, if we can’t have it localised, I’ll settle for YYYY-MM-DD. Because at the end of the day self-sorting dates are great. They’re hard to mis-understand, and most of us read them every day anyway. However in terms of user friendliness I’d urge for proper localisation. But that’s urge, not demand.

H

dwmw2:
Like many people, I actually hate seeing localised dates even in my own locale format. Because I *don't* only see dates in that format.

There are so many pieces of software out there that get it wrong, and so many *people* who get it wrong, with the result that *every* time I see a date like 11/12/2017 I have to stop and work it out. I even get US-form dates from Europeans while the US is asleep and I *ought* to be safe! It's all just horrid. The only date that I can see and *doesn't* cause a pipeline stall, is the international standard.

That's why showing dates in the "local" format is very much **not** user-friendly for many users. Show dates in a clear and unambiguous form which is meaningful to everyone and which they can cut and paste, instead.

Now, if you can go and fix the world so that locale-specific dates never escape their country of origin, then sure, eventually that problem might go away. But I think that means they can only *ever* by shown in something like Snapchat, and never anywhere that someone could cut and paste from. Perhaps then, we can get used to seeing 12/11/2017 dates again and knowing what they mean immediately.

dwmw2:
Note that text logs are often parsed by machine too, and now any scripts and tool which do that would have to cope. Seriously, unless it's *just* in Snapchat you can't say that it's only for human eyes and will never cause confusion elsewhere.

travellingkiwi:
I agree with you. With one exception… By localised I mean local to how the user wants it… Not the ‘google’ way that localises because of where you are...

i.e. you should be able as a user to tell Domoticz (or any web interface) put my dates in this damn format and don’t mess with me… And yes, I’d probably put them in YYYY-MM-DD as well. But as a developer I’m well aware that some people don’t like that. Hence my urging for able to be configured.


H

dwmw2:
If it was easy enough to do, then I would probably have done that already — despite the fact that I think it's a bad idea. Sometimes, what the user wants is not the same as what the user should have. All the time in this discussion I just keep remembering the old Red Hat installer locale option for "redneck" :)

But the UI side of Domoticz is all fairly alien to me, and it took me long enough to work out how to do what I've already done.

Remember, Domoticz is already using YYYY-MM-DD more often than not. It was only the setpoint editor that continually offended me before I went looking. So making it *consistent* seems like a decent first step anyway, and if someone really wants to come along and make it configurable, good luck to them. But please make it *default* to the standard, and let people change it only if the really insist.

dwmw2:
Just to clarify the point about consistency... here's a screenshot of my Domoticz dashboard right now.

I count 26 dates in ISO standard format, and precisely none in either my local dd/mm/yy or the US mm/dd/yy format. This is why I'm somewhat reluctant to accept an argument that users can't cope with standard dates and *have* to be mollycoddled with their own local format. We just aren't doing that *anyway*. We *do* use the standard in most places, and it's only a few oddities that I'm trying to fix.

domoticz

travellingkiwi:
Agree 100%

gizmocuz:
@travellingkiwi, no i do not agree, it is simpler with its own webserver, and that really besides this topic

gizmocuz:
@dwmw2 again, i agree for 1 locale format, and not on an ISO format.
Keep it simple for the end user... just use the systems locale/region, there are even c++ std/time functions for it

dwmw2:
@gizmocuz if it's for the end user, it's JS functions you want not C++. Those do exist too, but I have explained why it never really *is* just for the end user. Dates go out and are passed back in, and you have to be careful to do all the conversions in the right places or you introduce subtle sporadic bugs.

I do not believe it's possible to get that right, and have it *stay* right in the long term.

Have you *actually* had users who have really wanted a local format instead of the standard one that Domoticz currently shows in most places, like in the above screenshot?

dwmw2:
https://css-tricks.com/time-element/ looks interesting. Perhaps the answer is to do it like that — times and dates are put into the HTML thus:
```
    <time datetime="2017-11-18">Two days after Mummy's birthday</time>
```
We'd have JS which is responsible for generating the personalised local content in any `<time>` element, but the actual information would still be present for the cases when it needs to be sent back over the wire.

I still think it's pointless to show dates in a silly personalised form when there's a standard that everyone can understand and which Domoticz is showing most of the time anyway. But at least if we do it like this, we can separate the cosmetic display from the actual data, and avoid introducing bugs.

gizmocuz:
@dwmw2 i was also talking about the native log output of the application, and the lastseen variable that is used on the web interface, but that can be changed to a time_t value so the javascript code can translate this to the browsers locale time as we can do with highcharts

I have had no users complaining about date time yet

dwmw2:
So let's recap:

- In the vast majority of dates it shows to the user, Domoticz uses the international standard YYYY-MM-DD which is unambiguously understood throughout the whole world.
- No users have ever complained about this or asked for it to be changed to one of their other local date formats *(I say "other" because ISO8601 is ratified by local bodies  fairly much everywhere too, and some places have even more variants than that)*.
- There are a few places where Domoticz is inconsistent with itself and instead uses a local format which *is* ambiguous and, for many users, just plain wrong.
- I offered a patch to fix those few places, and make them consistent with the rest of the Domoticz UI.
- You don't want that patch, and you'd rather Domoticz remain inconsistent unless someone is going to implement full locale-dependent date views that no user has ever asked for.
- I have explained why converting to different forms opens up a horrible can of worms and will introduce bugs, just as it does with character sets, and you want it anyway.

Is that about right?

PR #1971, opened 2017-11-28
OnkyoAVTCP: Various improvements (raw EISCP command, net playback status, non-MRIQSTN support, compatibility fix)

dwmw2:
We don't probe for these at startup, and maybe we should. But it's fairly
forthcoming with spontaneous updates anyway, so this is a good start.

textinfo

Also support sending raw commands such as for radio presets:
```
http://${DOMOTICZ}/json.htm?type=command&param=onkyoeiscpcommand&idx=${IDX}&action=NPR04
```

### DIFF
--- a/hardware/OnkyoAVTCP.cpp
+++ b/hardware/OnkyoAVTCP.cpp
@@ -16,6 +16,74 @@
 
 #define RETRY_DELAY 30
 
+
+struct eiscp_packet {
+	uint32_t magic;
+	uint32_t hdr_size;
+	uint32_t data_size;
+	uint8_t version;
+	uint8_t reserved[3];
+	char start;
+	char dest;
+	char message[2];
+};
+
+// Do not change except to append to these, or you break compatibility.
+enum DevNr {
+	ID_MVL = 0,
+	ID_ZVL,
+	ID_VL3,
+	ID_VL4,
+	ID_PWR,
+	ID_ZPW,
+	ID_PW3,
+	ID_PW4,
+	ID_SLI,
+	ID_SLZ,
+	ID_SL3,
+	ID_SL4,
+	ID_HDO,
+};
+
+#define ARRAY_SIZE(x) (sizeof(x) / sizeof(x[0]))
+static struct {
+	const char *iscpCmd;
+	const char *iscpMute;
+	const char *name;
+	int switchType;
+	int subtype;
+	int customImage;
+	const char *options;
+} switch_types[] = {
+	{ "MVL", "AMT", "Master volume", STYPE_Dimmer, sSwitchGeneralSwitch, 8, NULL },
+	{ "ZVL", "ZMT", "Zone 2 volume", STYPE_Dimmer, sSwitchGeneralSwitch, 8, NULL },
+	{ "VL3", "MT3", "Zone 3 volume", STYPE_Dimmer, sSwitchGeneralSwitch, 8, NULL },
+	{ "VL4", "MT4", "Zone 4 volume", STYPE_Dimmer, sSwitchGeneralSwitch, 8, NULL },
+	{ "PWR", NULL, "Master power", STYPE_OnOff, sSwitchGeneralSwitch, 5, NULL },
+	{ "ZPW", NULL, "Zone 2 power", STYPE_OnOff, sSwitchGeneralSwitch, 5, NULL },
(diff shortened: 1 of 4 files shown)


dwmw2:
I updated the wiki at https://www.domoticz.com/wiki/Onkyo to describe the new functionality. I think it now obsoletes all the scripts described there.

gizmocuz:
Thanks, maybe combine a few?

dwmw2:
Combine the text sensors? I wondered about that and figured it might be better to expose the raw data then let people combine as they wish in scripting. Much more versatile that way.

Unless I make a new type which has all fields separately, but they *are* separate information from the server so that gets complex too

gizmocuz:
I meant make 1 text sensor with for example:
Title Name
Artist Name

dwmw2:
Yeah, I did think about that but had decided against it. All these fields come in asynchronously as completely separate updates. So if we put them into a single text sensor, we'd end up having to read the current value from the database, split it, change the one that we've just received, join it and then write it back.

That's a lot more complicated than the existing simple call to `SendTextSensor()`.

And I suspect the consumers of the data will also want individual fields. I haven't looked at doing a UI yet, like the Kodi remote and other things we have in Domoticz — but anything like that will want to display the fields separately, so they'd just be having to split and then process every one, when just a single one changes.

Scripts are the same — if users want to script things like turning the amp to a certain listening mode when listening to talk radio vs. music, for example, that ends up being based on only *one* of the text fields.

So it looked like it was extra work to put them together, and extra work to split them apart for display... and that's why I'd decided against it. Was that wrong?

gizmocuz:
Personally i do not see what this has to do with home automation... there are other dashboards applications for this

dwmw2:
All the volume and selector and other switches are part of the home automation. I can have scenes in Domoticz, triggered by Alexa, for turning the TV on for the kids, for turning the radio on when they leave the house, etc. 

That's not the part you're dubious about, I hope?

As for the text sensors... they're information that some of the scenes and scripting may want to react to. For example automatically changing the listening mode according to the type of radio station that's being listened to, etc.

gizmocuz:
Maybe we should enhance the text sensor a bit... like when needed to, scroll the text left and right... like an LCD of a receiver
Bdw of course using Domoticz to let Alexa trigger the television or radio is fine.. just all the track info on 5 separate text sensors not....

dwmw2:
Yeah, I'm going to look at the sensors' UIs at some point. Volume in particular is scary at the moment because a stray touch on the phone when you're actually trying to scroll the page, can set it to 100 and deafen you. I'm thinking of making a different sensor type for volume, which has +/- buttons and doesn't have a UI for **"hurt me!"** :)

Any UI for this would still end up breaking the strings apart though, so squashing them into one just seems like additional complexity on both producer (with read/modify/write) and consumer (with updates for all fields and having to work out which ones changed or just update UI elements when they didn't) side.

Am I missing a reason why we need to keep the overall *number* of devices to a minimum?

dwmw2:
After feedback from users, rolled in a couple more fixes.

gizmocuz:
Thanks!

ghost:
Is there any roadmap or timeframe for when the input selectors will be released into the stable version, and what's stopping them from being in the current stable release?

So far, all I have for my TX-NR414 is Master Power, Master Volume and Zone 2 Power.

dwmw2:
Please test the beta. If you report problems only after the next stable release happens, you get to wait even longer for the fixes you need :)

Komagain:
Hello,
I am using the latest beta ( 4.10301), the only devices shown for my NR616 are Master Power and Master Volume,
How can I have the others?

dwmw2:
Change them using the normal remote and they'll get "discovered".

Komagain:
done, thank you,
image
with my NR616:

- HDMI Output shows; main, sub, Main+Sub, 01 (which is main) and 02 (which is Sub)
01 and 02 have added after the Main, Sub, Main=Sub
is there a way to fix it?

- Master selector shows: 
image
GAME/TV, TV/CD, CBL/SAT, BD/DVD, PC, 2B, FM
it misses STB/DVR, NET. and I don't know what is 2B
- Master Power is always Off
  - There is a Zone 4 selector, I am not aware of Zone 4 on this model,
  - there is only one value 2B, I don't what is means.

PR #1973, opened 2017-11-30
JS: Add myglobals.DateFormat and honour it for all datepickers

dwmw2:
As requested in #1965, this makes the date format configurable — instead of hard-coding it in various places, we now have it in `myglobals.DateFormat` so it can be tweaked.

This still doesn't yet allow you to *change* it; that will require more work on the UI side to perform conversions, as discussed. But it's a step in that direction, which can be built on in future commits. And it does fix the immediate issue that Domoticz isn't consistent.

### DIFF
--- a/main/Scheduler.cpp
+++ b/main/Scheduler.cpp
@@ -1097,7 +1097,7 @@ namespace http {
 						int Year = atoi(sdate.substr(0, 4).c_str());
 						int Month = atoi(sdate.substr(5, 2).c_str());
 						int Day = atoi(sdate.substr(8, 2).c_str());
-						sprintf(szTmp, "%02d-%02d-%04d", Month, Day, Year);
+						sprintf(szTmp, "%04d-%02d-%02d", Year, Month, Day);
 						sdate = szTmp;
 					}
 					else
@@ -1169,7 +1169,7 @@ namespace http {
 						int Year = atoi(sdate.substr(0, 4).c_str());
 						int Month = atoi(sdate.substr(5, 2).c_str());
 						int Day = atoi(sdate.substr(8, 2).c_str());
-						sprintf(szTmp, "%02d-%02d-%04d", Month, Day, Year);
+						sprintf(szTmp, "%04d-%02d-%02d", Year, Month, Day);
 						sdate = szTmp;
 					}
 					else
@@ -1267,9 +1267,9 @@ namespace http {
 			{
 				if (sdate.size() == 10)
 				{
-					Month = atoi(sdate.substr(0, 2).c_str());
-					Day = atoi(sdate.substr(3, 2).c_str());
-					Year = atoi(sdate.substr(6, 4).c_str());
+					Year = atoi(sdate.substr(0, 4).c_str());
+					Month = atoi(sdate.substr(5, 2).c_str());
+					Day = atoi(sdate.substr(8, 2).c_str());
 				}
 			}
 			else if (iTimerType == TTYPE_MONTHLY)
@@ -1372,9 +1372,9 @@ namespace http {
 			{
 				if (sdate.size() == 10)
 				{
-					Month = atoi(sdate.substr(0, 2).c_str());
-					Day = atoi(sdate.substr(3, 2).c_str());
-					Year = atoi(sdate.substr(6, 4).c_str());
+					Year = atoi(sdate.substr(0, 4).c_str());
+					Month = atoi(sdate.substr(5, 2).c_str());
+					Day = atoi(sdate.substr(8, 2).c_str());
 				}
 			}
 			else if (iTimerType == TTYPE_MONTHLY)
@@ -1542,7 +1542,7 @@ namespace http {
 						int Year = atoi(sdate.substr(0, 4).c_str());
 						int Month = atoi(sdate.substr(5, 2).c_str());
 						int Day = atoi(sdate.substr(8, 2).c_str());
-						sprintf(szTmp, "%02d-%02d-%04d", Month, Day, Year);
+						sprintf(szTmp, "%04d-%02d-%02d", Year, Month, Day);
 						sdate = szTmp;
 					}
 					else
@@ -1604,9 +1604,9 @@ namespace http {
 			{
 				if (sdate.size() == 10)
 				{
-					Month = atoi(sdate.substr(0, 2).c_str());
-					Day = atoi(sdate.substr(3, 2).c_str());
-					Year = atoi(sdate.substr(6, 4).c_str());
+					Year = atoi(sdate.substr(0, 4).c_str());
+					Month = atoi(sdate.substr(5, 2).c_str());
+					Day = atoi(sdate.substr(8, 2).c_str());
 				}
 			}
 			else if (iTimerType == TTYPE_MONTHLY)
@@ -1700,9 +1700,9 @@ namespace http {
 			{
 				if (sdate.size() == 10)
 				{
-					Month = atoi(sdate.substr(0, 2).c_str());
-					Day = atoi(sdate.substr(3, 2).c_str());
-					Year = atoi(sdate.substr(6, 4).c_str());
+					Year = atoi(sdate.substr(0, 4).c_str());
+					Month = atoi(sdate.substr(5, 2).c_str());
+					Day = atoi(sdate.substr(8, 2).c_str());
 				}
 			}
 			else if (iTimerType == TTYPE_MONTHLY)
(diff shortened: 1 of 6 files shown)


gizmocuz:
Thanks! but..

If we let the user configure the 'DateFormat' and the user wants it to be displayed in mm-dd-yy or dd-mm-yy, then we still have a problem with the domoticz internals ?
So i would propose to make 1 standard API date/time format dd-mm-yy, and for the user interface(gui) use the 'myglobals.DateFormat'
We can also opt for a time_t value in the API communication ?

dwmw2:
> Thanks! but..
> 
> If we let the user configure the 'DateFormat' and the user wants it to be displayed in mm-dd-yy or dd-mm-yy, then we still have a problem with the domoticz internals ?

Yes. This is a *step* towards being able to configure it, but it doesn't get us all the way there yet. It makes it consistent, using the format that Domoticz already uses *almost* everywhere, which is OK for everyone and isn't interpreted differently in different countries. And it introduces the framework to *allow* it to be changed later, once we've implemented the rest of what's needed for that.

But as well as being a step towards configurability, in the meantime it fixes an actual bug in the UI which leads to confusion. I was working late on Monday night, and attempted to set a temperature setpoint in my study —  and got *very* confused because this is one of the *few* places where Domoticz doesn't already use the YYYY-MM-DD standard, and told me it was setting a setpoint for the 12th of November (12/11/2017). My brain kept interpreting that as the 12th of December, and that was extending the set point for a whole day too far. Despite the fact that obviously I'm *aware* of this bug and I should intellectually be able to compensate. But it's like when you're reading the word 'RED' written in green ink, when you're asked what colour the *ink* is.... you might be able to answer it eventually but your brain is playing tricks on you while you do. That's the effect this bug has, and why I'm so keen to fix it.

> So i would propose to make 1 standard API date/time format dd-mm-yy, and for the user interface(gui) use the 'myglobals.DateFormat'

Yes, that's what I've done. The standard is the ISO8601 one. Most APIs already accepted that, and *all* now do. The ones which took a different form before, will now accept both and automatically interpret it correctly.

What remains to be done, to complete the support for configurability, is implement conversion functions on the UI side which will convert to/from that API format, for display to the user.

> We can also opt for a time_t value in the API communication ?

Theoretically, perhaps. But ISO8601 is more typically used for this. And to use time_t we'd need to implement all those conversion functions *first*, rather than having an incremental path to get there. It doesn't seem worth it.

gizmocuz:
Then i am confused about the'/* Legacy US-local date format */ code/comments in the code.
Are they still needed if the API is using 1 format ?

dwmw2:
It's an API which we expose to the world. People drive it with curl and scripts and other stuff.

The people writing such scripts have already had to deal with *"er, Domoticz uses YYYY-MM-DD almost everywhere except for *this* call, which seems to only work with DD/MM/YY"*. 

So I'm fixing things to take YYYY-MM-DD everywhere... do I make a flag day and break the existing scripts? Or do I just look for a '/' in the third and sixth characters and make it do the right thing for those existing scripts too?

If we really don't care about that concern, we can drop the conditional stuff. It's just that it seemed relatively harmless to leave it there for now, and not risk breaking users' scripts etc.

dwmw2:
(That push just rebases onto development again to fix the build failure that wasn't my fault; I haven't removed the script compatibility conditionals)

gizmocuz:
So this gets confusing... as explained above, if we let a user choose the display format and there are options for
dd-mm-yy
mm-dd-yy

it still goes wrong internally

There should be one format only and the need for the if/else shall then not be needed.

The display (gui) should be handled differently then the internal API communication as this can be configured (later) to the users preference.
That could mean that displaying in the gui 11/01/2018 should still be send via the API as 2018-01-11
Internally we should only care/handle 2018-01-11 , also from the application to the GUI via the API 2018-01-11 should be displayed as 11/01/2018

Maybe that is why a lot of API's use the time_t value

dwmw2:
> So this gets confusing... as explained above, if we let a user choose the display format and there are options for
> dd-mm-yy
> mm-dd-yy
> 
> it still goes wrong internally
> 
> There should be one format only and the need for the if/else shall then not be needed.

For the UI, yes. That is absolutely correct. The UI will only ever use YYYY-MM-DD in the API, whatever it shows to the user. Before we allow that 'myglobals.DateFormat' option to be tweaked, we'll first implement all the conversions that are needed on the way in and out.

The compatibility isn't about the UI. It's purely for users' scripts which might invoke the JSON API directly. We don't *need* to break any scripts which might currently be submitting timer updates using the dd/mm/yyyy format. But then again, do people really write such scripts to add and modify timers? Those are the only JSON API calls we're talking about here; the rest already did use the ISO standard.

So I'm happy to drop the compatibility. As I said, I figured it was harmless to preserve it, just in *case*. But I suppose you're right, the fact that it gets confusing is harm enough, I'll drop it and make it unconditionally take the standard form. Anyone who has a script which sets timers, will have to fix their script. But that's unlikely, I think.

gizmocuz:
No people do not write such scripts... or very less people do, and they just need to update.
No need for extra code...

Maybe some users are liking the current display format... that is why a complete patch with user defined date display format in the settings and correct underlying API calls would be the best

dwmw2:
> No people do not write such scripts... or very less people do, and they just need to update.
> No need for extra code...

OK, I've dropped that and re-pushed (including rebasing onto a slightly older commit from the development branch, which actually compiles).

> Maybe some users are liking the current display format...

The "current display format" for Domoticz is to use YYYY-MM-DD almost everywhere, and that hasn't changed (and I think you already said nobody has ever asked for it to change). My dashboard has 30-something instances of that, and precisely none of any other form. The few cases I'm fixing here are hidden away in setpoint and timer UIs, and we're just making them consistent with what Domoticz does everywhere else.

> that is why a complete patch with user defined date display format in the settings and correct underlying API calls would be the best

Yes. This gives you a step in that direction. I still think there are better things we can be doing, as it's error-prone to do all that conversion and there seems to be very little user demand for it. But I've changed things around to both fix the immediate bug, and also allow for that work to be continued... by someone who really wants it :)

gizmocuz:
that must be you. Happy to accept a complete patch

dwmw2:
There are two separate things here — a bug, and a feature request. 

The bug is that in a *few* cases in the Domoticz UI, it doesn't do the standard thing that it does everywhere else, and it displays dates to the user in a way which is foreign and confusing. That's really easy to fix, and is done in this pull request.

The feature request — that you say no users have actually asked for — is to be able to configure the date format so that instead of the standard form that works for everyone, which Domoticz already uses *almost* everywhere and will *actually* use everywhere after my bug-fix, users will be able to choose a local date format of their own preference.

Even though the feature request involves a bunch of complicated JavaScript work and is likely to be continually out of date and buggy, which is the experience we've learned from character set handling over the years, and even though no user has even *asked* for this feature, you aren't happy with me simply designing the bug fix to *allow* for this feature to be added, but you actually want me to go through and do all the UI work that I think is a bad idea?

gizmocuz:
Please try to keep your comments a bit shorter...

dwmw2:
> Please try to keep your comments a bit shorter...

OK...

Please let me just fix the bug, and don't make me implement the additional feature that no user has even asked for.

gizmocuz:
Okey, let me be a bit clearer.
1 This is indeed requested by you (no other user)
2 Other users favor the other format (see your original issue)
3 It is best to make a complete pull request where everybody is happy, not half work

Nobody is in a hurry to have this fixed, please take the time you need and submit a complete sollution

dwmw2:
@travellingkiwi I suspect your comments in #1965 are being misinterpreted/overinterpreted.

Domoticz already uses the standard which works for everyone, almost everywhere. I just want to fix a few places where it gets it wrong, and uses a different non-standard form which doesn't work for everyone.

Your comments are being used (see above) to ask for this additional feature which I believe is complex and unnecessary. Could I ask you to retract them, or at least say that it's OK to fix the *bug* right now and leave the feature request for a later date? Thanks.

travellingkiwi:
I agree. I think my fervour has perhaps been misinterpreted. I would always plump for YYYY-MM-DD pretty much everywhere. It’s rare I’d use anything else. My comments were intended to highlight that not everyone uses the same date formats (As we can see already from the current state).

I would urge this pull request to be merged. Because it fixes the main problem. That is the inconsistency of the date formats.

I think further changes - to make the date formats actually configurable should be done separately. And possibly limited to just the display dates and not API dates (There’s no shame in localisation for user interface and not API because API’s are usually quite specific)

regards

dwmw2:
Thank you. I am not the right person to be doing the web-side JavaScript stuff; it took me long enough to do the bits I have already done. We really want someone with a different skill set for that anyway.

gizmocuz:
Duo a recent PR there is an issue in LightsController.js, is it possible to adjust this PR to fix this.
The code that was inside this controller is probably now in DeviceTimer.js

dwmw2:
Rebased and tested, and ready to merge again now; thanks.

dwmw2:
Updated again to cope with #2213

dwmw2:
Updated to resolve conflicts as @stas-demydiuk moved some of the US-only bits around again in #2316 

Would be nice to get this merged.

gizmocuz:
Thanks!!!

PR #1974, opened 2017-11-30
HardwareMonitor Domoticz process memory usage

jvandenbroek:
This will monitor process usage for the Domoticz process on Linux (unfortunately other platforms are much harder to monitor, but I think 99% of the users with memory issues use Linux anyway) by getting the relevant VmRSS + VmSwap values from it's status file. This way reports of memory leakages etc are much clearer, because we now know for sure whether it's generic memory usage or specific to Domoticz without having to dig too deep.

Changes to SendCustomSensor are needed to work correctly with child id > 255

### DIFF
--- a/hardware/DomoticzHardware.cpp
+++ b/hardware/DomoticzHardware.cpp
@@ -723,18 +723,19 @@ void CDomoticzHardwareBase::SendWaterflowSensor(const int NodeID, const int Chil
 
 void CDomoticzHardwareBase::SendCustomSensor(const int NodeID, const int ChildID, const int BatteryLevel, const float Dust, const std::string &defaultname, const std::string &defaultLabel)
 {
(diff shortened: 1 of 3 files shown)


gizmocuz:
Thanks! Let's hope it will be a steady line after an hour

Siewert308SW:
Great pull request.
But is it also possible to insert the requested Memory (VIRT) as this is for now the main culprit monitor.
This VIRT is increasing insane at the moment and a separate github thread about a possible memory leak is running.

jvandenbroek:
Virt as shown in top is not relevant, it might even grown as big as the total memory in the system. This is nothing to worry about and normal behavior, it's mapped but unused memory. You should really only be looking at RSS, which is what this patch does (both in RAM and swapped out).

PR #1978, opened 2017-11-30
Customisable start with no loss at upgrade

travellingkiwi:
This patch is to alter the default domoticz.sh startup script so that you can drop in place a local.settings file and

1. Run domoticz in a directory other than /home/pi
2. Run domoticz on a port other than 8080
3. Configure your logging how you like it
4. Not lose those customisations when domoticz upgrades itself and restarts using the domoticz.sh script that is provided with the distribution


gizmocuz:
Sorry, but you have to make commits against the development branch (which is the default when you checkout domoticz)

domoticz.sh
```diff
@@ -11,20 +11,34 @@
 
 # Do NOT "set -e"
 
+SCRIPT=$(readlink -f "$0")
```
gizmocuz, line 14: why not the 'pwd' command ?
for the local.settings, maybe better to use /etc/domoticz.conf ?

gizmocuz:
@travellingkiwi, when your back from traveling, could you have a look at my comments? Otherwise we need to close this pull request

domoticz.sh
travellingkiwi: It's not `pwd` because I'm not interested in what directory the script is being run from. It's interested in where the script is located. 

Sometimes they might be the same. But I know when I run it manually, they're usually not, so depending on the actual location is better.

travellingkiwi:
Branch to merge into changed to Development

domoticz.sh
gizmocuz: +1, thanks for the clarification!

gizmocuz: How about the use of /etc/domoticz.conf?

gizmocuz:
Maybe by excident you added changed to Rtl433 ? Is it possible to undo them and if needed create a separate pull request ?
Please try not to use boost lexical cast, an atoi(string.c_str()) will also work

travellingkiwi:
Seems to be an artefact of diff between my fork and development... I'm re-doing from scratch.

PR #1982, opened 2017-12-04
UpdateDevice now handles python plugin

guillaumezin:
UpdateDevice now handles python plugin (mainly for custom sensors modification from JSON calls)

Cf. http://www.domoticz.com/forum/viewtopic.php?f=6&t=20634&p=160396#p160396


gizmocuz:
Thanks!

Issue #1990, opened 2017-12-11
OWL CM180 missing 12s update


slybart77:
Hi

It seems there is a problem with the OWL CM180 during the update process.
Domoticz does not take into account every 12s update from the device but only the ones available every minute.

In main/mainworker.cpp **MainWorker::decode_Energy**, devType and subType are different when calling m_sql.GetLastValue => devType=pTypeENERGY and subType=sTypeELEC3
compared to when submitting new values with "decode_General" devType=pTypeGeneral and subType=sTypeKwh

What is the appropriate class for this device? pTypeENERGY/sTypeELEC3 or pTypeGeneral/sTypeKwh

An easy fix can be done by applying the appropriate parameters.

Thanks

PR #1991, opened 2017-12-11
Fix CM180 12s update not taken into account

slybart77:
Hi,
I don't know if this is the best way to fix this problem but it works. It solves issue #1990 
I think it should better to be clear about the use of subType and devType in this case.
Comment is welcome.
Thanks
Nicolas


gizmocuz:
Sorry, but you should make your patch against our development branch

PR #1992, opened 2017-12-11
Fix CM180 12s update not taken into account

slybart77:
Hi,
I don't know if this is the best way to fix this problem but it works. It solves issue #1990
I think it should be better to clarify things about the use of subType and devType in this case.
Comment is welcome.
Thanks
Nicolas

### DIFF
--- a/main/mainworker.cpp
+++ b/main/mainworker.cpp
@@ -8384,7 +8384,7 @@ void MainWorker::decode_Energy(const int HwdID, const _eHardwareTypes HwdType, c
 	unsigned char devType = pTypeENERGY;
 	unsigned char subType = pResponse->ENERGY.subtype;
 	std::string ID;
-	sprintf(szTmp, "%d", (pResponse->ENERGY.id1 * 256) + pResponse->ENERGY.id2);
+	sprintf(szTmp, "%08X", (pResponse->ENERGY.id1 * 256) + pResponse->ENERGY.id2);
 	ID = szTmp;
 	unsigned char Unit = 0;
 	unsigned char cmnd = 0;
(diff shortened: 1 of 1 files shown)


gizmocuz:
this should not be needed at is it translated to the new sensor type later on in the code ?

Line 8425
//Translate this sensor type to the new kWh sensor type

Or does it go wrong here ?

Line 8409
//Retrieve last total from current record

slybart77:
but line 8415 the **m_sql.GetLastValue** will use the wrong sensor type and subtype.
In this case, there is no match in the database. The "retrieve last total from current record" will never return a value.
 I forgot also to mention an other mistake line 8387 (I forgot to fix it in my commit...)
sprintf(szTmp, "%d", (pResponse->ENERGY.id1 * 256) + pResponse->ENERGY.id2);
it should be:
sprintf(szTmp, "%08X", (pResponse->ENERGY.id1 * 256) + pResponse->ENERGY.id2);
Otherwise, no match in database.

to illustrate the problem:
Query from **decode_General**:
 SQLQ query : SELECT ID,Name, Used, SwitchType, nValue, sValue, LastUpdate, Options FROM DeviceStatus WHERE (HardwareID=2 AND DeviceID='00008C42' AND Unit=1 AND Type=243 AND SubType=2

Query from **m_sql.GetLastValue**:
SQLQ query : SELECT nValue,sValue,LastUpdate FROM DeviceStatus WHERE (HardwareID=2 AND DeviceID='00008C42' AND Unit=0 AND Type=90 AND SubType=2) order by LastUpdate desc limit 1

As you can see, **Type** is different in the queries.

Do you want me to make another pull request with the missing fix?

gizmocuz:
You do not need to create another pull request, just committing to your branch was enough and it will be updated here.
So as i expected it is the retrieval of the last value (~ line 8409), so here and i think only here we might need to correct the type/subtype ?

slybart77:
Yes, we can change the Type and subType only there to keep compatibility with the device registered as pTypeGeneral/sTypeKwh. Otherwise, the "GetLastValue" function will use parameters above, meaning pTypeENERGY/sTypeELEC3.
It works for me. On Domoticz I have now the value refreshed every ~12s.

gizmocuz:
That would be great! You have a few changes in this commit that should not be there...
(createdevice changes)

slybart77:
I have performed a rebase on the development branch... not such a good idea after all...
I have just pushed a new commit with only useful stuff.

gizmocuz:
Thanks for this patch !!

gizmocuz:
i just made one small patch:
https://github.com/domoticz/domoticz/commit/971f74d0fe6147834fcc915f7e5fb9f22e03e386

PR #1994, opened 2017-12-13
Don't lose customised settings on update

travellingkiwi:
This patch is to alter the default domoticz.sh startup script so that you can drop in place a local.settings file and

Run domoticz in a directory other than /home/pi
Run domoticz on a port other than 8080
Configure your logging how you like it
Not lose those customisations when domoticz upgrades itself and restarts using the domoticz.sh script that is provided with the distribution

### DIFF
--- a/domoticz.sh
+++ b/domoticz.sh
@@ -11,20 +11,34 @@
 
 # Do NOT "set -e"
 
+SCRIPT=$(readlink -f "$0")
+BASENAME=$(dirname ${SCRIPT})
+
+USERNAME=pi
+HTTP_PORT=8080
+HTTPS_PORT=443
+DAEMON_OPT="-daemon"
+
+echo "Using BASE '${BASENAME}'"
+
+if [ -f ${BASENAME}/local.settings ];
+then 
+  . ${BASENAME}/local.settings
+fi
 PATH=/sbin:/bin:/usr/sbin:/usr/bin:/usr/local/bin
 DESC="Domoticz Home Automation System"
 NAME=domoticz
-USERNAME=pi
 PIDFILE=/var/run/$NAME.pid
 SCRIPTNAME=/etc/init.d/$NAME
 
-DAEMON=/home/$USERNAME/domoticz/$NAME
-DAEMON_ARGS="-daemon"
-#DAEMON_ARGS="$DAEMON_ARGS -daemonname $NAME -pidfile $PIDFILE"
-DAEMON_ARGS="$DAEMON_ARGS -www 8080"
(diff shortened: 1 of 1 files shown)


gizmocuz:
How about using /etc/domoticz.conf as the persistent configuration file ?
And if found, it should contain all startup parameters/arguments and skip all parameters defined in this file ?

Leave everything like it is, except under the DEAMON_ARGS issue

DAEMON=/home/$USERNAME/domoticz/$NAME
DAEMON_ARGS="-daemon"
#DAEMON_ARGS="$DAEMON_ARGS -daemonname $NAME -pidfile $PIDFILE"
DAEMON_ARGS="$DAEMON_ARGS -www 8080"
DAEMON_ARGS="$DAEMON_ARGS -sslwww 443"
#DAEMON_ARGS="$DAEMON_ARGS -log /tmp/domoticz.txt"
#DAEMON_ARGS="$DAEMON_ARGS -syslog"
 
```
if [ -f /etc/domoticz.conf ];
then 
  . /etc/domoticz.conf
fi
```

the content of /etc/domoticz.conf could be

DAEMON=/home/$USERNAME/domoticz/$NAME
DAEMON_ARGS="-daemon"
DAEMON_ARGS="$DAEMON_ARGS -www 8180"

So it only starts as a deamon on port 8180

travellingkiwi:
I don't like user/app config going into /etc - /etc is for system configs. Not app. Plus that limits you to 1 copy per host. Why should it be limited?

Turning it around makes it less clear what it's doing. And specific to just this. Having a set of defaults, then losing any changes to them and using the same code to process (i.e. completing DAEMON_ARGS) I believe makes it clearer and more easily extensible.

For example - if you add in a new parameter that's required that (should) have a sensible default you then break the startup. Which is what I was attempting to fix in the first place. 

Changing it around does not lend itself to extension.

gizmocuz:
Okey, but the whole idea is that you COPY this file in /etc/init.d so it does not get reverted with an update
if you use the 'sudo' method to install domoticz (see our website) this is the way it is done.
You can place multiple copies in /etc/init.d under different names, using different paths for different instances

travellingkiwi:
Yes. But if you have to manually copy the script, you also have to update it. 

This way you don't have to do either. Simply symlink the startup script in /etc/init.d to the real location and away you go. When the script itself changes. you get the changes. And get to keep your custom settings (Or continue to use the defaults).

gizmocuz:
There has been no startup changes for a long time.
You should not symlink the script
See our instructions on our wiki

Closing issue as it's not really an issue if you follow the guidelines.

PR #1995, opened 2017-12-13
Made handler a member variable of CWebsocket() instead of a pointer.

rkluwen:
As per discussed in the Github forum thread: Memory Leak.
This patch makes the request handler a member variable of the CWebsocket() class instead of an allocated pointer.


LemarinelNet:
Have you stress-tested it ?

gizmocuz:
Super thanks again! This indeed seems to solve the allocation problem!

PR #1998, opened 2017-12-15
Add support for Twilight and day length

ghost:
Twilight is the time between day and night when there is light outside, but the Sun is below the horizon. In the morning, twilight begins with dawn, while in the evening it ends with dusk.

Why would someone be interested in Twilight times? 
- In many parts of the world, as in northern Sweden right now for example there is no sunrise (and no sunset as well) and this "polar night" may last for a few months. The timeofday['SunriseInMinutes'] and timeofday['SunsetInMinutes'] both show "00:00" (Which is actually wrong). That is, all my Lua scripts based on those timeofday variables are useless.

- Furthermore in a more common scenario someone maybe likes to turn the lights on outside in the evening when it's usually gets dark. Civil twilight is the brightest of the 3 twilight phases. The Sun is just below the horizon, so there is generally enough natural light to carry out most outdoor activities. Problem is that the time between sunset and civil dusk varies, and it can vary a lot depending on where you live. just adding a fixed of time to approximate won't do well and it's certainly not easy to share the scripts with people living at other places in the world. 

That's why we need to make Domoticz aware of twilight.

While adding awareness of twilight it would also be valuable to add a time for when the sun is at the southernmost point. (That works actually well even when there is no sunrise). That time is in most cases not 12:00 at every place. It may actually differ a lot from that and even more when there is daylight savings time. "SunAtSouth" is a very versatile variable when making lua scripts.

My primary goal with this was to make Domoticz aware about location dependent Twilight and being able to provide those values into the Lua scripting environment using the timeofday table. I've also aimed to extend the Domoticz API /json.htm?type=command&param=getSunRiseSet so that it provides time stamps for the 3 most commonly used  twilights, when the sun is in the south and the length of the day. 

While implementing this I have also by looking at wherever Domoticz deals with sunrise and sunset and added code for twilight. That doesn't mean for example that scheduling using a switch device in the GUI support the new twilight timer options but rather that there is a "base" to start from if someone cares to implement it further. The same goes for Blockly. We could naturally choose to just discard that base if it's considered to be worthless. It's my believe though that this "base" is made in such a way that it shouldn't alter the way that things works today.

Naturally as this is my my first time to do a pull request for Domoticz, I can not fully grasp how Domoticz works and the changes needs to be reviewed by someone who's got the full picture so that the code changes that I have done isn't harmful nor useless.


ghost:
Just adding some information about the suggested JSON output from the API  `/json.htm?type=command&param=getSunRiseSet `

This is data from northern Norway today. No sunrise and no sunset but there is still twilight!

```
{
   "AstrTwilightEnd" : "16:29",
   "AstrTwilightStart" : "06:06",
   "CivTwilightEnd" : "13:13",
   "CivTwilightStart" : "09:22",
   "DayLength" : "00:00",
   "NautTwilightEnd" : "15:07",
   "NautTwilightStart" : "07:28",
   "ServerTime" : "2017-12-15 10:54:51",
   "SunAtSouth" : "11:17",
   "Sunrise" : "00:00",
   "Sunset" : "00:00",
   "status" : "OK",
   "title" : "getSunRiseSet"
}
```
This is data for Amsterdam for the same day:

```
{
   "AstrTwilightEnd" : "18:34",
   "AstrTwilightStart" : "06:37",
   "CivTwilightEnd" : "17:08",
   "CivTwilightStart" : "08:03",
   "DayLength" : "07:43",
   "NautTwilightEnd" : "17:52",
   "NautTwilightStart" : "07:19",
   "ServerTime" : "2017-12-15 11:37:02",
   "SunAtSouth" : "12:35",
   "Sunrise" : "08:44",
   "Sunset" : "16:27",
   "status" : "OK",
   "title" : "getSunRiseSet"
}
```

Below is an example LUA script that can be used to pull some information to the Domoticz log file:
```
commandArray = {}

-- get current time
timenow = os.date("*t")
minutesnow = timenow.min + timenow.hour * 60

print("\n==================================================================================")
-- test for sunset - 10 minutes
if (minutesnow == timeofday['CivTwilightEndInMinutes']) then
	print("Civil Twilight Ended!")
end

function MinutesToClock(minutes)
  local minutes = tonumber(minutes)

  if minutes <= 0 then
    return "00:00";
  else
    hours = string.format("%02.f", math.floor(minutes/60));
    mins = string.format("%02.f", math.floor(minutes - (hours*60)));
    return hours..":"..mins
  end
end

print (MinutesToClock(timeofday['AstrTwilightStartInMinutes']) .." AstrTwilightStartInMinutes. Number of minutes: ".. timeofday['AstrTwilightStartInMinutes']);
print (MinutesToClock(timeofday['NautTwilightStartInMinutes']) .. " NautTwilightStartInMinutes. Number of minutes: ".. timeofday['NautTwilightStartInMinutes']);
print (MinutesToClock(timeofday['CivTwilightStartInMinutes']) .. " CivTwilightStartInMinutes. Number of minutes: "..timeofday['CivTwilightStartInMinutes']);
print (MinutesToClock(timeofday['SunriseInMinutes']) .. " SunriseInMinutes. Number of minutes: "..timeofday['SunriseInMinutes']);
print (MinutesToClock(timeofday['SunAtSouthInMinutes']) .. " SunAtSouthInMinutes. Number of minutes: "..timeofday['SunAtSouthInMinutes']);
print (MinutesToClock(timeofday['SunsetInMinutes']) .. " SunsetInMinutes. Number of minutes: "..timeofday['SunsetInMinutes']);
print (MinutesToClock(timeofday['CivTwilightEndInMinutes']) .. " CivTwilightEndInMinutes. Number of minutes: "..timeofday['CivTwilightEndInMinutes']);
print (MinutesToClock(timeofday['NautTwilightEndInMinutes']) .. " NautTwilightEndInMinutes. Number of minutes: "..timeofday['NautTwilightEndInMinutes']);
print (MinutesToClock(timeofday['AstrTwilightEndInMinutes']) .. " AstrTwilightEndInMinutes. Number of minutes: "..timeofday['AstrTwilightEndInMinutes']);
print ("Length of the day: " .. MinutesToClock(timeofday['DayLengthInMinutes'])..". Number of minutes: "..timeofday['DayLengthInMinutes']);
local civilDawnLength = timeofday['CivTwilightEndInMinutes'] - timeofday['SunsetInMinutes']
print ("During "..MinutesToClock(timeofday['SunsetInMinutes']).." - ".. MinutesToClock(timeofday['CivTwilightEndInMinutes']).." is Civil Dusk. It lasts for: "..civilDawnLength.." minutes.")

print("\n==================================================================================\n")

return commandArray
```

gizmocuz:
Thanks! You patch is almost looking perfect except:

Do add the new timer types at the end of '_eTimerType' because this value is used in all timers and now you shift everything. (like TTYPE_ONTIME was for example 2, and now ~17)

It looks you now always return 'CivTwilightStart' for example. This will add a lot of traffic as it is queried every 10 seconds.
As it is currently only used in the event system, maybe do not provide it via the web server yet.

I dont know why it returns 00:00 in Norway, it does not seem to do so on the below link:
https://www.timeanddate.com/sun/norway/oslo

Have you considered using this excellent script ?
http://www.domoticz.com/wiki/Real-time_solar_data_without_any_hardware_sensor_:_azimuth,_Altitude,_Lux_sensor...

ghost:
Thanks @gizmocuz 

New timers in main/RFXNames.h has been moved. Thanks, I missed that they were referenced by their position. Committed.

About 'always return 'CivTwilightStart' ', I believe that you refer to `	int intSunAtSouth = getSunRiseSunSetMinutes("SunAtSouth");` etc in the CEventSystem::EvaluateLua function. I must agree that it doesn't look pretty. I will come up with a new suggestion shortly.

About 'returns 00:00 in Norway' you must go a bit further north than Oslo. Have a look at the Swedish city Kiruna for example https://www.timeanddate.com/sun/sweden/kiruna

And finally about the solar data script, yes I already use it and I've translated it for dzVents and share it to the public as I do with many other scripts. The solar script is nice, I agree. The current "Sun Azimuth" and the "Cloud Cover Data" is very useful for automating things. It lacks a lot of essential data though.

I'll get back shortly. Thanks!

ghost:
I have now done some changes that I hope will improve performance due to that various "sun-timestamp-strings" were recalculated to minutes at a very frequent basis. Those sun-timestamp-strings" are now calculated in MainWorker::GetSunSettings() that saves them in a vector m_SunRiseSetMins so that they can easily be retrieved whenever needed without the need to do any calculations at all.  MainWorker::GetSunSettings() should only happen at system start, new day and on the change of location settings so it won't happen often at all.

It's my hope and believe that this change can also contribute to some performance improvements in other parts of the code because the getSunRiseSunSetMinutes() function is used at 10 more places in EventSystem.cpp.

Furthermore there are 4 hardware source files (EnphaseAPI.cpp, GoodweAPI.cpp, SBFSpot.cpp and SolarEdgeAPI) that have their own private getSunRiseSunSetMinutes() function. (Still compatible though) They should all certainly benefit from if we moved the function to MainWorker and made it public. But that's not within the scope of this patch.

gizmocuz:
Those hardware classes still use the mainworkers 'm_LastSunriseSet' value which is calculated there
But a general function is always a good idea ;)

main/RFXNames.h
```diff
@@ -57,7 +57,21 @@ enum _eTimerType
 	TTYPE_MONTHLY_WD,
 	TTYPE_YEARLY,
 	TTYPE_YEARLY_WD,
-	TTYPE_END
+	TTYPE_END,
```
gizmocuz: TTYPE_END should be at the 'end'

ghost: Thanks. That has been taken care of now.

gizmocuz:
Thanks for this patch !!

imcfarla2003:
My lua script which has 
`if (timeofday['Nighttime']) then`
now doesn't know it is nighttime - what do I need to change?

ghost:
Please see this pull request: https://github.com/domoticz/domoticz/pull/2007/commits/dc9774ff2cd56fdce99a2ee8efae45251936bf9f

gordonb3:
Uhm... guys?

I have not installed this yet, but as I was looking into another matter I came across a change in Scheduler.cpp that doesn't seem right to me. I'm referring to this particular instance:

`	time_t allTimes[] = {m_tSunRise, m_tSunSet, m_tSunAtSouth, m_tCivTwStart, m_tCivTwEnd, m_tNautTwStart, m_tNautTwEnd, m_tAstTwStart, m_tAstTwEnd};`
`		for(unsigned int a = 0; a < sizeof(allSchedules)/sizeof(allSchedules[0]); a = a + 1)`
`...`
`				allTimes[a] = temptime;`

I am missing where that updates the value of e.g. m_tSunRise

ghost:
> I am missing where that updates the value of e.g. m_tSunRise

The allTimes vector and the allSchedules match each other, e.g the first element in allSchedules is the variable sSunRise that matches the allTimes vector element m_tSunRise (They share the same ordinal number)

So instead of having a code block for each timer type, we are iterating over those vectors and the statement `allTimes[a] = temptime;`should set the m_tSunRise when the iterator is 0.

Though the code is much shorter than before it's intended to do the same job and its also handling 7 additional timer types.

Do you have any reason to believe that it's not working as intended? If so I can try to debug it further.

gordonb3:
The way I read it is that `allTimes` is an array that is initialized with the values `m_tSunRise`, `m_tSunSet`, etc and then those array values get updated within that loop.

I may be missing something, but I'd say what you probably want to do is load the array with pointers to those values instead and then dereference those pointers when loading the new values that are passed to `SetSunRiseSetTimers()`

ghost:
Thanks a lot! I just discovered that when debugging! I will fix it immediately.

ghost:
@gordonb3 , does [this](https://github.com/allan-gam/domoticz/commit/7e536e827d6fd4824a186e13acda89781a5a7a92) look OK to you? 

I've ran it and it seems to set the values properly now.

gordonb3:
I had the same thought of what it should look like. Yes.

DebugBill:
Hi,

My scheduled switches based on sunset don't work anymore. Looking into the code I see an array "allTimes" that gets initialized with great effort... before being discarded apparently before it got a chance to be used. It is defined in a block that only exists for a few lines. Is that normal? 
The sure thing is that now the values for m_tSunSet or m_tSunRise around lines 460 are now 0. Looks like a bug to me be I did not have enoughh time to fully understand this part of the code.

ghost:
@DebugBill , have you had a look at the pull request #2008 The problem has been known and also fixed for a long time. It's just waiting to be merged.

It should fix the problem.

DebugBill:
Hi,

I hadn’t seen the fix, sorry, I should have checked to on-going pull requests. Forget about my message. I will test as soon as it is merged but indeed using pointers as you did should fix it. I’m aslo waiting for a bug fix merge on another part of the code.

Blaise


De : allan-gam [mailto:notifications@github.com]
Envoyé : vendredi 5 janvier 2018 18:38
À : domoticz/domoticz
Cc : Blaise Thauvin; Mention
Objet : Re: [domoticz/domoticz] Add support for Twilight and day length (#1998)


@DebugBill<https://github.com/debugbill> , have you had a look at the pull request #2008<https://github.com/domoticz/domoticz/pull/2008> The problem has been known and also fixed for a long time. It's just waiting to be merged.

It should fix the problem.

PR #2001, opened 2017-12-17
OpenWebNet: change password control for MHS1

Spugna85:


gizmocuz:
Thanks!, Could you also bump the version number in the appcache file, this needs to be done when making modifications to the www folder.

dannybloe:
Thanks

PR #2003, opened 2017-12-18
Make getSunRiseSet API handle queries for any geographical position

ghost:
Domoticz API can already return JSON data for the sunrise, sunset etc for the current location and the current date by using the URL `/json.htm?type=command&param=getSunRiseSet`

This little patch will make it possible to add additional command array parameters returning data for the specified position in longitude and latitude for any part of the world. Furthermore it allows to add a date, e.g. 

`/json.htm?type=command&param=getSunRiseSet&Latitude=52.370216&Longitude=4.895168&Date=2019-06-23`

Because all time stamps in the result shows the servers local time, the UTC offset used in case any calculations needs to be made.

Below is the result using the URL above (server time is UTC+1)
```
{
   "AstrTwilightEnd" : "00:00:00",
   "AstrTwilightStart" : "00:00:00",
   "CivTwilightEnd" : "21:56:00",
   "CivTwilightStart" : "03:29:00",
   "DayLength" : "16:48:00",
   "Latitude" : "52.370216",
   "Longitude" : "4.895168",
   "NautTwilightEnd" : "23:18:00",
   "NautTwilightStart" : "02:07:00",
   "Remarks" : {
      "AstronomicalTwilight" : "The astronomical day is 24 hours"
   },
   "ServerTime" : "2017-12-18 20:34:37",
   "SunAtSouth" : "12:42:00",
   "SunCalcDate" : "2019-06-23",
   "Sunrise" : "04:18:00",
   "Sunset" : "21:07:00",
   "UtcOffset" : 1.0,
   "status" : "OK",
   "title" : "getSunRiseSet"
}
```

### DIFF
--- a/main/EventSystem.cpp
+++ b/main/EventSystem.cpp
@@ -2798,6 +2798,12 @@ void CEventSystem::EvaluateLua(const _tEventQueue &item, const std::string &file
 	lua_pushstring(lua_state, "DayLengthInMinutes");
 	lua_pushnumber(lua_state, intDayLength);
 	lua_rawset(lua_state, -3);
+	lua_pushstring(lua_state, "Latitude");
+	lua_pushnumber(lua_state, m_mainworker.m_LatLong[0]);
+	lua_rawset(lua_state, -3);
+	lua_pushstring(lua_state, "Longitude");
+	lua_pushnumber(lua_state, m_mainworker.m_LatLong[1]);
+	lua_rawset(lua_state, -3);
 	lua_setglobal(lua_state, "timeofday");
 
 	{
--- a/main/SunRiseSet.cpp
+++ b/main/SunRiseSet.cpp
@@ -139,9 +139,8 @@ bool SunRiseSet::GetSunRiseSet(const double latit, const double longit, const in
 	//nautlen = day_nautical_twilight_length(year,month,day,longit,latit);
 	//astrlen = day_astronomical_twilight_length(year,month,day,longit,latit);
 
-	double _tmpH;
-	result.DaylengthMins = modf(daylen, &_tmpH)*60+0.5;
-	result.DaylengthHours = _tmpH;
+	result.DaylengthHours = (int)daylen;
+	result.DaylengthMins  = (int)((daylen-(int)daylen)*60+0.5);
 
 	rs = sun_rise_set(year, month, day, longit, latit, &rise, &set);
 	civ = civil_twilight(year, month, day, longit, latit, &civ_start, &civ_end);
@@ -150,8 +149,8 @@ bool SunRiseSet::GetSunRiseSet(const double latit, const double longit, const in
 
 	rise = UtcToLocal(rise, timezone);
 	set = UtcToLocal(set, timezone);
-	result.SunAtSouthMin = modf((rise+set)/2.0, &_tmpH)*60+0,5;
-	result.SunAtSouthHour = _tmpH;
+	result.SunAtSouthHour = (int)(rise+set)/2.0;
+	result.SunAtSouthMin  = (int)(((rise+set)/2.0-(int)(rise+set)/2.0)*60+0.5);
 
 	result.evtInfo.push_back(rs);
 	result.evtInfo.push_back(civ);
@@ -160,10 +159,10 @@ bool SunRiseSet::GetSunRiseSet(const double latit, const double longit, const in
 
 	switch(rs) {
 		case 0:
-			result.SunRiseMin = modf(rise, &_tmpH)*60+0.5;
-			result.SunRiseHour = _tmpH;
-			result.SunSetMin = modf(set, &_tmpH)*60+0.5;
-			result.SunSetHour = _tmpH;
+			result.SunRiseHour = (int)rise;
+			result.SunRiseMin  = (int)((rise-(int)rise)*60+0.5);
+			result.SunSetHour = (int)set;
+			result.SunSetMin  = (int)((set-(int)set)*60+0.5);
 			break;
 		case +1:
 		case -1:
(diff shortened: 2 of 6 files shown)


ghost:
I'm currently working on to address the issues that the AppVeyor build reported.

ghost:
So, that should be it. There are no more warnings or errors reported by AppVeyor  and Travis CI build caused by the code that I changed.

gizmocuz:
Thank you for the patch.
Domoticz returns the sun rise/set because this is needed in the main gui and timers can also use this information.
Could you describe why you need the additional data/command request as this is not used (tentatively) in Domoticz.
If it is used in your scripts, you could create your own python/bash script that returns this data?
This because it seems with this patch some more parameters are added in the lua stack that (for now) nobody is using

ghost:
You are correct. This is not necessary for Domoticz to work. It's just a feature allowing Domoticz to handle sun-times for any geographical position and any date. It comes at a low cost because almost all code is already in there.

If you believe that it doesn't add enough value to Domoticz you can close the PR but the three additional commits are fixes that should be merged regardless if the feature is merged. I'm sorry for not making a separate PR for those.

Thanks.

ghost:
Domoticz is currently limited to automate solar based (sunset, sunrise, day length etc) events for a single location.

This feature would allow using a single Domoticz instance by some scripting to handle solar based events for an arbitrary number of locations. You might have a residence in one place and a perhaps a vacation residence at a different location. Here in Sweden solar based events can differ by many hours from south to north. Of course such a situation can also be solved by setting up multiple Domoticz instances.

FWIW this feature is available in at least one more HA system while not existing in some others.

gizmocuz:
That is why you can setup multiple domoticz instances and user the client/server approach
Is it possible to split the pull requests ?

ghost:
First again, I apologize for not have separated the fixes into a separate PR.

I've spent a few hours trying to figure out how to remove 30234d4 and cherry-pick only the 3 specific commits fe70590, ae97088, and d81bbf1. I have failed in doing that because I don't have enough knowledge about git.

I will have to do it later when I get some more time unless someone is willing to give me a helping hand.

ghost:
You have probably already given it a thought, but before I close the PR and delete my repository fork, I'd like to say that it's not just a "new feature".

It will also make the Get sunrise and sunset times JSON API complete. Today without this PR many important elements are missing. Please have a look at the following result running `/json.htm?type=command&param=getSunRiseSet`

```
{
   "ServerTime" : "Sep 30 2013 17:15:21",
   "Sunrise" : "07:38:00",
   "Sunset" : "19:16:00",
   "status" : "OK",
   "title" : "getSunRiseSet"
}
```
There are two time stamps Sunrise and Sunset but they say nothing if you don't know what time zone they relate to. Furthermore the longitude and latitude values are missing. It could be very useful for script writers to be able to pull the longitude and the latitude from the JSON data instead of hard coding them into several scripts. It would also be easier to when sharing scripts with each other if those values were available. Yet another thing is that time stamps for different twilight times are missing.

If adding more parameters into the lua stack is the real culprit here, I suggest that we just comment them out.

Even though I may have implied earlier that the possibility to add arguments in the URL is the "main feature" of this patch, it's actually not. It makes the API result more complete and adding power to the script writers of Domoticz.

It's perfectly OK with me if you reject the PR, I really don't mind at all honestly but it's my believe that it would benefit Domoticz and it's users.

In case of rejection of the PR I can make a new fork and manually edit in the 3 fixes only. No problem.

Cheers! Please let me know your decision.

gizmocuz:
the sunset/rise time stamps are of course related to the local time of the hardware domoticz is running on, and controlling hardware.
In your other location (i have some remotes), you have your own Domoticz instance, with your own hardware.
It runs 100% negatively remotely,and does all needed work.
You might want to let it connect to your 'master' Domoticz instance, which is @home, and you might want to forward some sensors to the master (kwh meters, temp sensors wind, ... and maybe also some switches)
There is no need to know a sunset/rise for the slave domoticz (remote location) because all timers etc are programmed in the remote instance
... and thats how i use it

gordonb3:
> the sunset/rise time stamps are of course related to the local time of the hardware domoticz is running on, and controlling hardware.

Only roughly. In The Netherlands we use CET(D) which is UTC+0100 while in reality we are at UTC+0020. Without the option to set Lat/Lon I would therefore have to program my timers to 40 minutes before sunset/sunrise to make them run at the expected time.

The option to choose any date is in fact a useful idea as well. There have been many comments about the fact that sunset/sunrise timers don't fire at the correct time and it has been established that this is caused by the fact that these timers are usually set for the next day while using today's sunset/sunrise values. Obviously that doesn't require this function to be exposed through json.htm, but if it's there why not? It means people can also use it in scripts to add more flexibility to systems like Squeezebox, Sonox, Evohome, etc.

gizmocuz:
If there is a problem with the sun set/rise values, then (if) those should be addressed.
There is no need for other time functions.
But everything can be done via a script if needed

gizmocuz:
If those calculations are made for the next day, then maybe we could/should specify the previous day when calculating the sunset/rise, but as far as i researched it's for the requested day. Maybe those calculations needs adjustment.
One date system would be enough. Sure people can call external scripts (if needed)

gordonb3:
You should probably have another look. When a recurring timer is triggered we calculate the next time it should fire. This will normally be a question of adding 86400 seconds, but of course sunset/sunrise is variable so the calculation gets a bit more complicated. Our current routine takes today's sunset time, which is obviously in the past as our timer was just triggered, and then adds the additional day to set the timer to be in the future. Of course that is not the correct value, because tomorrow the sunset time will be different.

As a result, sunset/sunrise timers are currently only correct when you started Domoticz at some earlier time on the very same day. e.g. as I do a logrotate on Sunday at 3:00 AM all my sunset/sunrise timers run on the correct time on Sunday but they are wrong every other day of the week.

gizmocuz:
@gordonb3, there have been many poking around in the sunset/timer class over the last years...
From what i see, there is a routing in the mainworker called 'GetSunSettings', this is called at the startup of the application, and every hour.
When this is called, it will update/correct all sunset/rise based timers.
I would suspect that this will be OK that way ?

gordonb3:
Doesn't seem to work...

```
2018-05-14 21:27:01 | On
2018-05-13 21:27:01 | On
2018-05-12 21:25:01 | On
2018-05-12 11:19:52 | Off
2018-05-11 21:22:01 | On
2018-05-10 21:20:01 | On
2018-05-09 21:19:01 | On
2018-05-08 21:17:01 | On
2018-05-07 21:15:01 | On
2018-05-06 21:15:01 | On
2018-05-05 21:12:01 | On
2018-05-04 21:10:01 | On
2018-05-03 21:09:01 | On
2018-05-02 21:07:01 | On
2018-05-01 21:07:01 | On
2018-04-30 21:04:01 | On
2018-04-29 21:04:01 | On
2018-04-28 21:00:01 | On
```
Have had some intermediate restarts, but as you can see the firing times are identical on Sunday (29, 6, 13) when I do my logrotate and Monday. And so today it fired at 21:27 which was yesterday's sunset time instead of today's sunset time which was at 21:28

gizmocuz:
Then something is not updating the timer correctly. It should have been patched every hour looking at the code

PR #2010, opened 2017-12-21
Add Comm5 SM-xxxx sensors support

correa:
Add Comm5 SM-xxxx sensor support. SM sensors have temperature, humidity, and barometric readings.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -307,6 +307,7 @@ hardware/ASyncTCP.cpp
 hardware/AtagOne.cpp
 hardware/BleBox.cpp
 hardware/Comm5TCP.cpp
+hardware/Comm5SMTCP.cpp
 hardware/Comm5Serial.cpp
 hardware/csocket.cpp
 hardware/CurrentCostMeterBase.cpp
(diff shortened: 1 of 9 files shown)


gizmocuz:
Thanks for the patch. Some questions/remarks:

- Could this code be merged in a general Comm5 base class? Is it much different then the existing Comm5 code?
- I see classes for TCP and Serial (older Comm5 code), is the sensor code different? Maybe this can be done better via a base class.
- Could you add a link to the hardware website at the top of the .cpp file
- Could you remove the boost lexical cast include

correa:
Hi gizmocuz, thanks for reviewing this code. 
Yes it's quite different from the other Comm5 code, the existing code is an I/O module, that is, it provides sensors and relays (represented by switches in domoticz) while this one is a sensor with multiple physical measurements: Temperature, Humidity, Air Pressure and Luminosity.  If I were to create a base class I feel there would be minimal boiler plate code to be shared, such as, the error reporting code. May I have the base class issue investigated further and submit a different pull request after this one is merged if it makes sense?

I have inserted a link to comments and removed lexical_cast which wasn't used.

gizmocuz:
Thanks for the additions... Sure you make look at creating a base class... would be great if it supports all (and feature) sensor types

PR #2011, opened 2017-12-21
Bug fix for Teleinfo meters lacking the PAPP feature

DebugBill:
Some meters do not provide the PAPP value (apparent power). A workaround was built in the code but would not be triggered when necessary because of a bug affecting only those meters.

Thanks to H. Lertouani who reported it and provided a fix.


gizmocuz:
Thanks for the patch. But could you resubmit it against the development branch ?

PR #2013, opened 2017-12-22
new prototype for onCommand Python plugins, allows nValue and sValue

guillaumezin:
Following bugs reported in http://www.domoticz.com/forum/viewtopic.php?f=6&t=20634 I propose a new implementation on onCommand prototype for Python plugins. The new implementation is backward compatible, and allows plugins to receive nValue and sValue on updevice json command.

New prototype would be:
onCommand(Unit, Command, LevelOrNvalue, HueOrSValue)

if Command is "udevice", LevelOrNvalue is nvalue (integer) and HueOrSValue is svalue (string), otherwise it works as before, LevelOrNvalue is an integer or a float for thermostat, and HueOrSValue is an integer. I think I would prefer this implementation


dnpwwo:
Hi,

I have some concerns with this approach because it leads us on a path of complexity if other people add special handling for other devices later.  I realised after the fact that I made some mistakes with some of the function prototype due to my inexperience with Python.

In situations like these I now believe we should be passing dictionaries as parameters and then testing to see if a key exists.  I changed onMessage to be like that.  For simple protocols the data parameter is a string (to be backward compatible) and for complex ones (like HTTP) it is a dictionary with the HTTP message broken into component parts. That approach is much better suited to keeping the interface backwards compatible

I have never seen anyone use Hue so why don't you rename the parameter and pass "nValue" and "sValue" in a dictionary or "Hue" for the other cases

guillaumezin:
I agree with you, but I tried to keep backward compatibility. Another implementation is proposed here : https://github.com/guillaumezin/domoticz/commit/0b788890139396d0603636eec66be3d031ada6f6 where sValue is an additionnal parameter. Maybe we can pass the dict as an optionnal parameter?

Or we choose to break the compatibility, it is up to you, I can implement it, now that I already digged into the code.

dnpwwo:
Seems to me there are 3 options:

1.  Change what 2 existing params are used for under some circumstances. This seems like a hack and makes things more complicated and is not extensible without ever increasing complexity
2.  Add an extra optional parameter. From looking at this I believe it will break every existing plugin because they would need to change to add the parameter and give it a default value to cover for when it is not supplied.  This will annoy a lot of people (unless I've misunderstood how optional parameters work)
3. Re-purpose the Hue parameter from a string to a dictionary and rename it to "Details" or something. This may annoy 1 or 2 people (at most) who are using it but then there would be a way of adding more device specific information later to support things like Security Alarms.

Any others?

guillaumezin:
1. ok, no go
2. I've been doing some tests, I'm not that sure that this break old plugins, but cannot be 100% sure
3. I think it is more elegant, I will propose something

guillaumezin:
Ok, here I propose an implementation based on python dict

guillaumezin:
I updated https://github.com/guillaumezin/DomoticzEbusd to show an example of implementation

dnpwwo:
I see one path through the code only returns 3 parameters.  Won't Python throw an error if a default value for the 4th parameter isn't specified? That will break some existing plugins.

I was expecting the code to always return either :

    pParams = Py_BuildValue("isfO", m_Unit, m_Command.c_str(), m_fLevel, pDict);

or

    pParams = Py_BuildValue("isiO", m_Unit, m_Command.c_str(), m_fLevel, pDict);

where pDict was constructed to hold whatever was supplied rather than creating more cascading 'if' statements. Have a look at CPluginProtocolICMP::ProcessInbound to see code for creating dictionaries on the fly.

guillaumezin:
I had a look at CPluginProtocolICMP::ProcessInbound, and it seems overkill and risky to me: it would imply to create a PyObject* pDataDict = PyDict_New(); in the constructor of CommandMessage class, then create objects with PyObject*	pObj = Py_BuildValue(...) and inserting them with PyDict_SetItemString(pDataDict, ...) in each CommandMessage function, with the risk of a memory leak, because I'm really not sure to understand how the reference counter and the garbage collection of Python is working...

guillaumezin:
Ok, I tried as you suggest, please review my proposal, but before that, have a merry Christmas

gizmocuz:
@dnpwwo, did you found some time to review these changes ? thanks in advance!

dnpwwo:
@gizmocuz my concern with this is listed above.  The code as submitted can call onCommand with either 3 or 4 parameters which requires all existing plugins to be updated to make the 4th parameter optional otherwise Python *could* throw an error under some circumstances.

Now that the plugins are in the stable build we need to honor API contracts wherever possible.

gizmocuz:
I agree with that for sure !

guillaumezin:
But I changed my code to follow you suggestion, the 4th argument is not optional anymore

gizmocuz:
@guillaumezin , so now all current user scripts need to be changed ? that is not backwards compatible

guillaumezin:
No, it's backward compatible for plugin using hue parameter. I can code whatever you want, maybe I misunderstood what dnpwoo was expecting. Maybe I should code a different onUpdate function and keep onCommand as it was? We need a way to pass sValues to plugins.

guillaumezin:
Hello, here is a new implementation where I keep onCommand 100% compatible, and I add a onUpdate callback. Plugins are free to implement one, the other, or both. I checked, if a callback isn't defined in a plugin, it is harmless. Example of plugin implementation in last version of https://github.com/guillaumezin/DomoticzEbusd

Prototype :
onUpdate(Unit, Command, Details)
if Command is "udevice", Details contains a dict with {"iValue" : integer, "sValue": string}
if Command is "on", "toggle", "Set Level" or "off", Details contains a dict with {"iLevel": integer, "iHue": integer}
if Command is "Set Level" for a temperature setpoint, Details contains a dict with {"fLevel": float}

dnpwwo:
This PR overlaps with PR #2083 which has a proposed onDeviceModified callback

guillaumezin:
It is not the same purpose, is it ? Here I want to be able to pass sValue to plugins, and to handle the case where udevice command is used from JSON API on a plugin device

dnpwwo:
Your suggested callback covers a single use case of an update to a sValue from a single source.  My understanding is the suggested onDeviceModified seeks to inform the plugin if the device is updated anywhere in Domoticz: HTTP, MQTT, Web interface which seems more in keeping with the intent of the framework

guillaumezin:
Ok, I'll try to integrate #2083

guillaumezin:
Done, now the plugins have onUpdate(Unit, Command, Details) with:
if Command is "remove", Details contains an empty dict, it indicates to the plugin that the Unit has been removed from Domoticz DB as requested in #2083
if Command is "udevice", Details contains a dict with {"iValue" : integer, "sValue": string}
if Command is "on", "toggle", "Set Level" or "off", Details contains a dict with {"iLevel": integer, "iHue": integer}
if Command is "Set Level" for a temperature setpoint, Details contains a dict with {"fLevel": float}

emontnemery:
@guillaumezin Just a question if you don't mind, what do you actually want to use "udevice" for?

guillaumezin:
For plugins to get a callback when receiving an update from JSON RPC, cf. https://www.domoticz.com/wiki/Domoticz_API/JSON_URL's#Update_devices.2Fsensors

emontnemery:
OK, I see, I'll merge your changes and push a new commit in #2083 

PS:
It was a bit tricky to find your changes since you've been working on the development branch instead of branching out.. I find this guide really helpful: https://gist.github.com/Chaser324/ce0505fbed06b947d962

guillaumezin:
Thank you for the tip, I'm a beginner with Git, I'm more a hardware guy

gizmocuz:
If this pull request can be closed because the changes are merged together in another pull request, please do so.
Thanks!

guillaumezin:
Ok, I close this PR, taking the assumption that #2083 will be integrated

PR #2015, opened 2017-12-22
Comm5 TCP and Serial IO driver update

correa:
Update the serial IO hardware to do periodic updates of the IO State.
Fix issue when updating a switch state before the database is updated from the previus change was ignored, leaving the system in an outdated state.

### DIFF
--- a/hardware/Comm5Serial.cpp
+++ b/hardware/Comm5Serial.cpp
@@ -4,8 +4,6 @@
 #include "../main/Helper.h"
 #include "../main/localtime_r.h"
 #include "../main/mainworker.h"
-#include "../httpclient/HTTPClient.h"
-#include "../httpclient/UrlEncode.h"
 
 #include <iostream>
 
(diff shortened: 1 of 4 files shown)


gizmocuz:
Thanks for the patch, some questions/remarks:
- Could you make a base class for the Comm5 so there is no duplicate code for Serial and/or TCP ?
- Could you add the new SendSwitchUnchecked underneath the normal sendswitch code ? or maybe add a parameter to the default sendswitch code so we do not need additional code ?

correa:
Hi gizmocuz,

Thanks for reviewing the commit. There is almost no duplicate code between the serial and TCP classes since one is an ASCII based protocol and the other is a binary protocolo with CRC
    
About the SendSwitch, do you mean to just change the order in which the method appears in the file? I could change the original method's signature to add a bool parameter with default value as suggested, yes.

gizmocuz:
Thanks!

Issue #1981, opened 2017-12-01
1000W during 1h = 1000Wh


sifourquier:
Hi
I have create a fork of domoticz and i have found a bug and your code have the same bug
when a sensor of type General/kWh for example dummy/Electric (Instant+Counter)
have 1000W during 1hour or more. The power usage (Wh) are less of 1000Wh for example 917Wh (he lost 5 minute of power usage)
I join my patch. my patch can remove or add 1 minute of power but it is best of 5minutes

thanks you for your job

[0001-fix-power-kWh-usage.tar.gz](https://github.com/domoticz/domoticz/files/1523253/0001-fix-power-kWh-usage.tar.gz)

gizmocuz:
Thank you !, Could you make a pull request against our development branch with your patch ?

PR #2021, opened 2017-12-24
fix power [kWh] usage

sifourquier:
For calculate the power usage
The laste 5 minutes are not count
The result is in log of power usage if you use 1000w during 1 h the result are 917 kwh
This patch fix it

i have report this bug
https://github.com/domoticz/domoticz/issues/1981

(I don't found how to link a issues to this pull request)


flatsiedatsie:
Wow, that's quite a bug!

gizmocuz:
Thanks!

PR #2023, opened 2017-12-26
Fix web GUI timer days

ghost:
The web GUI needs to parse week days for the new timer types.


gizmocuz:
Thanks for the patch! If possible, in feature web gui (html/js) updates, please also update the appcach manifests internal version number

PR #2025, opened 2017-12-28
changes in I2C hardware module

OndrejLycka:
- added configuration option invert values for sensors witch GPIO

- added configuration option "Path to I2C bus" (into column SeriaPort in hw table) for all I2C sensors. Example value "/dev/i2c-1" or empty for autodetection

- changed store "i2c addres" from column "Port" to column "Address" in hw table

- some cosmetic changes in source code of I2C hardware

### DIFF
--- a/hardware/I2C.cpp
+++ b/hardware/I2C.cpp
@@ -145,19 +145,27 @@ const char* szI2CTypeNames[] = {
 	"I2C_MCP23017",
 };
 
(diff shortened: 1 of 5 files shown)


gizmocuz:
Thanks for the patch !! If possible in a feature web gui (html/js) update, also increase the version number in the appcache file

PR #2026, opened 2017-12-28
Rework of Nest plugin / NestOAuthAPI

dotms:
This pull request is the result of reworking the Nest thermostat plugin as described in [this Domoticz Forum thread](http://www.domoticz.com/forum/viewtopic.php?f=34&t=17370). This is intended as a replacement for the current Nest plugin (eventually) but it differs too much from the current one to be a drop-in replacement. 

Please consider this one experimental since it should work, fine, but needs additional testing by others. I for example do not have 2FA enabled on my Nest account so I cannot reliably tell if it works on those kind of accounts, although I do not see any reason why it wouldn't.

If I get confirmation that this pull request will be accepted I will start writing some basic documentation on the wiki. Until then I think a beta release channel would be suitable.

Comments are welcome.

afbeelding


dotms:
I have some difficulties getting the continuous integration stuff to work. Working on it.

edit: By the looks of it I am using an older version of the WindowsLibraries.7z to build. Project file is referring to the old paths which do no longer exist in the new version. That explains why it works on my machine but CI keeps failing.

dotms:
By the looks of it I can make it compile with the new Boost / 7z library by adjusting the paths in the msbuild/domoticz.vcxproj file. Not sure why this is needed but it appears to work.

gizmocuz:
Is it possible you could submit the pull request against the development branch ? Patches against the master branch are not allowed (Sorry)

poudenes:
I have add this NEST with OAUthAPI way. It work for a while but then i get errors.
2018-03-06 16:40:08.546 Error: NestOAuthAPI: Error login!
2018-03-06 17:01:54.006 Error: NestOAuthAPI: Error getting devices!
2018-03-06 17:03:43.153 Error: NestOAuthAPI: Error getting structures!
2018-03-06 17:04:09.218 Error: CheckAuthToken(4<TOKEN IS PRIVATE) : session id not found
2018-03-06 17:06:32.220 Error: NestOAuthAPI: Error getting devices!

Issue #2000, opened 2017-12-17
Memory leakage!


SargonofAssyria:
After upgrading from latest stable to beta v. 3.8789 there is still memory leakage.
The previous problem with Websocket might be solved, but there is still an other one.
My System:
Version: 3.8789
Platform: Linux domoticz 4.9.35-v7+ #1014 SMP Fri Jun 30 14:47:43 BST 2017 armv7l GNU/Linux
Plugin/Hardware: Raspberry PI 3 with RFXCOM, RazBerry, etc....
Using dzVents with several scripts, all kind of bash scripting to update for instance my APC UPS sensors.
Al you can see from the year graph it started with updating to the latest stable on the 1 aug 2017
Normally I logrotate every day and restart Domoticz, for not getting out of memory.

I have a slave system, which is still on version V 3.8153, with less scripting and devices.
The memory leak is present there too, its less but still there.
The slave has only RFLink as hardware and P1 Smart Meter, and some virtual hardware for Weather etc...

This is the script I use to get the memory use of Domoticz:
ps aux|grep "domoticz/domoticz"|grep -v grep| awk '{print $4}'

![2017-12-17_11-06-29](https://user-images.githubusercontent.com/24824113/34078456-10d38716-e31b-11e7-9d74-f60ec60355c0.png)

gordonb3:
That's not how you measure such a thing. It is normal for a process to gradually build up its physical memory usage as the Operating System does not grant an allocation request until the process actually starts using it. What you need to monitor if you suspect a memory leak is the virtual memory size (VSZ) which is the total memory that the process claims.

bizziebis:
So it's normal for Domoticz to grow 18MB in memory each day? I also noticed the memory leak which was not there in the last stable, but got introduced with the websockets, and got semi-solved with a recent patch.

gizmocuz:
The websocket leak I found has been patched which is one step closer in the right direction
It was also not complicated to find thanks to the help here with the bash script
And the use of the memory profiler on visual studio
Anyone including you can also perform these steps to hunt for additional problems (even if you do not have Windows you can use virtualbox.. all software is free to test)

SargonofAssyria:
You can discuss the way the measurement is done. Fact is that Domoticz is eating up memory which eventually end up with an out of memory ergo crash or lockup of the system. This memory leakage is not due to Websockets, but is introduced between that latest 2 stables as you can see from the graph.
[...]

PR #2028, opened 2017-12-29
Blocky (Eventsystem) memory leak fix

Siewert308SW:
This should fix Blockly (eventsystem) memory leak as reported here: https://github.com/domoticz/domoticz/issues/2000

Original fix was by @jvandenbroek on the 24th july 2017 https://github.com/domoticz/domoticz/pull/1714/commits/c8ec141f3b998ba5ef811074cd2214a677b2a51d
But according to @gizmocuz his advice the check for allocation was put back as discussed here: https://github.com/domoticz/domoticz/pull/1714
Don't know the reason for that but we still had memory leak.

After a couple of days testing with my pull request the leak seems to be fixed.
Vmem stays stable instead if increasing insanely as before the fix.
As mentioned here https://github.com/domoticz/domoticz/issues/2000 i put my production system to the test.
Which stays stable and all eventsystems (blocky, lua, python) worked like charm.

p.s
Don't know if it's related but could it be that the python plugin system is affected by this fix also.
As before the my pull request some plugins after a reboot had a error like (hardware stopped)
After the pull request i didn't see any of them.

### DIFF
--- a/main/EventSystem.cpp
+++ b/main/EventSystem.cpp
@@ -1599,8 +1599,6 @@ void CEventSystem::EvaluateEvent(const _tEventQueue &item)
 lua_State *CEventSystem::CreateBlocklyLuaState()
 {
 	lua_State *lua_state = luaL_newstate();
-	if (lua_state == NULL)
-		return NULL;
 
 	// load Lua libraries
 	static const luaL_Reg lualibs[] =


Siewert308SW:
After some reports not sure if it's the culprit.
On my system it looks okay but some other started reporting it aint, for now i close the pull request and monitor for a longer time period.

PR #2035, opened 2018-01-03
Update base.css

Manu512:
delete line 24 : background-image: linear-gradient(to bottom, #001c38 0%,#0f0c2b 100%);
because this line overwrite background in style.css

### DIFF
--- a/www/styles/default/base.css
+++ b/www/styles/default/base.css
@@ -21,7 +21,6 @@ This theme generally tries not to set any colours, and leaves that to the sub-th
 body {
 	color: #ddd; /*!important;*/
 	background-color: #013 !important;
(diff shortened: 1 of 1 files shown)


dannybloe:
Well, the problem is that this fix brings back that horrible background-image ;-) So shouldn't we just remove that in style.css?

flatsiedatsie:
In my humble opinion:
- Style.css should only provide minimal structural CSS. And it should do this with as minimally defined names as possible so that CSS further down the line (themes) can override it easily.
- Themes should provide things like background images, colours, widths and heights, etc.

So another option would be to move things from style.css to custom.css in the default theme.

gizmocuz:
The idea was to add the gradient in the default theme and not use the image anymore.

PR #2049, opened 2018-01-08
Python plugin MQTT support

emontnemery:
Added MQTT plugin support:
- MQTT support is using 'redboltz' MQTT library which fits nicely into the existing plugin framework
- Add redboltz include files to domoticz source tree with hacks for old boost and compiler related compile errors
- Enable c++11 support, builds on:
  - Ubuntu Trusty (through Travis)
  - OSX + XCode 8.3 (Through Travis)
  - VS2017 (Through appveyor)
  - Debian Jessie (x64)
  - Raspbian Stretch (RPi 3)
  - Ubuntu WSL (x64)

TODO:
- Add support for QoS
- Add support for CA, username and password
- Improve example plugin
- MQTT is a protocol, not a transport, the implementation should be adjusted accordingly (now MQTT is added as a transport)

Other unrelated changes:
- Escape '%'-characters in Python log messages
- Bugfix in HardwareController.js to not show HW dependent fields for plugins
- Restart plugin if plugin owned device is removed from Web UI


emontnemery:
Some additional background on this PR:
I first implemented MQTT support by interfacing with Mosquitto which is already included in Domoticz, but @dnpwwo suggested to stick with the boost::asio templates. Hence I modified the implementation to interface with the Redboltz MQTT library which is a header-only library using boost::asio.

However, @dnpwwo suggests to instead write a custom MQTT client for Domoticz plugin framework. This is fully doable, MQTT is not that complex, but it's not trivial either and will probably require a number of iterations to get it right and corner cases worked out.

The plugin framework already has its own custom HTTP client, so implementing a custom MQTT client is going down the same path. At the same time, the custom HTTP client is quite limited in supported functionality and is not widely tested.

What I want to ask is really: Is there a benefit of doing custom implementations instead of reusing what is already proven to work?

emontnemery:
Closing this since @dnpwwo pushed a custom client instead.

PR #2054, opened 2018-01-10
fix ConnectInternal overflow

Jarthianur:


gizmocuz:
Thanks for the patch! Could you perform patches against the development branch, pull requests to the master branch are not allowed

gizmocuz:
I patched it in the development branch, thanks for finding!

PR #2056, opened 2018-01-12
Icon for "Mobile Devices" in the menu + optimized filesize for all images.

flatsiedatsie:
Both an "on" version and a "off" version of a mobile phone icon, 48x48 pixels png. Has been size optimised.

Based on a public domain image.
https://openclipart.org/detail/19480/modern-touch-phone-mobile
- added little shadow and border spacing, conforming to the style of all other icons I've created for Domoticz.
- created an 'off' layer

In future It might be nice to use CSS Filters to greyscale icons when they are "off". That way only half the icons would need to be shipped. This has a number of advantages.
- Less things to download means a faster and more responsive interface, especially on mobile devices.
- Easier maintenance
- Less controller code required to switch icons. Just toggle an "on" or "off" class to the icon.


JBSoft:
The only problem in your solution to gray icon with css is that you forbid user to change active/non active signification of on/off : in my case, y play with that for my heaters: some are red when on and some are blue when on, depending if it is direct command or via «fil pilote»...

gizmocuz:
Thanks!!

PR #2058, opened 2018-01-12
Add support for new Xiaomi Square Wireless Switch

remcom:
Add support for new Xiaomi Square Wireless Switch that reports as sensor_switch.aq3. The switch is the same as the the one that reports as aq2.


remcom:
I’m closing this pull request. found a small difference between the switches. Switches having the same name but the aq2 supports click and double click and the aq3 click and shake. I'm making a new fix that will reflect this changes

PR #2061, opened 2018-01-12
Update LUA from 5.2 to 5.3.4

Siewert308SW:
Update LUA from 5.2 to 5.3.4
Tested it the past few days with out any issues.

Signed-off-by: Siewert308SW <s.j.lameijer@gmail.com>


Siewert308SW:
Is there something wrong with Travis CI.
I see also some other Pull Requests hanging at the moment.

dannybloe:
Looks like Travis is done.
I do remember Gizmocuz having revert an earlier attempt to upgrade to 5.3 in the past. Guess we really have to test this thoroughly on all platforms first before merging this.

Siewert308SW:
I can't agree more, last attempt was feb 2016 and didn't work out that well.
I can only test it on my Rpi3 and running it for 4 days now without issues.
Will see how it works on other platforms if someone is willing to test it off course ;-)

gizmocuz:
That's correct.. I updated Lua before and there where a lot of problems
There might still be responses about this on our forum

dannybloe:
I do have a lot of Lua based integration tests right now that are run (manually for now) on various platforms and likely will detect problems at an early stage. So after dzVents 2.4.0 is finished I'm glad to help you test this.

dannybloe:
And welcome back gizmocuz .. you were missed :)

koenkooi:
Lua treats increasing the minor version as a major version increase, it will break API and ABI compatibility, so moving from lua 5.2 to lua 5.3 is like moving from 5.x to 6.x for software with sane versioning.

Siewert308SW:
@koenkooi 
Did you try it out because you mention issues?
As for the API it's beyond my knowledge to fix if there are issues.
As mentioned earlier, i can only test it on my system (rpi3) and haven't seen issues and api calls work fine.
I can't speak for those on other platforms.

koenkooi:
@Siewert308SW I didn't try it out in domoticz, but I have had issues with lua 5.x upgrades for other software. It could very well be that the bits domoticz is using weren't broken from 5.2 to 5.3.

Siewert308SW:
@koenkooi 
Don't know how to describe it but Domoticz is using a internal Lua.
Installing upgrading external Lua Package doesn't affect Domoticz.
Actually it does, it will break a lot only if you update the internal Domoticz Lua version it will work.

gizmocuz:
Shall we close this pull request? Does it add any benefit to upgrade Lua @ all ?

Siewert308SW:
Sure, if you wan to close it then be my guest. ;-)
My main reason for upgrading:
- Some fixes
- UTF8 support
- more flexible libraries

If no one benefits this possible upgrade then it's a pitty.
Lucky me i can run it locally... ;-)

gizmocuz:
I try to avoid this : "Actually it does, it will break a lot only if you update the internal Domoticz Lua version it will work"
so if there is no need to update the Lua library version then i would say we shall not do it, otherwise it might result (again) that the eventsystem will break like in the past

Siewert308SW:
I understand and respect your choice.

PR #2064, opened 2018-01-13
Improved Xiaomi/Aqara water sensor detection.

slydiman:


remb0:
great work!!

22nra:
Thanks for the work on this but I have a small query.
I got one of these sensors a few week ago and discovered that when immersed in water they only send one reading when the state changes. The same happens when the sensor was dryed.  A flood sensor is very useful but I would have expected the sensor to transmit regular readings especially when wet. Maybe my sensor has old firmware, has anyone else seen this or got a solution?

slydiman:
I have 3 water sensors and a lot of another Xiaomi devices (~30 total).
It is the expected behavior.
It looks like Xiaomi Gateway can request the state of all sensors.
Domoticz restarting causes reinitialization of Xiaomi Gateway and the system receives the current states.
Probably it is possible to implement an internal timer and regenerate the sensor report if it is in alarm state.
Note the minimal motion sensor period is 2 minutes. Xiaomi Gatawey generates 5 minutes and more reports itself. I woluld like 30 sec at least as Broadlink.
Which kind of the solution do you mean? A user scripts or Domoticz code changes?
I don't like the code of XiaomiGateway.cpp
It contains too much hard coded logics.
A lot of useful data is just lost. It is hard to add support for new devices.
The code is not flexible. It will just stop working if something will be changed in the firmware.
I don't see the battery level from the classic Xiaomi wireless button.
The classic Xiaomi temp/humidity sensor reports 3 devices - temp, humidity and temp + humidity.
The temp device reports always 100% level, 2 other report a reliable value. 
The Aqara temp/humidity/pressure sensor resports 4 devices - temp, humidity, temp + humidity and pressure.
The temp device again reports always 100%, 3 other never report the battery level.
The battery level of light switch is always 110% :)

22nra:
Thanks for getting back to me so quickly and confirming this expected behavior. Have used these sensors for some time and noticed that they can sometimes (rarely) miss the odd reading. Sending one value when the state changes will extend battery life but makes this flood sensor a little risky. Great pity that we dont get a dry update every hour and if flood condition is triggered the update frequency is higher.
Have not tried Domoticz yet but from what you say it looks unreliable using these sensors :(

PR #2065, opened 2018-01-14
dzVents 2.4.0

dannybloe:
This pull request introduces dzVents 2.4.0, a major update. 

**Note to the reviewers:**
Unless you want to dive into the Lua part of dzVents runtime files you can limit your review to these files:

**EventSystem.cpp** 

- batch queued items for evaluation (lua / dzvents)
- extended queue to contain all current (json) data
- moved more dzvents related code to dzvents class (dzvents.cpp/h)
- added URL to queue / trigger reason

**HTTPClient.cpp/h**

- added reponse headerdata function, get headers and status code back in vector
concurrent_queue.h
- added size getter for batching queued items

**SQLHelper.cpp/h**
- extended AddTaskItem to allow cancel an item's all remaining queued actions
- extended GETURL handling to 
- added updatedevice as queue item 

**Other (to be specified)**
- hardware/HardwareMonitor.h
- main/Helper.cpp/h: better random generator for timing options
- main/Logger.cpp: coloring of errors in the log
- main/WebServer.cpp
- main/concurrent_queue.h
- main/dzVents.cpp/h
- www/eventsframe.html: extra dzVents boilerplates

**dzVents 2.4.0 Changelog**
- **BREAKING CHANGE**: The second parameter passed to the execute function is no longer `nil` when the script was triggered by a timer or a security event. Please check your scripts. The second parameter has checks to determine the type. E.g. `execute = function(domoticz, item) .. end`. You can inspect `item` using: `item.isDevice`, `item.isTimer`, `item.isVariable`, `item.isScene`, `item.isGroup`, `item.isSecurity`, `item.isHTTPResponse`. Please read the documentation about the execute function.
- Added ``.cancelQueuedCommands()`` to devices, groups, scenes and variables. Calling this method will cancel any scheduled future commands issued using for instance `.afterMin(10)` or `.repeatAfterMin(1, 4)`
- Added `.devices()` collection to scenes and groups to iterate (`forEach`, `filter`, `reduce`, `find`) over the associated devices.
- Added http response event triggers to be used in combination with `openURL`. You can now do `GET` and `POST` request and handle the response in your dzVents scripts **ASYNCHRONICALLY**. See the documentation. No more json parsing needed or complex `curl` shizzle.
- Added a more consistent second parameter sent to the execute function. When a timer is triggered then the second parameter is a Timer object instead of nil. This way you can check the baseType of the second parameter and makes third parameter (triggerInfo) kind of obsolete. Every object bound to the second parameter now has a baseType.
- Added locked/unlocked support for door-locks (locked == active).
- Moved utility functions from the domoticz object to `domoticz.utils` object. You will see a deprecation warning when using the old function like `round()`, `toCelsius()` etc.
- Added `lodash` as a method to `domoticz.utils`: `domoticz.utils._`
- Added `toJSON` and `fromJSON` methods to domoticz.utils.
- Added `afterXXX()` and `withinXXX()` support for device-update commands. E.g.: `myTextDevice.updateText('Zork').afterMin(2)`.
- Added support for Logitech Media Server devices (thanks to Eoreh).
- Added new timer rules: date rules: `on 13/07`, `on */03`, `on 12/*`, `on 12/04-22/09`, `on -24/03`, `on 19/11-`, week rules: `in week 12,15,19-23,-48,53-`, `every even week`, `every odd week`. See documentation.
- Added historical data helper `delta2(fromIndex, toIndex, smoothRangeFrom, smoothRangeTo, default)` to have a bit more control over smoothing. You can specify if want to smooth either the start value (reference) and/or the to value (compared value).
- Added historical data helper `deltaSinceOrOldest(timeAgo, smoothRangeFrom, smoothRangeTo, default)`. This will use the oldest data value when the data set is shorter than timeAgo.
- Added support for Lighting Limitless/Applamp RGBW devices. You can now set Kelvin and RGB values, NightMode, WhiteMode and increase and decrease the brightness and discoMode. See the documentation.
- Added device adapter for Onkyo receiver hardware.
- Added `scriptName` to the triggerInfo object passed as the third parameter to the execute function. This holds the name of the script being executed.
- Fixed bug in Domoticz where using forXXX() with selector switches didn't always work.
- Fixed bug in Domoticz where improper states were passed to the event scripts. This may happen on slower machines where several devices may have been updated before the event-system had a change to operate on them. In that case the event scripts received the current final state instead of the state at the moment of the actual event.
- Added support for webroot. dzVents will now use the proper API url when domoticz is started with the -webroot switch.
- Added support for event-bursts. If (on slower machines) events get queued up in Domoticz, they will be sent to dzVents in one-package. This makes event processing significantly faster.
- Added `domoticz.data.initialize(varName)`. This will re-initialize non-historical variables to the initial value as defined in the data section.
- You now no longer have to provide an initial value for a persistent variable when you declare it. So you can do `data = { 'a', 'b', 'c'}` instead of `data = {a={initial = nil}, b={initial=nil}, c={initial=nil} }`. You have to quote the names though.
- A filter can now accept a table with names/id's instead of only functions. You can now do `domoticz.devices().filter({ 'mySwitch', 'myPIR', 34, 35, 'livingLights'})` which will give you a collection of devices with these names and ids.


httpclient/HTTPClient.cpp
```diff
@@ -19,6 +19,24 @@ long		HTTPClient::m_iConnectionTimeout = 10;
 long		HTTPClient::m_iTimeout = 90; //max, time that a download has to be finished?
 std::string	HTTPClient::m_sUserAgent = "domoticz/1.0";
 
+size_t write_curl_headerdata(void *contents, size_t size, size_t nmemb, void *userp)
+{
```
gizmocuz: I think you could use the 'StringSplit' function to quickly get the vector filled ? (\n\r as delimiter)

httpclient/HTTPClient.cpp
```diff
@@ -111,15 +129,12 @@ void HTTPClient::SetUserAgent(const std::string &useragent)
 	m_sUserAgent = useragent;
 }
 
-void HTTPClient::LogError(void *curlobj)
+void HTTPClient::LogError(const long response_code)
```
gizmocuz, line 122: Could you undo this change? Now you create a lot more code that did not have to be created. This function was OK

httpclient/HTTPClient.cpp
gizmocuz: Why do you need the returned headers? Make it a VERY valid reason because the code gets dirty now.
If possible, revert this change!

gizmocuz (commented):
Thanks for the great work again!
I am not so happy with the additional header stuff in the HTTPClient class at all.
(And the status log)
Please revert those changes

httpclient/HTTPClient.cpp
dannybloe: Not sure if this is what I think it is (jvandenbroek has more details about the C++ implementation) but just to give you some background info about the http part of this PR: with this PR (dzVents 2.4) you can do a full *asynchronous* round-trip from script to Domoticz back to script with the response data when you invoke an openURL command using the commandArray.  As you may know many people struggle with this and have to use external mechanisms like curl to get some asynchronous data back into Domoticz including all the problems that could get you into like non-responsive scripts for more than 10s etc.

This PR solves this in a very neat elegant way. Just a few lines of Lua (dzVents) and you're done. Domoticz executes the url and when it's done it triggers the event-system with a callback trigger token and the response data and headers. dzVents will inspect the data and if the mime-type is application/json it will automatically convert it to a Lua table so you can directly access the response (and update your virtual devices or whatever).

Now, to make this asynchronous roundtrip more complete and not just hand over the body/data part of the script I wanted to have the response headers in the callback as well so you can check if the request was successful (inspect the status code) or perhaps grab the mime-type from the response headers to know if it is xml or json or maybe an access token for future requests etc.

Of course the world doesn't end if you don't get the response headers back to the script callback but it makes it more complete and versatile. But the least that has to be part of the callback data is the statusCode and mimeType.

Just to give you an example. A typical dzVents script to make such a round trip is this:

Here you have two triggers, one timer trigger that issues an url every 5 minutes and an httpResponse trigger that handles the url response:

```
return {
	on = {
		timer = {'every 5 seconds'},
		httpResponses = { 'mytrigger' }
	},
	execute = function(domoticz, item)
		if (item.isTimer) then
			domoticz.openURL({
				url = 'http://....',
				callback = 'mytrigger'
			})
		end
		if (item.isHTTPResponse) then
			-- we got data from the openURL call above
			if (item.headers.statusCode == 200 and item.isJSON) then						
				local t = item.json.sensors['outside'].temperature.raw
				domoticz.devices('virTemp').updateTemperature(t)
			end
		end
	end
}
```

As you can see here, `item` (second param to the script's execute function) is the full response datagram and has these attributes (excerpt from the doc):

- **data**: Raw response data.
 - **headers**: *Table*. Response headers.
 - **isJSON**: *Boolean*. Short for `response.headers['Content-Type'] == 'application/json'`. When true, the data is automatically converted to a Lua table.
 - **json**. *Table*. When the response data is `application/json` then the response data is automatically converted to a Lua table for quick and easy access.
 - **ok**: *Boolean*. `True` when the request was successful. It checks for statusCode to be in range of 200-299.
 - **statusCode**: *Number*. HTTP status codes. See [HTTP response status codes](https://developer.mozilla.org/nl/docs/Web/HTTP/Status).
 - **trigger**, **callback**: *String*. The callback string that triggered this response instance. This is useful if you have a script that is triggered by multiple different callback strings.

Of course, if passing the headers along has too many complications then we may just hand over a couple of necessary ones like statusCode and mimeType.

But I must say, this openURL-roundtripping works awesome and is so elegant. I already have a couple of those scripts that update some virtual devices like this. No more curl 'bashing' or os.executes or unresponsive scripts for more than 10s etc. Just a few lines of code and you are in business. 
This branch already runs on the machines of various testers with success.

gizmocuz: Thats for the headsup! Its clear, but i dont think you need the headers for this.
Next, this still might crash, image you do not have the credentials to access this url and you get a json back with an error message, then 'item.json.sensors['outside']' does not exists and needs to be checked.
A simple function IsJSon or IsXML could also be created.
Calling OpenURL could also just return true/false, if there is an error (status != 200) there is no need to continue parsing anyway.
When you run Domoticz in debug of have the trace enabled you already see the error/status code printed in the log.
Its just that i want to keep the HTTPClient class as clean as possible.
Also the new way of parsing the status code for logging is not great. It makes the code cluttered and additional code has to be created

dannybloe: If you do not have the permission you will get a 401 as a statuscode and item.isOk == false or item.headers.statusCode != 200. Of course you have to check that before you access the response data in your script. But even if you get a 401 there is a mime-type and only if that is json then dzVents will do the parsing (as a convenience to the user). 
So, as far as I am concerned I'm fine with handing over just the statuscode and the mime-type in the callback instead of all the headers. But I do need those two as the minimum.

gizmocuz: how about

```
if (domoticz.openURL(...)) then
 ... all is fine
end
```

httpclient/HTTPClient.cpp
jvandenbroek: Actually I tried to reduce as much as possible code, which is why I changed this part to use just the already retrieved response code (since we need this for event processing as well). Now we only use CURLINFO_RESPONSE_CODE once on each call. Could also extend the LogError function to process it for returning the code, but IMO that isn't the correct place to do it. 

I do believe the response code is very important, so one way or the other we need to include it somewhere (and I decided to put it at the end of the header return vector, to use as little as possible return vars)

gizmocuz: If a HTTP request (get/post/binary) does NOT return successfully , and only if (so there is a problem) that log function is called.
If it success, it returns TRUE... you can use this function in your dzVents to check if your call succeeded.
Next to this, for sure users will try the call in the browser and know the mime type
Next, when it should be json (call a function IsJSON for example) and you expect a temperature field, check for this field, else the call returned unwanted result.
No need for headers or mime type, just use the already returned Boolean to check if the call succeeded.
As for maybe in the feature... yes maybe in the feature

httpclient/HTTPClient.cpp
jvandenbroek: About StringSplit; I just copy/pasted and changed the already used callback function  write_curl_data_single_line which is used below. Didn't look into the possibility to use stringsplit for this, might be an option.

httpclient/HTTPClient.cpp
dannybloe: It is asynchronous so you cannot wait for the return value. OpenURL only fills the commandArray and long after the script is finished the results will be fetched by Domoticz and this will again trigger dzVents. With the callback trigger dzVents will see which script needs to be called with the response data.

httpclient/HTTPClient.cpp
dannybloe: It is asynchronous so you cannot wait for the return value. OpenURL only fills the commandArray and long after the script is finished the results will be fetched by Domoticz and this will again trigger dzVents. With the callback trigger dzVents will see which script needs to be called with the response data.

httpclient/HTTPClient.cpp
dannybloe: So if I understand it correctly, you only want to trigger dzVents with the response data only if there was a success? And if that's the case the raw response body is handed over to  dzVents and dzVents will call the callback-script with this raw data and the scripter needs to know what the mime-type is to be expected and he will convert it to json when needed himself?

jvandenbroek: Yes agree, could only parse / return status code if ret == false, but that's a minor change (moving a few lines)... However you would really only rely on return status true / false for processing in scripts? Makes things much less flexible unfortunately.

dannybloe: Yes.. it's a bit of a bummer indeed.

httpclient/HTTPClient.cpp
```diff
@@ -19,21 +19,11 @@ long		HTTPClient::m_iConnectionTimeout = 10;
 long		HTTPClient::m_iTimeout = 90; //max, time that a download has to be finished?
 std::string	HTTPClient::m_sUserAgent = "domoticz/1.0";
 
-size_t write_curl_headerdata(void *contents, size_t size, size_t nmemb, void *userp)
+size_t write_curl_headerdata(void *contents, size_t size, size_t nmemb, void *userp) // called once for each header
 {
-	// called once for each header
 	size_t realsize = size * nmemb;
 	std::vector<std::string>* pvHeaderData = (std::vector<std::string>*)userp;
```
gizmocuz: If a header line is terminated with crlf (\r\n) (which it should be according RFC 1945) then this might leave \r at the end in each vector item?
Is this for 1 header line or for all headers? as i think this will only work for the first header line

jvandenbroek: Hmm I admit only checked if new lines would occur printing them, which doesn't when stopping at \n, but you're right, stopping at \r also prevents printing them. So I'll assume it's always \r\n then (and change it to stop at \r).

It's indeed called per header.

dannybloe:
I'm still a bit confused here.. Do I have to rewrite all these http parts in dzVents not to support response headers and statusCodes or not? It's quite some work (rewriting code, changing unit-tests, integration tests, rewriting a substantial part of the new documentation and do the whole test-shebang again). So were do we stand?

gizmocuz:
Thanks again for this excellent patch !

PR #2071, opened 2018-01-15
Fix a few small bugs in the plugin framework

emontnemery:
Fix a few small bugs in the plugin framework.
Add support to modify device settings.
Corrections of inbound and outbound MQTT PUBLISH packages.

### DIFF
--- a/hardware/plugins/PluginProtocols.cpp
+++ b/hardware/plugins/PluginProtocols.cpp
@@ -115,7 +115,7 @@ namespace Plugins {
 		vData.insert(vData.end(), Message->m_Buffer.begin(), Message->m_Buffer.end());		// add the new data
 
 		std::string		sData(vData.begin(), vData.end());
-		int iPos = sData.find_first_of('\r');		//  Look for message terminator 
+		int iPos = sData.find_first_of('\r');		//  Look for message terminator
 		while (iPos != std::string::npos)
 		{
 			onMessageCallback*	RecvMessage = new onMessageCallback(Message->m_pPlugin, Message->m_pConnection, std::vector<byte>(&sData[0], &sData[iPos]));
@@ -124,7 +124,7 @@ namespace Plugins {
 				PluginMessageQueue.push(RecvMessage);
 			}
 
-			if (sData[iPos + 1] == '\n') iPos++;				//  Handle \r\n 
+			if (sData[iPos + 1] == '\n') iPos++;				//  Handle \r\n
 			sData = sData.substr(iPos + 1);
 			iPos = sData.find_first_of('\r');
 		}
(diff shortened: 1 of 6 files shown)


emontnemery:
Sorry about the number of commits, let me know if I should squash them.

hardware/plugins/PluginProtocols.cpp
```diff
@@ -883,169 +889,202 @@ namespace Plugins {
...
 				}
-			}
-			else
+			} while ((encodedByte & 128) != 0);
+
+			if (iRemainingLength > std::distance(it, m_sRetainedData.end()))
 			{
-				ADD_INT_TO_DICT(pMqttDict, "Status", -1);
-				ADD_UTF8_TO_DICT(pMqttDict, "Description", std::string("Invalid message length"));
+				_log.Log(LOG_TRACE, "(%s) Too little data (got %u, expected %u).", __func__, std::distance(it, m_sRetainedData.end()), iRemainingLength);
```
gizmocuz: Is it possible to rephrase this to 'Not enough data received' instead of 'Too little data' ?

gizmocuz (commented):
Thanks for this !!! Just one very minor request....

hardware/plugins/PluginProtocols.cpp
emontnemery: Yes, sure!  Should I squash the commits as well?

gizmocuz:
Thanks for the change! But are you sure the do->while(0) is needed ? Made a small test application and runs/compiles without issues:

```
#include <string>
#include <vector>

#define MQTT_STRING_W_LEN(vVector, sString)	\
 { \
  	int sLen = sString.length(); \
	for (int i = 0; i<sLen; i++) vVector.push_back(sString[i]); \
 }

int main()
{
	std::string szTest("Hello World!");
	std::vector<uint8_t> result;
	MQTT_STRING_W_LEN(result, szTest);
    return 0;
}

```

I would favor that your macro's follow the standard parameter method of Input/Output as now it is Output/Input

gizmocuz:
length() returns a size_t object...

gizmocuz:
I am not 100% in the code, but to convert a string to a vector you could simple use

`std::vector<uint8_t> result2(szTest.begin(), szTest.end());`

hardware/plugins/PluginProtocols.cpp
gizmocuz: No need, i could smash it as we merge ;)

emontnemery:
The purpose of the do { .. } while (0) is to ensure the macro can be used the same way a function call with no special cases. Here's a good summary explaining why it's necessary: https://gcc.gnu.org/onlinedocs/cpp/Swallowing-the-Semicolon.html

Agreed on the other comments, I'll update.

Edit:
Changed to functions instead of macros because macro expansion means multiple instances of passed string literals will be constructed, so the begin() and end() iterators will actually refer to different strings.

gizmocuz:
Thanks!

PR #2076, opened 2018-01-19
Enable type checking on CLogger::Log and fix warnings

emontnemery:
Enable gcc printf type checking on CLogger::Log and fix warnings
Most warnings were harmless but there were also a few which would cause undefined behavior / crash if printed.

Note:
Replacing this:
_log.Log(LOG_STATUS, sLine.c_str());
With:
_log.Log(LOG_STATUS, "%s", sLine.c_str());
Is necessary to ensure any % characters included in the string will not be parsed by the print function.


correa:
I guess it would be better to make an overload of Log without the variadic argument. The overload could take std::string and in that case there would be fewer conversions.

emontnemery:
@correa, done!

hardware/plugins/PluginProtocols.cpp
```diff
@@ -282,28 +282,28 @@ namespace Plugins {
 		}
 	}
 
-#define ADD_BYTES_TO_DICT(pDict, key, value) \
-		{	\
-			PyObject*	pObj = Py_BuildValue("y#", value.c_str(), value.length());	\
-			if (PyDict_SetItemString(pDict, key, pObj) == -1)	\
-				_log.Log(LOG_ERROR, "(%s) failed to add key '%s', value '%s' to dictionary.", __func__, key, value.c_str());	\
-			Py_DECREF(pObj); \
+	static void AddBytesToDict(PyObject* pDict, const char* key, const std::string& value)
```
gizmocuz: Could you undo these changes as they are not about the logging ? If needed, please create a separate pull request

gizmocuz: Probably all changes in this class

emontnemery:
@gizmocuz Thanks for the review! I pushed a new commit where the unnecessary macro-related changes are reverted.

gizmocuz:
Thanks for this great patch !

PR #2079, opened 2018-01-20
Change icon for Mobile Items in the menu to phone icon

flatsiedatsie:
There is a phone icon. So let's use that icon.


gizmocuz:
Thanks for the patch, but it should be 'Phone48_On.png', i don't know how the other icon got there, but it seems i need to do some cleanup in the images folder ;)
Could you also increment the version number in the appcache file if you make changes to the www folder?

gizmocuz:
No reaction, closing topic

PR #2075, opened 2018-01-19
New attempt to enable C++11 build and fix build errors

emontnemery:
Enable c++11 support, builds on:
  - Ubuntu Trusty (through Travis)
  - OSX + XCode 8.3 (Through Travis)
  - VS2017 (Through appveyor)
  - Debian Jessie (x64, but needs CMake from Jessie backports)
  - Raspbian Stretch (RPi 3)
  - Ubuntu WSL (x64)

CMake minimum version stepped, CMake 3.1.0 was first version to include support for CMAKE_CXX_STANDARD

PR #2081, opened 2018-01-20
Enable C++11 build

correa:
Tested on:
* Raspbian Jessie (cmake 3.0.2, gcc 4.9.2)
* Windows (MSVC 2017) 
* macOS Sierra (cmake 3.9.1, clang 9.0.0)
* Ubuntu 16.04.03 LTS (cmake 3.5.1, gcc 5.4.0)

Raspbian Wheezy starts building but we hit an internal compiler error on EventSystem.cpp, current development branch also causes a internal compiler error on SQLHelper.cpp so Jessie is currently minimum.


emontnemery:
I've tested and your changes to CMakeLists.txt unfortunately don't work for the precompiled header because flage std=c++11 is set when compiling pch instead of flag std=gnu++11
Edit: The warnings are seen also in the travis Linux build
```
[  0%] Generating stdafx.h.gch/.c++
/usr/bin/c++ -O3 -DNDEBUG -I/usr/include/python3.5m -I/mnt/d/development/github/domoticz_correa/main -I/mnt/d/development/github/domoticz_correa/tinyxpath -I/usr/include -I/usr/include -I/usr/include -I/usr/include -DBUILTIN_MQTT -DENABLE_PYTHON -DHAVE_EXECINFO_H -DHAVE_LINUX_I2C -DHAVE_LINUX_SPI -DWWW_ENABLE_SSL -DWITH_LIBUSB -std=c++11 -x c++-header -o /mnt/d/development/github/domoticz_correa/build/stdafx.h.gch/.c++ /mnt/d/development/github/domoticz_correa/main/stdafx.h
make[2]: Leaving directory '/mnt/d/development/github/domoticz_correa/build'
```
```
[ 20%] Building CXX object CMakeFiles/domoticz.dir/main/stdafx.cpp.o
/usr/bin/c++   -DBUILTIN_MQTT -DENABLE_PYTHON -DHAVE_EXECINFO_H -DHAVE_LINUX_I2C -DHAVE_LINUX_SPI -DWITH_LIBUSB -DWWW_ENABLE_SSL -I/usr/include/python3.5m -I/mnt/d/development/github/domoticz_correa/main -I/mnt/d/development/github/domoticz_correa/tinyxpath  -O3 -DNDEBUG   -include stdafx.h -Winvalid-pch -std=gnu++11 -o CMakeFiles/domoticz.dir/main/stdafx.cpp.o -c /mnt/d/development/github/domoticz_correa/main/stdafx.cpp
cc1plus: warning: ./stdafx.h.gch/.c++: not used because `__STRICT_ANSI__' not defined [-Winvalid-pch]
[ 20%] Building CXX object CMakeFiles/domoticz.dir/main/CmdLine.cpp.o
/usr/bin/c++   -DBUILTIN_MQTT -DENABLE_PYTHON -DHAVE_EXECINFO_H -DHAVE_LINUX_I2C -DHAVE_LINUX_SPI -DWITH_LIBUSB -DWWW_ENABLE_SSL -I/usr/include/python3.5m -I/mnt/d/development/github/domoticz_correa/main -I/mnt/d/development/github/domoticz_correa/tinyxpath  -O3 -DNDEBUG   -include stdafx.h -Winvalid-pch -std=gnu++11 -o CMakeFiles/domoticz.dir/main/CmdLine.cpp.o -c /mnt/d/development/github/domoticz_correa/main/CmdLine.cpp
cc1plus: warning: ./stdafx.h.gch/.c++: not used because `__STRICT_ANSI__' not defined [-Winvalid-pch]
```

Same output from #2075:
```
[  1%] Generating stdafx.h.gch/.c++
/usr/bin/c++ -O3 -DNDEBUG -I/usr/include/python3.5m -I/mnt/d/development/github/domoticz_emontnemery/main -I/mnt/d/development/github/domoticz_emontnemery/tinyxpath -I/usr/include -I/usr/include -I/usr/include -I/usr/include -DBUILTIN_MQTT -DENABLE_PYTHON -DHAVE_EXECINFO_H -DHAVE_LINUX_I2C -DHAVE_LINUX_SPI -DWWW_ENABLE_SSL -DWITH_LIBUSB -x c++-header -std=gnu++11 -o /mnt/d/development/github/domoticz_emontnemery/build/stdafx.h.gch/.c++ /mnt/d/development/github/domoticz_emontnemery/main/stdafx.h
make[2]: Leaving directory '/mnt/d/development/github/domoticz_emontnemery/build'
```
```
[ 20%] Building CXX object CMakeFiles/domoticz.dir/main/stdafx.cpp.o
/usr/bin/c++   -DBUILTIN_MQTT -DENABLE_PYTHON -DHAVE_EXECINFO_H -DHAVE_LINUX_I2C -DHAVE_LINUX_SPI -DWITH_LIBUSB -DWWW_ENABLE_SSL -I/usr/include/python3.5m -I/mnt/d/development/github/domoticz_emontnemery/main -I/mnt/d/development/github/domoticz_emontnemery/tinyxpath  -O3 -DNDEBUG   -include stdafx.h -Winvalid-pch -std=gnu++11 -o CMakeFiles/domoticz.dir/main/stdafx.cpp.o -c /mnt/d/development/github/domoticz_emontnemery/main/stdafx.cpp
[ 20%] Building CXX object CMakeFiles/domoticz.dir/main/CmdLine.cpp.o
/usr/bin/c++   -DBUILTIN_MQTT -DENABLE_PYTHON -DHAVE_EXECINFO_H -DHAVE_LINUX_I2C -DHAVE_LINUX_SPI -DWITH_LIBUSB -DWWW_ENABLE_SSL -I/usr/include/python3.5m -I/mnt/d/development/github/domoticz_emontnemery/main -I/mnt/d/development/github/domoticz_emontnemery/tinyxpath  -O3 -DNDEBUG   -include stdafx.h -Winvalid-pch -std=gnu++11 -o CMakeFiles/domoticz.dir/main/CmdLine.cpp.o -c /mnt/d/development/github/domoticz_emontnemery/main/CmdLine.cpp
```
Also, please fix indentation of CMakeLists.txt it becomes really hard to read the logic with mixed spaces/tabs.

gordonb3:
I don't really get the ;point of this update. I have been compiling Domoticz on gcc 4.9.x for ages. Since about a half year I'm building on gcc 5.4. The only issue I have is that the project favors static linking on some external libraries, which causes linker errors on my system.

emontnemery:
@gordonb3 c++11 has a number of really nice features, e.g. lambda expressions and auto keyword.

gordonb3:
Okay, but as far I'm aware gcc 5.4 will default to c++11 anyway, or is there something in the code that will make it fall back to the older ABI?

correa:
@gordonb3 It's good to know 5.4 will enable by default. Most other platforms do as well, but we still need to support older compilers, specially for the embedded OSes like the Raspbian. Unless we were to decide to drop Raspbian Jessie and go straight to Stretch as minimum required, then we would get gcc 6.3

Edit: Tried gcc 5.4 without the cmake changes but keeping the C++11 features in code, and it didn't build. Maybe it's a later version of gcc that enables C++11 by default

emontnemery:
@gordonb3 I tested by changing some code to use `auto` keyword introduced in c++11, and gcc 5.4 does not default to c++11, `auto` keyword is only accepted if -std=c++11 is added to gcc's commandline.

Edit: I didn't see @correa already tested, sorry.

Test:
```
$ gcc --version
gcc (Ubuntu 5.4.0-6ubuntu1~16.04.5) 5.4.0 20160609
```

```
diff --git a/hardware/plugins/PluginProtocols.cpp b/hardware/plugins/PluginProtocols.cpp
index c33bc2e..83f3a22 100644
--- a/hardware/plugins/PluginProtocols.cpp
+++ b/hardware/plugins/PluginProtocols.cpp
@@ -891,7 +891,8 @@ namespace Plugins {
                m_sRetainedData.insert(m_sRetainedData.end(), Message->m_Buffer.begin(), Message->m_Buffer.end());

                do {
-                       std::vector<byte>::iterator it = m_sRetainedData.begin();
+                       auto it = m_sRetainedData.begin();
```

```
[ 20%] Building CXX object CMakeFiles/domoticz.dir/hardware/plugins/PluginProtocols.cpp.o
/usr/bin/c++   -DBUILTIN_MQTT -DENABLE_PYTHON -DHAVE_EXECINFO_H -DHAVE_LINUX_I2C -DHAVE_LINUX_SPI -DWITH_LIBUSB -DWWW_ENABLE_SSL -I/usr/include/python3.5m -I/mnt/d/development/github/domoticz_fork/main -I/mnt/d/development/github/domoticz_fork/tinyxpath  -O3 -DNDEBUG   -include stdafx.h -Winvalid-pch -o CMakeFiles/domoticz.dir/hardware/plugins/PluginProtocols.cpp.o -c /mnt/d/development/github/domoticz_fork/hardware/plugins/PluginProtocols.cpp
/mnt/d/development/github/domoticz_fork/hardware/plugins/PluginProtocols.cpp: In member function ‘virtual void Plugins::CPluginProtocolMQTT::ProcessInbound(const Plugins::ReadMessage*)’:
/mnt/d/development/github/domoticz_fork/hardware/plugins/PluginProtocols.cpp:895:9: error: ‘it’ does not name a type
    auto it = m_sRetainedData.begin();
         ^
/mnt/d/development/github/domoticz_fork/hardware/plugins/PluginProtocols.cpp:897:20: error: ‘it’ was not declared in this scope
    byte  header = *it++;
                    ^
CMakeFiles/domoticz.dir/build.make:4142: recipe for target 'CMakeFiles/domoticz.dir/hardware/plugins/PluginProtocols.cpp.o' failed
```

gordonb3:
Possibly depends on platform then: https://www.gentoo.org/support/news-items/2015-10-22-gcc-5-new-c++11-abi.html

A warning appears to be in order here though, as it seems that c++11 ABI is not stable in pre-gcc5. Articles describing issues [here](https://developers.redhat.com/blog/2015/02/05/gcc5-and-the-c11-abi/) and [here](https://blogs.gentoo.org/blueness/2015/03/10/the-c11-abi-incompatibility-problem-in-gentoo/). Distributing binaries may become a problem.

emontnemery:
Aha, we were talking about different things then (language features vs ABI).
@gizmocuz indicated it would be enough to support Jessie and newer distributions (if I understood him correctly), and Jessie has gcc 4.9.2 Do you know when the ABI issues were solved? Your links describe that there was an issue almost 3 years ago, but not when or how the issue was solved.

gordonb3:
I don't think 4.9.x will do it. I remember when I upgraded to gcc5.4 I had to do a rebuild of every c++ based application (Gentoo packages build from source) as they displayed messages about missing symbols in the new glibc. On my main workstation I was recently upgraded to gcc6.4 and there were no compatibility issues with this version and code created by gcc5.4.

emontnemery:
OK, but we're not asking anyone to upgrade the compiler, so I don't think that problem is relevant...
Anyway, as stated in #2075 this builds - and works - on Debian Jessie which is with gcc 4.9.2
Also works on raspberry pi, although i've only verified on Raspbian Stretch.

gordonb3:
I probably should express myself more clearly. The problem is not so much with gcc but with glibc. Glibc is supposed to be backwards compatible, so in essence if you have a binary that is compiled with gcc 4.4 it will run with any version of glibc that is equal or higher than 2.17. This however proves to be untrue for applications that were compiled using the c++11 ABI from gcc version 4.7 up to 5.x. Effectively: if you compile a binary using gcc 4.9.2 I will not be able to run it on my system that has glibc 2.25. And since I am assuming that the standard build environment used for this project will be using an older version of gcc for compatibility with older systems you will also not be able to run its binary with your version of glibc (2.23?) if that binary was compiled using the c++11 ABI.

emontnemery:
OK, you're talking about distributing offical binaries, for example the ones available on https://domoticz.com/downloads/?
Those are compiled with gcc 6.3.0, would the problem you desribe occur then?
```
$ wget https://releases.domoticz.com/releases/beta/domoticz_linux_x86_64.tgz
$ tar xzf domoticz_linux_x86_64.tgz
$ strings domoticz | grep gcc
__gcclibcxx_demangle_callback
COMPILER=gcc-6.3.0 20170519
.gcc_except_table
__gcclibcxx_demangle_callback
```

gordonb3:
You're kidding?

Yes that would most definitely pose an issue. I'm amazed even that nobody has complained that the current distributed binaries don't run on their system. I'm assuming right now that this is because there hasn't been any change to the c++03 ABI since.

correa:
As far as I know there is no good answer for distributing binaries in Linux. It's not just libc and Domoticz builds against static libc by default. I have tried before as a commercial software developer and it's really not worth it. Open source software is usually distributed as source, even embedded is distributed as source and it takes several hours to build on those platforms when not cross-compiling.

correa:
AppImage might be a solution for Linux binaries.

emontnemery:
I agree there is no good answer for distributing binaries in Linux..
Anyhow, @gizmocuz, what versions of Linux are intended to be supported by the official binaries?

gizmocuz:
@emontnemery 
Armv6/v7 and x86_x64

Do note that this was working before, but since this openssl/python issue it is broken on quite some systems.
I know distributing binaries is not the right way, but for many users this is easy and the only way.

Before, years ago, i always made a stripped down version of raspbian for the RPi users and they always had to download 300MB. Those days are over (on the stable), but some have issues using the new 'stretch' platform and have issues with openssl libraries.

Hope it will be solved soon.

Next to distributing the binaries, the build systems are good to use for trouble shooting as well

emontnemery:
@gizmocuz Sure, but what versions of which distributions? The potential issue described by @gordonb3 might for example be an issue with debian wheezy (and derivates) which was released in 2013 and other distributions of similar age.
What is the "openssl/python issue" btw?

gordonb3:
I think you may have missed an earlier point I made. GCC does not guarantee C++ 11 ABI stability prior to version 5.1. That means that Jessie is affected as well, unless the distributed binary was created with the exact same compiler version.

emontnemery:
@gordonb3 No I didnt't miss that, as pointed out by @correa domoticz links statically with libc. I tested a binary built by travis (so it's the environment used to build official binaries, using gcc 6.3.0) with the changes from this PR and it runs just fine on Debian Jessie (although I only did some quick testing). I can make it available if you're interested?

gordonb3:
Actually, it doesn't. Domoticz links statically to boost an openssl, which means that it also links statically to every other dependency, including glibc. Domoticz does however also link dynamically to other libraries, including at least curl which in turn dynamically links to openssl and is the reason why you can't run Domoticz without having the correct version of openssl installed. And of course this also links glibc dynamically. As such it will depend on the functions you're using whether the application will crash on non matching symbols or not.

emontnemery:
@gordonb3 OK, thanks for explaining! Python is also (manually) linked dynamically btw.

gizmocuz:
Currently not all platforms can compile with these changes.
Hope to be able to convert soon.
As the discussion continues about linking, it is best to open an issue for this

Issue #1202, opened 2017-02-04
Philips Hue White group shows as RGBW


gielie:
#887 

problem still there, running latest beta.

Individual white light bulbs are correct but the group shows as rgbw.

Any idea how to solve this.

gizmocuz:
Probably with a code patch

Do the groups work ?

gielie:
Yes, groups works fine, but they are also shown as rgbw while the bulps are white ambiance

gizmocuz:
If you edit the group... can you set the type to 'Push On' ? Does it still work then ?

gizmocuz:
Maybe best to send me a PM on the forum, so i can ask you for some debug output

gielie:
Any news on this topic?

jjnj023:
If there is any need for testing i can help.

mattijsf:
I'm running V3.8276 and I still can't get the Hue groups with Hue Ambiance lights to work as a proper dimmer. It always recognizes the group as RGBW while a single light bulb from the group get a proper dimmers slider. The group is newly created and added after I updated to 3.8276.

The group only contains Hue Ambiance lights. I would expect it was possible to add the group and get the same slider / dim functionality as I would have when controlling the single light bulb.

![screen shot 2017-08-12 at 13 58 27](https://user-images.githubusercontent.com/706368/29240469-62582cea-7f66-11e7-91d1-d670c312e802.png)

While dimming works with the popup the UI is much less convenient when using the dashboard to control lights.

gizmocuz:
If you "Still" can not get it to work, maybe you should do a little more work and submit a patch ?

Nenodema:
Are there any updates on this bug/feature? My "older" hue color lights are also recognized as RGBW and not as AC.

I have 6 lights, 2 new ones (hue white Ambiance) and 4 color lights. The new ones are created with AC and dimmable but the older ones are recognized as RGBW lights.

![image](https://user-images.githubusercontent.com/33698556/33657736-7c590ee2-da7a-11e7-8650-fbfc76e96c82.png)

PR #2082, opened 2018-01-21
Fix bug where Philips Hue white group shows as RGBW

asteffey:
* Addresses issue #1202

### DIFF
--- a/hardware/PhilipsHue/PhilipsHue.cpp
+++ b/hardware/PhilipsHue/PhilipsHue.cpp
@@ -144,7 +144,7 @@ bool CPhilipsHue::WriteToHardware(const char *pdata, const unsigned char length)
 	if (packettype == pTypeLighting2)
 	{
 		//light command
-		nodeID = pSen->LIGHTING2.id4;
+		nodeID = (pSen->LIGHTING2.id3 << 8) + pSen->LIGHTING2.id4;
 		if ((pSen->LIGHTING2.cmnd == light2_sOff) || (pSen->LIGHTING2.cmnd == light2_sGroupOff))
 		{
 			LCmd = "Off";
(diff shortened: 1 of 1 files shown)


hardware/PhilipsHue/PhilipsHue.cpp
```diff
@@ -462,7 +462,7 @@ void CPhilipsHue::InsertUpdateSwitch(const int NodeID, const _eHueLightType LTyp
 	{
 		//Send as Lighting 2
 		char szID[10];
-		sprintf(szID, "%X%02X%02X%02X", 0, 0, 0, NodeID);
+		sprintf(szID, "%07X", (unsigned int)NodeID);
```
gizmocuz, line 465: Thanks for the patch!, Should this perhaps be %08X ? (as we use this a lot)

asteffey: For some reason Lighting 2 uses a 7 character DeviceID instead of 8 characters (as other hue light types do).  Things didn't work correctly when I used %08X.  Further, `sprintf(szID, "%X%02X%02X%02X", 0, 0, 0, NodeID);` seems to be used quite often to generate a 7 character DeviceID for Lighting 2 elsewhere in the code (e.g. [FritzboxTCP.cpp](https://github.com/domoticz/domoticz/blob/development/hardware/FritzboxTCP.cpp),  [Sterbox.cpp](https://github.com/domoticz/domoticz/blob/development/hardware/Sterbox.cpp), [RelayNet.cpp](https://github.com/domoticz/domoticz/blob/development/hardware/RelayNet.cpp)), so %07X does seem to be right, at least for Lighting 2.

The reason I changed it from  `sprintf(szID, "%X%02X%02X%02X", 0, 0, 0, NodeID);` to `sprintf(szID, "%07X", (unsigned int)NodeID);` was because the ID of Phillips Hue groups require 2 bytes instead of one.

gizmocuz:
Thanks for the patch!!

PR #2083, opened 2018-01-21
Notify plugin framework when device is changed

emontnemery:
**Background:**
The plugin framework has a local copy of rows in DeviceStatus, which is initialized when Domoticz is started.
This obviously causes problems if a device owned by a plugin is added, modified or deleted.

**Proposed change:**
I've added (working) code which will handle the case of deletion of plugin owned device from WebUI.

Not ready for merge, I hope to get some feedback first:
- The method added to CPluginSystem is not strictly necessary. The intention is to avoid duplication of code. Let me know if it should be removed or rewritten.
- I only notify the **plugin framework** to trigger a refresh, should the actual plugin also be notified?
- The name of new messages (PluginMessages.h) as well as new methods (Plugins.h) are probably not good.

Code to handle modification / addition of plugin devices will also be added after taking comments/suggestions into account..


dnpwwo:
I would like to understand how these will work from a plugin author perspective.

If you look at https://www.domoticz.com/wiki/Developing_a_Python_plugin#Callbacks there is a description of the callback and it would be useful to understand these callbacks to the same level.  When it will be called, what will be handed in to the plugin and what it allows the plugin author to do.

For example, the onDeviceRemoved callback:
- Is it called before or after the Device is deleted?
- Can the plugin prevent deletion be returning a value?
- Will the Plugin Framework handle all plugin Device deletion going forward?
- Is the Device object handed in as a parameter (which would align with the other callbacks)?
- Is the Device still in the Devices dictionary or has it been removed?
- If its still in the Devices dictionary should the plugin remove it or will the framework do after?

I had some question around the code but validating the functional design is right is more important

jvandenbroek:
Use #ifdef ENABLE_PYTHON where applicable.

emontnemery:
@dnpwwo, @jvandenbroek  Thanks a lot for the comments!!

@dnpwwo, what I had in mind was something much more simple.

For deletion of device:
- Inform **plugin framework** after deletion of device
  - Plugin framework would simply remove the device from `CPlugin::m_DeviceDict`
  - Plugin itself will not be informed 

If you think there is a benefit of informing the plugin, I would suggest an implementation like this based on your questions:
- Is it called before or after the Device is deleted?
  - Asynchronous call with no response, after deletion of device (to not keep Domoticz main thread waiting).
- Can the plugin prevent deletion be returning a value?
  - No, but it can of course re-add the device if needed.
- Will the Plugin Framework handle all plugin Device deletion going forward?
  - Sorry, I'm not sure what you mean..
- Is the Device object handed in as a parameter (which would align with the other callbacks)?
  - Yes. A device object based on cached data already in `CPlugin::m_DeviceDict` would be passed to the plugin
- Is the Device still in the Devices dictionary or has it been removed?
  - The device has been removed from Devices dictionary when callback is called.
- If its still in the Devices dictionary should the plugin remove it or will the framework do after?
  - Already removed, see above

Similar for modification / addition of device.

dnpwwo:
Seems okay.

One thing I would suggest is that rather that creating CPluginSystem::DeviceRemoved you make a CPluginSystem::MessagePlugin(CPluginMessageBase*) function that would allow for other messages to be added in the future with less change.

guillaumezin:
And as suggested by @dnpwwo  we can share our efforts (cf. #2013) to propose a common callback, I will propose something when I'll have time

emontnemery:
@guillaumezin My suggestion is to add three new callbacks, DeviceAdded, DeviceModified, DeviceRemoved, whereas you seek to extend the existing onCommand callback.
As commented by @dnpwwo  in #2013, with the suggested "DeviceModified" callback the plugin would be expected to refresh information from Domoticz database by calling Device.Refresh - this is where state is store after all - and then take action as needed.
I don't think it's very logical to handle all of this through an extended "onCommand" callback.
But I do think we should try to share our efforts!

guillaumezin:
See if you can integrate what you need with the onUpdate callback I created in #2013 , I already merged your pull request with mine 2 days ago, have a look.

emontnemery:
I updated to try to take comments from @dnpwwo into account. 
Adding calls to the new callbacks from Domoticz is not done pending review (onDeviceAdded / onDeviceModified not yet called from anywhere, onDeviceRemoved needs to be called from a few additional places in addition to from SQLHelper).
@guillaumezin I still kept onDeviceAdded / onDeviceModified / onDeviceRemoved as three separate callbacks, but it can of course be merged into just onCommand if @dnpwwo prefers it that way, it doesn't really matter to me.

Implementation is now for "onDeviceRemoved":
- Is it called before or after the Device is deleted?
  - Asynchronous call with no response, after deletion of device from mySQL database (to not keep Domoticz main thread waiting).
- Can the plugin prevent deletion be returning a value?
  - No, but it can of course re-add the device if needed.
- Is the Device object handed in as a parameter (which would align with the other callbacks)?
  - I reconsidered this, and the answer is now "no" to align with onCommand, just the Unit number is passed. I don't think a Device object is passed to any other callback either?
- Is the Device still in the Devices dictionary or has it been removed?
  - The device is still in Devices dictionary when callback is called, and will be removed immediately after it returns
- If its still in the Devices dictionary should the plugin remove it or will the framework do after?
  - The framework will remove it.

Regarding this:

> rather that creating CPluginSystem::DeviceRemoved you make a CPluginSystem::MessagePlugin(CPluginMessageBase*) function that would allow for other messages to be added in the future with less change.

The purpose of adding methods to CPluginSystem was to avoid duplication of code to first find the plugin and send messages to it. Since constructing a CPluginMessageBase object requires a pointer to the plugin, there would be no such benefit with a function CPluginSystem::MessagePlugin(CPluginMessageBase*) I think?

emontnemery:
@dnpwwo Do you have any further comments, or is what i pushed in my last commit OK?

dnpwwo:
Didn't really look at the SQLHelper part but the plugin components look good.

Only comment would be on what is handed to the plugin, for add and modify wouldn't it be friendlier to pass the actual device object rather than the unit number?

emontnemery:
@dnpwwo I cleaned up and also added the changes from #2013, hopefully ready for merge now.

dnpwwo:
Well now I'm a bit confused because I don't understand the difference between onDeviceModified an onUpdate. nValue and sValue are part of the device so if one gets modified (or is that updated) which callback fires (or is it both?)?

I thought it was better before #2013 was merged in because onDeviceModified could be called if **anything** about the device was modified outside the plugin's control. It's easy to document and easy for plugin authors to understand.

I would roll #2013 back and recommend to @gizmocuz that we merge that

emontnemery:
@dnpwwo, OK, I'll rollback the changes from #2013, I agree onUpdate is a less general case of onDeviceModified.

@guillaumezin, this call in MainWorker::UpdateDevice:
```
((Plugins::CPlugin*)m_hardwaredevices[hindex])->SendCommand(unit, "udevice",  static_cast<float>(atof(sValue.c_str())));
```
Will be be replaced with DeviceModified:
```
((Plugins::CPlugin*)m_hardwaredevices[hindex])->DeviceModified(unit);
```
That works for you, right?

guillaumezin:
How will the plugin get new values (sValue and nValue) with DeviceModified ?

emontnemery:
@dnpwwo I pushed a new commit.
@guillaumezin When you get the onDeviceModified callback in your plugin, the updated values are in the devices dictionary.
Something like:
```
def onDeviceModified(self, Unit):
    Domoticz.Debug("onDeviceModified Unit: "+str(Unit) + ", nValue: " + str(Devices[Unit].nValue) + ", sValue: " + Devices[Unit].sValue));
```

guillaumezin:
Ok, so it should work for my use case

emontnemery:
Yeah, I hope so.

By the way, with your changes in MainWorker::UpdateDevice committed here: e5e6a3f4fdd1ea038eb72a1ee66dc7d887b76234, you simply pass nValue and sValue to the plugin and then return.
I changed that to trigger the onDeviceModified callback at the very end, when the database has already been updated which is how it works for all other hardware.
Is there any reason why that won't work and you notify the plugin first?

guillaumezin:
This is historical: first I tried to extend on Command callback and pass values as arguments, but it was difficult to keep backward compatibility, then came onUpdate to pass sValue and nValue in a dict as argument, I didn't even had the idea to re-read them from Devices, so I didn't try to put the callback after database update

guillaumezin:
Hi, shouldn't CDevice_Refresh() be called before onDeviceModified (and maybe onDeviceAdded)?

emontnemery:
@guillaumezin that's done in onDeviceModifiedCallback::ProcessLocked handler, first m_pPlugin->onDeviceModified() is called which will in turn call CDevice_refresh, then the python callback onDeviceModified will be called.
```
class onDeviceModifiedCallback : public CCallbackBase
{
public:
	onDeviceModifiedCallback(CPlugin* pPlugin, int Unit) : CCallbackBase(pPlugin, "onDeviceModified") { m_Unit = Unit; };
protected:
	virtual void ProcessLocked()
	{
		m_pPlugin->onDeviceModified(m_Unit);
		PyObject*	pParams = Py_BuildValue("(i)", m_Unit);
		Callback(pParams);
	};
};
```

guillaumezin:
Sorry, that's obvious, I didn't see it, I must be tired...

emontnemery:
@guillaumezin No worries! Do you think the Python onDeviceModified callback is OK as is, or should it be modified such that a dictionary naming changed parameters is also passed in?

guillaumezin:
No, if sValue and nValue are refreshed and available in Devices dict, it is enough

guillaumezin:
@emontnemery What I didn't check is if onCommand AND onDeviceModified will be called for a single change, it could be a problem (plugin may receive a change via 2 differents callbacks), what do you think?

emontnemery:
@dnpwwo, @gizmocuz Is this OK to be merged, or do you want some changes?
My own remaining concern is if plugins should be blocked from implicitly adding devices by calls to CSQLHelper::UpdateValue, I think it's a bug in the plugin framework if that ever happens.
See comment on line 3937 in SQLHelper.cpp.

dnpwwo:
@emontnemery, from the comments above I thought the decision was not to have onUpdateCallbacks but they still seem to be in the code when I look at "Files Changed" are you going to remove them?

emontnemery:
@dnpwwo You're absolutely right, I forgot to push the last commit. Now it should be correct..

dnpwwo:
@gizmocuz I think this can bemerged, its new functionality so any issues can be shaken out without impact

PR #2085, opened 2018-01-21
OpenWebNet: active power and energy 2

Spugna85:
management of the active power and energy obtained from the F520, F521, F523 devices


gizmocuz:
I am not to happy about letting the user specify a rate-limit. 
It seems 300 seconds (5 minutes) is the minimum (and the default), and i think this is OK?
What could be the reason to make this larger?

Spugna85:
The openwebnet protocol is based on spontaneous updates of devices status via a monitor connection. However, it happened to lost some updates. The scanning of the state of the devices is done at the hardware start, but is repeated at intervals of minimum 300 seconds to perform a refresh of everything, this because a shorter time could create traffic problems on the scs bus. Nothing forbids to put a longer time or to disable the scan.
I apologize for making a copy paste, this is not a rate-limit but a scan interval.

gizmocuz:
Thanks!

Spugna85:
I can not see the change in the beta version
can you help me to understand why the merge has failed? I do not understand..

Issue #1955, opened 2017-11-20
Memory leak?


heggink:
From oct 20th to Nov 9th, my memory usage for the domoticz process was VERY stable. As of Oct 9th, it's starting to gradually go up. Not as bad as with the Blockly leak but around 1.7% per day. I updated a couple of times to new beta's (currently on 8740) but the increase remains (last update 17th).
![domo_mem](https://user-images.githubusercontent.com/3987735/33038088-ce53ee5e-ce33-11e7-8c25-bca609499ece.jpg)

http://www.domoticz.com/forum/viewtopic.php?f=6&t=19186&p=158521#p158521 references this although I am not sure this is a fair representation of facts (too many different betas).

Siewert308SW:
Could it be that something has been changed so the old fix for has been killed.
https://github.com/domoticz/domoticz/commit/d7841bbfc83e7ebf951eb708a92297f237ca992e

Struggling with this issues for a longtime until this https (ssl) thingy popped up in my mind.
To check if https is the issue again i disabled it a few hours ago.
It is to early to say but maybe this issues popped up again but for now the memory curve looks stable but will report tomorrow about my findings.
![chart_mem1](https://user-images.githubusercontent.com/25230393/33095419-e6a13a68-cf03-11e7-832a-70a3c27f4f49.png)

Siewert308SW:
Over a period of 12hr memory increased from 7.1 after reboot to 7.55%.
So it didn't increased that fast before i disabled https.
Still to early i think to make any conclusions, for now i will let Domoticz run and see if it stabilizes. 
Will report after work tonight...

Siewert308SW:
Quick report,
Memory has increased since 6am this morning from 7.55% to 7.98% thats aprox 10hr
Mean while i'm installing the latest Domoticz beta (same as daily driver) on my 2th Pi3.
Will let it run clean without scripts and such to see how that goes...

LemarinelNet:
Here is my contribution after almost 48H of memory tracking.
I'm on DEBUG build, and the memory usage is also growing, but it seems to grow by step every hour !?

![domoticzmemory](https://user-images.githubusercontent.com/10656218/33162142-206e2308-d028-11e7-8e23-77159578a1f9.png)

I'm letting this continue to grow, and will report later.

Edit : The first "big" grow on 21/11 20h00 is because of activation of xiaomi-robot plugin. I'm trying to disable it to check if memory continues to grow or settles...

[...]

PR #2087, opened 2018-01-23
EventSystem/Blockly reduced/optimized code

jvandenbroek:
- Renamed EvaluateBlockly to EvaluateDatabaseEvents
- Moved/merged DB events (m_events) evaluation to EvaluateDatabaseEvents (now m_events is iterated only once instead of twice per event)
- Moved redundant code from EvaluateBlockly to separate ParseBlocklyLua function
- Replaced some passed values with just one by reference

Tested with valgrind, no memory leaks ;)

### DIFF
--- a/main/EventSystem.cpp
+++ b/main/EventSystem.cpp
@@ -1509,7 +1509,6 @@ void CEventSystem::EvaluateEvent(const std::vector<_tEventQueue> &items)
 		// else _log.Log(LOG_STATUS,"EventSystem: ignore file not .lua or is demo file: %s", filename.c_str());
 
 #ifdef ENABLE_PYTHON
-
 		boost::unique_lock<boost::shared_mutex> uservariablesMutexLock(m_uservariablesMutex);
 		try
 		{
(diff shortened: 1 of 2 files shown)


gizmocuz:
Thanks!

emontnemery:
@jvandenbroek Are any changes to Domoticz needed to run with valgrind?

jvandenbroek:
@emontnemery (Especially) when working with pointers, you should always make sure no memory leaks are introduced.

emontnemery:
@jvandenbroek Yeah, and Valgrind seems to be a pretty awesome tool. What I meant to ask was: Do I need to do any modifications (special build options?) to run Domoticz through Valgrind?

pj-r:
@emontnemery Valgrind usage is bit off-topic but see this: https://github.com/domoticz/domoticz/issues/1955#issuecomment-351427668

xztraz:
This commit might have changed the behavior of blocky events. Stuff that worked with "all" before did go kind of nuts. Events now seem to differentiate between all and device and such. As it should do?

jvandenbroek:
@xztraz I figured that out, reverted that in https://github.com/domoticz/domoticz/pull/2090. I would also think it's probably better to differentiate the events, but previous behavior was there probably for convenience. It wasn't my intention to change this, thus reverted in other patch (not merged yet).

xztraz:
It just revealed my sloppy set up blocky events and i got disco lights for some minutes. all should be all i think but users might get confused a bit at first. maybe a release note or so would do.

PR #2089, opened 2018-01-25
dzVents 2.4.1

dannybloe:
Two fixes (Lua only):
- Fixed week number problems on Windows
- Fixed 'on date' rules to support dd/mm format instead of d/m (e.g. 01/02)

Bumped version and updated docs.


gizmocuz:
Thanks!

PR #2090, opened 2018-01-25
EventSystem/queue fix re-enable event system / Blockly time event fix

jvandenbroek:
Clearing queue function makes sure no segfault happens after re-enabling event system (possible invalid queued items leftover). Moved m_stoprequested to make sure it's set to false before spawning threads (so re-enabling now actually works).

Also make sure timeofday or weekday conditions are met for Blockly time events to trigger (behavior accidentally changed after my previous patch)

### DIFF
--- a/main/EventSystem.cpp
+++ b/main/EventSystem.cpp
@@ -150,6 +150,7 @@ void CEventSystem::StartEventSystem()
     Plugins::PythonEventsInitialize(szUserDataFolder);
 #endif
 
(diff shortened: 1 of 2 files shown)


jvandenbroek:
Wait for merge, want to include another fix... Will comment when ready

jvandenbroek:
Good to merge now :)

gizmocuz:
Thanks!!

PR #2104, opened 2018-02-01
OpenWebNet: change wait time between energy totalizer request

Spugna85:


gizmocuz:
Please see our build systems output:
.\hardware\OpenWebNetTCP.cpp(1073): error C2196: case value '18' already used [C:\projects\domoticz\msbuild\domoticz.vcxproj]

hardware/openwebnet/bt_openwebnet.h
```diff
@@ -195,6 +195,27 @@ typedef enum {
...
+|Description	|	Value
+--------------------------------------
+|Source			|	2#sourceID		  |
+|Speaker		|	3#area#point	  |
+|Speaker Area	|	4#area			  |
+|General		|	5#sender_address  |
+|All Source		|	6				  |
+--------------------------------------
+*/
+typedef enum {
```
gizmocuz: Please remove the typedef as it is not defined.
Could you place this enum in the .cpp file as it is not used in the header (saves compile time)

hardware/openwebnet/bt_openwebnet.h
```diff
@@ -249,6 +270,22 @@ typedef enum {
 	ENERGY_MANAGEMENT_DIMENSION_STATUS_STOPGO_MONTHLY_TOTALIZERS_DAILY_BASIS_LAST_YEAR_COMPARED_TO_32BIT_GRAPHICS_TOUCHX_PREVIUS_YEAR = 514
 }eEnergyManagementDimension;
 
+typedef enum {
```
gizmocuz: Please remove the typedef as it is not defined.
Could you place this enum in the .cpp file as it is not used in the header (saves compile time)

Spugna85: Another developer had declared all the enumerations of this file as typedef with its associated name, even if they are never defined.
Do I have to remove all typedefs and keep only the enums?

The enums should remain in the .h because they could also be used in the OpenWebNetUSB.cpp file when someone will develop new features already present in OpenWebNetTCP.cpp.

gizmocuz: Yes please remove the keyword 'typedef'

gizmocuz:
Thanks!

PR #2105, opened 2018-02-02
Replace all special chars for safe HTML.

Jarthianur:
There is actually such a function, called MakeHtml, but CRLF should not be replaced with br-tags in SafeHtml.

### DIFF
--- a/main/Helper.cpp
+++ b/main/Helper.cpp
@@ -881,25 +881,14 @@ std::string MakeHtml(const std::string &txt)
 //Prevent against XSS (Cross Site Scripting)
 std::string SafeHtml(const std::string &txt)
 {
-	std::string tmpstr = txt;
-	stdupper(tmpstr);
+    std::string sRet = txt;
 
-	bool bHaveFoundDirtyHTML = false;
-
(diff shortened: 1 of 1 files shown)


gizmocuz:
Do we need to change the & sign ? I assume people have names like Garden&Pool or similar

dannybloe:
Isn't everything utf8 already? Then you don't need this relic from the past.

Jarthianur:
No, replacing the ampersand is not really necessary, at least not for security reasons.
I've just compared this to the list of characters beeing replaced by php's htmlspecialchars function.

If the page is built with utf8 encoding, there is also no need to replace it for encoding reasons.
But it would be displayed as `&` either way, as this just effects the representation in the DOM.

I will adjust this.

gizmocuz:
Thanks!

dwmw2:
Now my son's bedroom temperature is reported as `Ned&apos;s room`.
nedaposroom

sweetpants:
@dwmw2, already upgrade toe latest beta?

dwmw2:
@sweetpants This is the latest development branch; I saw it when I updated. Looks like it's already reported as #2138

PR #2106, opened 2018-02-02
HTTPClient return curl error code when no response code

jvandenbroek:
Return Curl's error code (https://curl.haxx.se/libcurl/c/libcurl-errors.html) when no response code (!= CURLE_HTTP_RETURNED_ERROR) is received.


gizmocuz:
Thanks!

Issue #2096, opened 2018-01-30
WRONG CODE -> hitting selector switch


hargathor:
Hi and first thanks a lot for this development !

Domoticz_version: `3.8872`

Since several version I keep getting an error message while hitting a selector switch. It's a dummy switch with 5 states ('Normal', 'Eco', 'Absent', 'Vacances', 'Nuit'). 

I can trigger it using domoticz events but when I try to select a state I get a `WRONG CODE` error message  
Nothing comes up in the logs

ben33880:
same for me. moreover the selector goes into protected alone

ben33880:
lua command : commandArray['UpdateDevice'] = '470|0|40'  (470 the selector, 0 off, 40 level)

the following command causes the selector to become protected, is it normal ?

jvandenbroek:
@hargathor Not sure, need to investigate, will look into it later (but soon enough I hope :P)

@ben33880 Why are you trying to send both off and level? Just use '470|40' if you want it set to level 40 or '470|0' to set it off. You can indeed also set the protection flag with the 3rd value, so '470|44|1' would make it protected and '470|44|0' unprotected. However, just leaving out the last one won't change it's value.

Edit: hmm, sorry, you should be able to set both nvalue and svalue (off and level), could you try 470|0|40|0 ? If that works, I need to check the code.

ben33880:
it was working but not now !!
try 470|0|40|0 and it works fine.

Thx.

jvandenbroek:
@hargathor when the new beta is build (should be within a few hours max), it should no longer set the protection flag when updating it though the event system. You can remove the protection flag (which causes your WRONG CODE message) by clicking 'Edit' on the switch and toggle it there.

hargathor:
@jvandenbroek Thanks. Indeed my Switch was going into protected mode for the same reason as ben33880. An event with `idx|0|40` I tried your tips with only `idx|40` but it rose an exception. However with `idx|0|40|0` it seems to work fine (althought it seems to Switch off my selector)
I've upgraded tout the new beta. i will try again thé first and second way

jvandenbroek:
@hargathor yes idx|40 was my mistake and is wrong, but I hope it didn't crash on that (what kind of exception?)? Idx|1|40 would set the switch on and at level 40.

hargathor:
[...]

PR #2108, opened 2018-02-03
EventSystem fix Lua/dzVents UpdateDevice protected flag

jvandenbroek:
Fix old behavior for UpdateDevice, now protected flag is set when not provided (see also https://github.com/domoticz/domoticz/issues/2096)


gizmocuz:
Thanks!

PR #2117, opened 2018-02-07
Add support to update user variables and setpoint devices from python scripts

teeminus:
Currently, there is no way to update/change user variables from a python (event) script. The same goes for setpoint devices. With the patch, both features will be added..

The changes I made are in the CEventSystem::PythonScheduleEvent function but it might be put in in the CEventSystem::ScheduleEvent instead.


gizmocuz:
Thanks for the patch! One minor request (not your fault at all!),

could you replace (the default log)

EventSystem: EvohomeSetPoint

to

EventSystem: SetPoint

Also in the normal EventSystem.cpp class

As this is also for non-evohomesetpoints as in MainWorker::SetSetPoint the function SetSetPointInt is called for non-evo home systems

teeminus:
Thanks for your feedback. I changed the error message text.

gizmocuz:
Thanks!

PR #2118, opened 2018-02-07
Address erroneous values in counter meter readings

peahonen:
If a counter ever takes a step back, the graph may show a huge spike
due to an overflow as described in:
https://www.domoticz.com/forum/viewtopic.php?f=6&t=14125&p=170169#p170169

More complete and sustainable approach would be to enable
warnings on sign conversions (-Wsign-conversion) and treat them as errors


gizmocuz:
Would it not be better to better to check if ulFirstValue is higher then actValue and discard it ?

For example, you see the commented 'if' below, if you leave the total value unmodified, but you use the if statement like this

if (TotalValue >= 0)

then the problem will be solved too

gizmocuz:
Erroneous values caused by hardware (or any other process outside Domoticz) should not be concealed by Domoticz. Its good to see there are wrong values being pushed into the system

PR #2121, opened 2018-02-08
www/styles/dark-th3me/custom.css

TacticalLeader:
Fixed .dimslider size and position error. Tested on Chrome, IE, Firefox and tablet.


gizmocuz:
Thanks!

Issue #2110, opened 2018-02-05
Setting level values > 100 with switchSelector


elxsi:
Hi,

it seems impossible to set level values > 100 via the switchSelector function:

As per DzVents documentation:

switchSelector(level): Function. Switches a selector switch to a specific level (numeric value, see the edit page in Domoticz for such a switch to get a list of the values). Supports command options.

The edit page in domoticz for this device (IKEA tradfri light bulb) has levels from 0 (off) to 200 (warm white) in steps of 10. Setting the level via switchSelector (dzVents) to values > 100 is always set to 100, but switching the levels in the domoticz UI works as expected.

So this is either a bug in the switchSelector function or expected behaviour (level from 0 to 100%?).

elxsi:
Hi,

I think it would be better to allow levels > 100 for switchtype STYPE_Selector. In my opinion a selector switch has different states (and not percent values), so - in this case - the level just should be treated as a state number. For example - think of a radio device where you want to switch pre-configured radio stations, it would be perfectly make sense to have more than 100 stations.

I solved my issue by changing the level check in EventSystem.cpp at line 4091 (latest development branch), I just left the check for negative level values in place and removed the if (ilevel > 100) check.

I don't know if this would have any side effects, but it looks like this check is only in-place in the event system and not in the WebServer.cpp which explains why it already worked from within the domoticz web ui.

Maybe someone could review this and - if it is ok - add it to the domoticz code base.

Thanks!

PR #2125, opened 2018-02-10
EventSystem allow selector switch level set > 100

jvandenbroek:
https://github.com/domoticz/domoticz/issues/2110


gizmocuz:
255 values not enough ?

jvandenbroek:
Probably is yes, seems unlikely to use more than 25 items in a selector switch. However, looking down the road I saw that function SwitchLightFromTasker in SQLhelper.cpp already expects int for level. So while not necessarily needed, it seems more consistent to pass it through as int as well.

gizmocuz:
Thanks!

PR #2127, opened 2018-02-10
somewhat fixed white slider in rgbw popup in dark-element skin

xztraz:
Before
rgbw_before

After
rgbw_after

### DIFF
--- a/www/styles/element-dark/custom.css
+++ b/www/styles/element-dark/custom.css
@@ -253,9 +253,14 @@ width: 99.5% !important;
    margin-left: 95px !important;
 }
 
-#rgbw_popup .dimslider {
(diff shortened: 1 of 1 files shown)


gizmocuz:
And the light as well ? Maybe this needs a global css change ?

xztraz:
probably a global one(it's a mess to change skins). could not get it better than that in decent time. looks like a bunch of rowbreaks in html between sliders too.

gizmocuz:
Should be done globally and not for this specific theme

gizmocuz:
This should solve it globally i hope:

https://github.com/domoticz/domoticz/commit/7d890a97fb14ac489bedb0fb0b4bdabf9351ba0d

PR #2133, opened 2018-02-12
EventSystem Lua Classic use queued item value for otherdevices tables

jvandenbroek:
Had fixed the lastupdate / lastlevel issue before, but after changes for dzVents 2.4 the old (faulty) behaviour came back for classic. See https://www.domoticz.com/forum/viewtopic.php?t=22157
This will make sure the item's queued values for lastupdate, lastlevel and svalue are being evaluated.


gizmocuz:
Thanks!

PR #2139, opened 2018-02-14
Log error messages from python event scripts to domoticz error log

teeminus:
Currently when you write a python event script, there is no way to get diagnostic informations when the script failes. The script either runs - more or less as expected - or, in case of eg. a syntax, identation, ... errors, it does not. But there is not signalling to the user, if the script execution has failed. So you start wondering if either your script logic is not working or if there are any errors in the script und you start poking around until you eventually find the reason why your script is not working.

The reason why you can't get diagnoticz informations from the python interpreter is, that the PyRun_SimpleStringFlags and PyRun_SimpleFileExFlags functions have build in error handling which will print all errors to stderr.

To solve this problem, I implemented an override for the sys.stderr stream. Before running the event script code using PyRun_SimpleStringFlags/PyRun_SimpleFileExFlags my code injects a new python class into the python interpreter that overwrites the sys.stderr stream and buffers all messages written to stderr in an internal buffer variable. After the event script has been executed, the code pulls the buffer from python interpreter and - when a message has been written to stderr - writes the error messages to the domoticz log.

This solution might also be used to redirect the sys.stdout stream, so that all print(...) messages could be written to the domoticz log. As the messages from the redirected stderr stream are pulled from the python interpreter after the script has been run, the messages of the redirected stdout stream would only appear in the domoticz log after the script has been executed.

### DIFF
--- a/hardware/plugins/DelayedLink.h
+++ b/hardware/plugins/DelayedLink.h
@@ -111,6 +111,8 @@ namespace Plugins {
 		DECLARE_PYTHON_SYMBOL(Py_ssize_t, PyByteArray_Size, PyObject*);
 		DECLARE_PYTHON_SYMBOL(PyObject*, PyErr_Occurred, );
 		DECLARE_PYTHON_SYMBOL(long, PyLong_AsLong, PyObject*);
+		DECLARE_PYTHON_SYMBOL(PyObject*, PyUnicode_AsUTF8String, PyObject*);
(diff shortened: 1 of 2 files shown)


teeminus:
The solution currently only works for python versions >= 3. I can add the support for python versions < 3 if needed.

flatsiedatsie:
This would be very welcome! Perhaps it could output everything only of the plugin has the debugging enabled?

dnpwwo:
@teeminus,
Domoticz only handles Python 3.4 and above so I wouldn't bother supporting earlier versions.

PR #2142, opened 2018-02-16
Reencode windows error message from local code page to UTF-8

emontnemery:
This should solve the repeatedly reported issue with plugins not working on windows.

Root cause is that error messages passed to onConnect are not UTF-8 encoded as mandated by Python.

Big thanks to forum user Reconos who helped debug!

### DIFF
--- a/hardware/plugins/PluginMessages.h
+++ b/hardware/plugins/PluginMessages.h
@@ -117,6 +117,33 @@ namespace Plugins {
 		};
 	};
 
+#ifdef _WIN32
+static std::wstring string_to_wstring(const std::string &str, int codepage)
+{
+	if (str.empty()) return std::wstring();
+	int sz = MultiByteToWideChar(codepage, 0, &str[0], (int)str.size(), 0, 0);
+	std::wstring res(sz, 0);
+	MultiByteToWideChar(codepage, 0, &str[0], (int)str.size(), &res[0], sz);
+	return res;
+}
+
+static std::string wstring_to_string(const std::wstring &wstr, int codepage)
+{
+	if (wstr.empty()) return std::string();
+	int sz = WideCharToMultiByte(codepage, 0, &wstr[0], (int)wstr.size(), 0, 0, 0, 0);
+	std::string res(sz, 0);
+	WideCharToMultiByte(codepage, 0, &wstr[0], (int)wstr.size(), &res[0], sz, 0, 0);
+	return res;
+}
+
+static std::string get_utf8_from_ansi(const std::string &utf8, int codepage)
+{
+	std::wstring utf16 = string_to_wstring(utf8, codepage);
+	std::string ansi = wstring_to_string(utf16, CP_UTF8);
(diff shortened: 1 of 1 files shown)


emontnemery:
@dnpwwo do you think the changes are OK?

gizmocuz:
Maybe this was merged a bit quickly....

@emontnemery, i can't seem to find a discussion about this on the forum. Do you have an example python script, so i can test this ?

could this have worked as well:

```
std::string isostring_to_utf8(const std::string &input)
{
	std::string output;
	for (size_t ii = 0; ii < input.size(); ii++)
	{
		uint8_t ch = input[ii]; // assume that code points above 0xff are impossible since latin-1 is 8-bit 
		if (ch < 0x80) {
			output += ch;
		}
		else {
			output+=(0xc0 | (ch & 0xc0) >> 6); /* first byte, simplified since our range is only 8-bits */
			output+=(0x80 | (ch & 0x3f));
		}
	}
	return output;
}
```

^^code taken from:
https://stackoverflow.com/questions/5586214/how-to-convert-char-from-iso-8859-1-to-utf-8-in-c-multiplatformly

this might be a more general platform based solution, coz i do not think that 'It is safe to assume non-Windows will always be UTF-8'

emontnemery:
@gizmocuz  Thanks for the review!

There are many forum threads, a couple of examples:
- https://www.domoticz.com/forum/viewtopic.php?f=65&t=21602
- http://www.domoticz.com/forum/viewtopic.php?t=16096

The issue is that this Python call fails:
`Py_BuildValue("Ois", m_pConnection, m_Status, textUTF8.c_str())`
Because the string argument must be a char* with valid UTF8 data.
If it's not, the call return NULL and you get this error message:
` 'onConnect' failed 'TypeError':'onConnect() missing 3 required positional arguments: 'Connection', 'Status', and 'Description''`

You can reproduce by changing Windows display language, for example to Swedish or Danish, and then trying any TCP/IP based Python plugin, for example this one: https://github.com/emontnemery/domoticz_mqtt_discovery

Your suggestion will unfortunately not work because it's not a proper character encoding translation, it's a hack to make ISO8859-1 strings be accepted as UTF8 data. I think your suggestion might work/hide the issue for most western European users - Windows-1252 is very similar to ISO8859-1, but definitely not for others, for them it would make the error message illegible.

Linux moved to UTF-8 quite some time ago, but not sure about MacOS X.

A more general approach could be:
- For Windows always do proper character encoding translation as already merged
- Improve by checking return value of `Py_BuildValue("Ois", m_pConnection, m_Status, textUTF8.c_str())` and if it fails, print an error message and replace the error string with the empty string and call Py_BuildValue again. This will then hide the issue on non Windows platforms using localized error messages not in UTF-8

gizmocuz:
Thanks for the explanation. Then this is a perfect way to make a correct translation using the code_page.
Thanks again for the patch! I'm lucky i live in a country with easy characters ;)

PR #2147, opened 2018-02-18
Add SwitchType to Python Device object

emontnemery:
SwitchType was already available in the C++ class, but not exposed to Python.


gizmocuz:
Thanks!

PR #2157, opened 2018-02-21
 Added ZWave node ID to log

Olifant1990:
Added ZWave node ID to log, thanks to this commit:
7d7defd#diff-50d0461e793b63517112bdafa0ded628

Is there someone with a Z-wave device who can test this?
I have one production server with Z-wave, but no test environment with
Z-wave :(


gizmocuz:
Thanks for the patch!
Our GCC versions can not handle std::to_string yet.
You can not assume that zwave devices have at least 5 digits (substr), take for example a temperature sensor, this has a 16 bit ID.
Also here, the first BYTE is the node_id
About the logging, why is logging the node_id so important? If you name your sensor correctly, that, in my opinion, would be much more interesting then displaying a node_id
To hunt for your sensors that belong to a specific node, this can be done in the devices overview and the filter there.
If your zwave hardware has the name zwave, and your node_id is hex = 0x12 then in the filter you can type 'zwave 12' and look for all sensors that start with 12 (or 0012).
This is not that complicated, and needs to be done once when you label your sensors.

Ideally we should also have a database table with the original_node_id, the assigned_node_id (and sensor id's).
With this we can exclude a node, and include it again (so it has a new node_id), and map the old sensor to the new node_id.
This way your timers/scripts/... still uses the same sensor. The node_id you see in the devices overview in front of the sensor now also does not match anymore, so we need an additional column to be displayed here (the original_node_id)

Maybe a wiki page labeled something 'how to track back your zwave node_id from a device/sensor' helps a lot of users

Olifant1990:
Ok, let me explain my problem.

I change a lot in my home automation. Sometimes I switch some dimmer modules and try something new.
Therefore I reset a dimmer module and re-include it. This device will add about 10 devices in the device overview. When including, you'll see that "Node ID: 7" is added. Ok, then I go to the log page to see if the device is sending data correctly. What I see then is multiple times "(OpenZWave USB) General/kWh (kWh Meter)". I don't know from which Z-Wave Node this is comming from. Off course I know I could rename the devices to make it clear, but to rename all devices every time after including, takes a lot of extra effort. (I include via my mobile phone browser). So it would be nice to see directly which node is sending the information without renaming the devices.

Only a small addition of the node ID should do it for me, so it would be (OpenZWave USB 007 (0x07)) General/kWh (kWh Meter)

gizmocuz:
When you add a new node (with your mobile device) and it succeeds, it would always be a good option to go to the devices tab in domoticz.
For instance, your new node_id is 0x15
In the devices tab filter you press the button 'Not used'
in the filter you enter part of your zwave hardware name (zwave)
and the node id you are looking for (15)

zwave 15

you should see your new sensors, which you can label correctly, and add to the system

Because you disabled 'Accept new hardware' in the settings (disabled by default, enabled while including new node's), your hardware list is very clean and it is no problem looking up your new nodes.

A log message like "Node: 18 Kwh Lamp" ... the 18 does not make sense here, better have a log like "Kwh Outside Lamp"

Olifant1990:
True, off course it is possible, but this is quicker and sooner to see.

Just like the commit which I used as base for my idea: https://github.com/domoticz/domoticz/commit/7d7defdc0354dddb5866c7c6f19ffcfaa2c5d90a. In fact this is also extra info which is already visible in de ID of the device.

But when adding devices, I (and maybe that is personal) don't want to go to the device page and name the labels for (sometimes 6 or 7) new devices only to see in the log that that device is sending the needed information (on a mobile phone).

But if you think this is useless, please, feel free to close this pull request :)
Just was an idea ;)

gizmocuz:
The method i described above (and on various other places, like the forum, and maybe even the wiki) is quite easy ?

Olifant1990:
Sure! I didn't say it is difficult, but it would save me a few steps and maybe other users would like this too. But as said, Feel free to close this when you don't like it :)

gizmocuz:
Thanks for your effort but i am closing the PR.
As this is only relevant when you just include a node, and can be done by the method described above
Also when you next time exclude and include the node, it has a different id... no need to display this 'ID' as the ID is not relevant, it is the sensors name

Issue #2114, opened 2018-02-06
HTML improvement: add 'options' class


flatsiedatsie:
Currently there are quite a few inconsistencies in the Domoticz HTML output. Chrome is now grading some of these as errors. Perhaps some low-hanging fruit could be fixed.

Some examples:
- In item tables, instead of classes each TD is given an ID. This is one of the error Chrome complains about, as officially an ID should always be unique, while you can have more than one of a class. I would propose a slow transition. First, these TD's could a class with the same name as the ID they currently have. So for example, _td id="name"_ would become _td id="name" class="name"_ . Sometime in the (far) future, the ID's could be fased out. The Aurora theme, for example, currently uses javascript to copy the ID's into classes. This is important because JQuery assumes correct CSS. Many of its functions work on elements with classes, but don't work on ID's, requiring hacks.

- Some TD's don't have a name, which makes them very hard to target via CSS. In the Aurora theme I have suggested a few, which are again 'hacked in' via jquery. For example: give the TD that contains options like "edit" and "log" the _options_ class name.

Without names for elements, you get CSS like this:

	body table#itemtable tr td:first-child + td + td + td + td + td + td {
		font-size: 90%;
		clear: both;
		padding: 0 5% 0 5%;
		line-height: 130%;
	}

and:

	#settingscontent textarea + span,
	#settingscontent textarea + span + span,
	#settingscontent textarea + span + span + span,
	#settingscontent textarea + br + span,
	#settingscontent textarea + br + span + span,
	#settingscontent textarea + br + span + span + span {
		color: #ccc;
		font-size: 10px
	}

Which is currently in the Domoticz CSS.

This styling is very fragile. It breaks as soon as the order of the html changes. You could say that this is the opposite of that CSS is for: separate structure from styling. Here CSS is actually making Domoticz less easy to work on, instead of easier.

I could add this CSS (as I has done before in the Gui rework branch).

Should I?

gizmocuz:
Yes please go ahead! But can you show the errors chrome is giving ? As i don't have them here.

flatsiedatsie:
<strike>Oddly enough I don't see the errors now either. Perhaps it was on another browser. But I did see them.</strike>

<img src="https://www.pineapplejazz.com/public/domoticz/other/chrome-errors.png">
Chrome gives 54 errors like this.

So:
- I've directly added the 'options' class everywhere. It already existed on the utilities page.
- I've created pull requests for the addition of the classnames to all TD's that already had ID names.
[...]

PR #2158, opened 2018-02-23
CSS - Weather: copied ID names into class names

flatsiedatsie:


gizmocuz:
For me (looking at the description provided with this pull request), it is unclear why you need the id's also to be available as class name. (?)

is there a typo in this patch ? ('bigtext' -> 'name')?

`<td id="bigtext" class="name ng-cloak">`

flatsiedatsie:
Thanks, that is indeed a typo. Well spotted! Github does now allow me to edit it. Could you help me out here?

The reasons were discussed recently, and I even asked your permission to make these changes.
https://github.com/domoticz/domoticz/issues/2114

gizmocuz:
Thanks for the reference, makes things a lot clearer now.
Yes i also experienced these errors.. warnings actually.
For example (settings page), take the '#Mode1' '#Mode2' these are filled directly and are used by a lot of switches.
These days i would have used an angular approach and put them in a scope array and use this.
(Might still do this)
As a lot of Hardware has an 'apikey', we use the same ID, but actually since this is inside it's own scope, it is scope.apikey , which i find unique.
That chrome does not find this unique, thats chrome's disiccion

For example:

hardwareparamsunderground.apikey
hardwareparamssolaredgeapi.apikey
hardwareparamsnestoauthapi.apikey

Also in the controller they are referenced like 
$("#hardwarecontent #divunderground #apikey").val();
$("#hardwarecontent #divsolaredgeapi #apikey").val();

I see no problems with this.

But this is in the hardware section, not like in the PR with the Weather Widget

In the temperature and weather display pages, we are using Angular Directives, and we do not need the id's anymore.
You seem to want to create special classes/effects for certain elements, which is great, and i fully support this, but i would suggest to use other names, instead of 'class="name"' something like 'class="widget_name"' (widget_ prefix)

Hope to work soon on the other pages/directives

flatsiedatsie:
If you want to store a variable in a html element, then I'd suggest to use a custom attribute like "data-apicode" or something. You are free to create as many custom attributes as you like. But the ID and Class attribute are reserved for design and markup of HTML.

**ID**: Reserved for CSS use, to target html elements of which you know there is only one per page, like #mainmenu or #logo.
**Class**: Reserved for CSS use, to target html elements of which there are multiple on a page. Like giving a div, td or paragraph a name that signals a special version.

That is the standard way of doing these things.

If you want to use JQuery targetting, then you could do this:

A.
Change ID into a class. Then having multiple that are the same is no issue. For example:
_$("#hardwarecontent #divunderground .apikey").val();_ ( # become a . )

B.
Even better would be to store the key in a data attribute as I mentioned:
$("#hardwarecontent #divunderground [data-apicode='124729']").val();

See:https://stackoverflow.com/questions/4146502/jquery-selectors-on-custom-data-attributes-using-html5

C.
The best option would be to use the data function in JQuery:
https://api.jquery.com/data/



Ok, and then your other point:

> You seem to want to create special classes/effects for certain elements, which is great, and i fully support this, but i would suggest to use other names, instead of 'class="name"' something like 'class="widget_name"' (widget_ prefix)

The normal way of using CSS is to create 'sentences' out of your classes that, ideally, make sense to another coder or designer that has to later understand what you did. 

For example, it's better to have:
.dashboard .utility .item .name {color:red}
than
.dashboard .utility .item .item_name {color:red}

In CSS is is already clear that you are inside an item, because the parent element should have that as its name. It would be redundant (and a waste of space and bandwidth) to give an element the name "widget_name" when its already has a parent element called "widget".

gizmocuz:
it's best if you want to style all item names of all the widgets to have a class widget_name right ?

But you are mixing two different issues here now, the reference was for the hardware page, the temp+weather do not use the id's anymore as it's a angular directive

Also mentioned (and this supports what i was trying to say), please use the BEM naming schema, , it's good to start at 'some' point, might as well be this

flatsiedatsie:
> it's best if you want to style all item names of all the widgets to have a class widget_name right ?

I don't believe so, but I'm not against it either. Once these current changes are accepted it will be easy to do a mass replace, and turn it into BEM.

> But you are mixing two different issues here now, the reference was for the hardware page, the temp+weather do not use the id's anymore as it's a angular directive

I'm not sure what you are talking about? The issue I created was for the entire front-end. Classes were missing in the entire front-end. I've added them in the entire front-end.

> Also mentioned (and this supports what i was trying to say), please use the BEM naming schema, , it's good to start at 'some' point, might as well be this

Sounds like we have a plan. Once these current changes are accepted it will be easier to do a mass replace. That's because in this code I've already streamlined some things, like:
- adding class="img img1", which will allow themes to target all the images at once, or target images individually, whatever they prefer.
- I manually made sure that these classes worked on elements with multiple attributes. So there are no elements with "class="bla" class="ng-bla".

gizmocuz:
I was referring to:

The reasons were discussed recently, and I even asked your permission to make these changes.
#2114 

This i consider a complete different issue/pr.

Recently i updated some classes (also angularjs) because of XSS possibility. (still not finished)
Needed to make some changes on the temperature/widget directives too.
For now, for styling, i would say please add the classes (with BEM naming) and leave the id's in place.
I will have a look very shortly, but i think all id's can be removed already from the directives.
Next steps are to make the utility/lights/scenes and dashboard also with directives

flatsiedatsie:
Sounds great!

Only thing is: As a front-end designer, I don't use github. I only know how to use the web interface :-(

So there are two options:
1. I have to do all the work of adding classes all over again.
2. You accept the changes I already made, and then it's just a simpe Find-Replace action.

Let's go with option 2 :-)

Also, I haven't removed any ID's. I've only added classes.

gizmocuz:
Of course option 2 would be the easier.. but it is good to get yourself familiar with version control.
This was also the reason why it took me days to already implement a lot of your changes already.
I need to cross check every file, and every change, as i am not going to just accept the changes.
For example, someone currently is doing a great job separating the timer stuff from the lights/switches controller. This is causing some code changes as you might understand.
I also recently merged some PR requests that makes changes in the HTML
All these changes will be removed by just accepting your changes of course....

When your changes are in source control (git) , the system is most times smart enough to know what to change, and other developers can also still work on their stuff.

Maybe it's an idea to start with the current development code, and start with the core changes and PR them.
If i can help, drop me a mail

PR #2159, opened 2018-02-23
CSS - Temperature: copy ID names into class names

flatsiedatsie:


gizmocuz:

PR #2161, opened 2018-02-23
CSS - Scenes: copied ID names into class names

flatsiedatsie:


Kepro:
@flatsiedatsie use class ".name" is not very good, if you want then use BEM system

flatsiedatsie:
BEM is great, but let's fix one problem at a time. Currently Domoticz doesn't even have classes for most of its elements, as it somehow uses ID's everywhere instead. BEM, by contrast, tries to not use ID's.

If you want to do the work of changing this into BEM: be my guest!

//update: we're going to implement this.

gizmocuz:

PR #2162, opened 2018-02-23
CSS - Lights/Switches: copied ID names into class names

flatsiedatsie:
(and manually added "img1" class where necessary).


gizmocuz:

PR #2165, opened 2018-02-23
Given a name for unnamed TD on dashboard

flatsiedatsie:
On the dashboard there was an unnamed TD which contains things like sliders and switches. To give it a broad as possible future proof name, I thought it might be nice to call it "input". But I'm open to any name, as long as it gets a name :-)


gizmocuz:
Thanks. Please also update the appcache version number when you make changes to the www folder

PR #2167, opened 2018-02-23
Correct handling of newline

emontnemery:
Fix broken handling of empty line which cause the last received command to be executed again.
This can be triggered (with some bad luck) when reconnecting over TCP for example.
Also replace some magic numbers with character literals.


gizmocuz:
Thanks!

PR #2173, opened 2018-02-24
Now plugins control timeout for all device types

guillaumezin:
Now plugins control timeout for all device types : they enable and disable timeout for any device they control


gizmocuz:
I'm not sure if this is a wanted result all the time...
What happens with a switch or a text sensor ? (that is only updated very sporadic)

What do you mean with "they enable and disable timeout for any device they control"

guillaumezin:
Currently, the behaviour depends on type of device.

For these devices

`(dType == pTypeLighting1) || (dType == pTypeLighting2) || (dType == pTypeLighting3) || (dType == pTypeLighting4) || (dType == pTypeLighting5) || (dType == pTypeLighting6) || (dType == pTypeFan) || (dType == pTypeLimitlessLights) || (dType == pTypeCurtain) || (dType == pTypeBlinds) || (dType == pTypeRFY) || (dType == pTypeChime) || (dType == pTypeThermostat2) || (dType == pTypeThermostat3) || (dType == pTypeThermostat4) || (dType == pTypeRemote) || (dType == pTypeGeneralSwitch) || (dType == pTypeHomeConfort) || ((dType == pTypeRadiator1) && (dSubType == sTypeSmartwaresSwitchRadiator)) || ((dType == pTypeRego6XXValue) && (dSubType == sTypeRego6XXStatus))`

The timeoutd parameter given by plugins is used.

Otherwise it is ignored.

So for instance, for temperature sensor, if the plugin doesn't send a value on a regular basis, the device is shown in timeout state, even if is the normal behaviour from the plugin point of view. On the contrary, Domoticz never set timeout for text sensor (I think, I didn't check). With this PR, a plugin can decide to timeout a text sensor if it needs to.

The PR changes that so timedout is controlled by plugin for every devices type, so the plugin can decide to set a device as timed out, given that the device has been created and is controlled by plugin.

gizmocuz:
Thanks!!

PR #2179, opened 2018-02-28
Expose Description column to Python plugins

emontnemery:


dnpwwo:
@emontnemery, You need to make sure you update the wiki when you add these kinds of changes otherwise no one will know they are there.

emontnemery:
@dnpwwo Good point! Wiki now updated with:
- SwitchType which was recently added to Devices dictionary items
- Changes to Devices dictionary item's Update function.

I'll update with this change if/when merged.

gizmocuz:
Thanks for the patch!
@dnpwwo, i trust you to merge all python related PR's ? ;)

PR #2183, opened 2018-03-02
Support of EV1527 & Eurodomest switches in manual add dialog

lechunlx:
Support of EV1527 & Eurodomest switches in manual add dialog by adding in www views & app.
Also fixed formatting for RFLink.


gizmocuz:
I think you have merged two different topics in this pull request
Is it possible to only include the part for this change ?
I see you modified 'ConvertToGeneralSwitchType' which is used by RFLink, and i think this might cause problems.
Best to write your own function for  EV1527 & Eurodomest ?

lechunlx:
Hello,
Ok - now we have only changes for switches EV1527 & Eurodomest.
I will think how to resolve issue with manual adding devices with RFLink because there is definitively a bug.

gizmocuz:
Thanks!, Could you also undo the change in the PR for temperature_custom_temp_log.html ?

stuntteam:
Manual adding of RFlink devices is done through the hardware tab with the create device button that is located behind the RFlink hardware. 
In the popup you simply enter the 20;.... commands that RFLink receives from remote controls etc.

lechunlx:
Custom temp view back to original.

About RFLink - I know the way of creating switches from Hardware tab, nevertheless it's a bit messy way and no so elegant than clicking in "manual" way on Switch tab. Moreover, it is working, but wrong. When you trying to add a switch with "blinds" type, it will add it with lowercase and no leading zeros to database - but by adding the same device by RFLink (Hardware tab method), it will add it with leading zeros and uppercase - that's the whole problem.
I have made a little research, and found that ConvertToGeneralSwitchType is used twice in WebServer.cpp only during manual add (for addswitch and testswitch). So we can't mess anything else. I'll try to propose some fix later.

gizmocuz:
Thanks for the patch, those new table row id's, i think you could undo those changes as they are not used in the code ? (lighting1paramsHouseCode)

lechunlx:
Done - unnecessary IDs have been removed from change

lechunlx:
@gizmocuz : Do you have any more remarks/comments before merging this pull request?

gizmocuz:
I only see two files changed:

www/app/LightsController.js
www/views/lights.html

If the only change for this PR is in the HTML part of Domoticz, then also increase the version number in the appcache file

lechunlx:
Yes, for this PR only these 2 files. Version has increased.

gizmocuz:
Thanks!

PR #2185, opened 2018-03-02
Extract device timers editor to own page to simplify navigation

stas-demydiuk:
When we open device timers editor from LightSwitches page there is no way to refresh it or go to timers editor by direct link. Additionally browser "back" button doesn't work properly in this case.

Device timers editor was extracted to own angular route (page) to fix this. This change additionally simplifies future support of timers page and LightSwitches page as source code is divided from each other.

Additional changes:
- added babel-polyfill to support ES6 features in all browsers
- added dataTables.select plugin to simplify row selection in DataTables
- timers editor was rewritten in more angular way
- api-related logic was extracted from controller to factories


gizmocuz:
Great patch!! Thanks!! The more Angular the better!
Unfortunately i might have just merged a pull request that had changes in the LightsController file causing a merge-conflict for this patch.
Is it possible to have a look at this ?

There is one other place that is using timers, that is in the utility tab for Thermostats, could this also use this ?

stas-demydiuk:
@gizmocuz I've updated the PR with resolved merge conflict.

Thanks for the note about Thermostat timers, currently it is not possible to use my PR to replace this page because they use different API for timers (`json.htm?type=timers` and `json.htm?type=setpointtimers`) but they look pretty similar and I'm going to update Utility to use this new editor as well.

gizmocuz:
Great! And thanks for the PR!!

gizmocuz:
Something seems off with the Update/Delete and ADD buttons

image

There is a white border around them... maybe a style change ?

gizmocuz:
Found it, and made a small change when the timer list is refreshed (selectedindex=null to disable the update/delete button)

https://github.com/domoticz/domoticz/commit/48f283da42e886dfdb3d44822173f8635a115860

www/app/DeviceTimers.js
```diff
@@ -0,0 +1,837 @@
...
+
+        function initDatePicker() {
+            var nowTemp = new Date();
+            var now = new Date(nowTemp.getFullYear(), nowTemp.getMonth(), nowTemp.getDate(), 0, 0, 0, 0);
+            var element = $element.find("#sdate");
+
+            element.datepicker({
+                minDate: now,
+                defaultDate: now,
+                dateFormat: "mm/dd/yy",
```
dwmw2, line 464: You should never hard-code the US-local "mm/dd/yy" format. Please use the ISO standard YYYY-MM-DD. That's what Domoticz already uses in *almost* all of its user interface; please don't add new exceptions that make it inconsistent.
I've updated #1973  to fix this.

emontnemery:
@stas-demydiuk This is a really nice patch! Did you start working on scene timers already? I'm asking since I have to merge my changes done for timers in context of #2078 to the new structure.

PR #2187, opened 2018-03-03
Door Lock Inverted

gibbo520:
Added Door Lock Inverted switch type feature 
added delay off on Door Lock and Door Lock Inverted switch type
removed delay (on/off) on door contact switch


gizmocuz:
Thank you for this patch !
Still not sure why there would be a need for an 'inverted' door lock ... is the wiring OK ?

One minor request, could you remove the changes in hardware/OpenWebNetTCP.cpp ?

gibbo520:
If a door is closed , and i i want open it i send on for unlock and Not for lock it. 
wiring is ok yes !

gibbo520:
Other prople have same problem 
http://www.domoticz.com/forum/viewtopic.php?f=31&t=22191

gizmocuz:
Okey i understand. Please remove the changes in hardware/OpenWebNetTCP.cpp

gibbo520:
done!

gizmocuz:
Thanks !!

PR #2191, opened 2018-03-05
Allow up to 31days in Short Log History

lechunlx:
This change will allow to keep up to 31days in short-term history - it's a decision of user if can afford to store more data in database.
Also added new types of graphs in custom graphs to being less precise (15min, 30min and 1h).

### DIFF
--- a/main/WebServer.cpp
+++ b/main/WebServer.cpp
@@ -13125,8 +13125,21 @@ namespace http {
 				if (sensor == "temp") {
 					root["status"] = "OK";
 					root["title"] = "Graph " + sensor + " " + srange;
+					int nValue=0;	
+					m_sql.GetPreferencesVar("5MinuteHistoryDays", nValue);
+					if (nValue<=7)
+					{
+						result = m_sql.safe_query("SELECT Temperature, Chill, Humidity, Barometer, Date, SetPoint FROM %s WHERE (DeviceRowID==%" PRIu64 ") ORDER BY Date ASC", dbasetable.c_str(), idx);
+					}
+					else if(nValue<=14)
+					{
+						result = m_sql.safe_query("SELECT ROUND(AVG(Temperature),1) AS Temperature, ROUND(AVG(Chill),1) AS Chill, ROUND(AVG(Humidity),0) AS Humidity, ROUND(AVG(Barometer),0) AS Barometer, MIN(Date) AS Date, ROUND(AVG(SetPoint),1) AS SetPoint FROM %s WHERE (DeviceRowID==%" PRIu64 ") GROUP BY strftime('%%Y-%%m-%%d %%H',Date),(strftime('%%M',Date)/30) ORDER BY Date ASC", dbasetable.c_str(), idx);
+					}
+					else
+					{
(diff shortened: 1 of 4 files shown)


gizmocuz:
I dont think this is a wise idea as this will flood the database and produces lots more traffic to the graphs
You can always set this value yourself directly in the preferences in the database
The grouping of items is nice, but that would be great for all graphs, not only the custom temp graph.
Could be great if you could add this. Then we can reduce the amount of traffic to the graphs

lechunlx:
Yes, I agree with you about database flood etc., but please note, that default value of 5MinuteHistoryDays is still 1. So it depends of user choose for how long want to keep detailed data. Event if database will grow to few megabytes, it shouldn't hurt anyone.
I have made a little calculation and each day will cost us about (for 1 device per table):
Temperature - 20Kb
Rain - 12Kb
Wind - 14Kb
UV - 11,5Kb
Meter - 12Kb
MultiMeter - 22Kb
Percentage - 15Kb
Fan - 10Kb

So it's highly related to how many devices you have, but for eg. 10 Temperature, 1 Rain, 1 Wind, 5 MultiMeter will flood database with 336Kb of data each day. For 31 days it will consume around 10Mb.
I know that I can set up 5MinuteHistoryDays parameter manually in DB, but this change should help less experienced users too.

About reducing traffic to the graphs, I'll be happy to help with all graphs, please wait with this pull, let me work for it a bit more.

gizmocuz:
Thanks

gizmocuz:
Thanks, would be great if this was possible:

https://www.highcharts.com/demo/line-time-series

lechunlx:
Currently I have made modifications of temp, percentage and fan graphs.
It depends of 5MinuteHistoryDays (if <=7 days then 5mins, <=14 days - 30mins or more days - 1h step)
But, I'm not quiet sure if I'm doing well, because I don't have filled enough database with data.
Do you have some test database filled with types like wind, rain etc. to test rest types of graphs ?

gizmocuz:
Not with more then 2 days of data, thats why i think there will be a lot of data send, and the database will get larger
Maybe you test it for a week to see how the reality is ?

gizmocuz:
But it seems this chart example is not doing what i had in mind. The chart should request from the server the time range/interval, that would be the best. But maybe highcharts does not have an option for it.
In that case sending all data and allowing the chart to zoom like the example is maybe an option ?

5 minute interval is 12 samples/hour * 31 = 372 data points... maybe that is not to bad.
We can also set the interval to 1 minute, then it might be a different story

lechunlx:
12 samples per h x 24 x 31 gives 8928
that's the reason for 1h sample if we set up above 14days, so it gives 24x31 -> max 744 datapoints for whole month.
In worst scenarions:
7days -> 12x24x7=2016
14days -> 2x24x14=672
31 days -> 24x31 = 372.
So it looks, that even for 7 days (as could be set right now) it's very much.

btw. I have only one temp sensor connected and logs from 10 days - my database is 696Kb.

gizmocuz:
1 temp sensor is nothing... thats why it might be best if power users who needs this, poke this in the database directly

gizmocuz:
My devices tab:

image

gizmocuz:
that is quite much but from multiple locations too

lechunlx:
And could you check your DB size for so many devices and 2days history?
I'll also try to flood my database with some data.

gizmocuz:
Don't think it is a wise idea to allow that many short log history.
You can always set this directly in the database on your own risk

PR #2192, opened 2018-03-05
Update ASyncTCP::connect to reset IO service if stopped

emontnemery:
If ASyncTCP::disconnect() is called from subclass to ASyncTCP, any subsequent calls to ASyncTCP::connect() would fail.
Fixed by checking if the IO service is stopped in ASyncTCP::connect, and if it is, reset it.
This was an issue meaning keep-alive implementation fails to reconnect:
- EvohomeTCP (Based on source code, I don't have this HW)
- RFLinkTCP (Tested and verified with RFLink)
- ZiBlueTCP (Based on source code, I don't have this HW)

If it's not safe to reset the IO service from ASyncTCP::connect(), the affected HW implementations can be modified instead to not call disconnect() unless hardware is stopping.


gizmocuz:
Great patch! Thanks!

PR #2193, opened 2018-03-05
Send most missing notifications for devices handled by plugins

guillaumezin:
This corrects the problem described here: http://www.domoticz.com/forum/viewtopic.php?f=65&t=22438

For plugin devices that are not switch (for instance pressure, temperature, etc), notifications are now sent


guillaumezin:
I'm reworking the PR to have all notifications handled in one place, it will be easier to maintain

guillaumezin:
Ok, done, but someone should review the code

gizmocuz:
Thank you for this patch!! Looking good, but some small remarks:

- Please start function names with a capital letter

I think it is best to not make new functions in the mainworker, but in the Notification Base class
So we keep 1 function

m_notifications.CheckAndHandleNotification(....)

guillaumezin:
I'm not sure to understand, you want me to overload CheckAndHandleNotification() ? Or create a new function, like ManageNotification (with capital M ;-) )?

guillaumezin:
@gizmocuz ok, moved to NotificationHelper

gizmocuz:
Great thanks! Does that mean that all (except 1) CheckAndHandleNotification functions can be private ?

guillaumezin:
CheckAndHandleTempHumidityNotification, CheckAndHandleDewPointNotification, CheckAndHandleValueNotification, CheckAndHandleRainNotification and CheckAndHandleAmpere123Notification aren't used anymore outside of NotificationHelper.cpp

CheckAndHandleSwitchNotification and all CheckAndHandleNotification are used outside, by SQLHelper, mainworker, PythonObjects, Kodi, LogitechMediaServer and PanasonicTV

guillaumezin:
@gizmocuz I moved the above-mentioned functions to private

main/mainworker.cpp
```diff
@@ -9815,39 +9767,40 @@ void MainWorker::decode_General(const int HwdID, const _eHardwareTypes HwdType,
...
 		if (DevRowIdx == -1)
 			return;
-		m_notifications.CheckAndHandleValueNotification(DevRowIdx, procResult.DeviceName, pMeter->intval2);
 	}
 	else if (subType == sTypeTextStatus)
 	{
-		DevRowIdx = m_sql.UpdateValue(HwdID, ID.c_str(), Unit, devType, subType, SignalLevel, BatteryLevel, pMeter->text, procResult.DeviceName);
+		strncpy(szTmp, pMeter->text, 200);
+		szTmp[199] = '\0';
+		DevRowIdx = m_sql.UpdateValue(HwdID, ID.c_str(), Unit, devType, subType, SignalLevel, BatteryLevel, szTmp, procResult.DeviceName);
```
gizmocuz: Not sure why this is put in a temp variable, and pMeter->text is not used?

gizmocuz (commented):
One small question about a temp variable usage....
You would almost say that 'CheckAndHandleNotification' should only be put in m_sql.UpdateValue ?

main/mainworker.cpp
```diff
@@ -9815,39 +9767,40 @@ void MainWorker::decode_General(const int HwdID, const _eHardwareTypes HwdType,
...
 	}
+	m_notifications.CheckAndHandleNotification(DevRowIdx, HwdID, ID, procResult.DeviceName, Unit, devType, subType, cmnd, szTmp);
```
guillaumezin: Because I want only one call to CheckAndHandleNotification aftewards, and it needs szTmp to be filled, but pMeter->text could still be used on UpdateValue line

guillaumezin:
And for the idea to call CheckAndHandleNotification in m_sql.UpdateValue, I'm not sure it will work in any places, look at decode_Hum() for instance

main/mainworker.cpp
gizmocuz: Than i would just use strcpy as pMeter->text is (now) 64 chars

gizmocuz:
"And for the idea to call CheckAndHandleNotification in m_sql.UpdateValue, I'm not sure it will work in any places, look at decode_Hum() for instance"

It would have been very nice.... but i see what you mean. In this case, there is one hardware device (433Mhz) that is a Temp+Hum device, but this one transmits the temp and hum in two separate messages which i am trying to combine here...

gizmocuz:
Thanks ! I will now try to test each and every notification possible

gizmocuz:
Found some small issues that i am addressing at the moment

In a lot of switch cases, you did not call SplitString (CNotificationHelper::CheckAndHandleNotification), like pTypeWind, i will modify the code to always split the string at the beginning

If possible, next time also test all notifications possible via the JSON/API (udevice)

gizmocuz:
For pTypeTEMP_BARO, you where calling a CheckAndHandleRainNotification ;)

gizmocuz:
Air Quality was checked against NTYPE_TODAYCOUNTER

gizmocuz:
pTypeLux was not implemented, there seems to be a default catch that is working on cSubType, these are from type pTypeGeneral 

Made a commit with some changes, is it possible you could verify this ?
I tested as many notifications as possible

https://github.com/domoticz/domoticz/commit/6031a4f26a220414b3a0b155bb014c8b858633e3

guillaumezin:
I agree on most of this, I tried to review my code a few times, at different times, and did tests only on some devices, it wasn't enough obviously, I didn't take the time to build a complete test setup...

See my comments in code review.

gizmocuz:
Thanks for the review! I will do make adjustments..

gizmocuz:
This should cover it ?

https://github.com/domoticz/domoticz/commit/c863736949e36c8f2e8bae2afac8c0a252350844

gizmocuz:
(lets see if we get those non-handled log messages)

guillaumezin:
Wow, this is... violent! You removed all DecodeRXMessage, that are used to rebuild message and update values though sql calls. Honestly, I found the code path very... disturbing

gizmocuz:
The whole handling of decoderxmessage was to get the notifications to work, which by your patch is quite easy
Do you see different behavior ?

PR #2194, opened 2018-03-06
Fix travis config and fix Selector Switch type change

SoaR245:


SoaR245:
Travis builds sill failed. Need to work on it before next PR

PR #2197, opened 2018-03-06
Fix Travis build on macos x

SoaR245:
As explained in the homebrew blog, the formula for python was changed. https://brew.sh/2018/01/19/homebrew-1.5.0/

This is also discussed in this issue https://github.com/Homebrew/homebrew-core/issues/24702


gizmocuz:
Thanks!

PR #2198, opened 2018-03-06
Fix selector switch change

SoaR245:
When editing a Selector Switch, changing the type from Selector to any
other type delete the device options which cause some errors. To avoid
this, the device options are initialize, as the webserver does, when the
device is edited and the device options are required.
To reproduce :
- create a dummy selector
- Edit this device
- Change the type to Dimmer
- Save
- Edit again the device
- Nothing happened because the code try to call the split function on an
undefined variable.


gizmocuz:
Please undo the travis fix as they are two different issues (and the travis PR is already merged ;)

gizmocuz:
Thanks!

PR #2206, opened 2018-03-10
Logitech Media Server: Add option to remove unused nodes

EddyK69:
Added an option to the LMS hardware-setup to remove unused nodes from the WolNodes table.
Unused nodes (device) are obsolete devices that are are not recognized by the LMS server.


gizmocuz:
Thanks! Is it possible in the feature to also update the internal appcache version number when making changes to the www folder? (i will make a commit that will do so in a second)

EddyK69:
Ah sorry. Forgot that ;) Thnx

PR #2213, opened 2018-03-12
Add support of device setpoint timers to timers editor page

stas-demydiuk:
Add support of device setpoint timers to timers editor page
-- remove setpoint timers editor from utility page
-- replace links with buttons to proper support "disabled" property

To make timers table more universal it was required to change "Value" column for setpoint timers
Before:
setpoint_before
After:
setpoint_after


gizmocuz:
Another great patch ! Thanks!!

PR #2215, opened 2018-03-12
Work on notifications

guillaumezin:
Better log of unhandled notifications, some corrections and some changes for clarity

### DIFF
--- a/hardware/plugins/PythonObjects.cpp
+++ b/hardware/plugins/PythonObjects.cpp
@@ -923,7 +923,9 @@ namespace Plugins {
 				self->TimedOut = iTimedOut;
 			}
 			
-			m_notifications.CheckAndHandleNotification(DevRowIdx, self->HwdID, sDeviceID, sName, self->Unit, iType, iSubType, nValue, sValue);
+			if (!IsLightOrSwitch(iType, iSubType)) {
+				m_notifications.CheckAndHandleNotification(DevRowIdx, self->HwdID, sDeviceID, sName, self->Unit, iType, iSubType, nValue, sValue);
+			}
 
 			CDevice_refresh(self);
 		}
--- a/main/SQLHelper.cpp
+++ b/main/SQLHelper.cpp
@@ -5193,7 +5193,7 @@ void CSQLHelper::UpdateMeter()
 			{
 				sprintf(szTmp,"%d",nValue);
 				sValue=szTmp;
-				m_notifications.CheckAndHandleNotification(ID, devname, dType, dSubType, NTYPE_USAGE, (float)nValue);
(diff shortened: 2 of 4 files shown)


gizmocuz:
Thanks, that could have been done by a few lines of

_log.Log(LOG_ERROR, "Notification NOT handled (type: %d, subtype: %d), please report on GitHub!", cType, cSubType);

But i see a lot of other changes here as well, there is no need to cast a float to a const float, or a int to a const int
Could you undo these changes ?

gizmocuz:
going to add a small addition

gizmocuz:
Not sure if we need to cast every integer to a float, why is this ?
In the end, i think that a double is better then a float in the notification routines because of rounding issues

guillaumezin:
The cast is for clarity, to understand which CheckAndHandleNotification (nValue or fValue version) si called, it is not mandatory, it is for readability

gizmocuz:
no sorry, not what i meant. We have uint32 values and integers and you are casting them now to a float.
(like counter1 and counter2, but also others that where initially an integer)

guillaumezin:
I don't know where it comes from, your local repo I think, but the whole pTypeGeneralSwitch in NotificationHelper is pointless for now, because I added a condition `(!IsLightOrSwitch(cType, cSubType)) {
		return false;
	}` at the beginning

gizmocuz:
your changes like MainWorker.cpp line 10462

gizmocuz:
Removed 1 IsLightOrSwitch check in PythonObjects.cpp, there are no others anymore
We could call CheckAndHandleSwitchNotification so the whole notification handling/call could be done in UpdateValue

gizmocuz:
Currently you are handling pTypeLighting1/pTypeLighting2 in CheckAndHandleNotification but it can not be reached as you now check at the top is it's a light and then exit
Do those two types belong here?

gizmocuz:
"The cast is for clarity, to understand which CheckAndHandleNotification (nValue or fValue version) si called, it is not mandatory, it is for readability"
Sorry, this is not clear for me, and should not matter as we have different overloads

guillaumezin:
ok, no problem

guillaumezin:
> Currently you are handling pTypeLighting1/pTypeLighting2 in CheckAndHandleNotification but it can not be reached as you now check at the top is it's a light and then exit

I don't understand where this code comes from, same for `case pTypeGeneralSwitch`, I didn't write this part, this comes from a merge with your code

gizmocuz:
That's weird as i can't remember adding them, specially because i added the IfSwitch statement...

guillaumezin:
> Removed 1 IsLightOrSwitch check in PythonObjects.cpp

Right, I forgot to commit this one

guillaumezin:
> That's weird as i can't remember adding them, specially because i added the IfSwitch statement...

I looked everywhere in my code, it's not from me. My evil twin maybe, or yours ;-) Or booze...

PR #2217, opened 2018-03-12
fix cmake configure warning with cmake >= 3.3

gordonb3:
CMakeLists.txt references a debug variable that should no longer be used with cmake >= 3.3


gizmocuz:
Thanks!!!

PR #2218, opened 2018-03-12
Fix for some sensors

Olifant1990:
First image added is old situation
Second is new

Some changes:
- The kWh sensor
1. When refreshed, the today item had gone down a bit. in new.png you see everything will stay at the same position
2. Added the total value back to the view
3. When a device doesn't have data for today, don't show the total in the today value, but 0. This is for my "hoogtarief" meter (peak-hour sensor) as in the weekend it doesn't record usage at all and showed the counter instead...

- The gas sensor
1. Show total counter
old
new


gizmocuz:
I do not see a need for the total counter for the kWh sensor, in fact i removed that recently.
Yes, some people would like to see this, but probably those people are the once that have a 'return' meter, and not a 'usage' meter
My kWh sensor is now on 123.456 kWh ... next week 456.789 ... i don't know the value from last week...
So i go to the 'report' view where all is available when i 'need' to look at it

Olifant1990:
Ok, I will revert the change with the total counter in the view for kWh and gas again soon.
And the other changes are ok?
Point 1 and 3 of kWh?

gizmocuz:
Sure, but i think only point 1 remains ?

Olifant1990:
No, point 3 still exists.
If it is sunday, you have 1 full day the on-peak hours meter didn't receive updates, because it is not active.
Then it will display "Today: <full counter value> kWh". What I expect and what I have solved: It will display "Today: 0 kWh" when there is no today value.

gizmocuz:
Correct, sorry that's true! Yes please!

Olifant1990:
Ok, I reverted the changes for the total counters in the view.

gizmocuz:
Thanks!

Issue #1823, opened 2017-09-24
AreWeInLocalNetwork and Nginx


NewFolk:
Hello,
Looks like function AreWeInLocalNetwork 
in [cWebem.cpp](https://github.com/domoticz/domoticz/blob/1b19fd0d1b58e08342a03ec3e31e20825f8dee43/webserver/cWebem.cpp#L1131) works only with local proxy server.

I have Nginx as reverse proxy and configuration like

proxy_pass https://192.168.1.10:443; #DomoticzIP
proxy_set_header Host $host;
proxy_set_header X-Real-IP $remote_addr;
proxy_set_header X-Forwarded-For $remote_addr;

And settings Local Networks (no username/password):
127.0.0.1, 192.168.1.*

It does not ask password without depends if Im logging from intranet or from internet.

if I disable Local Networks settings in logs I see `Login successful from 192.168.1.1 for user 'XXXX'` #Router with Nginx

gizmocuz:
In the latest commit i moved the code
could you try removing the following line in

cWebemRequestHandler::handle_request

remove:
if (session.remote_host == "127.0.0.1")

gordonb3:
You wouldn't want to do that. This would allow anyone to spoof IP addresses by simply setting the appropriate proxy headers. What you need to do is replace the hardcoded localhost IP with the one belonging to the host that runs your Nginx server. Preferably this should be a configurable value in the settings screen.

NewFolk:
Yes, this will be a great option.

gizmocuz:
if the original socket is received on the localhost (127.0.0.1) which can not be spoofed by a header, we check for the existence of the NGINX header that is added in the NGINX (server) configuration file
I do not believe this is a header anyone can add right ?

gordonb3:
In essence anyone can add any header. We actually do that ourselves in various parts of the code to pass authentication keys etcetera. This is why you want to limit which callers may define another origin and possibly bypass authentication by doing so.

However by hardcoding the localhost address this limits the possibility of running a front end web server to be on the same machine as Domoticz only. By allowing a different IP address to be specified here one can offload the Nginx service and any other pages served by it (e.g. Wordpress) to different hardware without breaking logging functionality or authentication.

gizmocuz:
Sorry i was not talking about http headers but the headers that have to be manually added to the nginx configuration file like you can see above
Surely i would suspect nginx would filter out fake nginx headers as well?

gordonb3:
[...]

Issue #172, opened 2015-10-27
Critical Security Issue : Denial of Service caused by stalled HTTP sessions


gaudryc:
We should manage :
- the session timeout (the session should be deleted from the map but not from the database) ;
- the session expiration (the session should deleted from the database when the session expired).

gaudryc:
Please see https://github.com/domoticz/domoticz/pull/173

gizmocuz:
patch applied

PR #2223, opened 2018-03-14
Allow remote proxy

gordonb3:
In response to issue #1823: AreWeInLocalNetwork and Nginx

This update allows users to specify a list of IPs that are allowed to act as a proxy. The additional logging patch will show the originating (proxied) IP when a new session is started rather than just once for each unique IP.

### DIFF
--- a/main/WebServer.cpp
+++ b/main/WebServer.cpp
@@ -328,6 +328,20 @@ namespace http {
 				}
 			}
 
+			std::string WebRemoteProxyIPs;
+			int nValue;
+			if (m_sql.GetPreferencesVar("WebRemoteProxyIPs", nValue, WebRemoteProxyIPs))
+			{
+				std::vector<std::string> strarray;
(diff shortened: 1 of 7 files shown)


gordonb3:
Hold on... session expiration does not appear to update without logon

gordonb3:
Looks like something is f*ed up with session management. I can see the session refreshing every 5 minutes but it still starts a new session every 15 minutes...

Edit: 
Okay, that is weird. Has nobody else had issues with this before? There is in fact a mixup with two time values: `timeout` and `expires`. The latter one gets refreshed while the session is kept alive by the remote user, but the first one is only set at creation time and subsequently used in a periodic cleaning routine that will kill any session that was created more than 10 minutes ago, regardless of the value of `expires`.

Edit 2:
Right... This behaviour appears to be related to issue #172 and I need to reload the session info from store. Verifying solution right now.

gordonb3:
All working now. Good to go

gizmocuz:
Thanks!

PR #2225, opened 2018-03-15
Correction in pTypePower, sTypePercentage, pTypeUsage + silently ignore update notifications for general devices

guillaumezin:
because it seems to be a pain for users

I improved the log when the array size is not the expected one.

Notifications tested on every dummy devices.

Currently pTypeBARO,  pTypeWeight and pTypeP1Gas are not handled, but wasn't previously

Moved call to m_notifications.CheckAndHandleNotification() in MainWorker::UpdateDevice() to prevent 2 calls to CheckAndHandleNotification() for Thermostat SetPoint

### DIFF
--- a/main/mainworker.cpp
+++ b/main/mainworker.cpp
@@ -12919,9 +12919,6 @@ bool MainWorker::UpdateDevice(const int HardwareID, const std::string &DeviceID,
 	std::stringstream s_strid;
 	s_strid << std::hex << DeviceID;
 	s_strid >> ID;
-	uint64_t dID = 0;
-	std::string dName = "";
-
 
 	if (pHardware)
 	{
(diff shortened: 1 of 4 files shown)


main/mainworker.cpp
```diff
@@ -12913,6 +12915,10 @@ bool MainWorker::UpdateDevice(const int HardwareID, const std::string &DeviceID,
 {
 	CDomoticzHardwareBase *pHardware = GetHardware(HardwareID);
 
+	// Prevent hazardous modification of DB from JSON calls
+	if (!m_sql.DoesDeviceExist(HardwareID, DeviceID.c_str(), unit, devType, subType))
+		return false;
+
```
gizmocuz, line 12921: Extra tab here

main/mainworker.cpp
```diff
@@ -12928,7 +12934,7 @@ bool MainWorker::UpdateDevice(const int HardwareID, const std::string &DeviceID,
 
 		uint64_t dID = 0;
 		std::string dName = "";
-
+
```
gizmocuz, line 12937: extra tabs here

gizmocuz:
Thanks!, Merged ! There are a few extra tabs added on two lines

PR #2227, opened 2018-03-15
Fixed creation of new blinds type switch by RFLink

lechunlx:
When creating switch for blinds by "Manual Light/Switch" it create it with lowercase ID and without leading zeros. Therefore, if you "click" for such device, RFLink will create another one with proper ID (uppercase with leading zeros). This change fully repair that issue.


gizmocuz:
Thanks!

PR #2230, opened 2018-03-17
New correction in Multibloc management

ghost:
* fixed bug in the checkstates function (which always return true, so the witch is always update).
* fixed bug in the "learning entry method" / now we can see wich output is in learning by On/Off states of the switch associated.
subsidiary question: is there a way to show that an output (switch) is in a blink mode ? (because an output have a state bit: "ON" and another bit: "Blink ON", the blink mode is about 1sec).


gizmocuz:
Thanks!

ghost:
Hi gizmocuz ! do you think there is a way to find solution to my question ? ^^

PR #2231, opened 2018-03-18
Tiny bug in void CWebServer::RType_Plans

lechunlx:
To be honest, I don't know the purpose of this code and whether it's really bug, but doesn't looks legit for me. Found accidentally.


gordonb3:
Hmmm.... That's not a bug. That's just sloppy cut-and-paste programming. Obviously the original test will always return true as it is the condition for the loop in which it resides to be started. The replacement test is rather pointless as well though because the associated query will always return a result, unless of course at some point during development the query itself becomes invalid. You should simply drop the test completely in this PR

gizmocuz:
Not my code, but the fix looks OK, thanks !!

PR #2240, opened 2018-03-24
Own pages for light and temperature/humidity logs

stas-demydiuk:
Routing improvements for device logs + code reorganization

- Extract LightLog to own page
- Extract Temperature/Humidity Log to own page
- Extract Temperature/Humidity Report to own page
-- Add Temperature/Humidity variation chart
- General refactoring:
-- Add back-button directive for backButton logic encapsulation
-- Add translate directive for translation logic encapsulation
-- Add deviceApi factory for device-related api
-- Add dataTableDefaultSettings constant with common dataTables settings across the app
-- Move/refactor device logs-related logic from domoticz.js to corresponding files
-- Move device logs-related templates from index.html to corresponding views

**Temperature/Humidity Report page**
Before:
temperature_report_before
After:
temperature_report_after


gizmocuz:
Thanks for this super great patch ! I only miss the highlighted (red) month now
The Temperature variation is already in the table, instead of the new chart at the bottom, maybe this could be shown in the first chart? (but keep the bars verticle please)

stas-demydiuk:
Hey @gizmocuz, Originally the red highlighting of the month meant the month with the maximum temperature. It was quite unclear for me why do we want to highlight it, why for example max temperature is more important than min temperature (as we don't highlight it). So I decided to add this new chart with temperature variation and now it's quite easy to find the month or day (depending on view) with min or max temperature, so I removed highlighting from the original chart.

Honestly I don't like the idea of replacing the chart with average data in favor of variation (if I correctly got your point) as most likely the average value is what you ale looking for or at least nice to see feature, so I propose to leave both. But in any case I'm open for suggestions.

Here is the example of how this average chart will look like with vertical bars:
image

From my point of view its more informative with horizontal bars, but it could be only my opinion :)
Please let me know your thoughts and I'll update the PR

stas-demydiuk:
Or as an option we could combine these 2 charts into something like this:
image

gizmocuz:
Like your second/last chart, but please make it the same as we have on the temperature charts (looks)
I also agree (and thanks for the quick sample!) with the first chart look, it should be the same
I like the (liked) the idea of showing the hottest/coolest day of the year with a color, maybe we could set the value's color to blue(ish) for coldest day/month, and red(ish) for hottest ? (not average, but based on the real day value i suppose, coz the day values are already avaraged)

stas-demydiuk:
image

gizmocuz:
Thank you very much for this great patch !

gizmocuz:
There is a small issue, when i go to report, February, the graph below (min/max/avr) is from March
When i select march, then the graph is from April

stas-demydiuk:
@gizmocuz, yes looks like conversion is wrong, here the fix #2252

PR #2241, opened 2018-03-26
Changed polling time (again).

denkovi1:
Polling time changed to 1 sec as 5 sec is still reported for slow by some users.

### DIFF
--- a/hardware/DenkoviSmartdenIPInOut.cpp
+++ b/hardware/DenkoviSmartdenIPInOut.cpp
@@ -9,7 +9,7 @@
 #include "../main/SQLHelper.h"
 #include <sstream>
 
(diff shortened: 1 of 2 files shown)


gizmocuz:
Thats way to short....

denkovi1:
Is there any way how to make this polling time as configurable parameter (like IP, Port and pass) ? By default it could be 5 seconds and the user may decrease it to 1 sec (if need). We have many users reports that the polling time is still long and we thought it might be good idea to make this parameter configurable.
Thanks

gizmocuz:
Sure, there are other hardware classes that do this, let me try to implement this for you

gizmocuz:
Done, see here

https://github.com/domoticz/domoticz/commit/fea48d46d8dff437ef494ff78f8b9f48d9d28cc9

closing this PR

PR #2245, opened 2018-03-27
correct timer calculations when crossing a DST change

gordonb3:


gizmocuz:
Thanks! This seems like a never ending battle ;)

gordonb3:
Yeah...

The idea to not compensate for DST was good, but as it turns out we were in fact still compensating. There is also one new item added, because to my surprise my on-time timer that was set at 2:30 AM (non-existing time on Sunday) fired at 3:30 AM on Monday as well. I left a note on that in the code, so you probably read it.

PR #2246, opened 2018-03-28
limit amount of logging of localhost scripts as incoming connections

gordonb3:
Because these scripts don't accept and send cookies we can currently not track them and they are always seen as new connections, causing massive logging with some users as of 6b1f93d

This is a temporary fix until I can come up with something better.

### DIFF
--- a/webserver/cWebem.cpp
+++ b/webserver/cWebem.cpp
@@ -1458,6 +1458,7 @@ static void GetURICommandParameter(const std::string &uri, std::string &cmdparam
 bool cWebemRequestHandler::CheckAuthentication(WebEmSession & session, const request& req, reply& rep)
 {
 	session.rights = -1; // no rights
+	session.id = "";
 
 	if (myWebem->m_userpasswords.size() == 0)
 	{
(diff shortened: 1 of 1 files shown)


gizmocuz:
I don't mind logging it (as then you know some local process is doing something with your domoticz instance), but only when:
- session has expired
- or at startup of domoticz, IP address is not known
- or not seen for timeout period

This patch should maybe check/use the function 'cWebemRequestHandler::AreWeInLocalNetwork' as you can also specify that for instance 192.168.0.10 is allowed to connect without authentication/script, or even a IFTTT remote's IP address
Still logging the incoming connection is OK for those

gordonb3:
Working on a little idea. Trace looks good so far. Running some additional checks

gordonb3:
Took a bit more time than expected. I kept running into the automatic dropping of 'live' sessions which is impossible to recover from with scripts that do not identify themselves.

I'm unsure about the point of this mechanism and how this could lead to a DOS as stated by the person that submitted this part of the code. Given the conditions that are required to have the code accept traffic from a specific IP as belonging to a session I do not think overruling this timeout mechanism will cause any issues as these will be very limited.

gizmocuz:
Thanks!

PR #2229, opened 2018-03-15
Improved color dimmer support

emontnemery:
As discussed in #2078, improved support for color LED dimmers.

2018-03-29:
Updated binaries matching commit fae2ff0c78c0879b2ac61afe439ea8910df0ac40:
- Linux x86 64 binary [domoticz_linux_x86_64.tgz](http://montnemery.com/domoticz_rgbww/linux/133.1/fae2ff0c78c0879b2ac61afe439ea8910df0ac40/domoticz_linux_x86_64.tgz)
- Windows binary: [domoticz_windows_x86.zip](http://montnemery.com/domoticz_rgbww/windows/1.0.139/fae2ff0c78c0879b2ac61afe439ea8910df0ac40/domoticz_windows_x86.zip)

**Summary:**
- Color picker for dimmer updated with adjustable color temperature
- Possible to set colors other than fully saturated from Web UI
- Set color of device is stored in database and available through JSON API
- Full color data used internally by Domoticz instead of hue + white level
- Hue stored in existing scenes and timers is translated during database upgrade
- New Subtypes added to Limitlessled:
  - WW (white with adjustable color temperature)
  - RGBWZ - RGB + White, where RGB and White can be turned on at the same time
  - RGBWWZ - RGB + Warm white + Cold White, where RGB and White can be turned on at the same time

**API changes:**
- Field "Color" added to "devices" web API
- Filed "Color" added to MQTT status update from Domoticz
- Web API command "setcolbrightnessvalue" now accepts JSON formatted color data as well as hue
- Web API commands to set/get timers and set/get scene devices now accepts JSON formatted color data instead of hue
- JSON formatted color data passed to Python plugins instead of Hue

**Open issues / Not yet implemented:**
- Consider improving usability of the updated color picker in custom mode, it is not very intuitive, maybe grouping of the sliders would improve it
- Testing of the various HW is needed!

**HW status:**
Arilux:
- RGBWW / CW_WW hardware: Yes, according to comments in the source code
- RGBWW / CW_WW Domoticz support: No
- Test results: Not tested

Blebox:
- RGBWW / CW_WW hardware: No RGBWW devices according to homepage: http://blebox.eu/en/12-products
- RGBWW / CW_WW Domoticz support: No
- Test results: Not tested

EHouse:
- RGBWW / CW_WW hardware: ?
- RGBWW / CW_WW Domoticz support: No
- Test results: Not tested

Limitless:
- RGBWW / CW_WW hardware: Yes, RGBWW lights
[...]

PR #2249, opened 2018-03-28
Allow plugin devices to trigger a scene or device just like any other device

emontnemery:
Allow plugin devices to trigger a scene or device just like any other device.

Also adds new option SuppressTriggers to Device.Update, which if set will not call:
- m_sql.UpdateValue
- m_mainworker.CheckSceneCode
- m_mainworker.sOnDeviceReceived
- m_notifications.CheckAndHandleNotification

This is needed to be able to, for example, update SignalLevel or Battery without triggering some script or have some other unwanted side effect.

@dnpwwo Please review, another option could be to add new method `Device.ChangeSetting` which would be same as `Device.Update` but would not take nValue / sValue

### DIFF
--- a/hardware/plugins/PythonObjects.cpp
+++ b/hardware/plugins/PythonObjects.cpp
@@ -808,16 +808,17 @@ namespace Plugins {
 			uint64_t 		DevRowIdx;
 			char*		Description = NULL;
 			char*		Color = NULL;
+			bool		SuppressTriggers = false;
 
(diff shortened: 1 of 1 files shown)


emontnemery:
Rebased to fix merge conflict after merge of #2229

dnpwwo:
Happy with this approach.

`Device.ChangeSetting` is a good idea but the current interface has been in the stable release for a while and not breaking backwards compatibility is a good idea where there is no compelling case to do so.

PR #2250, opened 2018-03-29
repair remember me functionality in webserver

gordonb3:
I accidentally killed that earlier because that line in d8b3c70 didn't seem to make sense.


gizmocuz:
Thanks!

PR #2255, opened 2018-04-01
Fix #2135 Multiple Oregon THN132N are seen as the same device with a RFPlayer

### LINKED ISSUE #2135: Multiple Oregon THN132N are seen as the same device with a RFPlayer
I have 2 THN132N setup with different channels but they share the same id (id EA4C). I checked without the RFPlayer software and it sends a field with the channel. I guess this information is not used by domoticz. It's probably the same for others Oregon products.

ultrasuperpingu:
Fix issue #2135 
Device Ids will change so logs of those devices will be uncorrelated from new device. You will need to manually update ids in tables with sqlite3


gizmocuz:
Thanks!!

PR #2262, opened 2018-04-02
Fix crash in CPluginTransportICMP::handleTimeout

emontnemery:
Fix crash introduced in 18185a9a41cedcb5d274f4350a5e897a7d4005c0


dnpwwo:
I've never seen an error go through this code (apart from operation_aborted).

What error trigger this?

PR #2271, opened 2018-04-02
Add API for managed meters

guillaumezin:
This PR add JSON RPC calls and a new Python plugin function to add directly data to Meter and Meter_Calendar DB

Some french meters (Linky, Gazpar) grabs data that are available 24hours later, these API allows to include them in Domoticz DB without accessing it the bad way (editing domoticz.db from an external script).

JSON RPC calls:
For day view:
 json.htm?type=command&param=umeter&idx=14&value=32&date=2018-04-30&usage=2&day=1
For month/year view:
 json.htm?type=command&param=umeter&idx=14&value=32&date=2018-04-30&counter=3
And for multimeters:
 json.htm?type=command&param=umeter&idx=14&value=32&date=2018-04-30&min=2&max=4
Erase data (for day view and for month/year view):
 json.htm?type=command&param=umeter&idx=14&day=1&clearafterdate=2018-03-31&clearbeforedate=2018-06-01
 json.htm?type=command&param=umeter&idx=14&clearafterdate=2018-03-31&clearbeforedate=2018-06-01

Python calls:
For day view:
 Devices[14].UpdateMeter(Day=True, Values={'Value':12.2,'Date':'2014-04-02 18:14:30','Usage':2.0})
For month/year view:
 Devices[14].UpdateMeter(Day=False, Values={'Value':12.2,'Date':'2014-04-02 18:14:30','Counter':3.0})
And for multimeters:
 Devices[14].UpdateMeter(Day=False, Values={'Value':12.2,'Date':'2014-04-02 18:14:30'},'Min':2.0,'Max':20.0)
Erase data (for day view and for month/year view):
 Devices[self.iIndexUnit].UpdateMeter(Day=True, ClearAfterDate="2018-03-01 01:00:00", ClearBeforeDate="2018-04-30 20:00:00")
 Devices[self.iIndexUnit].UpdateMeter(Day=False, ClearAfterDate="2018-03-01 01:00:00", ClearBeforeDate="2018-04-30 20:00:00")

There will be a separate PR (#2272) to add an option on devices meter, preventing Domoticz to do average, transfer from Meter to Meter_Calendar, because it is useless for these kinds of meter.

### DIFF
--- a/main/SQLHelper.cpp
+++ b/main/SQLHelper.cpp
@@ -5149,17 +5149,17 @@ std::string CSQLHelper::getMeterFilter()
 	return result;
 }
 
-bool CSQLHelper::UpdateCalendarMeter(const int HardwareID, const char* DeviceID, const unsigned char unit, const unsigned char devType, const unsigned char subType, float value,  const char* date, float usage, float counter, float min, float max, const bool dayCalendar, const char* clearAfterDate, const char* clearBeforeDate)
+bool CSQLHelper::UpdateCalendarMeter(const int HardwareID, const char* DeviceID, const unsigned char unit, const unsigned char devType, const unsigned char subType, bool validValue, float value,  const char* date, float usage, float counter, float min, float max, const bool dayCalendar, const char* clearAfterDate, const char* clearBeforeDate)
 {
 	std::vector<std::vector<std::string> > result;
 	std::string filter = getMeterFilter();
 	char szTmp[200];
 	
-	if((date == NULL) || (date[0] == '\0')) {
+	if(validValue & ((date == NULL) || (date[0] == '\0'))) {
 		return false;
 	}
 	
-	result = safe_query("SELECT ID FROM DeviceStatus WHERE (HardwareID=%d AND DeviceID='%q' AND Unit=%d AND Type=%d AND SubType=%d AND (%s))", 
+	result = safe_query("SELECT ID, Name, SwitchType FROM DeviceStatus WHERE (HardwareID=%d AND DeviceID='%q' AND Unit=%d AND Type=%d AND SubType=%d AND (%s))", 
 		HardwareID, DeviceID, unit, devType, subType, filter.c_str());
 		
 	if (result.empty()) {
@@ -5172,11 +5172,18 @@ bool CSQLHelper::UpdateCalendarMeter(const int HardwareID, const char* DeviceID,
 	std::stringstream s_strid;
 	s_strid << sd[0];
 	s_strid >> DeviceRowID;
+	std::string devname = sd[1];
+	_eSwitchType switchtype=(_eSwitchType) atoi(sd[2].c_str());
 	
 	if (dayCalendar)
 	{
 		if((clearBeforeDate != NULL) && (clearBeforeDate[0] != '\0') && (clearAfterDate != NULL) && (clearAfterDate[0] != '\0')) {
-			safe_query("DELETE FROM Meter WHERE (DeviceRowID=='%" PRIu64 "') AND (Date<='%q') AND (Date<='%q')", DeviceRowID, clearBeforeDate, clearAfterDate);
+			_log.Log(LOG_ERROR,"DELETE FROM Meter WHERE (DeviceRowID=='%" PRIu64 "') AND (Date>='%s') AND (Date<='%s')", DeviceRowID, clearAfterDate, clearBeforeDate);
+			safe_query("DELETE FROM Meter WHERE (DeviceRowID=='%" PRIu64 "') AND (Date>='%q') AND (Date<='%q')", DeviceRowID, clearAfterDate, clearBeforeDate);
+		}
+
+		if (!validValue) {
+			return true;
 		}
 
 		sprintf(szTmp, "%.0f", value);		
@@ -5220,9 +5227,14 @@ bool CSQLHelper::UpdateCalendarMeter(const int HardwareID, const char* DeviceID,
 		if (!isMultiMeter(devType, subType))
 		{
 			if((clearBeforeDate != NULL) && (clearBeforeDate[0] != '\0') && (clearAfterDate != NULL) && (clearAfterDate[0] != '\0')) {
-				safe_query("DELETE FROM Meter_Calendar WHERE (DeviceRowID=='%" PRIu64 "') AND (Date<='%q') AND (Date<='%q')", DeviceRowID, clearBeforeDate, clearAfterDate);
+				_log.Log(LOG_ERROR,"DELETE FROM Meter WHERE (DeviceRowID=='%" PRIu64 "') AND (Date>='%s') AND (Date<='%s')", DeviceRowID, clearAfterDate, clearBeforeDate);
+				safe_query("DELETE FROM Meter_Calendar WHERE (DeviceRowID=='%" PRIu64 "') AND (Date>='%q') AND (Date<='%q')", DeviceRowID, clearAfterDate, clearBeforeDate);
 			}
-				 
+
+			if (!validValue) {
+				return true;
+			}
+
 			//insert into calendar table
 			result = safe_query(
 				"SELECT DeviceRowID FROM Meter_Calendar "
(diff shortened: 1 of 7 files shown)


gizmocuz:
Could it also if we modify the udevice function (webserver, update device) to include a date/time ?

guillaumezin:
I'm not sure to understand. You want to use "udevice" for JSON RPC, instead of "umeter"? I think it's possible, but a bit messy, and I like the idea to separate udevice, that will change DeviceStatus, and umeter, that will change Meter, Meter_Calendar, etc. tables.

gizmocuz:
I don't think it is messy. You could use the udevice method as usual for updating virtual devices, only now the addition to specify the log-time, is this not clearer then creating new methods just for a specific strange meter that logs after 24h ?

guillaumezin:
As you wish!

guillaumezin:
@gizmocuz Is my proposal ok or did you have something else in mind?

gizmocuz:
I still think this whole managed meter thing is a bit out of the scope of domoticz, and makes the configuration harder to read for end-users
Those two meter types you mentioned... no idea what they are, have you got links for them, and to the API as well ?
A meter that you can not query every xxx minutes is a (for me) useless meter! Better to install a good meter if you would ask me.
We have the TeleInfo meter already implemented, which is a french meter... could this be installed ?

guillaumezin:
The linky meter is the standard meter installed in France for almost everyone right now. It has TeleInfo, but you need hardware to connect as TeleInfo. Linky has a link to energy provider using powerline carrier, for smart grid, and collected data are available to customer, but only the day after. It creates a way to get data without additional hardware. There already is an external plugin that grabs data and put them in domoticz DB, but the crappy way, by modifying the DB directly with SQL queries from an external script (https://github.com/empierre/domoticz_linky).

I already made a plugin for Linky, if you want to have a look at the way I grab data
[plugin.zip](https://github.com/domoticz/domoticz/files/1935899/plugin.zip)

For now, I only found description in french...

There will be the same kind of meter for gas (named Gazpar).

gizmocuz:
Are you 100% sure there is no way to query the current usage (and if possible counter value) in real time ?
That sounds strange.... you would also want to see online your current usage.
This is the only data you will need, and be pushed to domoticz, and domoticz does the rest.
I see there is an API (big API with lots and lots of functions)

For example it is possible to query it per hour:

https://github.com/Pirionfr/pyLinky/tree/master/pylinky

Possible also realtime ? (My french is not that good to debug this)

If you can query the previous hour (or last hour) that would work too, maybe the current hour will increment in realtime ?

guillaumezin:
No you can't, you can get kWh data for every 30 minutes window, but data are only available for the day before...

It's TeleInfo option for real-time values, but you need to add hardware.

gizmocuz:
How much is the hardware ?

gizmocuz:
Seems only 22 $ ?

guillaumezin:
And wires, and holes, and discussion with my neighbor because the meter is on his wall, etc... Or the wireless version, which is Zigbee or KNX, and you need a Zigbee or KNX bridge, or the WiFi version, announced at 50€, but not existing yet, so the price is not that sure.

Here we have a (degraded, I agree) solution for free for 35 millions of counters

goulou:
As a french domoticz user for 4 years now, I do agree 1000% with guillaumezin : I have many devices, all of them connected to domoticz through different carriers (2.4Ghz, RF 433, Wifi...), but the french power provider installed a "smart" meter in my house, just like it does in 35 million other houses : with this PR I forsee that I can win a new sensor for free, just by connecting my domoticz to the Internet (which is already the case anyway) and adding a little piece of software... And by the way, my counter is not on my neighbor's wall, but on the street : I have no way to get a Teleinfo device there (the box where the counter is located does not belong to me, there is no power, no possibility to add a link to the house...) so "just" using the smart meter capability of the counter would be of a great help! And I think that all french users will definitely appreciate the capability!
@guillaumezin : how do you feed the data to the sensors throught the JSON API? You have an external daemon for that? Do you plan on releasing its code?
Thanks for the work anyway, I hope I could use it soon!

guillaumezin:
@goulou: There is an example using the Domoticz python framework attached to 8th post, and it will be released with GPL license. JSON RPC calls and Python framework methods are described in the 1st post

gizmocuz:
OK , I totally understand. But... the code gets bad and unreadable with a 'managed' meter...

So you want to have a metering device that you will update via JSON and use a counter value, so a direct 'day' value in the database.
How about we create a JSON function to directly add/update a day value for a device instead ?
First check if there was already a day value for this device, if yes, update it, if no, add it

Then we have the problem that at 00:00 domoticz calculates a day value automagically based on the short-log, that is a problem for these counter devices yes.

So we need a new Meter type (MTYPE_MANAGED), this type should not calculate a day value at 00:00
This meter type should not display the short-log in the charts
This can be very easily done, and the code stays very clear.

Then you can add code to python so it can also use the day value update code above (or it can use the JSON API directly, i don't know as i don't use/know python)

The new JSON call can also be used maybe in the feature to overwrite/add a day value to a sensor (at 00:05am for example, or to synchronize master/client device)

Is this an idea ?

gizmocuz:
Better would be if you can query the counter and actual usage via their API... must be possible... did you explored the full API ? Or asked them ?

guillaumezin:
My first intention was to add a meter type, but I thought it will make the code unreadable: there are a lot of meter type with switch/case everywhere in the code, javascript to add... I tried not to add one type for that reason.

I want to add the value per hour, this is the most interesting part: we can grab data for the month before with 30min granularity. It is interesting because we can log that in the short log and see when we consumed power.

I didn't ask or explore the full API, it is quite new and reverse engineering (they log the data on their website), and it is quickly changing (they made changes yesterday, I must look at what changed).

gizmocuz:
Even better as they might be willing to add a new function.
If you could ask them for a call to get the current counter value + actual usage, then no code changes need to be made.

guillaumezin:
I'm just not sure that they get live data, my understanding (but I may be wrong), is that they collect data by batch, using powerline carrier.

gizmocuz:
Hmmm if they are smart, they should get a value at least once per hour, so they know how much the total usage is and maybe change some internal routing to safe some money.
But it would be great if you can query this. Of course they will not be happy when users will query once every 5 minutes (hope they allow this), but maybe once an hour an update would be great

guillaumezin:
ok, what if I propose to use the udevice API for JSON calls and Devices[x].Update() for Python, add a new device type, and use the sValue with something like counterValue;date for instance? If date is not specified, it is the value for today, if date contains date AND hour, it is for short log, if date contains, well, date only, it is for history log

gizmocuz:
Yes that would work as well! But still it would be best if you can request the actual counter value + usage

goulou:
> I'm just not sure that they get live data, my understanding (but I may be wrong), is that they collect data by batch, using powerline carrier.

As I understand it, they have a paid option for users where they provide actual live data. It is in there business model not to provide this to all users. So I doubt that they will add a JSON API for domoticz users...

gizmocuz:
I think it is the same as here, that they are ordered by law to give a good insight in your usage and that they should provide this!
In here, this can be done via the P1 smart port (a RJ11 connector)

goulou:
Well, to be more precise, in France they are ordered by law to give precise and (almost) real time values to low-tax payers (people with small revenues) in order to help them reduce there power consumption... People with greater incomes are authorized to waste electricity...

gizmocuz:
Hehe.... so they must make this available in their API then ?

guillaumezin:
Maybe I will be authorized to access these data. I have not so small incomes, but I'm part of low-tax payers (lot of kids...)

main/SQLHelper.cpp
```diff
@@ -5130,6 +5145,158 @@ void CSQLHelper::UpdateMeter()
...
+			safe_query(
+				"INSERT INTO Meter (DeviceRowID, Value, Usage, Date) "
+				"VALUES ('%" PRIu64 "','%lld','%lld','%q')",
+				DeviceRowID, MeterValue, MeterUsage, date
+			);
+		}
+		else
+		{
+			safe_query(
+				"UPDATE Meter SET DeviceRowID='%" PRIu64 "', Value='%lld', Usage='%lld', Date='%q' "
```
dwmw2, line 5217: AFAICT we don't seem to do any validation of `date` at any point in the proceedings. The user could insert complete garbage into the database, couldn't they?

guillaumezin: That's true, I'm working on a different approach where the update will be possible on a specific managed device

guillaumezin:
I'm working on a different approach where the update will be possible on a specific managed device as suggested by @gizmocuz, so I close this PR

Issue #2269, opened 2018-04-02
P1 Smart Meter Annual Reports are broken


MAkerboom:
Device type: P1 Smart Meter
Error: Go to device -> Log -> Report: No data is shown, error in browser console

The `ShowP1YearReport()` function gives a javascript error in the Chrome browser:
```
Uncaught TypeError: Cannot read property 'match' of undefined
    at u (jquery.dataTables.min.js:63)
    at Ha (jquery.dataTables.min.js:59)
    at Y (jquery.dataTables.min.js:12)
    at t.<anonymous> (jquery.dataTables.min.js:123)
    at t.iterator (jquery.dataTables.min.js:98)
    at t.<anonymous> (jquery.dataTables.min.js:122)
    at t.<anonymous> (jquery.dataTables.min.js:102)
    at t.visible (jquery.dataTables.min.js:100)
    at w.fn.init.fnSetColumnVis (jquery.dataTables.min.js:83)
    at Object.success (domoticz.js:9323)
``` 

Did a bit of debugging:
Hiding the unused table columns when not delivering energy results in the error above:
```
			oTable.fnSetColumnVis(5, bHaveDelivered);
			oTable.fnSetColumnVis(6, bHaveDelivered);
			oTable.fnSetColumnVis(7, bHaveDelivered);
			oTable.fnSetColumnVis(8, bHaveDelivered);
```

Don't know the proper fix, so can't make a pull-request.

Domoticz version info: 
```
Version: 3.9168
Build Hash: 5f4b6858
Compile Date: 2018-04-02 15:54:41
dzVents Version: 2.4.2
Python Version: 3.5.3 (default, Jan 19 2017, 14:11:04) [GCC 6.3.0 20170124]
```

gizmocuz:
I have no issues with the electic meter.
Does it work from the Utility page ?
Could you clear the browser cache AND appcache (two different things)

MAkerboom:
@gizmocuz 
Today I read your message at work and was surprised it worked on Windows 10: Chrome Stable and IE10.
So back home on my Chromebook running 'Version 65.0.3325.184 (Official build) (64-bits)' it's still broken, and can't fix it.

Steps I took to clear browser and app cache:
1. open dev-tool (`control-shift-i`)
2. application tab
3. clear storage menu item
4. all checkboxes selected and 'Clear site data'
I even bumped my `domoticz/www/html5.appcache` to be sure. Also tried 'Incognito mode'.

[...]

PR #2283, opened 2018-04-04
bump dataTables from version 1.10.8 to 1.10.15

MAkerboom:
#2269 

dataTables 1.10.8 gave a javascript error on the latest chrome browser when hiding columns with `.fnSetColumnVis()`

The latest dataTables version 1.10.16 is incompatible with the used dataTables styling. Thefor bumping to version 1.10.15 which is still compatible with the used styling.


gizmocuz:
Thanks!!

PR #2286, opened 2018-04-04
Fix memory leak in python event script code

teeminus:
Long story short: I accidentally introduced a memory leak in my last merge request (#2139) which added the output of python event script errors to the domitcz log. The domitcz process leaks memory every time a python event script is executed because my code did not free all created python objects. I didn't noticed the bug while testing my code back then because I didn't triggered enough events to notice the mem leak.

I'm sorry the bug I introduced to the domoticz code. This merge requests fixes the memory leak. I tested the fix by triggering ~200k events which did not lead to a significant increase of the memory usage.

### DIFF
--- a/main/EventsPythonModule.cpp
+++ b/main/EventsPythonModule.cpp
@@ -367,36 +367,36 @@
                    }
 
                    // Get message from stderr redirect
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thanks for the quick fix !

PR #2287, opened 2018-04-04
Update subtype of Hue, Yeelight and ZWave devices when subtype is not matching

emontnemery:
There were changes in Hue, Yeelight and ZWave integration as part of the color switch changes:
- Philips Hue: Exact device capability is discovered and mapped to matching color switch subtype
- Yeelight: Exact device capability is discovered and mapped to matching color switch subtype
- ZWave: Legacy RGBW switch mapped to sTypeColor_RGB_W_Z, color switch mapped to sTypeColor_RGB_CW_WW_Z

This PR makes sure that if subtype is no longer matching after upgrade, subtype will be changed instead of creating duplicate devices.

@gizmocuz:
- I can test with Philips Hue only, please help to review (it's the same code for Yeelight and ZWave)
- Possible improvements, please let me know if I should add either of them:
  - If duplicate devices are present, print a warning message to alert the user
  - If duplicate devices are present, remove the newest one(s) from database and alert the user

### DIFF
--- a/hardware/PhilipsHue/PhilipsHue.cpp
+++ b/hardware/PhilipsHue/PhilipsHue.cpp
@@ -440,15 +440,22 @@ void CPhilipsHue::InsertUpdateSwitch(const int NodeID, const _eHueLightType LTyp
 
 		//Get current nValue if exist
 		vector<vector<string> > result;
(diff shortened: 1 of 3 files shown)


gizmocuz:
Thanks!

PR #2293, opened 2018-04-05
Made other energy rates possible in report views

Olifant1990:
You can edit a General kWh device and select the belonging rate
(T1/T2/R1/R2). This is now also calculated in the report views. Needed
this because my Z-Wave P1 dongle created two general kWh devices and
they used both T1 as costs by default. Now you can choose.

### DIFF
--- a/main/RFXNames.h
+++ b/main/RFXNames.h
@@ -34,11 +34,13 @@ enum _eSwitchType
 enum _eMeterType
 {
 	MTYPE_ENERGY = 0,			//0
-	MTYPE_GAS,				//1
-	MTYPE_WATER,			//2
-	MTYPE_COUNTER,			//3
-	MTYPE_ENERGY_GENERATED,	//4
-	MTYPE_TIME,				//5
+	MTYPE_GAS,					//1
+	MTYPE_WATER,				//2
+	MTYPE_COUNTER,				//3
+	MTYPE_ENERGY_GENERATED,		//4
+	MTYPE_TIME,					//5
+	MTYPE_ENERGY_T2,			//6
+	MTYPE_ENERGY_GENERATED_R2,	//7
 	MTYPE_END
 };
 
--- a/www/app/DashboardController.js
+++ b/www/app/DashboardController.js
@@ -3722,7 +3722,7 @@ define(['app'], function (app) {
 										}
 									}
 									else if ((item.Type == "Energy") || (item.Type == "Power") || (item.SubType == "kWh")) {
-										if (((item.Type == "Energy") || (item.Type == "Power") || (item.SubType == "kWh")) && (item.SwitchTypeVal == 4)) {
+										if (((item.Type == "Energy") || (item.Type == "Power") || (item.SubType == "kWh")) && ((item.SwitchTypeVal == 4) || (item.SwitchTypeVal == 7))) {
 											imagehtml += 'PV48.png" class="lcursor" onclick="ShowCounterLogSpline(\'#dashcontent\',\'ShowFavorites\',' + item.idx + ',\'' + escape(item.Name) + '\', ' + item.SwitchTypeVal + ');" height="40" width="40"></td>\n';
 										}
 										else {
--- a/www/app/UtilityController.js
+++ b/www/app/UtilityController.js
@@ -620,7 +620,7 @@ define(['app'], function (app) {
 								status = "";
 							}
 							else if ((item.Type == "Energy") || (item.Type == "Current/Energy") || (item.Type == "Power") || (item.SubType == "kWh")) {
-								if (((item.Type == "Energy") || (item.SubType == "kWh")) && (item.SwitchTypeVal == 4)) {
+								if (((item.Type == "Energy") || (item.SubType == "kWh")) && ((item.SwitchTypeVal == 4) || (item.SwitchTypeVal == 7))) {
 									xhtm += 'PV48.png" height="48" width="48"></td>\n';
 								}
 								else {
--- a/www/html5.appcache
+++ b/www/html5.appcache
@@ -1,6 +1,6 @@
 CACHE MANIFEST
 
-# ref 1514
+# ref 1502
 
 CACHE:
 # CSS
--- a/www/html5.appcache
+++ b/www/html5.appcache
@@ -1,6 +1,6 @@
 CACHE MANIFEST
 
-# ref 1501
+# ref 1514
 
 CACHE:
 # CSS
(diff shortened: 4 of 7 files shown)


gizmocuz:
Unfortunately i do not have such unit, which brand is this ? (BeNext?)
What would be better is to patch the domoticz openzwave code so it will update only 1 sensor, the P1 sensor with the T1 and T2 , R1 and R2 values.
Then we can continue using the normal P1 sensor we already have, and there is no need for another meter sensor, and you get all in one sensor
Could you modify the code for this ?

Olifant1990:
It is indeed the BeNext P1 dongle.

- Do you have an idea where to start to make it one normal P1 counter/sensor? Where does OpenZwave decide to use which sensor based on what Z-Wave class(es)?
- Also the changes I made doesn't create a new sensor, just add two "switchtypes" for this general kWh sensor. The usage T2 and the return R2. It is still a handy upgrade for the allready existing sensor.

ronaldPeer:
Has it something to do with the config?

`<CommandClass id="50" name="COMMAND_CLASS_METER" version="4" request_flags="3" innif="true">
				<Instance index="1" />
				<Instance index="2" endpoint="1" />
				<Instance index="3" endpoint="2" />
				<Instance index="4" endpoint="3" />
				<Value type="decimal" genre="user" instance="1" index="0" label="Energy" units="kWh" read_only="true" write_only="false" verify_changes="false" poll_intensity="0" min="0" max="0" value="1742.452" />
				<Value type="decimal" genre="user" instance="1" index="8" label="Power" units="W" read_only="true" write_only="false" verify_changes="false" poll_intensity="0" min="0" max="0" value="164" />
				<Value type="bool" genre="user" instance="1" index="32" label="Exporting" units="" read_only="true" write_only="false" verify_changes="false" poll_intensity="0" min="0" max="0" value="False" />
				<Value type="decimal" genre="user" instance="2" index="0" label="Energy" units="kWh" read_only="true" write_only="false" verify_changes="false" poll_intensity="0" min="0" max="0" value="1763.696" />
				<Value type="decimal" genre="user" instance="2" index="8" label="Power" units="W" read_only="true" write_only="false" verify_changes="false" poll_intensity="0" min="0" max="0" value="164" />
				<Value type="bool" genre="user" instance="2" index="32" label="Exporting" units="" read_only="true" write_only="false" verify_changes="false" poll_intensity="0" min="0" max="0" value="False" />
				<Value type="decimal" genre="user" instance="3" index="0" label="Energy" units="kWh" read_only="true" write_only="false" verify_changes="false" poll_intensity="0" min="0" max="0" value="1795.990" />
				<Value type="decimal" genre="user" instance="3" index="8" label="Power" units="W" read_only="true" write_only="false" verify_changes="false" poll_intensity="0" min="0" max="0" value="0" />
				<Value type="bool" genre="user" instance="3" index="32" label="Exporting" units="" read_only="true" write_only="false" verify_changes="false" poll_intensity="0" min="0" max="0" value="False" />
				<Value type="decimal" genre="user" instance="4" index="0" label="Gas" units="cubic meters" read_only="true" write_only="false" verify_changes="false" poll_intensity="0" min="0" max="0" value="1225.078" />
				<Value type="bool" genre="user" instance="4" index="32" label="Exporting" units="" read_only="true" write_only="false" verify_changes="false" poll_intensity="0" min="0" max="0" value="False" />
			</CommandClass>`

Olifant1990:
Has it to do with the BeNext command_class_meter?:

```
			<CommandClass id="50" name="COMMAND_CLASS_METER" version="4" request_flags="3" innif="true">
				<Instance index="1" />
				<Instance index="2" endpoint="1" />
				<Instance index="3" endpoint="2" />
				<Instance index="4" endpoint="3" />
				<Value type="decimal" genre="user" instance="1" index="0" label="Energy" units="kWh" read_only="true" write_only="false" verify_changes="false" poll_intensity="0" min="0" max="0" value="1742.452" />
				<Value type="decimal" genre="user" instance="1" index="8" label="Power" units="W" read_only="true" write_only="false" verify_changes="false" poll_intensity="0" min="0" max="0" value="164" />
				<Value type="bool" genre="user" instance="1" index="32" label="Exporting" units="" read_only="true" write_only="false" verify_changes="false" poll_intensity="0" min="0" max="0" value="False" />
				<Value type="decimal" genre="user" instance="2" index="0" label="Energy" units="kWh" read_only="true" write_only="false" verify_changes="false" poll_intensity="0" min="0" max="0" value="1763.696" />
				<Value type="decimal" genre="user" instance="2" index="8" label="Power" units="W" read_only="true" write_only="false" verify_changes="false" poll_intensity="0" min="0" max="0" value="164" />
				<Value type="bool" genre="user" instance="2" index="32" label="Exporting" units="" read_only="true" write_only="false" verify_changes="false" poll_intensity="0" min="0" max="0" value="False" />
				<Value type="decimal" genre="user" instance="3" index="0" label="Energy" units="kWh" read_only="true" write_only="false" verify_changes="false" poll_intensity="0" min="0" max="0" value="1795.990" />
				<Value type="decimal" genre="user" instance="3" index="8" label="Power" units="W" read_only="true" write_only="false" verify_changes="false" poll_intensity="0" min="0" max="0" value="0" />
				<Value type="bool" genre="user" instance="3" index="32" label="Exporting" units="" read_only="true" write_only="false" verify_changes="false" poll_intensity="0" min="0" max="0" value="False" />
				<Value type="decimal" genre="user" instance="4" index="0" label="Gas" units="cubic meters" read_only="true" write_only="false" verify_changes="false" poll_intensity="0" min="0" max="0" value="1225.078" />
				<Value type="bool" genre="user" instance="4" index="32" label="Exporting" units="" read_only="true" write_only="false" verify_changes="false" poll_intensity="0" min="0" max="0" value="False" />
			</CommandClass>
```

I only see three energy meters (T1, T2 and R1 I guess).
Also the gas meter. But not all values... This is from the zwcfg_xxxxx.xml file, so it has been build by openzwave. The command_class_meter isn't present in the BeNext P1.xml default config file

gizmocuz:
Then it is best that this dongle is threaded like a P1 smart meter, so to say, when you receive an update, update a P1 smart meter instead ? This way you also only get 1 device (the p1 smart meter)

If node manufacturer / id and type equals BeNext then.... something like this

Olifant1990:
I will have a look at it soon.

But apart from it, I think it is still a nice addon to be able to select which cost type belongs to the kWh device, as this device can be used as dummy too. If you go to the report page of a kWh device, it will use T1 costs by default. This PR gives a choice to the user which cost slot to use.
Why don't you like this?

gizmocuz:
A 'normal' kwh sensor in domoticz can already be configured to be 'Usage' or as 'Return'
In this case, the P1 smart meter via a P1 zwave dongle, it is best to also thread it as a real P1 smart meter and use the P1 Smart Meter (electra/gas) sensor for this so it will be already OK

gizmocuz:
I received such a sensor for testing before, but i remember it was not 100% working at that time.
I think i will order one as well

gizmocuz:
Hmmm maybe not if i look at the price.. there are also two types.
For now i stick to my P1 serial port. Hope you can patch the code to use it as a P1 Electra/gas sensor

PR #2299, opened 2018-04-06
attempt to fix stability issues with HarmonyHub

gordonb3:


gordonb3:
still doesn't look right. Webserver goes into stall when I want to switch off an activity.

gizmocuz:
@gordonb3, please let me know when you would like this to be merged... Looks much better already by code! Great!

gordonb3:
Need to run some more checks. Currently still have the code for the first set of commits running to verify whether I solved the stalls while only monitoring.  Need to check the additional two commits after that, but I think sending the correct PowerDown command will do most of the trick required.

Timing values probably need some looking into as well. Most of the issues coming from this module (and possibly MultiFun and OpenWebNetTCP) seem to be related to try listening to a socket that is not sending anything.

I'm thinking of changing the switch type of this module as well. Since there can be only one activity that has state "on" at any moment it makes more sense to make this a multi-select switch. Not sure if I can load the text values to select from through the json.htm API though.

gizmocuz:
They should be push on switches actually

gordonb3:
Yeah... Already noticed that the values for multi-select switches need to be programmed in the frontend, so that won't work with dynamic setups.

And I FU'd my duration test because I forgot about the logrotate on Sunday morning. So that will a bit longer and I also noticed a nasty error logging that needs to be silenced - I have the whole AV system including the HarmonyHub sitting behing a kaku switch, causing this module to produce an error every minute telling me that my IP/Port configuration is wrong.

gordonb3:
May want to have a look at d0e27bb

A default socket connect call will wait for the peer to confirm an it will take a considerable amount of time to return if the peer is dead. Thus long in fact that mainworker will generate an error that the hardware thread "seems to have ended unexpectedly" - which is of course because we did not update our heartbeat time value while we were waiting for the socket connect operation to return.

The code for Linux works and limits the connect time to 5 seconds. I just don't know how to do it for Windows.

gordonb3:
All done, ready for merge

Windows users may still see the `thread seems to have ended unexpectedly` error due to how socket connect() operates, but I currently have no clue how to fix that.

gordonb3:
Hold on. I think I got the solution for Windows

gizmocuz:
Thanks a million!

PR #2300, opened 2018-04-06
Fix handling of TRC02 RGB lights

### LINKED ISSUE #2282: Selector/dimmer switches missing after 3.9168
Hi

After updating to beta 9168 several switches are missing from both the dashboard and the switches page

Found this thread on the forum describing other user with the same issue:
[http://www.domoticz.com/forum/viewtopic.php?f=6&t=22902](url)

App cache has been cleared
I even tried it on a fresh computer and get the same issue there

Strange thing it works better on a mobile client, there the devices are visible on the dashboard but not under the switches tab

Notice it is the same room in all below screenshots

PC Dashboard:
image

PC Switches:
image

IPAD Dashboard:
image

IPAD Switches tab:
image

emontnemery:
Closes #2282


gizmocuz:
Thanks!

PR #2301, opened 2018-04-06
Handle not unique headers as multiline in headers dict for plugin HTTP connection

guillaumezin:
This is usefull mainly for "Set-Cookie" headers, but also "WWW_Authenticate" and "Proxy_Authenticate", if we received many time the same headers with different values, it combines headers into one, with \n as separator.

If we want to respect RFC 2616, we would use comma as separator, but with practical considerations, and because we are not designing a web browser or server, \n is more universal

https://stackoverflow.com/questions/3096888/standard-for-adding-multiple-values-of-a-single-http-header-to-a-request-or-resp
https://github.com/bnoordhuis/mozilla-central/blob/master/netwerk/protocol/http/nsHttpHeaderArray.h#L185

This should address the issue I described here: http://www.domoticz.com/forum/viewtopic.php?f=65&t=22972

### DIFF
--- a/hardware/plugins/PluginProtocols.cpp
+++ b/hardware/plugins/PluginProtocols.cpp
@@ -245,9 +245,19 @@ namespace Plugins {
 				if (uHeaderText == "CHUNKED")
 					m_Chunked = true;
 			}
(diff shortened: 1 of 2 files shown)


gizmocuz:
Thanks!

PR #2302, opened 2018-04-08
Extract light editing to own page

stas-demydiuk:
- Update angularAMD to support components lazy defining
- Update styles to properly display disabled buttons
- Add live validation to light editing page
- Extract all light-related api calls to own factory
- Code cleanup + refactoring to angular style

image


gizmocuz:
Great patch! Thanks!!

Issue #2311, opened 2018-04-11
Device dependant issues (content not loaded) with Web UI


odelma:
https://www.domoticz.com/forum/viewtopic.php?f=6&t=23013
https://www.domoticz.com/forum/viewtopic.php?f=6&t=23024

There is pretty good description of the issue already on the above forum threads. I'll copy it partly here:

> Version: 3.9203
Platform: Raspberry Pi3B / Raspbian Stretch
Plugin/Hardware: Z-wave, RFLink, HarmonyHub etc.
Description: Domoticz web interface does not open with browsers (Safari, Chrome) on iPad2 / iOS 9.3.4 after update to latest beta.

Web interface works well on Chrome & Safari on my Macbook and also with iPhones running iOS 11. However, after updating to latest beta (from a beta published about three weeks ago) I no longer cannot open the web interface (well, I can open the page, it loads, but items are more or less missing on each tab, e.g. on the Dashboard only an empty drop-down menu for "add new device" is shown right below the navbar) on iPad2 running iOS 9.3.4. The iPad is being used as a wall-mounted control dashboard.

Needless to say, I've cleared all offline content as well as I can on the iPad, tried with incognito mode etc. but behaviour is consistent. I hooked up the iPad to the Macbook to enable the Web Inspector and can see several errors on the console. However, I'm pretty clueless on what exactly is causing them as nothing like this is shown on e.g. Chrome on the Macbook. I've also tried with multiple themes but seems there is no effect, errors are still consistent.
![image](https://user-images.githubusercontent.com/10372952/38645619-61f80d42-3ded-11e8-81cd-22eb0f285d41.png)
![image](https://user-images.githubusercontent.com/10372952/38645629-6ac73010-3ded-11e8-928e-297f7f2d1866.png)

Any pointers on how to get my old iPad2 back on track displaying Domoticz would be appreciated. And yes, could always go for a custom dashboard but I quite like the standard interface and it has been serving me well thus far. Wouldn't like to give up yet :)

@stas-demydiuk - I did not find you on the forum and therefore decided to open the issue here. I noted you've recently made some improvements (much appreciated! 😃) to the web UI. Could there be any relation to these issues experienced with the web UI? And as you seem to know your way around the UI stuff very well, maybe you could take a look even if there isn't any relation...

emontnemery:
Edit: @odelma, I should have read the description more carefully, sorry about that!

odelma:
[...]

PR #2304, opened 2018-04-09
Restore color picker value from API response on light edit page

stas-demydiuk:
- Fix small typo
- Add On command for rgbwPicker

image

PR #2313, opened 2018-04-12
Replace 'let' and 'const' with 'var' in javascript

emontnemery:
#2311


gizmocuz:
Thanks!

gizmocuz:
Please increase the appcache version number when making changes to the www folder
In this case please create a pull request for this

emontnemery:
Appcache version is bumped in what you just merged..
Edit: Aha, bumped to same evrsion also by #2304
Is additional bumping really needed given they were both merged at almost the same time?

PR #2316, opened 2018-04-13
Move scene timers editor and log to own pages

stas-demydiuk:
Move scene timers editor and log to own pages
- extract common timer components into own file to reuse between scene and device timers
- add timersun component
- remove redundant getSurRiseSet api call
- simplify timers related api wrappers
- group js and html files by feature folders
- fix small template issues


gizmocuz:
Great patch! Thanks!

sweetpants:
This patch causes an app cache issue only 103 of 650 files being refreshed. It stops at DeviceTimers.js Seems the file does not exists or an appcache problem https://github.com/domoticz/domoticz/issues/2318#event-1574446396

stas-demydiuk:
@sweetpants yes, sorry for that, fix is already in #2337

PR #2322, opened 2018-04-16
Notify when Philips Hue devices change level or color

### LINKED ISSUE #2281: Hue Level and Color change doesn't notify Domo / MQTT
If you change Level on the Hue App / Switch MQTT isn't notified and when you change Color with the Hue App neither Domoticz or MQTT is notified. 

In my case I've changed Color quite a few times with the Hue App in the past few minutes and the Color variable is stuck at "{"b":81,"cw":0,"g":255,"m":3,"r":29,"t":0,"ww":0}" (last color change I did with domo)

emontnemery:
Fixes #2281


gizmocuz:
Thanks!

PR #2331, opened 2018-04-18
Move _tColor definition to source file

emontnemery:


gizmocuz:
Maybe the best solution in this case was to create a source/header file for _tColor ? Color.cpp/Color.h ? or a better name

emontnemery:
@gizmocuz OK, now moved to ColorSwitch.cpp / ColorSwitch.h, is it better?

gizmocuz:
Thanks, excellent !

PR #2306, opened 2018-04-09
Tado Smart Thermostat device support

dotms:
Device support for the Tado Smart Thermostat. Data source is the _my.tado.com_ web application API. Some measures have been taken to make it as robust as possible but since the API is unofficial it may change in the future require code changes.

Provides the following sensors:

**Home**
- Presence (home/away)

**Zone**
- Setpoint
- Temperature/humidity
- Heating power %
- Heating enabled (can thermostat turn on heating if required)
- Setpoint override in effect
- Heating On (is thermostat currently heating)

Please hold this PR until I have tested it for about a week. Will follow up with confirmation.

Related feature request: https://github.com/domoticz/domoticz/issues/1294

PR #2334, opened 2018-04-19
Use boost regex instead of c++11 std::regex

emontnemery:
Fix compile error introduced by #2306 due to use of c++11 feature std::regex
Closes #2334.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -538,7 +538,7 @@ set(Boost_USE_STATIC_LIBS ${USE_STATIC_BOOST})
 set(Boost_USE_MULTITHREADED ON)
 unset(Boost_INCLUDE_DIR CACHE)
 unset(Boost_LIBRARY_DIRS CACHE)
-find_package(Boost REQUIRED COMPONENTS thread date_time system)
+find_package(Boost REQUIRED COMPONENTS thread date_time system regex)
 
 if(USE_STATIC_BOOST)
    message(STATUS "Linking against boost static libraries")
--- a/hardware/Tado.cpp
+++ b/hardware/Tado.cpp
@@ -22,8 +22,8 @@
 #include "../json/json.h"
 #include "../webserver/Base64.h"
 #include "Tado.h"
-#include <regex>
-#include <initializer_list>
+//#include <regex> //Requires c++11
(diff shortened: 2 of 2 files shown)


emontnemery:
@dotms Please review

dotms (approved):
Looks fine to me, thanks for this fix!

@gizmocuz Is there any particular reason why c++11 extensions are not enabled/specified when building on Linux using gcc 4.x, while on more recent gcc versions this feature is enabled by default?

emontnemery:
@dotms it was discussed in #2081. My take is that c++11 unfortunately has to remain off until we can assume most users are on stretch or newer, so maybe next year, or until someone makes domoticz fully statically linked.

gordonb3:
Yeah... The problem is that in between gcc versions 4.7 and 5.1 the symbol naming of c++11 objects is not properly defined. As a result any revison or upgrade that has either its source or destination within this range will cause applications using it to break.

I'm unsure where Microsoft Visual Studio stands in this.

gizmocuz:
Could you try to avoid boost as well ? (or the regex, try a different method ?)

dotms:
Do you object against the use of regular expressions (whether it comes from std or boost)? Or boost altogether? I'm unsure what you mean.

gizmocuz:
I opt to limit boost usage as much as possible as we need to add this to the build systems
Sure when we could finally use c++x11, that would solve a lot of problems.
As this is not possible, you can therefore also not use the c++x11 regex function (unfortunately)

dotms:
I can work around it since I only need a few specific bits of information, although it will probably look a bit silly.

emontnemery:
@dotms OK, then I close this PR?

gizmocuz:
@dotms , or maybe update this pull request with other code ?

dotms:
@gizmocuz By the looks of it I cannot change much on this PR (makes sense since it's emontnemery's PR) nor the original one (most likely since it's already merged). 

I have a few changes ready to publish. If it's ok with you and @emontnemery close this PR and I will create a new one with the updated code. Code is ready, just test compiling it on gcc 4.8 now.

gizmocuz:
That would be great! Thanks in advance!

PR #2335, opened 2018-04-20
Address some issues with Harmony Hub

gordonb3:
Responding to forum topic https://www.domoticz.com/forum/viewtopic.php?f=6&t=23106

- do a quicker restore of connection after it was dropped
- compensate for possible heartbeat miss when fetching list of activities
- correct `PowerOff` switch state
- keep list of activities between session resets
- don't treat response that more data is coming as error
- increased timeout values to compensate for possible bad wifi reception with Hub hidden in cabinet

### DIFF
--- a/hardware/HarmonyHub.cpp
+++ b/hardware/HarmonyHub.cpp
@@ -648,16 +648,33 @@ bool CHarmonyHub::SendPing()
 	if (strData.empty())
 		return false;
 
-	if (strData.find("</message>") == std::string::npos) // no messages included
-		return (strData.find("errorcode='200'") != std::string::npos);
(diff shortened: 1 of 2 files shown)


gordonb3:
Hold back on this one. I accidentally reinstated the repeated error when the Hub is down.

gordonb3:
Never mind...

Wrong interpretation of the shown error. This particular error is actually one that we do like to show (repeated) - the timing just doesn't work out too well with my logrotate scheme. Which is more my problem than it should be someone else's.

Code is good to go.

gizmocuz:
Thanks!!

PR #2336, opened 2018-04-22
Fix device deletion

stas-demydiuk:
- Remove redundant question during removing (as it doesn't make any effect)
- Add disable button to device edit page
- Remove limit to max count of device levels:
  we doesn't have real restriction for that and some plugins already use more than 10 levels

image

### DIFF
--- a/www/app/DeviceLightEdit.js
+++ b/www/app/DeviceLightEdit.js
@@ -294,13 +294,12 @@ define(['app'], function (app) {
             }
 
             function isAddLevelAvailable() {
-                var isLessThanLimit = vm.ngModelCtrl.$modelValue.length < 11;
                 var isNotEmpty = !!vm.newLevelName;
                 var isUnique = !vm.ngModelCtrl.$modelValue.find(function (item) {
                     return item.name === vm.newLevelName
                 });
 
-                return isLessThanLimit && isNotEmpty && isUnique;
+                return isNotEmpty && isUnique;
             }
 
             function addLevel() {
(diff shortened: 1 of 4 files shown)


stas-demydiuk:
@gizmocuz what do you think about changing "Used/Not Used" naming to "Enabled/Disabled"?
From my point of view it sounds more clear what will actually happen with device.

gizmocuz:
@stas-demydiuk, your talking about the two buttons in the devices tab ? I think used/not used is better there, you can not disable a device

gizmocuz:
Please avoid making changes in the current messages as when you do this, this means every translator needs to translate this string again.
Is it really necessary ?

There should be no 'disable' option for a light/switch device, why ? You can delete it is you do not need it, or hide it

stas-demydiuk:
@gizmocuz, I've changed the naming because initial message is quite unclear for me. Especially if you look at json command, it is `type=setused&used=false` but the button completely removes device from domoticz.

BTW, do we really need this button on "Device Edit" page, maybe leave only "Set Unused" button is enough and user can only delete device on devices page?

gizmocuz:
The user should perform these actions from the devices tab. When doing so, we do not not need the button in the edit light device
The JSON command is not for end-users, setting used to false is not deleting the device

stas-demydiuk:
But looks like it deletes, unless I missed something (as I don't have any C experience)
void CWebServer::RType_SetUsed
```
if ((used == 0) && (maindeviceidx.empty()))
{
    //really remove it, including log etc
    m_sql.DeleteDevices(idx);
}
```
because maindeviceidx is always empty as we've never sent it when click "Delete" device

gizmocuz:
Yes, that seems like a delete indeed
But again, this should not be done in edit light/switch, there was no disable option here, should not be here
And when you press 'Remove', then of course you know what it does, that message box did not need any more explanation (translation)

gizmocuz:
Thanks!

PR #2337, opened 2018-04-22
Tado fix for build issues (regex/boost)

dotms:
As described in #2334 a PR to prevent future issues such as #2306.

Changes:
* Removed references to boost/regex
* Rewrote function MatchValueFromJSKey to use string manipulation functions from Helper instead of std/boost regex library (to allow for compatibility).
* Optimized logging (only write to trace log if tracing is actully enabled)
* Reference std namespace by default to allow for better readability.


dotms:
Noticed a warning on Tado.cpp on the macOS build. Will do an additional commit to fix this even though it's minor thing.

hardware/Tado.cpp
```diff
@@ -22,26 +22,28 @@
 #include "../json/json.h"
 #include "../webserver/Base64.h"
 #include "Tado.h"
-#include <regex>
-#include <initializer_list>
 
-#define round(a) ( int ) ( a + .5 )
+using namespace std;
```
gizmocuz: Please do not use the 'using namespace std', but write out the complete name like std::string, also in the header

hardware/Tado.cpp
```diff
@@ -22,26 +22,28 @@
...
+
 const int TADO_POLL_INTERVAL = 30;  // The plugin should collect information from the API every n seconds.
-const std::string TADO_API_ENVIRONMENT_URL = "https://my.tado.com/webapp/env.js";
+const string TADO_API_ENVIRONMENT_URL = "https://my.tado.com/webapp/env.js";
```
gizmocuz: instead of const int/string in the .cpp, could you use

`#define TADO_POLL_INTERVAL 30`
or

`#define TADO_API_ENVIRONMENT_URL "https://my.tado.com/webapp/env.js"`

hardware/Tado.cpp
```diff
@@ -22,26 +22,28 @@
...
+CTado::CTado(const int ID, const string &username, const string &password)
 {
 	m_HwdID = ID;
 	m_TadoUsername = username;
 	m_TadoPassword = password;
 
-	_log.Log(LOG_TRACE, "Tado: Started Tado plugin with ID=" + boost::to_string(m_HwdID) + ", username=" + m_TadoUsername);
+	if (_log.isTraceEnabled())
+	{
+		_log.Log(LOG_TRACE, "Tado: Started Tado plugin with ID=" + boost::to_string(m_HwdID) + ", username=" + m_TadoUsername);
```
gizmocuz: Please use the printf functionality of the log like

`_log.Log(LOG_TRACE, "Tado: Started Tado plugin with ID=%d, username=%s",m_HwdID, m_TadoUsername.c_str());`

Please avoid using boost functions like `boost::to_string`

(also further in the code)

hardware/Tado.h
```diff
@@ -58,29 +58,30 @@ class CTado : public CDomoticzHardwareBase
...
+		string m_TadoRefreshToken;
 
 		bool m_bDoLogin; // Should we try to login?
 		bool m_bDoGetHomes;
 		bool m_bDoGetZones;
 		bool m_bDoGetEnvironment;
 
-		std::vector<_tTadoHome> m_TadoHomes;
+		vector<_tTadoHome> m_TadoHomes;
 };
```
gizmocuz: Please add a return here

gizmocuz (changes requested):
Thank you for the changes.

Please make the requested changes, it might be better to create a new pull request because of the namespace usage

hardware/Tado.cpp
dotms: Reverted change I made. Both header and cpp file now use explicit std namespace reference.

hardware/Tado.cpp
dotms: Changed constants into #define statements.

hardware/Tado.cpp
dotms: Changed all the _log.Log lines to use printf format. 
Replaced string concatenations in the file with this format as well. Thanks for the suggestion!

hardware/Tado.h
dotms: Return added.

gizmocuz:
Thank you! Much appreciated !

Issue #2279, opened 2018-04-04
FloorPlans issue


kovainfo:
When i press on temp sensor or baro sensor - screen is cleared and nothing more.

quack3d:
Same thing reported here: https://www.domoticz.com/forum/viewtopic.php?f=6&t=22901

szczukot:
I have this same problem

dnpwwo:
Seems that someone is working their way through Domoticz changing how the Log page works for the HTML tabs but not the Floorplans.

Pushed a fix for Temperature sensors but only just saw the post for switches

kovainfo:
@dnpwwo Cool! But BARO sensor still have this issue. including dummy baro sensor...

pj-r:
Found one more: Custom sensors.

This one is compressor hz from MySensors heat pump node but this can be reproduced with dummy custom sensor.

`VM186 Floorplans:1 Uncaught ReferenceError: ShowGeneralLog is not defined
    at SVGImageElement.onclick (VM158 Floorplans:1)`

`      {
         "AddjMulti" : 1.0,
         "AddjMulti2" : 1.0,
         "AddjValue" : 0.0,
         "AddjValue2" : 0.0,
         "BatteryLevel" : 255,
         "CustomImage" : 15,
         "Data" : "0 Hz",
         "Description" : "",
         "Favorite" : 1,
         "HardwareID" : 4,
         "HardwareName" : "Mys-Wireless-LAN",
         "HardwareType" : "MySensors Gateway with LAN interface",
         "HardwareTypeVal" : 42,
         "HaveTimeout" : false,
         "ID" : "00000105",
         "Image" : "Heating",
         "LastUpdate" : "2018-04-16 13:51:15",
         "Name" : "PumppuCompressor",
         "Notifications" : "false",
         "PlanID" : "13",
         "PlanIDs" : [ 13 ],
         "Protected" : false,
         "SensorType" : 1,
         "SensorUnit" : "Hz",
         "ShowNotifications" : true,
         "SignalLevel" : "-",
         "SubType" : "Custom Sensor",
         "Timers" : "false",
         "Type" : "General",
         "TypeImg" : "Heating",
         "Unit" : 1,
         "Used" : 1,
         "XOffset" : "152",
         "YOffset" : "434",
         "idx" : "142"
      },`

kovainfo:
!Can't find variable: ShowGeneralLog
{
    function onclick(evt) {
[...]

PR #2342, opened 2018-04-23
#2279 Fix url for device log page

stas-demydiuk:
- Extract domoticzblocks.js, domoticzdevices.js from gzip
- Fix url for device log page from floor plan


gizmocuz:
Was there a change needed in the domoticzblocks ? (blockly part) ?

stas-demydiuk:
No, just unzipped both files because they contain domoticz related code as it would be useful to start having version history for them and I hope help to avoid such issues.

gizmocuz:
I understand, but then only for changed files please. If both are not changed, please undo this change

stas-demydiuk:
@gizmocuz I hope got your idea right: removed domoticzblocks.js.gz from pull-request.
But still don't completely understand what is the benefit of having this file compressed in the repository (except of RPI performance issue I guess)  as it doesn't allow us to have its history in git log

gizmocuz:
Its been explained quite some times, but basically in short, it saves a lot of CPU cycles on slow systems
Later we will make the compression realtime (for changes files) or during packaging .

gizmocuz:
Its been explained quite some times, but basically in short, it saves a lot of CPU cycles on slow systems
Later we will make the compression realtime (for changes files) or during packaging .

gizmocuz:
Thanks!

PR #2343, opened 2018-04-23
fix authentication issue with clients directly accessing json.htm

gordonb3:
Clients gaining admin rights through a direct call to json.htm may never receive a session cookie if another client or script is using the same IP address


gordonb3:
Hold back on this one

I think this may introduce a new conflict with scripts using basic auth

gizmocuz:
Thanks!

PR #2345, opened 2018-04-23
Harmony Hub: use receive size from socket operation to copy the correct number of bytes to our resultstring

gordonb3:
using the smaller data buffer for socket operations there is a chance that it may be filled all the way to the last character and we cannot use its content as a string.

### DIFF
--- a/hardware/HarmonyHub.cpp
+++ b/hardware/HarmonyHub.cpp
@@ -64,6 +64,11 @@ History:
 #define DATABUFFER_SIZE  READBUFFER_SIZE + 1
 
 
+// socket timeout values
+#define TIMEOUT_WAIT_FOR_ANSWER 2.0f
+#define TIMEOUT_WAIT_FOR_NEXT_FRAME 0.4f
+
+
 
 CHarmonyHub::CHarmonyHub(const int ID, const std::string &IPAddress, const unsigned int port):
 m_harmonyAddress(IPAddress)
@@ -252,7 +257,7 @@ void CHarmonyHub::Do_Work()
 				if (m_commandcsocket->read(databuffer, READBUFFER_SIZE, false) > 0)
 				{
 					strData.append(databuffer);
-					m_commandcsocket->canRead(&bIsDataReadable, 0.4f);
+					m_commandcsocket->canRead(&bIsDataReadable, TIMEOUT_WAIT_FOR_NEXT_FRAME);
 				}
 				else
 					bIsDataReadable = false;
@@ -503,7 +508,7 @@ bool CHarmonyHub::StartCommunication(csocket* communicationcsocket, const std::s
 	communicationcsocket->write(strReq.c_str(), static_cast<unsigned int>(strReq.length()));
 	std::string strData;
 	bool bIsDataReadable = true;
-	communicationcsocket->canRead(&bIsDataReadable, 2.0f);
+	communicationcsocket->canRead(&bIsDataReadable, TIMEOUT_WAIT_FOR_ANSWER);
 	if (bIsDataReadable)
 	{
 		memset(databuffer, 0, DATABUFFER_SIZE);
@@ -527,7 +532,7 @@ bool CHarmonyHub::StartCommunication(csocket* communicationcsocket, const std::s
 	strAuth.append("</auth>");
 	communicationcsocket->write(strAuth.c_str(), static_cast<unsigned int>(strAuth.length()));
 
-	communicationcsocket->canRead(&bIsDataReadable, 2.0f);
+	communicationcsocket->canRead(&bIsDataReadable, TIMEOUT_WAIT_FOR_ANSWER);
 	if (bIsDataReadable)
 	{
 		memset(databuffer, 0, DATABUFFER_SIZE);
@@ -545,7 +550,7 @@ bool CHarmonyHub::StartCommunication(csocket* communicationcsocket, const std::s
 	//strReq = "<stream:stream to='connect.logitech.com' xmlns:stream='http://etherx.jabber.org/streams' xmlns='jabber:client' xml:lang='en' version='1.0'>";
 	communicationcsocket->write(strReq.c_str(), static_cast<unsigned int>(strReq.length()));
 
-	communicationcsocket->canRead(&bIsDataReadable, 2.0f);
+	communicationcsocket->canRead(&bIsDataReadable, TIMEOUT_WAIT_FOR_ANSWER);
 	if (bIsDataReadable)
 	{
 		memset(databuffer, 0, DATABUFFER_SIZE);
@@ -583,7 +588,7 @@ bool CHarmonyHub::GetAuthorizationToken(csocket* authorizationcsocket)
 	authorizationcsocket->write(strReq.c_str(), static_cast<unsigned int>(strReq.length()));
 
 	bool bIsDataReadable = true;
-	authorizationcsocket->canRead(&bIsDataReadable, 2.0f);
+	authorizationcsocket->canRead(&bIsDataReadable, TIMEOUT_WAIT_FOR_ANSWER);
 	if (bIsDataReadable)
 	{
 		memset(databuffer, 0, DATABUFFER_SIZE);
@@ -598,14 +603,14 @@ bool CHarmonyHub::GetAuthorizationToken(csocket* authorizationcsocket)
 		return false;
 	}
 
-	authorizationcsocket->canRead(&bIsDataReadable, 2.0f);
+	authorizationcsocket->canRead(&bIsDataReadable, TIMEOUT_WAIT_FOR_ANSWER);
 	while(bIsDataReadable)
 	{
 		memset(databuffer, 0, DATABUFFER_SIZE);
 		if (authorizationcsocket->read(databuffer, READBUFFER_SIZE, false) > 0)
 		{
 			strData.append(databuffer);
-			authorizationcsocket->canRead(&bIsDataReadable, 0.4f);
+			authorizationcsocket->canRead(&bIsDataReadable, TIMEOUT_WAIT_FOR_NEXT_FRAME);
 		}
 		else
 			bIsDataReadable = false;
@@ -652,14 +657,14 @@ bool CHarmonyHub::SendPing()
 	m_commandcsocket->write(strReq.c_str(), static_cast<unsigned int>(strReq.length()));
 
 	bool bIsDataReadable = true;
-	m_commandcsocket->canRead(&bIsDataReadable, 2.0f);
+	m_commandcsocket->canRead(&bIsDataReadable, TIMEOUT_WAIT_FOR_ANSWER);
 	while (bIsDataReadable)
 	{
 		memset(databuffer, 0, DATABUFFER_SIZE);
 		if (m_commandcsocket->read(databuffer, READBUFFER_SIZE, false) > 0)
 		{
 			strData.append(databuffer);
-			m_commandcsocket->canRead(&bIsDataReadable, 0.4f);
+			m_commandcsocket->canRead(&bIsDataReadable, TIMEOUT_WAIT_FOR_NEXT_FRAME);
 		}
 		else
 			bIsDataReadable = false;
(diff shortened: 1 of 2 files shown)


gizmocuz:
I dont think that this was an issue, but it is better to change the 'read' part of the code to read with a size -1 if you think it is a potential problem

gordonb3:
The thing is I did an export of all the package sizes I get myself on this and I found that out of 27 fragments two of them were 1500 bytes long. This does not change if I double the data buffer size. The maximum stays 1500, but the normal size 1448 bytes.

The trouble here is that databuffer is used as a string input to a string.append() function, so it has to have a zero terminator and if I set the usable size of the buffer to 1499 than it will fragment those two bigger 1500 byte size frames into a 1499 byte frame and a 1 byte frame (likely padded to meet a minimum size), because that's how Ethernet works.

The other option is to verify the size of the received data and if that is equal to the buffer size of 1500 use the substring feature of string.append(). Adding the 1 byte is most definitely the cheaper solution though.

gordonb3:
... Bare in mind that I decreased the original buffer size from 2MB to the real life situation of 1500 bytes. So yes, this would have never been an issue before because the limit could never be reached and as a result there would always be a trailing \0 to allow the buffer to be referenced as a string type. I just happened to get too close to the fire.

gizmocuz:
True, but then i would increase it, and still in the read functions use -1, this sounds more logical right ?

gordonb3:
Depends on your point of view I guess. The way I see it using the data buffer as a string instead of a char[n] can be considered unexpected and it may be confusing for someone investigating the code in the future and thinking databuffer size should conform to some other networking standard if I set the define for the buffer size to 1501 (or worse in this sense: to something - hexadecimal - sane like 1536). Either way the spare byte needs to be preserved, so there's no avoiding of either a +1 or a -1 to the define.

gordonb3:
... The objective of course being to avoid needless fragmenting and related additional network traffic.

gordonb3:
How about this?

gizmocuz:
Thanks! Looking perfect, happy to merge this !
I think when the original code was committed i asked if it was possible to use the AsyncTCP class we already have, but as the code was derived from code on github, this was not possible for the original author.

The -1 could also be handled only in 'csocket::read' 

int numBytesRemaining = numBytesToRead - 1;

This might make the harmony code more readable ? If not ... happy to merge!

gordonb3:
Hadn't thought about that. Would still require me to set the data buffer to the odd 1501 size in the HarmonyHub class though. Also, I found that there are two other classes using the csocket class as well and I don't really know how these classes reference it.

The main thing is that we need the 1500 receive size. Which is something that I did not expect because I only saw the maximum of 1448 at first. There may be a point where we need to decide to increase the buffer size to 2324 if there are people running the Hub in a wireless only network (see note inside the code) but then again we will need the additional `stop` byte to allow us to safely use the data buffer as a string like we do in line [242](https://github.com/domoticz/domoticz/pull/2345/files#diff-de3388f081bf82e8086d8ae50957d366L242).

I can have a look at the AsyncTCP class. Also had not realized there might be other code doing more or less the same thing. Maybe combine the two if they complement each other - mostly thinking about the connect timeout that we got sorted now.

gordonb3:
Hmmm... Seems like a better idea to drop AsyncTCP in favour of csocket as it appears to have the same issues that I just eliminated in csocket. Most specific a large timeout on trying to connect to a dead peer. Probably a lot harder to tackle with all the boost code in there as well.

gizmocuz:
I would set the buffersize to 2500... then we are save, and those extra 1000 bytes... nobody will notice (when they use the hub hardware ;)

Another option is to add a readString function (that returns a string (in the paramaters std::string &), then you do not need a buffersize at all in the harmony class ?

Question, why you expect a timeout ? If everything is async, you get data when the hub sends data...
Only for connection ? We could set this to 5 seconds (with a boost option)

But it might be easier to implement the readString function ?

gordonb3:
I'd know :smile: 
And someone else might see it as well and make the same mistake at some time again, skipping over the `use as string` thingy. Funny you should mention a readString() function as that appears to be exactly what is meant to be done here, but it seems to me that faces us with the same challenge.

As an alternative to the current code:
```	commandcsocket->canRead(&bIsDataReadable, TIMEOUT_WAIT_FOR_ANSWER);
	while (bIsDataReadable)
	{
		memset(databuffer, 0, DATABUFFER_SIZE);
		if (commandcsocket->read(databuffer, READBUFFER_SIZE, false) > 0)
		{
			strData.append(databuffer);
			commandcsocket->canRead(&bIsDataReadable, TIMEOUT_WAIT_FOR_NEXT_FRAME);
		}
		else
			bIsDataReadable = false;
	}
```

I could also rewrite it like this:
```		commandcsocket->canRead(&bIsDataReadable, TIMEOUT_WAIT_FOR_ANSWER);
	while (bIsDataReadable)
	{
		int bytesReceived = commandcsocket->read(databuffer, READBUFFER_SIZE, false);
		if (bytesReceived > 0)
		{
			strData.append(databuffer, 0, bytesReceived);
			commandcsocket->canRead(&bIsDataReadable, TIMEOUT_WAIT_FOR_NEXT_FRAME);
		}
		else
			bIsDataReadable = false;
	}
```

Unsure which one would be more CPU friendly. Note that in the second case I do no longer need to do a memset so I will be saving some cycles there.

gordonb3:
About that AsyncTCP: I think we'll need to rewrite the whole class if we would want to make that work. This is because the packages I see coming in do not appear to be regular TCP fragments of one bigger size message (which could be up to 64kB in size). Instead it appears that the Hub itself is breaking up the message and sending the fragments out as individual packages that need to be reconstructed on the client side. I have a feeling clearing all the possible conflicts here will prove to be very challenging.

gizmocuz:
The second one is the best... but requires more code ;)
Maybe adding a readString function that does call the original 'read' (-1 for the length), that returns the string.
Not so worried about the cpu...

the hub code could then look like

`			std::string strData=m_commandcsocket->read(databuffer, BUFFER_SIZE, false, TIMEOUT_WAIT_FOR_NEXT_FRAME)
`

gordonb3:
You're forgetting that a string needs to be zero terminated, hence a 1500 byte string will occupy 1501 bytes in memory. We can't use the string definition though, because the underlying functions defined by sys/socket and winsock2 want the char[] parameter.

Since I'm not running Domoticz on a blade server I do like to consider memory and CPU usage. Thus, I'll do the second. I did some investigating about cost of string concatenation and found some benchmarks stating that string.append() is the quickest. While none of the benchmarks did anything with the substring parameters it was claimed that the required number of cycles for this operation came almost exclusively from moving characters from source to target. I'd say there has to be some overhead but I should still be looking at a net profit from eliminating the memsets. Considering we started off with 2MB buffers we already accomplished a huge gain here.

hardware/HarmonyHub.cpp
```diff
@@ -724,7 +717,7 @@ bool CHarmonyHub::SubmitCommand(const std::string &strCommand, const std::string
 		lstrCommand = GET_CURRENT_ACTIVITY_COMMAND;
 	}
 
-	std::string strData;
+	std::string strData = ""; // make sure to initialize this as we will be appending to it
```
gizmocuz: I think the constructor of this class will already do this

hardware/HarmonyHub.cpp
```diff
@@ -250,13 +247,13 @@ void CHarmonyHub::Do_Work()
 
 			boost::lock_guard<boost::mutex> lock(m_mutex);
 			std::string strData;
-			char databuffer[DATABUFFER_SIZE];
+			char databuffer[SOCKET_BUFFER_SIZE];
 			while (bIsDataReadable)
 			{
-				memset(databuffer, 0, DATABUFFER_SIZE);
-				if (m_commandcsocket->read(databuffer, READBUFFER_SIZE, false) > 0)
+				int bytesReceived = m_commandcsocket->read(databuffer, SOCKET_BUFFER_SIZE, false);
```
gizmocuz: Why not make a easy function in the socket class to return a complete string ?

gizmocuz (commented):
Would it not be better to create a function in the socket class to return a string ?
As this is used almost all over the hue class, also the code will be a lot cleaner and removes a lot of duplicate code.

gordonb3:
Because the type of communication this addresses is specific to Harmony Hub and csocket is a generic wrapper around the system's socket functions that is used by other hardware classes as well.

But granted, I could combine the socket reads in a function. Should note that they're not all the same: out of 9 there are 5 that look for a multi-part message and the other 4 do and should not as this will make things more complicated.

hardware/HarmonyHub.cpp
gordonb3: Likely, but a compiler might still warn about it.

gizmocuz:
The csocket class was included with the merge of the hub class. MultiFun and openwebnet should use AsyncTCP as also all other classes handling tcp should do.
I am also NOT a fan of the csocket class.
Adding a simple readstring function should not cause any harm.
The timeout's could also be added as a variable to the class, and could be set via a function once

hardware/HarmonyHub.cpp
gizmocuz: Should not... it perfectly legal and normal, nobody would declare (i hope) things like

std::string myString("");

gordonb3:
I understand, but that's not the point. The thing is that the Hub sends partials as individual messages. This is most likely the reason that the original code already did not use the functionality of csocket::read() to retrieve additional frames (which is the `false` parameter in the call).

Following your line of thinking it will actually make more sense to include the code from csocket in the HarmonyHub class (similar to what I did with the EvohomeWeb class) rather than extend the functionality of a class that you rather get rid of.

gizmocuz:
True, but one good class (ASyncTCP) i would prefer.

Another approach i see here

https://github.com/NovaGL/HarmonyHubControl

gordonb3:
That's not another approach. It is the very code where this class is based on and it is disfunctional with newer Hub firmware. See my fork for an up to date version: https://github.com/gordonb3/HarmonyHubControl

gordonb3:
Okay, just had another look at the AsyncTCP class. Didn't quite get it at first, but it is effectively a skeleton class that needs to be extended.

Problems:
- I can't see how it will react when trying to connect to a dead peer, which is a big issue
- it appears to generate needless overhead due to using a 1kB buffer rather than using the Ethernet frame size (MTU/MRU)
- it is unclear what polling interval async_read_some() uses for checking and accepting data from the peer and that may cause issues with this API
- the Hub probably will respond badly if we start sending data to it (including the required ping) while it is still sending data to use, which we will need to be able to detect through the asynchronous communication.
- it doesn't eliminate the need for what this PR does in reconstructing the full message, which is about not allowing the string operation to read past the buffer[] size, thereby avoiding inclusion of unreadable characters in the output string.

Can we please focus on the issue at hand and maybe talk about reinventing this class from top to bottom later?

gizmocuz:
Sure you should be able to send a command while other things are read.
But anyway, my initial issue with this PR could have been easily resolved by only changing the read function to read 1 byte less then the buffer size.
Next as this call is used a lot and in a while loop, it seems more then logical to just add a function to the csocket class that returns a complete string, and call this instead of all the extra code...
This also eliminates the buffer in the hub class as this can be in the socket class.
Please do not talk about 1KB more memory... that is nothing. If you are still using a system with only 32KB of memory, then maybe it is time for a little upgrade ;)

gordonb3:
I'm sorry, but you keep wanting me to add a load of needless CPU cycles where at the same time you keep pushing for making the code a few bytes smaller by combining similar lines in a function. That doesn't make sense.

If I decrease the read buffer length to 1499 than this will result in me reading 1499 bytes, 1 byte, 1499 bytes, 1 byte, 1499 bytes, 1 byte...etc in sequence if the peer sends me anything larger than 4500 bytes - which will be fractured into 1500 size frames by the network layer.

This is effectively even more of a problem with the AsyncTCP class using the 1024 buffer size. While the Hub appears to only send larger sized packages occasionally, for reasons unknown, the regular communication size is 1448, resulting in consecutive reads of 1024 bytes (possible overflow when used as input in `string.append()` ) and 424 bytes and thus I would be doing twice the number of reads if I would be tempted to try rewrite the class to use it. Which I am not, anymore.

I'm NOT doing the buffer size minus 1, period.

gizmocuz:
Okey okey ;) it's perfect already because of the new receive method you added and the append with the actual size returned.

gizmocuz:
Thanks!!!

gordonb3:
Right... Need to say I'm sorry I snapped at you. Shouldn't have done that but I really don't like to do things I know to be inherently wrong. Will you change the buffer size in the AsyncTCP class accordingly?

gizmocuz:
Ohh well that's not an issue... I receive packets of more data with this class but the receive is called more
But please feel free to make the needed adjustments here
You can test this by writing a small winsock test application perhaps

gordonb3:
Yeah... 've been reading into the asio library. Was thinking of maybe writing some stand-alone app for testing. Somewhat annoyed though to find that the non-boost version still requires boost for binding the handler function and I would like to avoid that dependency. Did notice that the AsyncTCP class is using the `read_some` method instead of normal `read` and I'm unsure how that works internally and whether that is in fact still related to the actual socket operation which as we know has a maximum payload size of 1500, or that we are referencing some in memory queue of frames already received by the network layer? It makes sense to assume that such a queue will keep the original Ethernet frames intact, but it might also store the reconstructed TCP frame that can be up to 65500 bytes in size.

In any case Domoticz is way too big to play with this and find out how different settings affect performance.

PR #2346, opened 2018-04-23
Send color as JSON object in MQTT senddeviceinfo instead of as JSON string dump

emontnemery:
Fixes issue reported in the forums: https://www.domoticz.com/forum/viewtopic.php?f=6&t=23115&p=178329#p178329


gizmocuz:
Thanks for the changes !

Could you change some code please

- In the header of ColorSwitch.h, please change

`#include "../json/json.h"`

to

```
namespace Json
{
	class Value;
};
```

In 'ColorSwitch.cpp' add under the last include a new one:
`#include "../json/json.h"`

This saves a lot of compile time when you make changes to the header files

- Visual Studio, The header file you placed under the 'Devices' folder/layout, that is ok, but the .cpp file you placed under the 'Sources' folder. Please move this as well under 'Devices', or move the header file to the 'Header' folder

I also get a compile error under visual studio:

```
1>PluginManager.obj : error LNK2005: "public: __thiscall _tColor::_tColor(void)" (??0_tColor@@QAE@XZ) already defined in ColorSwitch.obj
1>Plugins.obj : error LNK2005: "public: __thiscall _tColor::_tColor(void)" (??0_tColor@@QAE@XZ) already defined in ColorSwitch.obj
1>Plugins.obj : error LNK2005: "public: class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > __thiscall _tColor::toJSON(void)const " (?toJSON@_tColor@@QBE?AV?$basic_string@DU?$char_traits@D@std@@V?$allocator@D@2@@std@@XZ) already defined in ColorSwitch.obj
```

This is both with and without the json change

emontnemery:
@gizmocuz OK, updated as you suggest:
- Changed to forward declaration of Json::Value
- ColorSwitch.cpp / ColorSwitch.h are now both directly under "Devices" in Visual Studio's solution explorer

However, I don't see your build errors, it builds just fine on Appveyor.
I think the errors you see could be caused by old version of hardwaretypes.h, which included both declaration and definition of _tColor struct, maybe you have a stale version of precompiled headers?

gizmocuz:
@emontnemery , you could be right about an old hardwaretypes.h
Thanks for the changes!

PR #2347, opened 2018-04-24
Better management of TIC pulse 433MHz via RFPlayer

ultrasuperpingu:
I was sending the apparent power in methods SendKwhMeter and not the active power. This was confusing. So now, I estimate active power with the last values received. I send apparent power via a SendWattMeter call.


gizmocuz:
Thanks!

Issue #2332, opened 2018-04-18
Json command setcolbrighnessvalue not working properly


Janco59:
The json command .../json.htm?type=command&param=**setcolbrightnessvalue**&idx=418&hue=0&**brightness=0**&iswhite=false doesn't reduce brightness to zero but to an unknown low value. If a high(er) value is selected, it works properly.

Irrespective off the value, the command is always confirmed with:
{
   "status" : "OK",
   "title" : "SetColBrightnessValue"
}

This applies to Domoticz v3.9268 and maybe some earlier builds. Before it used to work as expected.

emontnemery:
@Janco59 This is same as https://www.domoticz.com/forum/viewtopic.php?f=6&t=23081&p=178130#p178130, and you see it on Philips Hue lights?

Janco59:
Yes, one of the people commenting on that thread suggested to open a thread here.
It's indeed on Philips Hue (old model Hub with Hue Bloom light, haven't tested it on other hardware).

emontnemery:
@Janco59 : I've made a potential correction and it would be great if you could verify it. Do you build Domoticz yourself, or should I prepare a test build and if so for which OS (Windows PC, MacOS, Linux x86, Linux Raspberry Pi)?

Janco59:
@emontnemery: Never built it myself. It would probably mean a lot of work for me. If you would be able to make a test build for Raspberry, I would be happy to test it.

emontnemery:
~~OK, please test this build:~~
~~http://montnemery.com/domoticz_hue/linux_armv7l/012/domoticz_linux_armv7l.tgz~~
Edit:
Updated build, please try this instead:
http://montnemery.com/domoticz_hue/linux_armv7l/013/domoticz_linux_armv7l.tgz

Janco59:
Downloaded the build. Is there an easy way to update? I haven't done this before. Just installed the system from scratch a long time ago.

emontnemery:
@Janco59 
OK, since this is an unofficial build for testing, I'd recommend to first do a full backup (assuming domoticz is installed in /home/pi/domoticz:
```
cd /home/pi
sudo systemctl stop domoticz
tar -zcf domoticz_full_backup.tgz domoticz
```
[...]

Issue #2314, opened 2018-04-13
Hue:  add options to not import groups and scenes


jonferreira:
as per #2281 would be great to have options to not import hue groups / scenes available

emontnemery:
@jonferreira Please test this build: http://montnemery.com/domoticz_hue/linux_armv7l/012/domoticz_linux_armv7l.tgz

jonferreira:
Thanks @emontnemery 

Add Groups seem to be working as expected but the Add Scenes doesn't seem to do anything, i.e. regardless if it's checked or not Scenes always get added

emontnemery:
OK, sorry about that.
Please test this build: http://montnemery.com/domoticz_hue/linux_armv7l/013/domoticz_linux_armv7l.tgz

jonferreira:
Looks fine now, awesome!!

jonferreira:
@emontnemery believe this was new publised to Beta so can be closed?

emontnemery:
@jonferreira Yes, please close the ticket.

jonferreira:
t

PR #2348, opened 2018-04-25
Small improvements of philips hue integration

emontnemery:
Fixes:
Optionally block new groups and scenes for Philips Hue (#2314)
Turn off when setting color with brightness 0 (#2332)

### DIFF
--- a/hardware/PhilipsHue/PhilipsHue.cpp
+++ b/hardware/PhilipsHue/PhilipsHue.cpp
@@ -19,6 +19,8 @@
 using namespace std;
 
 #define HUE_DEFAULT_POLL_INTERVAL 10
+#define HUE_NOT_ADD_GROUPS 0x01
+#define HUE_NOT_ADD_SCENES 0x02
 
 //#define DEBUG_PhilipsHue
 
@@ -50,15 +52,15 @@ string ReadFile(string filename)
 }
 #endif
 
-CPhilipsHue::CPhilipsHue(const int ID, const string &IPAddress, const unsigned short Port, const string &Username, const int PollInterval, const int AddGroupsScenes) :
+CPhilipsHue::CPhilipsHue(const int ID, const string &IPAddress, const unsigned short Port, const string &Username, const int PollInterval, const int Options) :
 m_IPAddress(IPAddress),
 m_UserName(Username)
 {
 	m_HwdID=ID;
 	m_Port = Port;
 	m_poll_interval = PollInterval;
-	m_add_groups = !(AddGroupsScenes&1);
-	m_add_scenes = !(AddGroupsScenes&2);
+	m_add_groups = (Options & HUE_NOT_ADD_GROUPS) == 0;
+	m_add_scenes = (Options & HUE_NOT_ADD_SCENES) == 0;
 	m_stoprequested=false;
 
 	// Catch uninitialised Mode1 entry.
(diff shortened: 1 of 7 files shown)


gizmocuz:
Thanks for the changes !!
One (small) request, could you add two '#defines' in the .cpp file

```
#define USE_GROUPS 0x01
#define USE_SCENES 0x02
```

And then in the constructor use these for 'AddGroupsScenes' like:

```
m_add_groups = (AddGroupsScenes & USE_GROUPS) != 0;
m_add_scenes = (AddGroupsScenes & USE_SCENES) != 0;
```

Maybe change 'AddGroupsScenes' to a better name like 'iOptions' ?

emontnemery:
@gizmocuz Thanks for reviewing! Your comments should now be fixed.

gizmocuz:
Thanks! Are you sure this works:

m_add_groups = (Options & HUE_NOT_ADD_GROUPS) == 0

Should this not be != 0 ?

if (1 & 1) == 0 --> add groups ?

gizmocuz:
Hmmm you are using reversed options... thats not so great... could you make it a bit enabled option field ? We might need more settings in the feature

emontnemery:
Yes, the reason is to be backwards compatible, Mode2 is set to 0 by default.
Otherwise, a new database version + update code is needed to set Mode2 = HUE_ADD_GROUPS | HUE_ADD_SCENES.

gizmocuz:
But you can change the value of mode2 ... even if it is 0, you could say, okey enable everything.
Or you can make a database patch (See SQLHelper) to set the default mode2 value to 0x03 ?

emontnemery:
Exactly, I'll make the database patch to update default mode2 value.

emontnemery:
@gizmocuz The options are now changed to be non-inverted

gizmocuz:
Thanks!!!

jonferreira:
Bit confuse wether this was publised to beta or not. I'm on 3.9398 and can't see any new options?

emontnemery:
Yeah, should be in beta. Can you try in incognito tab to make sure it's not hidden due to browser cache not being refreshed?

jonferreira:
I'll be damned, stupid cache!

Had to manually flush the cache and remove all application data for the thing to show up - simple ctrl + f5 wasn't enough!

Thanks

Wob76:
Hi,

Update (solved?): I switched between a few of the standard themes, It options don't show on default or elemental themes (or non-standard ThinkTheme), but it they do show on element-dark, so ticked and working. Theme issue??

I have just installed a Hue Bridge, running Domoticz 3.9476. I don't see any scenes or groups being imported, I can see the settings, but no check boxes.

I restored the default theme, completed a full browser flush and tried incognito mode all without success. Any idea?

image

emontnemery:
@Wob76 Yeah, @gizmocuz usually states only one theme is supported in beta (not sure which one that is though).

gizmocuz:
@emontnemery , that is the default theme for now

Wob76:
@gizmocuz "default" or "element-dark"? as I see the above issue with the "default" theme.

gizmocuz:
@Wob76 , default theme
Scenes/Groups being imported does not have something to do with the theme ?

jonferreira:
@gizmocuz think his problem is he doesn't see the options to enable Scenes / Groups...
@Wob76 you need to flush your browser application cache...In Firefox for example you do that on the settings page. After that when you open Domoticz you should see the usual "refreshing browser cache" message and you should see those options I reckon...

gizmocuz:
In the settings you can hide/show the scene/groups

jonferreira:
@gizmocuz exactly my point, he doesn't have that option (checkbox is missing) as per the above screenshot ... could be cache or something off on the theme

gizmocuz:
The above screenshot is not from the settings page, but from the hardware setup

jonferreira:
ah my bad @gizmocuz didn't realised that. I'm running 9462 and the options to show / hide Hue Scenes / Groups are in the Hardware setup

image

gizmocuz:
I see, they are also gone here ;)
The solution is simple... moment

gizmocuz:
Fixed, will be solved in an hour (or two)

jonferreira:
nice! thanks @gizmocuz

gizmocuz:
https://github.com/domoticz/domoticz/commit/56688ba670547be76f756cfb552d01f51d27fbb6

PR #2350, opened 2018-04-25
Prevent plugin start / stop races

emontnemery:
Fix three races when stopping a plugin:

1. Queued messages for a stopping plugin are not always cleaned causing them to be processed after the plugin has been already been deleted
   Attempt to fix by clearing the message queue from CPlugin::Stop()
2. If a plugin is stopped before m_bIsStarted has been set, it will not be properly stopped.
  Attempt to fix by adding a "starting" state which is set before sending the `InitializeMessage` and cleared after setting m_bIsStarted in CPlugin::Start().
3. A plugin manipulating dictionaries after another plugin has just called Py_EndInterpreter().
Prevent this by calling PyEval_RestoreThread also before processing events and directives.

The first two races cause access to already freed memory, by the plugin framework and by the python parser, and are easily triggered by enabling/disabling/changing settings of a plugin from the WebUI.

The third race causes crashes in the python dictionary code.

The error handling in CPlugin::Start() and CPlugin::Initialise() might need some additional improvement to make sure the plugin does not end up in a semi started state if we fail while starting the plugin.

@dnpwwo, @gizmocuz Please review.

### DIFF
--- a/hardware/plugins/Plugins.cpp
+++ b/hardware/plugins/Plugins.cpp
@@ -408,13 +408,11 @@ namespace Plugins {
 		Name = sName;
 		m_bIsStarted = false;
 		m_bIsStarting = false;
-		_log.Log(LOG_STATUS, "(%s) CPlugin::CPlugin()", Name.c_str());
 	}
 
 	CPlugin::~CPlugin(void)
 	{
 		m_bIsStarted = false;
-		_log.Log(LOG_STATUS, "(%s) CPlugin::~CPlugin()", Name.c_str());
 	}
 
 	void CPlugin::LogPythonException()
@@ -657,7 +655,6 @@ namespace Plugins {
 
 	bool CPlugin::StartHardware()
 	{
-		_log.Log(LOG_STATUS, "(%s) enter CPlugin::StartHardware", Name.c_str());
 		if (m_bIsStarted) StopHardware();
 
 		//	Add start command to message queue
@@ -666,7 +663,6 @@ namespace Plugins {
 
 		_log.Log(LOG_STATUS, "(%s) Started.", Name.c_str());
 
-		_log.Log(LOG_STATUS, "(%s) leave CPlugin::StartHardware", Name.c_str());
 		return true;
 	}
 
@@ -702,7 +698,6 @@ namespace Plugins {
 
 	bool CPlugin::StopHardware()
 	{
-		_log.Log(LOG_STATUS, "(%s) enter CPlugin::StopHardware", Name.c_str());
 		try
 		{
 			_log.Log(LOG_STATUS, "(%s) Stop directive received.", Name.c_str());
@@ -711,13 +706,13 @@ namespace Plugins {
 			while (m_bIsStarting)
 			{
 				int scounter = 0;
-				while (m_bIsStarting && (scounter++ < 600))
+				while (m_bIsStarting && (scounter++ < 300))
 				{
 					sleep_milliseconds(100);
 				}
 				if (m_bIsStarting)
 				{
-					_log.Log(LOG_ERROR, "(%s) Plugin did not finish start after 60 seconds", Name.c_str());
+					_log.Log(LOG_ERROR, "(%s) Plugin did not finish start after 30 seconds", Name.c_str());
 				}
 			}
 
@@ -778,7 +773,6 @@ namespace Plugins {
 		}
 
 		_log.Log(LOG_STATUS, "(%s) Stopped.", Name.c_str());
-		_log.Log(LOG_STATUS, "(%s) leave CPlugin::StopHardware", Name.c_str());
 
 		return true;
 	}
--- a/hardware/plugins/Plugins.cpp
+++ b/hardware/plugins/Plugins.cpp
@@ -626,7 +626,8 @@ namespace Plugins {
 
 	void CPlugin::Notifier(std::string Notifier)
 	{
-		delete m_Notifier;
+		if (m_Notifier)
+			delete m_Notifier;
 		if (m_bDebug & PDM_PLUGIN) _log.Log(LOG_NORM, "(%s) Notifier Name set to: %s.", Name.c_str(), Notifier.c_str());
 		m_Notifier = new CPluginNotifier(this, Notifier);
 	}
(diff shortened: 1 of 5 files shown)


hardware/plugins/Plugins.cpp
```diff
@@ -240,7 +240,7 @@ namespace Plugins {
 				if (type == 1) type = PDM_ALL;
 
 				pModState->pPlugin->m_bDebug = (PluginDebugMask)type;
-				_log.Log(LOG_NORM, "(%s) Debug logging mask set to: %s%s%s%s%s%s%s%s%s", pModState->pPlugin->Name.c_str(), 
+				_log.Log(LOG_NORM, "(%s) Debug logging mask set to: %s%s%s%s%s%s%s%s%s", pModState->pPlugin->Name.c_str(),
```
gizmocuz, line 243: could you add this in a #define _DEBUG block so this is only printed when compiled in debug mode ?

hardware/plugins/Plugins.cpp
```diff
@@ -657,19 +658,64 @@ namespace Plugins {
 		if (m_bIsStarted) StopHardware();
 
 		//	Add start command to message queue
+		m_bIsStarting = true;
```
gizmocuz, line 661: I have to review the whole code, but why would you start something in a destructor ?

> A destructor is a special member function that is called when the lifetime of an object ends. The purpose of the destructor is to free the resources that the object may have acquired during its lifetime

hardware/plugins/Plugins.cpp
```diff
@@ -657,19 +658,64 @@ namespace Plugins {
...
+		}
+	}
+
 	bool CPlugin::StopHardware()
 	{
 		try
 		{
 			_log.Log(LOG_STATUS, "(%s) Stop directive received.", Name.c_str());
 
+			// loop on plugin to finish startup
```
gizmocuz, line 705: Can't plugins be started in blocking mode ? Then we know when 'm_bIsStarted' is true, it was started, else not... we do not need a m_bIsStarting then
Then you can use the

boost::lock_guard<boost::mutex> l(PluginMutex);

in the stophardware ?

Also in this routine 'm_Notifier' is deleted, i would also set this then to NULL after deletion

hardware/plugins/Plugins.cpp
```diff
@@ -813,6 +834,7 @@ namespace Plugins {
 			if (!m_PyInterpreter)
 			{
 				_log.Log(LOG_ERROR, "(%s) failed to create interpreter.", m_PluginKey.c_str());
+				m_bIsStarting = false;
```
gizmocuz: Instead of 100 times writing m_bIsStarting=false and return false, i would (yes..) in this case use gotoExitFalse;
And add under in the function a
gotoExitFalse:
 m_bIsStarting=false;
 return false;

gizmocuz (commented):
Thank you big time looking into this problem !!!
For me, the biggest question is, why would you start something in the destructor ?
There is also on 'delete' that does not set the variable to NULL, better safe then sorry

It seems plugins are started non-blocking... if that could be made blocking, there might be no readon to use m_bIsStarting ?

hardware/plugins/Plugins.cpp
emontnemery: It's not in the destructor, it's in CPlugin::StartHardware()

hardware/plugins/Plugins.cpp
emontnemery: I agree adding the m_bIsStarting complicates the logic a bit, and increases risk of new issues..
Unfortunately, I don't think starting the plugins in blocking mode is a good idea, it would mean a plugin such as the one I used for testing, https://github.com/ffes/domoticz-buienradar, would block the domoticz main thread for a very long time. That plugin does some http access at startup, and in my own testing this needs up to 30 s.

Agreed on setting to NULL after deletion.

hardware/plugins/Plugins.cpp
emontnemery: OK, will fix.

hardware/plugins/Plugins.cpp
emontnemery: @dnpwwo very recently did big changes to plugin framework in bf379c441ff72085ebe51b32d9008546acc1cefa to have much more fine grained control over prints, I think the purpose of this print is to make the plugin developer and user know that the prints are in backwards compatible mode. It would be better if @dnpwwo removes/hides it once he's sure everyone has adjusted to the new print control..

emontnemery:
@gizmocuz The review comments are fixed in the new commit, except for removing the print at plugin start or making plugin start block.

emontnemery:
New commit which should fix a third race related to manipulating python objects after calling `Py_EndInterpreter()`.
Note: Dictionary helper macros replaced with functions to make them visible in stack traces.

hardware/plugins/Plugins.cpp
```diff
@@ -626,8 +626,7 @@ namespace Plugins {
 
 	void CPlugin::Notifier(std::string Notifier)
 	{
-		if (m_Notifier)
-			delete m_Notifier;
+		delete m_Notifier;
```
gizmocuz: Please do not remove the useless checks! Always good to check.
What happens when the system is out of memory, and a new value could not be allocated and Notifier is called again ?
Or what happens when a developer made a mistake and calls notifier twice
It might look unneeded, but it is always good to check for it's state

emontnemery: @gizmocuz They are useless not because you can be certain m_Notifier != NULL (you can not as you point out), they are useless because there's no need to pollute the code with NULL-checks before calling `free()` or `delete` - it's allowed to free or delete NULL pointers.
Here's a SO discussion on the topic with references to the spec: https://stackoverflow.com/a/4190733

gizmocuz:
Yes, as long as you set the pointer back to zero
What my coding style is (and always will be)

```
if (a != NULL) {
  delete a;
  a=NULL;
}
```

Now this is what i consider good coding/practace

emontnemery:
OK, I'll re add the NULL-checks if you like them, and I guess in both cases some meaningless operation is done 😁 
```
if (a != NULL) { // Meaningless comparison with NULL, since it's perfectly fine to `delete` a null pointer
  delete a;
  a=NULL;
}
```
And in the other case:
```
delete a;
a = NULL; // Meaningless to set to NULL if 'a' was already NULL
```

Does anything else need to be changed?

gizmocuz:
No other changes
But setting a = NULL is definitively always a good idea after a free/delete

gizmocuz:
Thanks!!

gizmocuz:
I will commit some sanity changes like m_Notifier = NULL;
Why ? The new might return NULL for some strange out of memory reason... who knows

Issue #2355, opened 2018-04-28
Segmentation fault


remcom:
After upgrading this morning I have the following issue. (Updated to hash: 5a186ddc4663a5faba367e7ddc1943a28fd44340). The issue occurs right at startup of domoticz

> 2018-04-28 13:45:57.679  OpenZWave: Value_Added: Node: 44 (0x2c), CommandClass: BATTERY, Label: Battery Level, Instance: 1
> 2018-04-28 13:45:57.679  OpenZWave: Value_Added: Node: 45 (0x2d), CommandClass: SWITCH BINARY, Label: Switch, Instance: 1
> 2018-04-28 13:45:57.679  OpenZWave: Value_Added: Node: 45 (0x2d), CommandClass: SWITCH MULTILEVEL, Label: Level, Instance: 1
> 2018-04-28 13:45:57.680  Error: Device creation failed, Domoticz settings prevent accepting new devices.
> Segmentation fault

The last update I did was last thursday. So something i between causing this error.

Already try'd a clean build but it didnt fix it

emontnemery:
Please start through gdb to get a stack trace:
```
gdb ./domoticz
run
<wait for segmentation fault>
bt
```

remcom:
Sure. Give me a moment. I just backed your change out to get it up and running. need to pull and make it again.

remcom:
> 2018-04-28 14:35:19.685  OpenZWave: Value_Added: Node: 45 (0x2d), CommandClass: SWITCH MULTILEVEL, Label: Level, Instance: 1
> 2018-04-28 14:35:19.685  Error: Device creation failed, Domoticz settings prevent accepting new devices.
> 
> Thread 13 "domoticz" received signal SIGSEGV, Segmentation fault.
> [Switching to Thread 0x7fffe77fe700 (LWP 14055)]
> 0x0000555555c9d6dc in std::vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > >::operator[](unsigned long) ()
> (gdb) bt
> #0  0x0000555555c9d6dc in std::vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > >::operator[](unsigned long) ()
[...]

PR #2356, opened 2018-04-28
Fix crash in CSQLHelper::UpdateValueInt (#2355)

emontnemery:


gizmocuz:
Thanks!

Dylantje:
Dear Gizmocuz.....
Is there now only a solution for X64?
When do you think the X32 version is working wel?

Thanks

gizmocuz:
@Dylantje , not sure what you mean ? If it is about a 32bit arm build, should be available now

Dylantje:
Dear gizmocuz,,,, Everything works fine again, after this update
THANKS for the great work!!

PR #2358, opened 2018-04-29
Initialize member variables

emontnemery:
Main issue is member variable `m_bRFDebug` which controls debug prints, but was not set anywhere. Hence debug prints would randomly show up or not each time RFLink HW was started.


gizmocuz:
Thanks!

PR #2359, opened 2018-04-29
Initial NetBSD port

tih:
These are the changes I had to make to get domoticz compiled and working on NetBSD/amd64-current. There's not a lot - just tweaks to CMakeLists.txt to help it get the USB stuff right, the addition of NetBSD clauses to a few ifdefs (mostly just aligning with other BSDs), and the addition of an extra include file for the type formatting in the logging in all the places where you already include inttypes.h.

With these changes, and some accompanying tweaks to open-zwave (for which I've submitted a pull request, as well), I'm happily running domoticz with Aeotec and Z-Wave.Me interfaces for Z-Wave, and an RFXCOM interface for 433MHz, all attached as USB serial devices.


emontnemery:
@tih Instead of adding this everywhere:
```
+#ifdef __NetBSD__
+#include <machine/int_fmtio.h>
+#endif
```

Try to instead add compiler flags:
`-D__STDC_LIMIT_MACROS -D__STDC_CONSTANT_MACROS  -D__STDC_FORM
AT_MACROS`

Otherwise the NetBSD build will break as soon as someone adds use of the C formatting macros anywhere.

main/EventSystem.cpp
```diff
@@ -20,6 +20,9 @@
 #include "../json/json.h"
 #define __STDC_FORMAT_MACROS
 #include <inttypes.h>
+#ifdef __NetBSD__
+#include <machine/int_fmtio.h>
```
krytarowski: Why this header?

CMakeLists.txt
```diff
@@ -129,7 +129,14 @@ IF(HAVE_EXECINFO_H)
       ADD_DEFINITIONS(-DHAVE_EXECINFO_H)
     ENDIF()
   ELSE()
-    ADD_DEFINITIONS(-DHAVE_EXECINFO_H)
+    IF(CMAKE_SYSTEM_NAME STREQUAL "NetBSD")
```
krytarowski: Why not `CMake/FindBacktrace.cmake`?

MQTT/net_mosq.c
```diff
@@ -38,6 +38,10 @@ and the Eclipse Distribution License is available at
 #  include <netinet/in.h>
 #endif
 
+#ifdef __NetBSD__
+#  include <netinet/in.h>
```
krytarowski: join with `__FreeBSD__` above?

hardware/serial/serial.cpp
```diff
@@ -2,11 +2,11 @@
 #include "stdafx.h"
 #include <algorithm>
 
-#if !defined(_WIN32) && !defined(__OpenBSD__) && !defined(__FreeBSD__)
+#if !defined(_WIN32) && !defined(__OpenBSD__) && !defined(__FreeBSD__) && !defined(__NetBSD__)
 # include <alloca.h>
 #endif
 
-#if defined (__MINGW32__)
+#if defined (__MINGW32__) || defined(__NetBSD__)
```
krytarowski: This looks like a need for `-std=g++11` instead of redefining alloca.

main/EventSystem.cpp
tih: I have no idea. I swear it failed to compile, complaining about text strings that were being built like "...  '%" PRIu64 "' ...", and when I (not having noticed that it would get included anyway) explicitly did include machine/int_fmtio.h, it stopped complaining.

I've removed all those extra includes now, and it's compiling just fine. Of course, I now have GCC 6.4, while I had 5.5 when I made that change, so it's no longer the same environment.

Anyway, thanks to both of you for pointing this out!

CMakeLists.txt
tih: Probably because I'm clueless about cmake. I was just using the mechanisms I could see being used in CMakeLists.txt already. If this should have been done differently, please show me how!

MQTT/net_mosq.c
tih: Ah, yes. A bit too quick, there. Fixed.

hardware/serial/serial.cpp
tih: I did that (well, actually -std=c++11), but it still ends up saying, at link time:

CMakeFiles/domoticz.dir/hardware/serial/serial.cpp.o: In function `serial::Serial::readline(std::string&, unsigned long, std::string)':
serial.cpp:(.text+0x4c9): warning: Warning: reference to the libc supplied alloca(3); this most likely will not work. Please use the compiler provided version of alloca(3), by supplying the appropriate compiler flags (e.g. not -std=c89).

tih:
Updated to fix the obvious problems pointed out by krytarowski. Haven't made any further changes to CMakeLists.txt, but I'd certainly like to see suggestions for improvements there. After removing the ugly alloca-related change in hardware/serial/serial.cpp, I now get a link time warning, that probably ought to be resolved before this is production ready.

Update: it works in this condition, too, though, including serial communication with the RFXCOM unit.

hardware/serial/serial.cpp
krytarowski: If you use c++11 instead of gnu++11, alloca is disabled and there is need for hack like this define to __builtin_alloca.

main/domoticz.cpp
```diff
@@ -353,7 +353,7 @@ void daemonize(const char *rundir, const char *pidfile)
 	{
 		return GetModuleFileNameA(NULL, pathName, (DWORD)pathNameCapacity);
 	}
-#elif defined(__linux__) || defined(__CYGWIN32__) /* elif of: #if defined(_WIN32) */
+#elif defined(__linux__) || defined(__CYGWIN32__) || defined(__NetBSD__) /* elif of: #if defined(_WIN32) */
```
krytarowski: Do you need to run this on NetBSD<8.0? New code should rather use sysctl(3) similar to FreeBSD below.

hardware/serial/serial.cpp
tih: Ah, I didn't know about gnu++11 - thanks!

main/domoticz.cpp
tih: Ah, yes - I see we're almost compatible with FreeBSD for this lookup. Fixing now.

tih:
No more warnings about alloca, and we now get our own executable name the modern way, using sysctl(3), as suggested by krytarowski.

Tested, and working.

main/domoticz.cpp
```diff
@@ -374,6 +374,19 @@ void daemonize(const char *rundir, const char *pidfile)
 		sysctl(mib, 4, pathName, &cb, NULL, 0);
 		return cb;
 	}
+#elif defined(__NetBSD__)
```
krytarowski: We could join this with FreeBSD and ifdef `mib[...]` part.

hardware/serial/serial.cpp
```diff
@@ -2,7 +2,7 @@
 #include "stdafx.h"
 #include <algorithm>
 
-#if !defined(_WIN32) && !defined(__OpenBSD__) && !defined(__FreeBSD__)
+#if !defined(_WIN32) && !defined(__OpenBSD__) && !defined(__FreeBSD__) && !defined(__NetBSD__)
 # include <alloca.h>
```
krytarowski: We can reverse this and include `<alloca,h>` for `__sun` only. Everybody else can include `<stdlib.h>`.

push/InfluxPush.cpp
```diff
@@ -375,4 +375,4 @@ namespace http {
 			root["title"] = "DeleteInfluxLink";
 		}
 	}
-}
```
krytarowski: unrelated change in unrelated file

push/FibaroPush.cpp
```diff
@@ -400,4 +400,4 @@ namespace http {
 			root["title"] = "DeleteFibaroLink";
 		}
 	}
-}
```
krytarowski: unrelated change in unrelated file

hardware/serial/serial.cpp
krytarowski: Alternatively we can detect alloca.h with cmake.

main/domoticz.cpp
tih: Less duplication - I agree.

hardware/serial/serial.cpp
tih: That does look nicer. As for detecting stuff with cmake, I'll take your word for it. :)

tih:
Tidied up as suggested by krytarowski. Tested and working.

hardware/serial/serial.cpp
```diff
@@ -2,8 +2,10 @@
 #include "stdafx.h"
 #include <algorithm>
 
-#if !defined(_WIN32) && !defined(__OpenBSD__) && !defined(__FreeBSD__)
+#ifdef __sun
```
krytarowski, line 5: I like more:

```
#ifdef __sun
# include <alloca.h>
#endif
#include <stdlib.h>
```
Just in case that SmartOS might want both for some reason.

tih: Agreed.

CMakeLists.txt
krytarowski:
```
find_package(Backtrace)
```

You can use `${Backtrace_FOUND}`, `${Backtrace_HEADER}`.

tih: I really don't understand that, Kamil. The fragment of cmakefile you tie this to is part of what's a pretty ugly nest of conditionals - is that what you'd like to see cleaned up? I've discovered that cmake has elseif(), which I would have used for the NetBSD part if I'd known. It should have been used already for the FreeBSD bit, but seeing as it wasn't, I just assumed it had to be that way, and continued in the same fashion.

krytarowski: Please detect Backtrace without homegrown OS-specific checks.

tih: I finally see what you mean. Sorry for being so slow. :)

I think changing the mechanism for resolving the execinfo/backtrace thing is beyond the scope of my just getting the existing implementation to work on NetBSD, and it would be overly ambitious for me to submit patches to do so.

What I will do is to simplify the conditional clauses a bit: I've added more complex nesting than I needed to in a couple of places, and I've also got that test for libusb1, which I can actually just lose, because I've since sorted that out in open-zwave itself, which is where it's needed.

tih:
Tidied up a bit more, now. As krytarowski points out, the execinfo stuff can be simplified by using the built-in find_package() mechanism in cmake, but I don't feel comfortable making that change, and I think it's a bit outside the scope of this pull request, anyway.

I now believe this is ready. It does what little it needs to to get things built, and working correctly, on NetBSD, and doesn't disturb anything else.

krytarowski:
Looks good to me.

gizmocuz:
Thanks!!

PR #2372, opened 2018-05-03
Decrease risk of deadlock in CLogger::Log

emontnemery:
Decrease size of critical region in CLogger::Log to decrease risk of deadlocks.

Without this change, there will for example be a deadlock if an exception is triggered in the call to  `vsnprintf` by trying to print a bad string pointer as the error handler is itself using CLogger::Log to print.


gizmocuz:
Thanks!

PR #2374, opened 2018-05-04
Harmony Hub bugfixes + user requests

gordonb3:
Bug: when communication is lost, switches can be in an impossible state of multiple activities running at the same time

Bug: possibility to power off the PowerOff state lead to a state where you cannot see the current activity

Bug: sending a switch off request on any activity would activate PowerOff activity regardless of whether the associated activity was running

Added: when we are disconnected from the Hub, do an instant connect if user sends a switch command

Added: allow silencing the error message when Hub cannot be reached after a Domoticz restart (by sending disallowed PowerOff `off` request)

### DIFF
--- a/hardware/HarmonyHub.cpp
+++ b/hardware/HarmonyHub.cpp
@@ -108,7 +108,12 @@ bool CHarmonyHub::WriteToHardware(const char *pdata, const unsigned char length)
 	}
 	else if ((pCmd->LIGHTING2.packettype == pTypeLighting2) && (pCmd->LIGHTING2.cmnd==0))
 	{
-		if (!SubmitCommand(START_ACTIVITY_COMMAND, "-1",""))
+		if (m_szCurActivityID == "-1") // already powered off
+		{
+			SubmitCommand(START_ACTIVITY_COMMAND, "-1", ""); // send it anyway, user may be trying to correct a wrong state
+			return false; // should I do something else here to prevent making this look like an error?
+		}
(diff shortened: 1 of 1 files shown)


heggink:
Hey @gordonb3 will this fix my log being spammed with:
```
2018-05-05 05:55:51.718 Error: Harmony Hub: Error pinging server.. Resetting connection.
2018-05-05 05:56:31.736 Error: Harmony Hub: Error pinging server.. Resetting connection.
2018-05-05 05:57:42.660 Error: Harmony Hub: Error pinging server.. Resetting connection.
2018-05-05 05:59:54.366 Error: Harmony Hub: Error pinging server.. Resetting connection.
2018-05-05 06:02:39.877 Error: Harmony Hub: Error pinging server.. Resetting connection.
2018-05-05 06:03:19.856 Error: Harmony Hub: Error pinging server.. Resetting connection.
2018-05-05 06:03:57.408 Error: Harmony Hub: Error pinging server.. Resetting connection.
2018-05-05 06:04:04.446 Error: Harmony Hub: Start communication failed
```
until I disable and enable the device? Happens every other day.

gordonb3:
No. Every individual loss of communication will still be logged like before.

The log entries shown do indicate something strange happening though. Apparently you can connect, but at the first ping (which is a protocol requirement of the Hub) you do after that the Hub gives a wrong or no response. The timings also do seem stretched. In between a connection reset and the next ping following a reconnect there should be ~35 seconds - you have 38-42 seconds

gizmocuz:
Thanks!!

PR #2375, opened 2018-05-04
Correct confusion between character literal and integer literal

emontnemery:
Fixes regressions introduced by #2174


gizmocuz:
Thanks!!

PR #2377, opened 2018-05-05
Prevent additional plugin start / stop race

emontnemery:
Call `PyEval_RestoreThread` earlier to prevent races when restarting the Python plugin framework.
Without this change, 100% reproducible crash when restoring Domoticz database from WebUI.

### DIFF
--- a/hardware/plugins/PluginMessages.h
+++ b/hardware/plugins/PluginMessages.h
@@ -71,6 +71,7 @@ namespace Plugins {
 		void Process()
 		{
 			boost::lock_guard<boost::mutex> l(PythonMutex);
+			m_pPlugin->RestoreThread();
 			ProcessLocked();
 		};
 		virtual const char* PythonName() { return m_Callback.c_str(); };
--- a/hardware/plugins/Plugins.cpp
+++ b/hardware/plugins/Plugins.cpp
@@ -707,13 +707,14 @@ namespace Plugins {
 			while (m_bIsStarting)
 			{
 				int scounter = 0;
-				while (m_bIsStarting && (scounter++ < 300))
+				int timeout = 30;
+				while (m_bIsStarting && (scounter++ < timeout*10))
 				{
 					sleep_milliseconds(100);
 				}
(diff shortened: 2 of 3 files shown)


emontnemery:
@dnpwwo, @gizmocuz Please review.

@dnpwwo: This timout:
```
while (m_bIsStarted && (scounter++ < 50))
{
	sleep_milliseconds(100);
}
if (m_bIsStarted)
{
	_log.Log(LOG_ERROR, "(%s) Plugin did not stop after 5 seconds, flushing event queue...", Name.c_str());

	ClearMessageQueue();
	m_bIsStarted = false;
}
```
Is really bad, if we get a timeout the Plugin will keep using freed memory which means the behavior of the domoticz process is completely undefined afterwards.
I suggest to:
- Increase the timeout significantly, let's say to 60s, to allow a blocking resolve (for example) to complete.
- If there is a timeout, kill the thread and call abort() so domoticz can be restarted instead of continuing with corrupted heap

Edit:
Can we interrupt the plugin by calling `PyThreadState_SetAsyncExc()` ?

dnpwwo:
I hate this timout as well and only put it in because @gizmocuz requested it.  

Boost asio is pretty good at not blocking and I used all the async (and abortable) variants of functions but I had a UDP/IP connection that didn't terminate the other day and the plugin exited after the timeout expired. I still saw updates from the connection from a ghost plugin after that.  No idea why it wasn't seg faulting.

 I agree this can cause more problems than it fixes.

emontnemery:
@dnpwwo The plugin worker thread might never exit since the `OnStopCallback` will be discarded when cleaning up the message queue after the timeout expires..
In your test where you saw the timeout, were you by any chance doing anything which might have triggered `MainWorker::StopDomoticzHardware`? I just found a deadlock in it, fixed in the latest commit.

dnpwwo:
@emontnemery you are probably right. It goes to show that forcing hardware to exit is not a good idea

gizmocuz:
Thanks!!

PR #2378, opened 2018-05-05
Plugins devicemodified bug 32bits

guillaumezin:
CPluginSystem::DeviceModified has no effect on 32bits system because ID is not parsed using PRIu64

 Call onDeviceModified in python plugin when UpdateDevice is used in Lua or dzEvents scripts


emontnemery:
@guillaumezin It doesn't build for Windows, I think you just need to `#include <inttypes.h>` in PluginManager.cpp
Also, why does the PR include a lot of unrelated commits?

guillaumezin:
inttypes included.

For the unrelated commits, I don't know, I think that's because I need to rebase my git, I tried but it was worse, I'm not an expert...

guillaumezin:
ok, this is a mess, I close it and I will repost it when clean

PR #2379, opened 2018-05-06
Extract text log to own page

stas-demydiuk:


gizmocuz:
Could you add a screenshot here what has changed ?
There is already a 'error' tab in the log

gizmocuz:
Sorry, I now understand what this is about.... Thanks!

PR #2382, opened 2018-05-08
Handle ID as 64 bits in CPluginSystem::DeviceModified

guillaumezin:
And call onDeviceModified in python plugin when UpdateDevice is used in Lua or dzEvents scripts


gizmocuz:
Thanks!!

PR #2386, opened 2018-05-12
Add generic CustomCommand handling for scripts

dwmw2:
The Onkyo-specific support in JSON and dzVents isn't really sufficient,
because it (quite rightly) needs authentication. But dzVents just assumes
it can make unauthenticated calls to localhost:8080, and doesn't work.

I've been working around this by spawning an external script to invoke
curl, but *that* only works over HTTPS not HTTP for some reason (without
HTTPS I just get an 'Unauthorized' response). And that takes literally
*seconds* to work, on the slow machine I'm using.

Fix it up so that the script command arrays can have a 'CustomCommand',
done generically so that it can easily cover devices other than Onkyo.

### DIFF
--- a/hardware/DomoticzHardware.h
+++ b/hardware/DomoticzHardware.h
@@ -16,6 +16,7 @@ public:
 	bool Start();
 	bool Stop();
 	virtual bool WriteToHardware(const char *pdata, const unsigned char length)=0;
+	virtual bool CustomCommand(const uint64_t idx, const std::string &sCommand);
 
 	void EnableOutputLog(const bool bEnableLog);
 
--- a/main/EventSystem.cpp
+++ b/main/EventSystem.cpp
@@ -2484,6 +2484,21 @@ bool CEventSystem::parseBlocklyActions(const _tEventItem &item)
 			actionsDone = true;
 			continue;
 		}
+		else if (deviceName.find("CustomCommand:") == 0)
+		{
+			int idx = atoi(deviceName.substr(14).c_str());
+			float afterTimerSeconds = 0;
+			size_t aFind = doWhat.find(" AFTER ");
+			if ((aFind > 0) && (aFind != std::string::npos)) {
+				std::string delayString = doWhat.substr(aFind + 7);
+				afterTimerSeconds = static_cast<float>(atof(delayString.c_str()));
+				doWhat = doWhat.substr(0, aFind);
+				StripQuotes(doWhat);
+			}
+			m_sql.AddTaskItem(_tTaskItem::CustomCommand(afterTimerSeconds, idx, doWhat));
+			actionsDone = true;
+			continue;
+		}
 		else
 		{
 			_log.Log(LOG_ERROR, "EventSystem: Unknown action sequence! (%s)", csubstr.c_str());
@@ -2626,6 +2641,19 @@ bool CEventSystem::PythonScheduleEvent(std::string ID, const std::string &Action
 				return false;
 		}
 
+		return true;
+	} else if(ID.find("CustomCommand:") == 0) {
+		int idx = atoi(ID.substr(14).c_str());
+		std::string doWhat = std::string(Action);
+		float afterTimerSeconds = 0;
+		size_t aFind = Action.find(" AFTER ");
+		if ((aFind > 0) && (aFind != std::string::npos)) {
+			std::string delayString = doWhat.substr(aFind + 7);
+			doWhat = doWhat.substr(0, aFind);
+			afterTimerSeconds = static_cast<float>(atof(delayString.c_str()));
+			StripQuotes(doWhat);
+		}
+		m_sql.AddTaskItem(_tTaskItem::CustomCommand(afterTimerSeconds, idx, doWhat));
 		return true;
 	}
 	return ScheduleEvent(ID, Action,eventName);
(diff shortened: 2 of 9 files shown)


dwmw2:
Ping @dannybloe

dwmw2:
Now I can cope with the fact that the Panasonic hardware driver can only turn the TV *off* not on. By having a script which checks for the attempt to turn it on, then sends 'CTVPWRON' via the amp.

And more to the point, it works within 5 seconds instead of taking the SSL setup time.

gizmocuz:
Is this a timed command or a queued command ? I do not see where you specify the delay as parameter.
This looks not like a queued_command

dwmw2:
Hm, because I don't actually pass in a DelayTime from the three separate ( :( ) script parsing things? I think I was cutting and pasting the SetSetPoint code, which doesn't either. In fact many of them seem not to...

gizmocuz:
Thats also the idea.
But you explicitly named this a 'timed' command, then in this case you should either add a delay time, so it is a timed command, or rename it to a queued command ?

dwmw2:
I hadn't actually noticed the distinction. Away for the (long) weekend but will try to snatch some time to power the computer up and look. Most likely Tuesday though. Thanks.

gizmocuz:
Thanks, or maybe 'non blocking' ... but you might also add a time so it can be used by others as well

dwmw2:
Erm... I need to cut and paste the code around `doWhat.find(" AFTER ");` in for example the "Variable:" implementation around EventSystem.cpp line 2238?

gizmocuz:
Question, you are trying to connect to HTTPS on your domoticz and get a 'Unauthorized' response ?
You are connecting to https://127.0.0.1 ? (localhost) and have local host setup in your local networks in domoticz so it does not need authentication ?

dwmw2:
HTTPS is fine (but slow as TLS setup takes ages especially when this little box is talking to itself); it was HTTP which failed. ISTR that happened even when using the real hostname, not just localhost. Editing the "local networks" list didn't make a difference. I think that even when local auth is bypassed it doesn't make it a request with "admin" privs.

dwmw2:
> Erm... I need to cut and paste the code around doWhat.find(" AFTER "); in for example the "Variable:" implementation around EventSystem.cpp line 2238?

Done that; testing now...

dwmw2:
Tested thus:
```
return {
        on = {
              	scenes = {
                        'Radio 4'
                }
        },
	execute = function(domoticz, device)
            domoticz.log('Device ' .. device.name .. ' was changed to ' .. device.state , domoticz.LOG_INFO)
            if (device.state == 'On') then
                domoticz.devices('Master power').onkyoEISCPCommand('SLI2B')
                domoticz.devices('Master power').onkyoEISCPCommand('MVL19').afterSec(0.2)
                domoticz.devices('Master power').onkyoEISCPCommand('ZPW01').afterSec(0.4)
                domoticz.devices('Master power').onkyoEISCPCommand('ZVL23').afterSec(0.6)
                domoticz.devices('Master power').onkyoEISCPCommand('NTC4').afterSec(0.8)
            end
        end
}
```

dwmw2:
This is actually really useful, since the amp is a bit unreliable at processing commands sent in quick succession. Thanks for making me fix that :)

dwmw2:
@dannybloe this should be ready to go now. Most of it is on the Domoticz side apart from the one-liner to fix OnkyoEISCPCommand to use the new feature. Do you want that split into a separate patch?

gizmocuz:
It will be merged when I checked it
There is a big patch coming up

dwmw2:
Thanks. Once the dust settles I may be inspired to clean up and unify some more of the script handling -- a lot of it has three separate copies of the same thing and could be consolidated.

hardware/DomoticzHardware.h
```diff
@@ -16,6 +16,7 @@ class CDomoticzHardwareBase
 	bool Start();
 	bool Stop();
 	virtual bool WriteToHardware(const char *pdata, const unsigned char length)=0;
+	virtual bool CustomCommand(uint64_t idx, const std::string &sCommand);
```
gizmocuz: could you make idx a const ?

main/EventSystem.cpp
```diff
@@ -2488,6 +2488,21 @@ bool CEventSystem::parseBlocklyActions(const _tEventItem &item)
 			actionsDone = true;
 			continue;
 		}
+		else if (deviceName.find("CustomCommand:") == 0)
```
jvandenbroek, line 2487: The colon isn't used when parsing Blockly and Python. You also forgot to set afterTimerSeconds in the AddTaskItem line.

hardware/DomoticzHardware.h
dwmw2: Done.

main/EventSystem.cpp
dwmw2: Hm, Blockly and Python really do have the colon for things like `Variable:`, `SetSetpoint:` etc.
Fixed the `afterTimerSeconds` one; thanks.

dwmw2: Hm, I note we are inconsistent about whether it's `SetSetPoint:` or `SetSetpoint:` though. Lua uses the former while Blockly and Python use the latter.

dwmw2: `$reasons_to_unify_this_stuff++;` :)

gizmocuz: Should be Setpoint (one word, see wikipedia)

dwmw2: We'll probably have to make it accept both now, for compatibility. But that's all for a different PR anyway.

gizmocuz: i agree, but should be changed soon else we keep using both ;)

PR #2388, opened 2018-05-13
Adding CharSize:8bits fallback to TeleinfoSerial (default is 7bits)

DenisMartin0:
I've added a fallback to 8bits if serial.open fail with the default 7bits value.
This feature is usefull to use any virtual device (like pseudo-terminal) to feed TeleinfoSerial.

I've also corrected a bug in ASyncSerial : if an exception occured while opening, the serial port will never be correctly closed.


DenisMartin0:
Successfully tested on Win10 & Raspberry.
I'll use such a feature with a tty multiplexer like this one : [ttybus](https://github.com/danielinux/ttybus)

gizmocuz:
Thanks!!

Issue #2254, opened 2018-04-01
Realtime compression (gzip) of html/js files during runtime/packaging


gizmocuz:
For speed (specially for arm systems) we probably need to continue serving compressed js files

JS libraries that are not part of Domoticz but from external sources we "could" already compress to GZip (.gz) like it is already so 90%

For libraries from Domoticz, we should have them uncompressed in our repo, so other developer can quickly work with them.

js/domoticz.js
js/domoticzblocks.js
js/domoticzblocks_messages_en.js
js/domoticzdevices.js

Also all controllers in the app/ folder

Question is, do we need to compress them ? Because are they not cached by the browser already ?
If not, we're going to need a PR that decompressed them and removes the compressed once.

If yes, we can think of compressing them during runtime or when packaging

What i think is a good idea, is to compress them in runtime.

During startup of domoticz, remove all .gz files. This means, no libraries should currently be compressed.
(we should however use the minimized versions where possible)
Then if we see that a javascript file is being requested, check if we have a .gz, if not, compress the library
serve the compressed .gz

That is once solution, and probably the easiest one.

Another is to use a (md5) hash, and after startup, and on first request, check the source and the uncompressed .gz, if there is a difference (or the compressed does not exist) (re) create the compressed and serve this

Any thoughts/ideas would be welcome

MAkerboom:
Why on runtime? It makes more sense to me to make this part of packaging

gizmocuz:
because a lot of (open source) users will compile/run domoticz from source
They do not use packaging so in runtime this will still work

But it was just a suggestion ... ;) Counter question, why not at runtime ?

MAkerboom:
I'm used to nodejs and react. There you compile a production build. Which creates a main.js and .css with all the needed code for the solution. So I guess making it part of the make-script seems logical

gizmocuz:
@MAkerboom , but those files you mentioned are not compressed ?

MAkerboom:
It's not default, but I guess if I look into it there is someone who added a module to do .gz while creating a production build for sure.

But the point I try to make: In development we retain a logical overview and readable source (uncompressed, separate and un-minified files). During the build, we create the binary, minify and compress for optimal performance. (also the appcache is now a static file and could be created during build, in my nodejs/react example a service-worker for caching is created automatically during production build)

emontnemery:
@gizmocuz Are you working on this? Otherwise I'll have a go at build time compression next week.

[...]

PR #2390, opened 2018-05-14
Runtime compression of .js and .html

emontnemery:
As discussed in #2254.

Implemented solution:
- .js and .html will be automatically compressed with gzip
- compressed version will be automatically recreated if it is outdated
- gzipped files stored separately, and ignored by git

Possible improvement points:
- Do something more advanced than just checking modification time

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,3 +44,4 @@ install_manifest.txt
 /.idea_/
 /www/images/floorplans/*.png
 /www/images/floorplans/*.jpg
+/www/.gz_cache/
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -547,7 +547,7 @@ set(Boost_USE_STATIC_LIBS ${USE_STATIC_BOOST})
 set(Boost_USE_MULTITHREADED ON)
 unset(Boost_INCLUDE_DIR CACHE)
 unset(Boost_LIBRARY_DIRS CACHE)
-find_package(Boost REQUIRED COMPONENTS thread date_time system filesystem)
+find_package(Boost REQUIRED COMPONENTS thread date_time system)
 
 if(USE_STATIC_BOOST)
    message(STATUS "Linking against boost static libraries")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -547,7 +547,7 @@ set(Boost_USE_STATIC_LIBS ${USE_STATIC_BOOST})
 set(Boost_USE_MULTITHREADED ON)
 unset(Boost_INCLUDE_DIR CACHE)
 unset(Boost_LIBRARY_DIRS CACHE)
-find_package(Boost REQUIRED COMPONENTS thread date_time system)
+find_package(Boost REQUIRED COMPONENTS thread date_time system filesystem)
 
 if(USE_STATIC_BOOST)
    message(STATUS "Linking against boost static libraries")
--- a/webserver/request_handler.cpp
+++ b/webserver/request_handler.cpp
@@ -204,6 +204,60 @@ void request_handler::handle_request(const request& req, reply& rep)
 	handle_request(req, rep, mInfo);
 }
 
+static bool endsWith(const std::string& str, const std::string& suffix)
+{
+    return str.size() >= suffix.size() && 0 == str.compare(str.size()-suffix.size(), suffix.size(), suffix);
+}
+
+static void validate_gzip(std::string full_path, std::string full_path_gzip)
+{
+	// Do nothing if browser requests a gzipped file
+	if (endsWith(full_path, ".gz"))
+		return;
+
+	struct stat result, resultgz;
+
+	bool gzip_valid = true;
+
+	if (stat(full_path.c_str(), &result) == 0)
+	{
+		if (stat(full_path_gzip.c_str(), &resultgz) == 0)
+		{
+			if (difftime(result.st_mtime, resultgz.st_mtime) > 0)
+			{
+				// gzipped version is old
+				gzip_valid = false;
+			}
+		}
+		else
+		{
+			// gzipped version does not exist
+			gzip_valid = false;
+		}
+	}
+
+	if (!gzip_valid)
+	{
+		std::ifstream is(full_path.c_str(), std::ios::in | std::ios::binary);
+		gzFile gzf = gzopen(full_path_gzip.c_str(), "wb");
+		if (is.is_open() && gzf != NULL)
+		{
+			std::string content((std::istreambuf_iterator<char>(is)),
+				(std::istreambuf_iterator<char>()));
+
+			bool err = false;
+			if (gzwrite(gzf, content.c_str(), content.size()) != content.size()) err = true;
+			if (gzclose(gzf) != Z_OK) err = true;
+
+			if (err){
+				std::remove(full_path_gzip.c_str());
+				_log.Log(LOG_ERROR, "Webserver: Failed to compress '%s' to '%s'", full_path.c_str(), full_path_gzip.c_str());
+			}
+			else _log.Log(LOG_STATUS, "Webserver: Compressed '%s' to '%s'", full_path.c_str(), full_path_gzip.c_str());
+		}
+	}
+}
+
 void request_handler::handle_request(const request &req, reply &rep, modify_info &mInfo)
 {
   mInfo.mtime_support = false;
@@ -272,6 +326,8 @@ void request_handler::handle_request(const request &req, reply &rep, modify_info
 		  //first try gzip version
 		  if (bHaveGZipSupport)
 		  {
+			  // Validate gzipped file
+			  validate_gzip(doc_root_ + request_path, doc_root_ + request_path + ".gz");
 			  std::string full_path = doc_root_ + request_path + ".gz";
 			  // Open the file to send back.
 			  std::ifstream is(full_path.c_str(), std::ios::in | std::ios::binary);
--- a/webserver/request_handler.cpp
+++ b/webserver/request_handler.cpp
@@ -13,7 +13,6 @@
 #include <fstream>
 #include <sstream>
 #include <string>
-#include <boost/filesystem.hpp>
 #include <boost/lexical_cast.hpp>
 #include <boost/scoped_array.hpp>
 #ifdef WIN32
@@ -26,6 +25,8 @@
 #include "request.hpp"
 #include "cWebem.h"
 #include "GZipHelper.h"
+
+#include "../main/Helper.h"
 
 // remove
 #include "../main/Logger.h"
@@ -240,14 +241,15 @@ static void validate_gzip(std::string full_path, std::string full_path_gzip)
 	if (!gzip_valid)
 	{
 		// Create directory
-		boost::filesystem::path p = full_path_gzip;
-		p.remove_filename();
-		try {
-			boost::filesystem::create_directories(p);
+		std::string dir = full_path_gzip;
+		std::string::size_type pos = dir.rfind('/');
+		if (pos != std::string::npos) {
+			dir.erase(pos);
 		}
-		catch (...)
+		int res = mkdir_deep(dir.c_str(), 0755);
+		if (res != 0 && errno != EEXIST)
 		{
-			_log.Log(LOG_ERROR, "Webserver: Failed to create directory '%s'", p.string().c_str());
+			_log.Log(LOG_ERROR, "Webserver: Failed to create directory '%s', errno: %d", dir.c_str(), errno);
 		}
 
 		// Compress
--- a/webserver/request_handler.cpp
+++ b/webserver/request_handler.cpp
@@ -13,6 +13,7 @@
 #include <fstream>
 #include <sstream>
 #include <string>
+#include <boost/filesystem.hpp>
 #include <boost/lexical_cast.hpp>
 #include <boost/scoped_array.hpp>
 #ifdef WIN32
@@ -238,6 +239,18 @@ static void validate_gzip(std::string full_path, std::string full_path_gzip)
 
 	if (!gzip_valid)
 	{
+		// Create directory
+		boost::filesystem::path p = full_path_gzip;
+		p.remove_filename();
+		try {
+			boost::filesystem::create_directories(p);
+		}
+		catch (...)
+		{
+			_log.Log(LOG_ERROR, "Webserver: Failed to create directory '%s'", p.string().c_str());
+		}
+
+		// Compress
 		std::ifstream is(full_path.c_str(), std::ios::in | std::ios::binary);
 		gzFile gzf = gzopen(full_path_gzip.c_str(), "wb");
 		if (is.is_open() && gzf != NULL)
@@ -326,12 +339,20 @@ void request_handler::handle_request(const request &req, reply &rep, modify_info
 		  //first try gzip version
 		  if (bHaveGZipSupport)
 		  {
-			  // Validate gzipped file
-			  validate_gzip(doc_root_ + request_path, doc_root_ + request_path + ".gz");
-			  std::string full_path = doc_root_ + request_path + ".gz";
+			  // Validate cached gzipped file
+			  validate_gzip(doc_root_ + request_path, doc_root_ + "/.gz_cache" + request_path + ".gz");
 			  // Open the file to send back.
+			  // Try to open from gzip cache
+			  std::string full_path = doc_root_ + "/.gz_cache" + request_path + ".gz";
 			  std::ifstream is(full_path.c_str(), std::ios::in | std::ios::binary);
-			  if (is)
+			  if (!is)
+			  {
+				  // Try to open directly
+				  full_path = doc_root_ + request_path + ".gz";
+				  is.open(full_path.c_str(), std::ios::in | std::ios::binary);
+			  }
+
+			  if (is.is_open())
 			  {
 				  mInfo.delay_status = false;
 				  if (not_modified(full_path, req, rep, mInfo)) {


emontnemery:
@gizmocuz Please review, I hope this is not too far off from what you had in mind in #2254.

gordonb3:
Looks like a lot better idea that doing on the fly compression on every call, but it seems like overkill to me to keep a server side cache of compressed static files that are supposed to be cached by the client as `site data`.

Also think the compressed files should not reside in the same file tree, allowing it e.g. to be put in a pseudo file system. And it should be optional.

gizmocuz:
@gordonb3 It should only be done after a start and it does not need to check or do this a second time while running
Next it is cached and should not even be requested again
The compressed files (except the libraries) should indeed be uncompressed in the file tree. That's the idea of these changes 👍

@emontnemery I will asap..

gizmocuz:
@emontnemery , i still have to do a good review, but what i had in mind:

at startup, check which files do not match anymore with the compressed files and re-compress them
i 100% agree a 'cache' folder would be perfect to store the compressed files

gordonb3:
I get the one time thing. It's the `should not even be requested again` thing that kind of hints towards `what's the point` of letting the machine do all that work to compensate something that in most cases does not require compensating?

As a side note: since these files are linked to the running Domoticz version, you could also choose to do the gzip during build of the application. Doing it at at first request like this PR will make Domoticz appear very slow for that first user hitting it.

In any case, you'll have to choose right now where to store the compressed files. Because if you decide to move them into their own subtree some time later that will not automatically clean up the folders where you stored them before. With no clear plan, this can make things very messy.

emontnemery:
@gizmocuz OK, I will change to a "cache" folder and add a .gitignore entry for that folder.
However, I don't understand this: _at startup, check which files do not match anymore with the compressed files and re-compress them_
If you do it like that, instead of checking mtime (or perhaps a hash if time travelling is a concern) on the fly, then domoticz has to be restarted after any change of html or js, and the whole compression story would make more sense as a build step.

@gordonb3 I tested the time needed to compress **all** currently uncompressed js and html on a raspberry pi 1:
```
pi@raspberrypi:~/domoticz_speed_test/www $ time find . \( -name '*.js' -o -name '*.html' \) -exec gzip --force --keep {} \;

real    0m3.520s
user    0m1.760s
sys     0m1.220s
```
This is no big deal timewise, and for sure no reason to believe it _will make Domoticz appear very slow for that first user hitting it_ :)

Size wise, 2MB is gained:
```
pi@raspberrypi:~/domoticz_size_test_gz $ tar xzf ../domoticz_speed_test/domoticz_linux_armv7l.tgz
pi@raspberrypi:~/domoticz_size_test_gz $ du -h --summarize www
13M     www
pi@raspberrypi:~/domoticz_size_test_gz $ find . \( -name '*.js' -o -name '*.html' \) -exec gzip {} \;
pi@raspberrypi:~/domoticz_size_test_gz $ du -h --summarize www
11M     www
```

I do however agree this would have made equally much sense as a build step, but @gizmocuz inisted for it to be done in runtime..

gordonb3:
Come to think a little more about it, if you are to serve the compressed files only there's really no need to keep the uncompressed versions in the live environment as well. Just serve the one that is there: compressed or uncompressed if the user wanted to make changes to it.

gordonb3:
@emontnemery,
Funny. That's even considerably slower than on my armv5te machine. I get
```
real    0m1.980s
user    0m0.440s
sys     0m0.240s
```

In any case it still adds two to four seconds to the regular load time and it doesn't seem likely that you'll ever see that returned in overall performance because it usually never takes that long to load the page.

emontnemery:
@gordonb3 You're describing what is already implemented...
The reason for this PR is to have uncompressed files in git without increasing the bandwidth consumed for a full refresh if the domoticz UI.

emontnemery:
@gordonb3 Good to hear yours is faster than mine..

_In any case it still adds two to four seconds to the regular load time_
No, it does not. It adds two to four seconds to the very initial load time by the first users first connected device after first installing domoticz.
Then it will add fractions of a second, again to only to the first users first connected device, each time any javascript or html is modified and the gzipped file is thus invalidated.

Edit:
BTW, please review the new commit..

gordonb3:
> You're describing what is already implemented...

No, I'm not. And it is also not a competition of whose is bigger. Or faster, whatever.

What I'm saying is that if it is really the intention to serve compressed files, which is truly only about bandwidth and has nothing to do with less CPU power,  we should gzip them during the build process. It makes no sense whatsoever to have both the source and the compressed version available in the live environment.

I'm also not that impressed by the compression ratio. 2 megabytes - that's 16 megabits. If that's consecutive data that will transfer through my network in 16ms. If I were to move a chess piece in that time and then move it back in another 16ms you wouldn't even be able to see that it ever moved.

Obviously with a Pi transfer speed will be somewhat lower due to the fact that it is limited by USB speed (540Mb), but that also applies if WiFi is involved. Either way, not counting whatever was conversed on this subject now and before, we're talking about using 3 seconds of system time to save 30ms in transfer time. So you'll need more than a 100 clients to get a net profit out of this.

emontnemery:
Yes, of course it makes sense to do it during build.
But please convince @gizmocuz

And again, the reason for the PR is that @gizmocuz listed compression of javascript as a requirement for storing uncompressed javascript in git.

gizmocuz:
Okey again, a web server normally these days serve compressed files, hence , that is why there has been support for this in the HTTP protocol.
It is really not strange that the domoticz internal server also does the same.
Yes it saves CPU if you serve already compressed files, did you try running this on a Pi Orange, or RPi 1?
Thats how i started development.
These days with a RPI3, it does not make much difference, and also memory wise. Still i hear people complaining if code is written that consumes to much memory.... (hint)
And yes, if the caching was working correct in ALL browsers, then it is requested only once.
But since there are 100 different browsers, it seems some are not working correctly, or i don't know what users do these days, switching back/forth between beta/stable, restoring folders...
To solve the appcache issue, maybe instead of the version number, we replace this (with a tag as we already do on some places) it with the build hash.
This will force a complete refresh for each new version

gordonb3:
Actually, most people complain about speed. There is a relation though, because once you start swapping that will be the end of any performance you can get out of a machine.

Also, according to the [Nginx documentation pages on this subject](https://docs.nginx.com/nginx/admin-guide/web-server/compression/) compression is not always on. You need to enable it and there appear to be quite some considerations whether you should or should not as well. Of course browsers that can handle javascript and CSS will no doubt also support gzipped content, but from reading that link you should probably stay away from wanting to compress json responses as well.

In automation essentially everything is a trade-off. There is always some limit or bottleneck that influences overall performance. On the internet that is the network with at least one very obvious bottleneck being that thing many of us still call a modem, limiting our bandwidth to a practical maximum of 100Mb but often much lower than that. Not that you would really notice the difference on anything but streaming or transferring really big files, because the real limit is about timings: latency. Just ping some machines and you'll see what I mean:
- pinging my Domoticz server in my local network returns in ~0.3 ms
- pinging a server located in my own country returns in ~4ms
- pinging a server on the other side of the Atlantic returns in ~106ms

Just do the math: how many objects (js, css, html, images) make up a single page? Or just stick to the files we store on the client side, currently at 613 and counting up to 5.5MB (=44 megabits) according to my browser. Assuming a 5Mb upload and ignoring parallel fetching (which will only reduce latency)):
- local net: 44Mb @ 1Gb/s + 613 x 0.3ms = 228ms
- away from home: 44Mb @ 5Mb/s + 613 x 4ms = 11.252s
- other side of the world: 44Mb @ 5Mb/s + 613 * 106ms = 73.778s

So if we can save 2MB by doing compression that will shave about 3 seconds from the total time with this upload speed. Only I have 50Mb, so that gain is only 300ms for me and because the compression itself requires close to 2 seconds that actually makes it slower.

For clarification: I'm not against compression. There has to be a point to it though and as far as I can see there isn't. It doesn't solve anything, it's mostly about an idea that really doesn't work that way in practice.

emontnemery:
I think the discussion has derailed a bit, sorry about that..

Trying to summarize the discussion:
- There is a small time penalty (a few seconds) for the very first user who connects to a new install of domoticz
- Some bandwidth is saved by compressing
- Compression could potentially be done as a build step instead
- As @gizmocuz has pointed out, other webservers support automatic compression (https://httpd.apache.org/docs/2.4/mod/mod_deflate.html, https://docs.nginx.com/nginx/admin-guide/web-server/compression/)
  Note the nginx documentation, I think it summarizes the discussion in this PR quite well :)
  > Compressing responses often significantly reduces the size of transmitted data. However, since compression happens at runtime it can also add considerable processing overhead which can negatively affect performance. NGINX performs compression before sending responses to clients, but does not “double compress” responses that are already compressed (for example, by a proxied server).

@gizmocuz You suggest that:
_at startup, check which files do not match anymore with the compressed files and re-compress them_
I really don't want to implement such a solution, that would be better done as a build step.
Can this PR please be accepted without that change?

gordonb3:
Yes, it would have been cleaner to keep this discussion in #2254 but for some reason the responses concentrated on this PR and my post in #2254 which explained my original thought better was ignored.

In any case, while I could probably live with the execution as done by @emontnemery I really do think the originating idea is a very bad one and you should expect users to complain about Domoticz having become extremely slow after this. I don't think adding `a few` seconds to something that normally loads in less than half a second can be considered a `small time penalty`. The fact that `only the first user` gets confronted with it also seems rather meaningless for users that don't normally use more than one client, which I assume to be many.

gizmocuz:
@emontnemery, i need some time to review this PR good. I am not actually fund with a solution that checks at startup if something needs to be changed. I rather have this real-time as this PR is.
PR looks good. We could add a commandline option to stop compressed responses (gzipoff)
I still want to have a peek before at NGINX as it does what i am looking for:

> "Compressing responses often significantly reduces the size of transmitted data. However, since compression happens at runtime it can also add considerable processing overhead which can negatively affect performance. NGINX performs compression before sending responses to clients, but does not “double compress” responses that are already compressed "

I don't think they will do this with 'stat' all the time, but i might be wrong

@gordonb3 , remember that a first time ever user... which has no cached compressed files, need a small moment for this to build. Existing users 'might' have a tiny delay for that 1 or 2 files that needs to be re-compressed. a simple 'ls' command on the prompt is fast right ? so why should that be slow in Domoticz ?

gordonb3:
You're assuming that people are running Domoticz from the build directory, or use an install routine that does not touch dynamically created files in "protected" directories and also does not change the timestamp on what are essentially to be considered source files in the www folder.

1. I don't think you should include source files in a distribution
2. I'm questioning the original idea as this appears to be about wanting to save CPU cycles where this is clearly about saving bandwidth - at the cost of CPU cycles
3. Beware of user experience. The chance that even a single user is using the application you developed the way you expect them to is practically zero. If you make things better, they will see new faults. If you add three seconds to the loading time a user will not appreciate how that improves the maintainability of the application.

gizmocuz:
@gordonb3, i think we talk about different issues/solutions

1. A web server should be able to serve compressed files (this could be disabled by an option)
2. This is done in real-time, not at the startup of an application
3. source files needs to be distributed, AND this makes it a lot easier for people working on the HTML part
4. That people install it 'their' way does not mean 'we' have to comply with that.
5. A 'cache' folder to store compressed files based on a fingerprint system is not a bad idea

gizmocuz:
@emontnemery, i will look into this solution today. I might change the cache folders name (default to www_cache), and make this configurable as well (maybe some want to place them in /tmp/domoticz_cache)

dannybloe:
As I mentioned in another thread, storing files with a build-hash or fingerprint is not sufficient for proper caching. That only benefits the js files but not all the other assets like images and fonts. Unless you hash these assets as well but then you need a build system that will hash the references in referencing files as well like html and css. Webpack does have loaders for this that will rewrite url() and src attributes using build hashes and surely there are other build systems supporting this as well.

gizmocuz:
@dannybloe, as you might see in the code, this is only done for JS/HTML files

dannybloe:
I know but assets are also a big part of the caching problem.

emontnemery:
@gizmocuz Please don't merge this.
I think the point raised by @gordonb3 about the user not behaving as we assume is a very good one: For example if the user is moving back to an earlier beta, to work around some bug or whatever, this solution will fail beacuse the gzip cache will not be invalidated. Hence the user will be served an undefined mishmash of compressed javascript and html from, most likely, incompatible versions.
Instead, we should move towards using a proper build system also for the Web UI as proposed by @dannybloe.

gizmocuz:
I am not going to compress files at build time, that is not a good idea at all.
Also, users SHOULD NOT go to a previous version, that is STUPID and DANGEROUS !!!!!!
There are database patches,migrations, going back IS NOT EVER AN OPTION !
(As said 1000 times *g*)

dannybloe:
Would be nice if you can't start domoticz at all if the db version is not compatible with the application version. So if the database version is too high (because user downgraded domoticz after an upgrade has been applied to the database) domoticz just throws the middle-finger.

emontnemery:
Then at least this PR has to be mended to not just look at larger mtime as a condition for invalidating a cache entry.
A very simple change would be to set mtime of the compressed file to that of the uncompressed file, then the condition for recompressing can be to different mtime instead of larger mtime.
@dannybloe Sure, but db version is not stepped with every change to web ui..

gizmocuz:
So to summarize:

1. A web server should be able to serve compressed files (gzip) as this is normal HTTP protocol
2. The whole point of this discussion is that we will unpack important files that are currently compressed in the repository that users should be able to edit
3. We could add a cache folder
4. We should have a 'good' fingerprint mechanism that knows when a file has been changed to recreate the cache

@emontnemery , i know, that i why i did not merge it yet, and i am going to look at the webserver part today/tomorrow
No the db version is only incremented by code (by someone needed to do this)

@dannybloe , yes, that could be a good option, going to submit a patch for that first (if the internal database version is higher then the database version of domoticz)

gordonb3:
> @gordonb3, i think we talk about different issues/solutions

Obviously, yes.
1. Please reread the comment in the Nginx admin-guide on this. Also note that compression is an optional feature that at least in Nginx is off by default and there is no rule that states it must be enabled.
2. There is a difference between a multi-purpose web server that is designed to serve random content that has to be added by the user and a service or application that generates content based on what the programmer fed into it.
3. That's not what I meant when I said `distribution`. Does it help if I replace that with `runtime environment`?
4. Unsure what you're trying to say here. I just run `make install`, which places all the files needed for running Domoticz in the target folder. I do not have any *.c, *.cpp or *.h files in that target directory. I fail to see why you would want to treat source files with a different extension differently and do some type of post build processing on those source files during runtime.
5. As indicated by @emontnemery: there are quite some messages on both the forum and here on GitHub where users state they are reverting betas that have issues. Simple timestamp checking whether a compressed file is newer than the associated source .js file will fail in these cases.

gizmocuz:
@gordonb3,

1. as i mentioned
2. ?
3. yes that helps as these are two major different ways
4. i update from the web browsers gui, but it should make no difference
5. as mentioned, and that should partly be handled now, and they should not do that

gordonb3:
@dannybloe Client side cache behaviour is a somewhat different issue. The source of that problem is however most likely related to similar timestamp matching giving the wrong result and thus the wrong version ending up in appcache

gizmocuz:
If we cannot trust the 'stat' output of a file (timestamp), then everything will fail of course
Yes, assuming the date/time from the end-users system is correct.

Assuming caching would/should work as intended, a 'non-compressed' file, that could be returned gzipped, should only have to be compressed once to a users browser.

I do not consider this to be a major CPU issue / show stopper

gordonb3:
@gizmocuz 

1. okay, so we agree on the point that it should be optional
2. Nginx developers can't know what content should be compressed and so they can't. We do and so we can
3. just stating: user point of view versus developer point of view
4. Never worked for me. Also compiling on the box itself probably takes about 10 hours now. I use a crossdev machine to build Domoticz
5. well, yeah... The fact of course is that they do. I've even had to do it myself at one time because my machine went into stall every 4-5 days. Up until about two months ago I was still on 3.8000 (with some custom patches) as this was the last stable version for me.

gizmocuz:
So... With the latest commit, all needed domoticz javascript files are uncompressed.
Every other library in the www/js folder is still compressed and i see no reason to decompress it
If a user wants to make changes in the angularjs.min.js file ? (ugh) then they need to unpack it, remove the compressed version. (And i would use a private browser)

So this whole timestamping/gzip thing is only for those 4 files now.
Maybe we can drop this PR....

Assuming the cache mechanism is working correctly (we do provide timestamps back to the browser), and those 4 files have to be served again because of a refresh, i would not consider this much CPU intensive, and should only have to be done once per browser

gordonb3:
>If we cannot trust the 'stat' output of a file (timestamp), then everything will fail of course
Yes, assuming the date/time from the end-users system is correct.

It's not that you can't trust the stat. The problem is that you can only draw a clear conclusion from it if the source is newer than the target object. If it's the other way around you can only assume that the target was created from that particular version of the source. You'll need something like a manifest to be able to uniquely bind the two files.

gordonb3:
> Every other library in the www/js folder is still compressed and i see no reason to decompress it

Absolutely. Files from external sources should be left the way they are. Keeping uncompressed versions of our own .js and .html files in the developer tree allows for better tracking of changes which are currently obscured by the binary format due to compression. Never really understood why these files had to be compressed in the first place.

gizmocuz:
@gordonb3 ,

Hmmm at the moment we cannot support non-gzip supported browsers (thankfully there are none)
This because of all the .gz files in our repo... 
If you really want to have support for non-gzip then this is going to be some work... but possible, but still i do not see any reason for so...
We definitely benefit extreme much bandwidth wise

For you other comment, how is it possible to have an older file in the www folder then belongs with the current domoticz version ?
Mind you ... this can currently ONLY go wrong for one of the library files !
This because .js , .htm(l) and .css files are realtime compressed and are not cached
That being said, so, if... a user replaces one of the libraries with an older version, then it could go wrong.
But, if we had a issue with a library, there will be a pull-request/commit to solve this, and the file will again have a newer creation date

gordonb3:
I don't really care about whether the files are compressed or not. I just dislike the idea of compressing them on the fly because the CPU cost for that is much higher than what you save in bandwidth. At least on a Pi - and my own ARM based system.

Why can files in www be older than the current Domoticz version? Well, obviously one can copy only the Domoticz executable into the target directory, but that wasn't what I was referring to. The problem would be with the static gzip `cache` files being created during runtime receiving current time. If I next install a version of Domoticz that is older than today (but may still be a newer version than I ran before) the `cache` files will be newer than the source files they are supposed to relate to and thus not refreshed. Clearly such an issue would not exist if generating the gzipped versions was part of the install routine.

emontnemery:
Regarding older files, a trivial correction is to set mtime of gzipped file to same as the uncompressed file and then recompress if mtime differs instead of if mtime is newer.

gordonb3:
Does that work with every platform?

gizmocuz:
> I don't really care about whether the files are compressed or not. I just dislike the idea of compressing them on the fly because the CPU cost for that is much higher than what you save in bandwidth. At least on a Pi - and my own ARM based system.

That's not correct, we save huge amounts of bandwidth. As said, those few files that are real-time being compressed and should be cached, would not be requested anymore!

> Why can files in www be older than the current Domoticz version? Well, obviously one can copy only the Domoticz executable into the target directory, but that wasn't what I was referring to.

No, that still leaves the files with their creation/modified dates

> The problem would be with the static gzip cache files being created during run-time receiving current time.

That was something that was not going to happen/coded.
Besides, this was mainly a problem for library files

@emontnemery, this PR is now only related to library files and those can stay compressed as they are not changed frequently, it's it is done by one of the team members.
Therefor i will close this PR.

It was good to have all this discussion ! Thanks !
Possibly there might be a caching problem which should be discussed/solved with another issue.

gordonb3:
Just investigating the code after a commit by @gizmocuz that didn't seem to make sense. Turns out that this discussion was rather pointless because all code is already compressed on the fly in cWebEm, regardless of whether the client informs that is capable of handling it.

Another thing I found is that you can let an installer compress all the web content and Domoticz will never look for its uncompressed version. Except for index.html, because Domoticz never tries to find its .gz version.

gizmocuz:
My commits always make sense ;) Yeah that was what i was saying... Did not start this discussion

PR #2396, opened 2018-05-17
Aqara Cube

1981nick1982:
The previous adjustment for integrating the new cube did not work as expected. I have now put the model in brackets.


gizmocuz:
Thanks!

PR #2399, opened 2018-05-20
Aqara cube

1981nick1982:
The following information I found on this page http://docs.opencloud.aqara.cn/development/gateway-LAN-communication/
1 of the selector properties has changed from clock_wise | anti_clock_wise to rotate where rotate happens in the form of degrees. I have created a separate line for the new cube model. I do not know if this will solve the problem but we can try it. I doubt whether I should use the command rotate or Rotate_degree.

Cube Sensor
(Device type model: sensor_cube.aqgl01)
Attributes	Instructions
Cube_status	Flip90/flip180/move/tap_twice/shake_air/swing/alert/free_fall/rotate (Turn 90 degrees/Flip 180 degrees/Pan/Double click/Shake/Hold/Stay for a while)
Rotate_degree	The angle of rotation, in degrees (°), is a positive number, indicating clockwise rotation, and negative number counterclockwise rotation.
Detect_time	Rotation sampling length in milliseconds (ms)
Battery_voltage	Button battery voltage value, unit mv, range 0~3300mv, under normal circumstances, less than 2800mv indicates low battery.


gizmocuz:
Thanks!

PR #2400, opened 2018-05-21
Floorplans now stored in database

gizmocuz:


gizmocuz:
All is expected to work fine, but to test make a backup of
- database
- www/images/floorplans

At start, all images found in www/images/floorplans are imported into the database, and the floorplan table is slightly modified.

gizmocuz:
@dnpwwo , as (original) author of the floorplan, is it possible you can test this PR ?

dwmw2:
Purely from code inspection, that looks like it's going to add the `example.jpg` floorplan back to the database when I update. I think I already deleted that once, when I added my real ones.

Should we iterate over the existing records in the `Floorplans` table and import those images, rather than importing *all* files (not just image files, it seems) that we find under `images/floorplans`? I had some old versions of my floorplans lying around too; I don't really want *those* added either... :)

I suspect that's going to fail on Windows too, as it wants to be `images\floorplans` instead? Perhaps we should have a generic function for this, btw, since we have a lot of `#ifdef WIN32` for precisely that `/` vs. `\` difference. A helper function which does the same as [`g_build_filename()`](https://developer.gnome.org/glib/stable/glib-Miscellaneous-Utility-Functions.html#g-build-filename) would probably be useful here.

gizmocuz:
1. Yes it does that, and when you had no floorplans before, you get an example floorplan, which you can always delete if you want

2. no, because maybe you still want those images.
You can always delete them later... not much work
And some cleaning in the actual folder won't hurt too ;), you can now also see if the floorplan is actually used in the first table, so if it is not, feel free to delete it

3. no, already tested

dwmw2:
My floorplans are PNG files, and this seems to be sending them to the browser with `Content-Type: image/jpeg`. It doesn't seem to matter but should probably be fixed.

It does the same for the text file that happened to be lying around in the `images/floorplans` directory too. That one *does* cause strange behaviour in the browser, as it causes all the images to show up really tiny.

```
> GET /images/floorplans/plan?idx=3 HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.59.0
> Accept: */*
> Cookie: SID=…
> 
< HTTP/1.1 200 OK
< Content-Disposition: attachment; filename=floorplan.jpg
< Content-Length: 10
< Content-Type: image/jpeg
< Cache-Control: max-age=3600, public
< 
Some text
```
Perhaps we could add only *image* files? And only add the example if there aren't any *other* floorplans already?

gizmocuz:
Well looking at the text file issue. The image is no problem/issue at all as is it loaded in IMG object that knows how to handle it
For the rest it is working correctly I hope?

gizmocuz:
@dwmw2 , made some changes to address the two comments. Hope all is ok now ?

dwmw2:
The blockly one is a non-sequitur but the "Only scan for images…" bit looks good; thanks. One final minor suggestion might be to name the floor after the filename. Otherwise if you have a bunch of images lying around they all end up called 'Example'. But in general, all looks good to me. Thanks.

gizmocuz:
With the current patch the floorplans will be named correctly as well.
Question, did you tested the patch also ?

dwmw2:
Yes, I tested it and am now running it (except the latest change) on my production instance. All good; thanks!

gizmocuz:
Thanks for the feedback !

PR #2405, opened 2018-05-22
Allow libusb to be disabled

dwmw2:
Sometimes, even if libusb is actually present in the build system we
don't want to link against it. For OpenWRT builds, for example, it does
exist and some optional packages *do* use it, but we don't want to
unconditionally make it a dependency for the Domoticz build.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -596,20 +596,23 @@ ENDIF(CURL_FOUND)
 #
 # USB
 #
-find_path(LIBUSB_INCLUDE_DIR usb.h
-   HINTS ${PC_LIBUSB_INCLUDEDIR} ${PC_LIBUSB_INCLUDE_DIRS})
-find_library(LIBUSB_LIBRARY NAMES usb
-   HINTS ${PC_LIBUSB_LIBDIR} ${PC_LIBUSB_LIBRARY_DIRS})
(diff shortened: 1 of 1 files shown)


dwmw2:
Again without the whitespace:
```
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -596,6 +596,8 @@ ENDIF(CURL_FOUND)
 #
 # USB
 #
+option(WITH_LIBUSB "Enable libusb support" YES)
+if(WITH_LIBUSB)
   find_path(LIBUSB_INCLUDE_DIR usb.h
      HINTS ${PC_LIBUSB_INCLUDEDIR} ${PC_LIBUSB_INCLUDE_DIRS})
   find_library(LIBUSB_LIBRARY NAMES usb
@@ -610,6 +612,7 @@ IF(LIBUSB_FOUND)
   else()
     MESSAGE(STATUS "==== LibUSB not found, support for TE923/Voltcraft disabled!")
   ENDIF(LIBUSB_FOUND)
+endif(WITH_LIBUSB)
 
 ## support lua popen on Linux platforms
 #IF(${CMAKE_SYSTEM_NAME} MATCHES "Linux")

```

gizmocuz:
Thanks!

PR #2406, opened 2018-05-22
Fix precompiled header with -Wl,… in compile flags

dwmw2:
Arguably, GCC ought to ignore stuff like -Wl,-z,relro when it's building
a precompiled header. It doesn't, so strip it out.

Some platforms do enable relro by default, and this causes a build
failure otherwise.


gizmocuz:
Thanks!

PR #2407, opened 2018-05-22
webserver: automatically reload SSL certificates and DH params

dwmw2:
When something like LetsEncrypt is used to issue certs, they rotate
fairly frequently. Instead of having to restart Domoticz to pick up
the new cert, just reload the cert each time.

### DIFF
--- a/webserver/server.cpp
+++ b/webserver/server.cpp
@@ -188,20 +188,37 @@ void ssl_server::init_connection() {
 	char cipher_list[] = "ECDH+AESGCM:DH+AESGCM:ECDH+AES256:DH+AES256:ECDH+AES128:DH+AES:RSA+AESGCM:RSA+AES:!aNULL:!MD5:!DSS";
 	SSL_CTX_set_cipher_list(context_.native_handle(), cipher_list);
 
+	struct stat st;
 	if (settings_.certificate_chain_file_path.empty()) {
 		_log.Log(LOG_ERROR, "[web:%s] missing SSL certificate chain file parameter !", settings_.listening_port.c_str());
-	} else {
+	} else if (!stat(settings_.certificate_chain_file_path.c_str(), &st)) {
+		cert_chain_tm_ = st.st_mtime;
 		context_.use_certificate_chain_file(settings_.certificate_chain_file_path);
+	} else {
+		_log.Log(LOG_ERROR, "[web:%s] missing SSL certificate chain file %s!", settings_.listening_port.c_str(), settings_.certificate_chain_file_path.c_str());
 	}
+
 	if (settings_.cert_file_path.empty()) {
 		_log.Log(LOG_ERROR, "[web:%s] missing SSL certificate file parameter !", settings_.listening_port.c_str());
-	} else {
+	} else if (!stat(settings_.cert_file_path.c_str(), &st)) {
+		cert_tm_ = st.st_mtime;
 		context_.use_certificate_file(settings_.cert_file_path, boost::asio::ssl::context::pem);
+	} else {
+		_log.Log(LOG_ERROR, "[web:%s] missing SSL certificate file %s!", settings_.listening_port.c_str(), settings_.cert_file_path.c_str());
 	}
+
+
 	if (settings_.private_key_file_path.empty()) {
 		_log.Log(LOG_ERROR, "[web:%s] missing SSL private key file parameter !", settings_.listening_port.c_str());
-	} else {
+	} else if (!stat(settings_.private_key_file_path.c_str(), &st)) {
+		// We don't actually bother to track the mtime of the private
+		// key file as it can't sanely change without changing the
+		// certificate too. And may in fact change *before* the
+		// certificate does, while the cert is being issued. We
+		// don't want to update until the *cert* file changes.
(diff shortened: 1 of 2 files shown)


gizmocuz:
With this patch it is reloaded (initialized) for every query to the webserver. That is a bit to much disk/cpu wise.
How about a file hook or a once/day method ?

dwmw2:
I figured it was kind of in the noise in comparison with the actual TLS setup, which has to get random numbers and do all the crypto. Just loading the cert is relatively trivial in comparison to that, and there should be no actual disk access as it'll be entirely in cache.

I can do it only if one of the files changes... but that still has to touch the disk anyway. Except again, it's going to be in cache.

Do we have a simple inotify-style trigger that can be used? It doesn't *hurt* and I'd be happy enough to do it; I just didn't see the point.

dwmw2:
Better?

gizmocuz:
There is unfortunately not a cross platform solution (Win32 has file/directory system watches)
A common way to do this is to write a script that refreshes the certificate, and sends a SIGHUP
a simple check to check a time_t variable could not hurt to see if one day (or 12 hours, or hour) have been passed, set a Boolean to reload the certificate on the next accept

gizmocuz:
Looking good, but i do not see how this reloadcerts is being called ? If you are going to call it from an external script, then a SIGHUP might be more ideal ?
I do like this reload, we might add a button in the settings page to do this ?
Is it thread safe ? I would still use a Boolean that is set to true, and on the next accept call , reload it, that should be thread safe, and it is also called almost immediately

dwmw2:
Yeah, SIGHUP would have been nice but I eschewed that because it doesn't work on Windows. Right now I use `curl` to trigger the cert reload, after renewing the cert _(actually I lie I have tested manually but my production system is still running the previous version which does it automatically)_.

I may revert to Plan A -- avoid the thread issues and  do it directly when the next connection is set up, **but** just `stat` the cert file and only really do the reload if it's changed.

dwmw2:
Here, back to Plan A. Automatically reload, which is much nicer for the admin anyway — no need to set up special hooks to "prod" Domoticz to notice the new cert. But do a simple `stat()` to check if the file is updated, and only actually reload it if it has.

gizmocuz:
Won't that query 3 stats and checks for each http request again ?
Happy to merge this if you think this would not have a performance impact...
Imagine a full cache reload, 1000 files... 3000 stats/checks hmmm lots of disk access
Whats wrong with once an hour check ?

dwmw2:
There's no actual disk access. If you have real spinning rust disks, they won't even spin up for it because it'll all be in cache. The `stat()` call is really lightweight — you'll generally find that even libc functions  like `localtime()` will `stat("/etc/localtime", …)` every time they're called just to see if it's changed. Build this and strace it, for example:
```
#include <time.h>
#include <stdio.h>

int main(int argc, char *argv[])
{
  int i;
  time_t t = 0;

  for(i=0; i<10; i++)
    localtime(&t);

  return 0;
}
```
It's about 50ns a time, after the first one (which takes a whole 60ns!). It is going to be *utterly* in the noise compared with all the entropy-gathering and crypto calculations that are needed to set up a TLS session.
```
10:03:50.917122 stat("/etc/localtime", {st_mode=S_IFREG|0644, st_size=3687, ...}) = 0
10:03:50.917182 stat("/etc/localtime", {st_mode=S_IFREG|0644, st_size=3687, ...}) = 0
10:03:50.917236 stat("/etc/localtime", {st_mode=S_IFREG|0644, st_size=3687, ...}) = 0
10:03:50.917289 stat("/etc/localtime", {st_mode=S_IFREG|0644, st_size=3687, ...}) = 0
10:03:50.917341 stat("/etc/localtime", {st_mode=S_IFREG|0644, st_size=3687, ...}) = 0
10:03:50.917394 stat("/etc/localtime", {st_mode=S_IFREG|0644, st_size=3687, ...}) = 0
10:03:50.917446 stat("/etc/localtime", {st_mode=S_IFREG|0644, st_size=3687, ...}) = 0
10:03:50.917498 stat("/etc/localtime", {st_mode=S_IFREG|0644, st_size=3687, ...}) = 0
10:03:50.917551 stat("/etc/localtime", {st_mode=S_IFREG|0644, st_size=3687, ...}) = 0
```
The reason I'm not keen on a once-an-hour check is because people *test* these things when they're being deployed. And _"renew your cert and then wait an hour and come back to see if Domoticz has noticed"_ is not a great user experience for that :)

dwmw2:
Note it's also for each new connection, *not* for each request. If we aren't doing HTTP keepalive and we're forcing people to do a whole new TLS session for every request, we have much more important performance concerns to look at... :)

gizmocuz:
Thanks for the patch and explanation! Interesting about localtime, hopefully this is only under linux/unix ;)

dwmw2:
FWIW I get a whole *millisecond* back (as well as reducing the executable size by about a quarter(!)) if I build with `-flto`. A primitive benchmark of `time curl -sk 'https://[::1]:8443/json.htm?type=command&param=getversion'` takes mostly 0.018s with `-flto` and 0.019s without. Although it's quite noisy so it's hard to be sure :)

Of course the additional time to do the final link takes up a lot of those milliseconds...

PR #2408, opened 2018-05-23
MochadTCP/P1MeterBase: fix C++ One Definition Rule violations

dwmw2:
Fix the following build warnings seen with GCC 8.
Untested as I have neither hardware, but Obviously Correct™.

```
hardware/MochadTCP.cpp:20:3: warning: type ‘MatchType’ violates the C++ One Definition Rule [-Wodr]
 } MatchType;
   ^
hardware/P1MeterBase.cpp:19:3: note: an enum with different value name is defined in another translation unit
 } MatchType;
   ^
hardware/MochadTCP.cpp:29:16: warning: type ‘struct _tMatch’ violates the C++ One Definition Rule [-Wodr]
 typedef struct _tMatch {
                ^
hardware/P1MeterBase.cpp:59:16: note: a different type is defined in another translation unit
 typedef struct _tMatch {
                ^
hardware/MochadTCP.cpp:30:12: note: the first difference of corresponding definitions is field ‘matchtype’
  MatchType matchtype;
            ^
hardware/P1MeterBase.cpp:60:12: note: a field of same name but different type is defined in another translation unit
  MatchType matchtype;
            ^
hardware/MochadTCP.cpp:20:3: note: type ‘MatchType’ itself violates the C++ One Definition Rule
 } MatchType;
   ^
hardware/P1MeterBase.cpp:19:3: note: the incompatible type is defined here
 } MatchType;
   ^
```

### DIFF
--- a/hardware/MochadTCP.cpp
+++ b/hardware/MochadTCP.cpp
@@ -11,30 +11,30 @@
 
 #define RETRY_DELAY 30
 
-typedef enum { 
+enum _eMochadMatchType {
 	ID=0, 
 	STD, 
 	LINE17, 
 	LINE18, 
 	EXCLMARK 
-} MatchType;
+};
 
-typedef enum {
+enum _eMochadType {
 	MOCHAD_STATUS=0,
 	MOCHAD_UNIT,
 	MOCHAD_ACTION,
 	MOCHAD_RFSEC
-} MochadType;
+};
 
-typedef struct _tMatch {
-	MatchType matchtype;
-	MochadType type;
+typedef struct {
+	_eMochadMatchType matchtype;
+	_eMochadType type;
 	const char* key;
 	int start;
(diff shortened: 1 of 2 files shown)


gizmocuz:
Thanks! You're on a roll ! While you are hear, could you make it more compliant with the current code ?
start enums with _e and no need for the typedefs

dwmw2:
Comme ça?

gizmocuz:
Thanks!
I don't think you have to declare it in a structure like this ?

enum _eMochadMatchType matchtype

_eMochadMatchType matchtype

Would be ok ?

dwmw2:
Er... really? _<tests…>_ Yes, that actually does work in C++ just not C.
Your turn to teach me something today :)

gizmocuz:
Thanks!

Issue #2409, opened 2018-05-23
Mismatch mode for Daikin AC


pipiche38:
When referring to this https://github.com/ael-code/daikin-control and also while checking on my system ( Domoticz beta + Daiking Emura AC) , I have a miss-match of Mode

DEHUMIDIFICATOR gives: "ret=OK,pow=0,mode=3,....." instead of mode=2
HEAT gives : "ret=OK,pow=0,mode=6,...." instead of mode 4
COLD gives: "ret=OK,pow=0,mode=4,... "instead of mode 3
FAN gives: "ret=OK,pow=0,mode=4...." instead of mode 3
AUTO gives: "ret=OK,pow=0,mode=2,..." instead of mode 0

I used the following URL to check the controlinfo:
http://10.0.0.33/aircon/get_control_info

Version of AIRCON Interface: ver=1_2_48

pipiche38:
While putting domoticz in Debug , here is what I get:

Selecting AUTO Mode:
May 23 17:36:40 pi3 domoticz[11137]: 2018-05-23 17:36:40.412  User: support initiated a switch command (92/Mode/Set Level)
May 23 17:36:40 pi3 domoticz[11137]: 2018-05-23 17:36:40.414  Daikin: Worker 10.0.0.33, Write to Hardware...
May 23 17:36:40 pi3 domoticz[11137]: 2018-05-23 17:36:40.414  Daikin: Worker 10.0.0.33, Set General Switch ID 5, command : 1, value : 10
May 23 17:36:40 pi3 domoticz[11137]: 2018-05-23 17:36:40.414  Daikin: Mode ...
May 23 17:36:40 pi3 domoticz[11137]: 2018-05-23 17:36:40.538  Error: Daikin: Invalid response


Selecting DEHUMIDIFICATOR MODE
May 23 17:37:12 pi3 domoticz[11137]: 2018-05-23 17:37:12.969  User: support initiated a switch command (92/Mode/Set Level)
May 23 17:37:12 pi3 domoticz[11137]: 2018-05-23 17:37:12.971  Daikin: Worker 10.0.0.33, Write to Hardware...
May 23 17:37:12 pi3 domoticz[11137]: 2018-05-23 17:37:12.971  Daikin: Worker 10.0.0.33, Set General Switch ID 5, command : 1, value : 20
May 23 17:37:12 pi3 domoticz[11137]: 2018-05-23 17:37:12.971  Daikin: Mode ...
[...]

PR #2411, opened 2018-05-23
Fixing #2409

pipiche38:
The Mode setting was not correct and not aligned with the Switch/Selector created. Was also missing the Fan mode.


pipiche38:
The Mode setting was not correct and not aligned with the Switch/Selector created. Was also missing the Fan mode.
This fix #2409

gizmocuz:
Thanks!

PR #2414, opened 2018-05-24
Add .gif file format to floorplan

sweetpants:
.gif was missing and users where complaining


gizmocuz:
Thanks, did not know we supported gif ? There is another place where we need to add support for this, it is where we return the image.. i will make a commit for this in a few minutes

done..

https://github.com/domoticz/domoticz/commit/a74f7c4a5e7e03d9ff9b9fde0ce07c77a8ccdd8d

PR #2416, opened 2018-05-24
Compression and cache behaviour 

gordonb3:
Following up on earlier discussions

I've added a command line switch that allows three modes of compression behaviour:
- `on` = always compress unless browser doesn't support it [default] - same as current
- `off` = serve uncompressed files only - decompresses gz files
- `static` = serve compressed if available but do not compress on the fly

Cache behaviour is troubled by the fact that the browser maintains a double cache when we use appcache. The second problem is that our web server sends incorrect time stamps for modification time and an expiration date on images that is not accepted by at least Firefox (on Linux). Third is that our web server tries to be nice and checks the original files against the bad (newer) timestamps that we sent to the browser ourselves to return a HTTP-304.

I've corrected the timestamps and added an exception for theme files which should always be renewed from the server when a client requests them. This allows users to switch themes without needing to flush both caches.


gizmocuz:
Thanks for the PR. Could you make an "enum" for the mode ?
For me the description in the "help" is not very clear.
Currently we have .gz files and normal files, and normal files will be gzippid, the gz is send passthrough
So that will be "static" mode ? Shall this not be default then ?

Maybe with the "incorrect time stamps" fix, it will work a lot better already ?

gizmocuz:
I might also be better to just have to options, YES/NO , when no, decompress the .gz

gordonb3:
Yeah, well... I intended to mimic the way Nginx supports compression with a single command line switch. The name `static` was also taken from Nginx, meaning that it will only serve compressed files if they are already compressed in the webroot.

The `off` setting actually started more as a play thing, just to see what result that would have in my browser and the network. I just left it in there because it might have a practical use to someone, e.g. if the user runs both client and server in embedded Windows on an atom and constantly referencing something that needs to be decompressed only makes it heavier on the CPU. And of course compressing and decompressing doesn't make much sense either in such an environment.

Internally the values translate as -1, 0 and 1 which allows easy testing by only referencing the value 0. I guess I could export these to the command line. I'm kind of a fan for stuff that is readable myself though.

The cache thing is somewhat unrelated actually. I just happened to stumble over it while working on the compression issue and not directly seeing the responses I expected to see. I'm not saying this will solve all the caching issues users have reported though. To do that you would have to reference the appcache file and simply ignore the cache verification request that a browser does on every file that is in there. I'd say that is probably counterproductive, but there may be more files than the theme specific files that you might want to add to the ignore list.

gizmocuz:
Maybe it is better to make a separate PR for the cache problem fixing then.
I like enum's, they are much easier to understand then -1, 0,1,2,...
If OFF is not used, and probably static not, the default behavior will stay the same with this PR except the caching might work better ?
Two separate patches would be great then, as i am not to fond about implementing the on/off/static command line option yet
Those people that will run on an Atom etc... for 30 bucks you have a good raspberry pi

dwmw2:
FWIW as I read this I was planning to use the `static` option for the OpenWRT builds. Those boxes are built for network not CPU so doing the compression dynamically would rarely be a win. 

I *will* probably make sure we compress everything at install time though, even though I don't think we have that in the upstream makefiles yet? Which I suppose means it ends up being equivalent to the `yes` option in the end?

gizmocuz:
@dwmw2 , there is no need to compress anything else now and there is also no need to actually
You might confuse your users as there is a file mismatch with the original repo
Also they need to be able to edit css/html/js scripts if they want to make changes

gordonb3:
Just mentioned Atom as it seems more straightforward to run embedded on Windows. Obviously if you attach a screen to a Pi and run X on it the same applies: with the absence of bandwidth to consider when running server and client on the same machine compressing and decompressing only costs CPU time.

Also not a big fan of a Pi myself. I mean: it's a nice thing to play with, but running the OS from an SD card is really not something one should want to do with a machine that is supposed to be up 24/7. Of course you can also use the SD card to start a pre-boot environment and then run the actual OS from a USB attached disk, but that means your whole system is restricted to USB speed.

In any case I'll have a look into splitting this PR in two. Or maybe three, because there's also a bit of refactoring in handle_request to allow searching for an `index.html.gz` first which does add quite some gain because precompressed files are skipped searching for `#embed` tags.

dwmw2:
We are typically looking at a read-only ROM image for the "static" files. If they want to edit, they can do that before building/installing. And as flash is short, they really *do* want to have as much as possible compressed. And would be likely to edit on a different box... so putting the .gz file onto the machine (if it does have a writeable file system at all) would be OK.

There's already a file mismatch with the original repo. The original repo has all these readable .CPP files and what gets installed is different... I think we can cope with gzip too :)

main/domoticz.cpp
```diff
@@ -170,6 +171,7 @@ bool g_bStopApplication = false;
 bool g_bUseSyslog = false;
 bool g_bRunAsDaemon = false;
 bool g_bDontCacheWWW = false;
+signed char g_wwwCompressMode = 0x1; // 0x1 = on, 0x0 = static, 0xFF = off
```
dwmw2, line 175: 0xFF doesn't fit in a signed char.

gordonb3: It works here. 0xFF = -1

gordonb3: Sorry, maybe a bit short
It's still a char type, so I can go from 0 - 255. Or in hex: 0x0 - 0xFF. Defining it as `signed` makes the compiler treat everything from 0x80 - 0xFF as negative numbers. This allows me to test whether the number is less than zero, which should be more CPU friendly than testing for specific values. Or so I believe.

gizmocuz:
@gordonb3, would be great if you can split them
Ps, there is not much problem with SD Cards these days, but i also run Pi machines with iSCSI drives which is also very cool ;)
And you can of course rebuild a android tv box with an amlogic 905 chip ($32 dollar) to a nice embedded android system

gordonb3:
Hmmm... I just realized splitting may be not be possible because the resulting PRs would cause merge conflicts.

gizmocuz:
Could you close this one, and create new ones perhaps ?

PR #2417, opened 2018-05-24
Add pTypeManaged/sTypeManagedCounter devices for plugins and JSON

guillaumezin:
Mainly for counter externally managed (Linky, Gazpar) where only history can be grabbed and that will be managed by scripts

This is another implementation proposal following discussion in PR #2271

### DIFF
--- a/main/EventSystem.cpp
+++ b/main/EventSystem.cpp
@@ -941,6 +941,44 @@ void CEventSystem::GetCurrentMeasurementStates()
 						isUtility = true;
 					}
 				}
+				else if (sitem.subType == sTypeManagedCounter)
+				{
+					if (splitresults.size() > 1) {
+						float usage = static_cast<float>(atof(splitresults[1].c_str()));
+                                                
+						if (usage < 0.0) {
+							usage = 0.0;
+						}
+
+						char szTmp[100];
+						sprintf(szTmp, "%.02f", usage);
+
+						float musage = 0;
+						_eMeterType metertype = (_eMeterType)sitem.switchtype;
+						switch (metertype)
+						{
+						case MTYPE_ENERGY:
+						case MTYPE_ENERGY_GENERATED:
+							musage = usage / EnergyDivider;
+							sprintf(szTmp, "%.03f kWh", musage);
+							break;
+						case MTYPE_GAS:
+							musage = usage / GasDivider;
+							sprintf(szTmp, "%.02f m3", musage);
+							break;
+						case MTYPE_WATER:
+							musage = usage / WaterDivider;
+							sprintf(szTmp, "%.02f m3", musage);
+							break;
+						case MTYPE_COUNTER:
+							break;
+						default:
+							continue; //not handled
+						}
+						utilityval = static_cast<float>(atof(szTmp));
+						isUtility = true;
+					}
+				}
 			}
 		}
 		break;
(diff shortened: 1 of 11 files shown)


gizmocuz:
Have not looked at the patch yet 100%, but please remove the new pTypeManaged thing and use the
pTypeGeneral with a new subtype (pTypeManaged)
There are to many pType's already in the system

main/EventSystem.cpp
```diff
@@ -1061,6 +1061,13 @@ void CEventSystem::GetCurrentMeasurementStates()
 				}
 			}
 			break;
+		case pTypeManaged:
+			if (sitem.subType == sTypeManagedCounter)
+			{
+				utilityval = static_cast<float>(atof(sitem.sValue.c_str()));
```
gizmocuz: This counter only has a counter and no actual usage ? Also not in the feature ?
I say thinking about the feature, there might be someone willing to use this and also report the current usage
So a sValue like COUNTER(number);USAGE(Watt) would be a good idea?

main/SQLHelper.cpp
```diff
@@ -3587,6 +3587,9 @@ uint64_t CSQLHelper::CreateDevice(const int HardwareID, const int SensorType, co
 		case sTypeCounterIncremental:		//Counter Incremental
 			DeviceRowIdx = UpdateValue(HardwareID, ID, 1, SensorType, SensorSubType, 12, 255, 0, "0", devname);
 			break;
+		case sTypeManagedCounter:		//Counter Incremental
```
gizmocuz: I think the comment here is wrong, could be removed (also for the counter incremental if you would)

main/SQLHelper.cpp
```diff
@@ -5176,6 +5194,100 @@ void CSQLHelper::UpdateUVLog()
...
+	}
+
+	std::vector<std::string> sd = result[0];
+	std::stringstream s_strid;
+	s_strid << sd[0];
+	s_strid >> DeviceRowID;
+	std::string devname = sd[1];
+	_eSwitchType switchtype=(_eSwitchType) atoi(sd[2].c_str());
+
+	if (dayCalendar)
```
gizmocuz: this can not be the right boolean for the Meter table, the meter table is for short-log, and tables ending with _Calendar are for day values

main/SQLHelper.cpp
```diff
@@ -5176,6 +5194,100 @@ void CSQLHelper::UpdateUVLog()
 	}
 }
 
+bool CSQLHelper::UpdateCalendarMeter(const int HardwareID, const char* DeviceID, const unsigned char unit, const unsigned char devType, const unsigned char subType, bool dayCalendar, float counter, float usage, const char* date)
+{
```
gizmocuz: Maybe a small check if the 'date' field is correct, for instance it's length + dashes on the right possition

gizmocuz (commented):
Some small comments

main/SQLHelper.cpp
guillaumezin: The name of the boolean is misleading, I will change it

gizmocuz:
Thanks!

gizmocuz:
Made some small changes... i will rebuild our old code as well to use longlong as the database fields are not floats

https://github.com/domoticz/domoticz/commit/043e46ea7375381d357570d5d690c7974c9a030a

Issue #178, opened 2015-10-31
Official Debian packages for domoticz?


lnussbaum:
Hi,

I'm a happy domoticz user, and involved in Debian. I was looking into creating official domoticz packages for Debian. I could maintain backport packages as well.

Is that something you would be interested in?

I haven't looked in detail yet, but it could require some changes to the building infrastructure on the domoticz to make things easier on the packaging side. That's most likely the only thing I'll need from your side.

Please let me know what you think,

Lucas

gizmocuz:
Hello Lucas,

Sounds very good ! We also just using Travis-CI and AppVeyor.

Is this something that can be integrated with travis ?

Let us know the build changes you need to change

lnussbaum:
Hi!

Regarding travis support for building Debian packages, that's likely. I'll look at that later.

I've done some preliminary work to support building domoticz as a Debian package. My work is in
https://github.com/lnussbaum/domoticz

The main obstacle, as far as I can see, is that you currently install everything under /opt. Is there support for installing domoticz following the Linux File Hierarchy Standard, that is, executables in /usr/bin, support data in /usr/share, etc? If not, would you be willing to add such a support?

I've uploaded the result of Debian's lintian tool, that checks for common problems in Debian packages, to http://blop.info/pub/lintian-domoticz.txt. I haven't looked at all the issues listed there, and many of them can probably be fixed on the packaging side, but some of them might be worth fixing on your side as well. Anyway, the most important thing to do is to figure out a strategy to deal with the install paths problem.

Thanks

gizmocuz:
What is wrong with /opt ? everything is installed there, very easy to maintain and update right ?
via command line arguments other paths can be set... but everything in 1 folder, very easy i would say ;)

lnussbaum:
Well installing stuff using the FHS is not negociable for Debian packages. That's part of the role of Linux distributions to standardize on such things over all the software they ship.

gizmocuz:
Okey, i think this is already possible by using the command line arguments:

-wwwroot file_path (for example /opt/domoticz/www)
-dbase file_path (for example /opt/domoticz/domoticz.db)
-userdata file_path (for example /opt/domoticz)

lnussbaum:
Hi,

yes, this could work (installing in a custom location + changing the default search path in domoticz).
I'll look into that in a few weeks: I broke my arm in the meantime.

gizmocuz:
Ai... good luck with your arm... painful when typing

[...]

PR #2419, opened 2018-05-25
Update to clean minizip from zlib 1.2.11 and allow dynamic linking

dwmw2:
We had a version of contrib/minizip from zlib, that had been partly
C++-ised... and then we weren't even building unzip.cpp because it
had been concatenated to unzip.h and all the functions marked 'inline'.

Kill it with fire, and drop in a clean copy from zlib 1.2.11. Also
make it possible to build with -DUSE_BUILTIN_MINIZIP=NO to use a
system dynamic library instead.

As discussed in #178 

Tested uploading custom icons both with and without the builtin version.


dwmw2:
I probably need to do some local build testing *without* the dynamic library installed...

gizmocuz:
Thanks!!

PR #2420, opened 2018-05-25
Improve caching behaviour

gordonb3:
as discussed in [PR 2416](https://github.com/domoticz/domoticz/pull/2416)

### DIFF
--- a/webserver/cWebem.cpp
+++ b/webserver/cWebem.cpp
@@ -1893,6 +1893,7 @@ void cWebemRequestHandler::handle_request(const request& req, reply& rep)
 	session.forcelogin = false;
 	session.rememberme = false;
 
+	rep.status = reply::ok;
 	rep.bIsGZIP = false;
 
 	bool isPage = myWebem->IsPageOverride(req, rep);
@@ -1901,7 +1902,6 @@ void cWebemRequestHandler::handle_request(const request& req, reply& rep)
 	// Respond to CORS Preflight request (for JSON API)
 	if (req.method == "OPTIONS")
 	{
-		rep.status = reply::ok;
 		reply::add_header(&rep, "Content-Length", "0");
 		reply::add_header(&rep, "Content-Type", "text/plain");
 		reply::add_header(&rep, "Access-Control-Max-Age", "3600");
--- a/webserver/cWebem.cpp
+++ b/webserver/cWebem.cpp
@@ -28,6 +28,8 @@
 #define SHORT_SESSION_TIMEOUT 600 // 10 minutes
 #define LONG_SESSION_TIMEOUT (30 * 86400) // 30 days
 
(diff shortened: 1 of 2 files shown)


webserver/cWebem.cpp
```diff
@@ -28,6 +28,8 @@
 #define SHORT_SESSION_TIMEOUT 600 // 10 minutes
 #define LONG_SESSION_TIMEOUT (30 * 86400) // 30 days
 
+#define HTTP_DATE_RFC_1123 "%a, %d %b %Y %H:%M:%S %Z" // Sun, 06 Nov 1994 08:49:37 GMT
```
gizmocuz: Is this different then using the make_web_time function ? 
If it is, one of the two is correct ?

webserver/cWebem.cpp
```diff
@@ -1991,13 +1993,31 @@ void cWebemRequestHandler::handle_request(const request& req, reply& rep)
...
-	if (!myWebem->CheckForPageOverride(session, requestCopy, rep))
+	if (myWebem->CheckForPageOverride(session, requestCopy, rep))
+	{
+		if (session.reply_status != reply::ok) // forbidden
+		{
+			rep = reply::stock_reply(static_cast<reply::status_type>(session.reply_status));
+			return;
+		}
+
+		if (!rep.bIsGZIP) // this is probably always true?
```
gizmocuz: Seems like it, please remove the comment

webserver/cWebem.cpp
gordonb3: Don't really know. Seems similar. The define is copied from request_handler.cpp and it seemed to explain the hardcoded entry that was in the original cWebem.cpp on line 2101 (which happened to be an override on a header field that already exists)

gordonb3: Looks like the strftime_t function is actually the one that should be preferred as it uses the thread safe localtime_r() function where make_web_time() uses the old gmtime() function. Not sure which one actually returns the time we want and is not distorted by a TZ offset.

webserver/cWebem.cpp
gordonb3: and leave the test just in case?

gizmocuz: From what i tested with breakpoints, at this point it is not compressed
As the boolean is set after compression, i don't expect it to be compressed again
Only some additional headers will be added

gordonb3: Didn't verify it. Was just wondering if there could be some action or override that returned already compressed data. I'll leave the test in place.

gizmocuz: The test is good, but remove the comment
The test could also be added in the function 'CompressWebOutput' so it will never recompress

gordonb3:
Seems to have become messy due to parts of this already having been committed to development branch?

gordonb3:
created a new PR for the additional commit

PR #2421, opened 2018-05-25
allow manipulation of www compress behaviour

gordonb3:
Definitely something that belongs in the advanced users guide. Is
CPU more important, or is bandwidth? This switch allows the user to
make a different choice than what is considered to be sane for most
systems.


main/domoticz.cpp
```diff
@@ -171,6 +172,7 @@ bool g_bStopApplication = false;
 bool g_bUseSyslog = false;
 bool g_bRunAsDaemon = false;
 bool g_bDontCacheWWW = false;
+signed char g_wwwCompressMode = 0x1; // 0x1 = on, 0x0 = static, 0xFF = off
```
gizmocuz: could you make this an enum please

main/domoticz.cpp
```diff
@@ -884,6 +886,21 @@ int main(int argc, char**argv)
...
+		{
+			_log.Log(LOG_ERROR, "Please specify a compress mode");
+			return 1;
+		}
+		std::string szmode = cmdLine.GetSafeArgument("-wwwcompress", 0, "on");
+		if (szmode == "off")
+			g_wwwCompressMode = 0xFF;
+		else if (szmode == "static")
+			g_wwwCompressMode = 0x0;
+		// should I verify possible invalid input? doesn't seem to be done with other parameters
```
gizmocuz: Please do not add comments/questions in a PR
Always good to discuss it
No you do not as you set the default at the top of the code

main/domoticz.cpp
gizmocuz: Plz (also) make (it) a member of myWebem, and add a call to set the compression mode
that way you also do not have to work with externals in the webserver class which might be a stand-alone library later

gizmocuz (commented):
Few small remarks

main/domoticz.cpp
gordonb3: So invalid equals default? As the comment states: parameter validation doesn't seem to be done on other inputs as well. Was more of a remark to myself and didn't really think about it when entering the PR (which in this case is mostly cut-and-paste from the original PR)

main/domoticz.cpp
gordonb3: less straightforward. need to rethink the decision points.

webserver/cWebem.h
```diff
@@ -19,6 +19,12 @@ namespace http {
 			AUTH_LOGIN=0,
 			AUTH_BASIC,
 		};
+		enum _eWebCompressionMode
+		{
+			WWW_USE_GZIP=0,
```
gizmocuz, line 24: Could you make no-compression value 0 ?

gizmocuz: Is this enum public ? Then please use it everywhere (g_wwwCompressMode, in SetWebCompressionMode)

gizmocuz (commented):
Could you consider the following changes

webserver/cWebem.h
gordonb3: It is public, but I need to transfer the value from domoticz.cpp to mainworker.cpp before I can use it. This is effectively the same method used for setting authentication method and webroot.

gordonb3: There is actually no real logic to the enum. It merely seemed appropriate to use 0 for the default. If it is to be about (not) doing something special than the logical choice would be for `static` to be 0 as this is the setting that does not actively zip or unzip files.

gizmocuz: Correct, the Authentication method should also be an enum

0 = Off, or Disabled , not On or Enabled

gordonb3: I think it is? The two available choices are actually auth_login and auth_basic.

gizmocuz: Now it is:
https://github.com/domoticz/domoticz/commit/5a00ba936ab9b39606add5ab7a783297fcd1446f

gordonb3: Haha... I did not want to question existing constructs that were outside the scope of this PR. Hence it seemed appropriate to use the same method as there would likely be a good reason for it.

gordonb3:
Looks like the build system is messed up?

webserver/cWebem.h
gizmocuz: I know, but this is how it was supposed to do, and could inspire you to also use enums !

main/domoticz.cpp
```diff
@@ -171,6 +172,7 @@ bool g_bStopApplication = false;
 bool g_bUseSyslog = false;
 bool g_bRunAsDaemon = false;
 bool g_bDontCacheWWW = false;
+_eWebCompressionMode g_wwwCompressMode;
```
gizmocuz: Could you set the default here (http::server::WWW_USE_GZIP)
Then the 'else' if (cmdLine.HasSwitch("-wwwcompress")) here is not needed

gizmocuz (commented):
Could you add the default compression mode
This is also needed as when there is no -wwwcompress command line specified it is undefined.
(it's also good to always initialize variables)

main/domoticz.cpp
gordonb3: Aha! ***That*** is why I had it initialized to 0 when it was still an int...

gizmocuz: Yes, that's why it is good everything is triple checked ;)

gizmocuz:
Thanks!!

PR #2422, opened 2018-05-25
Add systemd unit file

FedericoCeratto:
Initial version. Related to #178

### DIFF
--- a/misc/domoticz.service.in
+++ b/misc/domoticz.service.in
@@ -1,0 +1,46 @@
+[Unit]
+Description=domoticz - Home Automation System
+Documentation=https://www.domoticz.com/
+Documentation=https://www.domoticz.com/wiki
+After=time-sync.target network.target multi-user.target
+ConditionPathExists=/etc/domoticz.conf
+
+[Service]
+Type=simple
+TimeoutStopSec=30
+KillMode=mixed
+KillSignal=SIGTERM
+
+User=domoticz
+Group=domoticz
+ExecStart=@SBINDIR@/domoticz -f /etc/domoticz.conf
+Restart=on-failure
+RestartSec=5s
+
+# execve() cannot regain privileges
+NoNewPrivileges=true
+CapabilityBoundingSet=CAP_NET_BIND_SERVICE
+
+# makes / read-only except /dev /proc /sys
+ProtectSystem=strict
+ProtectKernelModules=yes
+ProtectKernelTunables=yes
+ProtectControlGroups=yes
+
+# hides /home /root /run/user
+ProtectHome=yes
+
+# dedicated volatile /tmp
+PrivateTmp=true
+
+ReadWriteDirectories=/proc/self
+ReadWriteDirectories=@SHAREDSTATEDIR@/domoticz
+
+# Access to devices is required
+PrivateDevices=no
+
+StandardOutput=syslog
+StandardError=syslog
+
+[Install]
+WantedBy=multi-user.target


gizmocuz:
Thanks! Could you also add the config file
Could you make all options in the config file including the startup exe
I was planning to add this for but of course for
/home/pi/domoticz
Now I understand people want to place everything in different places and folders... Coz that makes it easy to backup
So then a good configuration for that I can patch during the install script under a raspberry or dervish system will solve that problem

pipiche38:
I Strongly recommend to add in the Unit section the following statement, in order to have Domoticz starting only when Multi-user and network are up-and-running
After=time-sync.target network.target multi-user.target

Otherwise there is a risk that it start too early before the system is ready to operate

koenkooi:
The paths don't have to be in the config file, they should get filled in by cmake, like https://git.icinga.com/github-mirror/icinga2/commit/69c06115d2dd27f665bfe9955e9648580f52ef8b

gizmocuz:
Let us assume that 90% of all installations are binary installs via our setup script (see our website/download page)
I am willing to accept a systemd service file as standard
This means, the install script has to patch the service file ? Or the config file ? I think the later is nicer ?
Why, suppose the config file can overwrite any possible command line option available in domoticz
you have to search/edit in one place, the config file, not the service file, and the config file

koenkooi:
You'll need to process the systemd unit regardless of what the config file will be able to override, since the unit needs to have the path to the config file.

Traditionally environment scripts in /etc/sysconfig and /etc/default have been used to set all the commandline option for the daemon, but not the path to the daemon itself.

FedericoCeratto:
I'm updating the PR with a cmake template.

gizmocuz:
Why not /etc/domoticz.conf or /etc/domoticz/domoticz.conf ? Normal location

koenkooi:
Files in /etc/default/ or /etc/sysconfig/ get parsed by the startup script, files like /etc/domoticz.conf get parsed by the daemon itself.

gizmocuz:
If thats so, then we should make a patch for domoticz that it used/reads the config file when it is found in /etc/domoticz.conf
That will simplify the startup script too

gizmocuz:
We could add a line DEAMON_OPTS="-f /etc/domoticz.conf" in the startup script and add support for this in domoticz

FedericoCeratto:
@gizmocuz /etc/default/domoticz has the (minor) advantages of having a somewhat standard format (EnvironmentFile) that can be more easily understood by configuration tools and scripts, and requiring no code changes.

gizmocuz:
i don't agree on that, a simple /etc/domoticz.conf is much easier, and more widely known, or if other configuration files are expected /etc/domoticz/domoticz.conf

FedericoCeratto:
Ok, then: updated to act as a cmake template, use @SYSCONFDIR@ and related variables and -f

gizmocuz:
I will build support for the -f parameter.

gizmocuz:
Added an Initial configuration file (not yet implemented in domoticz), feel free to make adjustments

https://github.com/domoticz/domoticz/commit/21dd200b628593ae4a300744a58a0d6b31be8ac7

will give this a few days before starting implementation in the application

FedericoCeratto:
@gizmocuz few suggestions: you usually don't need a license in a conf file. Also, it could be shipped as a cmake template in order to use @LIBDIR@ and friends.

( Linux distributions should patch out the  *_root, daemon, pid_file, userdata_path examples and hardcode the correct values in the init and systemd unit file. )

gizmocuz:
"usually" yes, and it is also common to have one, so lets just have one to make life simple right ?
So the patch should be done in the configuration file
Again, i want a global usable systemd file, putting the options in a configuration file (... that is what they are for) seems the way to go

gizmocuz:
Could someone please respond to pipiche38 message, it sound logical

FedericoCeratto:
@gizmocuz the systemd file needs to use @libdir@ and others anyways in order to find the executable and allow read/write permissions to the right directories.

pipiche38:
I had a look to the propose domoticz.service and I think my suggestion was taken into account.

gizmocuz:
@pipiche38 , great , could not find an comment about this

@FedericoCeratto , for the raspberry/debian installer, i will still use one folder in the users home folder (domoticz)
Reason is extreme simple, and i also urge you to do the same in the /opt folder

"backup/restore"

When you scatter the executable, scripts, www and log/database patch all over the system, it is very complex to make a complete backup/restore
Specially when users are also installing their own scripts, web pages, etc etc 
(monolithic method)

For this reason, i like the systemd file (startup script) to be as transparent as possible so I can also use it.

So if you patch the config file to those folders, that's perfect, i need to do the same for the startup folder

gizmocuz:
Just committed the code that handles the configuration file parsing

gizmocuz:
these variables can be taken from  the configuration file:

ExecStart=@SBINDIR@/domoticz \
+  -wwwroot @LIBDIR@/domoticz/www/ \
+  -dbase @SHAREDSTATEDIR@/domoticz/domoticz.db \

gizmocuz:
Some questions/comments

1.
these variables can be set in the configuration file and removed from the systemd file

-wwwroot
-dbase

2.
What happens when we update via the web interface? Internally a service restart is done that could take some time, let's say a minute.

Will this setting be a problem ?

RestartSec=5s

Should this be set to 70s ?

3.
Admin is allowed to reboot the system from the menu, will this not cause a SIGSYS signal ?

SystemCallFilter=@reboot

4.
PrivateDevices is default false, and can be removed from this service file

5.
when domoticz terminates, it tries to stop all hardware nicely.

TimeoutStopSec=10

might be to short ! and could result in a database flush issue

I would recommend setting this to 30

6.
As per documentation i would suggest the following two lines as there is no system man page

Documentation=https://www.domoticz.com
Documentation=https://www.domoticz.com/wiki

7.
The domoticz process runs as a Daemon, should the service type not be set to 'forking' ?
Do we need to supply the PIDFile ?

8.
Instead of a vague /usr/local/etc folder, let's just use /etc/domoticz.conf

FedericoCeratto:
Replying to your questions below. Of course feel free to change the unit file when merging.

> What happens when we update via the web interface? Internally a service restart is done that could take some time, let's say a minute.
> RestartSec=5s

That's an interval between restarts done by systemd, not the service warmup time, so it's ok.

> Admin is allowed to reboot the system from the menu, will this not cause a SIGSYS signal ?
> SystemCallFilter=@reboot

Normally a service or a non-root user should not be allowed to reboot a system.

> PrivateDevices is default false, and can be removed from this service file

That's meant to make the configuration more explicit - also the default might change in future.

> The domoticz process runs as a Daemon, should the service type not be set to 'forking' ?
Under systemd stuff should not demonize if possible.

>Do we need to supply the PIDFile ?

Not needed anymore under systemd

>Instead of a vague /usr/local/etc folder, let's just use /etc/domoticz.conf

The template file uses @SYSCONFDIR@/domoticz.conf  to make it configurable.

FedericoCeratto:
> these variables can be set in the configuration file and removed from the systemd file
> -wwwroot
> -dbase

The unit file needs to have the right paths to use ReadWriteDirectories , also distributions-specific install scripts needs to know which directories to create and assign to the daemon user.
Also, those paths are not meant to be changed by the user after installation.

FedericoCeratto:
> When you scatter the executable, scripts, www and log/database patch all over the system, it is very complex to make a complete backup/restore

Files are grouped by category and not by application on purpose because they have different lifetime and access pattern. E.g. /usr/bin are host-specific, /etc is not, /var/log files allow rotation and compression, /var/lib is backed up. Configuration, backup and security tools treat them differently and in this way all installed applications can be managed together.

gizmocuz:
@FedericoCeratto 

>>Admin is allowed to reboot the system from the menu, will this not cause a SIGSYS signal ?
>>SystemCallFilter=@reboot
>Normally a service or a non-root user should not be allowed to reboot a system.

Yes, but only an admin user can perform this action, so it's OK and should be able to do so

>>PrivateDevices is default false, and can be removed from this service file
>That's meant to make the configuration more explicit - also the default might change in future.

Correct, better not to declare it at all

>>Instead of a vague /usr/local/etc folder, let's just use /etc/domoticz.conf
>The template file uses @sysconfdir@/domoticz.conf to make it configurable.
Thats nice, and in certain cases do not point to /etc/domoticz.conf

Please change this to /etc/domoticz.conf as this is commonly used

>>these variables can be set in the configuration file and removed from the systemd file
>-wwwroot
>-dbase

They are declared in the domoticz.conf file and do not need to be declared here as well

>>When you scatter the executable, scripts, www and log/database patch all over the system, it is very >>complex to make a complete backup/restore
>Files are grouped by category and not by application on purpose because they have different lifetime >and access pattern. E.g. /usr/bin are host-specific, /etc is not, /var/log files allow rotation and >compression, /var/lib is backed up. Configuration, backup and security tools treat them differently and in this way all installed applications can be managed together.

Sure... but practice in this case ? No!
Anyway, i don't mind that users can not do a update via the web interface, but don't forget to disable that option in the configuration file (domoticz.conf)


>>The domoticz process runs as a Daemon, should the service type not be set to 'forking' ?
>>Under systemd stuff should not demonize if possible.
>>Do we need to supply the PIDFile ?
>Not needed anymore under systemd

Does this mean domoticz will not run as a daemon then ? That is fine by me
If it needs to run as a daemon, i have a problem now that when using a normal user,
the daemon is not allowed to create the PID file under /run

emontnemery:
@FedericoCeratto, @gizmocuz Do you think this will be ready any time soon?
I want to open a PR with changes to improve the stability/reliability when some HW doesn't stop properly, for this it would be great if systemd is managing the domoticz process.

FedericoCeratto:
@gizmocuz I updated the PR.

gizmocuz:
@FedericoCeratto , thanks !
I would remove 'SystemCallFilter', where also the @debug is mentioned twice
The current initd is places in /scripts, i would prefer this one as well

FedericoCeratto:
Updated. Feel free to make changes to the PR.

gizmocuz:
Thanks! Can you move it to the scripts folder instead of the misc folder ?

muhammadtajuddin78:
halo

Lucasmarina (commented):
Don't do this again plz!

gizmocuz:
Thread not active anymore, closing the issue. Feel free to re-open if needed

PR #2423, opened 2018-05-25
Serial Helper : Add support for symlink /dev/serial/by-id/

badzz:
/dev/serial/by-id/ folder now exits that point to the real ttyUSBx device.
It is much more convenient as it is based on the USB vendorId and productID which are stable and independant of usb insertion order.

With this for each heardware, the drop down list of serial entries include those by-id devices.

Signed-off-by: Matthias Badaire <mbadaire@gmail.com>

### DIFF
--- a/main/Helper.cpp
+++ b/main/Helper.cpp
@@ -299,6 +299,25 @@ std::vector<std::string> GetSerialPorts(bool &bUseDirectPath)
 		closedir(d);
 	}
 
+        //also scan in /dev/serial/by-id
+        d=opendir("/dev/serial/by-id");
+	if (d != NULL)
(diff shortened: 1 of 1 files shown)


gizmocuz:
That won't work:
- in the database, the length of the serial port field (which was normally /dev/ttyUSB102 or \\COM10) is 50 characters long

With this patch, i get names like 'usb-Silicon_Labs_CP2102_USB_to_UART_Bridge_Controller_0001-if00-port0'
This also does say nothing to anyone, but it is 70 characters long
You can name your own port if you like, see

https://www.domoticz.com/wiki/PersistentUSBDevices

Issue #2410, opened 2018-05-23
The different Daikin devices status are not updated 


pipiche38:
In case of a change via an other remote control, of any Daiking devices parameters, this is not reflected into the Domoticz Devices.

For instance if with the Daiking remote controller you power on or power off the AC, the corresponding Power Devices in domoticz is not updated accordingly.
This is the same for Mode, Fan rate and Fan direction

Information is available through a /aircon/get_control_info URL

pipiche38:
Did some investigation while reading the code.
(1) The polling is done only every Daikin_POLL_INTERVAL 300  ( 300s / 15' ) which seems to be quiet long. 30'' or at least 1' would be more favourable to handle update done by an other controller.

(2) I tried to understand how the decoding is done, but my knowledge on cpp syntax is limited. But I did a test of reducing Daikin_POLL_INTERVAL to 30 (seconds), and there is no update done. So I suspect that the response of  `/aircon/get_control_info` is not processed correctly.
Should be somewhere in 
`CDaikin::GetControlInfo()` function on the section
` for (itt = results.begin(); itt != results.end(); ++itt)`

gizmocuz:
Did you enable 'accept new hardware' under settings ?
Else it might be your username/password, please contact me on the forum (gizmocuz) and we can try to debug together. When i wrote this i tested online to another user, was working great

pipiche38:
The Devices are already present in Domoticz, it is about polling information from the Daikin interface to the Domoticz devices.
I have currently identified that the Polling is set to 300s ( 15 minutes ) which seems to be quiet long. There is maybe a Domoticz needs for that, but I think 60s would be more acceptable.

I also started to instrument the code a bit in order to understand what is happening when update is needed.

Will come back with recommendation ;-)

pipiche38:
Discussion on Forum is available here: https://www.domoticz.com/forum/viewtopic.php?f=34&t=23566

pipiche38:
I think , I found the issue in the code. If I'm not mistaken the Switch name is hardcoded, while on my setup and because I have 3 Daikin hardware I have renamed them.

For instance , is the following is not hardcoding "Ventillation", "Winds", "Mode" and "Power"
```InsertUpdateSwitchSelector(6, true, 10, "Ventillation");
InsertUpdateSwitchSelector(7, true, 10, "Winds");
InsertUpdateSwitchSelector(5, true, 50, "Mode");
[...]

PR #2425, opened 2018-05-25
Update domoticz Daikin status in case changes are detected during poll

pipiche38:
Defect #2410
In case they are some changes done on the Daikin HW via for instance remote control , we need to update the Domoticz devices (Selectors) accordingy

Up to now, those selector were created after addition of the new HW, but were never updated during the polling exercise

### DIFF
--- a/hardware/Daikin.cpp
+++ b/hardware/Daikin.cpp
@@ -749,6 +749,7 @@ void CDaikin::InsertUpdateSwitchSelector(const unsigned char Idx,  const bool bI
 	result = m_sql.safe_query("SELECT nValue, BatteryLevel FROM DeviceStatus WHERE (HardwareID==%d) AND (DeviceID=='0000000%d') AND (Type==%d) AND (Unit == '%d')", m_HwdID, sID, xcmd.type, xcmd.unitcode);
 	if (result.size() < 1)
 	{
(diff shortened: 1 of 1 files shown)


pipiche38:
Let's put it on hold, as they is some side effects to be verified

pipiche38:
Ready for merge ;-) Please let me know if anything should be done.

gizmocuz:
Okey, a little change maybe
Just before if (result.size()<1)
could you move there 
m_mainworker.PushAndWaitRxMessage(this, (const unsigned char *)&xcmd, defaultname.c_str(), 255);

so you get

```
	result = m_sql.safe_query("SELECT nValue, BatteryLevel FROM DeviceStatus WHERE (HardwareID==%d) AND (DeviceID=='0000000%d') AND (Type==%d) AND (Unit == '%d')", m_HwdID, sID, xcmd.type, xcmd.unitcode);
	m_mainworker.PushAndWaitRxMessage(this, (const unsigned char *)&xcmd, defaultname.c_str(), 255);
	if (result.size() < 1)
	{
```

this way you do not need the 'else'

could you also change

if (result.size() < 1)

to

if (result.empty())

pipiche38 (commented):
Ready for merge

gizmocuz:
Thanks!

PR #2432, opened 2018-05-27
Small UI improvements for Managed Counter

guillaumezin:
Don't show 0 as last value for week/month/year log view


gizmocuz:
Thanks!

PR #2436, opened 2018-05-28
Extra commit for PR 2420

gordonb3:
Use thread safe gmtime_r/gmtime_s in make_web_time

Replace use of the strftime_t method with make_web_time for setting the images expiration timestamp.


gizmocuz:
Thanks!!

PR #2450, opened 2018-05-31
correct position of sun times and room selector in themes

gordonb3:


gizmocuz:
Thanks for the patch. I was not aware there was a problem with the 'default' theme ?
Could you show a screenshot where it goes wrong, and what changed ?

gordonb3:
No, the default theme is good. The problem is with the `#holder .container-fluid` class that is set to have 20px top margin in the default theme but references the following definition in bootstrap.css for all the other themes. This is why the named items are mostly hidden under the top bar in those themes.

```
.container-fluid {
    padding-right: 20px;
    padding-left: 20px;
    *zoom: 1;
}
```

images here: https://www.domoticz.com/forum/viewtopic.php?f=6&t=23671&p=182134

gizmocuz:
Does this not corrupt the default theme ?
I see a change in legacy.css, but not in the default css theme, but all other themes do have a change
Is it not better to just correct the other themes ?

gordonb3:
Actually, some of the themes were in fact adapted on an individual base to counter the effect. Which is why I had to change those specific themes again to counter the countermeasures not to let the content shift down too much.

No this does not corrupt the default theme. Not sure if that even includes legacy.css, but the entry in that file is the exact same entry as specified in base.css from the default theme.

gordonb3:
Oh... guess I need to increment the appcache counter?

gizmocuz:
@gordonb3 , no we never have to increase the version number in appcache anymore

gizmocuz:
I had a look in the bootstrap css file, and here it is also 20/20,
so i guess we can remove it from legacy and base css ?

gordonb3:
Look closer. In bootstrap it is left-right. In base it is top-bottom. It is the latter that makes the positioning correct in default theme, but not in e.g. simple-grey where this is missing.

gordonb3:
Effective rules in default theme according to Firefox inspector:

_____________________//base.css
#holder .container-fluid {
    margin-top: 20px;
    padding-bottom: 20px;
}
_____________________//legacy.css
#holder .container-fluid {
   ~~margin-top: 20px;~~
   ~~padding-bottom: 20px;~~
}
_____________________//bootstrap.css
.container-fluid {
    padding-right: 20px;
    padding-left: 20px;
    ~~*zoom: 1;~~ // -> invalid property value
}

So that leads up to:
```
#holder .container-fluid {
    margin-top: 20px;
    padding-bottom: 20px;
    padding-right: 20px;
    padding-left: 20px;
}
```

gizmocuz:
Thanks, that tricked me ;)

PR #2452, opened 2018-06-01
correct for broken lua pkg-config

gordonb3:
Standard Lua 5.2 pkg-config does not correctly return the include path and library name. If the distro did not create links to provide backward compatibility with older Lua (if the API itself wasn't changed as well) then we need to find these ourselves.


gizmocuz:
Thanks!

gordonb3:
Sloppy....

On line 67 the ENDIF argument does not match the IF argument (~NOT~ LUA_INCLUDE_DIRS) causing a cmake warning at build time. Can you add the missing `NOT`?

gizmocuz:
Sure, done in this commit
https://github.com/domoticz/domoticz/commit/af1acd4a0024dee986622846221660d85a9506fa

gordonb3:
Thanks

PR #2461, opened 2018-06-05
Added new hardware class for Denkovi devices

dzelyaydinov:
Support for DAEnet IP4, SmartDEN IP 16, SmartDEN IP 32 In, SmartDEN Logger, SmartDEN Watchdog, SmartDEN Notifier and SmartDEN IP-Maxi.

### DIFF
--- a/main/SQLHelper.cpp
+++ b/main/SQLHelper.cpp
@@ -2427,6 +2427,7 @@ bool CSQLHelper::OpenDatabase()
 		{
 			safe_query("UPDATE Hardware SET Mode1 = 5000 WHERE Type = %d", HTYPE_DenkoviSmartdenIPInOut);
 			safe_query("UPDATE Hardware SET Mode1 = 5000 WHERE Type = %d", HTYPE_DenkoviSmartdenLan);
+			safe_query("UPDATE Hardware SET Mode1 = 5000 WHERE Type = %d", HTYPE_DenkoviDevices);
 		}
 		if (dbversion < 124)
 		{
--- a/main/SQLHelper.cpp
+++ b/main/SQLHelper.cpp
@@ -35,7 +35,7 @@
 #define __STDC_FORMAT_MACROS
 #include <inttypes.h>
 
-#define DB_VERSION 129
+#define DB_VERSION 130
 
 extern http::server::CWebServerHelper m_webservers;
 extern std::string szWWWFolder;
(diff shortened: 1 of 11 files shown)


gizmocuz:
Thanks for the patch! There are build errors, see the problem here above/below
Also when using parameters of a function that do not change, please declare them as 'std::string &parameter'

main/SQLHelper.cpp
```diff
@@ -2427,6 +2427,7 @@ bool CSQLHelper::OpenDatabase()
 		{
 			safe_query("UPDATE Hardware SET Mode1 = 5000 WHERE Type = %d", HTYPE_DenkoviSmartdenIPInOut);
 			safe_query("UPDATE Hardware SET Mode1 = 5000 WHERE Type = %d", HTYPE_DenkoviSmartdenLan);
+			safe_query("UPDATE Hardware SET Mode1 = 5000 WHERE Type = %d", HTYPE_DenkoviDevices);
```
gizmocuz: You need to increase the database version (to 130 as we speak) and make a new 'if' line for this change

dzelyaydinov: Thanks for the remind. Is this mean to remove this line 
```
			safe_query("UPDATE Hardware SET Mode1 = 5000 WHERE Type = %d", HTYPE_DenkoviDevices);
```
and add
```
if (dbversion < 130)
		{
			safe_query("UPDATE Hardware SET Mode1 = 5000 WHERE Type = %d", HTYPE_DenkoviDevices);
		}
```
Thanks in advance.

gizmocuz: yes it is, thanks !

main/SQLHelper.cpp
```diff
@@ -2428,6 +2428,10 @@ bool CSQLHelper::OpenDatabase()
 			safe_query("UPDATE Hardware SET Mode1 = 5000 WHERE Type = %d", HTYPE_DenkoviSmartdenIPInOut);
 			safe_query("UPDATE Hardware SET Mode1 = 5000 WHERE Type = %d", HTYPE_DenkoviSmartdenLan);
 		}
+		if (dbversion < 130)
```
gizmocuz: Something seems not right here
you placed a check for version 130 above 124 ?
Could you place it where it belongs ?
You also need to increase the actual database version with 1 at the top of the class

gizmocuz:
Thanks for this great addon !
Shall i remove the following hardware types?:

HTYPE_DenkoviSmartdenLan:
HTYPE_DenkoviSmartdenIPInOut:

gizmocuz:
Another question, your new hardware class is

HTYPE_DenkoviDevices

Then i think you do not have to patch the database setting the type to 5000, as it is a new hardware class correct ?

		if (dbversion < 130)
		{
			safe_query("UPDATE Hardware SET Mode1 = 5000 WHERE Type = %d", HTYPE_DenkoviDevices);
		}

dzelyaydinov:
Thanks again for the support. Yes, you are right about this:
```
Then i think you do not have to patch the database setting the type to 5000, as it is a new hardware class correct ?
```
Please do not remove old hardware types if this will make trouble to our current clients. We just do not want our clients to lose their hardware and devices.

gizmocuz:
@dzelyaydinov , true, but it is a one time job they have to do...
But maybe we will start removing the old classes after the new stable release

PR #2471, opened 2018-06-08
Master ctrl_86plug.aq1 wall socket plug

AlbrtoSam:


gizmocuz:
Sorry, you patched the wrong file here, you need to undo the changes from Rtl433.cpp

Issue #2470, opened 2018-06-08
ctrl_86plug.aq1 wall socket plug


AlbrtoSam:
I cant see wall plug consumption using V3.9587 ( i dont know if it worked before ).

I can see this in logs:

Status: {"cmd":"heartbeat","model":"ctrl_86plug.aq1","sid":"158d000221b1b2","short_id":54728,"data":"{\"voltage\":3600,\"status\":\"on\",\"inuse\":\"1\"}"}

I cant see power usage also...

AlbrtoSam:
I already receive power usage on logs, but domoticz doesnt create a consumption device:

2018-06-08 01:16:23.668 Status: {"cmd":"heartbeat","model":"ctrl_86plug.aq1","sid":"158d000221b1b2","short_id":54728,"data":"{\"voltage\":3600,\"status\":\"on\",\"inuse\":\"1\",\"power_consumed\":\"1\",\"load_power\":\"3.71\"}"}

I saw where the problem is and i solved changing code. It works for me now.

On domoticz/hardware/XiaomiGateway.cpp

I Solved changing on line 872 name = "Xiaomi Smart Wall Plug";   with name = "Xiaomi Smart Plug";
I changed that to trigger line 531 if (Name == "Xiaomi Smart Plug") {
			if (load_power != "" && power_consumed != "") {
				int power = atoi(load_power.c_str());
				int consumed = atoi(power_consumed.c_str()) / 1000;
				SendKwhMeter(sID, 1, 255, power, consumed, "Xiaomi Smart Plug Usage");
			}

So i guess that a better way to solve it is just change on line 531:
if (Name == "Xiaomi Smart Plug") {
if ((Name == "Xiaomi Smart Plug") || (Name == "Xiaomi Smart Wall Plug")) {

gizmocuz:
@AlbrtoSam , 

"if ((Name == "Xiaomi Smart Plug") || (Name == "Xiaomi Smart Wall Plug")) {"

sounds good, could you make a pull request for this change ?

Do you expect any more different "plugs" ?

If yes, then you can maybe also use

if (Name.find("Plug")!=std::string::npos) {
...

AlbrtoSam:
I have never did a pull request before but im going to try.  Regarding the second part I saw the same structure along the code, and i dont think there will be another type of plug. ( there are some other xiaomi plugs but they are wifi )

gizmocuz:
Great, here is the steps you need to take to make a PR
[...]

PR #2472, opened 2018-06-08
Update XiaomiGateway.cpp

AlbrtoSam:
Commented here: https://github.com/domoticz/domoticz/issues/2470

Aqara wall plug recognized as ctrl_86plug.aq1 doesnt show power usage:

So i guess that a better way to solve it is just change on line 531:
if (Name == "Xiaomi Smart Plug") {
if ((Name == "Xiaomi Smart Plug") || (Name == "Xiaomi Smart Wall Plug")) {


gizmocuz:
Thanks for the fix/update!

AlbrtoSam:
Thank you for you help !!! :)

PR #2475, opened 2018-06-08
When compiling LUA, include FreeBSD as a target in addition to Linux …

Whiskey24:
…when adding -DLUA_USE_LINUX

This fixes dzVents not working on FreeBSD. See this forum post: http://www.domoticz.com/forum/viewtopic.php?f=59&t=23743


gizmocuz:
Thanks!

PR #2478, opened 2018-06-09
GoodweAPI fixes

llagendijk:
hi
I had to update the Goodwe API as Goodwe splitted their server infrastructure per region. 
I added a server selection field in the hardware screen for that.

I also added a fix for the case where the inverter lost contact, as that might result in 0 values that mess up the database. So unless the status is "normal" we skip updating the database.

Please review the changes, as I am not a very good (C++) programmer as you may remember

Kind regards, Louis

### DIFF
--- a/hardware/GoodweAPI.h
+++ b/hardware/GoodweAPI.h
@@ -7,11 +7,12 @@
 class GoodweAPI : public CDomoticzHardwareBase
 {
 public:
-	GoodweAPI(const int ID, const std::string &userName);
+	GoodweAPI(const int ID, const std::string &userName, const int ServerLocation);
 	~GoodweAPI(void);
 	bool WriteToHardware(const char *pdata, const unsigned char length);
 private:
 	std::string m_UserName;
+	std::string m_Host;
 	volatile bool m_stoprequested;
 	boost::shared_ptr<boost::thread> m_thread;
 
@@ -20,6 +21,7 @@ private:
 	bool StopHardware();
 	uint32_t hash(const std::string &str);
 	int getSunRiseSunSetMinutes(const bool bGetSunRise);
+	bool GoodweServerClient(const std::string &sPATH, std::string &sResult);
 	float getPowerWatt(const std::string &str);
 	float getEnergyWh(const std::string &str);
 	void Do_Work();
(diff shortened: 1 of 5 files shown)


hardware/GoodweAPI.h
```diff
@@ -20,13 +21,19 @@ class GoodweAPI : public CDomoticzHardwareBase
...
 	int getSunRiseSunSetMinutes(const bool bGetSunRise);
+	bool GoodweServerClient(const std::string &sPATH, std::string &sResult);
 	float getPowerWatt(const std::string &str);
 	float getEnergyWh(const std::string &str);
 	void Do_Work();
 	void GetMeterDetails();
 	void ParseStation(const std::string &sStationId, const std::string &sStationName);
 	void ParseDeviceList(const std::string &sStationId, const std::string &sStationName);
 	void ParseDevice(Json::Value device, const std::string &sStationId, const std::string &sStationName);
+	enum _eGoodweLocation {
```
gizmocuz: Can you move the enum just right underneath:

```
class GoodweAPI : public CDomoticzHardwareBase
 {
```

That's more easier to find for all developers, and where i always place them ;)

www/views/hardware.html
```diff
@@ -1662,6 +1662,15 @@ <h2><span data-i18n="Hardware">Hardware</span>:</h2>
                         <br>
                         <table class="display" id="hardwareparamsgoodweweb" border="0" cellpadding="0" cellspacing="20">
                                 <tr>
+					<td align="right" style="width:110px"><label for="comboserverselect"><span data-i18n="Select a Server">Select a Server</span>:</label></td>
```
gizmocuz: How about a rename from "Select a Server" to "Location", this string is already in use and translated in all/most languages

www/views/hardware.html
```diff
@@ -1662,6 +1662,15 @@ <h2><span data-i18n="Hardware">Hardware</span>:</h2>
...
+					<td><select id="comboserverselect" style="width:200px" class="combobox ui-corner-all">
+						<option value="0">Global</option>
+						<option value="1">Oceania</option>
+						<option value="2">Europe</option>
+						</select>
+					</td>
+				</tr>
+				<tr>
 					<td align="right" style="width:110px"><label><span data-i18n="username">Goodwe-power Username</span>:</label></td>
```
gizmocuz: Could you change data-i18n="username" to use "Username" (capital U), this one is already in use and translated.
Also >Goodwe-power Username rename to 'Username' only, like the data-i18n, on better, you can remove that

gizmocuz: Best to open 'a' language file from www/i18n folder, and search what you can use

gizmocuz:
Patch is looking perfect, i just have some tiny remarks...

llagendijk:
HI
I made the requested changes. Thanks for the hints. I removed the duplication of the Location as well as the duplication on Username for consistency

gizmocuz:
Thanks, can you move the enum to the cpp file ?

llagendijk:
no problem, done

gizmocuz:
Thanks !!

PR #2492, opened 2018-06-14
Remove unecessary test

pipiche38:
due probably to some copy-paste from other Hardware code, remove unecessary test (which was always true)


gizmocuz:
Thanks!

PR #2503, opened 2018-06-19
 full overhaul of HarmonyHub code

gordonb3:
Now using a centralized reader function rather than write and wait for Harmony Hub to respond. More or less mimics asynchronous operation, but with full control when the reader may be triggered. Should be possible to move on to using the Async class from here, but can currently not fully predict how that will work out in real life and possibly lead to loss of data due to overwriting/re-using of buffers, so I won't be going that way this time.


gizmocuz:
Thanks for the great work again! Much appreciated!

Issue #2504, opened 2018-06-20
GCM device notifications not working in latest beta


bostjane:
Hi,

I've upgraded from stable v3.8153 to beta v3.9657 on Windows and Mobile Device notification no longer works. Log says:
Error: SQL Query("SELECT SenderID, DeviceType FROM MobileDevices WHERE (ID IN (pushover))") : no such column: pushover

bostjane:
I wasn't specific enough. I was talking about GCM messaging. The log message appeared when send via dzVents. When pressing "Test" button for GCM in Settings->Notifications there was no such message, but messaging still doesn't work.
After going back to stable release it works just fine...

gizmocuz:
Not a fix, but could it be you specified a certain device ? ( |Device= )
(Just looking at the code)

bostjane:
You mean if I specified a device when sending notification in dzVents? No I haven't, here is the line:
domoticz.notify('Vlom v Bitnjah#devid=v92E48E15FA935AA&DETAILS='..device.name)

I did a few more test and when hitting the "Test" in Settings->Notifications notification arrive almost instantly in stable release, while in beta notification arrive with minutes delay at best, in most cases does not arrive at all. Interestingly, log says success.
Using dzVents with line above, notification never arrive and SQL error appears in log. 

I have two mobile devices in Setup->More options->Mobile Devices, both enabled for notification. I don't know if it has anything to do with it, but I tested beta on different computer with imported Domoticz database and OZW config file. Everything other seems to work just fine.

Further, I use two notification systems, GCM and Custom http. I would like to point out that when notification is called from dzVents, custom http works, only GCM fails.

gizmocuz:
@bostjane , i think there is no difference in code between the stable or beta, but lets find out by looking at the history of the GCM notification helper:

https://github.com/domoticz/domoticz/commits/development/notifications/NotificationGCM.cpp

Okey, there seems to be 5 changes, some of non importance

Thats all i can do as i don't use GCM, and don't know how to debug.

Maybe @galadril or @Olifant1990 could have a small look onto this one ?

galadril:
If you use the test button on the mobile devices page to test a push notification.. does that one work?
It's the green arrow icon at the right side:

![afbeelding](https://user-images.githubusercontent.com/14561640/41707833-53f7476a-753f-11e8-8de5-cabbdfbbac39.png)

Olifant1990:
@bostjane, If you can receive the test message, GCM is working fine.
[...]

PR #2509, opened 2018-06-21
Rewritten GCM Specific phone selection method

Olifant1990:
Rewritten GCM Specific phone selection method

Just now you need to do the following:
table.insert(commandArray, {['SendNotification'] = "Subject#Message###midx_7;8#gcm"})

instead of:
table.insert(commandArray, {['SendNotification'] = "Subject#Message###7;8#gcm"})

I think, because I build it and I did not change the wiki's, this wasn't used, so no (much) problems for other users.

midx_ was introduced by @gizmocuz to do a single test message from the mobile devices overview


Olifant1990:
#2504

gizmocuz:
Thanks for the fast response !

PR #2522, opened 2018-06-27
Adding eurotronic spiritz Radiator Valve support

pipiche38:
Adding eurotronic spiritz Radiator Valve support


gizmocuz:
You have to submit this directly to the openzwave team (based on their master branch), also include the 'eur_spiritz.xml'

PR #2524, opened 2018-06-27
UI improvements for managed counters

guillaumezin:
Broken during code refactory 41fcb49f7d5d7b495ad906e651d488255a2e7e8b

### DIFF
--- a/main/WebServer.cpp
+++ b/main/WebServer.cpp
@@ -13255,6 +13255,9 @@ namespace http {
 			else if ((dType == pTypeRego6XXValue) && (dSubType == sTypeRego6XXCounter))
 				metertype = MTYPE_COUNTER;
 
+			// Special case of managed counter: Usage instead of Value in Meter table, and we don't want to calculate last value
+			bool bIsManagedCounter = (dType == pTypeGeneral) && (dSubType == sTypeManagedCounter);
+
 			double AddjValue = atof(result[0][3].c_str());
 			double AddjMulti = atof(result[0][4].c_str());
 			std::string sOptions = result[0][5].c_str();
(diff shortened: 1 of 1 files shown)


gizmocuz:
What was the problem with the refactoring ?
I only see new code ? I can not remember i removed any of those 'if/then/else' lines

guillaumezin:
On dc3287f5ae807c07912b264fd8d2da7e552c6632#diff-b178aa1b57f3c23ff1c6fbce4b495e30 you replaced "if (!((dType == pTypeGeneral) && (dSubType == sTypeManagedCounter))) {" with "else {" lines 15173, 16370, 16702 and 17434 of the previous git version (https://github.com/domoticz/domoticz/blob/dc3287f5ae807c07912b264fd8d2da7e552c6632/main/WebServer.cpp)

From lines 14238 to 14383 there was a block to trace the graph for Managed Counter, I implemented it differently this time, using if/then/else in the next else statement (lines 14384 to 14595 of the previous git version), to not have almost the same code twice, I think that's why you don't remember part of those if/then/else.

gizmocuz:
That seems not related to "Broken during code refactory 41fcb49"

Can you click on the above link you posted ? There are no changes in the webserver.cpp code related counters at all

https://github.com/domoticz/domoticz/commit/dc3287f5ae807c07912b264fd8d2da7e552c6632

guillaumezin:
Just to a diff between Webserver.cpp from dc3287f and 41fcb49, and you'll see the changes I described, dc3287f was before the changes, 41fcb49 after

gizmocuz:
Ok...
I do remember there was a crash a short month ago, just after these changes, and i reverted them, and there was no crash anymore...

To make things clearer (for me or fellow developers), could you just add a boolean 'bIsanagedCounter', and use this instead of duplicating the code for type/subtype ? That might clear things up

guillaumezin:
> could you just add a boolean 'bIsanagedCounter', and use this instead of duplicating the code for type/subtype ?

Sure, I'll do it

guillaumezin:
Done!

gizmocuz:
Thanks!

PR #2534, opened 2018-07-02
Python 3.7 GIL changes

dnpwwo:
Python 3.7 forces Python threading to by initialized which means that Domoticz needs to handle the Global Interpreter Lock (GIL) now.  This was avoided previously for simplicity and because its use is poorly documented for use with sub-interpreters.

This pull request turns on threading for all versions of Python (3.4+) for consistency and manages the GIL through matched 'SaveThread' and 'RestoreThread' operations on the Python threads.

To be clear, the Plugin Framework is still as single threaded as it always was and this PR is backwards compatible with existing plugins.

Tested on Python 3.5, 3.6 and 3.7 on Windows 10 and Linux (Ubuntu 18.04).

This PR also adds a new 'Trace' function that plugin developers can toggle inside plugins to get line level tracking of their code executing.  This is a 1st version with minimal details and will be built out later.


gizmocuz:
Thanks!!

PR #2536, opened 2018-07-02
concurrent_queue dropped boost dependency

jvandenbroek:
Now we're able to use C++11 extensions, let's drop some unneeded boost dependencies :) Running this a few days without issues.

### DIFF
--- a/hardware/MySensorsBase.cpp
+++ b/hardware/MySensorsBase.cpp
@@ -2361,7 +2361,7 @@ void MySensorsBase::Do_Send_Work()
 	while (true)
 	{
 		std::string toSend;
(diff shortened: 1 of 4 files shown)


gizmocuz:
Thanks! I also did this, but then reverted it as c++11 does not have a 'scoped_lock'.
c++17 does... 
Would there be a problem when using a 'unique_lock' ?

jvandenbroek:
Nop essentially the same (I even think it's just a typedef for boost unique_lock..).

gizmocuz:
Yes, that's what i found on one site, the other side indicates is not just a typedef

https://en.cppreference.com/w/cpp/thread/scoped_lock

But i seems to run perfectly, so going to merge this ! Thanks again!

PR #2537, opened 2018-07-02
Fix OSX build override warnings

jvandenbroek:
Added override to derived virtual functions to remove XCode compilation warnings.


gizmocuz:
Thanks!

PR #2539, opened 2018-07-03
Converted many boost functions to C++11

jvandenbroek:
Converted a lot boost functions to C++11, needed to add some includes because of dropping generic boost/thread/mutex.hpp include in stdafx.h (which includes a lot headers used by many files). Tested it on both Linux and Windows, seems fine, but of course test it yourself before merging please ;)

joinable() is needed for safe calling join(), this wasn't the case with boost.


gizmocuz:
Great work! (And a lot of work..😎)

gizmocuz:
Question, why all the  std::thread::joinable() calls ?
Every thread that is not 'nullptr' is joinable...

gizmocuz:
Also this is joinable:

m_thread = std::shared_ptr<std::thread>(new std::thread(std::bind(&CAtagOne::Do_Work, this)));
return (m_thread != NULL && m_thread->joinable());

the thread will start right away

Please undo all joinable calls

jvandenbroek:
@gizmocuz hmm yes I would normally agree, but I had a situation which caused a race condition without it. But this probably is unnecessary now thinking about it for hardware anyway, since stophardware would normally only be called from the mainworker thread. Maybe there is a little bug somewhere which caused a stop event called twice, resulting in that race. So I concluded that before the boost join call is thread safe and std isn't, but that might not be entirely true... Will do some tests without, double join should indeed never happen..

jvandenbroek:
@gizmocuz yes this is a search & replace thing, of course not needed in any case ;) tnx for noticing

Edit: the return value

jvandenbroek:
Ok, removed almost all joinable's, they seem redundant indeed. But except for TCPServer, which gets called twice on shutdown, throwing some nasty exception. Before with boost thread this didn't happen, I guess there is some additional check on join() for boost (clever boost ;))

Also changed the shared ptrs in EventSystem to normal ones, because disabling/enabling ES in settings no longer worked. By using joinable() before, this did work. However, after some debugging, this seems not the correct approach. The ES object doesn't get destroyed on disabling, so the pointers to it's threads remained valid, preventing to re-enable it (logic is first call stop before start on enable for some reason..). As a bonus Valgrind no longer gives a single leak (before 16 bytes were definitely lost, now 0, woohoo ;))

Last but not least, fixed a compilation issue with GCC 6.3 (ARM64) erroring on ambiguous std::abs in EnOceanESP2/3 by including cmath. Always love it when different compilers handle things differently :P

jvandenbroek:
Using shared_ptr again for eventsystem threads, calling reset is the correct way of handling it. Changed 1WireKernel stop logic, there is no interrupt for std::thread which it used before. Now using m_stoprequested as all other hardware does.

gizmocuz:
Thanks for the great work ! Question, can the event system be started/stopped again via the settings since you moved back to the shared_ptr in the eventsystem ?

jvandenbroek:
Yes that's no problem, the reset call will delete the object and set the ptr refcount to 0. Could also use unique_ptr in this case, but I like consistency.

gizmocuz:
Many Thanks!!!

gizmocuz:
Maybe we should call 'reset' after the join ? Or indeed always check if we can join

Edit: working on this
Edit: Done

PR #2545, opened 2018-07-08
[Linux] GetSerialPorts return /dev/ttySx

marcusbirkin:
GetSerialPorts didn't return UART devices labelled /dev/ttyS0 under Linux, unless operating under certain situations on a RPI.


gizmocuz:
Thanks, but this has been discussed before. This will cause (on some systems) to add about 20 serial ports that are not even used/connected.
Please see our wiki about "persistent devices"

Issue #2549, opened 2018-07-09
Crash in ~CHarmonyHub() 


pierrenicolasr:
Version 4.9769

{code}
2018-07-09 15:21:35.075  Error: Thread 11 (Thread 0x7fc34cdf2700 (LWP 2827)):
2018-07-09 15:21:35.075  Error: #0  0x00007fc35926623a in __waitpid (pid=4241, stat_loc=0x7fc34cdf040c, options=0) at ../sysdeps/unix/sysv/linux/waitpid.c:30
2018-07-09 15:21:35.075  Error: #1  0x00005568f4f3c992 in dumpstack_gdb() ()
2018-07-09 15:21:35.076  Error: #2  0x00005568f4f3cdce in signal_handler(int, siginfo_t*, void*) ()
2018-07-09 15:21:35.076  Error: #3  <signal handler called>
2018-07-09 15:21:35.076  Error: #4  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
2018-07-09 15:21:35.076  Error: #5  0x00007fc358901801 in __GI_abort () at abort.c:79
2018-07-09 15:21:35.076  Error: #6  0x00005568f5716375 in __gnu_cxx::__verbose_terminate_handler() ()
2018-07-09 15:21:35.076  Error: #7  0x00005568f5716196 in __cxxabiv1::__terminate(void (*)()) ()
2018-07-09 15:21:35.076  Error: #8  0x00005568f57211c9 in __cxa_call_terminate ()
2018-07-09 15:21:35.076  Error: #9  0x00005568f5720d48 in __gxx_personality_v0 ()
2018-07-09 15:21:35.076  Error: #10 0x00005568f57bbfa3 in _Unwind_RaiseException_Phase2 ()
2018-07-09 15:21:35.076  Error: #11 0x00005568f57bc7bf in _Unwind_RaiseException ()
2018-07-09 15:21:35.076  Error: #12 0x00005568f5716717 in __cxa_throw ()
2018-07-09 15:21:35.076  Error: #13 0x00005568f579cc3e in std::__throw_system_error(int) ()
2018-07-09 15:21:35.076  Error: #14 0x00005568f574435c in std::thread::join() ()
2018-07-09 15:21:35.076  Error: #15 0x00005568f51b4cef in CHarmonyHub::~CHarmonyHub() ()
[...]

PR #2550, opened 2018-07-09
Fix crash - ensures we don't join twice the thread

pierrenicolasr:
See https://github.com/domoticz/domoticz/issues/2549 - Makes sure we don't "join" the thread twice.


gizmocuz:
Thanks for the patch !
There was a lot more for other classes that needed to be done.
Made some changes here:

https://github.com/domoticz/domoticz/commit/c12ea2fba7c778dff2f2d02119c4b816a7e18992

Will be merged when the build systems passes compiling asap

Issue #2540, opened 2018-07-04
Meter readings and Graphs


jonferreira:
following the discussion at https://github.com/domoticz/domoticz/commit/9cfe88a6ced8d62de6206ae76306552fb9e87c16

The way graphs are implemented, if N is lower than N-1, graph will go crazy.

I reckon there should be a check such as

if ( n > (n-1)) {
value = n - (n-1)
else
value = n

gizmocuz:
Correct, but HOW can a counter value be lower then a previous value ? a counter is ALWAYS incremental

jonferreira:
Theoretically: yes. Real world: no. 
Take the RFXMeter / WPC3 counter, if power goes out the pulse count is reset. Same happens if the pulse count max out (think it's a 32bit number)
Should be a small fix that won't have any negative impact and should ensure the graphs don't get all messed up when / if that happens. 
Honestly, right now it's the one thing I think Dominicz needs improving :)

Edited: Removed email stuff

gizmocuz:
Okey, please make a PR

jonferreira:
I would if I could, unfortunately I can't seem to get to grips with domo development :(

I'd say the necessary edit would be at WebServer.cpp@14205

changing
`
float TotalValue = float(actValue - ulFirstValue);`

to 

```
if(actValue > ulFirstValue) {
float TotalValue = float(actValue - ulFirstValue);
} else {
float TotalValue = float(actValue);
}
```

I may be wrong though...

gizmocuz:
@jonferreira , what problems are you struggling with to compile Domoticz ?
The easiest would be visual studio maybe ? Very good debugging/breakpoint options
I think you are close with the fix, but don't want untested code....

jonferreira:
No problem **compiling**,  just me being a noob and struggling to get started :)

gizmocuz:
You could try these instructions:

https://www.domoticz.com/wiki/Raspberry_Pi_-_Build_Domoticz_from_source

It could be that you need to patch 'ulFirstValue' as well

So two options, your sollution:

`float TotalValue = (actValue > ulFirstValue) ? float(actValue - ulFirstValue) : actValue`

or

```
float TotalValue;
if(actValue > ulFirstValue) {
        TotalValue = float(actValue - ulFirstValue);
} else {
        TotalValue = float(actValue);
        ulFirstValue = actValue;
}
```

jonferreira:
[...]

PR #2551, opened 2018-07-09
Update WebServer.cpp

jonferreira:
following #2540 this prevents graph from going crazy if the meter counter resets


gizmocuz:
Thanks!!

PR #2557, opened 2018-07-10
 Fixed temperature sorting in report

Olifant1990:
Due to the degrees symbol after the numbers, sorting asc and desc happend text-wise. By removing all symbols, sorting works as number value.
Put the symbols in the column headers


Olifant1990:
Somthing is going wrong in my repo. Will check it soon! :(

Olifant1990:
#2558 Is now the correct one

PR #2561, opened 2018-07-14
Safe HW shutdown

emontnemery:
In some HW types relying on worker threads, there there are currently timeouts waiting for the worker threads to finish when stopping or restarting the HW. If this timeout expires, Domoticz simply soldiers on.
This **will** cause the worker threads to use freed memory in case they are not truly stuck, which causes crashes and other undefined behavior.

This change:
- Increases the timeout such that a blocking call for exampleto resolve an address or do some network related operation should timeout
- Will instead print an error and then call abort() in case the timeout expires, to allow Domoticz to be restarted, e.g. by systemd or by Windows service management.

### DIFF
--- a/hardware/Comm5Serial.cpp
+++ b/hardware/Comm5Serial.cpp
@@ -105,6 +105,7 @@ bool Comm5Serial::StopHardware()
 	{
 		m_thread->join();
 		// Wait a while. The read thread might be reading. Adding this prevents a pointer error in the async serial class.
+		// TODO: Add proper synchronized shutdown of read thread instead of fixed timeout
 		sleep_milliseconds(10);
 		m_thread.reset();
 	}
--- a/hardware/CurrentCostMeterSerial.cpp
+++ b/hardware/CurrentCostMeterSerial.cpp
@@ -76,6 +76,7 @@ bool CurrentCostMeterSerial::StopHardware()
 	{
 		m_thread->join();
 		// Wait a while. The read thread might be reading. Adding this prevents a pointer error in the async serial class.
+		// TODO: Add proper synchronized shutdown of read thread instead of fixed timeout
 		sleep_milliseconds(10);
 		m_thread.reset();
 	}
--- a/hardware/DavisLoggerSerial.cpp
+++ b/hardware/DavisLoggerSerial.cpp
@@ -61,6 +61,7 @@ bool CDavisLoggerSerial::StopHardware()
 		m_thread.reset();
 	}
 	// Wait a while. The read thread might be reading. Adding this prevents a pointer error in the async serial class.
+	// TODO: Add proper synchronized shutdown of read thread instead of fixed timeout
 	sleep_milliseconds(10);
 	terminate();
 	m_bIsStarted = false;
--- a/hardware/DomoticzHardware.cpp
+++ b/hardware/DomoticzHardware.cpp
@@ -90,6 +90,7 @@ void CDomoticzHardwareBase::StopHeartbeatThread()
 	{
 		m_Heartbeatthread->join();
 		// Wait a while. The read thread might be reading. Adding this prevents a pointer error in the async serial class.
+		// TODO: Add proper synchronized shutdown of read thread instead of fixed timeout
 		sleep_milliseconds(10);
 		m_Heartbeatthread.reset();
 	}
--- a/hardware/EnOceanESP2.cpp
+++ b/hardware/EnOceanESP2.cpp
@@ -658,6 +658,7 @@ bool CEnOceanESP2::StopHardware()
 	{
 		m_thread->join();
 		// Wait a while. The read thread might be reading. Adding this prevents a pointer error in the async serial class.
+		// TODO: Add proper synchronized shutdown of read thread instead of fixed timeout
 		sleep_milliseconds(10);
 		m_thread.reset();
 	}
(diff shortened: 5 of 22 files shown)


gizmocuz:
Thank you for the pull request.
I was also looking at this, and i don't like the 'sleeps' that much, and wanted to do this

https://thispointer.com/c11-how-to-stop-or-terminate-a-thread/

How about this ? We could add part of the needed variables in the base class, so it should not be to hard to implement this in every class

emontnemery:
@gizmocuz The real root issue is that the worker thread might be waiting on some synchronous I/O which in case of TCP/IP might take in the order of minutes to time out.

The purpose of this PR is not to fix the HW implementation, but to remove undefined behavior which causes hard to debug crashes.

What could be done to improve the likelyhood of waiting is to either make sure the HW implementation is using I/O methods which can be safely interrupted when needed, for example using boost:
https://www.boost.org/doc/libs/1_67_0/doc/html/thread/thread_management.html, see section "interruption", or by using asynchronous I/O.

The suggested scheme at https://thispointer.com/c11-how-to-stop-or-terminate-a-thread/ is just a slightly more complicated version of the `m_stoprequested` member variable used in Domoticz, isn't it? You **still** need to wait for the worker threads to finish, from the example:
```
std::cout << "Asking Task to Stop" << std::endl;
// Stop the Task
task.stop();
 
//Blocking wait for thread to join, might never return
th.join();
```
And, since th.join() is synchronous and doesn't time out, you will end up with something like:
```
// Stop the Task
task.stop();
 
//Waiting for thread to join
while ( t.join_for(boost::chrono::milliseconds(500)) && retry_cntr++ < 120)
{
    // Failed to join, give up
    abort();
}
```
And this is essentially the same as the current implementation..

gizmocuz:
I think it is a much better improvement. But the join will wait until the worker returns, so i don't see the problem.
I see a problem that you (and maybe more) added sleeps, that is not a good solution and should not be needed and avoided.
Join does not return until the Do_Work function terminates

emontnemery:
@gizmocuz OK, but if the wait() are removed, there has to be another way of detecting deadlocks when shutting down HW, i.e. a watchdog/heartbeat for Domoticz' main thread. Do you think that would be a better solution?

By the way, the link you point to will give 100% the same behavior as simply removing the timeout from existing code (example form hardware/Pinger.cpp):

```
//Make sure all our background workers are stopped
int iRetryCounter = 0;
while ((m_iThreadsRunning > 0) && (iRetryCounter < 15))
{
	sleep_milliseconds(500);
	iRetryCounter++;
}
```

to:
```
while (m_iThreadsRunning > 0)
{
	iRetryCounter++;
}
```

gizmocuz:
@emontnemery , i know, but it looks cleaner this way.
It is the hardware's responsibility that the shutdown functions properly.
So if it get's a shutdown request, then it should shutdown (in the Worker thread)
That is also the reason why i don't like a simple Sleep (1000) or Sleep (10000) but prefer a wait of 100ms or 500ms at the most
If a serial port is closed, it should not cause errors. If it does, instead of trying to fix it in 100 classes with a possible sleep (no comment to you of course... !!) then it should be solved (if there is a problem) in the serial class.
A thread 'join' should not have to call wait, as it blocks the current thread until the thread finishes its execution
I also never had an issue with a serial port class, so it could be one class you have issues with ?

gizmocuz:
if workers keep their sleeps small, then 15*500 = 7.5 seconds, that is a long time...
For instance, if you have a blocking socket, close the socket, it will break out of the blocking call.
Sure that needs to be in another thread, and thats why we have the AyncTCP class

emontnemery:
@gizmocuz 
Regarding hang in worker thread:
I agree for native HW implementation, it is the HW responsibility to ensure no long sycnhronous call is done. For example use only asynchronous I/O methods.
Python plugins is different though, how to guarantee nice behavior of Python plugin which are not even in domoticz source tree?

Maybe Domoticz' Python framework should print error message and automatically disable a Python plugin which (repeatedly?) needs long time to execute??

Anyhow, I'm fine to close this PR and replace with a new PR where the timeout is replaced with some sort of SW watchdog if you think that's better.

Regarding issue with serial port:
```
// Wait a while. The read thread might be reading. Adding this prevents a pointer error in the async serial class.
// TODO: Add proper synchronized shutdown of read thread instead of fixed timeout
sleep_milliseconds(10);
```
I just added the TODO comment because the 10 ms sleep seems extremely dodgy..

gizmocuz:
Thanks!. But i see this 10 ms sleep and comment in quite some source files now... 
Python threads can be closed differently.
But do you have a problem with a certain native hardware class ?

emontnemery:
> do you have a problem with a certain native hardware class

Yes!! I've seen crashes caused by the unsafe timeout with Python and Kodi hardware classes.
Because of those crashes, I went through all hardware classes to search for the same design pattern, and it turns out the same timeout is used in: Kodi, LogitechMediaServer, PanasonicTV, Pinger, Python


When I went through all the hardware classes, I also noticed the 10 ms sleep for hardware classes using the serial port.
Do you know/remember why it was added?

gizmocuz:
@emontnemery , no sorry, i have to look into the history for this. But it should not be needed.
For instance, the RFXCom classes use ASyncSerial as well  (And TCP) without issues
I think it is a matter of closing ports in the thread, making sure no calls can be made etc..

emontnemery:
I will replace this with a new PR soon

PR #2575, opened 2018-07-18
Sw watchdog

emontnemery:
Replaces #2561 
- Add SW watchdog functionality, triggered by lack of heartbeat from: main thread, main worker thread, and any thread registered by MainWorker::HeartbeatUpdate, including webservers
- Remove unsafe timeouts when trying to shut down HW, instead rely on watchdog functionality
- If watchdog barks, print stack trace for all threads (to aid debugging of deadlocks), then exit domoticz process


gizmocuz:
Thank you for the patch. Would this not cause a stop/abort/signal when one thread is not working correctly? Meaning, one bad class will cause everything to stop ? Can't that thread be restarted ? (Like we do with the data-timeout option)

emontnemery:
@gizmocuz Yeah, the purpose is exactly to:
- Print a detailed error message to aid debugging of a deadlock
- Raise a signal to terminate Domoticz
- Let the OS (e.g. systemd, init, Windows service framework) safely restart Domoticz

The reason is there is no way to safely restart a thread without risking stability of the process.

gizmocuz:
If a thread is in a deadlock phase, the user can use the data-timeout option that will restart the thread (or a new instance)
With your patch, if a dev made a error, instead of writing this deadlock in the error log, you stop domoticz.
That is not desired., what happens if the process can not be restarted because of a open database handle, or other veague reason ?
I think it is better to send a notification or a error log message

emontnemery:
> If a thread is in a deadlock phase, the user can use the data-timeout option that will restart the thread (or a new instance)

No, it is **not possible** to safely restart a thread, that's the point of the patch!
Of course, the normal case would not be to restart Domoticz just because something is slow, this is an absolute last resort if a critical thread (web server, mainworker, etc.) has been stuck for more than a minute.

_what happens if the process can not be restarted because of a open database handle_
Do you mean what happens if domoticz is writing to the database when the signal is triggered? There could be risk of database corruption?

gizmocuz:
Ahh okey, so this patch is ONLY for very critical threads, not for all hardware classes then ?

About the database, well i noticed a lot, when a database is open (which is the case when domoticz is running), you can not access/preform actions to this database with other applications (another instance)
If domoticz is stopped not cleanly, and the database is not closed, there might be some data loss as well.

emontnemery:
> Ahh okey, so this patch is ONLY for very critical threads, not for all hardware classes then ?

Yes, right.
To make it more clear in the code, MainWorker::HeartbeatUpdate can be updated with an additional parameter "Critical", and only if that parameter is set Domoticz will restart if the process is stuck.

For the potential database issue, I can try to update the signal handler to request the database to finish any pending operation and not start any new ones to make that less likely.

gizmocuz:
That would be a good idea, i just added a CloseDatabase function to SQLHelper.cpp for that, that you should be able to call.

Some more little things i found:

Why the new mutex in the time function, this is not needed there (see the time function)
(std::unique_lock<std::mutex> lock(TimeMutex2_);)

m_LastHeartbeat does to have to be atomic, this is also not the case for the normal hardware classes.

I also would initialize this the normal way, as it is not an array/structure

time_t m_LastHeartbeat = 0;

emontnemery:
OK, I will try to update with using the new CloseDatabase function.


Also:
> Why the new mutex in the time function, this is not needed there (see the time function)
> (std::unique_lockstd::mutex lock(TimeMutex2_);)

Yes, it is needed to synchronize update of m_lasttime. However, it is not related to the watchdog functionality, I'll move this change to #2598 instaed.

>m_LastHeartbeat does to have to be atomic, this is also not the case for the normal hardware classes.

Yes, access to it needs to be synchronized, e.g. with std::atomic because it is accessed by multiple threads. Please see similar changes in #2598 

> I also would initialize this the normal way, as it is not an array/structure
> time_t m_LastHeartbeat = 0;

OK, good point

gizmocuz:
Still i see no reason for the new mutex in the time function, are you really saying that 1 second could give a problem ?

I see no atomic discussion in '2598', this is also not used in the normal hardware classes.
There 'should' also be no problem that multiple threads reading this value, only 1 is writing it
Next, as above, 1 second could not make a difference
It's an expensive declaration, and in this case not needed (as is the mutex)

emontnemery:
My concern is not 1 second delta, that's of course not an issue..
My concern is that there's no guarantee that a write of an integer is atomic, hence comparing m_LastHeartbeat or m_lasttime might give unexpected results if there is a thread switch in the middle of a write to m_LastHeartbeat or m_lasttime.
(m_lasttime in r_localtime can be made std::atomic instead of protected by mutex, std::atomic is supposed to be more lightweight)

Edit: By the way, I'm not confident about calling CloseDatabase() from the signal handler. There's a risk this deadlocks if the reason we're in the signal handler in the first place is a crash or deadlock while m_sqlQueryMutex is held.

gizmocuz:
Okey better not use the CloseDatabase in the exit code then.

About the atomic, there is one thread only that 'writes' to the variable, and one other that 'reads' from it.
We are talking about an integer type, not an object, and it is a simple write (a=b)

It is best if you write a test program to see whats going on (or prove that it goes wrong)

Sure i believe that "maybe" "sometimes" you "might" get a **stale** value when you try reading the value when at that exact time the thread writes the value, that is why i say, 1 second difference (at the most)

I believe in test applications that proves me wrong ;)

emontnemery:
Please have a look here:
https://stackoverflow.com/questions/54188/are-c-reads-and-writes-of-an-int-atomic

Updating an integer is NOT guaranteed to be atomic.

By using std::atomic for variables that are shared between threads, we are protecting the variable from being written to and read from at the same time with undefined results. I think this is common sense..
Please note std::atomic is not very heavy or expensive..

emontnemery:
@gizmocuz I tried to fix the review comments, can this be merged now?

Let's continue the discussion about std::atomic in #2626 :)

gizmocuz:
Patch looks good, i only have one request (when possible)
Would it be OK if you created a watchdog class cpp/h file, where all watchdog stuff, including signal handling resided in ?
This way, the domoticz.cpp keeps clean, it is getting messy, AND for each change in the domoticz repo, this file has to be re-compiled, so it takes longer

IsWSL
printRegInfo
printSingleThreadInfo
printSingleCallStack
heartbeat_check
Signal handling
watchdog handling

This way it is also quickly accessible in the feature

emontnemery:
Sure, that's very reasonable!
Moved signalhandler, bactrace generation and watchdog to own file (SignalHandler.cpp)

gizmocuz:
Thanks ! Great patch !!

gizmocuz:
This is not working well while debugging and setting breakpoints, and taking more then 60 seconds to debug!

gizmocuz:
https://github.com/domoticz/domoticz/commit/1b2f7a6c1cee14ec21019a893a35d015584531e6

gizmocuz:
Locking two muxtex in HeartbeatCheck could be dangerous...

gizmocuz:
I imagine that in the webserver, server_base::heart_beat, is called in a separate thread all the time.
So if the server hangs for other reasons (invalid parsing code or...) this server_base::heart_beat will keep going ?

gizmocuz:
@emontnemery , sorry to say, but this breaks my 64bit arm compile system

> [ 21%] Building CXX object CMakeFiles/domoticz.dir/main/SignalHandler.cpp.o
> /home/pi/dev-domoticz/main/SignalHandler.cpp: In function 'bool printRegInfo(siginfo_t*, ucontext_t*)':
> /home/pi/dev-domoticz/main/SignalHandler.cpp:71:109: error: 'mcontext_t {aka struct sigcontext}' has no member named 'arm_lr'
>   _log.Log(LOG_ERROR, "siginfo address=%p, address=%p", info->si_addr, ((ucontext_t *)ucontext)->uc_mcontext.arm_lr);
>                                                                                                              ^~~~~~
> CMakeFiles/domoticz.dir/build.make:283: recipe for target 'CMakeFiles/domoticz.dir/main/SignalHandler.cpp.o' failed
> make[2]: *** [CMakeFiles/domoticz.dir/main/SignalHandler.cpp.o] Error 1
> CMakeFiles/Makefile2:104: recipe for target 'CMakeFiles/domoticz.dir/all' failed
> make[1]: *** [CMakeFiles/domoticz.dir/all] Error 2
> Makefile:151: recipe for target 'all' failed
>

gizmocuz:
Maybe related to

https://github.com/torvalds/linux/blob/master/arch/arm64/include/uapi/asm/sigcontext.h

emontnemery:
> This is not working well while debugging and setting breakpoints, and taking more then 60 seconds to debug

Good point, but instead of compile time flag (1b2f7a6), try calling IsDebuggerPresent() in SignalHandler.cpp befor triggering watchdog.
Should I open PR for such change?

> I imagine that in the webserver, server_base::heart_beat, is called in a separate thread all the time

No, I don't think it is called from separate thread, look at server_base::server_base() constructor
I'll double check though, just to make sure.

> this breaks my 64bit arm compile system

Ooops! I tried to fix in #2638

Jumbotroll:
#2638 fixed #2637 build error for Synology arch-rtd1296 and arch-qoriq.

gizmocuz:
IsDebuggerPresent is not working for win32 at the moment i am afraid... i had put them in #ifdef now

Issue #2576, opened 2018-07-18
Seems that Build is broken with gcc 8.1.1 on Fedora


pipiche38:
When building domoticz, I got a compilation error 

See here after a recommendation I found

https://stackoverflow.com/questions/20759750/resolving-redefinition-of-timespec-in-time-h

```

make
[  1%] Built target minizip
[  8%] Built target mqtt
[  9%] Built target sqlite
[ 20%] Built target lua
-- Found Git: /usr/bin/git (found version "2.17.1")
-- domoticz has been modified locally: adding "-modified" to hash
[ 20%] Built target revisiontag
[ 21%] Built target domoticz_gch
[ 22%] Building CXX object CMakeFiles/domoticz.dir/main/Camera.cpp.o
In file included from /usr/include/sys/stat.h:30,
                 from /usr/local/include/boost/asio/detail/impl/eventfd_select_interrupter.ipp:23,
                 from /usr/local/include/boost/asio/detail/eventfd_select_interrupter.hpp:80,
                 from /usr/local/include/boost/asio/detail/select_interrupter.hpp:25,
                 from /usr/local/include/boost/asio/detail/epoll_reactor.hpp:28,
                 from /usr/local/include/boost/asio/detail/reactor.hpp:21,
                 from /usr/local/include/boost/asio/detail/impl/scheduler.ipp:23,
                 from /usr/local/include/boost/asio/detail/scheduler.hpp:212,
                 from /usr/local/include/boost/asio/system_context.hpp:19,
                 from /usr/local/include/boost/asio/impl/system_executor.hpp:22,
                 from /usr/local/include/boost/asio/system_executor.hpp:129,
                 from /usr/local/include/boost/asio/associated_executor.hpp:21,
                 from /usr/local/include/boost/asio.hpp:21,
                 from /usr/share/domoticz/beta-domoticz/main/../tcpserver/TCPClient.h:4,
                 from /usr/share/domoticz/beta-domoticz/main/../tcpserver/TCPServer.h:4,
                 from /usr/share/domoticz/beta-domoticz/main/mainworker.h:10,
                 from /usr/share/domoticz/beta-domoticz/main/Camera.cpp:7:
/usr/include/bits/types/struct_timespec.h:9:8: error: redefinition of ‘struct timespec’
 struct timespec
        ^~~~~~~~
In file included from /usr/include/sched.h:32,
                 from /usr/include/pthread.h:23,
[...]

PR #2579, opened 2018-07-19
redefinition of ‘struct timespec’  struct timespec when Building with…

pipiche38:
Fixing G++ issue at least on Linux Fedora distribution.

Fixing #2576

It has been tested with gcc-c++ version 8.1.1 on fedora 28 hosted on armv7 (RPI3B+)


gizmocuz:
Thanks for the PR. After reading the issue, i don't think this is needed anymore ?

gizmocuz:
Closing PR, feel free to re-open if needed

PR #2593, opened 2018-07-26
Change MQTT idx to Integer for scenes/groups inconsistency

sweetpants:
IDX is published as string while integer for other devices.


gizmocuz:
Thanks!!

PR #2596, opened 2018-07-27
Added native support for multiple Xiaomi gateways without any restrictions

JobsIsAlive:
Added native support for multiple Xiaomi gateways. 
You can simple add, remove and modify settings for Xiaomi gateway hardware without any restrictions. 
Devices tied to the gateways will be managed and displayed correctly in the Domoticz. .


kovainfo:
Cool! With this patch no longer need to follow the instructions: https://www.domoticz.com/wiki/Xiaomi_Gateway_(Aqara)#Multiple_gatevays

corbinmunce:
this is awesome!

gizmocuz:
Thank you for the patch! Could you make the new functions a member of the XiaomiGateway:: class ?
GatewayByIp
AddGatewayToList
RemoveFromGatewayList

Whiskey24:
Any chance this will be picked up? I'd very much like to get #2609 merged ;-)

JobsIsAlive:
I will do all after 1 sept. I don't understand why this functions must be in xiaomi class ( this functions don't implemented class functionality just list of that class members), but I will do it ( static functions and static list ).

Whiskey24:
Super, thanks!

gizmocuz:
@JobsIsAlive , those functions are only used in this hardware class, so they must be a member. No need for static functions
Thanks for your consideration

JobsIsAlive:
all done

gizmocuz:
Great thanks !!
It's perfect now... just two suggestions:

StartHardware, just use

```
if (result.empty())
  return false;
```

No need for the {} makes it a bit clearer...

In the feature (i think I will do this, but anyone can beat me) i will make a ASyncUDP class... but for now, your solution is perfect.
Only i think it is possible to create 1 time the UDP socket correct? Now it is created allways in the sendmessage
It's probably working now... but could be a little bit cleaner

Just my two cents

JobsIsAlive:
Maybe you rigth about UDP, but please write it to MI developers, becouse they dont indexed request and if you use only one UDP port you cannot understand on that command you receive answer. I think in that case developer of that XiaomiGateway class open port in SendMessage, see . Its my 1 cent.

kovainfo:
Xiaomi gateway uses multicast packets over UDP and current release of code is correct for this gateway. In case, when system have many gateways, Xiaomi class must workhard with many multicast packets. @corbinmunce

JobsIsAlive:
Or you need to do list of commands and send its in SYNC mode, but in this case you dont will turn on 2 lamp in one time )) And if command take much time all other commands will waiting when it ends. I think its not right

gizmocuz:
Okey, let me know when you want this merged... i see you are still pushing changes

JobsIsAlive:
i just change "if (result.empty())return false;", you ask for this i think. Now all done

gizmocuz:
Thanks !!

PR #2598, opened 2018-07-28
Fix various data races

### LINKED ISSUE #2541: Signal 11 crash
Version: 4.9741 (But has been happening on previous betas)
Platform: Ras Pi 3B
Plugin/Hardware: No Hardware. Weather/Hue/Tado/Harmony Hub/MQTT Client Gateway/Sonoff MQTT discovery

I'm seeing fairly regular crashes in the latest beta, although they've been happening for a week or two now going back through previous beta versions. Happens roughly once every day our two. Restarts cleanly afterwards. Looks like it might be related to the Sonoff MQTT Discovery plugin as the crash happens so soon afterwards? 

This is what gets dumped into the logs

2018-07-04 09:45:14.486 (Sonoff) Heartbeating...
2018-07-04 09:45:14.486 (Sonoff) Reconnecting
2018-07-04 09:45:14.486 (Sonoff) MqttClient::Open
2018-07-04 09:45:14.537 (Sonoff) MqttClient::onConnect
2018-07-04 09:45:14.537 (Sonoff) Successful connect to: 192.168.200.25:1883
2018-07-04 09:45:14.538 (Sonoff) MqttClient::Connect
2018-07-04 09:45:14.538 Error: Domoticz received fatal signal 11 !...
2018-07-04 09:45:14.541 Error: /home/pi/domoticz/domoticz() [0x261f54]
2018-07-04 09:45:14.541 Error: /home/pi/domoticz/domoticz(_Z14signal_handleri+0x4c) [0x26200c]
2018-07-04 09:45:14.542 Error: /lib/arm-linux-gnueabihf/libc.so.6(__default_sa_restorer+0) [0x76c8c6b0]

emontnemery:
Fix various data races, mainly in the Plugin framework.
This fixes #2541

### DIFF
--- a/hardware/DomoticzHardware.h
+++ b/hardware/DomoticzHardware.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <atomic>
 #include <boost/signals2.hpp>
 #include "../main/RFXNames.h"
 
--- a/hardware/DomoticzHardware.h
+++ b/hardware/DomoticzHardware.h
@@ -100,7 +100,7 @@ protected:
 	//Barometric calculation (only for 1 sensor per hardware device!)
 	int CalculateBaroForecast(const double pressure);
 
-	bool m_bIsStarted = { false };
+	std::atomic<bool> m_bIsStarted = { false };
 
 private:
     void Do_Heartbeat_Work();
--- a/hardware/plugins/PluginManager.cpp
+++ b/hardware/plugins/PluginManager.cpp
@@ -260,31 +260,12 @@ namespace Plugins {
 
 		// Create initial IO Service thread
 		ios.reset();
+		// Create some work to keep IO Service alive
+		auto work = boost::asio::io_service::work(ios);
 		boost::thread bt(boost::bind(&boost::asio::io_service::run, &ios));
 
 		while (!m_stoprequested)
 		{
-			if (ios.stopped())  // make sure that there is a boost thread to service i/o operations if there are any transports that need it
-			{
-				bool bIos_required = false;
-				for (std::map<int, CDomoticzHardwareBase*>::iterator itt = m_pPlugins.begin(); itt != m_pPlugins.end(); ++itt)
-				{
-					CPlugin*	pPlugin = reinterpret_cast<CPlugin*>(itt->second);
-					if (pPlugin && pPlugin->IoThreadRequired())
-					{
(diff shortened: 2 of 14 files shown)


dnpwwo:
Not sure why you are guarding a Connection object with a mutex when it is a Python object.

If you want to guard against it being released you should INCREF it at the start of the I/O operation and DECREF once the event has been queued. The Python GC won't respect a mutex.

emontnemery:
@dnpwwo The problem is that the Python Connection object is being accessed by the async IO threads via handleRead functions and so on. INCREF and DECREF are not thread safe, so reference counting can't be trusted without a lock.
My idea was to reuse the existing Python mutex which is already taken before calling the Python interpreter to ensure the Python interpreter is not running when messing with the Connection object.

dnpwwo:
But the crashes that @andyfitter has been having look they are because of the object being deleted rather than referenced.  As long as the object still exists, Python and the other code should be fine. 

Perhaps a review is required over Connection object lifecycle management overall is required because if the ref count goes to zero the Python garbage collector will clean up regardless of the mutex.

I must say I do like the boost::asio::io_service::work change though, I wish I had seen that when I wrote the framework. Much more straightforward.

emontnemery:
> But the crashes that @andyfitter has been having look they are because of the object being deleted rather than referenced. As long as the object still exists, Python and the other code should be fine.

Right, but since INCREF and DECREF are not thread safe, this could very well be the symptom when the async IO thread was touching the connection object without synchronization, right? For example, it could very well happen that an INCREF has no effect if a DECREF is happening at the same time, for example like so:
```
ref_ctr = 2
t1: call inc_ctr()
t1:   load ref_ctr
t2: call dec_ctr()
t2:   load ref_ctr
t1:   ref_ctr += 1
t1:   store ref_ctr = 3
t2:   ref_ctr -= 1
t2:   store ref_ctr = 1
```

emontnemery:
@gizmocuz OK to merge this? Do you need me to rebase since there are now some merge conflicts?
If you want, I can split the PR in two: one to fix only the plugin framework races as reported in #2541, and another to fix the various other races.

gizmocuz:
Yes please rebase, but i do not see the use for atomic yet again. But i am going to rebuild all threads to use the signal way to close as described before.

PR #2603, opened 2018-07-30
Updated GoodweAPI to use new API with additional data

llagendijk:
I updated GoodweAPI to use a new API that provides additional measurements  (Mains voltage/current per phase and string voltages/ current (per string)


gizmocuz:
Sorry, you need to make a PR against the development branch, not the master branch

PR #2606, opened 2018-08-02
Adding new hardware support for Denkovi plus improvements

dzelyaydinov:


gizmocuz:
Is it possible to make 1 hardware class for 'Denkovi' as we now have about 5 of them ?
Maybe initialize the constructor with a 'type'.
This way it is a lot clearer i hope

dzelyaydinov:
Hello Rob. 
Can we remove the older classes (written by you) and just leave the three we wrote? It is better experience for the users and for us to separate the devices in different hardware classes as they are separated by type of communication. We have as you can see more than 10 devices that currently are supported by Domoticz and we plan to support more in the near future. I hope you understand our willing to make more user friendly interface. 
Also can you please help with resolving the issue with pull request?

gizmocuz:
That is of course also an option. Please label them TTYPE_RESERVED_FOR_YOU_XXX (where XXX is 1,2 and 3), and of course some code in the webserver/mainworker as well.

dzelyaydinov:
I made changes and want to edit SQLHelper.cpp and remove old classes files but I do not see how it can be done without opening a new pull request. Please help me resolving issues. Thanks in advance.

gizmocuz:
@dzelyaydinov , not sure i understand, why can't you work on SQLHelper.cpp with this patch ?

dzelyaydinov:
Ok I was wrong. It is edited now. So can you help resolving issues? Thanks again.

gizmocuz:
It should be resolved now, don't forget to do git pull if you continue development on this PR

gizmocuz:
Thank you for the PR, i have a question and a wish ...

Could you undo the formatting of in the file RFXNames.cpp

There are now 3 Denkovi devices:

HTYPE_DenkoviDevices
HTYPE_DenkoviUSBDevices
HTYPE_DenkoviTCPDevices

What is 'HTYPE_DenkoviDevices' for device ? Is the USB or TCP ?

dzelyaydinov:
Thank you for the support again. _HTYPE_DenkoviDevices_ are actually HTTP communicating devices. So if it is ok for you we could make it _HTYPE_DenkoviHTTPDevices_. About undo formatting you may be want to make *"Denkovi Modules with LAN/WiFi (HTTP) Interface"* to be *"Denkovi Modules with LAN (HTTP) Interface"*. Is this right? If it is right, that's no problem for us.

gizmocuz:
That sound perfect!

dzelyaydinov:
Is there something else that we can do or some other issue with the pull request? When you plan to merge the pull request? Thanks in advance.

gizmocuz:
Whey you say it'r ready .... so that will be now ;), thanks for the PR!

gizmocuz:
There are quite a few compile warnings in this commit.. most of them i will resolve now
But what is this function 'copyBuffer' do ? Would a memcpy not work ?
I assume everything is text and not binary correct? if that is the case, please work with std::string (append/insert etc), as this is dangerous code

gizmocuz:
Could you also rename ALL member variables of the class with a m_ prefix ? like m_receivedDataLength

dzelyaydinov:
Ok, I will edit these problems and will make a new pull request. Is it OK?

gizmocuz:
You can continue pushing changes to this one

gizmocuz:
(Or create a new one, because you closed this)

PR #2609, opened 2018-08-03
Xiaomi get local IP fallback: parsing from ifconfig

Whiskey24:
A fallback to determine the Domoticz local IP address for the Xiaomi gateway in case boost asio does not work. This was the case for me using an iocage FreeNas jail (based on FreeBSD), but others seem to encounter this problem as well, e.g. with NAS4Free (http://www.domoticz.com/forum/viewtopic.php?f=47&t=22813).

The fallback parses the output from ifconfig (present on almost every linux system) to determine the local IP. It can be made more robust (for example by checking if 127.0.0.1 is not given as answer), but it does the job for me.

### DIFF
--- a/hardware/XiaomiGateway.cpp
+++ b/hardware/XiaomiGateway.cpp
@@ -678,16 +678,29 @@ void XiaomiGateway::Do_Work()
 		try {
 	
 			// get output from ifconfig
-			std::string ifconfigStr;	
-			FILE * stream = popen("ifconfig", "r");
-			const int max_buffer = 256;
-    			char buffer[max_buffer];
+			std::string ifconfigStr;
+                        const int max_buffer = 256;
+                        char buffer[max_buffer];
 
-	    		if (stream) {
-  				while (!feof(stream))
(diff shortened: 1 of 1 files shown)


Whiskey24:
Anything I can do to do to progress this? The Travis-CI build failed, as far as I can see because it kept waiting for command line input: 

``patching file cpp/build/Makefile
Reversed (or previously applied) patch detected!  Assume -R? [n]``

``No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
The build has been terminated``

Whiskey24:
I have updated the patch so the fallback will only match an IP in the same network range as the Xiaomi gateway and changed the WIN32 section to use ipconfig. This should also make it work on Windows, but I've not tested this.

Whiskey24:
If #2596 is added, I can submit a new patch for that code base. Just let me know.

gizmocuz:
There are a few issues with the other patch that needs to be resolved before

Whiskey24:
Closed in favor of #2702

PR #2612, opened 2018-08-05
Single route for different device log types

stas-demydiuk:
Use single `/Log` route for different device log views instead of

- /TextLog
- /GraphLog
- /TemperatureLog
- /LightLog


gizmocuz:
Thanks! Great patch again ! :)

gizmocuz:
There might be a problem with the temperature log

Issue #2506, opened 2018-06-20
[RTL433] AlectoV1 Wind Sensor: Unhandled sensor reading


HansVanEijsden:
Hi all,

`rtl_433 -G` finds this (currently there's no wind though):

```
trying device  0:  Realtek, RTL2838UHIDIR, SN: 00000001
Found Rafael Micro R820T tuner
Using device 0: Generic RTL2832U OEM
Exact sample rate is: 250000.000414 Hz
[R82XX] PLL not locked!
Sample rate set to 250000.
Bit detection level set to 0 (Auto).
Tuner gain set to Auto.
Reading samples in async mode...
Tuned to 433920000 Hz.
2018-06-20 23:35:09 :	AlectoV1 Wind Sensor
	House Code:	 44
	Channel:	 1
	Battery:	 OK
	Wind speed:	 0.00 m/s
	Wind gust:	 0.00 m/s
	Direction:	 180
	Integrity:	 CHECKSUM
```

But Domoticz 3.9656 shows this:

> 2018-06-20 23:38:46.246 Status: Rtl433: Unhandled sensor reading, please report: (2018-06-20 23:38:46,AlectoV1 Wind Sensor,44,1,OK,,CHECKSUM,,,,,,,,,,,,,,,,,,,,0.000,0.000,180,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,
> 2018-06-20 23:38:46.246 )

All my other RF-sensors are detected by rtl_433 and parsed by Domoticz, but this wind sensor is unknown to Domoticz.
I hope you can add the sensor, with the above information. Thanks!

HansVanEijsden:
Some more readings, so you can see the values:

> 2018-06-21 02:03:58.460 Status: Rtl433: Unhandled sensor reading, please report: (2018-06-21 02:03:58,AlectoV1 Wind Sensor,44,1,OK,,CHECKSUM,,,,,,,,,,,,,,,,,,,,1.200,1.600,180,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,
2018-06-21 02:03:58.460 )
[...]

PR #2619, opened 2018-08-09
rtl_433: Add support for wind_speed wind_gust wind_direction

TimelessNL:
description:
We had support for windstrength and winddirection, based on #2506 there are wind sensors that use wind_speed/wind_gust & wind_direction.
For all we know these values cannot appear simultaneously so it's either:
windstrength vs wind_speed
winddirection vs wind_direction

Add support for:
wind_speed/wind_gust & wind_direction with or without temperature.

Tested on:
AlectoV1 Wind Sensor (wind_speed/wind_gust/wind_direction/!temperature)
TFA 30.3168 Solar Windmeter (windstrength/winddirection/temperature)

### DIFF
--- a/hardware/Rtl433.cpp
+++ b/hardware/Rtl433.cpp
@@ -194,6 +194,8 @@ void CRtl433::Do_Work()
 				bool hasdepth = false;
 				float wind_str;
 				bool haswind_str = false;
+				float wind_gst;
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thanks!

TimelessNL:
Quick response! Thanks!

Issue #2620, opened 2018-08-09
Beta 4.9811+ crashes straight on startup


heggink:
Just tried to update to 4.9812. Already saw https://www.domoticz.com/forum/viewtopic.php?f=6&t=24437&p=188284#p188284 mentioning the 4.9811 crash on startup. Same happens with 9812.

Aug  9 12:00:39 domoticz domoticz: Domoticz is starting up....
Aug  9 12:00:39 domoticz domoticz: Domoticz running...
Aug  9 12:00:39 domoticz domoticz: Sunrise: 06:11:00 SunSet: 21:17:00
Aug  9 12:00:39 domoticz domoticz: Day length: 15:06:00 Sun at south: 13:05:00
Aug  9 12:00:39 domoticz domoticz: Civil twilight start: 05:31:00 Civil twilight end: 21:56:00
Aug  9 12:00:39 domoticz domoticz: Nautical twilight start: 04:38:00 Nautical twilight end: 22:49:00
Aug  9 12:00:39 domoticz domoticz: Astronomical twilight start: 03:29:00 Astronomical twilight end: 23:59:00
Aug  9 12:00:39 domoticz domoticz: EventSystem: reset all events...
Aug  9 12:00:40 domoticz domoticz: PluginSystem: Started, Python version '3.5.3'.
Aug  9 12:00:40 domoticz domoticz: Domoticz received fatal signal 11 (Segmentation fault)
Aug  9 12:00:45 domoticz domoticz: Thread 1 (Thread 0x76f5c000 (LWP 21559)):
Aug  9 12:00:45 domoticz domoticz: #0  0x76e8b4cc in __waitpid (pid=21591, stat_loc=0x7ecbadb0, options=0) at ../sysdeps/unix/sysv/linux/waitpid.c:29
Aug  9 12:00:45 domoticz domoticz: #1  0x00228abc in dumpstack_gdb() ()
Aug  9 12:00:45 domoticz domoticz: #2  0x00228f1c in signal_handler(int, siginfo_t*, void*) ()
Aug  9 12:00:45 domoticz domoticz: #3  <signal handler called>
Aug  9 12:00:45 domoticz domoticz: #4  0x76e8cb80 in pthread_setname_np (th=0, name=0x7ecbb598 "AsyncSer_Backgr") at ../sysdeps/unix/sysv/linux/pthread_setname.c:47
[...]

PR #2623, opened 2018-08-09
Fix for #2620

emontnemery:
The issue is that the default constructor for std::thread constructs an object which does not represent a thread.
That made backgroundThread.native_handle() return a NULL pointer.


gizmocuz:
A hardware class is derived from ASyncSerial/TCP, does it make sense to name these threads ?
When using 10 hardware classes depending on the async serial/tcp class, would this not create 20 named threads, and 10 have the same name, and is confusing and can not be used ?

emontnemery:
OK, that's a pretty good point..

PR #2625, opened 2018-08-10
Fix races in plugin framework

emontnemery:
This PR includes only changes from #2598 which are critical for fixing #2541.


gizmocuz:
Thanks!

PR #2626, opened 2018-08-10
Fix various data races

emontnemery:
This PR includes changes from #2598 which are **not** critical for fixing #2541.

Reason for these changes:
1. Write of integer is NOT an atomic operation, see https://stackoverflow.com/questions/54188/are-c-reads-and-writes-of-an-int-atomic, the current way of sharing some between threads in member variables has undefined behavior
2. Make it possible to use helgrind data race detection tool without a huge number of complaints about non thread safe sharing of data

### DIFF
--- a/hardware/DavisLoggerSerial.cpp
+++ b/hardware/DavisLoggerSerial.cpp
@@ -114,7 +114,7 @@ void CDavisLoggerSerial::Do_Work()
 		sec_counter++;
 
 		if (sec_counter % 12 == 0) {
-			mytime(&m_LastHeartbeat);
+			m_LastHeartbeat = mytime(NULL);
 		}
 		if (m_stoprequested)
 			break;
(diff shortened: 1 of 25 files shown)


emontnemery:
@gizmocuz You mentioned in #2598 you have plans to implement proper signalling/messaging between the threads instead of communication through member variables. If so, most of these changes are definitely not needed. Is such a change something you might want help with?

gizmocuz:
@emontnemery , i think i want to do this for one or two classes, then have a review, and then we could do this together (i do classes started with A-M, and you from M - Z ;))

In here i would undo the atomic changes, and initialize these header variables just normally like

time_t m_LastHeartbeat = 0;
time_t m_LastHeartbeatReceive =  0;
bool m_bSkipReceiveCheck =  false;
bool m_bIsStarted = false;

etc...

There should be no need for atomic objects

emontnemery:
Closing, better solved by refactoring the HW classes as proposed by @gizmocuz

PR #2627, opened 2018-08-10
[1Wire] fix for all old devices

szczukot:
http://www.domoticz.com/forum/viewtopic.php?f=6&t=24177


gizmocuz:
Thanks, that was a stupid type from me i quess

PR #2628, opened 2018-08-10
Fix build on FreeBSD

xbeaudouin:
On FreeBSD we don't have pthread_setname_np() nor PTHREAD_MAX_NAMELEN_NP so the build fails.

### DIFF
--- a/main/Helper.cpp
+++ b/main/Helper.cpp
@@ -38,7 +38,7 @@
 #endif
 
 #if defined(__FreeBSD__) 
-// Check if NetBSD / OpenBSD or DragonFly need that at well?
+// Check if OpenBSD or DragonFly need that at well?
 #include <pthread_np.h>
 #ifndef PTHREAD_MAX_MAMELEN_NP
 #define PTHREAD_MAX_NAMELEN_NP 32 	// Arbitrary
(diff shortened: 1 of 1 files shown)


emontnemery:
@xbeaudouin According to the docs you have it: http://nixdoc.net/man-pages/FreeBSD/man3/pthread_set_name_np.3.html
Maybe need to `#include <pthread_np.h>`?

xbeaudouin:
pthread_set_name_np() is here but : use of undeclared identifier 'PTHREAD_MAX_NAMELEN_NP'
        char name_trunc[PTHREAD_MAX_NAMELEN_NP];
Is mostly what break the build.

gizmocuz:
https://github.com/joyent/rfd/blob/master/rfd/0096/README.md

That could be defined if it is not elsewhere defined.

I dont know the internals of FreeBSD, maybe there is no limitation... But in that case i would define it as 32 as well

xbeaudouin:
I have done test with length to 10 (should be ok) maybe we can change that...

main/Helper.cpp
```diff
@@ -37,6 +37,14 @@
 #include <time.h>
 #endif
 
+#if defined(__FreeBSD__) 
+// Check if NetBSD / OpenBSD or DragonFly need that at well?
+#include <pthread_np.h>
+#ifndef PTHREAD_MAX_MAMELEN_NP
+#define PTHREAD_MAX_NAMELEN_NP 10 	// Arbitrary
```
gizmocuz: I think this is a bit to small. Why not make it 32 like on other systems ?

krytarowski: Please do not include <pthread_np.h> on NetBSD, it does not exist and it will break the build.

main/Helper.cpp
```diff
@@ -1128,12 +1136,18 @@ int SetThreadName(std::thread::native_handle_type thread, const char *name)
 	strncpy(name_trunc, name, sizeof(name_trunc));
 	name_trunc[sizeof(name_trunc)-1] = '\0';
 	return pthread_setname_np(thread, name_trunc);
```
krytarowski: while there, there is a bug
`return pthread_setname_np(thread, name_trunc);` -> `return pthread_setname_np(thread, "%s", (void *)name_trunc);`

xbeaudouin: Fixed.

gizmocuz:
Thanks you !

PR #2632, opened 2018-08-10
Store energy meter mode in options map

### LINKED ISSUE #2631: "Options" column meaning for utility devices clashes with Python Plugins
"Options" column in "DeviceStatus" table has special meaning for some utility devices.
This clashes with how the Plugin framework stores data by using CSQLHelper::SetDeviceOptions

Suggestion:
- Harmonize utility device use of options column to also use CSQLHelper::SetDeviceOptions

emontnemery:
Should fix #2631


gizmocuz:
Thanks!

PR #2640, opened 2018-08-12
Options fix for selector level from selector switch

Olifant1990:
- You can now move down all selector levels but the last one
- You can now move up all selector levels but the first one
- You can now rename/delete first row too.

When removing all selector levels, default off level is comming back after save.


gizmocuz:
thanks !!

PR #2646, opened 2018-08-15
Implement IsDebuggerPresent() for other OS than Linux

emontnemery:
For Windows: Call built in function bool IsDebuggerPresent(void);
For BSD/MacOS: Try to attach to the domoticz process, which will fail if Domoticz is already being debugged
Also: Fix/cleanup Linux implementation.

### DIFF
--- a/main/Helper.cpp
+++ b/main/Helper.cpp
@@ -1,11 +1,15 @@
 #include "stdafx.h"
 #include "Helper.h"
+#include "Logger.h"
 #ifdef WIN32
 #include "dirent_windows.h"
 #include <direct.h>
 #else
 #include <dirent.h>
 #include <unistd.h>
+#endif
+#if !defined(WIN32)
+#include <sys/ptrace.h>
 #endif
 #include <sys/types.h>
 #include <sys/stat.h>
@@ -1125,6 +1129,7 @@ int SetThreadName(std::thread::native_handle_type thread, const char *name)
 	return pthread_setname_np(thread, name_trunc);
 #elif defined(macintosh) || defined(__APPLE__) || defined(__APPLE_CC__)
 	// Not possible to set name of other thread: https://stackoverflow.com/questions/2369738/how-to-set-the-name-of-a-thread-in-linux-pthreads
+	return 0;
 #elif defined(__NetBSD__)
 	char name_trunc[PTHREAD_MAX_NAMELEN_NP];
 	strncpy(name_trunc, name, sizeof(name_trunc));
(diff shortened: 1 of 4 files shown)


emontnemery:
@gizmocuz Please help to test Windows functionality.
I have no way to test the BSD/MacOS versions myself, @krytarowski, @xbeaudouin, is this something you can review and test?
Otherwise, maybe it would be better to leave BSD/MacOS version empty, perhaps something like:
```
bool IsDebuggerPresent(void)
{
#if defined(__linux__)
	// Linux implementation
	...
#else
	// MacOSx / BSD
#ifdef _DEBUG
	return true;
#else
	return false;
#endif
}

```

main/Helper.cpp
```diff
@@ -1140,6 +1145,117 @@ int SetThreadName(std::thread::native_handle_type thread, const char *name)
 }
 #endif
 
+#if !defined(PTRACE_ATTACH) && defined(PT_ATTACH)
```
krytarowski: Why this?

main/Helper.cpp
```diff
@@ -1140,6 +1145,117 @@ int SetThreadName(std::thread::native_handle_type thread, const char *name)
...
+#  define PTRACE_ATTACH PT_ATTACH
+#endif
+#if !defined(PTRACE_DETACH) && defined(PT_DETACH)
+#  define PTRACE_DETACH PT_DETACH
+#endif
+
+#define _PTRACE(_x, _y) ptrace(_x, _y, NULL, 0)
```
krytarowski: Why this?

main/Helper.cpp
```diff
@@ -1140,6 +1145,117 @@ int SetThreadName(std::thread::native_handle_type thread, const char *name)
...
+	{
+		if (::isspace(*characterPtr))
+			continue;
+		else
+			return ::isdigit(*characterPtr) != 0 && *characterPtr != '0';
+	}
+
+	return false;
+#else
+	// MacOSx / BSD: Try to attach as suggested in https://stackoverflow.com/questions/3596781/how-to-detect-if-the-current-process-is-being-run-by-gdb
```
krytarowski: This is questionable approach.

gizmocuz:
@emontnemery , 'IsDebuggerPresent' is working for me in VStudio!

main/Helper.cpp
emontnemery: OK, good point, it is not needed.
(The idea is to be compatible with both Linux and Mac OSX / BSD, but since this code is not not used for Linux it is not needed)

main/Helper.cpp
emontnemery: OK, good point, it is not needed.
(The idea is to be compatible with both Linux and Mac OSX / BSD, but since this code is not not used for Linux it is not needed)

main/Helper.cpp
emontnemery: You mean trying to attach is a questionable approach?
If so, I'd rather replace with a compile time hard-coded return value and let someone who is actually debugging Domoticz under MacOS X / BSD do a good implementation.

emontnemery:
@gizmocuz Cool, thanks for testing!

main/Helper.cpp
krytarowski: Right, for NetBSD the appropriate approach is to use sysctl(3) and check flags of the process whether it has traced bit turned on or not.

There is need to check `p_slflag` for `PSL_TRACED`.

I don't use Darwin and other BSDs, but the algorithm is the same.

krytarowski: I wrote it somewhere... but I cannot find it now. I will rewrite it from scratch.

krytarowski: http://netbsd.org/~kamil/gdb/isdebugged.c

emontnemery: OK, I pushed a new commit with empty implementation for BSD as starting point.

krytarowski:
NetBSD check: http://netbsd.org/%7Ekamil/gdb/isdebugged.c

emontnemery:
@gizmocuz I removed the BSD version. Please merge this version and let developers using BSD contribute proper implementation (such as the clean NetBSD implementation linked to by @krytarowski).

gizmocuz:
Thanks for the patch!

PR #2647, opened 2018-08-15
Fix build error when compiling with older (but supported) boost versions

emontnemery:


gizmocuz:
Which boost version did you build against ? It's always a good idea to use a recent version

emontnemery:
steady_timer::expires_after() was just introduced in Boost 1.66, isn't that a too tough requirement?
Please note it my own mistake to use steady_timer::expires_after() in #2575 :)

gizmocuz:
Thanks! Yes thats a bit to new *g*

PR #2648, opened 2018-08-16
The devices are not created correctly

pipiche38:
The function PushAndWaitRx must be call after the safe_query in order to have the if (result.empty(s) statement working correctly !

That is a not a great code, but this is how it was originally implemented. the latest changes which have been made in 0af09b28e215e31d726081e07ea4a63176919bbe have introduce the issue.


gizmocuz:
Thanks!

Issue #2126, opened 2018-02-10
Libssl 1.1 vs libssl 1.0.2


kofec:
Sorry but I don't have time to continue previous investigation but today I have a while and what I found
https://github.com/domoticz/domoticz/issues/2098

Thread 3 "domoticz" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0xb5613430 (LWP 31660)]
__GI_strcmp (p1=, p2=) at strcmp.c:38
38	strcmp.c: No such file or directory.
(gdb) bt
#0 __GI_strcmp (p1=, p2=) at strcmp.c:38
#1 0x00687cd0 in getrn ()
#2 0x00687ef8 in lh_insert ()
#3 0x0066dcec in OBJ_NAME_add ()
#4 0x0068dfd4 in EVP_add_digest ()
#5 0xa7a87e5c in ?? () from /usr/lib/arm-linux-gnueabihf/libcrypto.so.1.1
Backtrace stopped: previous frame identical to this frame (corrupt stack?)

so I decide to remove(change name of) library /usr/lib/arm-linux-gnueabihf/libcrypto.so.1.1 to
/usr/lib/arm-linux-gnueabihf/libcrypto.so.1.0.1
and now Domticz is working fine but 2 python plugins which are using libcrypto.so.1.1 failed (as it is missing).

ldd show that Domoticz should use 1.0.2 so I don't know why Domoticz crash when use 1.1

~/domoticz $ ldd domoticz
	/usr/lib/arm-linux-gnueabihf/libcrypto.so.1.0.2 (0xb6d7b000)
	/usr/lib/arm-linux-gnueabihf/libarmmem.so (0xb6d65000)
	libusb-0.1.so.4 => /lib/arm-linux-gnueabihf/libusb-0.1.so.4 (0xb6d4d000)
	libudev.so.1 => /lib/arm-linux-gnueabihf/libudev.so.1 (0xb6d30000)
	librt.so.1 => /lib/arm-linux-gnueabihf/librt.so.1 (0xb6d19000)
	libresolv.so.2 => /lib/arm-linux-gnueabihf/libresolv.so.2 (0xb6cf4000)
	libpthread.so.0 => /lib/arm-linux-gnueabihf/libpthread.so.0 (0xb6ccb000)
	libz.so.1 => /lib/arm-linux-gnueabihf/libz.so.1 (0xb6ca4000)
[...]

PR #2652, opened 2018-08-17
Block openssl higher than 1.0.2

emontnemery:
Abort if cmake finds openssl version greater than 1.0.2 (i.e. 1.1)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,6 +7,10 @@ project(domoticz)
 SET(DOMO_MIN_LIBBOOST_VERSION 106000)
 ##
 
+## required max. OpenSSL version
+SET(DOMO_MAX_OPENSSL_VERSION 1.0.2)
+##
+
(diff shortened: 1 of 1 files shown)


Porzeczek:
But you're aware that this will block compiling i.e. on Debian Stretch?

gizmocuz:
If there an issue with 1.1 specifically ?

emontnemery:
@Porzeczek No that's not the case. On debian stretch, and distros derived from it, you need to:
```
sudo apt-get remove libssl-dev
sudo apt-get install libssl1.0-dev
```

@gizmocuz Yeah, many reports of users compiling against libssl1.1 causing crashes, for example here:
#2126.

Edit: Domoticz won't compile on Ubuntu 18.04 (Bionic) with this change though (libcurl4-openssl-dev now depends on libssl-dev (>= 1.1))

emontnemery:
Closing this, need to find a better way to block incompatible libssl version.

PR #2656, opened 2018-08-19
Reenable check of formatting strings

emontnemery:
Reenable checking of formatting strings (was removed by a188a925f9a85df822f7f7be79cf2bd935b0d6c5)
Fix associated warnings

### DIFF
--- a/main/WebServer.cpp
+++ b/main/WebServer.cpp
@@ -16814,7 +16814,7 @@ namespace http {
 									root["result"][ii]["sx"] = sx;
 									char szTmp[1024];
 									sprintf(szTmp, "%.1f %.1f %.1f", sm, se, sx);
-									_log.Log(LOG_STATUS, szTmp);
+									_log.Log(LOG_STATUS, "%s", szTmp);
 
 								}
 								ii++;
--- a/hardware/plugins/PluginProtocols.cpp
+++ b/hardware/plugins/PluginProtocols.cpp
@@ -163,32 +163,32 @@ namespace Plugins {
 				{
 					PyObject*	pObj = JSONtoPython(&pRef);
 					if (!pObj || (PyList_SetItem(pRetVal, Index++, pObj) == -1))
-						_log.Log(LOG_ERROR, "(%s) failed to add item '%d', to list for object.", __func__, Index - 1);
+						_log.Log(LOG_ERROR, "(%s) failed to add item '%zd', to list for object.", __func__, Index - 1);
 				}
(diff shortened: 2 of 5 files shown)


main/WebServer.cpp
```diff
@@ -16814,7 +16814,7 @@ namespace http {
 									root["result"][ii]["sx"] = sx;
 									char szTmp[1024];
 									sprintf(szTmp, "%.1f %.1f %.1f", sm, se, sx);
-									_log.Log(LOG_STATUS, szTmp);
+									_log.Log(LOG_STATUS, "%s", szTmp);
```
gizmocuz: This won't be needed as there is a specific override for that (void Log(const _eLogLevel level, const std::string& sLogline);)

gizmocuz:
@emontnemery , explain a bit more what specifically was removed as this patch has a lot more then that

emontnemery:
@gizmocuz It was removed here: https://github.com/domoticz/domoticz/commit/a188a925f9a85df822f7f7be79cf2bd935b0d6c5#diff-786ce91e0059756f71e70a800667a663

gizmocuz:
This was related to the new hardware log method.
This patch contains a lot of formatting changes that i didn't do in that patch

emontnemery:
So you want me to split this PR in two:
1. Re-enable  type checking of formatting string
2. Fix the warnings caused by 1)

?

gizmocuz:
The extra type check attribute in the log header i would not do. '...' is used a lot, even in sqlite, never used
The warnings you receive, they should also be found with GCC or cppcheck, regardless the extra attribute ?
Also all formatting errors have nothing to do with my previous patch, they where already wrong.
So you can make a PR with as subject "Corrected some string formatting's'' or similar ;)

emontnemery:
GCC can only give warnings if it knows there is an printf-style formatting string, '...' is not only used for printf.
For example, in case of sqlite, '...' does not follow same rules as printf.

So, by rewriting function declaration like this:
void Log(const _eLogLevel level, const char* logline, ...) **__attribute__ ((format (printf, 3, 4)))** ;
gcc will check that the type and number of variadic arguments match the formatting string, and that the formatting string is valid.

gizmocuz:
then i suggest to create a PR for the attribute, and also add this in

1WireForWindows.cpp
EvohomeBase.h
Logger.h
SQLHelper.h

And create a separate PR for fixing incorrect formatting

emontnemery:
@gizmocuz OK

PR #2657, opened 2018-08-19
Use std::stoull / std::stoll instead of std::stringstream

emontnemery:


gizmocuz:
Thanks for another big refactoring ;)

gizmocuz:
Some users are reporting problems since this patch.
I happen to revert this patch in the UpdateMeter code a while ago as this was causing the exact same issue

```
terminate called after throwing an instance of 'std::invalid_argument'
  what():  stoll
```

See this forum thread:
https://www.domoticz.com/forum/viewtopic.php?f=6&t=24601

For now i will revert this patch "only" in that function, but it needs to be checked.... 
I know std::stoll is not thread safe, but something is used as the argument that is causing this crash

gizmocuz:
Make a commit that hopefully solves the problem, i suspect that sValue could be empty

https://github.com/domoticz/domoticz/commit/78ed6e28af2a21bf11327efce220980776ad4bec

std::stoll crashes when the argument is empty, or is a non-number

susage is always 0 or the usage, i don't think that's the issue

gizmocuz:
Sorry, reverting this patch. Will test with the user to see if that solves the problem

emontnemery:
Sorry about that!!

gizmocuz:
It could be related to a wrong database sValue (for instance when someone used a json query to update a sensor, or a wrong script)
Hope to receive the database and then apply the PR again and see whats wrong

emontnemery:
Yeah, but I think it was a bit aggressive to throw() on bad data, it requires error handling everywhere..
I opened #2682 instead, which has same behavior as the stringstream: return 0 if string to number conversion fails.

PR #2660, opened 2018-08-19
Update NotificationPushsafer.cpp

appzer:
add new [Pushsafer](https://www.pushsafer.com) parameter, priority, retry, expire, answer

### DIFF
--- a/notifications/NotificationPushsafer.cpp
+++ b/notifications/NotificationPushsafer.cpp
@@ -23,6 +23,7 @@ bool CNotificationPushsafer::SendMessageImplementation(
 	const std::string &Name,
 	const std::string &Subject,
 	const std::string &Text,
+	const int Priority,
 	const std::string &ExtraData,
 	const std::string &Sound,
 	const bool bFromNotification)
--- a/notifications/NotificationPushsafer.cpp
+++ b/notifications/NotificationPushsafer.cpp
@@ -65,18 +65,12 @@ bool CNotificationPushsafer::SendMessageImplementation(
 			_apikey = PS_k;
 		}
 
-		if (PS_m.length() >= 1)
-		{
-		}
-		else
+		if (PS_m.empty())
 		{
 			PS_m = Text;
 		}
-
-		if (PS_t.length() >= 1)
-		{
-		}
-		else
+		
+		if (PS_t.empty())
 		{
 			PS_t = cSubject;
 		}
--- a/notifications/NotificationPushsafer.cpp
+++ b/notifications/NotificationPushsafer.cpp
@@ -23,10 +23,11 @@ bool CNotificationPushsafer::SendMessageImplementation(
 	const std::string &Name,
 	const std::string &Subject,
 	const std::string &Text,
-	const int Priority,
 	const std::string &ExtraData,
+	const int Priority,
 	const std::string &Sound,
 	const bool bFromNotification)
+	
 {
 	//send message to Pushsafer
 	std::string cSubject = (Subject == Text) ? "Domoticz" : Subject;
--- a/notifications/NotificationPushsafer.cpp
+++ b/notifications/NotificationPushsafer.cpp
@@ -42,9 +42,9 @@ bool CNotificationPushsafer::SendMessageImplementation(
 
 	std::vector<std::string> splitresults;
 	StringSplit(CURLEncode::URLDecode(Text), "|", splitresults);
-	if (splitresults.size() == 15)
+	if (splitresults.size() >= 11)
 	{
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thanks for the PR.
The priority is already in the function call (SendMessageImplementation)

If you check above for 'if (splitresults.size() == 11)', are you sure you can do 'std::string PS_a = splitresults[14];'

Is this verified ?

How are you using this ?

appzer:
the priority in SendMessageImplementation is not needed i delete it.
All other is OK!

gizmocuz:
@appzer , no you can not do this, and you should use this if you want to use the priority in the message system.
This priority can be set on various places in Domoticz.
How are you using the other fields ?
I do not think are checking the code changes you are pushing as they contain build errors, and also the split size was incorrect in your initial post.
Please be carefull

appzer:
OK, the priority func I have involved again.
The split size adjusted to 15
All the parameter are send to the Pushsafer API to generate a push-notification based on the [API](https://www.pushsafer.com/en/pushapi)

Now it works fine!

gizmocuz:
@appzer , yes i believe it does,  but 'how' are you using it with domoticz?
How do you use the 'answer' or the 'retry' or expire ?

appzer:
here a documentation to use the pushsafer plugin in domoticz
[https://www.pushsafer.com/en/domoticz](https://www.pushsafer.com/en/domoticz)

gizmocuz:
@appzer , thanks for the documentaton. Maybe this can de added to our wiki page.
But these 4 parameters Priority|Retry|Expire|Answer are optional, and do not needed to be send
It will probably also break the implementation for current users, who do not use the optional 3 parameters
Because of this, i would say, if size()>11 (meaning 12 or higher) then we can use the Retry, if size()>12, we have the Expire, if size()>13 the user entered the answer

In the documentation, the priority can be removed as this can already be specified in the userinterface/script

appzer:
i changed the documentation and also the plugin.
Now the domoticz priority are passed throught the pr parameter of pushsafer.

I also changed the split size to >=11, so user with new and older domoticz version can work with parameters.

Kevin

gizmocuz:
You have some strange code here

```
if (PS_m.length() >= 1)
		{
		}
		else
		{
			PS_m = Text;
		}
```

Do we need those extra parameters at all ? Just that it is possible, does not mean it is useful right ?

appzer:
@gizmocuz  it is ok so, yes many domoticz customers want the new parameter.

gizmocuz:
Sure thats not OK code, maybe you mean this:

```
if (PS_m.empty())
{
	PS_m = Text;
}
```

Please point me to some references where uses want these parameters please

appzer:
i fix the code.
I got many mails from domoticz users, they ask me, when i implement the new pushsafer parameters.
especially the priority parameter, because domoticz also handle this.

Kevin

gizmocuz:
But the sendmessage implementation had already a priority field.
You can use this.
So if the other parameters are not needed, please only implement the priority

appzer:
@gizmocuz 
i only use the implement priority.
`const int Priority`
The other 3 new pushsafer parameters also needed.
Why it is wrong to add them!

Kevin

gizmocuz:
@appzer , first, as no other pusher implement these new tree actions, not is there any facility to invoke these actions straight from domoticz without using a third party script.
Next, i do not see any use of these tree new actions, and have not seen any post that people desperately need them.
And it's not the best code practice. 
So it you would like to see the priority added, please undo the other changes, or create a new PR with only the priority option.
Or please show me (like asked before) the forum thread where many people are requesting the other tree options

appzer:
@gizmocuz  I get the requests to add the new parameters via email, for privacy reasons, I can not post complete emails here.

Priority, Retry and Expire are really needed.

With retry and expire the user has the possibility to resend notifications in a given time > independent of domoticz or other platforms. It is really usefull.

What part it's not the best code practice? I already changed the if clauses to .empty().

I ask you to release this changes

Thanks Kevin

gizmocuz:
Okey Kevin, first things first, i do not need to accept any pull requests.
But i am willing to consider them, if i think they are correct.
So, going back to my question then, please submit a new PR with the priority implementation

Next we can discuss (here, on on slack in a private channel), why you need retry and expire, and how this will be implemented in domoticz.
As you say it is requested on the forum, show me a link for that

appzer:
@gizmocuz 
OK, i update the PR and added only the priorty value.
How we can discuss it on a private way here?

Kevin

gizmocuz:
Thanks! , Feel free to join https://domoticz.slack.com

appzer:
@gizmocuz  thanks for merging!
How i can create an domoticz slack account?

gizmocuz:
Good question, i need your email address for that. Can you send me a PM on the forum ?

appzer:
@gizmocuz is sent out

PR #2661, opened 2018-08-20
Edit Denkovi Classes

dzelyaydinov:


hardware/DenkoviTCPDevices.h
```diff
@@ -86,15 +65,13 @@ class CDenkoviTCPDevices : public CDomoticzHardwareBase, ASyncTCP
...
-	//_sDenkoviTCPModbusRequest sReq;
-	_sDenkoviTCPModbusRequest *pReq;// = new _sDenkoviTCPModbusRequest;// &sReq;
-	//_sDenkoviTCPModbusResponse sResp;
-	_sDenkoviTCPModbusResponse *pResp;// = new _sDenkoviTCPModbusResponse;// &sResp;
-	uint16_t receivedDataLength = 0;
-	uint16_t transactionCounter = 0;
+	bool m_bReadingNow = false;
+	bool m_bUpdateIo = false;
+	bool m_bFirstTime = true;
+	_sDenkoviTCPModbusRequest *m_pReq;
```
gizmocuz: Question, why do this have to be a pointer ? Why not

`_sDenkoviTCPModbusRequest m_Req;`

Same for m_Resp

dzelyaydinov: I do it in this way because of ASyncTCP returns the respond of request in portions of bytes and for me is easier to handle the respond as a pointer as in this way I just copy received bytes. What do you suggest?

gizmocuz: Keeping it a member... i am not so happy with new/delete if that is not needed.
Also memcpy and pointers .... i don't like
I think the response of your devices is text based or really hexadecimal byte based?
Still you can use a std::string (.append) maybe ?

dzelyaydinov: The response is hexadecimal data and not text based. What are the other ways to handle hexadecimal responses without using new/delete?

gizmocuz: still you can use a std::string as byte buffer

dzelyaydinov: std::string a = "";
char arr[20] = { 'a',0x00,'b' };
a.append((const char*)arr);//null terminates this std::string
int b = a.length();//length is 1

This does not work for me, but I will look for another way.

dzelyaydinov: Is it ok not to be a member of the class and only to be created/deleted (using new/delete) in receive callback?

dzelyaydinov:
Is it OK now?

gizmocuz:
Looking much better now! Only the memcpy/poking in the memory buffer worries me a bit.
So you get bytes back from a TCP connection, and you know when it terminates and it is ready to handle ?
You could look at the way of having a std::string (m_rxbuf) and use m_rxbuf.append to add the new bytes till your done, then do the processing, then clear the rx buffer with m_rxbuf.clear() ?

dzelyaydinov:
What about these changes?

hardware/DenkoviTCPDevices.h
gizmocuz: No new/delete please
please have a test with std::string with append,insert etc...
Use it quite a lot when working with small memory

gizmocuz:
Better! Thanks
How about the ByteSwap routine ? Did you know the function 'htons' ?

dzelyaydinov:
ByteSwap function is not used anymore, so I think it does not matter the functionality. But it is good to know about htons(). :)

dzelyaydinov:
I think we are ready and if you do not have any other remarks, please merge. Thanks again.

gizmocuz:
Good to hear, can you remove that obsolete function ? I will merge after this

dzelyaydinov:
Done.

gizmocuz:
Thanks!!

PR #2665, opened 2018-08-21
Fix incorrect printf format strings and arguments

emontnemery:


gizmocuz:
Thanks for the corrections !

PR #2666, opened 2018-08-21
Check printf format string and arguments

emontnemery:
Notes:
- This functionality is gnu specific and not available in msvc, so no point enabling for Log function in hardware/1Wire/1WireForWindows.cpp
- The SQLHelper functions safe_query, safe_queryBlob and safe_exec_no_return do take format strings and variable amount of arguments, but the format strings do not follow same rules as printf


gizmocuz:
Thanks!

Issue #2601, opened 2018-07-29
V4.9796 - Crashes when disable hardware


poudenes:
When disable hardware i get this error. Doesn't matter what kind of hardware. After restart domoticz, the hardware is disabled. So the action itself has been executed. Enabling is working without errors.

Version: 4.9796
Build Hash: 4952b51e
Compile Date: 2018-07-26 17:01:10
dzVents Version: 2.4.7
Python Version: 3.6.4 (default, May 12 2018, 20:07:58) [GCC 6.3.0 20170516]

```
2018-07-29 11:00:04.521  Status: dzVents: Info:  ------ Finished Script Weather
2018-07-29 11:00:04.521  Status: EventSystem: Script event triggered: /home/pi/domoticz/dzVents/runtime/dzVents.lua
2018-07-29 11:00:13.738  (P1) P1 Smart Meter (Watt)
2018-07-29 11:00:13.749  (P1) General/Voltage (Voltage)
2018-07-29 11:00:13.752  (PiHole Remote) General/Percentage (PiHole - CPU)
2018-07-29 11:00:13.841  Error: Thread 3 (Thread 0x758ac430 (LWP 9383)):
2018-07-29 11:00:13.841  Error: #0  0x76e604cc in __waitpid (pid=28239, stat_loc=0x758ab4e0, options=0) at ../sysdeps/unix/sysv/linux/waitpid.c:29
2018-07-29 11:00:13.842  Error: #1  0x00228abc in dumpstack_gdb() ()
2018-07-29 11:00:13.842  Error: #2  0x00228f1c in signal_handler(int, siginfo_t*, void*) ()
2018-07-29 11:00:13.842  Error: #3  <signal handler called>
2018-07-29 11:00:13.842  Error: #4  _PyObject_Alloc (ctx=<optimized out>, elsize=44, nelem=1, use_calloc=0) at Objects/obmalloc.c:1258
2018-07-29 11:00:13.842  Error: #5  _PyObject_Malloc (ctx=<optimized out>, nbytes=44) at Objects/obmalloc.c:1437
2018-07-29 11:00:13.843  Error: #6  0x76245340 in _PyObject_GC_Alloc (basicsize=<optimized out>, use_calloc=0) at Modules/gcmodule.c:1714
2018-07-29 11:00:13.843  Error: #7  _PyObject_GC_Malloc (basicsize=<optimized out>) at Modules/gcmodule.c:1736
[...]

PR #2668, opened 2018-08-23
fix crash on disable hardware

gordonb3:
See issue [V4.9796 - Crashes when disable hardware](https://github.com/domoticz/domoticz/issues/2601)

### DIFF
--- a/hardware/CurrentCostMeterTCP.cpp
+++ b/hardware/CurrentCostMeterTCP.cpp
@@ -114,8 +114,13 @@ void CurrentCostMeterTCP::disconnect()
 	m_stoprequested=true;
 	if (m_socket==INVALID_SOCKET)
 		return;
-	closesocket(m_socket);	//will terminate the thread
+	closesocket(m_socket);
 	m_socket=INVALID_SOCKET;
+	sleep_seconds(1);
+	if (m_thread)
+	{
+		m_thread->join();
+	}
 }
 
 
--- a/hardware/DomoticzTCP.cpp
+++ b/hardware/DomoticzTCP.cpp
@@ -217,8 +217,13 @@ void DomoticzTCP::disconnectTCP()
 	if (m_socket != INVALID_SOCKET)
 	{
 		shutdown(m_socket, SHUT_RDWR);
-		closesocket(m_socket);	//will terminate the thread
+		closesocket(m_socket);
 		m_socket = INVALID_SOCKET;
+		sleep_seconds(1);
+	}
+	if (m_thread)
+	{
+		m_thread->join();
 	}
 }
 
--- a/hardware/P1MeterTCP.cpp
+++ b/hardware/P1MeterTCP.cpp
@@ -115,8 +115,13 @@ void P1MeterTCP::disconnect()
 	m_stoprequested=true;
 	if (m_socket==INVALID_SOCKET)
 		return;
-	closesocket(m_socket);	//will terminate the thread
+	closesocket(m_socket);
 	m_socket=INVALID_SOCKET;
+	sleep_seconds(1);
+	if (m_thread)
+	{
+		m_thread->join();
+	}
 }
 
 
--- a/hardware/SolarMaxTCP.cpp
+++ b/hardware/SolarMaxTCP.cpp
@@ -173,11 +173,14 @@ void SolarMaxTCP::disconnect()
 	m_stoprequested = true;
 	if (m_socket != INVALID_SOCKET)
 	{
-		closesocket(m_socket);	//will terminate the thread
+		closesocket(m_socket);
 		m_socket = INVALID_SOCKET;
 		sleep_seconds(1);
 	}
-	//m_thread-> join();
+	if (m_thread)
+	{
+		m_thread->join();
+	}
 }
 
 void SolarMaxTCP::Do_Work()


emontnemery:
The sleep(1) is no good, please do something like this instead:
```
void SolarMaxTCP::disconnect()
 {
	m_stoprequested = true;
	if (m_thread)
	{
		m_thread->join();
		m_thread.reset();
	}
	if (m_socket != INVALID_SOCKET)
	{
		closesocket(m_socket);
		m_socket = INVALID_SOCKET;
	}
}
```

By the way, this pattern occurs in a few hardware classes:
```
bool CHEOS::StopHardware()
{
	m_stoprequested = true;
	if (isConnected())
	{
		try {
			disconnect();
		}
		catch (...)
		{
			//Don't throw from a Stop command
		}
	}
	try {
		if (m_thread)
		{
			m_thread->join();
			m_thread.reset();
		}
	}
	catch (...)
	{
		//Don't throw from a Stop command
	}

	m_bIsStarted = false;
	return true;
}
```

Shouldn't it be changed like below to avoid threading issues:
```
bool CHEOS::StopHardware()
{
	m_stoprequested = true;
	try {
		if (m_thread)
		{
			m_thread->join();
			m_thread.reset();
		}
	}
	catch (...)
	{
		//Don't throw from a Stop command
	}
	if (isConnected())
	{
		try {
			disconnect();
		}
		catch (...)
		{
			//Don't throw from a Stop command
		}
	}
	

	m_bIsStarted = false;
	return true;
}
```

gordonb3:
Yeah, it seems to have become something of a mess, changing from boost to c++11

Unsure about that first one. Is it safe to join the thread when the (is it an async?) socket is still open? As you can see in the diff the `sleep` was already part of `SolarMaxTCP` and it seemed to make sense.

gizmocuz:
Thanks, yeah it's a bit of a mess, going to implement a signal method soon

gordonb3:
Excellent. A bit of standardization should definitely help to prevent problems like these.

emontnemery:
Hmm, but I **really** don't like the sleep.
If there is a potential threading issue, a sleep can't be the solution as it just hides the real issue (most of the time) and makes the real problem much harder to find and debug.

gordonb3:
I partly agree. The disagreeing part here being that the 1 second is a TCP protocol related timeout and since with async we're not actively monitoring the traffic I'd say it makes more sense to obey the timeout value than risk a crash due to some external disturbance. 

But I assume the new signal method means to handle all of that.

emontnemery:
What TCP protocol related timeout????

The sleep is completely, 100% useless and does absolutely nothing because a call to join() does nothin else than waiting until the thread has terminated:
```
	m_stoprequested=true; // Signal to worker thread that it's time to die
	if (m_socket==INVALID_SOCKET)
		return;
	closesocket(m_socket); // Not thread safe, and will cause crashes because the socket is also used by the worker thread
	m_socket=INVALID_SOCKET;
	sleep_seconds(1); // Make sure UI / domoticz shutdown is slower than necessary
	if (m_thread)
	{
		m_thread->join(); // Wait for worker thread to exit
	}
```

This is what the code should be change to:
```
	m_stoprequested=true; // Signal to worker thread that it's time to die

	if (m_thread)
	{
		m_thread->join(); // Wait for worker thread to exit
	}
	if (m_socket==INVALID_SOCKET)
		return;
	closesocket(m_socket); // Thread safe, because we have join()ed the thread
	m_socket=INVALID_SOCKET;
```

emontnemery:
@gizmocuz I'm more than interested in helping out with implementing the signalling. Do you have some prototype already or a particular API in mind? If not, I'd propose to use boost_asio::post.

gordonb3:
You are forgetting that async I/O is a different thread as well. And it looks for an end point in the thread that we are closing. That is the whole point: destroying the thread means destroying the end point which introduces the possibility of causing a memory access violation.

Again: I'm not really sure what we are looking at with these TCP related hardware classes. If these are synchronous I/O then yes the sleep has no function whatsoever. As far as I'm concerned the main concern at this very moment is to stop the crashes and since a new method is currently being developed I assume that as a temporary fix this will suffice.

emontnemery:
Please note the classes you modified are not exending ASyncTCP, they're all doing synchronous IO in the worker. But nevermind, I'll open another PR to remove the sleep :)

By the way, any idea what the purpose of this might be:
```
bool CDomoticzHardwareBase::Stop()
{
	std::lock_guard<std::mutex> l(readQueueMutex); // Why do we need the mutex??
	return StopHardware();
}
```

This causes deadlocks when shutting down the HW class if there is some data arriving after taking the mutex and before the socket / serial port has been shut down.

Edit: I opened #2671 for the deadlock

gordonb3:
In honesty, with respect to TCP connectivity I have no clue what these classes do extend since I do not see any direct reference to any library I know to provide the socket functions. As far as I'm concerned this code is way too big, too diverse, too undocumented to try figure out whatever the person that wrote any particular part was or wasn't thinking.

emontnemery:
Every single HW class extends CDomoticzHardwareBase, meaning the deadlock in #2671 also affects every single HW class. Let's hope someone else has some good idea or good memory then..

PR #2669, opened 2018-08-23
repair graphs and reports for Evohome devices

gordonb3:
a bit dirty, but I don't really see any other way to do it


gizmocuz:
Thanks for the patch!

PR #2674, opened 2018-08-24
Fix TeleinfoSerial frame analysis

llbrt:
- strlen was reading random memory area
- add debug log

### DIFF
--- a/hardware/TeleinfoSerial.cpp
+++ b/hardware/TeleinfoSerial.cpp
@@ -164,8 +164,13 @@ void CTeleinfoSerial::MatchLine()
 	#endif
 
 	// Is the line we got worth analysing any further?
-	if ((strlen((const char*)&line)<4) || (line[0] == 0x0a))
+	if ((strlen(line)<4) || (line[0] == 0x0a))
+	{
+		#ifdef DEBUG_TeleinfoSerial
+		_log.Log(LOG_NORM,"Frame #%s# ignored, too short or irrelevant", line);
+		#endif
 		return;
+	}
 
 	// Extract the elements, return if not enough and line is invalid
 	StringSplit(line, " ", splitresults);


gizmocuz:
Thanks for the PR! I will replace the debug logging with the new debug logging methods after merging, and make it a std::string

gizmocuz:
Got some questions where i hope you can help me with, i know it is not your code, but hope you have some answers and can debug :

- Why is 	StartHeartbeatThread and StopHeartbeatThread() called ? The heartbeat is updated in this class itself with `mytime(&m_LastHeartbeat);// keep heartbeat happy`
- How often is data received ? multiple times a minute ?

- I suspect a complete packet starts or ends with "(label == "MOTDETAT")" ?

Why is this code here?

```
	// at 1200 baud we have roughly one frame per 1,5 second, check more frequently for alerts.
	if (m_counter >= m_iBaudRate / 600)
	{
		m_counter = 0;
		_log.Debug(DEBUG_HARDWARE, "(%s) Teleinfo frame complete, PAPP: %i, PTEC: %s", m_Name.c_str(), m_teleinfo.PAPP, m_teleinfo.PTEC.c_str());
		ProcessTeleinfo(m_teleinfo);
		mytime(&m_LastHeartbeat);// keep heartbeat happy
	}
```

Is it not better to keep a Boolean m_bHaveReceivedMOTDETATBefore, and when it has found it 2 or more times, we call ProcessTeleinfo ?

I moved the CRC checking in the base class, so this can be called by the EFD class for example
(anyone who likes to implement that, please do so)

Made a few changes here:
https://github.com/domoticz/domoticz/commit/a93e2f9cd04f8eaaadc4f49787b508ab6b7c0dbe

Hope they are OK ?

llbrt:
Thanks for the merge 👍 
I had a reproducible situation where the module did not provide any data, so I looked for what could be wrong.

- I did not looked that the architecture of domoticz modules, so I can't tell what should be done for the heart beat.
- Yes, the device sends data continuously, so we get multiple frames per minute.

As in the example in the comments, a frame contains several lines. The first line starts with "ADCO" and the last line with "MOTDETAT". The lines of a frame depends on the pricing (3 kinds of pricing at least).
So, the class TeleInfo contains all the possible datas as documented by ENEDIS (owner of the french electricity meters), but not all the fields are set when a "MOTDETAT" arrives.

I don't know why ProcessTeleinfo() is called only when counter>=2 (to reduce CPU load ?), but the "positive" effect is to avoid to have an incomplete frame at startup (we may start to read the device in the middle of a frame). I would have wrote that in another way, but this code works...

I tested the last version of the code of Teleinfo (not a long run, just a few hours) and it's ok for me  :smile:

gizmocuz:
@llbrt , thanks for the answers ! Yes, i would have written that function also very differently.
The problem with the current implementation is that it is doing some funky logic.
It should just check if there was previously a 'MOTDETAT', and when it happens again, then you know all fields are always available from that moment on (hence my idea for a Boolean)
For example, if you buy a small tp-link wifi router (20 euro's), you can install 'ser2net' on this device, and place it in your meterroom and connect that to the Ethernet.
Then from anywhere/system, you can access your meter via tcp.
This can be extreme easy to implement (because of the base class) (see for example the RFXCom TCP class),  but with the current logic, depending on the baudrate, this can not be done,

But if it works, it works ;)

PR #2675, opened 2018-08-24
Fix difference between domoticz.time object and trigger

trandbert37:
Fix problem for when "at sunrise" is triggered, this script logs "night" instead of "day":
```
return {
    on = {
        timer = {
            'at sunrise',
            'at sunset'
        }

    },
    execute = function(domoticz)
        if (domoticz.time.isNightTime)) then
                domoticz.log('Night', domoticz.LOG_INFO)
        elseif (domoticz.time.isDayTime) then
                domoticz.log('Day', domoticz.LOG_INFO)
        end
    end
}
```

### DIFF
--- a/History.txt
+++ b/History.txt
@@ -3,7 +3,7 @@ Version 4.xxxx (July xxth 2018)
 - Implemented: OpenZWave, support for Seismic Intensity sensor
 - Implemented: InfluxDB Data push, support for Username/Password
 - Fixed: Philips Hue, should work correctly now again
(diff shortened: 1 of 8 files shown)


dannybloe:
Ah.. great. But this seems more like a bug in Domoticz then dzVents so perhaps the dzVents version didn't have to be udpated?

gizmocuz:
Yep, seems only a patch for "EventSystem.cpp" is needed, could you undo the other changes ?

trandbert37:
You are right ;-) done

gizmocuz:
Thanks for the PR!

PR #2680, opened 2018-08-26
Remove RFXCom specifics from CDomoticzHardwareBase

emontnemery:
Move RFXCom specifics from CDomoticzHardwareBase to CRFXBase, get rid of readQueueMutex from most HW.

### DIFF
--- a/hardware/Comm5SMTCP.cpp
+++ b/hardware/Comm5SMTCP.cpp
@@ -51,7 +51,6 @@ bool Comm5SMTCP::StartHardware()
 
 	//force connect the next first time
 	m_bIsStarted = true;
-	m_rxbufferpos = 0;
 
 	//Start worker thread
 	m_thread = std::make_shared<std::thread>(&Comm5SMTCP::Do_Work, this);
@@ -100,7 +99,6 @@ void Comm5SMTCP::Do_Work()
 			bFirstTime = false;
 			if (!mIsConnected)
 			{
-				m_rxbufferpos = 0;
 				connect(m_szIPAddress, m_usIPPort);
 			}
 		}
@@ -173,7 +171,6 @@ bool Comm5SMTCP::WriteToHardware(const char* /*pdata*/, const unsigned char /*le
 
 void Comm5SMTCP::OnData(const unsigned char *pData, size_t length)
 {
-	std::lock_guard<std::mutex> l(readQueueMutex);
 	ParseData(pData, length);
 }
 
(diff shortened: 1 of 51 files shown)


emontnemery:
@gizmocuz I would appreciate if you could help verify that DomoticzTCP is working as expected with this change.

gizmocuz:
Why not remove the whole readQueueMutex ? But i think this was needed when the hardware class was terminating/closing
DomoticzTCP is used in the client/server setup. For instance, you have a main domoticz, install another instance of domoticz, let it connect to your main. (with a specific user with viewer rights and maybe some specific sensors or all)

emontnemery:
The readQueueMutex is still needed to ensure modifications to the readbuffer are done in a safe way, for example these lines in mainworker:
```
CRFXBase *pRFXBase = reinterpret_cast<CRFXBase *>(pHardware);
 pRFXBase->SendResetCommand();
```

OK, I'll give client/server setup a try.

hardware/OpenWebNetTCP.cpp
```diff
@@ -419,6 +419,7 @@ void COpenWebNetTCP::MonitorFrames()
 				}
 				else
 				{
+					//TODO: This mutex is only needed if it's not safe to call ParseData concurrently, is that the case?
```
gizmocuz: Yes this seems possible from the senddata function

emontnemery:
@gizmocuz DomoticzTCP is working well. I rebased and fixed review comments, should be ready for merge!

gizmocuz:
Thanks ! But the proxyclient as well as DomoticzTCP should send valid messages right ? Why the bypass of the correctness check ? This should be done

emontnemery:
@gizmocuz Sure, proxyclient and DomoticzTCP should send valid messages, it's just that there's no guarantee they're RFXCom messages.
CheckValidRFXData() checks that the message is an RFXCom message, so calling CheckValidRFXData() doesn't make sense.

gizmocuz:
Thanks!!

gizmocuz:
Your comment is right, but as soon as we are going to go for a json serialization  approach everything has to go via a big checking/de-serializing routine

gysmo38:
Hello I use domoticz and rfxcom. Since last update if i send lighting4 with rfxcom it stop working. On rfxcom i get a red led stuck. I need to unplug and plug the rfxcom to get it work again.

This commit could be my problem?

emontnemery:
Can you open an issue for that problem please?

gizmocuz:
That could be an endless loop in sending commands... do you have other switches that you control (non-lighting 4, but maybe lighting 1 or 2) ... do they still work correctly ?

gysmo38:
yes i have other switch that use blyss and somfy. It work well if I not use lighting4. If I use one lighting4 switch it stop working for all rfxcom devices.

PR #2681, opened 2018-08-26
Fix non threadsafe HW shutdown

emontnemery:
Many HW classes shuts down the connection before the worker thread. This is unsafe and causes crashes.
Also add forgotten join() of the worker in MochadTCP.

### DIFF
--- a/hardware/Tellstick.cpp
+++ b/hardware/Tellstick.cpp
@@ -238,7 +238,11 @@ bool CTellstick::StopHardware()
     m_bIsStarted = false;
     m_cond.notify_all();
     lock.unlock();
-    m_thread.join();
+    if (m_thread)
+	{
+		m_thread->join();
+		m_thread.reset();
+	}
     return true;
 }
 
--- a/hardware/Tellstick.cpp
+++ b/hardware/Tellstick.cpp
@@ -238,8 +238,7 @@ bool CTellstick::StopHardware()
     m_bIsStarted = false;
     m_cond.notify_all();
     lock.unlock();
-    if (m_thread.joinable()) //TODO: Why do we only join the thread if it's joinable?
-        m_thread.join();
+    m_thread.join();
     return true;
 }
 
--- a/hardware/Tellstick.cpp
+++ b/hardware/Tellstick.cpp
@@ -238,7 +238,7 @@ bool CTellstick::StopHardware()
     m_bIsStarted = false;
     m_cond.notify_all();
     lock.unlock();
-    if (m_thread.joinable())
+    if (m_thread.joinable()) //TODO: Why do we only join the thread if it's joinable?
         m_thread.join();
     return true;
 }
--- a/hardware/CurrentCostMeterTCP.cpp
+++ b/hardware/CurrentCostMeterTCP.cpp
@@ -68,6 +68,12 @@ bool CurrentCostMeterTCP::isConnected()
 
 bool CurrentCostMeterTCP::StopHardware()
 {
+	m_stoprequested=true;
+	if (m_thread)
+	{
+		m_thread->join();
+		m_thread.reset();
+	}
 	if (isConnected())
 	{
 		try {
@@ -111,16 +117,10 @@ bool CurrentCostMeterTCP::ConnectInternal()
 
 void CurrentCostMeterTCP::disconnect()
 {
-	m_stoprequested=true;
 	if (m_socket==INVALID_SOCKET)
 		return;
 	closesocket(m_socket);
 	m_socket=INVALID_SOCKET;
-	sleep_seconds(1);
-	if (m_thread)
-	{
-		m_thread->join();
-	}
 }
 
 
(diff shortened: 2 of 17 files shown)


hardware/Tellstick.cpp
```diff
@@ -238,7 +238,7 @@ bool CTellstick::StopHardware()
     m_bIsStarted = false;
     m_cond.notify_all();
     lock.unlock();
-    if (m_thread.joinable())
+    if (m_thread.joinable()) //TODO: Why do we only join the thread if it's joinable?
```
gizmocuz: this line can be removed

emontnemery:
OK, I fixed the review comments.
Please note this needs to merged at the same time as #2680.

emontnemery:
@gizmocuz Rebased, should be ready for merge!

hardware/Tellstick.cpp
```diff
@@ -238,8 +238,7 @@ bool CTellstick::StopHardware()
     m_bIsStarted = false;
     m_cond.notify_all();
     lock.unlock();
-    if (m_thread.joinable()) //TODO: Why do we only join the thread if it's joinable?
-        m_thread.join();
+    m_thread.join();
```
gizmocuz: Everywhere in the code we perform

```
if (m_thread)
{
 m_thread,join();
}
```

Of course it is safe (now....!) to assume there is always a thread.. if StartHardware is called... but who knows in the feature...

gizmocuz: and a m_thread.reset

gordonb3:
I have a feeling that the assumption being made here is not correct. If it is unsafe for the worker to access a closed connection, which will also affect mainworker accessing WriteToHardware(), that issue should be targeted directly within the hardware class itself because this situation may also be triggered by the remote at any given time.

What causes the thread to finish is this:
```
m_stoprequested = true;
```

Depending on the hardware class that may actually take some time though, particularly if locking I/O operations are used by that worker. If there are crashes during hardware shutdown - which I have not seen for a long time - I'd say the most likely cause for that will be the instant `join` and `reset` after the informing the worker it should finish what it is doing, in effect expecting that the stop will be instant.

The logical solution for this being to let the worker report back that it has in fact finished, e.g. by setting a boolean value `m_stopreached` to true. Or.... Didn't we already have a signal for that?

emontnemery:
@gordonb3 There is no race if there is a write request while the HW is shutting down. Such write will never be sent to the HW because the HW is deleted from the m_hardwaredevices before calling stop:
Please see here:
```
void MainWorker::RemoveDomoticzHardware(CDomoticzHardwareBase *pHardware)
{
	// Separate the Stop() from the device removal from the vector.
	// Some actions the hardware might take during stop (e.g updating a device) can cause deadlocks on the m_devicemutex
	CDomoticzHardwareBase *pOrgHardware = NULL;
	{
		std::lock_guard<std::mutex> l(m_devicemutex);
		std::vector<CDomoticzHardwareBase*>::iterator itt;
		for (itt = m_hardwaredevices.begin(); itt != m_hardwaredevices.end(); ++itt)
		{
			pOrgHardware = *itt;
			if (pOrgHardware == pHardware) {
				m_hardwaredevices.erase(itt);
				break;
			}
		}
	}

	if (pOrgHardware == pHardware)
	{
		pOrgHardware->Stop();
		delete pOrgHardware;
	}
}
```

(Edit since I misunderstood the point at first)

gordonb3:
I think you are purposely trying not to understand what I'm saying.

But granted, I do in fact not know all the ins and outs of every possible library that may be pulled in. That doesn't change anything though about the fact that this PR does not prevent a crash from the worker accessing a socket that was closed for another reason than a stop being requested.

In effect I think this PR is about trying to mask a problem that should not be a problem and probably isn't, but in the rare case that it might be it makes more sense to address the actual issue than try to mask it.

PS Am I correct in understanding that you are saying that `thread::join()` is a locking operation?

hardware/Tellstick.cpp
emontnemery: OK, fixed!

emontnemery:
@gordonb3 No offence meant, I misunderstood your comment at first.
But yeah, exactly, std_thread::join() is a synchronous/blocking call which will wait for the thread to exit.

> this PR does not prevent a crash from the worker accessing a socket that was closed for another reason than a stop being requested.

Right, this PR prevents crashes from the worker accessing a socket that was closed for the reason of a stop being requested. Is there some other case where a thread other than the worker is messing with the socket while the thread is running?

emontnemery:
@gizmocuz hardware/Tellstick.cpp updated according to your comment

gizmocuz:
thread::join is indeed a synchronous operation. I also agree that hardware classes should cleanup there open IO connections.
I see a lot of classes there after a join/reset a 'terminate' routine is called to close a serial connection.
This should be done in the worker thread.
I will try as well to make a start for signalling a thread to stop instead of a Boolean as discussed in another issue

gizmocuz:
Thanks!!

emontnemery:
> I will try as well to make a start for signalling a thread to stop instead of a Boolean as discussed in another issue

Did you start working on this already?

gizmocuz:
@emontnemery , no not yet, hopefully today with one or two to start with

PR #2682, opened 2018-08-26
Use std::strtoull / std::strtoll instead of std::stringstream

emontnemery:
Like #2657, but now use std::strtoull / std::strtoll which don't throw exceptions on bad input


gizmocuz:
Thanks! But is this an expensive function ? (cpu wise ?) i suspect this might be needed on only a few places

emontnemery:
No, it's not CPU intensive.
In fact, std::stoull /  std::stoll are wrapper functions for std::strtoull / std::strtoll

gizmocuz:
Thanks!!

PR #2683, opened 2018-08-26
Fix Ec3kMeter crash on large integer

### LINKED ISSUE #2678: Ec3kMeter may crash on certain ec3k readings
Using V4.9913 I got this error when enabling Ec3kMeter. It was able to connect to ec3k-server but Ec3kMeter crashes after a few sensor readings with the following message.

2018-08-19 17:49:32.373  Status: Ec3kMeter: connected to: 127.0.0.1:3001
terminate called after throwing an instance of 'Json::LogicError'
2018-08-19 17:49:34.270  (EC3K) General/kWh (EC3K meter e2a5 Usage)
  what():  LargestUInt out of Int range
Aborted

I will try to find the root-cause of this issue.

@llagendijk

Batilan:
Ec3kMeterTCP::ParseData will crash on large integers for data[WS] (which will happen after some time as the unit is Watt-seconds). Simply converting the Int to Float will prevent these range issues while the math can stay the same. Fixes #2678

### DIFF
--- a/hardware/Ec3kMeterTCP.cpp
+++ b/hardware/Ec3kMeterTCP.cpp
@@ -260,7 +260,7 @@ void Ec3kMeterTCP::ParseData(const unsigned char *pData, int Len)
 	// update only when the update interval has elapsed
 	if (m_limiter->update(id))
 	{
-		float ws = data[WS].asFloat();
+		double ws = data[WS].asDouble();
 		float w_current = data[W_CURRENT].asFloat();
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thanks!
Maybe it is better to use asDouble() , as later in the code, this value is handled as a double ? (/3600/1000)

Batilan:
OK, will update it accordingly. Currently not able to test, will come back on this later.

gizmocuz:
Thanks!!

PR #2684, opened 2018-08-27
rewrite P1MeterTCP to use ASyncTCP

gordonb3:


gizmocuz:
The logic in Do_Work is a bit complex with the milliseconds ... an introduction to a second variable might be better a sec_counter and a msec_counter.
Could you have a look at the 'OTGWTCP.cpp' implementation, specially at the 'bool bFirstTime=true;'

gordonb3:
I actually started off with sleep_second(1) but then I remembered that the ESP module that is mostly used for this, fragments the P1 datagrams. While it is okay for my DSMR 4 meter to process the full datagram 1 second later, that kind of gets weird with the ESMR 5+ meters that send every second. Hence the smaller interval.

`bFirstTime` is in fact obsolete when you increment the counter value at the end of the loop in stead of at the beginning of the loop.

IMO using the sec_counter for reconnect as done in `OTGWTCP` (and others) is wrong because you can't predict what value that counter has when the connection drops out for whatever reason. At modulo 30 the first reconnect attempt may therefore occur anywhere between 0 and 30 seconds. The way it is written here that will always be at 29 seconds (because retry counter will be at +1 from a successful connect)

Last, I found that this:
```
				connect(m_szIPAddress,m_usIPPort);
			}
			update();
```
doesn't work well. The connection would only register after the second connect attempt (i.e. after another 30 seconds) at which time it would log being connected twice. Another option is to add a `sleep()` after `connect()` but I think this is a neater method.

gordonb3:
Ah, hold on.

You are referring to this construct?
```
(RETRY_DELAY_SECONDS * 1000 / SLEEP_MILLISECONDS)
```

That is really meant to keep the defines on top comprehensible. It also makes them independent of each other in the sense that you can change the update interval without needing to adjust the heartbeat value for the counter being either hastened or slowed. Of course this is all preprocessor stuff - the final code will contain the single value `150` for this highlighted example.

gizmocuz:
Thanks!!

PR #2685, opened 2018-08-28
Default value 6.4 confusing after creating device

sweetpants:
After creating a "Ampere" device the default value 6.4 A is confusing https://www.domoticz.com/forum/viewtopic.php?f=6&t=24647&p=189946#p189937


gizmocuz:
Thanks, of course it was a virtual sensor and the value needs to be updated externally

PR #2694, opened 2018-08-31
Convert DomoticzTCP hardware class to use ASyncTCP

gordonb3:


gizmocuz:
Thanks! I will wait a bit with this patch as there is another patch waiting that is making changes to this class.

gizmocuz:
I was afraid of this... there are some merging issues caused by accepting PR 2880/2881

gordonb3:
Saw that. Was unavoidable due to the nature of this patch. Would have been better if the changes to these two files had been reverted in those PRs

gizmocuz:
I am not to keep about that change in DomoticzTCP either, maybe it's best to make a copy of your modified files, revert to the repo, and copy your files back and update this PR

gordonb3:
I'll have a look. Seems I should implement 2680.

Was having a discussion with Erik. Still don't think 2681 solves anything and otherwise this would still be a problem with WriteToHardware() that is accessed from mainworker. We both do agree though that there is something fishy with the ASyncTCP update() method. We think this is in reality polled I/O.

gordonb3:
Just noticing something: is it right to set m_bIsStarted to true if StartHardware() returns false?

gordonb3:
Seems to be something wrong with the switch to RFXBase:

```
2018-09-01 10:08:19.363  Error: RFXCOM: input buffer out of sync, going to restart!....
2018-09-01 10:08:19.363  Error: UNHANDLED PACKET TYPE:      FS20 36
```

Never mind. overlooked a reference to what was already a pointer

emontnemery:
@gordonb3  _Still don't think 2681 solves anything and otherwise this would still be a problem with WriteToHardware() that is accessed from mainworker._

#2681 solves a threading error when shutting down HW
As you correctly state, accessing WriteToHardware() is _also_ an issue because the asio::socket class is not thread safe (https://www.boost.org/doc/libs/1_67_0/doc/html/boost_asio/reference/ip__tcp/socket.html). Hence, thread safe WriteToHardware() should be implemented _as well_ as #2681, and I'm currently working on it by refactoring the ASyncTCP class to use signals instead of direct calls.

gizmocuz:
@gordonb3 , no 'm_bIsStarted' should not be set to true is starthardware cannot succeed

gordonb3:
Figured as much. Needed to ask though because the following is a common construct in practically every (if not all) hardware class:

```
	m_bIsStarted = true;

	//Start worker thread
	m_thread = std::make_shared<std::thread>(&Arilux::Do_Work, this);
	SetThreadName(m_thread->native_handle(), "Arilux");

	return (m_thread != nullptr);
```

gizmocuz:
@gordonb3 , that can never happen, we could also have put an asset there, so not, this i consider still valid

gizmocuz:
@gordonb3 , needed to rebuild ASyncTCP ... causing this class to 'need' some small additional changes (Callbacks, RequestStop, IsStopRequested, ...) Please see RFXComTCP.cpp for an example, also the Boolean 'm_stoprequested' is obsolete

gordonb3:
Seems stuff got added and then removed, leaving the fact that you changed mIsConnected from `protected` to `private`.

That causes a conflict with the MyDomoticz code in this class, both the existing and the new one, because the `proxy client` wants the `IsConnected()` function to return something different than `mIsConnected` and we therefore need to override this function but still allow it to return the value of `mIsConnected`.

Really nothing for me to adapt here. Only way I can make that work is to change ASyncTCP.h again.

gordonb3:
Alternative is to give the MyDomoticz hardware type its own class.

gizmocuz:
@gordonb3 , hmmm a seperate class sounds good too.
I am not sure what you mean by returning something different with 'IsConnected' ? this is a function from ASyncTCP and returns the boolean mIsConnected.
If you class has the same function somewhere, then i would suggest to rename that function

gordonb3:
`IsConnected` is called from webserver/proxyclient. I'm currently unsure how far that stretches

gordonb3:
May also like to note that `mDoReconnect` is not a state variable but a control variable. Which becomes rather pointless if it can't be changed.

gizmocuz:
@gordonb3 , it is changes with connect/disconnect

gordonb3:
Don't think so. The only place I see it being assigned a value is at the class construct. Just to be clear, I mean this one:
```

		if (!mDoReconnect)
		{
			if (mAllowCallbacks)
				OnDisconnect();
			return;
		}
```
This value must be allowed to be changed, because there are situations where you cannot just do a reconnect, e.g. when a portmapper is being used.

gizmocuz:
@gordonb3 , oh yes, that value (was confused with the connected value), yes, mentioned before, we can enrich the class. Let's get a great working class first, then add extra functionality.
It could/should be part of the connect call, does not make sense at the moment to create a dedicated function for this. Also when setting the reconnect duration to zero it should not reconnect

gordonb3:
As this isn't leading anywhere I'm closing this PR

I'll create a new one using the functions from csocket

PR #2696, opened 2018-09-01
Fixed MCP23017 initializacion issue with inverse logic

gneluka:
Fixed the inverse logic initialization issue with MCP23017


gizmocuz:
Thanks!

PR #2697, opened 2018-09-01
Refactor ASyncTCP class

emontnemery:
Changes:
- No need to poll
- Thread safe write() since WriteToHardware() is accessed from mainworker as pointed out by @gordonb3
- Fix access modifiers: ASyncTCP does not need any public members or methods, only private and some protected (for use in the derived HW classes)
- Refactor all derived HW classes, including removal of dead / useless code from Do_Work() loop

### DIFF
--- a/hardware/ASyncSerial.cpp
+++ b/hardware/ASyncSerial.cpp
@@ -127,7 +127,6 @@ void AsyncSerial::open(const std::string& devname, unsigned int baud_rate,
     pimpl->open=true; //Port is now open
 }
 
-//TODO: Remove this function!
 void AsyncSerial::openOnlyBaud(const std::string& devname, unsigned int baud_rate,
 	boost::asio::serial_port_base::parity /*opt_parity*/,
 	boost::asio::serial_port_base::character_size /*opt_csize*/,
--- a/hardware/ASyncSerial.cpp
+++ b/hardware/ASyncSerial.cpp
@@ -127,6 +127,7 @@ void AsyncSerial::open(const std::string& devname, unsigned int baud_rate,
     pimpl->open=true; //Port is now open
 }
 
+//TODO: Remove this function!
 void AsyncSerial::openOnlyBaud(const std::string& devname, unsigned int baud_rate,
 	boost::asio::serial_port_base::parity /*opt_parity*/,
 	boost::asio::serial_port_base::character_size /*opt_csize*/,
--- a/hardware/ASyncTCP.cpp
+++ b/hardware/ASyncTCP.cpp
@@ -15,7 +15,7 @@ struct hostent;
 #define RECONNECT_TIME 30
 
 // TODO: Delete default constructor and require thread name to be set
-ASyncTCP::ASyncTCP()
+ASyncTCP::ASyncTCP(const std::string thread_name)
 	: mIsConnected(false), mIsClosing(false),
 	mSocket(mIos), mReconnectTimer(mIos),
 	mDoReconnect(true), mIsReconnecting(false),
@@ -26,12 +26,13 @@ ASyncTCP::ASyncTCP()
 
 	//Start IO Service worker thread
 	m_tcpthread = std::make_shared<std::thread>(boost::bind(&boost::asio::io_service::run, &mIos));
-	SetThreadName(m_tcpthread->native_handle(), "ASyncTCP");
+	SetThreadName(m_tcpthread->native_handle(), thread_name.c_str());
 }
 
 ASyncTCP::~ASyncTCP(void)
 {
 	disconnect();
+
 	// tell the IO service to stop
 	mIos.stop();
 	try {
@@ -45,14 +46,6 @@ ASyncTCP::~ASyncTCP(void)
 	{
 		//Don't throw from a Stop command
 	}
-}
-
-void ASyncTCP::update()
-{
-	if (mIsClosing)
-		return;
-	// calls the poll() function to process network messages
-	//mIos.poll();
 }
 
 void ASyncTCP::connect(const std::string &ip, unsigned short port)
@@ -99,12 +92,6 @@ void ASyncTCP::connect(boost::asio::ip::tcp::endpoint& endpoint)
 
 	mEndPoint = endpoint;
 
-	// restart IO service if it has been stopped
-	if (mIos.stopped()) {
-		_log.Log(LOG_ERROR, "ASyncTCP::connect: IO service stopped!");
-		//mIos.reset();
-	}
-
 	// try to connect, then call handle_connect
 	mSocket.async_connect(endpoint,
         boost::bind(&ASyncTCP::handle_connect, this, boost::asio::placeholders::error));
@@ -114,9 +101,6 @@ void ASyncTCP::disconnect()
 {
 	// tell socket to close the connection
 	close();
-
-	// tell the IO service to stop
-	//mIos.stop();
 
 	mIsConnected = false;
 	mIsClosing = false;
@@ -130,6 +114,27 @@ void ASyncTCP::close()
 	mIos.post(boost::bind(&ASyncTCP::do_close, this));
 }
 
+void ASyncTCP::read()
+{
+	if (!mIsConnected) return;
+	if (mIsClosing) return;
+
+	mSocket.async_read_some(boost::asio::buffer(m_readbuffer, sizeof(m_readbuffer)),
+		boost::bind(&ASyncTCP::handle_read,
+			this,
+			boost::asio::placeholders::error,
+			boost::asio::placeholders::bytes_transferred));
+}
+
+void ASyncTCP::write(const uint8_t *pData, size_t length)
+{
+	write(std::string((const char*)pData, length));
+}
+
+void ASyncTCP::write(const std::string &msg)
+{
+	mIos.post(boost::bind(&ASyncTCP::do_write, this, msg));
+}
 
 // callbacks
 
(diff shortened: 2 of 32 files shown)


emontnemery:
@gordonb3 Have a look at the changes to write() please
@gizmocuz Quite a lot of changes, please help to review

gordonb3:
Currently unable to verify the code for expected behaviour. I see you made `class name` a variable to allow ASyncTCP to use the child class name for logging. That will definitely allow for better identification of which hardware is failing at some given time.

Just wondering: if the hardware worker is just about maintaining the heartbeat, why do `sleep_seconds(1)` and not `sleep_seconds(12)` (i.e. heartbeat interval)?

Also, could you make retry interval a variable too? Or at least add the option to silence it? I'm not a big fan of that fixed value, especially when that causes me to receive an error every 30 seconds when the reason for Domoticz not being able to communicate with that device is not a communication issue. Think maintenance or simply being shut down.

What is the connect timeout on the boost::asio procedures anyway? If I remember correctly Windows defaults to 72 seconds on this.

emontnemery:
@gordonb3 Thanks for very good points!

> Just wondering: if the hardware worker is just about maintaining the heartbeat, why do sleep_seconds(1) and not sleep_seconds(12) (i.e. heartbeat interval)?

It's not so good because mainworker then needs to wait for 12s in order to shutdown the HW.
With just a few TCP connected HW, this easily means waiting time of a minute when shutting down Domoticz. On the other hand, when the hardware worker has no real work, the loop could be replaced with a periodic timer.

> Also, could you make retry interval a variable too? Or at least add the option to silence it? I'm not a big fan of that fixed value, especially when that causes me to receive an error every 30 seconds when the reason for Domoticz not being able to communicate with that device is not a communication issue. Think maintenance or simply being shut down.

You mean the length of the reconnect period, right?
Are you suggesting something automatic in the ASyncTCP class like:
- 1st try: sleep 15s, then try reconnect
- 2nd try: sleep 30s, then try reconnect
- 3rd try: sleep 60s, then try reconnect
- 4th try: sleep 120s, then try reconnect
- 5th..nth try: sleep 240s, then try reconnect

> What is the connect timeout on the boost::asio procedures anyway? If I remember correctly Windows defaults to 72 seconds on this.

The timeout depends on the type of error.. 
The very long timeouts is an issue if the TCP packets can be routed, but the receiver just drops the packet.
For local devices, for example:
- If the device is not there, it will be much quicker (no MAC is matching the IP)
- If the device is up, but the service is down, it will fail instantaneously (connection refused)

gordonb3:
Hmmm, in hindsight...

If you set that to 12 seconds that will likely also mean that it may take up to 12 seconds for the hardware to shut down when instructed. For some more snappiness you'd actually need to make the sleep shorter.

Retry:
Yeah something like that. Possibly not as a general preconfigured sequence for every hardware type though as this consideration might not always apply.

gizmocuz:
I think we have done the same... somehow

gizmocuz:
Okey, not 100%. I just changed the ASyncTCP class so it is using callbacks like the serial class.
This, so when we call 'terminate' (of course from the thread), there can be no message been send (data/disconnect/error/...)
Therefor i also needed to change all hardware classes using this class, so call 'terminate' the right way
Maybe this patch is obsolete now ?

emontnemery:
@gizmocuz Yeah, not really the same, I also did a big cleanup of the ASyncTCP class.
Are you planning on doing more changes in ASyncTCP?
If not, I'll rebase this PR.

gizmocuz:
@emontnemery , i was thinking about partly revert my changes. Making booleans to tell if an 'override' should be called, which is default enabled, but when calling terminate, they will be disabled. This way, also, there will no callbacks/overrides being called when closing down
I will do this now, hold on a sec

gizmocuz:
Done, found some other issues as well

Simplifying ASyncTCP usage
Fixed double IsConnected call
Classes don't directly use private variables from ASyncTCP
Fritzbox now uses it's own buffer instead of ASyncTCP!
Moved and removed correct defines between headers (BaseHardware, RXBase)

gizmocuz:
Soon (Not tonight) i will implement the stop signal in all threads so there will be changes in all conflicted classes again

emontnemery:
@gizmocuz Right, I did quite a lot of cleanup..
 So, should I wait until you're done with the stop signal, or are you OK to merge this first after a rebase (tomorrow)?

gizmocuz:
I am not a fan of the thread in ASyncTCP, should not be needed.
Class works perfectly now.

emontnemery:
> I am not a fan of the thread in ASyncTCP, should not be needed.

The thread is sort of needed though, otherwise we're just polling for data, which is slow (need to wait until next poll) and a waste of CPU resources.
I was thinking to go the opposite direction: Remove the worker thread from the derived classes, let them  use the ASyncTCPs IO thread for the periodic timers (mostly heartbeat, and in very few cases some additional polling of data).

gordonb3:
@gizmocuz Your change doesn't look good to me. As mentioned in PR2694 Erik and I concluded that the AsyncTCP class is in fact not Async but polled. Which causes various timing issues, specifically with connect() as addressed in PR2684 for P1MeterTCP. Latter one being in fact what got us on track.

gizmocuz:
@emontnemery , no please keep the workers in the 'worker' classes.
In 90% of the cases it is data polling.
I am not a fan of including other classes (helper.h/log.h), SetThreadName etc.

@gordonb3 , my change ? My change was about making sure no messages are send (callbacks/overrides), nothing about threads/polling. Or you meant my comment ?

gizmocuz:
@emontnemery , okey, a thread sounds good, like ASyncSerial, but don't name things and do other stuff.
Want to keep this class very clean, i think i want to remove the logging here as well. (Should be handled by the overrides anyway)

emontnemery:
@gizmocuz:

> In 90% of the cases it is data polling

Yeah, but what about the other 10%?
For example in case of RFLink or RFXCom, both classes support receiving input from remote controls.

In RFLink:
Poll once a second, this means the remote control response time is completely unacceptable but CPU waste is negligible.

In RFXCom:
Poll every 40 ms, this means acceptable remote control response time, but a significant waste of CPU resources.

With proper - **non polled** - I/O these issues go away.

gizmocuz:
I agree, see above, a separate thread in the ASyncTCP. Classes that need to do polling, need their own thread, else no thread.
No funky stuff with overrides that are called from the TCP thread where a hardware class can do their stuff in. Then its best to keep it clean

Thanks

emontnemery:
@gizmocuz OK, fair enough! I'll rework this PR then :)

gizmocuz:
Great! I am going to work on serial classes for now ;)

emontnemery:
@gizmocuz Ah, I opened a PR on some changes I already started in the AsyncSerial class: #2699

gizmocuz:
@emontnemery , Oeps sorry yes just noticed after i committed my work...
Same comments regarding threadnames/helper/log here

gordonb3:
@gizmocuz Sorry. starting to get confused about who did what or did not do, or dislikes what appears to make sense but admittedly doesn't necessarily look nice.

In any case, ASyncTCP at present is not async so the statement that it works perfectly doesn't really apply. If it were really async then derived classes could become a lot simpler. In fact most of them would not need a worker at all, except for maintaining the heartbeat which is effectively bogus in this case.

emontnemery:
@gizmocuz 

> Want to keep this class very clean, i think i want to remove the logging here as well.

I don't really agree with this...
Isn't it better if ASyncTCP is a "complete" base class which it is very easy to build a HW integration from?
If each derived class event must invent their own error message for everything that can possibly go wrong, that just seems like unecessary code duplication?

gizmocuz:
@gordonb3 , the way the TCP class works now, is workable, but could be simplified with a thread.

@emontnemery , true, a stand-alone-class, not depending on helper/log etc.
Using overrides like now, but with an internal thread for the polling.
(could also use another way like the webserver does but ok...with a separate thread it works)
So any class that does inherit ASyncTCP must override 5 functions (like now), thats it

gizmocuz:
@emontnemery , yes, every class needs to handle connect/disconnect/data and error messages for sure

gordonb3:
@gizmocuz 
Workable isn't exactly perfect as well. Speaking for myself I found the behaviour I saw in `connect()` to be very unexpected and you said you thought the additional `update()` loop that is required to make that work the way it is supposed to was `complex`. Which I can agree with to some extend because I did in fact not intend to do it that way, but I saw no better way around it. It's something of a catch-22.

@emontnemery 
I actually think it is better to let the derived class do its own error handling. Because not every error is an error and sometimes need to be interpreted differently. An example of this can be found in 7aa6bf9 where a specific occurrence of EOF is used to determine a login failure. Which is of course specific to that particular hardware.

gizmocuz:
@gordonb3 , remember that is not my class... Anyway, let us do the refactoring first and it would be great if you can review it once we're done

emontnemery:
Closing this one and open #2701 instead since the scope changed a bit due to the parallel changes by @gizmocuz

emontnemery:
@gizmocuz "a stand-alone-class, not depending on helper/log etc."

I'm sorry, but I don't understand this reasoning. I see the ASyncTCP is a base class to make it easy and simple to implement support for network connected HW, and also keep the tricky code in one place without duplication.
It seems unnecessarily restrictive if the base class shall not be allowed to aid debugging by logging or setting thread name?

gizmocuz:
@emontnemery , when the class is OK, no need for that extra debug stuff.
Want to keep it clean.

PR #2698, opened 2018-09-01
Add scan of /dev/serial/by-id/* and /dev/serial/by-path/* on Linux

JonathanLondon:
Modern Linux systems allocate /dev/ttyUSB* in an unpredictable order at boot time.
This requires udev rules for consistency at reboot otherwise Domoticz doesn't reliably find USB devices.
This change exposes /dev/serial/by-id/* and /dev/serial/by-path/* when configuring USB devices, which are always predictable.
As a result, the Domoticz user does not have to dive down to the Linux command line to create udev rules, making it a less daunting experience.
See http://www.domoticz.com/forum/viewtopic.php?f=6&t=24619

### DIFF
--- a/main/Helper.cpp
+++ b/main/Helper.cpp
@@ -196,7 +196,7 @@ std::vector<std::string> GetSerialPorts(bool &bUseDirectPath)
 
 #else
 	//scan /dev for /dev/ttyUSB* or /dev/ttyS* or /dev/tty.usbserial* or /dev/ttyAMA* or /dev/ttySAC*
-	//also scan /dev/serial/by-id/* and /dev/serial/by-path/* on Linux
+	//also scan /dev/serial/by-id/* on Linux
 
 	bool bHaveTtyAMAfree=false;
 	std::string sLine = "";
@@ -314,8 +314,8 @@ std::vector<std::string> GetSerialPorts(bool &bUseDirectPath)
 		// Loop while not NULL
 		while ((de = readdir(d)))
 		{
-			// Only consider character devices and symbolic links
-                        if ((de->d_type == DT_CHR) || (de->d_type == DT_LNK))
+			// Only consider symbolic links and limited to 99 chars incl prefix i.e. 81 chars
+                        if ((de->d_type == DT_LNK) && (strlen(de->d_name) <= 81))
                         {
 				std::string fname = de->d_name;
 				ret.push_back("/dev/serial/by-id/" + fname);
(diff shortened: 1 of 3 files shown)


gizmocuz:
Thank you for your patch!
I have a few issues with this, but i like the idea.
I don't think non-unix minded users would know what to do with the result of /dev/serial/by-path
So maybe better drop this, as /dev/serial/by-id should be self explaining

Then we have to make sure that the result of 'by-id' has a string that has no more then 49 characters as in the database this field is limited (hardware table)
Which we can not, as then we have to chop of characters.

We need to create a database patch to increase the field length to 100 characters, which we can do in SQLHelper.cpp with a temp table
And for this patch to work, you also need to increase the width of the 'Serial Port' dropdown in the hardware setup as it is to small now

JonathanLondon:
Hi

Thank you for your constructive feedback!

I did wonder about the by-path devices.  What I don't know (and am 
unable to test) is what happens if you have TWO devices of the same type 
and serial number.  The RFXtrx433 has a unique serial number so that's 
fine, but the ZWave stick is serial number 0001 (thanks Aeotec).  I 
don't know whether their entries in by-id would be invariant across 
reboots for devices without a unique serial number - I am guessing not.  
Do you have two of anything, in order to test?  The by-path device names 
relate to the physical sockets, so these would definitely be invariant, 
which is why I included them.  I appreciate they are also confusing!

So now I can reveal my complete ignorance of the code base, and of 
AngularJS.  It would be nice if the drop-down could just list the 
devices (from by-id) and the physical USB sockets (from by-path) .  
Indeed it would be great if it could present "RFXtrx433 USB socket 2" 
and do the right thing behind the scenes.  But that's a step too far for 
me right now!

In terms of the database schema, I suggest that technically the field 
should be MAXPATHLEN bytes long, or simply a string rather than a 
varchar.  A slightly dirty hack would be to only present devices which 
are less than 49 (or 100) bytes long.

I am not promising I can spend much time on this, although I am keen 
to.  I would welcome your feedback on what I have written above.

Kind regards
Jonathan.

gizmocuz:
Thanks for the PR !
After doing some reading on the differences between VARCHAR and TEXT in relation with SQLite, there seems to be no difference and both have the same size (independent on the specified length of the VARCHAR)
For this reason, it might be a good idea to replace VARCHAR with TEXT and skip the length limit ?
If you want to keep the length limit, i think we should log if some serial devices will be skipped

JonathanLondon:
OK, I have switched to a TEXT and removed the length check.  I have also 
bumped the DBversion due to a clash with another change.

I am struggling a little to merge versions (not enough experience with 
github).  Hopefully you can merge my patch.

gizmocuz:
Thanks for the changes !
Yes,i also see you reply via email on git...

Next time (because i will merge this of course!, and thanks!!), when you create a PR, you can just make changes, and keep committing them like normal. Then the PR will be automatically updated

JonathanLondon:
Hi

I am hopelessly confused by GitHub.

I have just submitted an SMTP patch.  Please let me know if you can see it.

http://www.domoticz.com/forum/viewtopic.php?f=6&t=24120

Very many thanks

Jonathan.

gizmocuz:
Please do not reply with your email client but use github directly.
I do not see a pull request for the changes.

In your repo, delete the domoticz fork
fork domoticz again (from the domoticz repo with the 'fork' button) to your account
Checkout your forked domoticz, make the changes, commit the changes
On your forked domoticz, you should see a 'create pull request' button
press this button, a PR is made
If you need to commit more changes, just make them, commit them, thats it

PR #2701, opened 2018-09-03
Refactor ASyncTCP class to not require polling

emontnemery:
This PR replaces #2697 

Changes:
- No need to poll
- Thread safe write() since WriteToHardware() is accessed from mainworker (as pointed out by @gordonb3)
- Fix access modifiers: ASyncTCP does not need any public members or methods, only private and some protected (for use in the derived HW classes)
- Add documentation in ASyncTCP.h
- Simplify DoWork() of all derived classes since there is no longer any polling needed

### DIFF
--- a/hardware/ASyncTCP.cpp
+++ b/hardware/ASyncTCP.cpp
@@ -33,16 +33,10 @@ ASyncTCP::~ASyncTCP(void)
 
 	// tell the IO service to stop
 	mIos.stop();
-	try {
-		if (m_tcpthread)
-		{
-			m_tcpthread->join();
-			m_tcpthread.reset();
-		}
-	}
-	catch (...)
+	if (m_tcpthread)
 	{
-		//Don't throw from a Stop command
+		m_tcpthread->join();
+		m_tcpthread.reset();
 	}
 }
 
--- a/hardware/ASyncTCP.cpp
+++ b/hardware/ASyncTCP.cpp
@@ -4,8 +4,6 @@
 #include <boost/asio.hpp>
 #include <boost/bind.hpp>
 #include <boost/system/error_code.hpp>     // for error_code
-#include "../main/Helper.h"                // for SetThreadName
-#include "../main/Logger.h"                // for CLogger, _log, _eLogLevel:...
 struct hostent;
 
 #ifndef WIN32
@@ -27,7 +25,6 @@ ASyncTCP::ASyncTCP()
 
 	//Start IO Service worker thread
 	m_tcpthread = std::make_shared<std::thread>(boost::bind(&boost::asio::io_service::run, &mIos));
-	SetThreadName(m_tcpthread->native_handle(), ASYNCTCP_THREAD_NAME);
 }
 
 ASyncTCP::~ASyncTCP(void)
@@ -75,8 +72,9 @@ void ASyncTCP::connect(const std::string &ip, unsigned short port)
 			}
 			else
 			{
-				//we will fail
-				_log.Log(LOG_ERROR, "TCP: Unable to resolve '%s'", fip.c_str());
+				// Failed to resolve hostname
+				if (mAllowCallbacks)
+					OnError(boost::system::errc::make_error_code(boost::system::errc::host_unreachable));
 			}
 		}
 
@@ -88,7 +86,6 @@ void ASyncTCP::connect(const std::string &ip, unsigned short port)
 	{
 		if (!mAllowCallbacks)
 			return;
-		_log.Log(LOG_ERROR, "TCP: Exception: %s", e.what());
 		OnError(e);
 	}
 }
@@ -136,7 +133,6 @@ void ASyncTCP::StartReconnect()
 	if (m_reconnect_delay != 0)
 	{
 		mIsReconnecting = true;
-		_log.Log(LOG_STATUS, "TCP: Reconnecting in %d seconds...", m_reconnect_delay);
 		// schedule a timer to reconnect after xx seconds
 		mReconnectTimer.expires_from_now(boost::posix_time::seconds(m_reconnect_delay));
 		mReconnectTimer.async_wait(boost::bind(&ASyncTCP::do_reconnect, this, boost::asio::placeholders::error));
@@ -202,7 +198,6 @@ void ASyncTCP::handle_connect(const boost::system::error_code& error)
 
 		if (mAllowCallbacks)
 			OnError(error);
-		OnErrorInt(error);
 
 		if (!mDoReconnect)
 		{
@@ -302,7 +297,6 @@ void ASyncTCP::do_reconnect(const boost::system::error_code& /*error*/)
 	}
 	mReconnectTimer.cancel();
 	// try to reconnect, then call handle_connect
-	_log.Log(LOG_STATUS, "TCP: Reconnecting...");
 	mSocket.async_connect(mEndPoint,
         boost::bind(&ASyncTCP::handle_connect, this, boost::asio::placeholders::error));
 	mIsReconnecting = false;
@@ -320,6 +314,7 @@ void ASyncTCP::do_write(const std::string &msg)
 	}
 }
 
+/*
 void ASyncTCP::OnErrorInt(const boost::system::error_code& error)
 {
 	if (
@@ -342,3 +337,4 @@ void ASyncTCP::OnErrorInt(const boost::system::error_code& error)
 	else
 		_log.Log(LOG_ERROR, "TCP: Error: %s", error.message().c_str());
 }
+*/
--- a/hardware/ASyncTCP.cpp
+++ b/hardware/ASyncTCP.cpp
@@ -4,6 +4,7 @@
 #include <boost/asio.hpp>
 #include <boost/bind.hpp>
 #include <boost/system/error_code.hpp>     // for error_code
+#include "../main/Helper.h"                // for SetThreadName
 #include "../main/Logger.h"                // for CLogger, _log, _eLogLevel:...
 struct hostent;
 
@@ -11,41 +12,46 @@ struct hostent;
 	#include <unistd.h> //gethostbyname
 #endif
 
-/*
-#ifdef WIN32
-	#include <Mstcpip.h>
-#elif defined(__FreeBSD__)
-	#include <netinet/tcp.h>
-#endif
-*/
-
 #define RECONNECT_TIME 30
 
 ASyncTCP::ASyncTCP()
 	: mIsConnected(false), mIsClosing(false),
 	mSocket(mIos), mReconnectTimer(mIos),
 	mDoReconnect(true), mIsReconnecting(false),
+	m_tcpwork(mIos),
 	mAllowCallbacks(true),
 	m_reconnect_delay(RECONNECT_TIME)
 {
+	// Reset IO Service
+	mIos.reset();
+
+	//Start IO Service worker thread
+	m_tcpthread = std::make_shared<std::thread>(boost::bind(&boost::asio::io_service::run, &mIos));
+	SetThreadName(m_tcpthread->native_handle(), ASYNCTCP_THREAD_NAME);
 }
 
 ASyncTCP::~ASyncTCP(void)
 {
 	disconnect();
+
+	// tell the IO service to stop
+	mIos.stop();
+	try {
+		if (m_tcpthread)
+		{
+			m_tcpthread->join();
+			m_tcpthread.reset();
+		}
+	}
+	catch (...)
+	{
+		//Don't throw from a Stop command
+	}
 }
 
 void ASyncTCP::SetReconnectDelay(int Delay)
 {
 	m_reconnect_delay = Delay;
-}
-
-void ASyncTCP::update()
-{
-	if (mIsClosing)
-		return;
-	// calls the poll() function to process network messages
-	mIos.poll();
 }
 
 void ASyncTCP::connect(const std::string &ip, unsigned short port)
@@ -96,12 +102,6 @@ void ASyncTCP::connect(boost::asio::ip::tcp::endpoint& endpoint)
 
 	mEndPoint = endpoint;
 
-	// restart IO service if it has been stopped
-	if (mIos.stopped()) {
-		_log.Log(LOG_STATUS, "ASyncTCP::connect: reset IO service");
-		mIos.reset();
-	}
-
 	// try to connect, then call handle_connect
 	mSocket.async_connect(endpoint,
         boost::bind(&ASyncTCP::handle_connect, this, boost::asio::placeholders::error));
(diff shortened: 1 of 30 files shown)


hardware/ASyncTCP.cpp
```diff
@@ -4,50 +4,56 @@
 #include <boost/asio.hpp>
 #include <boost/bind.hpp>
 #include <boost/system/error_code.hpp>     // for error_code
+#include "../main/Helper.h"                // for SetThreadName
```
gizmocuz: Please undo this, and also the log option as discussed

hardware/ASyncTCP.cpp
```diff
@@ -156,70 +151,27 @@ void ASyncTCP::close()
...
+			this,
+			boost::asio::placeholders::error,
+			boost::asio::placeholders::bytes_transferred));
+}
 
-	// Set the timeout before the first keep alive message
-	int ret_sokeepalive = setsockopt(native_fd, SOL_SOCKET, SO_KEEPALIVE, (void*)&on, sizeof(int));
-	int ret_tcpkeepalive = setsockopt(native_fd, IPPROTO_TCP, TCP_KEEPALIVE, (void*)&timeout, sizeof(int));
-	int ret_tcpkeepintvl = setsockopt(native_fd, IPPROTO_TCP, TCP_CONNECTIONTIMEOUT, (void*)&intvl, sizeof(int));
+void ASyncTCP::write(const uint8_t *pData, size_t length)
```
gizmocuz, line 156: Please do not reverse the write operations. Write with uint8_t* make much more sense to be the default

hardware/ASyncTCP.h
```diff
@@ -8,61 +8,81 @@
...
 
 class ASyncTCP
 {
-public:
+protected:
 	ASyncTCP();
 	virtual ~ASyncTCP(void);
+	// Async connect and start async read from the socket, data is delivered in OnData()
+	// No action if mIsConnected = true. If the connection is lost, it will automatically be setup again
+	void connect(const std::string &hostname, unsigned short port);
```
gizmocuz, line 22: setting mIsConnected could be part of the connect call (maybe default false)

gordonb3:
As explained in #2694, please don't make the `state variables` private.

1. mDoReconnect is not a state variable, but a control variable
1. DomoticzTCP defines `IsConnected()` as a public function meant to be accessed by the My Domoticz link. This makes converting that class to async a lot more complicated or it will require something silly like maintaining a self-maintained duplicate (possibly tricky with shutdown).


Just tested with the prior code (just intermediate AsyncTCP.cpp and .h). Looks good, in particular connect() and read(). Not seeing much difference in write(), but I'd say that is also good.

gizmocuz:
I do not understand 100 the following commit details:

>>Thread safe write() since WriteToHardware() is accessed from mainworker (as pointed out by @gordonb3)
Maybe this is why there is a reverse of the write function, but i like to see where this is pointed out

>>Fix access modifiers: ASyncTCP does not need any public members or methods, only private and some protected (for use in the derived HW classes)

I thought all variables/functions where already made private with my previous patch ?

hardware/ASyncTCP.h
gordonb3: That would be assuming things. mIsConnected should and is set by `handle_connect()` when the peer responds to that request.

hardware/ASyncTCP.cpp
emontnemery: "mainworker" is not a good threadname which helps when debugging, so please suggest how to set the threadname without this dependency?

hardware/ASyncTCP.cpp
emontnemery: You're right, but the reason for reversing is that the data needs to be in a copyable container to be sent to the IO thread.

emontnemery:
@gordonb3 It's just asking for trouble to (re)expose the variables and use that as communication with the thread. If there's a need to expose some state variable, that should be done through thread safe getter/setter methods.

emontnemery:
@gizmocuz 
>> Thread safe write() since WriteToHardware() is accessed from mainworker (as pointed out by @gordonb3)

>Maybe this is why there is a reverse of the write function, but i like to see where this is pointed out

Yes, exactly! It's because we need a copyable container to do this:
```
void ASyncTCP::write(const std::string &msg)
{
	mIos.post(boost::bind(&ASyncTCP::do_write, this, msg));
}
```

>>Fix access modifiers: ASyncTCP does not need any public members or methods, only private and some protected (for use in the derived HW classes)
> I thought all variables/functions where already made private with my previous patch ?

Not really, check the diff of ASyncTCP.h @_@

gordonb3:
Shamefully it probably makes even less sense to rename `IsConnected()` for the My Domoticz link as it makes sense to rename it for this skeleton class. Due to the other change to conditionally disallow the callback methods I can also not reliably track what value this var is supposed to have at any given time.

In effect this change is blocking me in changing this:
```
			//this could take a long time... maybe there will be no data received at all,
			//so it's no good to-do the heartbeat timing here
			m_LastHeartbeat = mytime(NULL);

			int bread = recv(m_socket, (char*)&buf, sizeof(buf), 0);
```

Which is nothing short of TCP Hell

emontnemery:
@gordonb3 But that code is simply replaced by the async receive,  error, and disconnect callbacks.

emontnemery:
@gizmocuz I started removing the logs, agree most logs are redundant and can be replaced by logging in the derived classes.
What about this one though, this is really essential:
`_log.Log(LOG_ERROR, "TCP: Unable to resolve '%s'", fip.c_str());`

Also, please comment on how to set the threadname. If (just an example) there are is a deadlock in RFLink, it's really nice to have the threadname because all the callstacks will look the same.

gordonb3:
Nope...

You missed this bit:
```
bool DomoticzTCP::isConnected()
{
#ifndef NOCLOUD
	if (b_useProxy) {
		return isConnectedProxy();
	}
	else {
		return isConnectedTCP();
	}
#else
	return isConnectedTCP();
#endif
}
```

emontnemery:
@gizmocuz Maybe OnError() callbacks from ASyncTCP can be extended with a string which can be used for debug prints?

emontnemery:
@gordonb3 But the reworked ASyncTCP class has the IsConnected() function?

gordonb3:
Yes, but I can't use that. Because this is an endless loop:
```
#ifndef NOCLOUD
bool DomoticzTCP::isConnected()
{
	if (b_useProxy) {
		return isConnectedProxy();
	}
	return isConnected();  // return mIsConnected
}
#endif
```

emontnemery:
@gordonb3 Not only endless loop, but infinite recursion as well :)
I think you can do like this if you really don't want to rename DomoticzTCP::isConnected()
```
#ifndef NOCLOUD
bool DomoticzTCP::isConnected()
{
	if (b_useProxy) {
		return isConnectedProxy();
	}
	return ASyncTCP::isConnected();  // return mIsConnected
}
#endif
```

gordonb3:
It's not about wanting.

It's about needing to investigate every occurance of IsConnected() in every source file to find whether that is related to this so-called `proxy` method. Either way it leads to the same issue that one of the functions has to be renamed (to what?).

I'll verify that suggestion of yours. Don't like it though. Looks like shoestring operation. Ugly.

gordonb3:
Appears to work - on Linux anyway

Doesn't feel good though. Seems easier to just copy the functions from AsyncTCP and make them an integral part of DomoticzTCP

emontnemery:
@gordonb3 It's nothing strange, it's proper C++ syntax to disambiguate between the IsConnected() function in the derived class and the IsConnected() function in the super class.

gordonb3:
I know it is proper disambiguation. I think it will be **extremely** uncommon though to use this construct in an extended class.

IMO you and Rob are making an issue of something that is not an issue. Protected is not public and the ASyncTCP class in itself is rather pointless without being extended. There's really only one variable that could be considered vital enough to shield from the hardware class that would normally be able to access them all (if this skeleton class did not exist) and that is `mAllowCallbacks` which by the looks of it is nothing more than a reverse shadow of `IsStopRequested` - i.e. redundant.

emontnemery:
@gordonb3 I don't understand what makes DomoticzTCP so special that you need to mess with / have access to ASyncTCP states? But if it's really needed, just add DomoticzTCP as a friend of ASyncTCP class.

The reason for restricting access and making most members private is that:
- It gives you a well tested base, which can be trusted to work
- You can fix bugs in ASyncTCP without breaking other classes who are relying on implementation details

Anyway, maybe we can take this discussion in #2694 instead?

gordonb3:
Pointless. It's not about DomoticzTCP being special or not. That class just happens to contain a direct conflict with what you and Rob are doing.

You're also not responding to the fact that mDoReconnect is not a state variable, but a control variable. And no it is also no good to allow control over that variable through the `connect()` function because it may depend on the error state whether it is desirable to attempt a reconnect. Like a hang-up due to bad credentials.

emontnemery:
Edit: No need to expose mDoReconnect, you can call SetReconnectDelay with Delay = 0:
```
// Set reconnect delay in seconds, set to 0 to disable reconnect
void SetReconnectDelay(const int Delay = 30);
```

@gizmocuz this code is a bit strange IMHO:
```
if (!mDoReconnect)
{
	if (mAllowCallbacks)
		OnDisconnect();
	return;
}
```

In this way, since ASyncTCP is not allowed to log, how can the user know there was an error causing the socket to disconnect? Or it's implied by the OnError call that there has been a disconnection?

emontnemery:
@gizmocuz Updated as you requested, I hope this can be merged now:
- Debug prints are gone. In case there is a name resolution error, call OnError():
```
// Failed to resolve hostname
if (mAllowCallbacks)
	OnError(boost::system::errc::make_error_code(boost::system::errc::host_unreachable));

```
- Remove setting of thread name

hardware/ASyncTCP.cpp
```diff
@@ -4,50 +4,53 @@
...
+	m_tcpthread = std::make_shared<std::thread>(boost::bind(&boost::asio::io_service::run, &mIos));
 }
 
 ASyncTCP::~ASyncTCP(void)
 {
 	disconnect();
+
+	// tell the IO service to stop
+	mIos.stop();
+	try {
```
gizmocuz: I don't think we need a try/catch here...

gizmocuz:
Looking good, i only think the try/catch can be removed

gordonb3:
Does this comment apply here as well?

```
		m_Heartbeatthread->join();
		// Wait a while. The read thread might be reading. Adding this prevents a pointer error in the async serial class.
		sleep_milliseconds(10);
		m_Heartbeatthread.reset();
```

emontnemery:
@gordonb3 This:
```
// Wait a while. The read thread might be reading. Adding this prevents a pointer error in the async serial class.
sleep_milliseconds(10);
```
Makes absolutely no sense at all.
If there are bugs in AsyncSerial, they should be fixed there, not by sprinkling the code with delays which may or may not hide the bug.
(I'm being a bit snarky, but my point is that there is no guarantee whatsoever that the AsyncSerial thread is scheduled within the 10ms. This means, IF there is a bug, adding the delay just makes it harder to debug because it is now likely to appear when system load is high and the AsyncSerial thread is nto scheduled)

gordonb3:
I couldn't agree more, but who added it and why, and is the code in this PR unaffected by whatever caused that comment and code line to be added?

emontnemery:
Ah, but do you see that line in any file affected by this PR? I thought it's only in the classes deriving AsyncSerial?

gordonb3:
Nope. It's in the CDomoticzHardware class

emontnemery:
I say: Remove it :)

emontnemery:
@gizmocuz The unecessary try/catch is removed now.

gordonb3:
I dare not, sir ;)

Not without knowing the why and whether or not that still applies.

gordonb3:
You on the other hand might want to use that standardized function if you are converting the workers in hardware classes to just provide the heartbeat.

gordonb3:
Also, it may not be a bug. It may just be how the Beast works and my professional guess would be that even though with a serial connection you will be more likely to encounter this apparent race condition that doesn't mean it can't occur when any other type of communication is involved.

In any case the comment is too specific to be the result of being cautious.

gizmocuz:
@gordonb3 :
the wait miliseconds(10) was added by someone. This was to avoid a problem that a callback/override was called while terminating the thread.
The serial/tcp connection should be terminated inside the thread, so this can be removed

> // Wait a while. The read thread might be reading. Adding this prevents a pointer error in the async serial class.
> 		sleep_milliseconds(10);

I am going to refactor the hardware classes soon, i've seen you also started to do this unfortunataly
The heartbeat should be implemented correctly, if one does not use a thread to do polling, and is expecting data every xx seconds, then it's best to set the heartbeat timestamp in the receiving call.
But that's outside the scope of this PR

hardware/ASyncTCP.h
gizmocuz: Sorry, i meant the reconnect 'mDoReconnect' and the retry delay
Actually only the retry_delay, we do not need 'mDoReconnect'. When the retry_delay = 0, there won't be a reconnect

emontnemery:
Heartbeat is not same as expecting data though, there's a separate function for that? Heartbeat is do detect deadlock I thought, and as such that should be done by the io-worker thread which might very well deadlock?
Unless you're working on it already I'll open a PR to show how I'm thinking.

gizmocuz:
@emontnemery , no, but if you have a polling class, and there is a problem(deadlock, anything) with that code, sure you can not send the heartbeat
If you are 100% sure you should expect every 10 seconds data (like for the p1 smart meter), it is very fair to say that if you have not received any data for 1 minute, something is very wrong.
Anyway, that is completely outside the scope of the PR

gordonb3:
@gizmocuz 
Okay. May like to know that I spotted a hardware class or two that starts both the heartbeat tread from DomoticzHardware as its own worker. If I remember correctly LMS is one of them. Maybe Kodi also.

emontnemery:
@gizmocuz So, can this PR be merged now, or any outstanding issue?

emontnemery:
@gordonb3 that's really useless, the separate heartbeat thread is only useful if the hw does not have its own thread..
Or do those classes override the heartbeat thread such that it actually does work?

gizmocuz:
Yes, unfortunately i have reverted all my current changes to rebuild the remaining hardware classes to use the stop signal. Will redo this again after this.
Thanks for your work on this and patience with me ;)

gizmocuz:
@gordonb3 , we have indeed some very funky classes. Best to check that after every is changed to signals.
... some classes i will not touch now, you can find them later if you search for the variable 'm_stoprequested'

PR #2702, opened 2018-09-03
Fix for finding local ip for Xiaomi GW for e.g. FreeBSD

Whiskey24:
Atempts to parse ipconfig (ifconfig for WIN32) to determine ip address of domoticz host in case boost asio fails (as it does on FreeBSD).

This replaces #2609

### DIFF
--- a/hardware/XiaomiGateway.cpp
+++ b/hardware/XiaomiGateway.cpp
@@ -12,6 +12,7 @@
 #include <openssl/aes.h>
 #include <boost/asio.hpp>
 #include <boost/bind.hpp>
+#include <regex>
 
 /*
 Xiaomi (Aqara) makes a smart home gateway/hub that has support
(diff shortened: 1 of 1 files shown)


xbeaudouin:
Hello, 
I am not sure that using ifconfig from a module is really a good idea (thread safe stuff or whatever).

I used a long time ago this code (that still work on FreeBSD 11.2) :  

```
#include <sys/types.h>
#include <sys/socket.h>

#include <netinet/in.h>
#include <net/if.h>
#include <net/if_dl.h>

#include <arpa/inet.h>

#include <assert.h>
#include <ifaddrs.h>
#include <netdb.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>

char *mangle(char *addr)               /* addr is xx0:addr, strip xx0: bit */
{
  char *colon;

  colon = strchr(addr, ':');
  if (colon == NULL)
    return addr;                       /* no link addr */
  return colon + 1;
}

int main(int argc, char *argv[])
{
  struct ifaddrs *ifaddr;
  struct sockaddr *a;
  int ret;
  int specific;

  ifaddr = malloc(sizeof(struct ifaddrs));
  assert (ifaddr != NULL);

  ret = getifaddrs(&ifaddr);
  if (ret != 0) {
    perror("getifaddrs");
    exit(1);
  }
  specific = (argc == 2);

  /* ugly code follows, rewrite as needed */
  for (a = ifaddr->ifa_addr; ifaddr->ifa_next;
       ifaddr = ifaddr->ifa_next, a = ifaddr->ifa_addr)
    switch(a->sa_family) {
      case AF_LINK:
        if (!specific || (specific && strcmp(ifaddr->ifa_name, argv[1]) == 0))
          printf("%s link %s\n", ifaddr->ifa_name,
                 mangle(link_ntoa((struct sockaddr_dl *)a)));
        break;
      case AF_INET:
        if (!specific || (specific && strcmp(ifaddr->ifa_name, argv[1]) == 0)) {
          char hbuf[NI_MAXHOST], sbuf[NI_MAXSERV];

          getnameinfo(a, a->sa_len, hbuf, sizeof(hbuf), sbuf, sizeof(sbuf),
                      NI_NUMERICHOST | NI_NUMERICSERV);
          printf("%s inet %s\n", ifaddr->ifa_name, hbuf);
        }
        break;
      case AF_INET6:
        if (!specific || (specific && strcmp(ifaddr->ifa_name, argv[1]) == 0)) {
          char hbuf[NI_MAXHOST], sbuf[NI_MAXSERV];

          getnameinfo(a, a->sa_len, hbuf, sizeof(hbuf), sbuf, sizeof(sbuf),
                      NI_NUMERICHOST | NI_NUMERICSERV);
          printf("%s inet6 %s\n", ifaddr->ifa_name, hbuf);
        }
        break;
      default:

        if (!specific || (specific && strcmp(ifaddr->ifa_name, argv[1]) == 0)) {
          int i;

          printf("%s %d ", ifaddr->ifa_name, a->sa_family);
          for (i = 0; i < a->sa_len; i++)
            putchar(a->sa_data[i]);
          putchar('\n');
        }
    }
  exit(0);
}


```

Output is pretty good : 

```
bge0 link a0.48.1c.b8.58.c8
bge0 inet 192.168.0.9
bge0 inet6 fe80::a248:1cff:feb8:58c8%bge0
bge0 inet6 2001:67c:25e8:8888::9
bge1 link a0.48.1c.b8.58.c9
lo0 link lo0
lo0 inet6 ::1
lo0 inet6 fe80::1%lo0
```

Regards :)

Whiskey24:
Thanks for chiming in. I'm on very thin ice here (my c++ experience being limited to 2 Domoticz pull requests) , but I get your point about parsing ifconfig maybe not being the best idea. However, the consensus [here](https://stackoverflow.com/questions/1702442/popen-locks-or-not-thread-safe) seems to be that popen should be thread safe (or is that not the issue?).

That being said,  if your example really is the better way to do it, I'll happily give it a go to try and adapt it.

xbeaudouin:
Well I am not sure that FreeBSD's popen() thread safe or not. Maybe if I ask some FreeBSD dev I can get the reply. Anyway doing a popen forks /bin/sh and then /sbin/ifconfig with is not really light comparing of the usage of internal libc stuff (which is already in memory).
Maybe this function can be central to domoticz, what do you think of that @gizmocuz ?
Regards

Whiskey24:
@gizmocuz do you have a preferred approach for this? Getting the local ip must be something that is needed more often.

gizmocuz:
Sorry, did not follow this discussion as i don't have this hardware (yet).
So the webserver and others need a local IP address too, we use '::' which works on IPv4 and IPv6
(see mainworker.cpp, webserverp.cpp, tcpserver.cpp)
(That's for listening to a incoming TCP/UDP message)

But after reading the code of this class, it seems there is a kind of discovery going on at startup.
I hope those gateways have a fixed IP address and you can use those addresses as well ?

In that case, you might indeed need a real IP address, and hopefully end-users only have one network adapter, but i know a lot don't ... so... a challenge (think of both WiFi and ETH)

You need a platform independent approach to detect the machine's IP addresses ! (and ignore the localhost and non-network adapter addresses)

Have you tried this?

https://stackoverflow.com/questions/2146191/obtaining-local-ip-address-using-getaddrinfo-c-function

But maybe entering fixed IP addresses is a better way ?

Whiskey24:
Sorry, got a few projects and this one fell to the background for a bit.
You both, @gizmocuz and @xbeaudouin, refer to the same getifaddrs() function, so basically the same solution. 
So I'll give that another go with and I agree it seems a better approach.

gizmocuz:
PR #2778 is merged, does this mean this one can be closed ?

Whiskey24:
This was replaced by #2778. Closing this PR.

PR #2705, opened 2018-09-04
Correct possible misalignment of m_bIsStarted

gordonb3:
- override value of m_bIsStarted in CDomoticzHardware:
  most hardware classes set m_bIsStarted unconditionally to true  where it should  be the return value of StartHardware()
- change Start() and Stop() functions return type to void:
  their return is never used
- add Restart() function
- allow passing thread name to StartHeartbeatThread()


hardware/DomoticzHardware.cpp
```diff
@@ -26,15 +26,32 @@ bool CDomoticzHardwareBase::CustomCommand(const uint64_t /*idx*/, const std::str
...
+}
+
+void CDomoticzHardwareBase::RestartWithDelay(const long seconds)
+{
+	if (StopHardware())
+	{
+		sleep_seconds(seconds);
+		m_bIsStarted = StartHardware();
+	}
 }
```
emontnemery: From which context will this function be called?

hardware/DomoticzHardware.cpp
```diff
@@ -65,11 +87,9 @@ void CDomoticzHardwareBase::Do_Heartbeat_Work()
 {
 	int secCounter = 0;
 	int hbCounter = 0;
-	while (!m_stopHeartbeatrequested)
+	while (!IsStopRequested(1000))
```
gizmocuz: If you use IsStopRequested(), you should use the same value for the sleep_milliseconds, and delete the sleep function

hardware/DomoticzHardware.cpp
```diff
@@ -26,15 +26,32 @@ bool CDomoticzHardwareBase::CustomCommand(const uint64_t /*idx*/, const std::str
 	return false;
 }
 
-bool CDomoticzHardwareBase::Start()
+void CDomoticzHardwareBase::Start()
```
gizmocuz: please keep the boolean as also with the stop.
If it is not used, ok, but at least we also return true/false.

gizmocuz:
Thanks for the PR.
Could you remove the 'RestartWithDelay' and make it optional for the 'Restart' function ? ( Restart(const int seconds=0) )
(int should be fine)

The heartbeat thread in the basehardware class, can you make it not depending on the default StoppableTask class, but add a StoppableMember, like i did in some hardware classes (1Wire.h/cpp, StoppableTask m_TaskSwitches)

gizmocuz:
Thanks!!

PR #2712, opened 2018-09-09
Remove worker from classes derived from ASyncTCP

emontnemery:
- Replace derived class' worker thread with periodic timers
- Expose ASyncTCP's thread to allow derived class to set thread name
- Start ASyncTCP's thread on connect() and stop it on terminate()

### DIFF
--- a/hardware/ASyncTCP.cpp
+++ b/hardware/ASyncTCP.cpp
@@ -20,24 +20,11 @@ ASyncTCP::ASyncTCP()
 	mAllowCallbacks(true),
 	m_reconnect_delay(RECONNECT_TIME)
 {
-	// Reset IO Service
-	mIos.reset();
-
-	//Start IO Service worker thread
-	m_tcpthread = std::make_shared<std::thread>(boost::bind(&boost::asio::io_service::run, &mIos));
 }
 
 ASyncTCP::~ASyncTCP(void)
 {
-	disconnect();
-
-	// tell the IO service to stop
-	mIos.stop();
-	if (m_tcpthread)
-	{
-		m_tcpthread->join();
-		m_tcpthread.reset();
-	}
+	terminate();
 }
 
 void ASyncTCP::SetReconnectDelay(int Delay)
@@ -47,6 +34,16 @@ void ASyncTCP::SetReconnectDelay(int Delay)
 
 void ASyncTCP::connect(const std::string &ip, unsigned short port)
 {
+	//Start IO Service worker thread if not started
+	if (!m_tcpthread)
+	{
+		// Reset IO Service
+		mIos.reset();
+
+		//Start IO Service worker thread
+		m_tcpthread = std::make_shared<std::thread>(boost::bind(&boost::asio::io_service::run, &mIos));
+	}
+
 	// connect socket
 	try
 	{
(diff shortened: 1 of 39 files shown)


gizmocuz:
Thanks for the PR.
Could you split this in two separate PR's ?
I am not a big fan of all the other modifications... 

#include <boost/asio/placeholders.hpp>
#include <boost/asio/steady_timer.hpp>
boost::asio::steady_timer m_heartbeat_timer;
Do_Work(const boost::system::error_code& error)

There is no reason to set the thread name is the class functions correctly ?

I have planned to modify the base class and allow to add a short-name in the constructor of the class, and override the log functions.
This way the short-name is always at the start of the log, and we can also use the short-name for threadnames

emontnemery:
@gizmocuz 
**Regarding _not a big fan of all the other modifications_**
It's not clear to me which changes are OK and which are not.
Do you mean these are OK:
```
#include <boost/asio/placeholders.hpp>
#include <boost/asio/steady_timer.hpp>
boost::asio::steady_timer m_heartbeat_timer;
Do_Work(const boost::system::error_code& error)
```
But this is not:
```
SetThreadName(m_tcpthread->native_handle(), "RFLinkTCP");
```

**Regarding thread name:**
Setting the thread name **greatly** improves readability of Linux stack trace + coredump.
Please note that if thread name is not set, the HW class thread will be named either:
- 'mainworker' if the HW was started at Domoticz startup
- 'webserver' if the HW has been added or configured after Domoticz start

Your idea about short name which is automatically used in logging and so on is very good, but why not merge this first until that's done?

gizmocuz:
@emontnemery , there are to many changes in this patch i do not like, and that is the includes from above and the Do_Work changes
The start/stop in ASyncTCP sounds like a good idea.
The other changes are (for now) not mandetory

emontnemery:
Ok, but a heartbeat thread that does nothing but update a variable is better?
Anyway, closing this.

gizmocuz:
That's for another pull request... it depends.. All hardware classes need to set the heartbeat otherwise it gets restarted. If the hardware should receive data every xx seconds or polls frequently... it should set the heartbeat in that function.

emontnemery:
Yeah, but if all the work happens in the io-worker thread of ASyncTCP (or AsynsSerial) the heartbeat in a separate thread is meaningless.
My intention with this PR was to move heartbeat to the context that is doing the work (and could possibly get stuck).

gizmocuz:
Some classes using ASyncTCP do poll on their own and could update the heartbeat.
Still don't like the idea of these changes:

```
#include <boost/asio/placeholders.hpp>
#include <boost/asio/steady_timer.hpp>
boost::asio::steady_timer m_heartbeat_timer;
Do_Work(const boost::system::error_code& error)
```

So maybe i should first implement the short-name/log, and see if we can simplify things

gizmocuz:
I also renamed all threads now 'm_thread' , and now you change it to another name... lets keep it uniform for feature developers

emontnemery:
> Some classes using ASyncTCP do poll on their own and could update the heartbeat.

Yeah, but I'm also concerned about possible threading related bugs, e.g. races and deadlocks.

> I also renamed all threads now 'm_thread' , and now you change it to another name

Thread was renamed because there are in fact two threads involved for classes extending ASyncTCP (and AsyncSerial):
- m_thread - The thread explicitly started by the HW class itself
- m_tcpthread - The thread started when ASyncTCP object is constructed

Anyway, I will wait for your short-name/log changes and then we can discuss if and / how this PR is relevant and its potential split to multple PRs.

PR #2714, opened 2018-09-11
Update SMTP line endings to comply with RFC

JonathanLondon:
Change line endings from LF to CRLF as per RFC and libcurl SMTP examples
Resolves complaint about protocol violation from Exim MTA
Also fix the attempt to send a dot on a line by itself to teminate the message
Libcurl does this anyway, and now the CRLF issue is fixed, the dot appears in the message body, although outside any MIME attachments, so not visible

### DIFF
--- a/smtpclient/SMTPClient.cpp
+++ b/smtpclient/SMTPClient.cpp
@@ -368,10 +368,9 @@ const std::string SMTPClient::MakeMessage()
 		ret += "--" + std::string(szBoundaryMixed) + "--\r\n";
 	}
 
-	// end the data in the message.
-	// Do not do this - libcurl will escape the dot so it will appear in the body of the sent message
-	// libcurl sends the terminating dot at end of data
-	// ret += "\n.\n";
+	// We do not have to end the message with "<CRLF>.<CRLF>" as libcurl does this for us
+	// ret += "\r\n.\r\n";
+	
 	ret += "\r\n";
 
 	return ret;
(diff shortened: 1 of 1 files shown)


smtpclient/SMTPClient.cpp
```diff
@@ -269,107 +269,109 @@ const std::string SMTPClient::MakeMessage()
...
-		ret += "--" + std::string(szBoundaryMixed) + "--\n";
+		ret += "--" + std::string(szBoundaryMixed) + "--\r\n";
 	}
 
-	// end the data in the message.
-	ret += "\n.\n";
+	// We do not have to end the message with "<CRLF>.<CRLF>" as libcurl does this for us
+	// ret += "\r\n.\r\n";
+	
+	ret += "\r\n";
```
gizmocuz: I am 99,999% sure this last \r\n is no needed, can you verify this

JonathanLondon: libcurl just sends <dot>CRLF so my code is defensive in case anything sends a plain text mail which doesn't terminate with a CRLF.  If everything is sent MIME encoded, and the MIME encoding is performed by SMTPClient.cpp then you are correct that the `ret += "\r\n";` is not needed, and I am very happy for you to remove it.

gizmocuz: If possible please test it

JonathanLondon:
Latest patch just clarifies the 'defensive' additional CRLF

gizmocuz:
Great, thanks for the patch !!

PR #2715, opened 2018-09-16
Fix Tellstick compilation errors (from boost to C++11 conversion)

hoxi:
When migrating some boost usage to C++11 (commit 34816ac623ecd29f246d4364c0931d2221443283 ), some errors were introduced in the Tellstick files. The Tellstick support is only turned on when the tellstick dev-libraries are found, so that is probably why it goes undetected for normal builds.

This commit removes boost usage completely in Tellstick.cpp/.h.


gizmocuz:
Thanks!

PR #2716, opened 2018-09-17
Fixed oldest bug no data updates from some Xiaomi components

JobsIsAlive:
Fixed oldest bug: no data updates from some Xiaomi components before first command to gateway is sended.


hardware/XiaomiGateway.h
```diff
@@ -16,7 +16,7 @@ class XiaomiGateway : public CDomoticzHardwareBase
 
 	int GetGatewayHardwareID(){ return m_HwdID; };
 	std::string GetGatewayIp(){ return m_GatewayIp; };
-	std::string GetGatewaySid(){ return m_GatewaySID; };
+	std::string GetGatewaySid(){ if (m_GatewaySID == "") m_GatewaySID = XiaomiGatewayTokenManager::GetInstance().GetSID(m_GatewayIp); return m_GatewaySID; };
```
gizmocuz, line 19: Maybe place this in the .cpp file as it has more code lines.
But maybe this can be solved somewhere else in the cpp file/code as well

gizmocuz: Is this patch still needed as i see no feedback ?

PR #2723, opened 2018-09-23
Voltage reporting of Neo Coolcam power plugs does not work anymore

Claude2666:
After upgrade from the latest stable 4.9700 to the latest beta releases, the NEO Coolcam plugs don't report voltage anymore. 

In the beta's, a change was introduced changing in the voltage reporting functions ChilldID from an int to a uint8_t, which caused the high byte to be lost. This update changes uint8_t back to an integer uint16_t for ChildID.  I have tested this raspberry pi and win10

### DIFF
--- a/hardware/DomoticzHardware.cpp
+++ b/hardware/DomoticzHardware.cpp
@@ -625,7 +625,7 @@ void CDomoticzHardwareBase::SendRGBWSwitch(const int NodeID, const uint8_t Child
 	sDecodeRXMessage(this, (const unsigned char *)&lcmd, defaultname.c_str(), BatteryLevel);
 }
 
-void CDomoticzHardwareBase::SendVoltageSensor(const int NodeID, const uint8_t ChildID, const int BatteryLevel, const float Volt, const std::string &defaultname)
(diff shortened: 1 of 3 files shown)


gizmocuz:
Thanks for the PR.
I think you got/get a new device... could you enable 'accept new hardware' in the settings and check this
It does not make sense to have the ChildID as an uint16_t, the way it is used:

`gDevice.intval1 = (NodeID << 8) | ChildID`

In fact, if you look at the zwave code, it should be an uint32_t:

```
	else if (pDevice->devType == ZDTYPE_SENSOR_VOLTAGE)
	{
		int sid = (int)(ID1 << 24) | (ID2 << 16) | (ID3 << 8) | ID4;
		SendVoltageSensor(0, sid, BatLevel, pDevice->floatValue, "Voltage");
	}
```

i think it's best to use it at uint32_t ?

Claude2666:
Indeed, It should be uint32_t. How to update? As I am new to github. A new fork and new pull requests?

gizmocuz:
No just change it and commit.. it should update this PR

Claude2666:
Update done to uint32_t

gizmocuz:
Thanks!!

PR #2729, opened 2018-09-26
Add routing for log

stas-demydiuk:
- Use /log route for CounterLog
- Use /log route for CounterSplineLog
- Use /log route for SmartLog
- Extract P1Year report to own component
- Extract P1Gas report to own component
- Use single /report route for all existing report components


gizmocuz:
Thanks for this great patch again !!

gizmocuz:
Is it possible you could have a look at this issue:
https://github.com/domoticz/domoticz/issues/2753#issuecomment-427679360

gizmocuz:
@stas-demydiuk , you also display 'return' values even if the user does not deliver energy back (and the return counters are zero, or less then 1)
I will revert this patch for now, hope you can correct it as this patch was almost great!

stas-demydiuk:
@gizmocuz I wonder if you could share data for real P1 Smart Meter (maybe SQL script to insert data into table or JSON from real request), as I don't have one and could test only with fake device.

gizmocuz:
Sure, can you send me a PM on the forum ? I will send you a database backup
(Also handy for possible other sensors)

PR #2737, opened 2018-09-28
Code improvements

denkovi1:
- Digital input/outputs are changed to SendSwitch instead of SendGeneralSwitch in order to prevent too many logs and events;
- Analog inputs are populated now with raw and scaled value;
- Increased the maximum refresh time constant to one hour;
- Added counters for the digital inputs if they are supported;
- Fixed the temperature C/F recognition for the temperature inputs.


gizmocuz:
Thanks!

PR #2740, opened 2018-09-30
Allow the update of DeviceID from the Python framework

pipiche38:
DeviceID is a field handled by the plugin. 
This change allows the plugin to update DeviceID as other fields

### DIFF
--- a/hardware/plugins/PythonObjects.cpp
+++ b/hardware/plugins/PythonObjects.cpp
@@ -808,6 +808,7 @@ namespace Plugins {
 			PyObject*	pOptionsDict = NULL;
 
 			char*		Name = NULL;
+			char*		DeviceID = NULL;
 			char*		TypeName = NULL;
 			int			iType = self->Type;
 			int			iSubType = self->SubType;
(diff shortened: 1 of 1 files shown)


pipiche38:
I'm not sure that the continuous-integration/travis-ci/pr  failure is due to the changes proposed.

gizmocuz:
Will this not break existing implementations ?

pipiche38:
What do you mean, not breaking existing implementation ?
- I'm just adding one more possibility on update to update the DeviceID

If now the question is about the impact of changing a DeviceID, this is the responsibility of the plugin to manage it as soon as it change it, and the Update in all cases is limited to the Hardware ID. So for instance you cannot update the DeviceID of an entry which doesn't belong to the plugin/hardware

gizmocuz:
Okey thanks, that's clear. But now, why does one need to change the device ID via python ? DeviceID/Unit is a unique combination that should always be fixed. If , for some reason, there is a new DeviceID, domoticz should create a new sensor. (So the user will see this, and can decide to move all data from the old device into the new device (the already existing replace function))

pipiche38:
Valid point 

As part of ZigBee stuff it seems that time to time devices decides to change their Short address to a new one. There is also cases, where if you pair again the device it will come with a new adress. And of course we don't want to loose all existing activities log against that device.

Now, I can accept your point , then the Device doesn't have a Unique address. We except yes, they have a kind of IEEE/Mac@, and here is our purpose to move from the old address mechanism to a new one. ( we are not fully sure on Xiaomi - as we are much more in a reverse engineering mode , rather than working on open specification from the manufacturers - definitively not like Open Zwave -

Basically what we want to implement on the plugin is an upgrade process which will automatically migrate deviceID from ShortAdress to IEEE.

Last , as the DeviceID is relevant to the Plugin, I feel that makes sense to have such capability offer by domoticz plugin framework to update it.

gizmocuz:
When you have a new device id, sure you do not know the previous device id ?
If you do know that, then the current device-id is not generated unique/ok
Exclude/include, a new device id, sure, that sounds logical
But in domoticz you can press 'edit' on a sensor and click on the button 'replace' to replace an old device with a new one

pipiche38:
@gizmocuz to give you the full context, this is for the Zigate plugin, which basically provide similar functionalities than OpenZWave based on an other communication protocol ZigBee.

So we want to make it transparent to the user, as this is part of the operational flow between the ZigBee devices and the Zigate Plugin.

As today the HW Zigate Plugin is based on the Domoticz framework an so we have only the accessible features from the framework.

As you express , if from Domoticz you can press the 'edit' and do a 'replace' why cannot we do it from the plugin itself . This is what this feature will provide, no ?

pipiche38:
@gizmocuz do you need more information ? 
As mentioned in the early command, DeviceID is a field as all others and belongs to the hardware which manage them. So it is the full responsibility of the Hardware to know what it is doing.

gizmocuz:
No i dont, but a you should not be able to change a DeviceID from a script/python.
Any hardware should not change a deviceID for a sensor in normal running mode... that is crazy
When batteries are changed... okey
You say you know what device/deviceid should be changed... that means that the the current deviceid you gave to a sensor is not unique, not OK, and maybe that can be changed so it will always be that ID ?

gizmocuz:
it is the same for zwave, when you exclude and include a node again, you get a new ID ad the node number increases... but that will normally not have to be done

pipiche38:
> You say you know what device/deviceid should be changed... that means that the the current deviceid you gave to a sensor is not unique, not OK, and maybe that can be changed so it will always be that ID ?

That is what we plan to do , but in order to accommodate the existing users we want to provide a smooth migration path, and make it automatically and transparently, without loosing any historical data (like temperature, pressure and other graphs)

> it is the same for zwave, when you exclude and include a node again, you get a new ID ad the node number increases... but that will normally not have to be done

Agree, when the user Exclude and Include as it is a human action. But when it belongs to the plugin behaviour trying to make think transparent, I do consider important that we keep it transparent to the user (domoticz one) and keep all historical data.

Last, I agree with you , this is not a typical and regular thing to do, but in particular cases, like here I'm mainly referring to an upgrade from one way to manage DeviceID to an other way, the function makes sense.

eiaro:
> As part of ZigBee stuff it seems that time to time devices decides to change their Short address to a new one. There is also cases, where if you pair again the device it will come with a new adress. And of course we don't want to loose all existing activities log against that device.

Each device has a IEEE and a NWK address. The IEEE is global unique and the NWK is network unique. The device has no route to change it's NWK as it is assigned by it's parent. This is how it should be, but unfortunately not all bridges report this properly.

If this is reported wrong you could end with a actuator that receives the number of another device and suddenly the event that turn on the lights in the bathroom now turns the lock on your door, or something really bad. 

In any automation I prefer intervention over general assumptions.

pipiche38:
@eiaro fully agree with you, that is exactly what we want to do from a ZigBee perspective without having the user to loose datas. Thansk for contribuing.

PR #2744, opened 2018-10-01
Denkovi Fixed switch / dimmer recognition

denkovi1:
Changed the recognition of switch / dimmer for sending correct command.


denkovi1:
I am not sure where is there issue as the only change we've made is an "if" condition modification.

gizmocuz:
There is no issue

gizmocuz:
Thanks!!

PR #2745, opened 2018-10-01
Fix </a> tags for DashboardController.js

kovainfo:


gizmocuz:
Could you show us a screenshot of before and after ? Seems the name needs to be in the <a> {name} </a> tag, you removed this

kovainfo:
previously (before #2729) the name was out of the tag. after #2729 the name of some devices became in the tag. it's not beautiful. It should be monotonous.

gizmocuz:
Ahhh this clears things up! Thanks for the patch, will merge this!

gizmocuz:
need to revert this as PR 2729 needs to be reverted

PR #2752, opened 2018-10-03
 Stop sensors refresh if Netatmo is locally disconnected

999LV:


gizmocuz:
Thanks for the PR.
Trying to see if this can be simplified:

I do not see why you need to query the domoticz database all the time for the lastupdate field.
This is a field that you could also store as a member inside the class ? (m_LastUpdateTime), you can use it as a string, and store root["time_utc"] in there, then do a simple string compare, if they are the same, the might be no update.... ?
Not sure about update intervals for Netatmo and the polling interval, but you could check if the timestamp is the same and then return.
If you set a data timeout on the hardware (in the hardware setup), you do get a log message from this

999LV:
Thanks for your feedback Gizmocuz. Let me address your comments/suggestions:

1. The polling of the database for LastUpdate is only performed every 10 minutes for sensors and 1 hours for thermostats. I also thought about a local class variable, but felt it was simpler (and more elegant ?) to keep all new code for this functionality locally contained within the ParseDashboard function, given the limited overhead of infrequent database pools.
2. Regarding compare mechanism, eventhough polling intervals between domoticz hardware and Netatmo cloud are of a similar duration, there is not guarantee the polls will be synchronized. This is why I preferred to compare timestamps rather than date/time strings, at the expense of a little conversion work since domoticz does not store Posix style timestamps in its database.
3. I take your point about log entries when devices timeout as per the preferences setting (I did not know that), but the error logging that I included was meant to be more informative to the user, and it is only logged once when an error occurs (not repeated at each polling cycle).

gizmocuz:
Sorry, as mentioned, there is no need to query the database and you can store the root["time_utc"]
compare it on the next poll, if it is the same, you know there was no update

Netatmo updates the "time_utc" value (i hope), so if there is any change in that string (does not matter what changed), there was a change.
An initial string compare is sufficient.
If you want to display how long there was no data, then you have to convert the strings to a timestamp

It does not matter if the log comes from the hardware class, or from Domoticz globally, either way it will tell the user that there was no data received....
If you use the data timeout in the hardware setup, it will restart the class, so for whatever reason, we restart. And regarding Netatmo which is working with tokens, that might not be a bad idea at all

999LV:
I see (though I do not fully understand why) that you want to avoid querying the database. Indeed what you propose (compare last update timestamps for change) works well while domoticz is running, but still you need to query the database at startup to check when the last domoticz devices update took place. Otherwise you would have to wait for the next poll after startup to check for updates, meaning that if even if there is no connectivity issue with the Netatmo devices/modules you wait 10 minutes before the first update.

Anyway, the patch I wrote works for me as it is, so I will carry on with it on my own production systems. Please reject the PR

Thanks

gizmocuz:
Does it matter that when you restarted domoticz, and there where no Netatmo updates, you get the last value again ? How often will this situation arise and what is the problem with it ?
Again a simple string compare is more then sufficient.
I see from your response you do not wish to make changes.
If you do want to contribute in the feature please make a new PR

pipiche38:
Very surprise that a change submitted which can benefits to the community is not to go anywhere !

@999LV I would be interested to your comment 

> but still you need to query the database at startup to check when the last domoticz devices update took place.
I'm not sure if I understood the issue ?

@gizmocuz  I'm by the way surprise that you close the PR, as someone else could be interested to take it over. I'm for instance having Netamo Valves, for which BTW you implemented the code for me, so very much interested !

999LV:
@pipiche38 , I appreciate your interest in this. 

The rationale for the PR was to stop updating domoticz sensors associated with a Netatmo station that lost its local wifi connection and therefore no longer updates the data in the Netatmo API. That happens from time to time at my vacation place and I need to manually reconnect the Netatmo station (when I am present).

Eventhough in that case the Netatmo API still responds to domoticz queries, domoticz devices should no longer be updated and ultimately be marked as timed out (after a time that depends on the domotice settings). This is important to me as my Netatmo station’s temperature readings feed the SmartVirtualThermostat python plugin that I wrote (and that I know has many users)... the plugin excludes timed out temperature devices from its calculations for accuracy or even safety (automatic switch off if no valid temperature reading, to avoid overheating).

Not checking the domoticz devices LastUpdate (as @gizmocuz has suggested) will result in domoticz updating the devices on startup (or rather restart) even if invalid data is reported and these devices would no longer be marked as timedout until the longer of the next poll cycle and the overall sensor timeout setting. This ruins the accuracy of the data fed to the SmartVirtualThermostat plugin (or any other application or script that relies on checking timeouts).

This is unacceptable to me in light of my needs, hence I am NOT willing to drop the querying of the domoticz database for LastUpdate info. A smaller point is that I definitely want to log a specific message when the issue with the local Netatmo connection happens.

That’s why I was not willing to edit my PR and lose the functionalities I need... (I am however open to receiving advise on writing a possibly cleaner code with a similar logic flow than was I proposed... I am not proficient in c++ and certainly willing to learn)

pipiche38:
Fully share your concern and the PR must Re-open.  @gizmocuz didn’t understand the rationale and we should find a way to fix it otherwise all Netatmo users could fall into the same issue 

Envoyé de mon iPhone

gizmocuz:
@pipiche38 , as said before, comparing the JSON root["time_utc"] would be sufficient (as a string)

Next to that, there are other flags that could be check, for example "reachable : true"

Depending on your hardware, the main indoor module is powered, should also be trusted. If you can't trust that one, i would consider searching for a 433Mhz temperature module. If that is not received you do get a timeout notification

I don't even have a field 'time_utc' in my Netatmo responses

So the best thing to do is check the "reachable" flag.

As a (small) reminder, a pull request, is a "request" to get something merged in the repository.

pipiche38:
ok, answer clear

PR #2760, opened 2018-10-06
Add Boost license file

FedericoCeratto:


gizmocuz:
why a boost license ? There is already a license in the root.
Maybe code has been modified and added in this folder as well.

FedericoCeratto:
Various files in that directory are under the Boost license, which should be shipped with them:
`[..snip..]
The copyright notices in the Software and this entire statement, including
the above license grant, this restriction and the following disclaimer,
must be included in all copies of the Software, in whole or in part, and
all derivative works of the Software, unless such copies or derivative
works are solely in the form of machine-executable object code generated by
a source language processor.
`

gizmocuz:
Yes i agree with that, but how about a boost license in the root ? Which file is from boost bdw ? (sure it is using boost)
And i would name the license differently

FedericoCeratto:
In theory the name should stay the same
`webserver/request_parser.cpp:// Distributed under the Boost Software License, Version 1.0. (See accompanying
webserver/request_parser.cpp-// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
`

There are files outside of webserver with the same license indeed:

`
hardware/ASyncSerial.cpp
hardware/pinger/icmp_header.h
hardware/pinger/ipv4_header.h
hardware/plugins/icmp_header.hpp
hardware/plugins/ipv4_header.hpp
hardware/TCPProxy/tcpproxy_server.cpp
webserver/connection.cpp
webserver/connection.hpp
webserver/connection_manager.cpp
webserver/connection_manager.hpp
webserver/header.hpp
webserver/mime_types.cpp
webserver/mime_types.hpp
webserver/reply.cpp
webserver/reply.hpp
webserver/request_handler.cpp
webserver/request_handler.hpp
webserver/request.hpp
webserver/request_parser.cpp
webserver/request_parser.hpp
`

gizmocuz:
true and modified... so... best to have a boost_license.txt file in the root ?

gizmocuz:
or... copy at http://www.boost.org/LICENSE_1_0.txt
let's leave it at the online copy... so no change is needed

FedericoCeratto:
Using an online copy is bad form from a legal standpoint because it's not always reachable and it cannot be proven that the contents are the same over time, hence the PR.

gizmocuz:
Then they should not put that in their headers... and it is an link on their website.
There is a version number there, and if it will be different then the original, not only domoticz developers will complain

PR #2772, opened 2018-10-11
Implemented connectivity to deCONZ gateway

eiaro:
Giving Domoticz a basic connectivity to deCONZ REST API. deCONZ is one of the big efforts into getting a gateway that can communicate with Zigbee hardware from multiple manufacturers. It is similar to Philips Hue, but different enough to not being supported by the current Domoticz implementation.

[deCONZ REST API](https://github.com/dresden-elektronik/deconz-rest-plugin)
[Phoscon Gateway](https://www.dresden-elektronik.de/funktechnik/solutions/wireless-light-control/gateways/phoscon-gateway/?L=1&cHash=c9c902ccdb43164696acccf81b62b2bd)
[RaspBee](https://www.dresden-elektronik.de/funktechnik/solutions/wireless-light-control/raspbee/?L=1)

### DIFF
--- a/hardware/deconz/Deconz.cpp
+++ b/hardware/deconz/Deconz.cpp
@@ -10,6 +10,13 @@
 #include "../../main/mainworker.h"
 #include "../../main/SQLHelper.h"
 
+/* Hardware to support deCONZ REST API 
+ * Zigbee gateway project to support all brands Zigbee devices in one gateway. 
+ * - deCONZ REST API (https://github.com/dresden-elektronik/deconz-rest-plugin)
+ * - Phoscon Gateway(https://www.dresden-elektronik.de/funktechnik/solutions/wireless-light-control/gateways/phoscon-gateway/?L=1&cHash=c9c902ccdb43164696acccf81b62b2bd)
+ * - RaspBee(https://www.dresden-elektronik.de/funktechnik/solutions/wireless-light-control/raspbee/?L=1) 
+ */
+
 Deconz::Deconz(const int id, const std::string& ipAddress, const unsigned short port,
                const std::string& username, const int poll, const int options)
 {
@@ -368,7 +375,8 @@ bool Deconz::GetStates()
 		return false;
 	}
 
-	SaveString2Disk(result, "E:\\deconz.json");
+	// Debugging removed.
+	//SaveString2Disk(result, "E:\\deconz.json");
 
 	Json::Value root;
 	Json::Reader jReader;
--- a/hardware/deconz/Deconz.cpp
+++ b/hardware/deconz/Deconz.cpp
@@ -1,0 +1,824 @@
+#include "stdafx.h"
+#include <string>
+#include "Deconz.h"
+#include "../../main/Logger.h"
+#include "../../main/EventsPythonDevice.h"
+#include "../../main/localtime_r.h"
+#include "../../json/reader.h"
+#include "../../main/WebServer.h"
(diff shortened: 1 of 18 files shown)


www/views/hardware.html
```diff
@@ -1623,35 +1623,56 @@ <h2><span data-i18n="Hardware">Hardware</span>:</h2>
 				</tr>
 			</table>
 		</div>
-		<div id="divphilipshue">
-			<br>
-			<table class="display" id="hardwareparamsphilipshue" border="0" cellpadding="0" cellspacing="20">
+        <div id="divphilipshue">
```
gizmocuz: Could you undo the formatting changes here? Looks like tabs have been replaced by spaces, not only on this line.
Only changes related to the new hardware please

Also please use tabs instead of spaces

hardware/deconz/Deconz.cpp
```diff
@@ -0,0 +1,824 @@
...
+#include "Deconz.h"
+#include "../../main/Logger.h"
+#include "../../main/EventsPythonDevice.h"
+#include "../../main/localtime_r.h"
+#include "../../json/reader.h"
+#include "../../main/WebServer.h"
+#include <platform/Log.h>
+#include "../../main/mainworker.h"
+#include "../../main/SQLHelper.h"
+
```
gizmocuz, line 12: Could you add a link to the github page somewhere at the top here?
Maybe with some more information (zigbee, supported hardware etc)

Otherwise we do not know what this kind of hardware is

gizmocuz (commented):
The ligts/node/sensor classes are small, is it possible to embed them in the hardware class them itself ?

www/views/hardware.html
eiaro: The whole file is alternating tab/space. Even the section you here show is originally both types.
image

To correct this, the best way would be to do all lines *space -> tab*. Would that be OK? However, I corrected my part and copied in the original section above so it should be even now.

SwiftingSpeed:
What happened with this pull request? getting a universal zigbee controller into Domoticz seems like a great addition.

gizmocuz:
From what i remember the author wanted to use some fields in the database that should not be used by plugins.
Did not used the sensor routines from DomoticzHardware.h and was experiencing a 'lag' that was caused by a HTTP call (and thought it was a Domoticz issue)
Maybe work is still being done and a new PR will be made. Another option is to create a new PR based on this one

eiaro:
@SwiftingSpeed Due to waiting almost a week with no feedback, I had to close it as it comes on top of another engagement and I would have no time to support it.

The current state (again, not in this PR) is working IKEA and Philips motion sensors. Switches is also working, but they aren't mapped to Domoticz events as of yet. The lag @gizmocuz mention is solved, but the database fields I'm clueless of as all database code is copy paste from Philips Hue.

If my schedule opens I might make a new PR, but then with sensors and switches.

Smanar:
The lag that you are speaking about is this one that was coming from the missing socket connexion feature ?

eiaro:
Sorry, I don't understand what you are asking.

Smanar:
My fault ^^.
On the forum you are speaking about a missing socket connexion feature  (websocket) to have direcly informations from sensor instead of using polling. It's that the "lag" you are speaking about ?

eiaro:
There is no websocket connection in this PR. 

The lag was when calling the wrong function to update switch state, but that isn't related to this PR.

Issue #2726, opened 2018-09-24
Weather -> Forecast Button Link to 260 Broadway, New York City, NY


Dtrieb:
If I click on weather and the the forecast Button I get the weather from 260 Broadway, New York City, NY not from my city.

gizmocuz:
Is your latitude and longitude correct?

Dtrieb:
Yes it is!
![chrome_2018-09-30_18-52-34](https://user-images.githubusercontent.com/3650284/46260084-1f34b100-c4e2-11e8-86f4-5ddf297570e3.png)
![chrome_2018-09-30_18-54-00](https://user-images.githubusercontent.com/3650284/46260095-370c3500-c4e2-11e8-9709-61e37f1ecd7f.png)

Pooj2:
The url that is requested when pushing on the forecast button is to see in the api output. See <your domotics url>/json.htm?type=devices&used=true&order=Name. Look for forecast_url. Translate the url at [https://www.w3schools.com/jsref/tryit.asp?filename=tryjsref_win_atob](https://www.w3schools.com/jsref/tryit.asp?filename=tryjsref_win_atob). You will see something that starts with `https://darksky.net/#/f/<your settings>,` which always leads to the address above.

I had to make the following change to make it work in  domoticz/www/app/ForecastController.js:
Old: `			htmlcontent += '<iframe class="cIFrameLarge" id="IMain" src="//darksky.net/#/f/' + $scope.config.Latitude + ',' + $scope.config.Longitude + '"></iframe>';`
New: `			htmlcontent += '<iframe class="cIFrameLarge" id="IMain" src="//darksky.net/forecast/' + $scope.config.Latitude + ',' + $scope.config.Longitude + '/ca12/en"></iframe>';`

The /ca12/en above are my preferences, you can get your preference by pushing on the upper right buttons in the darksky.net page.

Than restart domoticz and your browser (my browser cache get cleaned), push the button and see.

Don't mind the output of the api, it will stay the same wrong or dated url.

Dtrieb:
> The url that is requested when pushing on the forecast button is to see in the api output. See /json.htm?type=devices&used=true&order=Name. Look for forecast_url. Translate the url at https://www.w3schools.com/jsref/tryit.asp?filename=tryjsref_win_atob. You will see something that starts with `https://darksky.net/#/f/<your settings>,` which always leads to the address above.
> 
> I had to make the following change to make it work in domoticz/www/app/ForecastController.js:
[...]

PR #2775, opened 2018-10-12
Fix: DarkSky url fixed.

eiaro:
Fix: #2726 
Corrected the URL.

Did not do the proposed changes as it would override changes from Domoticz settings and also force a new unit system on DarkSky for the user.


gizmocuz:
Thanks!!

PR #2776, opened 2018-10-13
Log routing

stas-demydiuk:
Add routing for log
- Use /log route for CounterLog
- Use /log route for CounterSplineLog
- Use /log route for SmartLog
- Extract P1Year report to own component
- Extract P1Gas report to own component
- Use single /report route for all existing report components

PR #2777, opened 2018-10-16
Fix </a> tags

kovainfo:


gizmocuz:
When making changes to the default layout, could you post a before and after screenshot with the changes ?
This does not seem to be a 'fix' but a 'change'

kovainfo:
previously (before #2776) the name was out of the tag. after #2776 the name of some devices became in the tag. it's not beautiful. It should be monotonous.
pic

gizmocuz:
Looking at your patch it is the other way around ? The name is not part of the <a> tag anymore
Also probably caused by that patch is that i am missing names here...

image

PR #2778, opened 2018-10-16
Xiaomigw getifaddrs for local ip

Whiskey24:
Adds a function to obtain local ip addresses using getifaddrs. This is used as fallback when Boost.Asio fails (as happens on FreeBSD).

Only tested on FreeBSD (but code will only be executed when Boost.Asio does not come up with a local ip).

This replaces #2702.

FYI @gizmocuz and @xbeaudouin


gizmocuz:
Thanks!!

gizmocuz:
This will not work or compile under win32

gizmocuz:
Also using exit() in your code is VERY bad... please change this

gizmocuz:
Had to make a commit to let it compile, already changed the exit() call, but please check if you can add win32 support for this, if you google for "getifaddrs win32" you find some answers

Whiskey24:
Sure, no problem. Forgot to consider it needs to compile on win32 and missed the exit() call completely.

I figured out a way to do this for win32, will try to get it in soon.

gizmocuz:
Great thank you ;)!

PR #2779, opened 2018-10-17
DomoticzTCP fix no timeout and hanging on shutdown

jvandenbroek:
I've had a few times the remote connection dropped without notice, keeping the socket alive forever (well at least for about a day..). Using keepalive is a good method to prevent this, set to 600 sec instead of default 7200 sec to speed things up (except for win32, only default possible).

Data timeout set on hardware was ignored when m_bSkipReceiveCheck is set, which is used / should only be used for ignoring heartbeats. I've moved that out of it's scope, changed m_DataTimeout to int so we can check for -1 to ignore certain hardware. I've changed those that were already affected by the bSkipReceiveCheck, the authors can decide themselves if DataTimeout should be ignored or not (just m_DataTimeout = -1 to get default value of 0).

Moved disconnectTCP() to StopHardware(), so the thread continues on stopping (shutdown on socket effectively interrupts the blocking read). This fixes hanging for minutes on shutdown.

### DIFF
--- a/hardware/DomoticzHardware.h
+++ b/hardware/DomoticzHardware.h
@@ -28,7 +28,7 @@ public:
 
 	int m_HwdID = { 0 }; //must be uniquely assigned
 	bool m_bSkipReceiveCheck = { false };
-	unsigned long m_DataTimeout = { 0 };
+	int m_DataTimeout = { 0 };
 	std::string m_Name;
 	_eHardwareTypes HwdType;
 	unsigned char m_SeqNr = { 0 };
--- a/hardware/DomoticzTCP.cpp
+++ b/hardware/DomoticzTCP.cpp
@@ -6,24 +6,22 @@
 #include "../main/mainworker.h"
 #include "../main/WebServerHelper.h"
 #include "../webserver/proxyclient.h"
-
-#ifdef WIN32
-#define SHUT_RDWR SD_BOTH
-#endif
+#include "csocket.h"
 
 #define RETRY_DELAY 30
 
 extern http::server::CWebServerHelper m_webservers;
 
 DomoticzTCP::DomoticzTCP(const int ID, const std::string &IPAddress, const unsigned short usIPPort, const std::string &username, const std::string &password) :
-	m_username(username), m_password(password), m_szIPAddress(IPAddress)
+	m_username(username),
+	m_password(password),
+	m_szIPAddress(IPAddress),
+	m_usIPPort(usIPPort),
+	m_retrycntr(RETRY_DELAY),
+	m_connection(nullptr)
 {
 	m_HwdID = ID;
-	m_socket = INVALID_SOCKET;
-	m_usIPPort = usIPPort;
-	info = NULL;
 	m_bIsStarted = false;
-	m_retrycntr = RETRY_DELAY;
 #ifndef NOCLOUD
 	b_useProxy = IsValidAPIKey(m_szIPAddress);
 	b_ProxyConnected = false;
@@ -32,15 +30,6 @@ DomoticzTCP::DomoticzTCP(const int ID, const std::string &IPAddress, const unsig
 
 DomoticzTCP::~DomoticzTCP(void)
 {
-#if defined WIN32
-	//
-	// Release WinSock
-	//
-#endif
-	if (NULL != info) {
-		freeaddrinfo(info);
-	}
-
 }
 
 #ifndef NOCLOUD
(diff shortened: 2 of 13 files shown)


hardware/DomoticzTCP.cpp
```diff
@@ -127,6 +127,7 @@ bool DomoticzTCP::StopHardware()
 	if (m_thread)
 	{
 		RequestStop();
+		disconnectTCP();
```
gizmocuz: This should actually be done in the worker thread once the request for stop is received at the end of the worker

main/mainworker.cpp
```diff
@@ -12885,59 +12885,57 @@ void MainWorker::HeartbeatCheck()
...
+			//Check Receive Timeout
+			double diff = difftime(now, pHardware->m_LastHeartbeatReceive);
+			if (diff > pHardware->m_DataTimeout)
 			{
-				//Check Receive Timeout
-				double diff = difftime(now, pHardware->m_LastHeartbeatReceive);
-				if (diff > pHardware->m_DataTimeout)
+				std::vector<std::vector<std::string> > result;
+				result = m_sql.safe_query("SELECT Name FROM Hardware WHERE (ID='%d')", pHardware->m_HwdID);
+				if (result.size() == 1)
```
gizmocuz: This should always be 1... can't be different ;) for this reason, please change this to if (!result.empty())

hardware/DomoticzTCP.cpp
jvandenbroek: Problem is that the worker thread is hanging on read() before it receives any stop requests.. By calling disconnectTCP before it will send a shutdown / interrupt to wake it up. That's why it's not working when putting it at the end of the worker thread.

main/mainworker.cpp
jvandenbroek: I didn't change anything in this part of the code ;) Just moved that block out of the if !m_SkipReceiveCheck scope. But will change, no problem ;)

jvandenbroek: Maybe you can change it? Screwing things up here with staging and stuff and can't get a good new commit for this branch somehow without committing all changes I made (which is more than this ;)) Haven't use this system for a while..

hardware/DomoticzTCP.cpp
gizmocuz: Okey, but then we could use a 'select' call for a timeout... for example see

int csocket::canRead( bool* readyToRead, float waitTime )
bool CLimitLess::IsDataAvailable(const SOCKET sock)

I know it's a bit outside the scope of the PR, but it is cleaner to clean things at the end of the worker.

main/mainworker.cpp
gizmocuz: Sure i can do this later ;)
Edit: There seem to be 20+ more lines of code that could be changed

hardware/DomoticzTCP.cpp
```diff
@@ -136,34 +88,32 @@ bool DomoticzTCP::StopHardware()
...
-	#else
-		unsigned long nTimeout = 120*1000;
-		setsockopt(m_socket, SOL_SOCKET, SO_RCVTIMEO, (const char*)&nTimeout, sizeof(DWORD));
-	#endif
-	*/
-	// connect to the server
-	int nRet;
-	nRet = connect(m_socket, info->ai_addr, info->ai_addrlen);
-	if (nRet == SOCKET_ERROR)
+#ifndef WIN32
```
gizmocuz, line 99: Can you make a 'SetKeepAlive' function in csocket ? I think other implementations could benefit from this in the feature

hardware/DomoticzTCP.h
```diff
@@ -47,10 +45,8 @@ class DomoticzTCP : public CRFXBase
 	std::string m_password;
 	int m_retrycntr;
 	std::shared_ptr<std::thread> m_thread;
-	sockaddr_in6 m_addr;
-	struct addrinfo *info;
-	int m_socket;
-	unsigned char mBuffer[512];
+	csocket *m_connection;
```
gizmocuz, line 48: I'm not a big fan of new/delete usages.... It saves a header inclusion ...
Another approach would be to inherit csocket

class DomoticzTCP : public CRFXBase, csocket

I am not a bit fan of the class name 'csocket' and member functions at the moment... 
Not even sure it is working for IPV6 (as this class (DomoticzTCP) was working with

I am a fan of ASyncTCP, this is a much better class. Is it possible to use this instead ?

hardware/DomoticzHardware.h
```diff
@@ -28,7 +28,7 @@ class CDomoticzHardwareBase : public StoppableTask
 
 	int m_HwdID = { 0 }; //must be uniquely assigned
 	bool m_bSkipReceiveCheck = { false };
-	unsigned long m_DataTimeout = { 0 };
+	int m_DataTimeout = { 0 };
```
gizmocuz, line 31: I would like to propose a declaration here that classes (and mainworker) can use instead of -1 ... because that does not say a lot.

#define TIMEOUT_IGNORE -1

so classes can use 

`m_DataTimeout = TIMEOUT_IGNORE;`

..

`if (pHardware->m_DataTimeout != TIMEOUT_IGNORE)`

I still don't get it why you need -1 ... if it is set to 0 (by default), there is no check on timeout
That should be sufficient ?

Hardware that does not receive anything, or not much (like only switch commands once in a while) should set 'm_bSkipReceiveCheck', so the datatimeout is ignored.
Actually this whole boolean could also be removed completely in the code, as users should not be allowed to specify a datatimeout, or should not do so
(Maybe we should call it 'receive timeout' in the user interface to clear things up)

gizmocuz (commented):
Thank you for the changes, i have some comments/suggestions/questions

jvandenbroek:
Created a new PR https://github.com/domoticz/domoticz/pull/2796 based on ASyncTCP

PR #2782, opened 2018-10-19
Various NetBSD updates

tih:
Here's a couple of updates needed to build cleanly under NetBSD:

- The change from "-std=c++11" to "-std-=gnu++11" is because GNU C++, as configured under NetBSD, needs this to handle malloc correctly.

- The change in StoppableTask.h works around a problem that occurs on some recent combinations of NetBSD versions and platforms, where the wait_for() function busy waits, and does so with a huge amount of overhead, thus both wasting CPU and messing up the timing.

- I've taken the liberty of submitting the #include of string.h in localtime_r.cpp without any #ifdef wrapping, as I feel it really ought to be there for all ports, to supply the correct declaration for memcpy().

- Finally, NetBSD does have a 64 bit off_t, so the zip code should be built accordingly.


gizmocuz:
Thanks for the PR and patches !

I do not understand why you need to use gnu++11 as this is used for extensions which we do not use.
This should not be needed, and i could also not find a post via google explaining this

About the StopppableTask.h patch, could you point me to a post explaining this ? I can't believe this function behaves differently for BSD systems, that would be really bad. Nobody wants to use ifdefelse code

tih:
Thanks for your comments!  The gnu++11 thing was something someone suggested because of this warning during linking:

CMakeFiles/domoticz.dir/hardware/serial/serial.cpp.o: In function `serial::Serial::readline(std::string&, unsigned long, std::string)':
serial.cpp:(.text+0x4c9): warning: Warning: reference to the libc supplied alloca(3); this most likely will not work. Please use the compiler provided version of alloca(3), by supplying the appropriate compiler flags (e.g. not -std=c89).

However, it turns out that a much simpler fix is to activate the relevant #define, already present in hardware/serial/serial.cpp, which I have now done (and reverted the CMakeLists.txt change).

tih:
The problem with the future object wait_for() call is, I believe, in certain versions of libstdc++.  On the NetBSD hosts where I observe it, I have "gcc (nb2 20180327) 6.4.0" -- in other words, a tool chain imported back in March.  With a newer system, and "gcc (nb3 20180905) 6.4.0", the problem is not present.  A binary compiled on the former, and run on the latter, does not exhibit the problem.

It doesn't take all of Domoticz to demonstrate it: the first complete example program on https://thispointer.com/c11-how-to-stop-or-terminate-a-thread/ will do it.

Since I don't know exactly which versions of NetBSD have and do not have the buggy libstdc++, I chose to suggest the patch as I did to cover all the bases; the changed code is no worse than the original, and works in both environments.

gizmocuz:
Thanks for the change! Great find !

About the StopppableTask.h patch, i find it hard to believe that on some systems it is not working correctly. (that would not be great!)
But if i read your comment , it looks like there was a bug in BSD and they solved it with an update ?
In that case, please revert this patch. Users should update their system, who knows what else was wrong with their c++11 implementation

tih:
Good point - I don't know exactly in which versions of NetBSD the wait_for() problem might be present, but it has been fixed, so the proper fix is, as you say, to upgrade the system, not the application.  Reverted.

gizmocuz:
Thanks!!

krytarowski:
We shall use gnu++11 for alloca, as this is an extension shipped with a compiler. `#define alloca` is a hack that might work, but beware from using alloca() symbol, as it will resolve to libc's internal fallback version of alloca() that tries to implement it without compiler support, but it's usually breaking. In general it's considered to be difficult (impossible?) to implement alloca() functionally out of a compiler.

PR #2783, opened 2018-10-19
Make DomoticzTCP stoppable and restartable

tih:
The recent changes to introduce stoppable tasks and watchdogs has turned the DomoticzTCP mechanism for receiving measurements from a remote Domoticz into a problem, because of its recv() call with no timeout.

- If the data stream stops, the watchdog will detect this, and tell it to stop, so it can be restarted - but as it's sitting in a recv() that doesn't return, it won't react.

- During shutdown, it'll likewise be told to stop, but again won't react, so Domoticz as a whole fails to terminate.

My change reinstates the setting of a socket timeout option, which was there, but commented out, but a) reduces the timeout to ten seconds, to make the code acceptably responsive during shutdown, and b) reorganizes things a little bit, to handle the return values from recv() correctly.

### DIFF
--- a/hardware/DomoticzTCP.cpp
+++ b/hardware/DomoticzTCP.cpp
@@ -142,17 +142,17 @@ bool DomoticzTCP::ConnectInternal()
 		_log.Log(LOG_ERROR, "Domoticz: TCP could not create a TCP/IP socket!");
 		return false;
 	}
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thanks for the patch!!

PR #2784, opened 2018-10-19
Add a quirk for the Popp Stop water cut-off valve

tih:
I recently supplied a configuration file for the Popp Stop water cut-off valve to OpenZwave, and the Domoticz project has imported this (Config/popp/009501.xml). The Popp unit claims to have a multilevel setting, which is untrue (I've even opened it up to verify that the hardware can't support one; it has no position sensing, just microswitches at the end stops). My XML code, as submitted to OpenZwave, includes the rejection of this part of the configuration (using \<CommandClass id="38" action="remove" />), but while this keeps the offending bits out of the generated configuration, Domoticz still tries to send multilevel switch commands to the device.

This change adds a quirk for the device, patterned on existing ones for other devices.

### DIFF
--- a/hardware/OpenZWave.cpp
+++ b/hardware/OpenZWave.cpp
@@ -1238,6 +1238,14 @@ bool COpenZWave::SwitchLight(const int nodeID, const int instanceID, const int c
 				bHandleAsBinary = true;
 			}
 		}
+		if (pDevice->Manufacturer_id == 0x0154)
+		{
+			if ((pDevice->Product_id == 0x0003) && (pDevice->Product_type == 0x0005))
+			{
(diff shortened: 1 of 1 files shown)


hardware/OpenZWave.cpp
```diff
@@ -1238,6 +1238,14 @@ bool COpenZWave::SwitchLight(const int nodeID, const int instanceID, const int c
 				bHandleAsBinary = true;
 			}
 		}
+		if (pDevice->Manufacturer_id == 0x0154)
```
gizmocuz: please change this to 'else if'

gizmocuz:
Thank you for the patch! Made a small change

PR #2791, opened 2018-10-23
Find python3.x lib on MacOS

moroen:
Improved search for libpython3.x.dylib on MacOS. 

When trying to dynamically load the python library, domoticz will now find libpython3.x.dylib installed either using the official download from www.python.org or python3 installed by using homebrew.


gizmocuz:
Thanks!!

PR #2796, opened 2018-10-24
DomoticzTCP rewrote to use ASyncTCP

jvandenbroek:
It seems like someone else already changed this and got merged yesterday, while I was also working on it.. Just wanted to test it for a day to see if all is working fine and it seems so :) Had to patch ASyncTCP to resolve IPv6 addresses, this didn't work before.. 

Anyway, this will stop immediately on shutdown instead of a 10 sec delay, and as a bonus I've included a authentication check ;) Not sure why the current code does keep the connection alive when no user is specified at all? Maybe that has it's reason, so decided no to alter that (but shouldn't we always authenticate anyway?)


gizmocuz:
Thanks!

Issue #2583, opened 2018-07-21
Build broken with GCC 6.4.0 on Alpine


rhaamo:
When updating my Domoticz image (alpine:latest) it doesn't build anymore and I get an error somewhat similar to #2576 but not identical.

The build is done using alpine docker image (latest) and it use the current GCC of alpine which looks like 6.4.0.

build log:
```
-- The C compiler identification is GNU 6.4.0                                                                                                                     
-- The CXX compiler identification is GNU 6.4.0
[...]
[ 23%] Building CXX object CMakeFiles/domoticz.dir/main/EventsPythonDevice.cpp.o                                                                                  
[ 24%] Building CXX object CMakeFiles/domoticz.dir/main/Helper.cpp.o
[ 24%] Building CXX object CMakeFiles/domoticz.dir/main/IFTTT.cpp.o
In file included from /src/domoticz/main/IFTTT.cpp:3:0:
/src/domoticz/main/SQLHelper.h:74:10: error: field '_DelayTimeBegin' has incomplete type 'timeval'                                                                
  timeval _DelayTimeBegin;                                                                                                                                        
          ^~~~~~~~~~~~~~~                                                                                                                                         
In file included from /src/domoticz/main/SQLHelper.h:6:0,                                                                                                         
                 from /src/domoticz/main/IFTTT.cpp:3:                                                                                                             
/src/domoticz/main/Helper.h:71:21: note: forward declaration of 'struct timeval'                                                                                  
 int getclock(struct timeval *tv);                                                                                                                                
                     ^~~~~~~                                                                                                                                      
make[2]: *** [CMakeFiles/domoticz.dir/build.make:279: CMakeFiles/domoticz.dir/main/IFTTT.cpp.o] Error 1                                                           
make[1]: *** [CMakeFiles/Makefile2:105: CMakeFiles/domoticz.dir/all] Error 2                                                                                      
make: *** [Makefile:152: all] Error 2                                                                                                                             
```

rhaamo:
#2592 fixes the timeval, however I'm hitting another issue later:

```
/src/domoticz/hardware/csocket.cpp: In member function 'int csocket::connect(const char*, unsigned int)':                                                         
/src/domoticz/hardware/csocket.cpp:129:39: error: 'u_short' was not declared in this scope                                                                        
  m_remoteSocketAddr.sin_port = htons((u_short)m_remotePort);                                                                                                     
                                       ^~~~~~~                                        
```                                                                            
[...]

PR #2802, opened 2018-10-26
Alpine/MUSL: Fix pthread_getname_np only available on GLIBC

rhaamo:
Used ```__GLIBC__``` to conditionally call ```pthread_getname_np``` and same for the Log right after.

For #2583


gizmocuz:
Thanks!!

PR #2803, opened 2018-10-26
Alpine/MUSL: Replace u_short by uint16_t

rhaamo:
u_short doesn't exist under Alpine/MUSL, changed to uint16_t as discussed into #2583


gizmocuz:
Thanks!!

PR #2592, opened 2018-07-26
Fix missing sys/time.h include

rhaamo:
Partly fixes #2583

PR #2804, opened 2018-10-26
Alpine/MUSL: Fix timeval by better importing sys/time.h

rhaamo:
Musl on Alpine doesn't auto include ```sys/time.h```, including it as discussed in #2583

### DIFF
--- a/hardware/HardwareMonitor.cpp
+++ b/hardware/HardwareMonitor.cpp
@@ -26,7 +26,6 @@
 	#include <fstream>
 	#include <string>
 	#include <limits>
-	#include <sys/time.h>
 	#include <unistd.h>
 
 	struct _tDUsageStruct
--- a/hardware/csocket.cpp
+++ b/hardware/csocket.cpp
@@ -13,7 +13,6 @@
 #include <arpa/inet.h>
 #include <netdb.h>
 #include <errno.h>
-#include <sys/time.h>
 #include <sys/errno.h>
 #include <unistd.h>
 #include <fcntl.h>
--- a/hardware/serial/impl/unix.cpp
+++ b/hardware/serial/impl/unix.cpp
@@ -24,7 +24,6 @@
 #endif
 
 #include <sys/select.h>
-#include <sys/time.h>
 #include <time.h>
 #ifdef __MACH__
 #include <AvailabilityMacros.h>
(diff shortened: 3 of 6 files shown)


gizmocuz:
Please do not include it in stdafx.h, this header is include in every cpp file
Best to add the include in the cpp files directly when needed

rhaamo:
I've done what you told me in https://github.com/domoticz/domoticz/issues/2583#issuecomment-420524427

The first issue is:
```
[ 23%] Building CXX object CMakeFiles/domoticz.dir/main/IFTTT.cpp.o
In file included from /src/domoticz/main/IFTTT.cpp:3:0:
/src/domoticz/main/SQLHelper.h:75:10: error: field '_DelayTimeBegin' has incomplete type 'timeval'                                                                      
  timeval _DelayTimeBegin;                                                                                                                                              
          ^~~~~~~~~~~~~~~                                                                                                                                               
In file included from /src/domoticz/main/SQLHelper.h:6:0,                                                                                                               
                 from /src/domoticz/main/IFTTT.cpp:3:                                                                                                                   
/src/domoticz/main/Helper.h:71:21: note: forward declaration of 'struct timeval'                                                                                        
 int getclock(struct timeval *tv);                                                                                                                                      
                     ^~~~~~~                                                                                                                                            
make[2]: *** [CMakeFiles/domoticz.dir/build.make:193: CMakeFiles/domoticz.dir/main/IFTTT.cpp.o] Error 1                                                                 
make[1]: *** [CMakeFiles/Makefile2:105: CMakeFiles/domoticz.dir/all] Error 2                                                                                            
make: *** [Makefile:152: all] Error 2                                                                                                                                   
```

Putting it in ```main/Helper.h``` will also act as a "global fix" as already said in the previous PR about the same issue: https://github.com/domoticz/domoticz/pull/2592#issuecomment-417607909 and it was rejected.

The only thing I can do is either put ```#include <sys/time.h>``` on top of impacted .cpp files, or right before the ```SQLHelper.h``` include in thoses sames cpp files.

What do you prefers ?

gizmocuz:
Then i see no other way !.... thanks for the patch !!

rhaamo:
Thanks for merging !
I was rebuilding after adding one-by-one an include in each cpp files and yeah there is only two ways, global or in like (for now) ~5 cpp files.
I don't think sys/time will do any harm anyway, but still it would have been better if there was a better way.

stas-demydiuk:
Awesome, it is possible to use alpine again! @rhaamo thank you many times for the patch!

PR #2809, opened 2018-10-27
Fix energy total cost calculation

stas-demydiuk:


gizmocuz:
Already solved, thanks for this patch!

Issue #2368, opened 2018-05-02
Hex code,  near hue and yeelight bulb


Dylantje:
Dear developers....
Until a few months ago there was a option to set a hex code in the hue or yeelight.
Now this gone..
Is there perhaps a option to get this hex code option back?
This was a very great option to set all bulbs into the same color.
With the colorpicker is this almost impossible.

emontnemery:
Where would you like to see this, is it in timer and scene editors?

gizmocuz:
This was in the JSON api

Dylantje:
Dear...
In this area was in older versions a place where you set the color with hex...
https://drive.google.com/open?id=1dHFUUglCOUDKZHVJ4yuHLeKTyUXhPEWi

gizmocuz:
Ahhh okey..yes that is indeed gone now....

Dylantje:
That was a great feature ...
All the lights the same color..
And the same level...

emontnemery:
@dylantje Where would you like to see this, is it in timer and scene editors?

Edit: Also, which format do you suggest? There is now both RGB picker and color temperature picker..

Dylantje:
Dear,,, When you can implement this great!!
The type/format should be in my eyes..:
http://www.color-hex.com/  [ in the top there is a color picker that give the color a Hex code ]

Where u can place this.. like it on the place where i let you see....
Perhaps also in the timer section.

If possible...
Perhaps als in blockley..:-) [ set a bulb in blockley, to a hex code. ]

And when you are my superhero....
If possible make a fade option in the timer and blockley section..:-)


And when you will be a super super hero for the WAF.... and the kids..
A disco option ,... :-)

Only i think i'am  dreaming out loud ....
Thanks..!!
 xx

Dylantje:
mmm Dream this night about it :-)
Perhaps also in the groups/scene section..

x
x

gizmocuz:
@emontnemery , i think he is talking about the colorwheel popup. In the new popup you can not enter a hex value, so you have to click on a point to set the color.

gizmocuz:
According the colorwheel pickers github page, it should be possible to add a textbox for the RGB value, do you think this is possible to implement ?
Maybe also at the bottom of the sliders a small numeric textbox for the slide values

emontnemery:
[...]

Issue #2307, opened 2018-04-10
MiLight and timers+brightness


Tontze:
Not a bug, but improvement for setting timers and brightness level.

Since there is no more field where to write brightness level percentage, you need to try many times with slider to find percentage you want to use.  Please bring the field back :)

Tontze:
For clarification, you need to slide slider to some point, then accept it, and then you see what % position  slider was. If its not the one you want, you need to edit it and move slider a bit to find correct % ... 

Not very handy do you think ?

emontnemery:
No, not very handy at all and I agree UI should be same for both normal dimmers and color dimmers.
@stas-demydiuk, I think I broke this when doing the RGBW stuff. Should I fix it, or are you working on the edit/timer/scenes stuff?

Also, I don't think it's handy to have a drop-down where you have to scroll to choose 0..100, but not sure how to make the UI both easy/quick to use and allow exact setting.

stas-demydiuk:
@emontnemery currently I'm not working on the edit/timer/scenes stuff, so feel free to update it if required

Tontze:
Perhaps you could use slider, but show percentage when sliding ?

Tontze:
Will this be improved ? I was about to do outdoor lightning that lights up bit by bit during sunset.. But its impossible with timers now ..

emontnemery:
Yeah, I'll try to get it fixed this weekend. Sorry about the delay!
It would be great if you could help testing, are you running Domoticz on Raspberry Pi?

Please have a look in #2810, and let me know if I understood your request correctly.

Tontze:
Yup i can help, if instructed carefully :D

waltervl:
Fixed, can be closed

PR #2810, opened 2018-10-27
Small fixes for rgbww related UI

emontnemery:
Fix several small issues:

- Show hex rgb input in light edit, timer editor and scene editor for RGB lights, when color mode is RGB, as requested in #2368:
image

- Show dim level combo box in timer editor for RGB lights, as requested in #2307:
image

- Fix scene editor bug when default device was RGB type

- Fix timer editor bug causing incorrect default level to be shown for RGB lights

### DIFF
--- a/www/app/ScenesController.js
+++ b/www/app/ScenesController.js
@@ -566,23 +566,22 @@ define(['app'], function (app) {
 			$("#scenecontent #LevelDiv").hide();
 			if (isLED(SubType)) {
 				$("#ScenesLedColor").show();
-			} else {
-				if (bShowLevel == true) {
-					var levelDiv$ = $("#scenecontent #LevelDiv");
-					levelDiv$.find("option").show().end().show();
+			}
+			if (bShowLevel == true && !isLED(SubType)) { // TODO: Show level combo box also for LED
+				var levelDiv$ = $("#scenecontent #LevelDiv");
+				levelDiv$.find("option").show().end().show();
 
-					var dimmerValues = [];
+				var dimmerValues = [];
 
-					$.each(dimmerLevels.split(','), function (i, level) {
-						dimmerValues[i] = level;
-					});
+				$.each(dimmerLevels.split(','), function (i, level) {
+					dimmerValues[i] = level;
+				});
 
-					levelDiv$.find("option").remove();
-					for (var levelCounter = 0; levelCounter < dimmerValues.length; levelCounter++) {
-						var option = $('<option />');
-						option.attr('value', dimmerValues[levelCounter]).text(dimmerValues[levelCounter] + "%");
-						$("#scenecontent #combolevel").append(option);
-					}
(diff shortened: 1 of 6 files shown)


www/app/ScenesController.js
```diff
@@ -566,23 +566,22 @@ define(['app'], function (app) {
 			$("#scenecontent #LevelDiv").hide();
 			if (isLED(SubType)) {
 				$("#ScenesLedColor").show();
-			} else {
-				if (bShowLevel == true) {
-					var levelDiv$ = $("#scenecontent #LevelDiv");
-					levelDiv$.find("option").show().end().show();
+			}
+			if (bShowLevel == true && !isLED(SubType)) { // TODO: Show level combo box also for LED
```
emontnemery, line 570: @stas-demydiuk What would be a good way to connect level combo-box with rgbww picker here? In timer editor this was already done by you..

www/views/timers.html
```diff
@@ -177,7 +177,7 @@ <h2>{{ ::'Name'|translate }}: {{::vm.itemName}}</h2>
                 </select>
             </td>
         </tr>
-        <tr id="TimersLedColor" ng-show="::vm.isColorSettingsAvailable" class="ng-hide">
+        <tr id="TimersLedColor" ng-show="::vm.isColorSettingsAvailable" class="ng-show">
```
emontnemery, line 180: @stas-demydiuk I changed this since the rgbww picker does not render correctly when hidden, is that OK?

gizmocuz:
@emontnemery , is it possible you could change the style of the RGB editbox to capitals ?

www/views/timers.html
gizmocuz: @emontnemery , no this is not the way. You have to either make it visible by letting the routine "vm.isColorSettingsAvailable" return true, or remove the whole ng-show attribute when it should 'allways' be visible

emontnemery: @gizmocuz  The problem is that vm.isColorSettingsAvailable initially returns true **after** the color picker has already been created (initLEDLightSettings function in app/timers/components.js:136). Thus, the colorpicker is created with zero size which causes display bugs.
By setting the element to default visible instead of default invisible there are no display bugs.

emontnemery:
>  is it possible you could change the style of the RGB editbox to capitals

Fixed now!

www/app/ScenesController.js
stas-demydiuk: As there is almost jQuery stuff, I see the only way just to add listeners for each change and manually do the related updates

www/views/timers.html
emontnemery: @gizmocuz There is a proper fix in #2815, can this be merged first? Otherwise I can revert this change.

gizmocuz: Sure!

gizmocuz:
Thanks!!!

PR #2812, opened 2018-10-28
Events-editor: minor improvements

stas-demydiuk:
-- Show disabled label for disabled scripts
-- Highlight active script in list
-- Highlight On/Off buttons with green/red
-- Show notification after script successful removal


gizmocuz:
Thanks!!

PR #2817, opened 2018-10-29
inclreased Nest_Poll_Interval due to Nest rate limits

molekuul:
For the connection to the Nest Service, there is a rate limit, see 
[Nest](https://developers.nest.com/guides/api/data-rate-limits)

 To avoid errors, we recommend you limit requests to one call per minute, maximum.

kind regards,

Joris Weijters


gizmocuz:
Thanks!!

PR #2820, opened 2018-10-29
Update html5.appcache

syrhus:
Domoticz is not an Offline application and doesn't work Offline, just have to store the offline page and image ONLY


gizmocuz:
@syrhus , it does work. Maybe first discuss this before you assume this and make a PR ?
Sure it needs to be online to communicate, but all pages/images/scripts/... are loaded offline, this speeds up everything!

PR #2821, opened 2018-10-29
RFLink: Add missing trailing semicolon

emontnemery:
Send trailing semicolon as documented in RFLink protocol reference: http://www.rflink.nl/blog2/protref
Same as #2816, but fix also for various RGB lights


gizmocuz:
Thanks!!

PR #2825, opened 2018-10-31
fixed working of httpclient redirect

molekuul:
After compilation of the domoticz, the redirection of URL by the HTTPClient didn't work anymore.
It seems:
"Starting in 7.58.0, libcurl will specifically prevent "Authorization:" headers from being sent to other hosts than the first used one, unless specifically permitted with the CURLOPT_UNRESTRICTED_AUTH option."
see https://curl.haxx.se/libcurl/c/CURLOPT_HTTPHEADER.html
 
I added parameter to make libcurl continue to send authentication (user+password) credentials when following locations, even when hostname changed.


gizmocuz:
Thanks for the PR! What't the URL that is being called ? Enabling this option could be dangerous

molekuul:
The NestOAuthAPI connects to https://developer-api.nest.com  This gets forwarded to https://firebase-apiserver44-tah01-iad01.dapi.production.nest.com:9553

In older versions of the libcurl, prior 7.58.0, the authentication headers were redirected, however starting from this version authentication headers will only be forwarded when The parameter: " CURLOPT_UNRESTRICTED_AUTH",  is set.

Kind regards,
Joris Weijters

gizmocuz:
Thanks!!

PR #2828, opened 2018-10-31
Fix inability to disable the last active dzvents script (fixes #2827)

### LINKED ISSUE #2827: Deactivating a dzvents script with no other dzvents or lua scripts active fails to work
Playing with the (excellent) new events GUI, I've noticed that there's a bug that affects deactivation of dzvents scripts. If I understand correctly, the activation of an internally edited dzvents script is done by writing it to the "generated_scripts" directory and resetting the lua environment - and deactivation is similary done by removing the script from said directory. I guess lua reads all the scripts it finds in "scripts" (external) and "generated_scripts" (internal), and these then make up the currently active dzvents set.

This all works fine, except for deactivating an internal dzvents script when it is the only internal lua or dzvents script to be active. In this case, it is set as inactive in the database, but *not* removed from the generated_scripts directory, and it will thus stay active. (If you then delete the script from the GUI, it'll be removed from the database, but still remain in the file system, and will continue to run, even after a restart of Domoticz.) With a script in the "deactivated but erroneously active" state, activating another internal script, either lua or dzvents, will cause the error to be corrected; the script is then removed from the generated_scripts directory, and stops running.

If I were to hazard a guess, I'd look for a place where the code says approximately "if at least one lua or dzvents script is currently active, let's make sure everything is correctly set up, and reset the lua environment" - failing to catch the situation where no script is active, but the reason is that we've just deactivated the very last one, which happens to be a dzvents script.

tih:
This is a fix for issue #2827 - see the issue text for problem description and rationale for change.

### DIFF
--- a/main/EventSystem.cpp
+++ b/main/EventSystem.cpp
@@ -240,24 +240,26 @@ void CEventSystem::LoadEvents()
 			m_events.push_back(eitem);
 		}
 	}
+
+	std::vector<std::string> FileEntries;
+	std::string filename;
+
+	// Remove dzVents DB files from disk
+	DirectoryListing(FileEntries, dzv_Dir, false, true);
+	std::vector<std::string>::const_iterator itt;
+	for (itt = FileEntries.begin(); itt != FileEntries.end(); ++itt)
+	{
+		filename = dzv_Dir + *itt;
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thanks !!!

PR #2830, opened 2018-11-01
Fix bugs related to dimmer and on/off switch

### LINKED ISSUE #2780: Hue dimmer recognition
Hi!

Since the beta updates from early October I have some issues with my Hue lights. Some lights, even bundles of lights, are recognized as dimmers. When I change them to an on/off type switch, they work fine as on/off going to the last state they were in. This is how I used them the last years. 

Since the beta updates from early October the light will switch back to a dimmer type switch after a while, usually after one of these lights are updated. I see that there is a commit concerning switch/dimmer recognition, but I'm not sure if this is what results in this bug. If I select to use a Hue light as an on/off type switch, it would be great if it just stays an on/off light switch and doesn't revert back to a dimmer type switch. 

I tried to find the code where it goes wrong, but I'm unfortunately not able to find what causes it. Thanks in advance!

emontnemery:
Closes #2780 

Note: Not ready for merge yet, needs testing

### DIFF
--- a/hardware/PhilipsHue/PhilipsHue.cpp
+++ b/hardware/PhilipsHue/PhilipsHue.cpp
@@ -475,6 +475,7 @@ std::string CPhilipsHue::RegisterUser(const std::string &IPAddress, const unsign
 
 void CPhilipsHue::InsertUpdateSwitch(const int NodeID, const _eHueLightType LType, const _tHueLightState tstate, const std::string &Name, const std::string &Options, const std::string &modelid, const bool AddMissingDevice)
 {
+	_log.Log(LOG_STATUS, "Philips Hue: InsertUpdateSwitch called for device '%s' with type %u", Name.c_str(), LType);
 	if (LType == HLTYPE_RGB_W || LType == HLTYPE_CW_WW || LType == HLTYPE_RGB_CW_WW)
 	{
 		char szID[10];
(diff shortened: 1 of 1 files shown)


bbogey:
Yes it works now for both individual switch and for the hue group. Also log shows the status OK.

image

hardware/PhilipsHue/PhilipsHue.cpp
```diff
@@ -475,6 +475,7 @@ std::string CPhilipsHue::RegisterUser(const std::string &IPAddress, const unsign
 
 void CPhilipsHue::InsertUpdateSwitch(const int NodeID, const _eHueLightType LType, const _tHueLightState tstate, const std::string &Name, const std::string &Options, const std::string &modelid, const bool AddMissingDevice)
 {
+	_log.Log(LOG_STATUS, "Philips Hue: InsertUpdateSwitch called for device '%s' with type %u", Name.c_str(), LType);
```
gizmocuz: Could you remove this debug line, or use the DEBUG_HARDWARE call for this ?

gizmocuz:
Thanks for this PR. There are lots of new debug/log lines added. Is it possible to remove them all, or use the .Debug(HARDWARE,...) option ?

emontnemery:
@gizmocuz Sure, the logs were just added for testing and I will remove them.

emontnemery:
@gizmocuz The extra prints are removed now, this is ready for merge.

gizmocuz:
Thanks!!!

PR #2836, opened 2018-11-02
Ensure / ending to a couple more directory paths

### LINKED ISSUE #2826: The -userdata parameter is required to end in a / - but this is not enforced
Domoticz normally determines the path to its various component directories working from the path to the invoked executable itself, by using that path up to and including the last '/' as a starting point. Several paths are derived from this, by appending directory names to it. When this is done, it is assumed that the original path (typically in the szUserDataFolder variable) ends in a '/'. Picking an example at random:

std::string sbackup_DirD = szUserDataFolder + "backups/daily/";

If the user chooses to give the user data folder as a command line parameter, though, he or she might understandably specify e.g. "-userdata /opt/domoticz", without the trailing slash. Domoticz will accept this parameter without complaint, and end up making mistakes like (using the example above) trying to put daily backups in "/opt/domoticzbackups/daily/".  Similarly, the scripting subsystems will be looking for "/opt/domoticzscripts/", which again does not exist.

My suggestion would be to have the -userdata switch (and probably the other switches that are used to supply directory paths) check for a trailing slash, and silently append one if it's missing.

tih:
Fixes #2826, using Rob Peters' recent mechanism for sanitising the -userdata parameter.

### DIFF
--- a/main/domoticz.cpp
+++ b/main/domoticz.cpp
@@ -658,9 +658,11 @@ bool ParseConfigFile(const std::string &szConfigFile)
 		}
 		else if (szFlag == "app_path") {
 			szStartupFolder = sLine;
+			FixFolderEnding(szStartupFolder);
 		}
 		else if (szFlag == "userdata_path") {
 			szUserDataFolder = sLine;
+			FixFolderEnding(szUserDataFolder);
 		}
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thanks!!

PR #2846, opened 2018-11-07
Proxyclient ASsyncTCP + secure connetions in ASyncTCP

rkluwen:
This pull request implements the CProxyClient class as a base class of ASyncTCP.
At the same time the ASyncTCP class has been extended to be able to connect via ssl sockets.

This patch hasn't been tested for a long period of time, so a message to all testers is to use this branch for a while.

### DIFF
--- a/hardware/ASyncTCP.cpp
+++ b/hardware/ASyncTCP.cpp
@@ -3,8 +3,6 @@
 #include <boost/asio.hpp>
 #include <boost/bind.hpp>
 #include <boost/system/error_code.hpp>     // for error_code
-// todo: remove logging
-#include "./../main/Logger.h"
 struct hostent;
 
 #ifndef WIN32
@@ -14,7 +12,7 @@ struct hostent;
 #define RECONNECT_TIME 30
 
 ASyncTCP::ASyncTCP(bool secure)
-	: mIsConnected(false), mIsClosing(false), mSecure(secure),
+	: mIsConnected(false), mIsClosing(false), mSecure(secure), mWriteInProgress(false),
 	mContext(boost::asio::ssl::context::sslv23),
 	mSocket(mIos), mSslSocket(mIos, mContext), mReconnectTimer(mIos),
 	mDoReconnect(true), mIsReconnecting(false),
@@ -177,7 +175,15 @@ void ASyncTCP::write(const uint8_t *pData, size_t length)
 
 void ASyncTCP::write(const std::string &msg)
 {
-	mIos.post(boost::bind(&ASyncTCP::do_write, this, msg));
+	boost::unique_lock<boost::mutex> lock(writeMutex);
+	if (mWriteInProgress) {
+		writeQ.push(msg);
+	}
+	else {
+		mWriteInProgress = true;
+		//do_write(msg);
+		mIos.post(boost::bind(&ASyncTCP::do_write, this, msg));
+	}
 }
 
 // callbacks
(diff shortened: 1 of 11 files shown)


rkluwen:
At the same time a lot of shared pointers are removed in favor of non-pointer members.

hardware/ASyncTCP.h
```diff
@@ -1,11 +1,15 @@
 #pragma once
 
 #include <stddef.h>                        // for size_t
+#include <queue>						   // for write queue
 #include <boost/asio/deadline_timer.hpp>   // for deadline_timer
 #include <boost/asio/io_service.hpp>       // for io_service
 #include <boost/asio/ip/tcp.hpp>           // for tcp, tcp::endpoint, tcp::s...
+#include <boost/asio/ssl.hpp>			   // for secure sockets
```
gizmocuz, line 8: I think you have to use the preprocessor 'WWW_ENABLE_SSL' also in this class (openssl has to be found correct?, see webserver/server.hpp)

hardware/ASyncTCP.h
```diff
@@ -15,7 +19,7 @@ namespace boost { namespace system { class error_code; } }
 class ASyncTCP
 {
 protected:
-	ASyncTCP();
+	ASyncTCP(bool secure = false);
```
gizmocuz, line 22: could you make the argument 'const' ?

hardware/ASyncTCP.cpp
```diff
@@ -260,6 +310,18 @@ void ASyncTCP::write_end(const boost::system::error_code& error)
 				StartReconnect();
 			}
 		}
+		else {
+			boost::unique_lock<boost::mutex> lock(writeMutex);
+			if (writeQ.size() > 0) {
```
gizmocuz, line 315: Should this be a 'while > 0' ? could it be that there are more items in writeQ ?
Please use if (!writeQ.empty())

hardware/ASyncTCP.h
```diff
@@ -83,6 +90,11 @@ class ASyncTCP
 	std::shared_ptr<std::thread> 	m_tcpthread;
 	boost::asio::io_service::work 	m_tcpwork; // Create some work to keep IO Service alive
 
+	boost::asio::ssl::context		mContext; // ssl context
 	boost::asio::ip::tcp::socket	mSocket;
+	boost::asio::ssl::stream<boost::asio::ip::tcp::socket> mSslSocket; // the ssl socket
 	boost::asio::ip::tcp::endpoint	mEndPoint;
+	std::queue<std::string>			writeQ; // we need a write queue to allow concurrent writes
```
gizmocuz, line 97: would a dequeue be better/faster?

webserver/proxyclient.h
```diff
@@ -23,45 +24,27 @@ namespace http {
...
-				size_t bytes_transferred);
-
-			void ReadMore();
-
 			void WebsocketGetRequest();
 			bool parse_response(const char *data, size_t len);
 			std::string websocket_key;
 			std::string compute_accept_header(const std::string &websocket_key);
-			std::string SockWriteBuf;
+			std::string readbuf;
```
gizmocuz, line 45: If possible, please start 'all' member variables with m_

gizmocuz (commented):
Thanks for the PR, some small questions/suggestions

rkluwen:
Should this be a 'while > 0' ? could it be that there are more items in writeQ ?

- No, there should be only one outstanding async_write(). That's exactly the purpose of writeQ.

would a dequeue be better/faster?

- I'm not sure

About the other comments: Roger & Wilco. I will implement those and come back with an update on this PR.

rkluwen:
The pull request doesn't work on the Pi.

hardware/ASyncTCP.h
rkluwen: Doine

hardware/ASyncTCP.h
rkluwen: Done

hardware/ASyncTCP.cpp
rkluwen: writeQ,empty() used now. The other comment was a negative.

hardware/ASyncTCP.h
rkluwen: Yeah... Makes much more sense. Done.

webserver/proxyclient.h
rkluwen: Done.

Issue #2842, opened 2018-11-06
Access credentials for IP cameras are not urlencoded when used


sq5gvm:
(tested on 4.10174)

Short story:
Domoticz does not url encode password (maybe username too) when requesting images from cameras. May work in many cases but will probably never work when login/password contains : (used to separate login and password in url) or @ (used to separate password and host in url).

Longer explanation:
Let's assume that I've set my IP camera so it requires authentication with "secure" credentials:
user: admin
pass: Admin1@
If, while setting up camera in domoticz, I enter "Admin1@" as password, quick preview/test will work, but domoticz will not be able to display it (when accessing thru /camsnapshot.jpg?idx=). 
On the other hand, if I encode special characters and enter password as "Admin1%40" domoticz wil be able to access camera.
In the first case domoticz requests url: http://admin:Admin1@@1.1.1.1/source and gets "401 Unauthorized".

Example:
![screenshot_2018-11-06](https://user-images.githubusercontent.com/2500520/48036796-28acdb00-e16a-11e8-8a65-154e46d40f0e.png)
First camera set with password with no url encoding in configuration, second with manually url encoded password.

Additionally - maybe there should be some kind of notice that "test connection" works by requesting image from user local machine (browser) and not from domoticz host and also it's using two different mechanisms for that.

sq5gvm:
IMHO user and password should probably be url encoded here: 
https://github.com/domoticz/domoticz/blob/development/main/Camera.cpp#L135

gizmocuz:
Could you try this ?

sq5gvm:
What do you mean by "this"? I don't see any changes that could address this issue.

gizmocuz:
i mean, add the code to encode it, and test if that works, and if it works, please make a pull request

sq5gvm:
If I had development env that would allow me to make and test quick fix I would have already submitted a patch. But ok, I'll try to install dev env and fix it hopefully in the upcoming days.

gizmocuz:
Thanks for the PR... did you test the patch as i do not see any feedback here ?

Also you do not have to cast to c_str() when you are using a stringstream object, and you don't need to provide a const char* to the function.

can you modify your patch to

`s_str << szURLPreFix << "://" << CURLEncode::URLEncode(pCamera->Username) << ":" << CURLEncode::URLEncode(pCamera->Password) << "@" << pCamera->Address << ":" << pCamera->Port;
`

(just update the code and commit/push should do the trick)

[...]

PR #2847, opened 2018-11-07
Fixes issue #2842

sq5gvm:
This pull request fixes issue #2842 - Username and password used in URL for downloading images from cameras should be url encoded.


gizmocuz:
Thanks! Could you remove the comments as well ?

gizmocuz:
Thanks!!

PR #2851, opened 2018-11-08
Add parsing of current readings to P1MeterBase code

fjgalesloot:
The DSMR telegram contains current values for all phases L1, L2, L3. The current P1MeterBase code does not parse those values.

This PR adds code to read those values and output them to a current sensor.

### DIFF
--- a/hardware/P1MeterBase.cpp
+++ b/hardware/P1MeterBase.cpp
@@ -33,7 +33,7 @@ enum _eP1MatchType {
 #define P1VOLTL3	"1-0:72.7.0"	// voltage L3 (DSMRv5)
 #define P1AMPEREL1	"1-0:31.7.0"	// amperage L1 (DSMRv5)
 #define P1AMPEREL2	"1-0:51.7.0"	// amperage L2 (DSMRv5)
-#define P1AMPEREL3	"1-0:71.7.0"	// amerage L3 (DSMRv5)
+#define P1AMPEREL3	"1-0:71.7.0"	// amperage L3 (DSMRv5)
 #define P1GTS		"0-n:24.3.0"	// DSMR2 timestamp gas usage sample
 #define P1GUDSMR2	"("		// DSMR2 gas usage sample
 #define P1GUDSMR4	"0-n:24.2.1"	// DSMR4 gas usage sample
--- a/hardware/P1MeterBase.cpp
+++ b/hardware/P1MeterBase.cpp
@@ -31,6 +31,9 @@ enum _eP1MatchType {
 #define P1VOLTL1	"1-0:32.7.0"	// voltage L1 (DSMRv5)
 #define P1VOLTL2	"1-0:52.7.0"	// voltage L2 (DSMRv5)
 #define P1VOLTL3	"1-0:72.7.0"	// voltage L3 (DSMRv5)
+#define P1AMPEREL1	"1-0:31.7.0"	// amperage L1 (DSMRv5)
+#define P1AMPEREL2	"1-0:51.7.0"	// amperage L2 (DSMRv5)
+#define P1AMPEREL3	"1-0:71.7.0"	// amerage L3 (DSMRv5)
 #define P1GTS		"0-n:24.3.0"	// DSMR2 timestamp gas usage sample
 #define P1GUDSMR2	"("		// DSMR2 gas usage sample
 #define P1GUDSMR4	"0-n:24.2.1"	// DSMR4 gas usage sample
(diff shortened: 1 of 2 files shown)


gizmocuz:
Thanks !!

gizmocuz:
Corrected the ID of the current sensor , please do not use 0

gizmocuz:
Is it a coincidence i have rounded values at the moment (2.0 A, 3.0A, 2.0A) ?

fjgalesloot:
No that is no coincidence. The smart meters report the current in integer values.

gizmocuz:
Thats a pitty... i don't think this value is of any use then.
You can calculate it, by using P=U*I ==> I=P/U
Now a sensor is added to the internal database, used in calculations, scripts etc, and it's not accurate.
Shall we remove it ?

fjgalesloot:
It is a pitty indeed…
But I still have a use case for it. To charge my electric cars, I built a charge station. I want to use as much current as possible for the cars, so I need the know the current in use from the smart meter. Then I can load-balance the cars and the house, so there is always as much current flowing into the cars as possible.
As the circuit breakers are triggered on the amount of current passing through I need the Current or the Apparent Power and not the Real Power, that the smart meter outputs. To calculate the Current or Apparent Power from the Real Power, you need the power factor as well (https://en.wikipedia.org/wiki/Power_factor). I think the easiest way (at least for me) is to read the current values from the smart meter. To calculate the current from the Power readings from the meter, we will have to know the power factor.

Maybe this is a specific use case, but I think more people can benefit from it. Maybe I could make the current reading an optional feature in the hardware setup of the P1 meter with an explanation about the values being integers.

gizmocuz:
Is the power factor supplied in P1 protocol ?
But even without the power factor, it is more accurate then what we have now ? (the rounding to whole numbers) correct ?

fjgalesloot:
I don’t think the power factor it is supplied by the smart meter. So I guess the accuracy of calculating the current versus reading it depends on value of the power factor, thus is unknown. I don’t know how the meter rounds the current values though.

In my case I have to deal with a max value, so the accuracy of everything below the max is not really important to me. When the usage of my charging station exceeds the maximum allowed for my mains (25A for me) the main circuit breaker will cut out and I will have to call the electricity company to fix that. Using the current value supplied by the meter looks like the safest route to determine that max value.

So I guess it depends on the use case wether this addition can be useful. ;-)

gizmocuz:
Thats true... are you able to make a dzVents script that creates a current sensor based on voltage/power ?

fjgalesloot:
I was planning on toying with dzVents, so I can try. If not, a LUA script can do that calculation as well.
Do you want me to compare those values to the integer values supplied by the meter?

JVaassen:
According to what I've seen on the smart meter installed last week, the actual power out per phase (L1/L2/L3) is also reported in the dataset. Whether that is compensated for power factor I don't know. But at least it has more significant digits than the current... I can take some time this week and also add the other values in the data-package to domoticz. It seems pretty straightforward.

gizmocuz:
@llagendijk , that would be great.
@JVaassen , yes indeed. That's why i think we should drop the current , and if someone wants the current, make a (virtual) sensor with Lua/dzVents

fjgalesloot:
I did some googling on a typical value of the power factor in households. It seems the value will be somewhere between 0.7 and 0.9 typically (forgot to make note of the URLs though).  The power values reported by the meter are RMS power, so these are corrected. We need the 'uncorrected' power (or apparent power) to calculate the current accurately.

Calculating the current only with the voltage and (RMS) power will actually be less accurate compared to the rounded values supplied by the meter, especially for 'higher' current situations. 
Let's say the RMS power reading is 2300W with 230V voltage level. The calculated current would be 10A assuming a power factor of 1. But if the power factor would be say 0.9, the 'real' current flowing through the wires would be 11,1A. Rather significant difference using only the 'best' power factor in the typical range. It gets worse when the power factor is less.

Perhaps there is a way to calculate the power factor without additional hardware, but I do not know of any...

gizmocuz:
Isn't the value reported for power taking the power factor into consideration as well ?
Isn't the voltage reported the actual voltage, and the power reported the actual power ?
I=p/u

fjgalesloot:
According to what I found (and what you expect to pay for to the electricity company) the power reported is the real power used (so with the power factor correction), not the actual power needed to calculate the current.

gizmocuz:
For now i removed the current sensors from the P1 Smart Meter hardware

PR #2854, opened 2018-11-09
EventSystem Lua AFTER support for OpenURL

jvandenbroek:
Quick fix for forcing correct order calling multiple URLs.


gizmocuz:
Thanks for this PR!! Could you name a use-case for this behavior ?
If there is a logical use-case, then also the blockly widget should be updated.
Is there a minimum requirement of 0.2f ? In blockly there is nu 'after' currently, what happens when this will be 0.0 ?

jvandenbroek:
I use it to do 2 different calls to my Kodi mediacenter, which need to be executed in order to function. When both are send with the same default delay, it's 50/50 change which one gets called first. Simply calling the second one with 0.3 or 0.4 delay will do the trick. 

I think it's also nice to have it for other use cases, so you can use AFTER just like you'd use it for devices or variables. You could also do a local call to DZ after a predefined delay if for some reason you'd want to do something.

I'm not using Blockly, so just skipped it. Also need to look at it deeper, this was just easier to do ;) But sure, I can look into it if you want. But if you could update it for Blockly, that would be great ;)

I think the default delay of 0.2 sec was introduced before the event queue was implemented, which could cause lockups. But to be on the safe side, I decided for this patch just to keep it. But I think it's not necessary anymore, a test with AFTER 0 does work fine here. Also now tested Blockly with this patch, URL gets called and no other issues.

gizmocuz:
Thanks for the verification.
But if you call two url's after each other, they should be handled in sequence... at least, in domoticz... what happens in kodi i don't know... 
If you could have a look at the corresponding Blockly and Lua part as well, that would be great! Also the wiki needs updating.
(Then it's a complete patch....)

jvandenbroek:
I'm not sure if the commandArray get's executed in order as well? I'm using commandArray[#commandArray + 1] in a single script to execute 2 OpenURL actions.. You could be right, but my testings with AFTER 0.3, so 0.1 sec later than the first one, seems to always work as intended now (while before it didn't).

Will look into Blockly part as well!

jvandenbroek:
Added Blockly support for after, tested and works as expected. Also replaced redundant AFTER code with ParseActionString function.

gizmocuz:
Thanks!!

PR #2856, opened 2018-11-09
Enable domoticz to run as non-root

tvwerkhoven:
PIDFILE is now writable by regular user
do_start() now also starts domoticz as regular user (testing if it was already done as regular user?)

### DIFF
--- a/domoticz.sh
+++ b/domoticz.sh
@@ -15,12 +15,14 @@ PATH=/sbin:/bin:/usr/sbin:/usr/bin:/usr/local/bin
 DESC="Domoticz Home Automation System"
 NAME=domoticz
 USERNAME=pi
-PIDFILE=/var/run/$NAME.pid
+# Use this PID file to run as regular user
+PIDFILE=/tmp/$NAME-$USERNAME.pid
+
 SCRIPTNAME=/etc/init.d/$NAME
 
 DAEMON=/home/$USERNAME/domoticz/$NAME
 DAEMON_ARGS="-daemon"
-#DAEMON_ARGS="$DAEMON_ARGS -daemonname $NAME -pidfile $PIDFILE"
+DAEMON_ARGS="$DAEMON_ARGS -daemonname $NAME -pidfile $PIDFILE"
 DAEMON_ARGS="$DAEMON_ARGS -www 8080"
 DAEMON_ARGS="$DAEMON_ARGS -sslwww 443"
 #DAEMON_ARGS="$DAEMON_ARGS -log /tmp/domoticz.txt"
@@ -59,8 +61,9 @@ do_start()
         #   2 if daemon could not be started
         start-stop-daemon --chuid $USERNAME --start --quiet --pidfile $PIDFILE --exec $DAEMON --test > /dev/null \
                 || return 1
-        start-stop-daemon --start --quiet --pidfile $PIDFILE --exec $DAEMON -- \
-                $DAEMON_ARGS \
+
+        start-stop-daemon --chuid $USERNAME --start --quiet --pidfile $PIDFILE --exec $DAEMON -- \
+		$DAEMON_ARGS > /dev/null \
                 || return 2
 }
 
--- a/domoticz.sh
+++ b/domoticz.sh
@@ -19,12 +19,14 @@ USERNAME=pi
 PIDFILE=/tmp/$NAME-$USERNAME.pid
 
 SCRIPTNAME=/etc/init.d/$NAME
+SSLCERT=/home/$USERNAME/domoticz/server_cert.pem
+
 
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thanks for the PR! How would this go for sysFS, GPIO, motherboard sensors like temp/cpu and other processes that requires root privileges ?

xbeaudouin:
Maybe in this case a warning should be done. On my FreeBSD port, I have added a defaut user "domoticz", but some people may want to run it as root for the same purpose, but this is really end user controled...

tvwerkhoven:
> Thanks for the PR! How would this go for sysFS, GPIO, motherboard sensors like temp/cpu and other processes that requires root privileges ?

Good point, not sure, but I think/suspect `USERNAME=root` could do the trick. In any case I think it's a good idea to be able to run as non-root user.

domoticz.sh
```diff
@@ -15,12 +15,14 @@ PATH=/sbin:/bin:/usr/sbin:/usr/bin:/usr/local/bin
...
-PIDFILE=/var/run/$NAME.pid
+# Use this PID file to run as regular user
+PIDFILE=/tmp/$NAME-$USERNAME.pid
+
 SCRIPTNAME=/etc/init.d/$NAME
 
 DAEMON=/home/$USERNAME/domoticz/$NAME
 DAEMON_ARGS="-daemon"
-#DAEMON_ARGS="$DAEMON_ARGS -daemonname $NAME -pidfile $PIDFILE"
+DAEMON_ARGS="$DAEMON_ARGS -daemonname $NAME -pidfile $PIDFILE"
```
gizmocuz, line 27: The PID file is already specified as argument of the start-stop-daemon command... could be removed here. Maybe this parameter could be changed to include the 'chuid' ?

gizmocuz:
Some tests would be great to know if this could cause issues for current (raspberry) users

tvwerkhoven:
> Thanks for the PR! How would this go for sysFS, GPIO, motherboard sensors like temp/cpu and other processes that requires root privileges ?

One solution is to add the user running domoticz to the groups that have access to that hardware (for example https://askubuntu.com/a/112572). I think running domoticz as root is a security risk and not required most of the time.

domoticz.sh
tvwerkhoven: Not sure if that's true, it could be that `start-stop-daemon` also needs to know the pid in order to check if it's running (e.g. when restarting). `chuid` could be included here indeed.

tvwerkhoven:
I've just added another fix I ran into: when running as non-root/non-pi user, 1) the ssl cert cannot be found, explicitly adding the path fixes that. I don't know why this is, but I don't see where the default value is set in the source. 2) the port should be >1024.

domoticz.sh
```diff
@@ -15,14 +15,18 @@ PATH=/sbin:/bin:/usr/sbin:/usr/bin:/usr/local/bin
...
+SSLCERT=/home/$USERNAME/domoticz/server_cert.pem
+
 
 DAEMON=/home/$USERNAME/domoticz/$NAME
 DAEMON_ARGS="-daemon"
-#DAEMON_ARGS="$DAEMON_ARGS -daemonname $NAME -pidfile $PIDFILE"
+DAEMON_ARGS="$DAEMON_ARGS -daemonname $NAME -pidfile $PIDFILE"
 DAEMON_ARGS="$DAEMON_ARGS -www 8080"
-DAEMON_ARGS="$DAEMON_ARGS -sslwww 443"
+DAEMON_ARGS="$DAEMON_ARGS -sslwww 10443 -sslcert $SSLCERT"
```
gizmocuz, line 29: This will break all current raspberry installations where people expect it to be found on port 443 (and probably other installations as well)
In the source it should load the ssl certification from the path where the executable resides (/home/$USERNAME/domoticz), this has always worked before...
need to double check this

tvwerkhoven: I realise this will break compatibility, so it shouldn't be default or at least give a warning.

Regarding server_cert.pem: I found it in the source here: 

https://github.com/domoticz/domoticz/blob/72aea2965b0bf63b2b7be3fcbd8d9cb215bf68bf/main/mainworker.cpp#L211.

It's a relative path and appears to depend on the location where domoticz is run: if I run it manually (`/home/domoticz/domoticz/domoticz -daemonname domoticz -pidfile /tmp/domoticz-domoticz.pid -www 0 -sslwww 10443`) it doesn't work if I'm not in the domoticz dir, but if I start it from `/etc/init.d/domoticz.sh` it does work. Perhaps it's an idea to add this to `/etc/domoticz/setupVars.conf` to make it an absolute path.

gizmocuz: Nope, best to use /etc/domoticz.conf, see the example configuration in the git repo.

PR #2860, opened 2018-11-12
Add Telegram as notify option in domoticz.notify 

rwaaren:
By adding the line ['NSS_TELEGRAM'] = 'telegram', to the definitions in Domoticz.lua, this notification method can be used in domoticz.notify


rwaaren:
Looks like I used the wrong branch

gizmocuz:
correct, you need to patch against the default branch (development)

PR #2861, opened 2018-11-12
Update Domoticz.lua to include telegram as notification method

rwaaren:
Add the possibility to use Telegram as notification method in domoticz.notify


dannybloe:
Hah.. and you thought you were ready ;-)
Next step is to update the version number everywhere and update the readme.md file and history.md (and make sure the history list in the readme at the end is the same. And, if you are thorough you generate the wiki document using pandoc (instructions are in the documentation folder) and commit that as well.

gizmocuz:
Thanks for the PR... @dannybloe , that's indeed a lot of work for 1 line ;) but indeed, the documentation should be up2date (and also the wiki)

rwaaren:
OK will try and create a new PR for this

PR #2864, opened 2018-11-14
Fix EvoHome hot water graphs

sidepipe:
Identical to the fix in merge 38486716 from gordonb3, but include "Hot Water" sub type as being a temperature device.


gizmocuz:
Thanks!!

PR #2874, opened 2018-11-19
Implementing P1 Power / Delivery per phase (DSMR5)

JVaassen:
Added Power Usage and Power Delivery sensors.
Test on Linux PC with P1 meter. Usage sensors are seen.
Fixes contains fixes after testing/reviewing myself.


gizmocuz:
Are you sure this is only for (E)SMR 5 ? I found an old document (i think this is from version 3) that already explains it

https://www.netbeheernederland.nl/_upload/Files/Slimme_meter_15_a727fce1f1.pdf

JVaassen:
I would have to look that up and then see whether it is easy to also
support older versions.  Disadvantage is that I can only test v5.

Regards

Jos

gizmocuz:
Thanks for the PR! I will test this on an old meter

PR #2875, opened 2018-11-20
Honeywell EvoHome Radiator Relay/Valve position metrics.

cocooma:
Make the Honeywell EvoHome Radiator Relay/Valve position metrics
 available for external metrics collectors.


cocooma:
@gizmocuz thx for implementing the above.

gizmocuz:
I implemented it as 'value', is it in 'percentage' ? (0-100), in that case, please make a PR for using 'Percentage'

cocooma:
It should be percentage, but it seems like it measures between 0-200. I think it follows the LUX sensor scale. I need to figure out where to change the scale to make it run between 0-100.

PR #2877, opened 2018-11-21
Update MQTT.cpp

jonferreira:
adds support for Stop command

### DIFF
--- a/hardware/MQTT.cpp
+++ b/hardware/MQTT.cpp
@@ -227,8 +227,8 @@ void MQTT::on_message(const struct mosquitto_message *message)
 		else if (szCommand == "switchlight")
 		{
 			std::string switchcmd = root["switchcmd"].asString();
-			if ((switchcmd != "On") && (switchcmd != "Off") && (switchcmd != "Toggle") && (switchcmd != "Set Level") && (switchcmd != "Stop"))
-				goto mqttinvaliddata;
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thanks for the PR!
Question, there are "Many" more switch commands... i don't think we need a validation check...
It's nice, but it depends on the switch type. Have a look at RFXNames.cpp, for example:

"Group On"
"Group Off"
"Set Group Level:"
"Dim"
"Bright"
"Sound 0 till Sound 8"
"Set to White"
Set Color
NightMode
Disco Mode 1 till 9
"Arm Home",
"Disarm",
"Light Off",
"Normal"
...
...
a lot of commands here....

jonferreira:
OK I wasn't aware of all those but in that case maybe remove the validation entirely then?

gizmocuz:
Yes, maybe comment it out for now...

jonferreira:
as requested

gizmocuz:
Not working.... please always test the code, at the very extreme least, compile it

jonferreira:
sorry had a typo. my tests instance is down so don't have anywhere to test it

gizmocuz:
Thanks!!!

jonferreira:
No problem, pretty basic PR.

Just for curiosity, how does one know if PR X is part of beta / stable release Y?

gizmocuz:
All PR's are based against the beta version. Releases are made from time to time...
When we are going to make a new release (should happen not to long from now) there will be a code freeze. At that time, only mandatory patches are taken and merged in both beta and the pre-release branch

PR #2885, opened 2018-11-26
Evohome Web API corrections

gordonb3:
Honeywell added some limits to the usage of the Evohome web portal, causing Domoticz HTTP client to return more unspecified errors. This PR changes the session renewal method as well as changes all accesses to the httpclient class to functions that allow access to the return codes and respond to them in a sane manner.


gizmocuz:
Thanks!!

PR #2887, opened 2018-11-26
Honeywell thermostat fixes and improvements

roldoe:
- Make the API Key and API secret configurable because of quota limitations of the Honeywell API
- Fixed minor bugs in the OAuth2 handling
- Added two new devices Away (geofence) and Equipment state

More info on Domoticz Forum: https://www.domoticz.com/forum/viewtopic.php?f=31&t=22430&start=20

### DIFF
--- a/hardware/Honeywell.cpp
+++ b/hardware/Honeywell.cpp
@@ -63,7 +63,7 @@ CHoneywell::~CHoneywell(void)
 
 void CHoneywell::Init()
 {
-	mTokenExpires = std::time(0);
+	mTokenExpires = mytime(NULL);
 }
 
 bool CHoneywell::StartHardware()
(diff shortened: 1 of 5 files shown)


gizmocuz:
Thanks for the PR!!!

Could you replace std::time with normal time (without std?)

It looks like you now have to provide a API key in the hardware setup, but it is optional by the looks of the code.
Maybe you could make a link to where to generate such an API key in the hardware setup (html) like we do for some notification systems (for example setup->settings->notifications->prowl, To get a free account/API key click >>Here<<)

roldoe:
Replaced the std::time with normal time. 

The embedded (default) API key is being used by too many users. This causes a quota exceeded response from the Honeywell API server. It takes 10 minutes to hit the quota per hour limit. During the next 50 minutes no more data can be pulled from the Honeywell API server.
To keep it backwards compatible I made the API key optional. The old API key is functional if the user is satisfied with an update of once an hour. :-) 

I've included a link to the Domoticz wikipedia describing how to get the necessary keys. The link was already included on line 1699 of www/view/hardware.html.

gizmocuz:
Great thanks !!!

PR #2891, opened 2018-11-30
Restructure HTTPClient class

gordonb3:
A user pointed me to an issue with HTTPClient class allowing other classes to set timeout values as a global parameter. This causes classes that need to access resources on the internet to fail if another class sets the timeout values expecting to talk to a local(net) resource.

While investigating I found that the various web access methods did not all offer the same functionality, with only one of the GET functions allowing an override of the global timeout value. I have regrouped the functions according to their input parameter complexity and extended functionality where needed. This includes allowing timeout override for all web access methods and allow access to either the HTTP or Curl return code which was previously only offered by the GET and POST methods.

This PR includes an update to EvohomeWeb class making use of these extended functions, because this was the reason for wanting to redo the HTTPClient class.

### DIFF
--- a/hardware/EvohomeWeb.cpp
+++ b/hardware/EvohomeWeb.cpp
@@ -25,6 +25,7 @@
 #define LOGONFAILTRESHOLD 3
 #define MINPOLINTERVAL 10
 #define MAXPOLINTERVAL 3600
+#define HTTPTIMEOUT 30
 
 #ifdef _WIN32
 #define gmtime_r(timep, result) gmtime_s(result, timep)
@@ -1907,7 +1908,7 @@ std::string CEvohomeWeb::send_receive_data(std::string url, std::vector<std::str
 	std::vector<unsigned char> vHTTPResponse;
 	std::vector<std::string> vHeaderData;
 
-	bool httpOK = HTTPClient::GETBinary(url, headers, vHTTPResponse, vHeaderData, -1);
+	bool httpOK = HTTPClient::GETBinary(url, headers, vHTTPResponse, vHeaderData, HTTPTIMEOUT);
 
 	return process_response(vHTTPResponse, vHeaderData, httpOK);
 }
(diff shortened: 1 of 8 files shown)


gordonb3:
Note: I added MainWorker as a friend class to HTTPClient with the intention to allow this class to call the Cleanup() function and at the same time block any other class from doing so by making this function protected. I have actually not added this to MainWorker and thus as it was Cleanup() is not called from anywhere, even though the already existing comment states it should.

gizmocuz:
Thanks for the great rework !! One question/comment, why choose for a timeout in MS ? Everywhere you now issue *1000, i also don't think we ever going to need a that accurate timeout... could you redo this ?

gordonb3:
That is actually because of logitech media server. The setup routine of that class asks to enter the timeout in milliseconds, which I think makes sense. The next thing happening though is that it divides that number by 1,000 which is then treated as a `long`. In effect, if you enter something like 400ms this means that timeout will be disabled.

I agree that the multiplication is ugly, but this is needed for backward compatibility. Another option would be to change all existing calls to these functions. I think there's probably just about 10-20 throughout the whole code.

gordonb3:
I mean this part here:
https://github.com/domoticz/domoticz/blob/development/hardware/LogitechMediaServer.cpp#L66..L71

 	sPostData << sPostdata;

	HTTPClient::SetTimeout(m_iPingTimeoutms / 1000);
	bool bRetVal = HTTPClient::POST(sURL.str(), sPostData.str(), ExtraHeaders, sResult);

	if (!bRetVal)

gizmocuz:
Thanks, but then it's better to adjust those calls/parameters.... a timeout in MS makes perfect sense for worker threads, but not for HTTP calls... setting up a connection (dns lookup etc) already takes time
Looking at the code, it's only the logitech media class doing this.
I actually would change that particular call to SetTimeout(5) ... which should be more then sufficient when using this internally

gordonb3:
Yes. Quick scan shows that other classes changing the timeout value use fixed values. LMS for some reason has this user configurable, with a default of 1000 milliseconds.

Should we just remove that setup option?

gizmocuz:
Yes that would be great! maybe using a default of 5 seconds for now until we start to hear objections

gizmocuz:
What is this function?

 //Cleanup function, should be called before application closed
static void Cleanup();

i do not see a implementation or call... maybe it should be removed?

gordonb3:
Looks like I missed one call attempting to set the timeout...

Yes: Cleanup() was never called. Seems to intent clearing resources that may otherwise remain in memory after the program is closed. Adding it now.

gizmocuz:
Thanks !!!

PR #2899, opened 2018-12-04
 start implementing the use of SendTemp() method in order to get adju…

pipiche38:
I have started the implement of using SendTemp ... instead of the m_sql.Update(), however I'm stop at compilation due to the fact that those methods are not in the scope.

So I'm pushing this code in order to get support:

[ 22%] Building CXX object CMakeFiles/domoticz.dir/hardware/plugins/PythonObjects.cpp.o
/usr/share/domoticz/pipiche-domoticz/hardware/plugins/PythonObjects.cpp: In function ‘PyObject* Plugins::CDevice_update(Plugins::CDevice*, PyObject*, PyObject*)’:
/usr/share/domoticz/pipiche-domoticz/hardware/plugins/PythonObjects.cpp:850:20: error: ‘SendTempHumBaroSensorFloat’ was not declared in this scope
                    SendTempHumBaroSensorFloat((const unsigned char)self->Unit, iBatteryLevel, temp, hum, baro, nforecast, sName, iSignalLevel);
                    ^~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/share/domoticz/pipiche-domoticz/hardware/plugins/PythonObjects.cpp:857:13: error: ‘SendTempHumSensor’ was not declared in this scope
             SendTempHumSensor((const unsigned char)self->Unit, iBatteryLevel, temp, hum, sName, iSignalLevel);
             ^~~~~~~~~~~~~~~~~
/usr/share/domoticz/pipiche-domoticz/hardware/plugins/PythonObjects.cpp:871:13: error: ‘SendTempSensor’ was not declared in this scope
             SendTempSensor((const unsigned char)self->Unit, iBatteryLevel, temp, sName, iSignalLevel);
             ^~~~~~~~~~~~~~
make[2]: *** [CMakeFiles/domoticz.dir/build.make:2429: CMakeFiles/domoticz.dir/hardware/plugins/PythonObjects.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:73: CMakeFiles/domoticz.dir/all] Error 2
make: *** [Makefile:152: all] Error 2

### DIFF
--- a/hardware/plugins/PythonObjects.cpp
+++ b/hardware/plugins/PythonObjects.cpp
@@ -838,8 +838,42 @@ namespace Plugins {
 			}
 			if (!SuppressTriggers)
 			{
-				DevRowIdx = m_sql.UpdateValue(self->HwdID, sDeviceID.c_str(), (const unsigned char)self->Unit, (const unsigned char)self->Type, (const unsigned char)self->SubType, iSignalLevel, iBatteryLevel, nValue, sValue, sName, true);
+				// Translate from iType, iSubType and iSwitchType if we have Temp+Hm+Baro, Temp+Hum or Temp
 
+				_log.Log(LOG_NORM, "(%s) %d/%d/%d Updating device from %d:'%s' to have values %d:'%s'.", sName.c_str(), iType, iSubType, iSwitchType, self->nValue, PyUnicode_AsUTF8(self->sValue), nValue, sValue);
+
+				if ((iType == pTypeTEMP_HUM_BARO  ) and ( iSubType == sTypeTHB1 ))
+				{ // Temp+Hum+Baro: sValue = "0.0;50;1;1010;1"
+					int temp, hum, baro, nforecast;
+
+					sscanf( sValue, "%d;%d;%d;%d;%d", &temp, &hum, &baro, &nforecast);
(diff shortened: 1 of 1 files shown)


dnpwwo:
My feedback is not to go this way.

IMHO the C++ device handling inside Domoticz is a bit of a mess (check the forum for '<device type> doesn't seem to update properly' etc) so I specifically didn't put any of this in the framework.

Someone started a nice python module that handled devices more specifically.  If you want specific functionality I would recommend finding that and maybe uplifting it to be be officially supported.  There are a few plugins that use it already.  It has a base device with a device hierarchy to handle specific cases.

dnpwwo:
Have a look at this: https://github.com/stas-demydiuk/domoticz-zigbee2mqtt-plugin/tree/master/devices

pipiche38:
@dnpwwo 
`IMHO the C++ device handling inside Domoticz is a bit of a mess (check the forum for ' doesn't seem to update properly' etc) so I specifically didn't put any of this in the framework.`

Thanks for taking the time. 
 I could agree, as it looks a bit strange to have this 'adjustment' value taken in consideration during Update, while I would consider them to be taken during display and so it would be independent of the raw data in the database. This is the main difference between Windows OS and other modern OS which store a Universal time on the clock and use the TZ+DST for display.

gizmocuz:
Okey lets close this then

PR #2901, opened 2018-12-05
fix compiler error in SQLHelper due to missing newline at end of file

gordonb3:
See C++03 Standard [2.1.1.2]

Note:
According to the specifications for C++11 this should no longer apply
because the compiler is supposed to handle the file as if it has the
newline appended to it. Although GNU 7.3 compiler (and possibly others)
state that it does add the additional newline, this still causes the
make command to terminate during the first run and you have to run it
a second time to complete the job.


gizmocuz:
Thanks!

PR #2909, opened 2018-12-11
refactor P1MeterBase

gordonb3:
Points of interest:
 - fully ignore telegrams that are to be discarded due to rate limiting
 - don't retest the same character sequences over and over again
 - drop L1 instantaneous power devices on single phase installations
   (this data is the same as the total instantaneous power)

To do:
 - add a switch for the tariff indicator (possible use in scripts)
 - add a web helper for advanced settings (e.g. manual selection of gas m-bus channel)

### DIFF
--- a/hardware/P1MeterBase.cpp
+++ b/hardware/P1MeterBase.cpp
@@ -137,21 +137,22 @@ void P1MeterBase::Init()
 	l_exclmarkfound = 0;
 	l_bufferpos = 0;
 
-	m_voltagel1 = 0;
-	m_voltagel2 = 0;
-	m_voltagel3 = 0;
+	m_voltagel1 = -1;
+	m_voltagel2 = -1;
+	m_voltagel3 = -1;
 
-	m_amperagel1=0;
-	m_amperagel2=0;
-	m_amperagel3=0;
+	m_bReceivedAmperage = false;
+	m_amperagel1 = 0;
+	m_amperagel2 = 0;
+	m_amperagel3 = 0;
 
-	m_powerusel1 = 0;
-	m_powerusel2 = 0;
-	m_powerusel3 = 0;
+	m_powerusel1 = -1;
+	m_powerusel2 = -1;
+	m_powerusel3 = -1;
 
-	m_powerdell1 = 0;
-	m_powerdell2 = 0;
-	m_powerdell3 = 0;
+	m_powerdell1 = -1;
+	m_powerdell2 = -1;
+	m_powerdell3 = -1;
 
 	memset(&m_buffer, 0, sizeof(m_buffer));
 	memset(&l_buffer, 0, sizeof(l_buffer));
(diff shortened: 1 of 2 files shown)


gizmocuz:
Thanks for this !!!
Question, has this been tested on a 3 phase meter and single meter ?

Isn't the voltage for all 3 phases always transmitted ? In this case, aren't you always setting 'phase' to 3 on line 385 ?

gordonb3:
I have a single phase connection, so this is untested for triple phase. I did an offline test on the logic for mapping numbers 3,5,7 and 2,4,6 to L1, L2, and L3 though and the live situation does give me the L1 instantaneous power. Since my meter is DSMR 4.2 I do not get Volts but I would be surprised if a single phase meter would display information for unconnected phases. In fact that would quickly become very puzzling because a single phase meter always displays the attached phase as being L1, even if the outside line is connected to L2 or L3.

Luxembourg specs appear to only differ with Dutch specs on the tariff indicator, so this code should be fully set to handle that as well (still requires decrypting, so input must be proxied at this time)

gordonb3:
Sorry. Just catching on with what you were asking. You mean this line:
>phase = ((m_lbuffer[4] & 0xFE) ^ 0x30) >> 1;

So what that does is subtract 1 from every odd number by applying a bitwise mask (0xFE -> 11111110), followed by an XOR with 0x30 which is the ASCII code for the number `0` and lastly a division by 2 by shifting everything to the right.

Example:
 - in OBIS ID `1-0:52.7.0` `'5'` first becomes `'4'`, then `4` and finally `2` -> L2
 - in OBIS ID `1-0:61.7.0` `'6'` stays `'6'`, then becomes `6` and finally `3` -> L3

gordonb3:
ToDos completed ;)

gordonb3:
Ow! This is a bit ugly:
Screenshot
Obviously that sensor only updates twice per day max.

gordonb3:
...and it seems it won't change what is supposed to be the unit once this is set (so `High` does not change to `Low` when it switches to tariff 1)

Suggestions?

gizmocuz:
Could you make it a switch ? or a user variable ?

gordonb3:
Yeah, I was thinking I could maybe add a modified version of the `contact` switch type. Obviously it's not a switch that users should be able to toggle.

gizmocuz:
Well sure not, but they probably could use it for some automation purposes.
I would just use a default on/off switch, nothing fancy

gordonb3:
That was the idea, yes. In any case the logic for the `contact` switch type already exists. It's just the textual state values that don't match.

gordonb3:
Screenshot of new switch

screenshot

gizmocuz:
Thanks!! Looking good, but it might be confusing as in the scripts it will be threaded as On/Off (1/0)
A normal switch would be sufficient ? 0=off=low tarrif, 1=on=high tarrif ?

There are to optional files that should not be in this PR i think

HTTPClient.cpp.orig
HTTPClient.h.orig

gizmocuz:
Or a simple user variable ?

gordonb3:
Hmmm... that's weird. Don't know where those .orig files came from. I did not touch the httpclient files.

With regard to user variable: I don't think you can set an on-change trigger on that, can you?

Yes I did consider a normal switch, but I liked the concept of the switch types `Contact`, `Door Contact` and `Dusk Sensor` better because these disallow the user to manually access it. The problem with these types was that they used states like `Open` and `Closed` and I didn't like the icons that came with them.

The logic is actually reversed because this was sort of dictated by the colour scheme: low tariff = gswitch_sOn (1) and shows green numbers in the icon.

gordonb3:
So you think I should reset the status text to the default `On` and `Off` and just leave the highlight text to indicate whether that refers to high or low tariff?

gizmocuz:
Hmmm i would use a normal On/Off switch, and keep the status text also On/Off (just a regular switch)
Where on represents a state. I would set the default name to a good name, so users will know the corresponding switch state. Like if you name the switch "P1 Low Tarrif" then when the switch is ON, it's Low Tarrif. If you name it "P1 High Tarrif" then when the switch is ON, it's High Tarrif

But a user variable might be even better.

Looking at CSQLHelper::UpdateUserVariable , it seems the EventSystem can react on UserVariable changes
If you update it via the web interface it should work, and if you call the above function from the P1 class it should work.

Looking at the code of UpdateUserVariable, i think it should combine SaveUserVariable/SetUserVariable in here, and we should drop the later..
Select ID from UserVariable, if not found, add it, else update it, but maybe thats another PR
I'm going to have a look at this now

gizmocuz:
If possible please have a look in Slack

gordonb3:
On/Off state now matched and displayed in the front-end:
screenshot1

And here are screenshots from the Hardware page:
screenshot2screenshot3

gizmocuz:
Is it a general switch ? As i see 'tarrif indicator' as switch type... don't think that will be needed? A reguar switch can handle on/off ?

gordonb3:
Yes it's a general switch. A modified version of the `smoke detector` switch type to be precise.

gizmocuz:
But do we need another type of switch ? Can't you use the default general switch which is already present in Domoticz ?

gordonb3:
No. Because then it will get `Turn On` and `Turn Off` links and allow timers to be set, which must not be possible. And the other already existing switch types have dedicated icons and redefined state values, so these are useless as well.

I think I have a better idea though. Something more generic that could also be used by those other special purpose switch types.

gizmocuz:
Thats also an idea, but we already have a contact type for a switch (lighting2)
As it's not a switch, and in the feature it needs to support more then two states, is a User Variable not a better idea ? I don't need to see the current low/high tarrif, we know when it is... But we could use it in scripts

gordonb3:
You mean the switch type `contact`? This is different. The existing `contact` type is a special purpose type, just like the tariff indicator you don't want. It is useless because it redefines the On/Off status values to Open/Closed. The subtype `contact` when similar switch types are moved to it allows for a generic rule in the front-end to disallow user initiated changing of their state.

This would thus apply to (at least) the following switch types:
- contact
- door contact
- door lock
- door lock inverted
- smoke detector
- dusk sensor
- doorbell
- motion

gordonb3:
And no I do not think a user variable is a good idea. It is way too hidden and the user cannot control whether he wants it or not. A switch apart from being visible has the added bonus of offering direct actions through the On and Off Action fields and assigning slave devices.

With respect to your comment that we know when the switch between low en high tariff occurs: I have a feeling we may see some change in there as the number of EV owners grows an nationwide usage may actually become higher at night then is is during daytime. At the same time we have a huge problem with solar panels generating too much power at certain times and the result may be that low tariff times become more or less variable.

PS The hardware type as it reads now on the switches page is: `Light/Switch, Contact, On/Off`

gordonb3:
Just forget it

gizmocuz:
I did not had time to test it, but all changes seems perfect ? what changed your mind ?

PR #2914, opened 2018-12-13
Make Telegram notification silent when priority is negative

macjl:
Hello,

I don't know if it's the right method, but after opening a topic in the forum, I was advised to open a pull request. Here is the related topic :
https://www.domoticz.com/forum/viewtopic.php?f=31&t=26083&p=200264&hilit=macjl#p200264

The main point, is to have silent notifications with the Telegram service. It's possible with Telegram API to send messages that don't produce sound or screen notifications when they are received, by adding the 'disable_notification' parameter.

Am I doing right?


gizmocuz:
Thanks!!

PR #2915, opened 2018-12-15
Blinds percentage switch fix

Pelicjan:
Fixing blinds percentage switch (before: on - LastLevel = 100, off - LastLevel = nothing changed now: on - LastLevel = 100, off - LastLevel = 0).


gizmocuz:
Okey, but now in the GUI you probably don't see what the percentage level is set to when you press the "ON" button ?
What problem did you experience ?

Pelicjan:
blinds
If you click on the right button (Close Blinds), the slider is set to 100 (LastLevel in db is 100).
If you click on the left button (Open Blinds), the slider doesn't change its position. Why? In my opinion, it should be 0.

gizmocuz:
Thanks!! Makes indeed perfect sense, but it seems it was not done so on purpose

Issue #2931, opened 2018-12-31
Please add support for double click and long press for "Xiaomi Wireless Dual Wall Switch"


itpansic:
I am using domoticz beta version of commit: [`630bf6b`](https://github.com/domoticz/domoticz/commit/630bf6bbf5067b1f1ad0f7fdcfa2e54654cfeed7)
It seems that Domoticz hardware `Xiaomi Gateway` don't recogonize `double click` and `long click` action from `Xiaomi Wireless Dual Wall Switch`.
Here are logs when i `double click` and `long click` a switch:
```
2018-12-31 17:15:50.624  Status: {"cmd":"report","model":"remote.b286acn01","sid":"158d0002c3f7bd","short_id":51955,"data":"{\"channel_0\":\"double_click\"}"}
2018-12-31 17:15:52.288  Status: {"cmd":"report","model":"remote.b286acn01","sid":"158d0002c3f7bd","short_id":51955,"data":"{\"channel_0\":\"long_click\"}"}
```

Please add support for double click and long press for `Xiaomi Wireless Dual Wall Switch`

PR #2933, opened 2018-12-31
Support double click and long press feature for "Xiaomi Wireless Dual Wall Switch"

itpansic:
Support double click and long press feature for "Xiaomi Wireless Dual Wall Switch" , issue #2931


gizmocuz:
Thanks!!

PR #2936, opened 2019-01-03
Update LightsController.js - Livolo switch - manual adding

lechunlx:
Based on experience with livolo switches and remotes (and internet), it should be defined by 16bit ID and 8bit Unit code.


gizmocuz:
Thank you for the patch. Could you change the number order so that 57, 55, 100 will become 55, 57, 100 (logical order)

gizmocuz:
Thanks!!

PR #2941, opened 2019-01-05
Show default icon based on device switch type

stas-demydiuk:


gizmocuz:
Thanks !!!!! Could not get the change working! Big thanks!

Issue #2947, opened 2019-01-07
Group/Scene state changes initiated by a group device are not passed to eventSystem


rwaaren:
In the eventSystem the group/scene state does not get updated when an underlying device has changed state. In the database the update is done (by updating nValue) but this change is not propagated to the eventSystem. A call to M_mainworker. M_eventsystem. GetCurrentScenesGroups () seems to be missing. If I add this call to SQLHelper.cpp (see code snippet)  the state change to the group/scene does make it to the eventSystem. Tested on  V4.10321

```
		//Set new Scene status
		safe_query("UPDATE Scenes SET nValue=%d WHERE (ID == %" PRIu64 ")",
			int(newValue), Idx);
            	if (m_sql.m_bEnableEventSystem)  // Only when eventSystem is active
			m_mainworker.m_eventsystem.GetCurrentScenesGroups(); 
```
My c++ skills are not yet at the level required to change this to a more efficient approach where only the affected scenes/groups (a device can belong to multiple groups/scenes) are called.
Happy to turn this into a PR if this is the right way of handling.

gizmocuz:
Thanks! PR merged

PR #2949, opened 2019-01-07
Fix issue #2947: Add call to GetCurrentScenesGroups in SQLHelper  on device state change

rwaaren:
As described in issue #2947 a change in state of an individual device can result in a state change of the groups/scenes where the changed device is part of. This group/scene state-change is performed by SQLHelper but not propagated to the event System.


gizmocuz:
Thanks !!

Issue #2927, opened 2018-12-23
rtl_433 not working with latest version.


markputters:
### Issue Guidelines

i update my rtl_433 source from "https://github.com/merbanan/rtl_433" and now i got this message in the logging.

2018-12-23 12:46:31.178 Status: Rtl433: rtl_433 startup failed. Make sure it's properly installed (rtl_433 -F csv -q -I 2 -G -f 433.92e6 -f 433.42e6 -H 60 2>/dev/null). https://github.com/merbanan/rtl_433

Mhrok:
Some hints, why it is not working now:
$ rtl_433 -F csv -q -I 2 
rtl_433 version 18.12-44-gd5bb623 branch master at 201812251302
quiet option (-q) is default and deprecated. See -v to increase verbosity
include_only (-I) is deprecated. Use -S none|all|unknown|known

Little digging,
-q is now default and not needed anymore
-I 2 is equal to -S known as says help of rtl_433

gizmocuz:
So a patch might be needed on line 398 of Rtl433.cpp ?

sergray:
Possible workaround is to get version before non-backward compatible changes https://github.com/merbanan/rtl_433/tree/a50fe6a14bc94612aa2cfc85dd9b9e7f83fbe9df

sergray:
Opened PR #2951 with a fix. In order to support broader versions of rtl_433 software, made it possible to put version dependent command line arguments as extra parameters of RTL-SDR hardware in Domoticz.

PR #2951, opened 2019-01-07
Fix issue #2927: Remove hardcoded legacy rtl_433 command line arguments

sergray:
Hardcoded arguments break RTL-SDR integration for newer versions of rtl_433 software after 18.05

It should be possible to specify extra arguments in additional command-line arguments for RTL-SDR version, depending on rtl_433 version in use, so supporting old and new versions of rtl_433 software.


gizmocuz:
Thanks !!

PR #2954, opened 2019-01-10
Managed counter report

guillaumezin:
Handle reports for Managed Counter devices, send 0 as counter value for Managed Counter devices instead of nothing

### DIFF
--- a/main/WebServer.cpp
+++ b/main/WebServer.cpp
@@ -15873,6 +15873,10 @@ namespace http {
 							sprintf(szTmp, "%d", atoi(sValue.c_str()));
 							root["counter"] = szTmp;
 						}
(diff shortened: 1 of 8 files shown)


gizmocuz:
Thank you for this PR !!!
Question, is this report view/controller very different then the current report controller/view ?
(I mean, is it better to implemented this in the original report controller/view)

guillaumezin:
No it's not very different than the Counter, but it doesn't show the counter value column, only the usage (kWh). Please wait a bit before merging, or reject, I think I made a mistake in DeviceLog.js (I use CounterLogSpline, I think I should revert to CounterLog) and I don't have access to my test setup right now.

PR #2953, opened 2019-01-10
Add On/Off chart to Lights log

stas-demydiuk:
2 new charts on Lights Log page:
* On/Off chart for current day
* On/Off chart for last 7 days

Other changes:
- Introduce [luxon.js](https://moment.github.io/luxon/) library to simplify working with Dates
- Fix Log display for MediaPlayers
- Move views next to corresponding Controller files

image

PR #2956, opened 2019-01-11
Min point width for On/Off chart in Lights log

stas-demydiuk:
Based of @sweetpants comment
https://github.com/domoticz/domoticz/pull/2953#commitcomment-31906609


gizmocuz:
Thanks!

PR #2958, opened 2019-01-11
Managed counter report

guillaumezin:
Handle reports for Managed Counter, send 0 as counter value, this time without adding a new view for Managed Counter but with slight modification to normal counter view


gizmocuz:
Thanks!!

PR #2963, opened 2019-01-12
Update dzVents to 2.4.10

rwaaren:
Added option to use afterXXX() and withinXXX() functions to updateSetPoint() in thermostat devices
Changed Z-Wave thermostat adapter to display mode as string in domoticz logfile on update
Update to integration-test and deviceTest
Update to documentation and WiKi


gizmocuz:
Thanks!

PR #2964, opened 2019-01-13
Handle other units (m3, s) for Managed Counters in reports

guillaumezin:
Mainly for future (for instance Gazpar smart gas meter)


gizmocuz:
Thanks for the PR, i would say start with 0 = kwh, then the rest... the default/current value is 0 right ?
Is it possible to change this ?

guillaumezin:
Default is "?", I will change as requested anyway

guillaumezin:
Done, I took the opportunity to correct reports for customized counters, that now display configured quantity and units in axis and total usage and hide costs that return NaN anyway

gizmocuz:
Thanks !!

Issue #2955, opened 2019-01-11
New Light log


Ittiz:
New light log feature only shows the light is on for dimmers when they are at 100%.  It seems to me it should show it as on when the dimmer is on at any percent.

stas-demydiuk:
Agree. Good catch!

Ittiz:
I'm not sure if this was caused by the new log as well, it may be a separate issue. But I also noticed the lighting graphs show the time in GMT instead of the local time zone.  The text log shows the correct local time.  So when the text log says a light was turned on at 17:00 the graph shows 22:00 since I'm in EST 5 hours behind GMT.

stas-demydiuk:
Could you please provide screenshot, because I wasn't able to reproduce this issue on my machine, have GMT+2 timezone

Ittiz:
It actually looks like it just the light percentage graph on the bottom.  The new graphs you created look like they have the correct times.

Ittiz:
Just checked and it all seems fixed now.  Keep up the good work.

PR #2968, opened 2019-01-14
Light log chart improvements

stas-demydiuk:
* #2955 Include Timezone offset in DeviceLevelChart
* Add initial value to DeviceOnOffChart


gizmocuz:
Thanks!!

PR #2971, opened 2019-01-15
Providing WWWFolder instead of WebRoot

pipiche38:
We should have access to szWWWFolder instead of szWebRoot which is pointing the Web folder for url ( Web root path (for instance mydomoticz, Domoticz url will become https://127.0.0.1:8080/mydomoticz/ )

### DIFF
--- a/hardware/plugins/Plugins.cpp
+++ b/hardware/plugins/Plugins.cpp
@@ -1134,7 +1134,7 @@ Error:
 					ADD_STRING_TO_DICT(pParamsDict, "HomeFolder", m_HomeFolder);
 					ADD_STRING_TO_DICT(pParamsDict, "StartupFolder", szStartupFolder);
 					ADD_STRING_TO_DICT(pParamsDict, "UserDataFolder", szUserDataFolder);
-					ADD_STRING_TO_DICT(pParamsDict, "WebRoot", szWebRoot);
+					ADD_STRING_TO_DICT(pParamsDict, "WWWFolder", szWWWFolder);
 					ADD_STRING_TO_DICT(pParamsDict, "Database", dbasefile);
 					ADD_STRING_TO_DICT(pParamsDict, "Version", m_Version);
 					ADD_STRING_TO_DICT(pParamsDict, "Author", m_Author);


gizmocuz:
Not sure if that's a good idea... what do you want to do with this ?

pipiche38:
It is following a discussion with @dnpwwo , we want to have access to the various Domoticz Folders from the Python Plugin framework

WWWFolder allows for instance the plugin to reach the WWWFolder/templates where custom pages could be found

gizmocuz:
The templates folder can also be accessed via HTTP...
Not sure if a "plugin/script" should be allowed access to them

gizmocuz:
I find it a security risk if scripts are allowed to access folders outside it's script folder

pipiche38:
@gizmocuz, this is not about script accessing the folder , it is Plugin accessing it . 
BTW they can do it already with full qualified or relative path to the HomeFolder, so I' very surprise that the PR has been closed without much discussion

gizmocuz:
Plugin/Script, it's the same. If that is so, maybe we should limit it

pipiche38:
I'm currently develop a full plugin for Domoticz and we have also started to use the www/template folder in order to provide Plugin specific web pages.

In order to have the plugin web pages accessing data, we need the plugin to push data to www/template, this is why it is critical to have access to the WWWFolder.

This PR was just easing the process to access it, we can however deduct the place from the HomeFolder and the StartupFolder parameters.

SO I really do not understand the issue here

gizmocuz:
@pipiche38 , a plugin should NOT have access to the WWW folder, or any folder outside the plugin's path
This is a security issue.. 
I have no problems if a script has it's own folder to store/load files from

pipiche38:
@gizmocuz I share your point, but what do you suggest to have the plugin bringing datas information the plugin web page ?

Can we go in an other direction, where the Domoticz WW Custom pages could be located somewhere else, like a plugin folder

In other terms, how a Custom Tempalte Web page ( www/templates) can get access to plugin data ?
Should we store those data directly in the Domotciz DB ?

gizmocuz:
@pipiche38 , i don't think a custom template page needs access to a plugin or it's data, nor any web page in domoticz except the hardware setup page.
All data is available via the JSON api (device status/data/....)

Maybe if you share what you want to accomplish ?

pipiche38:
I send you MP on the forum

gizmocuz:
Send you a PM back... maybe better to discuss on slack

pipiche38:
@gizmocuz , I just send you an other MP. I'm on slack, but don't know how to reach you ;-)

PR #2979, opened 2019-01-18
Add a weekly planning user interfarce for timers (heating, lights/switches, switches levels)

syrhus:
Require  to change the DeviceTimersController.js file to load the planning.js file and trigger 2 events + adding the planning.css file

image
image


gizmocuz:
Thanks!!
Could you add it as well to the html5.appcache file ?
Where is this script being called ? Could you rebuild it to AngularJS like the other controllers?

gizmocuz:
@syrhus , why you closed your PR ?

PR #2983, opened 2019-01-19
fix(python domoticz): Log should be upper case

### LINKED ISSUE #2097: Update python event handler default code
The current default event handler code when creating a new python plugin does not work. This is because of two reasons (as far as I noticed):
1) The variables defined by the domoticz.py (scripts/python/domoticz.py) are not present
     * changed_device: the current device that changed (object of Device)
     * changed_device_name: name of current device (same as changed_device.name)
     * is_daytime: boolean, true when it is is daytime
     * is_nighttime: same for the night
     * sunrise_in_minutes: integer
     * sunset_in_minutes: integer
     * user_variables: dictionary from string to value
2) The domoticz.log(message_type, message_body) function is broken. Fixing this problem would break the DomoticzEvents.Log() function so a new function must be defined.

I would recomment to remove the domoticz.py helper and replace the default python plugin code with the code from the following file: domoticz/scripts/python/script_device_demo.py

alfonso-presa:
fixes: #2097

I think there's an error in the script causing errors when using `domoticz.log` in python scripts for events. Changing `log` to `Log` seems to solve it.

### DIFF
--- a/scripts/python/domoticz.py
+++ b/scripts/python/domoticz.py
@@ -39,10 +39,10 @@ import re
 #	pass
 
 def log(*args):
-    domoticz_.log(0, " ".join([str(k) for k in args]))
+    domoticz_.Log(0, " ".join([str(k) for k in args]))
 
 def error(*args):
-    domoticz_.log(1, " ".join([str(k) for k in args]))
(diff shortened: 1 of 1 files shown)


gizmocuz:
thanks! Great find !!

PR #2995, opened 2019-01-27
Stop updating domoticz devices when Netatmo data is not updated

999LV:
If Netatmo cloud data for modules is not refreshed in the past 10 minutes (meaning either the modules/station is not connected to the cloud, or the cloud itself is no longer current), then do not keep updating domoticz devices sensors with outdated data.


hardware/Netatmo.cpp
```diff
@@ -443,6 +443,37 @@ bool CNetatmo::ParseDashboard(const Json::Value &root, const int DevIdx, const i
 
 	int batValue = GetBatteryLevel(ModuleType, battery_percent);
 
+	std::time_t tNetatmoLastUpdate = 0;
+	std::time_t tNow = time(NULL);
+
+	// initialize the relevant device flag
+	if ( m_bNetatmoRefreshed.count(ID) == 0 )
```
gizmocuz: Should this not be
`if ( m_bNetatmoRefreshed.find(ID) == m_bNetatmoRefreshed.end())`

gizmocuz:
Thanks for the PR!, I have one small comment on this

hardware/Netatmo.cpp
999LV: I understand from litterature (I am not a c guy) that both methods to test existence of a key are widely accepted as equivalent, but if you perfer .find == .end(), this is fine for me.

gizmocuz: Yes i would like to have the find option please... i think it's the clearest and fastest way...
Found some small discussion about it as well

https://stackoverflow.com/questions/25490357/checking-for-existence-in-stdmap-count-vs-find

i normally go even further and use/update the resulting item of the find command instead of writing into the map like this

`m_bNetatmoRefreshed(ID)=true`

In this case it's a very simple Boolean map, but if it contains a structure, it's easier to find and update the element

gizmocuz:
Thanks!!

PR #2997, opened 2019-01-27
Clean TRIGGER during ParseActionString

guillaumezin:
Makes Logitech Media Server device command startPlaylist works in dzVents


gizmocuz:
Thanks !!

PR #3004, opened 2019-01-31
Update html

DewGew:
This fix makes the tabs (switches, scenes, temperatures e.t.c.) to have same behavior when creating themes.

### DIFF
--- a/www/views/utility.html
+++ b/www/views/utility.html
@@ -217,7 +217,7 @@
 
 <div class="container">
 	<gz-back-to-top></gz-back-to-top>
-	<div id="utilitycontent" class="container"></div>
+	<div id="utilitycontent"></div>
 </div>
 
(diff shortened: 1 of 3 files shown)


www/views/utility.html
```diff
@@ -215,7 +215,7 @@
     </form>
 </div>
 
-<div>
+<div class="container">
 	<gz-back-to-top></gz-back-to-top>
 	<div id="utilitycontent" class="container"></div>
```
gizmocuz: i ithink the 'class="container"' needs to be removed here ?

gizmocuz:
Thanks for the PR, i have one question (see above), there is a double class="container" in utility.html with this patch, and i think the last one should be removed

DewGew:
I have Edit that one now

gizmocuz:
Thanks!!

Issue #3006, opened 2019-02-01
AnnaThermostat not working since firmware 3.1.4


Ectholian:
Since firmware 3.1.4 of Anna Thermostat (Domoticz Version: 4.9700) the plugin is not working anymore. The only thing logged is:
Error: AnnaThermostat: Invalid data received!

Scanning the source it seems a xml parsing failure, but if i try to parse the xml locally with the source it doesn't report any problems. There are more users with this problem:

https://www.domoticz.com/forum/viewtopic.php?f=6&t=26741&p=205387

gizmocuz:
Could you contact me on the forum with some login credentials ? Or post the XML here please

p-hagen:
I am about to push a fix for this

gizmocuz:
Thanks for the PR, does this solve the issue ?

Ectholian:
I am unable to test it myself since i dont have a linux development environment set up and i am using a jadahl dsm package from the stable line. I will ask in the forum topic for someone to test.

gizmocuz:
@Ectholian , if you own a windows version, you could download/install/test the windows version

p-hagen:
It resolves the issue . I have tested it from a Mac and raspberry and in has been reported in the forum by an other user that it solved it for them too

gizmocuz:
Great, Thanks for the feedback and PR!

Ectholian:
Thanks guys!

p-hagen:
np

ivooovi:
I am using for many years the Anna thermostat with Domoticz but with the upgrade to the latest beta I get now invalid data received in my logs. It is not working. My Anna is on firmware 1.8.20.

Can you please assist

WilliamRoxit:
> I am using for many years the Anna thermostat with Domoticz but with the upgrade to the latest beta I get now invalid data received in my logs. It is not working. My Anna is on firmware 1.8.20.
> 
> Can you please assist

This issue is fixed in the development branch. So you need to install one of the beta's, version 4.10406 or newer. For support it is better to visit the forums: http://www.domoticz.com/forum/viewtopic.php?f=6&t=26741

ivooovi:
> > I am using for many years the Anna thermostat with Domoticz but with the upgrade to the latest beta I get now invalid data received in my logs. It is not working. My Anna is on firmware 1.8.20.
> > Can you please assist
> 
[...]

PR #3008, opened 2019-02-02
Fix for Anna thermostat firmware 3.1.4 XML change (Issue #3006)

p-hagen:
Fix is handling  CR/LF issues and Setpoint no longer returning data


gizmocuz:
Thanks!!

p-hagen:
np!

PR #3014, opened 2019-02-03
Devices page update

stas-demydiuk:
- Ability to filter data by Usage, Hardware, Device Type, Room
- Correct icons for different devices (especially light/switches)
- Display battery level in corresponding battery icon
- Replace sync calls with async ones
- Fix different chrome warnings
- Fix layout issues

image

image

Any feedback and suggestions are highly appreciated!


dannybloe:
Nice. When you change one of the checkboxes, does it only refresh the table? (I hope it does so you don't have to scroll down after each tick)
And it would be nice you could collapse the main filter groups (and remember the state).
Nice work.

stas-demydiuk:
@dannybloe 

> When you change one of the checkboxes, does it only refresh the table?

It loads all available devices when you open the page and then hide ones which not fit the filter when you change checkboxes

> And it would be nice you could collapse the main filter groups (and remember the state).

Good idea!

gizmocuz:
I like all the changes, except one... The part on the left, the checkboxes/device types, making my table small.

We (I) need to have a view where we can see as much as possible, as it is now.

There is a search box on the top right, if you type here Temp, you already get all temperature devices
It's very easy already to find anything you want using the search box

I have absolutely no problem at all with the left part... that is... if you can make a floating button that the user can click/pin to show/hide it.
Is this possible ? Maybe something like:

https://bootstrapious.com/p/bootstrap-sidebar

http://jsfiddle.net/hThGb/1/

https://www.w3schools.com/howto/tryit.asp?filename=tryhow_js_sidenav_push_opacity

Maybe you can store the state(s) in the HTML5 Web Storage ?

stas-demydiuk:
@gizmocuz This left panel is collapsed by default to save space for table and you can show it using button near the table (like in events editor). Or you've meant something else?

gizmocuz:
@stas-demydiuk , thats yes exactly what i meant... did not noticed it ... maybe the color of the button should be slightly more brighter/shiny
The 'used' option i would make with radio buttons 'All, No, Yes'

But everything you have in the sidebar can already be found using the 'search box' and many more

gizmocuz:
The 'used' maybe keep them always on top, so you can make your selections at the left (it will scroll up/down if you have 40 hardware/rooms ?).
This way you can quickly toggle between all/used/not used.
(Basically keep the 3 buttons we had like before)
Also because some would not use this bar on the left (because of the search box...) users still need to be able to toggle between the three options (without toggling the bar visible)

gizmocuz:
Maybe instead of the toggle bar/button on the left, do this with a (small) icon in the top bar (left of 'used')
This way it does not take any space

stas-demydiuk:
The idea was to implement something similar to filters in online stores. When there is no options selected in the group it equals the filter doesn't apply (the same as to select all options under the group). That's why there is no reason in "All" option.

For example if you want to select only used devices you need to select option "Yes" under "Used"section. If you want All you can do both: select Yes and No options or select nothing in this section.

gizmocuz:
Yes that's why I vote for a radiobutton solution instead of checkbox
And as the left bar is Not visible (per default) we need a top bar at the top like we have now

stas-demydiuk:
I will move "Used" filters out of the filters panel to the top (like it was before) if you insist, but for me its still not clear why do you think this type of filter will be used more often than others. From my personal experience I almost haven't used it but always miss option to filter or group devices for example by Hardware. 
And another my point, again just personal, I would prefer all filters in one place than 2 different panels: one at the top and another at the left.

gizmocuz:
@stas-demydiuk
For me, i like/need to have a view that is as much wide as possible
I also will not use the filter box on the left
I will use the search box for 'all' filtering
I need to switch between used/not used and all like before

So to keep the behavior exactly like it is now (which is good!), we could add a small icon in the top bar to hide/show the new filter box
This button also needs to have a color if any filters are set/stored

freemann:
Maybe a "compromise" could be to hide columns in the devices table?
Some of them are in 90% of the cases totally useless such as signal, subtype, unit. The take a lot of space and by hiding columns, you get more space for data and the left filterpanel?!

It would also be a great feature to hide colums :)

stas-demydiuk:
@gizmocuz I've restored the original panel at the top

image

gizmocuz:
Looking great!😁

gizmocuz:
Thanks!

salopette:
Stas since the PR is a mistake in the list!

https://github.com/domoticz/domoticz/issues/3052

PR #3022, opened 2019-02-06
Make Anne Thermostaat work with both old and new XML file format

p-hagen:
Patch will make allow the plugin  to read both version of XML input/output created by the two version of the ANNA (smile gateway) that seem to exist
When Domoticz starts or when new hardware is added the plugin will attempt to read the info of the gateway, based on the result it  will set a static flag that will be used to  handle the input out put for the two version

Tested against firmware 3.1.4  from Domoticz running on mac and on raspberry pi 3
No test has been done against older firmware as no device was available, however on forum volunteers are willing to test is as soon as it s merged in the beta build

### DIFF
--- a/hardware/AnnaThermostat.cpp
+++ b/hardware/AnnaThermostat.cpp
@@ -16,12 +16,15 @@
 // Plugwise Anna Thermostat
 
 const std::string ANNA_GET_STATUS = "/core/appliances";
+const std::string ANNA_SET_STATUS = "/core/appliances";
 
 #ifdef _DEBUG
-	//#define DEBUG_AnnaThermostat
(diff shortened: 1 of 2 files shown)


jgaalen (commented):
It's a whole lot for fix the parsing. This fix does work too on both versions:
Remove the replacing/removing of new lines, that is not necessary. Change the parse part with the line below and it works on both versions / xml's.

if (doc.Parse(sResult.c_str(), 0, TIXML_ENCODING_UTF8) && doc.Error())

gizmocuz:
Please only add the version check and leave all formatting/debug code as is...

p-hagen:
I’ll close this as it seems there is a other patch

PR #3024, opened 2019-02-06
Fixed Anna to work with both firmwares

jgaalen:
Fixed Anna to work with both firmwares
Added different boiler state switches
Improved log messages

### DIFF
--- a/hardware/AnnaThermostat.cpp
+++ b/hardware/AnnaThermostat.cpp
@@ -420,27 +420,11 @@ void CAnnaThermostat::GetMeterDetails()
                 {
                     if(strcmp(tmpstr.c_str(), "on") == 0)
                     {
-                        SendSwitch(8, 1, 255, true, 0, sname);
+                        SendTextSensor(8, 1, 255, "On",sname);
                     }
                     else
                     {
-                        SendSwitch(8, 1, 255, false, 0, sname);
-                    }
-
-                }
-            }
-            else if (sname == "domestic_hot_water_state")
-            {
-                tmpstr = GetPeriodMeasurement(pElem);
-                if (!tmpstr.empty())
-                {
-                    if(strcmp(tmpstr.c_str(), "on") == 0)
-                    {
-                        SendSwitch(9, 1, 255, true, 0, sname);
-                    }
-                    else
-                    {
-                        SendSwitch(9, 1, 255, false, 0, sname);
+                        SendTextSensor(8, 1, 255, "Off",sname);
                     }
 
                 }
@@ -452,11 +436,11 @@ void CAnnaThermostat::GetMeterDetails()
                 {
                     if(strcmp(tmpstr.c_str(), "on") == 0)
                     {
-                        SendSwitch(10, 1, 255, true, 0, sname);
+                        SendTextSensor(9, 1, 255, "On",sname);;
                     }
                     else
                     {
-                        SendSwitch(10, 1, 255, false, 0, sname);
+                        SendTextSensor(9, 1, 255, "Off",sname);;
                     }
 
                 }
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thanks for the PR
Is it possible to let every log message start with "AnnaTherm: " ?

jgaalen:
There you go

p-hagen:
Tested the code for 3.1.4
Result: Fail
Reason: The new switches do not update Domoticz as the code to implement switches other then the setpoint was never completed in the Anna plugin.
Recommendation : Change them to sensors for now  as that will allow this patch to function

New Switches:     Fail
Delete hardware : Ok
Add hardware :     Ok
Delete devices:     Ok
Auto add devices: Ok
Disable/Enable:     Ok
Set Setpoint:         Ok


Result: The new switches do not update as the code to implement switches other then the setpoint was never completed in the Anna plugin.

jgaalen:
It seemed that 'domestic_hot_water_state' doesn't update, its a fixed setting in the Anna Therm. But boiler_state updates (on and off) at least at 1.8.20 once the boiler starts heating. Perhaps we only need this switch. flame_state seemed to be a new state for 3.1.x version (not available on the 1.8.20).

Can you test the switch for 'boiler_state' only for 3.1.4? Then we can leave that switch in there. It works on my end for 1.8.20 and with a static xml file of 3.1.4

p-hagen:
@Jgaalen  I've tested the boiler state  and it gets an update.
 
However it is implemented as a switch that the user could update. this causes an error messagebox to appear if they click on the switch : Problem sending switch command 
using a text sensor for example is a quick solution:
` 
else if (sname == "flame_state")
            {

                tmpstr = GetPeriodMeasurement(pElem);
                if (!tmpstr.empty())
                {
                    if(strcmp(tmpstr.c_str(), "on") == 0)
                    {
			SendTextSensor(10, 1, 255, "On",sname);
                    }
                    else
                    {
                        SendTextSensor(10, 1, 255, "Off", sname);
                    }
                }
            }
`

jgaalen:
@p-hagen Thanks for your response. I see what you mean now, and yes the switch shouldn't be user-controllable, so text sensor seems the way to go.
I've updated the fix and also removed the domestic_hot_water_state as thats a config from Anna, not a state that switches when you turn on/off the hot water.

p-hagen:
Tested on 3.1.4 and no issues found.  Ok to merge

gizmocuz:
Thanks!!

jgaalen:
I'm not happy by how the text sensor works for the state. It always adds a new line 'On' or 'Off' and overwriting the time. Therefor you can't see for how long the state 'On' is on that state for which time. We need a simple switch that cannot by controlled from the interface. Not sure how to do that, but there are many switches like that, like motion sensor, door switch, those are switches which cannot be controlled from the interface. Any idea how to change the text sensor to an actual switch?

gizmocuz:
like a 'contact' type of switch ?

gizmocuz:
Instead of a text sensor, the code needs to change to use a general switch for On/Off

jgaalen:
@gizmocuz That's what it was before `SendSwitch` , but that seems to be a user controllable switch from the user interface, which should be possible. Any function available to set a switch on/off which can only be controlled from the code?

jgaalen:
I've thought about it and think the `SendSwitch` command is the better way to go compared to `TextSensor`.  It's a on/off switch, which users can't control. If they somehow try to turn on the boiler by using the switch, they get an error and that is actually good, because they cannot switch it which is nicely shown in the interface. The TextSensor is a worse implementation in my opinion. 
What do you think?

gizmocuz:
Yes better use the switch. Users can set it to a contact type

ivooovi:
I have noticed one strange behaviour with Anna 1.8.20 I get battery low on my return water temperature from Anna. 

See this notification:
status problems 
2019-02-08 18:03:03.040  Status: dzVents: Error (2.4.13): Low battery warning: device temp - water (retour) low battery level : 56

PR #3028, opened 2019-02-07
add latitude / longitude and test for (local) Allowed networks without password

rwaaren:


gizmocuz:
Why would you need the 'allowed networks' to be available in a script ? I think this is a security issue

rwaaren:
> Why would you need the 'allowed networks' to be available in a script ? I think this is a security issue

It is not so much that it must be available in an end user script but by passing it, it can be used in an additional check to ensure that dzVents functions needing this, will work. It's along the same lines as the check that latitude / longitude are set to ensure sunset / sunrise and the likes can be calculated correctly.
On the forum I have seen a number of posts complaining that certain dzVents functions do not work, where the solution was to add local (127.0.0.1) in the 'allowed network' setting, despite the fact it's clearly stated in the documentation in the wiki.  
 
If you think it presents a security risk I will take it out but I don't see it as an additional risk as the user has set the 'allowed network' him-/ herself.

gizmocuz:
Maybe the c++ code could be changed to always allow dzVents to run 'locally' ?

rwaaren:
Thx for your comments but I don't want to go that way. For one thing because it's beyond my current c++ skills and knowledge of the domoticz internals and I guess (and please correct me if I am wrong) it would be a kind of backdoor approach.
I will close this PR and will come later today or this weekend with a new one without passing the 'Allowed local network' setting. I will try to get the check and message inside dzVents.cpp or pass a (kind of) true / false to dzVents framework.

PR #3030, opened 2019-02-08
Add latitude / longitude and test for (local) Allowed networks without password

rwaaren:


gizmocuz:
I don't think we need to expose 'localNetworksAllowed' to plugins.
If DzVents needs to have localnetwork access (always allowed to execute) this might be done in the c++ code ?
But the reason why this is needed is not clear to me, unless the system is calling some funky http code.
In the later case, it might be best to add some functionality so this won't be needed?

rwaaren:
For some commands dzVents do call the domoticz API. Only reason to  expose the boolean localNetworksAllowed to dzVents is tp enable dzVents Lua code to present user with a meaningful message before the call to the API is actually made. 
If you think it is a security issue I will take it out and move the check and message to dzVents.cpp

rwaaren:
Will send a new one without allowedNetworks passed to dzVents Lua code

dannybloe:
The thing is that not every http-API command is available through the commandArray. So in an ideal situation we extend the commandArray handler with the missing commands but I don't have the skills to do so. So instead, we rely on the http API to still be able to issue those commands. After all, it would be strange that you can issue the command through the GUI but not from a script.

But, in order to do so, the scripting engine needs access to the local API address but that must be enabled in the settings.

gizmocuz:
Yes ok, as long as you do not expose the lat/long and local network 'settings' to the end-user.
You ONLY get those (via the web page/json) if you have admin rights
(So if your not in the local network, then you also do not have admin rights)

PR #3040, opened 2019-02-12
Updated Anna Thermostat Added ability to select presets on the Anna  via selector switch. Added ability to update  the state of the proximity switch in the Anna thermostat.  

p-hagen:


p-hagen:
Tested against firmware 3.1.4. openend the plugin to be used with the Plugwise Adam as well.

gizmocuz:
Thank you for the great changes !
Could you set your editor to use 'tabs' as idents, it looks it is configured for 4 spaces (like line 290, or 3? like line 250)
Don't know what editor you are using, but if you use codelite try:
https://stackoverflow.com/questions/12711230/codelite-formatting

Could you correct the idents ?

PR #3054, opened 2019-02-19
dzVents bugfix for RGBWW Set Color state and inverted logic in dzVents.cpp

rwaaren:
modified:   dzVents/runtime/device-adapters/Adapters.lua
modified:   main/dzVents.cpp


gizmocuz:
Thanks!

PR #3055, opened 2019-02-19
Fix file paths for z-wave topology window

stas-demydiuk:
Fix the error when z-wave topology window can't load particular scripts


gizmocuz:
Thanks!!

PR #3062, opened 2019-02-22
Show counter value instead of usage (undefined) on dashboard

guillaumezin:
for managed counters, should correct https://github.com/guillaumezin/DomoticzLinky/issues/7


gizmocuz:
Thanks!!

PR #3077, opened 2019-02-27
Update dzVents to 2.4.15 

rwaaren:
Add option to use camera-name in snapshot command.
Domoticz- and dzVents version string available 
Some acceptance tests adjusted
Some acceptance tests added


rwaaren:
Need extra test

PR #3082, opened 2019-02-28
Detect Python boost library

Infern1:
Boost Python library wasn't detected. Based on the example from here:
https://github.com/TNG/boost-python-examples
more exactly
https://github.com/TNG/boost-python-examples/blob/master/CMakeLists.txt


gizmocuz:
Thank you for the PR !!
As far as i know, there is no dependency on boost_python needed, are you sure about this ?
This because all build scripts also do not 'build' the boost-python library, and python is still working

Infern1:
> As far as i know, there is no dependency on boost_python needed, are you sure about this ?
> This because all build scripts also do not 'build' the boost-python library, and python is still working

Yes python is working, since this is detected elsewhere.
The reason for this PR is that in the [install manual ](https://www.domoticz.com/wiki/Mac_OSX) for MacOS is mentioned:
```
brew install cmake
brew install boost --with-python
brew install boost-python
.....
```

So when I need to install boost-python I suppose it need to be detected by cmake. So this is why I made the PR. If it is really need for functionality I don't know.

gizmocuz:
Could you have a look at the Travis build script here:

https://github.com/domoticz/domoticz/blob/development/.travis.yml

I do not see a reference to it, maybe it is not needed (anymore)

gizmocuz:
Then again... the travis OSX build is not working...

gizmocuz:
No response for a long time.. maybe the setup page for osx on the wiki needs updates?
Closing issue for now

PR #3086, opened 2019-03-01
Correct display log for device type: Soil Moisture

blizniukp:


gizmocuz:
Moisture is not the same as Humidity, it is another type of sensor, with different meaning

```
const char *Get_Moisture_Desc(const int moisture)
{
	if (moisture < 10)
		return "saturated";
	else if (moisture < 20)
		return "adequately wet";
	else if (moisture < 60)
		return "irrigation advise";
	else if (moisture < 100)
		return "irrigation";
	else
		return "dangerously dry";
}
```

When you add a dummy moisture sensor you will see how it appears

You need to add a new device type for this.

Olifant1990:
Doesn't an "irrigation advise" also depends on the kind of plant placed in the soil?
I have plants which love dry soil, but I have also plants which love wet soil.

gizmocuz:
@blizniukp, if you have some (more) free time, please make a new PR with the needed changes

PR #3091, opened 2019-03-04
Proxycereal

rkluwen:
Proxycereal revisited.
Can someone that had troubles with the previous patch please test this first?


rkluwen:
This PR still has issues with inter-domoticz connections via MyDomoticz.

rkluwen:
Is there anybody who was able to review this PR?
Thanks,

Rene

gizmocuz:
@rkluwen , i think checking out a PR and building is a bit to much asked...
Maybe we should merge ?

rkluwen:
I did something but I'm not sure if I merged or not.
Either way, as far as I'm concerned, this PR can be applied now.

rkluwen:
Some checks turn out unsuccessful. But I think those are not related to the proxycereal patch.
I tested positively on the Pi, Ubuntu & Windows.

rkluwen:
Apolgies. It did have a merge conflict that I didn't see. Now it doesn't anymore.

gizmocuz:
Thanks for the updates ! One question, do we need the changes in 'domoticz.vcxproj' or can they be reverted ? I would exprect the version number to go up (VS2019 161.0 P2), but there is also a include dir change

rkluwen:
I made a change to the vs142 toolset. Is this better?

gizmocuz:
Thank you for the change... it's mainly this part ' <IncludePath>$(IntDir);$(IncludePath)</IncludePath>'  ($(IntDir); addon)

rkluwen:
Appveyor doesn't seem to like v142. Even though this is the one used in the development branch?

gizmocuz:
Yes that happens.. if you revert this file it would be fine.. revert as in the current development branch

rkluwen:
I'm afraid you will have to do that part. I upgraded my VS already completely.

gizmocuz:
I reverted the include path and the version number change to VS2017.
Still i read that AppVeyor should support VS2019... Anyway, for now i take care not to submit changes related to this ;)

gizmocuz:
Question, is it possible to change the boost::mutex to std::mutex ? This is not used in domoticz anymore

I also removed the boost lexical dependency, trying to go to cxx11

rkluwen:
Change from boost::mutex => std::mutex: consider that done in the latest commit.
Note: I needed to do a make clean in visual studio in order to avoid a segfault after applying the patch.

gizmocuz:
Thanks !!

PR #3096, opened 2019-03-05
Create nas-wr01ze.xml

Beddie:
This file is already part of/mapped in manufacturer xml but the file itself is not copied from openzwave. Hereby the file+content

Manufacturer xml: Product type="0200" id="1027" name="Power plug 12A" config="shenzen_neo/nas-wr01ze.xml"

### DIFF
--- a/Config/shenzen_neo/nas-wr01ze.xml
+++ b/Config/shenzen_neo/nas-wr01ze.xml
@@ -1,0 +1,89 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Product xmlns='http://code.google.com/p/open-zwave/'>
+  <!--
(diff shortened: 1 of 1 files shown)


gizmocuz:
@Beddie , thank you for the PR
As this is for the openzwave library, you need to make a pull request to their repository so it will be included in their repo
We poll from time to time their repository to update the configuration files

Beddie:
@gizmocuz The file is already part of the openzwave repository and yesterday you did update the configuration files, except this file was left out.

https://github.com/OpenZWave/open-zwave/blob/master/config/shenzen_neo/nas-wr01ze.xml

gizmocuz:
@Beddie , indeed... very strange it was not included...
Should be included now

PR #3100, opened 2019-03-06
POST request fix where content-length header equals 0.

rkluwen:


rkluwen:
An http request was not parsed properly when a post request was issued and a content length of 0 was posted.

gizmocuz:
Thanks!

PR #3102, opened 2019-03-07
Missing SET in the Update statement

pipiche38:
In the UPDATE clause , the SET statement was missing to get the UPDATE correctly xecuted


pipiche38:
@dnpwwo for your info

Issue #3103, opened 2019-03-08
Devices: Last Seen sort issue in Safari


HansVanEijsden:
Thanks for the new devices list. I like it better!
Unfortunately I had to install Chrome to use it, because in Safari the list doesn't get sorted on date, and also if I click on the Last Seen sort column (upper right, "Laatst gezien" in Dutch) the list doesn't sort by date.
There are no errors visible in the Safari console.

Safari 12.0.3 (14606.4.5), MacOS Mojave 10.14.3 (18D109). No Safari plugins enabled. Language nl_NL.
Domoticz Version: 4.10547, Build Hash: a76a0fdb, Compile Date: 2019-03-08 03:04:06.

![schermafbeelding 2019-03-08 om 16 59 03](https://user-images.githubusercontent.com/11890145/54039562-8bb78880-41c3-11e9-9e38-3ed64f218e78.png)

Let me know if you need more info. Thank you!

stas-demydiuk:
@HansVanEijsden the issue should be fixed in latest Domoticz beta

HansVanEijsden:
@stas-demydiuk confirmed working again. Thank you very much for the fix!

Issue #3093, opened 2019-03-05
With the new Device List - if the Hardware is disable you still see the devices 


pipiche38:
In the V4.9700, when you go in the Delive List and the Hardware is disabled, you don't any of the related devices.

Here with the new version, ALL devices are shown despite the fact that some Hardware could be disabled !

gizmocuz:
@stas-demydiuk , could you have a look at this ? there is a option 'displaydisabled' that should be set to false

stas-demydiuk:
@pipiche38 Are you sure it was working differently with old Device List, as I use exactly the same request to get all devices:
```
/json.htm?type=devices&displayhidden=1&displaydisabled=1&filter=all&used=all
```
I guess `displaydisabled` param affects will disabled devices be shown or not but it was hardcoded to `1` even before my update

pipiche38:
May be it is not related to your changes, but for me it was indeed such behaviour, Hardware disabled, no devices shown.

pipiche38:
thanks

Issue #3084, opened 2019-03-01
WIth the new List of Devices on beta there is no device list refresh


pipiche38:
Beta 4.10531

Since the new Device List layout (which is great), there is no refresh of such list, when going from NOt Used, to Used or to All of Devices.
It was really convenient before, when having new devices created by a Python Framework plugin !

So I think when moving from one tab, to the other, or when simply re-asking Hardware-Devices (even if we are on such screen) a full refresh of the Device List is a needed

gizmocuz:
I agree, that was the same behavior as before
@stas-demydiuk , is this something to add ?

pipiche38:
agree. I think the Refresh button is a great enhacement

PR #3105, opened 2019-03-09
Devices page improvements

stas-demydiuk:
#3084 Devices: Ability to manually refresh device list
#3093 Devices: Hide disabled devices if such option is enabled in settings
#3103 Devices: Fix sorting by Date in Safari

Added refresh button at the top of the table
image


gizmocuz:
Thanks !!

PR #3106, opened 2019-03-09
Serial port not detected

Kebims:
ttyS aren't detected because it's just ttySAC before.

### DIFF
--- a/domoticz.sh
+++ b/domoticz.sh
@@ -14,7 +14,7 @@
 PATH=/sbin:/bin:/usr/sbin:/usr/bin:/usr/local/bin
 DESC="Domoticz Home Automation System"
 NAME=domoticz
-USERNAME=dietpi
+USERNAME=pi
 PIDFILE=/var/run/$NAME.pid
 SCRIPTNAME=/etc/init.d/$NAME
 
--- a/domoticz.sh
+++ b/domoticz.sh
@@ -14,7 +14,7 @@
 PATH=/sbin:/bin:/usr/sbin:/usr/bin:/usr/local/bin
 DESC="Domoticz Home Automation System"
 NAME=domoticz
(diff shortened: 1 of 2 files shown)


gizmocuz:
Please check our wiki about persistent devices

Kebims:
> 
> 
> Please check our wiki about persistent devices

Sure, but depending on the system, the serial ports are ttySx or ttySACx so if we do a search with ttyS *, it works for both without persistent devices, isn't it?

gizmocuz:
I have a system where i see 20 or more ttyS*** devices which are not real and connected
What type of hardware are you using ?
I don't like to display non-connected serial ports...

Kebims:
I see what you are talking about.
I use a NanoPi Neo Air, with ttyS[0-7]. Only ttyS[0-2] are accessible on GPIO port and ttyS[4-7] are not used.
I will try to use persistent device for ttyS1->ttySAC1 and ttyS2->ttySAC2 ( ttyS0 is for console ).

Kebims:
Hi,
On my NanoPi Neo Air, I was created a file /etc/udev/rules.d/90-local.rules with :
`KERNEL=="ttyS[0-2]", SYMLINK+="ttySAC%n"`
In hardware my ttySAC* are display correctly.

gizmocuz:
How about you symlink it to ttyUSBMyDevice ? Then we do not need this patch ?

Kebims:
The change only affects UART serial ports. On NanoPi, the devices name are ttyS [0-7] and the symlink allows you to create an alias to become ttySAC *. By making this change, the patch is no longer needed. You can add the trick on a wiki if you wish.

gizmocuz:
We have already a document on our wiki explaining how to deal with persistent USB devices 
Thanks for the feedback

PR #3110, opened 2019-03-11
Fix for auto update script when running as non-root user

mpiederiet:
Fixed update script to also work when running as systemd service unit as non-privileged user (but with sudo rights).


gizmocuz:
Thanks!!

PR #3116, opened 2019-03-15
Fix Evohome heat demand bug and add SYNC command

Evsdd:
Hi,  
A small update to fix a couple of bugs in EvohomeRadio which affected heat demand logging.  Also added the 0x1F09 (SYNC) to the list of decoded commands.
Dan


gizmocuz:
Thanks !!

bramvreugd:
The new sync causes Domoticz to crash within a few minutes with a number of people on the forum.  
[Item on forum](https://www.domoticz.com/forum/viewtopic.php?f=6&t=27484)

PR #3125, opened 2019-03-24
Fixed rain device for DarkSky

Olifant1990:
Fixed rain device for DarkSky.
DarkSky only sends a rate/hour. So you have to build a total yourself.


Olifant1990:
Note: I had to add a new rain subdevice type. Because of the different way of calculations.

Olifant1990:
Should wait to have some real rain to test, before pulling. Sorry!

Issue #3109, opened 2019-03-10
Issue with rain graph from DarkSky


Olifant1990:
Since WU is not working anymore since they closed the free API, I switched to DarkSky.
Nice service!
But, the week and month graphs are incorrect.
See screenshot. It rained a lot today, but the week graph is hangen on 0 mm...
Maybe something to do with issue https://github.com/domoticz/domoticz/issues/2736?

For the today week state, I see the row SELECT Total, Total, Rate FROM Rain WHERE (DeviceRowID=%" PRIu64 " AND Date>='%q') ORDER BY ROWID DESC LIMIT 1. (WebServer.cpp line 14551) So I can imagine that this will stay on 0 forever due to the order by rowid desc and limit 1. Doesn't have this row to do some calculations or something?
![afbeelding](https://user-images.githubusercontent.com/25199588/54090973-9e27f280-437a-11e9-81a8-c27793d25a23.png)

Olifant1990:
I am close to a solution. Just need some tuning and testing. Hang on!
It was needed to rebuild the calculation logic, because you only get a rainRate from DarkSky, no rain totals.
Herefore I needed to make a new subtype for DarkSky.

gizmocuz:
@Olifant1990 , i think we get a daily rain, not a total rain.
The rain-rate is something different, this is the amount of rain in the last hour
But maybe a more generic sensor name for the new rain sensor is better ? It could be used by other hardware as well

Olifant1990:
No, we get a precipIntensity only from DarkSky, which is described as "The intensity (in inches of liquid water per hour) of precipitation occurring at the given time. This value is conditional on probability (that is, assuming any precipitation occurs at all)."
So we get only for example a rainrate of 0.5 mm/hour at current time.
Then we need to calculate this and get a total for our selfs over the day.
All this is working at my testversion right now and seems ok, but need some real rain, it is dry for a few days here or I wasn't home those rainy days :P
I will give it a more generic name. Like sTypeRAINRate?
Have sneak preview at https://github.com/domoticz/domoticz/pull/3125
I closed it because a lack of testing

gizmocuz:
Sounds very bad, and not good at all ... Sure the rain rate will be adjusted during an hour... you can not use this for a counter
01:00-02:00 5mm rain, started at 01:15
01:20- 3 mm rain.. the 2mm from 01:15 is already taken out
It keeps on changing

As this is not reliable, i would suggest to drop support

Bdw, WU is working again

[...]

PR #3131, opened 2019-03-27
Fixed rain device for DarkSky

Olifant1990:
Fixed rain device for DarkSky.
DarkSky only sends a rate/hour. So you have to build a total yourself.

See https://github.com/domoticz/domoticz/issues/3109


gizmocuz:
Thanks for the patch !!
Could you make a SendRainRateSensor function ? (Instead of doing it directly in the DarkSky class)
This way we might be able to use it in other feature hardware classes that need the same type of rain calculation

Olifant1990:
The calculations are done in mainworker.cpp ;)
So not depending on Darksky already, right?

gizmocuz:
Yes, but the object creation is done in DarkSky, that should be moved to DomoticzHardware.cpp/h

RBUF tsen;
memset(&tsen,0,sizeof(RBUF));
...

gizmocuz:
Thanks!!

Olifant1990:
You were fast, just committed it xD

PR #3134, opened 2019-04-01
Fixed: Evohome SYNC bug

Evsdd:
This fixes a bug which was introduced in the recent update to EvohomeRadio to add the SYNC command which causes Domoticz to crash. 

See discussion of bug here:

https://www.domoticz.com/forum/viewtopic.php?f=6&t=27484

Thanks,

Dan


gizmocuz:
Thanks!!

PR #3139, opened 2019-04-04
Added Evohome DATE_TIME message decoding

Evsdd:
Hi,
A small update to add 0x313F (DATE_TIME) to the list of decoded Evohome messages.
Dan

### DIFF
--- a/hardware/EvohomeRadio.cpp
+++ b/hardware/EvohomeRadio.cpp
@@ -47,6 +47,7 @@ enum evoCommands
 	cmdDeviceInfo = 0x0418,
 	cmdBatteryInfo = 0x1060,
 	cmdSync = 0x1F09,
+	cmdDTime = 0x313F,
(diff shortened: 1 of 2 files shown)


gizmocuz:
Thank you for this PR.
It seems the new function 'DecodeDTime' is not used and currently only outputs some logging ?

Evsdd:
Hi,

Yes, this is a really small update to simply add the decoding of this command to the log.  I'm gradually expanding the list of Evohome commands that are successfully decoded (to reduce/remove the UNKNOWN COMMAND flagging) and this helps with Evohome system monitoring and troubleshooting by identifying true corrupt messages caused by weak signal due to device positioning.

Let me know if would be better to group the updates to reduce the number of pull requests rather than submit small updates such as this one.

Dan

gizmocuz:
@Evsdd , it might be better to submit a PR when the code is being called/implemented.
I understand you are working on this (Great!!), but currently the new function is not called, so there are no new changes for domoticz (for the end-user)

Evsdd:
No problem.  I'll close this pull request and re-submit in a few weeks when I can combine it with additional functionality using the additional decoded commands.

Regards,

Dan

Issue #2546, opened 2018-07-08
Use of compiler security flags


Avamander:
It would be nice if domoticz could be compiled with all the security features compilers currently provide. This could be done incrementally.

Small collection of flags domoticz could be compiled with, with descriptions (thanks to stackoverflow):

```
-Wall -Wextra 
Turn on all warnings to help ensure the underlying code is secure.

-Wconversion -Wsign-conversion
Warn on unsign/sign conversion.

-Wformat­-security
Warn about uses of format functions that represent possible security problems.

-Werror 
Turns all warnings into errors.

-arch x86_64 
Compile for 64-bit to take max advantage of address space (important for ASLR; more virtual address space to chose from when randomising layout).

-mmitigate-rop 
Attempt to compile code without unintended return addresses, making ROP just a little harder.

-fstack-protector-all -Wstack-protector --param ssp-buffer-size=4 
Your choice of "-fstack-protector" does not protect all functions (see comments). You need -fstack-protector-all to guarantee guards are applied to all functions, although this will likely incur a performance penalty. Consider -fstack-protector-strong as a middle ground. 
The -Wstack-protector flag here gives warnings for any functions that aren't going to get protected.

-pie -fPIE 
For ASLR.

-ftrapv 
Generates traps for signed overflow (currently bugged in gcc, and may interfere with UBSAN).

-­D_FORTIFY_SOURCE=2 -­O2
Buffer overflow checks. See also difference between =2 and =1.

­-Wl,-z,relro,-z,now
RELRO (read-only relocation). The options relro & now specified together are known as "Full RELRO". You can specify "Partial RELRO" by omitting the now flag. RELRO marks various ELF memory sections read­only (E.g. the GOT).

-fsanitize=float-divide-by-zero
detect floating-point division by zero

-fsanitize=float-cast-overflow
check that the result of floating-point type to integer conversions do not overflow;

-fsanitize=bounds
enable instrumentation of array bounds and detect out-of-bounds accesses;

-fsanitize=alignment
enable alignment checking, detect various misaligned objects;

-fsanitize=object-size
enable object size checking, detect various out-of-bounds accesses.

-fsanitize=vptr
enable checking of C++ member function calls, member accesses and some conversions between pointers to base and derived classes, detect if the referenced object does not have the correct dynamic type.
```

[...]

PR #3150, opened 2019-04-15
Bunch of code fixes

dextercd:
Fixes a bunch of code that's obviously incorrect. These were found by enabling -Wall and looking at the compiler output (see #2546). It'd by nice if more warnings were fixed so they can be enabled by default to avoid introducing detectable bugs.

In commit 28c9207 I had to choose between changing the indentation or introducing braces, I chose to change the indentation because that preserves the current meaning of the code, and seems to be how the code was introduced in 1d1d5404b (function Assegna_chi_cosa_dove_quando).

Sorry about basing commits on the development branch instead of master. I didn't look at the contributing guidelines before writing these fixes, and rebasing them causes a bunch of conflicts.

### DIFF
--- a/main/SunRiseSet.cpp
+++ b/main/SunRiseSet.cpp
@@ -154,7 +154,7 @@ bool SunRiseSet::GetSunRiseSet(const double latit, const double longit, const in
 
 	rise = UtcToLocal(rise, timezone);
 	set = UtcToLocal(set, timezone);
-	result.SunAtSouthMin = static_cast<int>(modf((rise+set)/2.0, &_tmpH)*60+0,5);
+	result.SunAtSouthMin = static_cast<int>(modf((rise+set)/2.0, &_tmpH)*60+0.5);
 	result.SunAtSouthHour = static_cast<int>(_tmpH);
 
 	switch(rs) {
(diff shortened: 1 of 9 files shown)


main/SunRiseSet.cpp
```diff
@@ -154,7 +154,7 @@ bool SunRiseSet::GetSunRiseSet(const double latit, const double longit, const in
 
 	rise = UtcToLocal(rise, timezone);
 	set = UtcToLocal(set, timezone);
-	result.SunAtSouthMin = static_cast<int>(modf((rise+set)/2.0, &_tmpH)*60+0,5);
+	result.SunAtSouthMin = static_cast<int>(modf((rise+set)/2.0, &_tmpH)*60+0.5);
```
dextercd, line 157: Should this get turned into a call to `std::round`? `static_cast<int>(x + 0.5)` doesn't round negative numbers in the same way as `std::round`.

gizmocuz: Hi, yes that makes more sense indeed!

gizmocuz: And thanks for looking at these 'warnings' (errors)

gizmocuz:
Thanks for this great PR! and looking into these issues. Is it possible to make another PR for the SunRiseSet::GetSunRiseSet?

PR #3153, opened 2019-04-18
Update setup.html

ressof:
Change log interval


gizmocuz:
Sorry, this has been intentionally left out as it can cause damage to usage if their database gets to full.
You can always set this value in the database directly.
For a patch, i would also suggest to use a dropdown and never an edit box option ;)

PR #3157, opened 2019-04-23
Use round instead of adding 0.5 and casting.

dextercd:
Per https://github.com/domoticz/domoticz/pull/3150#issuecomment-485749552.

### DIFF
--- a/main/SunRiseSet.cpp
+++ b/main/SunRiseSet.cpp
@@ -144,7 +144,7 @@ bool SunRiseSet::GetSunRiseSet(const double latit, const double longit, const in
 	//astrlen = day_astronomical_twilight_length(year,month,day,longit,latit);
 
 	double _tmpH;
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thanks! Is it possible that you could you add 'std::' before the 'round' call, so we know it is using the 'std' round function ?
(This because there are also other round functions defined in the software, for example see 1wire.cpp at the top)

dextercd:
I've made that change, but this file currently includes `<math.h>`, not `<cmath>`, so it's not guaranteed that round will be in the std namespace. (See https://en.cppreference.com/w/cpp/header#C_compatibility_headers).

It seems fine in libstdc++.

gizmocuz:
Thanks!!, i am going to do some test build on a windows, raspberry pi (32/64) and ubuntu and let you know if it's working

PR #3162, opened 2019-04-26
Arilux driver brightness level support and code cleanup

nutki:
This PR changes the following in the arilux hardware driver:
* RGBW controllers changed from `sTypeColor_RGB_W` to `sTypeColor_RGB_W_Z` so all 4 channels can be used independently
* `value` is no longer ignored in `Color_SetColor` RGB mode (this prevented dimming in most situations)
* removed some dead code
* simplified IP address manipulation


gizmocuz:
Thanks !!

PR #3163, opened 2019-04-27
Prevent touch scroll while using jQuery wheel color picker

nutki:
When using touch and slide on jQuery wheel color picker elements the page will scroll as well.
The is fixed here using the same method as in the component used for level slider:
https://github.com/jquery/jquery-ui/blob/master/themes/base/slider.css#L22


gizmocuz:
Thanks !!

PR #3164, opened 2019-04-27
Add support for evohome setpoint background colours

fullTalgoRythm:
Please can you add back support for setpoint colours for evohome temperature devices

### DIFF
--- a/www/app/TemperatureController.js
+++ b/www/app/TemperatureController.js
@@ -104,23 +104,6 @@ define(['app'], function (app) {
 			$("#dialog-editstate").i18n();
 			$("#dialog-editstate").dialog("open");
 		}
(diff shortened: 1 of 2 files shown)


gizmocuz:
There already seems to be a function 'EvoSetPointColor' in this class, maybe this function should be called ?

fullTalgoRythm:
That function is legacy code it's not compatible with the new code which atm requires a css style. The old function can be remove or left for reference as preferred. The colours were default behaviour until 4.9700 so the patch just restores original behaviour.

gizmocuz:
@fullTalgoRythm , thats for the explanation. Could you remove them in your patch ?

fullTalgoRythm:
Right that should be done now are we OK to merge?

gizmocuz:
Thanks !!

PR #3172, opened 2019-05-03
avoid running as root, actually use --chuid

### LINKED ISSUE #3145: domoticz.sh runs domoticz as root
The `domoticz.sh` script does not actually pass `--chuid` to `start-stop-daemon` because it is only added in the line testing whether the daemon is already running. See https://github.com/domoticz/domoticz/blob/ce68fa52ca292b5f784af47e45440196c272df1e/domoticz.sh#L60 versus https://github.com/domoticz/domoticz/blob/ce68fa52ca292b5f784af47e45440196c272df1e/domoticz.sh#L62

I wanted to make a pull request. The [instructions](https://github.com/domoticz/domoticz/blob/development/.github/CONTRIBUTING.md#fixing-bugs) say to base against the `master` branch, but the `development` branch seems to be the one actually used. Which one should I use?

m4lvin:
fix https://github.com/domoticz/domoticz/issues/3145

### DIFF
--- a/domoticz.sh
+++ b/domoticz.sh
@@ -59,7 +59,7 @@ do_start()
         #   2 if daemon could not be started
         start-stop-daemon --chuid $USERNAME --start --quiet --pidfile $PIDFILE --exec $DAEMON --test > /dev/null \
                 || return 1
-        start-stop-daemon --start --quiet --pidfile $PIDFILE --exec $DAEMON -- \
+        start-stop-daemon --chuid $USERNAME --start --quiet --pidfile $PIDFILE --exec $DAEMON -- \
                 $DAEMON_ARGS \
                 || return 2
 }


gizmocuz:
@m4lvin , Thanks for the patch!
I need to test this more, as some hardware on the raspberry pi needs root privileges and i wonder if they continue to work with this patch
(For example GPIO)

Is that something you could test ?

m4lvin:
I am running domoticz on a raspberrypi 3 as a non-root user. I do not use the GPIO pins for anything and cannot test this, but I remember that I added the domoticz user to the `dialout` group which has access to `/dev/ttyUSB0`. My guess would be that GPIO can also be accessed by non-root users if they are in the relevant groups?

gizmocuz:
I don't think that is the issue as it's not using TTY ... 
Would be great if someone can test this... Even if you do not have anything connected to the gpio you should be able to test it and check if there are errors in the log

m4lvin:
Okay, I had to do `adduser domoticz gpio` and `echo 4 > /sys/class/gpio/export` as mentioned [here](https://www.domoticz.com/wiki/GPIO). After that the "Generic sysfs GPIO" driver works fine on my pi 3.

rkluwen:
Just a question: Can we still bind to ports < 1024 on Linux?

m4lvin:
I think not directly, but searching for it leads to [several methods](https://serverfault.com/questions/268099/bind-to-ports-less-than-1024-without-root-access), of which iptables forwarding or `sudo setcap cap_net_bind_service=ep the-domoticz-binary` seem to be the ways to go.

In any case, the default config of domoticz runs on 8421, correct? Then this should not be a problem.

gizmocuz:
The default port for HTTP is 8080, and HTTPS is 443

m4lvin:
Oh, sorry, I probably changed this in my configuration then. So 8080 should work fine, and is SSL enabled or disabled by default?

gizmocuz:
It is enabled by default...

molusk:
Hi,
I created a systemd unit to run domoticz with an unprivileged user on ports 80 and 443 for my personal use if it can help in anyway, here it is : https://github.com/molusk/domoticz-scripts/tree/master/systemd
I didn't check gpio's were working fine with that config but rfx433, z-wave and zigate are.

gizmocuz:
I am closing this issue as it seems it can not be tested 100%

m4lvin:
Fine with me, but for the record: If the recommended/default setting is to run domoticz as root, then this should be documented. In particular I would then suggest to remove `USERNAME=pi` and all uses of it from `domoticz.sh`.

Issue #3169, opened 2019-05-02
Issue with posting data to influxdb through dzvnet


Kratax:
I have an issue with the dzvent script posting data to influxdb ending with error log (despite the fact that the data are properly posted):

Please see below...
```
2019-05-02 13:16:00.542 Status: dzVents: Info: ------ Start internal script: Post_influx_data:, trigger: every 1 minutes
2019-05-02 13:16:00.565 Status: dzVents: Info: Zapisuji do influx teplota venek
2019-05-02 13:16:00.565 Status: dzVents: Info: Zapisuji do influxdb teploty konec
2019-05-02 13:16:00.565 Status: dzVents: Info: TEMP Cels
2019-05-02 13:16:00.565 Status: dzVents: Info: 21.3
2019-05-02 13:16:00.565 Status: dzVents: Info: HUM
2019-05-02 13:16:00.565 Status: dzVents: Info: 46
2019-05-02 13:16:00.565 Status: dzVents: Info: TFA_sensor temp=21.3,hum=46
2019-05-02 13:16:00.565 Status: dzVents: Info: Zapsano do influxdb
2019-05-02 13:16:00.566 Status: dzVents: Info: ------ Finished Post_influx_data
2019-05-02 13:16:00.567 Status: EventSystem: Script event triggered: /home/pi/domoticz/dzVents/runtime/dzVents.lua
2019-05-02 13:16:00.709 Status: dzVents: Info: Handling httpResponse-events for: "mycallbackstring
2019-05-02 13:16:00.709 Status: dzVents: Info: ------ Start internal script: Post_influx_data: HTTPResponse: "mycallbackstring"
2019-05-02 13:16:00.715 Status: dzVents: !Info: response from influxDB:
2019-05-02 13:16:00.715 Status: dzVents: !Info: ASCII values of response from influxDB:
2019-05-02 13:16:00.715 Status: dzVents: !Info: return code from influxDB: 12103200
2019-05-02 13:16:00.715 Status: dzVents: !Info: Header Date ==>> Thu, 02 May 2019 11:16:00 GMT
2019-05-02 13:16:00.715 Status: dzVents: !Info: Header X-Request-Id ==>> aa7a98a9-6ccb-11e9-bef6-b827eb116147
2019-05-02 13:16:00.715 Status: dzVents: !Info: Header X-Influxdb-Version ==>> 1.7.6
[...]

PR #3176, opened 2019-05-04
Get HTTP statusCode from HTTP header

rwaaren:
#3169 	modified:   dzVents/runtime/HTTPResponse.lua
	modified:   main/dzVents.cpp

### DIFF
--- a/dzVents/runtime/HTTPResponse.lua
+++ b/dzVents/runtime/HTTPResponse.lua
@@ -4,18 +4,19 @@ local utils = require('Utils')
 local function HTTPResponce(domoticz, responseData)
 
     local self = {}
-
+     
     self.headers = responseData.headers or {}
     self.baseType = domoticz.BASETYPE_HTTP_RESPONSE
-
     self.data = responseData.data or nil
-
+    self.statusCode = _.get(responseData, {'statusCode'}, 0)
     self._contentType = _.get(self.headers, {'Content-Type'}, '')
-
+    
     self.isJSON = false
 
-    self.statusCode = responseData.statusCode
-
+    if self.headers.status then
+        self.statusCode = tonumber((self.headers.status):match("%s+(%S+)")) 
+    end
+    
     self.ok = false
     if (self.statusCode >= 200 and self.statusCode <= 299) then
         self.ok = true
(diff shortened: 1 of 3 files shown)


gizmocuz:
Thanks for the fast change !!

gizmocuz:
Can't this also be used with a GET ?

rwaaren:
> Can't this also be used with a GET ?

Yes. tested and working. I also keep using the 2nd method and deal with in HTTPResponse.lua in the theoretical case the header method should not give the expected result but have not seen such a case using my test-set and in my production environment.

gizmocuz:
@rwaaren , i would drop the additional adding of the status code to the http headers in SQLHelper
You always get a result code back in the original headers, and i think also always on the first line
Better to scan against the string "HTTP/1" if that's found, split it in an array (separated by a space), and you should have your result/return code

http://crimsonsecurity.blogspot.com/2016/08/format-of-http-response.html

"The very first line of a valid HTTP Response is consists of the protocol name, it's version, status code of the request, and a short description of the status code"

rwaaren:
@gizmocuz, sorry but I am not sure I completely understand.

Can you please confirm that this is what you want me to do ?

in dzVents.cpp CdzVents::ProcessHttpResponse 

- take statusCode out as separate value
- take status out of lua-subtable "headers" and create new lua-subtable "status" with keys: "protocol/version", "statusCode" and "statusMessage"    // by splitting line that begins with HTTP/

gizmocuz:
Sorry, yes it would be great if you could take the status code from the first HTTP Header array.
And remove the code in SQLHelper.cpp that pushes a status code in the array before it is send to CdzVents::ProcessHttpResponse

rwaaren:
Will try and test but might take a while. My C++ is not at the same level as my Lua

rwaaren:
PR #3181 is ready for review

dannybloe:
Lua part looks ok to me.

gizmocuz:
@rwaaren , thanks for the PR !
I would do it a bit differently... domoticz already contains a 'StringSplit' in helper.h/cpp,
you can use this in CdzVents::ProcessHttpResponse
Then for lua you can provide the status code only, you do not need a stringsplit function in lua

Please revert the change in History.txt, i will add it manually later...

PR #3177, opened 2019-05-04
Minor ui fixes for evohome

fullTalgoRythm:
The menu height for the evohome popup menu is squashed in the elemental theme. This is because the elemental custom css overrides the normal li line-height. The popup looks bad and is difficult to use on a touch device therefore add a specific ui-popup lil override in the elemental custom css that restores the  default line height for the evohome popup menu.

The human readable mode text for the evohome controller device is no longer being displayed. The original behaviour has been restored with a minor update to LightsController.js


gizmocuz:
Thanks !!

PR #3181, opened 2019-05-05
Changed method to return HTTPResponse statusCode to dzVents

rwaaren:
modified:   History.txt
	modified:   dzVents/documentation/README.md
	modified:   dzVents/documentation/README.wiki
	modified:   dzVents/documentation/history.md
	modified:   dzVents/runtime/HTTPResponse.lua
	modified:   dzVents/runtime/Utils.lua
	modified:   dzVents/runtime/tests/testHTTPResponse.lua
	modified:   main/SQLHelper.cpp
	modified:   main/dzVents.cpp

### DIFF
--- a/main/dzVents.cpp
+++ b/main/dzVents.cpp
@@ -16,7 +16,7 @@ extern http::server::CWebServerHelper m_webservers;
 CdzVents CdzVents::m_dzvents;
 
 CdzVents::CdzVents(void) :
-	m_version("2.4.18")
+	m_version("2.4.19")
 {
 	m_bdzVentsExist = false;
 }
@@ -93,8 +93,8 @@ void CdzVents::ProcessSecurity(lua_State *lua_state, const std::vector<CEventSys
 void CdzVents::ProcessHttpResponse(lua_State *lua_state, const std::vector<CEventSystem::_tEventQueue> &items)
 {
 	int index = 1;
-	int statusCode;
-
+	std::string status = "" ;
+	
 	lua_createtable(lua_state, 0, 0);
 	std::vector<CEventSystem::_tEventQueue>::const_iterator itt;
 	for (itt = items.begin(); itt != items.end(); ++itt)
@@ -117,25 +117,14 @@ void CdzVents::ProcessHttpResponse(lua_State *lua_state, const std::vector<CEven
 						lua_pushstring(lua_state, (*itt2).substr(pos + 2).c_str());
 						lua_rawset(lua_state, -3);
 					}
-					else
-					{
-						size_t pos = (*itt2).find("HTTP/");
-						if (pos != std::string::npos)
-						{
-							lua_pushstring(lua_state, "status");
-							lua_pushstring(lua_state, (*itt2).c_str());
-							lua_rawset(lua_state, -3);
-						}	
-					}	
 				}
-				// last item in vector is status code when method = GET
-				itt2 = itt->vData.end() - 1;
-				std::stringstream ss(*itt2);
-				ss >> statusCode;
+				size_t pos = (*itt->vData.begin()).find("HTTP/");
+				if (pos != std::string::npos)
+					status = *itt->vData.begin();
 			}
 			lua_rawset(lua_state, -3);
-			lua_pushstring(lua_state, "statusCode");
-			lua_pushnumber(lua_state, (lua_Number)statusCode);
+			lua_pushstring(lua_state, "status");
+			lua_pushstring(lua_state, status.c_str());
 			lua_rawset(lua_state, -3);
 			lua_pushstring(lua_state, "data");
 			lua_pushstring(lua_state, itt->sValue.c_str());
--- a/main/dzVents.cpp
+++ b/main/dzVents.cpp
@@ -90,10 +90,14 @@ void CdzVents::ProcessSecurity(lua_State *lua_state, const std::vector<CEventSys
 	lua_setglobal(lua_state, "securityupdates");
 }
 
-void CdzVents::ProcessHttpResponse(lua_State *lua_state, const std::vector<CEventSystem::_tEventQueue> &items)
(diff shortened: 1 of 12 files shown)


main/dzVents.cpp
```diff
@@ -117,25 +117,14 @@ void CdzVents::ProcessHttpResponse(lua_State *lua_state, const std::vector<CEven
...
-							lua_rawset(lua_state, -3);
-						}	
-					}	
 				}
-				// last item in vector is status code when method = GET
-				itt2 = itt->vData.end() - 1;
-				std::stringstream ss(*itt2);
-				ss >> statusCode;
+				size_t pos = (*itt->vData.begin()).find("HTTP/");
+				if (pos != std::string::npos)
```
gizmocuz: I would compare it against 0 as it needs to be at the beginning

if (pos != 0)

main/dzVents.cpp
```diff
@@ -93,8 +93,8 @@ void CdzVents::ProcessSecurity(lua_State *lua_state, const std::vector<CEventSys
 void CdzVents::ProcessHttpResponse(lua_State *lua_state, const std::vector<CEventSystem::_tEventQueue> &items)
 {
 	int index = 1;
-	int statusCode;
-
+	std::string status = "" ;
```
gizmocuz: I would split the string into a status_code (integer) and a status_text (string) below

main/dzVents.cpp
```diff
@@ -117,25 +117,14 @@ void CdzVents::ProcessHttpResponse(lua_State *lua_state, const std::vector<CEven
...
-						}	
-					}	
 				}
-				// last item in vector is status code when method = GET
-				itt2 = itt->vData.end() - 1;
-				std::stringstream ss(*itt2);
-				ss >> statusCode;
+				size_t pos = (*itt->vData.begin()).find("HTTP/");
+				if (pos != std::string::npos)
+					status = *itt->vData.begin();
```
gizmocuz: i would use the helper function StringSplit (split with a space as delimiter) and get the status_code (integer) and status_text, and push these to Lua

rwaaren: Happy to do so but don't know how in C++
In dzVents / Lua I now split this in protocol, statusCode and statusMessage (which can be multiple words)
Did search for something I could 'borrow' but no luck yet

main/dzVents.cpp
rwaaren: Thx !
Will do 
if (pos == 0)

main/dzVents.cpp
gizmocuz: Okey i will look into this.. it is more complex.... if an invalid url is entered, we do not get a response back from the server of course...
Now it returns a number '7' ... this is a curl return code CURLE_COULDNT_CONNECT (7)
it's not a header...
I am going to generate the header... hold on for a while

gizmocuz: Okey, please check with the latest beta code... the HTTP status code is now always in the header.
Also when there is an internal curl error (or when it can't connect), this is inside a header

This is the code that (i think) should work, only when using 'statusText' inside a dzVents script i get an error complaining about a nil field... so i do not think i did it correctly

I see a table being allocated which is one larger then the amounts of headers, i think that is correct? (as statusText is an extra field)

```
void CdzVents::ProcessHttpResponse(lua_State* lua_state, const std::vector<CEventSystem::_tEventQueue>& items)
{
	int index = 1;

	lua_createtable(lua_state, 0, 0);
	std::vector<CEventSystem::_tEventQueue>::const_iterator itt;
	for (itt = items.begin(); itt != items.end(); ++itt)
	{
		if (itt->reason == m_mainworker.m_eventsystem.REASON_URL)
		{
			lua_pushnumber(lua_state, (lua_Number)index);
			lua_createtable(lua_state, 0, 0);
			lua_pushstring(lua_state, "headers");
			lua_createtable(lua_state, (int)itt->vData.size() + 1, 0);
			if (!itt->vData.empty())
			{
				std::vector<std::string>::const_iterator itt2;
				for (itt2 = itt->vData.begin(); itt2 != itt->vData.end(); ++itt2)
				{
					std::string header = (*itt2);

					size_t pos = header.find(": ");
					if (pos != std::string::npos)
					{
						lua_pushstring(lua_state, header.substr(0, pos).c_str());
						lua_pushstring(lua_state, header.substr(pos + 2).c_str());
						lua_rawset(lua_state, -3);
					}
					else
					{
						if (header.find("HTTP/") == 0)
						{
							std::vector<std::string> results;
							StringSplit(header, " ", results);
							if (results.size() >= 3)
							{
								lua_pushstring(lua_state, "statusCode");
								lua_pushnumber(lua_state, (lua_Number)atoi(results[1].c_str()));
								lua_rawset(lua_state, -3);

								pos = header.find(results[1]);
								std::string statusText = header.substr(pos + results[1].size() + 1);

								lua_pushstring(lua_state, "statusText");
								lua_pushstring(lua_state, statusText.c_str());
								lua_rawset(lua_state, -3);
							}
						}
					}
				}
			}
			lua_rawset(lua_state, -3); //end of table

			lua_pushstring(lua_state, "data");
			lua_pushstring(lua_state, itt->sValue.c_str());
			lua_rawset(lua_state, -3);
			lua_pushstring(lua_state, "callback");
			lua_pushstring(lua_state, itt->nValueWording.c_str());
			lua_rawset(lua_state, -3);
			lua_settable(lua_state, -3); // number entry
			index++;
		}
	}
	lua_setglobal(lua_state, "httpresponse");
}
```

rwaaren: Combination of dzVents.cpp and HTTPResponse.lua handle this situation by generating a statusCode 0

rwaaren: Will do but no time left this morning. Will come back on this later today.

rwaaren: Should be OK in latest commit

gizmocuz:
Thanks !!

PR #3190, opened 2019-05-08
hardware/openzwave/Manager.h: fix build with latest openzwave

ffontaine:
openzwave in version 1.6 added an optional _pos parameter to
GetValueHelp and GetValueLabel, as a result link fails on:

OpenZWave.cpp:(.text+0x63bc): undefined reference to `OpenZWave::Manager::GetValueHelp[abi:cxx11](OpenZWave::ValueID const&)'

To fix this error, update hardware/openzwave/Manager.h but a better
solution would be to use the headers provided by openzwave instead of
including them in domoticz. However, this would be a major change.

### DIFF
--- a/hardware/openzwave/Manager.h
+++ b/hardware/openzwave/Manager.h
@@ -887,12 +887,13 @@ OPENZWAVE_EXPORT_WARNINGS_ON
 		/**
 		 * \brief Gets the user-friendly label for the value.
 		 * \param _id The unique identifier of the value.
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thanks for contributing !

It's on my to-do list to integrate 1.6 and also another way to handle labels...
I updated the wiki, for now please issue

git checkout v1.4-3335-g74e05982

inside the openzwave folder

PR #3196, opened 2019-05-09
Fixed dzVents error messages and test script

rwaaren:


gizmocuz (commented):
Small question .. did you know a lot of idents have changed? (Tabs)
If this was python.....

rwaaren:
Yes I know, Agreed with @dannybloe that we use tabs to indent where possible for Lua / dzVents code. So if I notice "left over spaces" in code that I modify, I change that to tabs.

PR #3202, opened 2019-05-10
[BleBox] support for tempSensor

szczukot:


szczukot:
I have a big problem. Where can I save IP address for temp sensor ? 
In method CDomoticzHardwareBase::SendTempSensor i see:
	tsen.TEMP.id1 = (NodeID & 0xFF00) >> 8;
	tsen.TEMP.id2 = NodeID & 0xFF;
That only 2 bytes used :(
This same problem exist for example for 1wire: 
https://www.domoticz.com/forum/viewtopic.php?t=10808

gizmocuz:
@szczukot , 1wire is kind of the same problem, but not quite
You have the option for 65K ID's, that should be enough, i won't expect one BleBox user to have a conflict?
You can use the 'WOLNodes' table, table needs to be renamed, but hardware is using this as helper table.
Is it possible you could undo the formatting changes ? The formatting was just find (switch/case for example)
If you are in visual studio, and press in your document control-k then control-d it will format it back
(Depending on the Visual Studio options)

gizmocuz:
Question, why do you need to save IP address ?
If you use the last two numbers as ID, should that not be clear enough ?

szczukot:
All devices used different IP.
Ok - i check WOLNodes table.

Formatting: I used always CTRL+K+D. But sometimes You have different formating (https://github.com/domoticz/domoticz/commit/b60437c400656aa96f08019a83d32b4bd1bf2b2d)

gizmocuz:
The above commit just makes clear code, is not about indent spacing (tabs/spaces)
(Sorry i meant indent/spacing before)

gizmocuz:
Thanks!

Issue #3205, opened 2019-05-11
switch log graphics and timestamp  wrong


GravityRZ:
domoticz V4.10717
cleared cache

the switchlog seems wrong
the part that holds the data is correct but the parts which hold the graphics are wrong.

graphics above the textlog are empty.
graphics below the textlog do not represend the textlog data, moments it was switche on/off and also the popup data is completely wrong (all timestamps are jan 1 abnd time is wrong)
checked multiple switches either physical or virtual but they al have the same problem.

![switchlog-1](https://user-images.githubusercontent.com/16853535/57568675-1f3b4e80-73eb-11e9-8208-c69efdd71b61.JPG)
![switchlog-2](https://user-images.githubusercontent.com/16853535/57568678-24000280-73eb-11e9-90fb-06292ed602ff.JPG)
![switchlog-3](https://user-images.githubusercontent.com/16853535/57568685-295d4d00-73eb-11e9-85a0-79972cf0b95f.png)
![switchlog-4](https://user-images.githubusercontent.com/16853535/57568688-2cf0d400-73eb-11e9-931e-d0b4725683b7.jpg)

GravityRZ:
tested it. fixed in build 734

PR #3206, opened 2019-05-11
#3205 Fix date parsing in device log

stas-demydiuk:


grzejszop:
Hi,

I changed files as you wrote: 
www/app/log/DeviceLog.html
www/app/log/components/DeviceLevelChart.js 
www/app/log/components/DeviceOnOffChart.js
 and still have problem with shown graphs of sensors, still have time 1:00:00:00 (date 1970). In whole domoticz i have correct local time. Do you know whats wrong?

gizmocuz:
Thanks for the fast response !

PR #3217, opened 2019-05-15
Fix Detection of Python 3.x libraries on FreeBSD (Delayed Link).

xbeaudouin:
On FreeBSD the shared lib used for Python has the suffix "m" (don't ask me why)..
This trivial patch fix the delayed linker to find the right lib on FreeBSD.

Regards


gizmocuz:
Verified via https://www.freshports.org/lang/python37 (lib/libpython3.7m.so)

gizmocuz:
Thanks for the patch !

PR #3220, opened 2019-05-15
fixed memory leak in XPath library

xbeaudouin:
This patch was produced on SourceForge. Was never merged into tinyxpath code.
Seems the maintainer of tinyxpath doesn't maintain that anymore :(  

The XPath library did not take into consideration that a query would be run more
than once.  As a consequence, if it was it was leaking memory in the overridden
assignment operator (=).  I added a check to free the memory to prevent the
memory leak.

This was occurring when I would call the "u_compute_xpath_node_set()" function
in a loop to retrieve the number of nodes in the set.

Credit :  Benjamin Rood <brood@attotech.com>


gizmocuz:
Thanks!!

PR #3239, opened 2019-05-19
Properly display water sensor on dashboard

stas-demydiuk:


gizmocuz:
What's a watersensor type switch ? ... I know a watersensor as a metering sensor, which is supported on the dashboard/utility page.
I am not aware of a 'watersensor switch' in domoticz ? It looks like it's a normal switch that you can edit with a custom icon ?

stas-demydiuk:
Yes, it is just regular contact switch with icon set to "Water"

gizmocuz:
I don't have this in my switch types ? I don't think domoticz/gui recognizes this?
Where is this used ?
Why not use a default on/off switch with a custom icon ?

stas-demydiuk:
It is a default contact type. The issue is dashboard (svg one) currently doesn't draw the icon properly as it assumes "On" and "Off" states while device has "closed" and "open" instead. So this commit is created to fix this.

stas-demydiuk:
The best option would be to create one more device type "WaterLeakSensor" because currently there is no one to represent water leakage device in Domoticz, and the only way is to use some workarounds: as example create "contact" sensor and manually change the icon.

gizmocuz:
What hardware is this?
Yes if it is a sensor like I have it is a connect sensor
Nothing new has to be created for this
Or a custom icon switch... Water detected or not

gizmocuz:
I would just use the contact sensor.. not sure if we can change it's icon yet....
Same like a soil sensor
I have added a water icon a long time ago 😀

stas-demydiuk:
Yes, we can change icon of contact sensor to water one, but this icon isn't displayed properly on floor plan (dashboard). This PR should fix that.

gizmocuz:
Thanks, sorry, got confused because the PR said dashboard...

PR #3251, opened 2019-05-21
appversion.default: bump version to 10717

ffontaine:


gizmocuz:
Thanks! Still wondering why we need this file... build systems are online and should be able to get this number (see makebeta/makerelease scripts)

ffontaine:
Some build systems such as buildroot are not "online" and does not want to depends on an Internet connection to build packages. As a result, the domoticz tarball must contain the correct version otherwise the wrong version is displayed in the web ui and when the user checks for an update, domoticz wrongly says that a new version is available. This bug was reported by one of our users in a previous domoticz version (see http://lists.busybox.net/pipermail/buildroot/2018-August/227787.html).

gizmocuz:
@ffontaine , when the build succeeds, won't it be uploaded 'somewhere' ? does this not imply that there should be a internet connection?
This particular file has caused the same issue in the past , and will cause the same issue in the feature

Can the build number not be taken from the first line of the history file?

Anyway, i vote to check this online

ffontaine:
When the build succeeds, it remains on the PC of our users. However, I agree that the current mechanism is not satisfying. I'll send a PR soon.

gizmocuz:
@ffontaine , thanks for the PR

PR #3256, opened 2019-05-22
Remove appversion.default

ffontaine:
Use History.txt to retrieve version and remove appversion.default


gizmocuz:
Thanks!

Issue #3261, opened 2019-05-23
Buffer Overflow Vulnerability in YouLess class


Drumstickx:
## Description
The YouLess hardware class contains a buffer overflow vulnerability due to a missing bounds check when writing user input into a char array buffer.
This can be exploited by an attacker to crash the application or manipulate the execution in other ways (e.g., executing arbitrary code).

## Affected Version(s)
Current release version (4.10717) and before.

## Steps to Reproduce
1. Create a fresh clone of the repository and compile it. Using an existing installation may result in data loss!
2. Run domoticz (`./domoticz`)
3. Visit the web interface tab for adding hardware (`http://localhost:8080/#/Hardware`) in your browser
4. Add a new device of type `YouLess Meter with LAN interface` with the following parameters:
	- remote address: `AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA`
	- port: `12345`
5. The parameters used in 4. result in an overflow of the `szURL` buffer when one of the methods `GetP1Details` or `GetMeterDetails` is called
6. The program crashes as result of the buffer overflow

## Fix
Use `snprintf` instead of `sprintf` and supply the size of the target variable to the function to prevent buffer overflows.
See the attached PR.

gizmocuz:
Thanks for reporting! Fixed this a bit differently, as we do in all other classes using URL's

Drumstickx:
The second occurance of the buffer overflow should also be fixed. See PR.

xbeaudouin:
Well : 
```
~/domoticz.dev $ grep -r sprintf * | wc -l
    1624
```

I think some of have to check all of these.  One that are scary are ones in webserver.cpp I think. Maybe they correctly tested but they can be abused...

Unfortunatly I will be busy next month, so... I'll have a nice look during july... :)

gizmocuz:
Thanks! I guess most have been looked at already but another is very welcome

PR #3262, opened 2019-05-23
Fixed Buffer Overflow.

Drumstickx:
See #3261


gizmocuz:
Thanks for the PR!!!!
I solved it a bit differently as we do this also in the other classes

https://github.com/domoticz/domoticz/commit/e8c05e7c29130eac6465ec0664a41dd0790069d3

PR #3266, opened 2019-05-24
Fixed Buffer Overflow.

Drumstickx:
See #3261


gizmocuz:
Thanks, but the changes where already applied

PR #3269, opened 2019-05-25
Youless, fixed Buffer Overflow.

Drumstickx:
This fixes the second occurance of the buffer overflow that was not fixed.

See my last comment in #3261


rwaaren (commented):
This fix breaks the connection between domoticz and my Youless.
Checked with previous version, other system and CLI curl and locally with reverted fix and confirmed that the error Error: YouLess: Error connecting to: 192.168.192.222
that was introduced with this fix is solved by reverting.

Issue #3274, opened 2019-05-27
fix #3269 Breaks the connection between domoticz and Youless.


rwaaren:
#3269 

This fix breaks the connection between domoticz and my Youless.
Checked with previous version, other system and CLI curl and locally with reverted fix and confirmed that the error Error: YouLess: Error connecting to: 192.168.192.222
that was introduced with this fix is solved by reverting.

gizmocuz:
Can you debug the code? It should generate the same url

rwaaren:
Will do.

rwaaren:
> Can you debug the code? It should generate the same url

It did not generate the same url.  Was it because the /e compared to /a ? or was there something else ?
#3277 does fix it (tested)

gizmocuz:
@rwaaren , thanks for the PR!
Verified the stringstream, no problems detected. It's also used everywhere and is perfect c++
If you set a breakpoint at the beginning of HTTPClient::GET you should see this
This should be solved in beta 10836

PR #3277, opened 2019-05-28
fix for issue #3274 ( connection to Youless failed )

rwaaren:


gizmocuz:
@rwaaren , thanks for the PR, yes, this was a small typo introduced by Drumstickx
The password is optional

rwaaren:
> @rwaaren , thanks for the PR, yes, this was a small typo introduced by Drumstickx
> The password is optional

Was still working to pinpoint the exact reason why it failed and removing the changes I made during that process step by step and check if it was still working.
On the optional password: the url with "&w=" wil work with an empty password.  So extra check is not really needed. Is it better practice to do the extra check anyway ?

gizmocuz:
@rwaaren , yes it is, and it makes the code clearer. Who knows in the feature when you get a new YouLess firmware it breaks because an empty &w= is provied

PR #3279, opened 2019-05-28
Added Hue LCS001 to GAMUT_C_BULBS_LIST and extra command to OpenTherm to re-enable PS=0

w1ll14m:
I recently got Hue Outdoor Lily Spots and discovered that the bulbs are not known to Domoticz yet. I added LCS001 GAMUT_C_BULBS_LIST.

I use Domoticz and OtMonitor (OpenTherm Monitor) at the same time, Domoticz uses PS=1 to gather information but that causes OtMonitor to stop receiving OT messages. I added a little sleep after sending the PS=1 command followed by sending PS=0. With this Domoticz and OtMonitor can coexist :)

### DIFF
--- a/hardware/OTGWBase.cpp
+++ b/hardware/OTGWBase.cpp
@@ -250,7 +250,7 @@ void OTGWBase::GetGatewayDetails()
 	strcpy(szCmd, "PS=1\r\n");
 	WriteInt((const unsigned char*)&szCmd, (const unsigned char)strlen(szCmd));
 
-	/* Wait for 0.75 seconds for the command to complete and switch back to PS=0 to fix use of OtMonitor */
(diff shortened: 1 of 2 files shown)


Olifant1990:
Please consider splitting your pull requests on subject.
That way, for example, the PR about the Philips Hue can be accepted, but the second can be held to have some thoughts about it (not that I personaly have to say something about it ;))

w1ll14m:
> Please consider splitting your pull requests on subject.
> That way, for example, the PR about the Philips Hue can be accepted, but the second can be held to have some thoughts about it (not that I personaly have to say something about it ;))

Hi Olifant, 

I'm fine with that. Actually this is the first time really using GitHub so I'm still figuring things out :). Can I still split this pull request or should I create a new one?

Regards,
William

PR #3286, opened 2019-05-29
fix comparison operator COMMAND_CLASS_SCENE_ACTIVATION

petergebruers:


gizmocuz:
Thanks!! Good catch !

PR #3296, opened 2019-06-04
Speed optimization

Olifant1990:
Some speed increasement for database update


gizmocuz:
Thanks but it is an old patch and not being used anymore

PR #3297, opened 2019-06-04
Line endings supported in text log

Olifant1990:
\r \n line endings are parsed to <br \> for line endings in log. As continuation of commit https://github.com/domoticz/domoticz/pull/1942

### DIFF
--- a/main/WebServer.cpp
+++ b/main/WebServer.cpp
@@ -13097,7 +13097,7 @@ namespace http {
 			root["status"] = "OK";
 			root["title"] = "TextLog";
 
-			result = m_sql.safe_query("SELECT ROWID, sValue, Date FROM LightingLog WHERE (DeviceRowID==%" PRIu64 ") ORDER BY Date DESC",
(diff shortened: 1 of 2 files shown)


Olifant1990:
afbeelding
Line endings are supported now

gizmocuz:
Thanks! Could you undo the change in WebServer.cpp, the previous code was OK and the date field is later then sValue in the database as well

Olifant1990:
Yes I can, but is it an alternative to change the order of the rows

root["result"][ii]["idx"] = sd[0];
root["result"][ii]["Date"] = sd[2];
root["result"][ii]["Data"] = sd[1];

to

root["result"][ii]["idx"] = sd[0];
root["result"][ii]["Data"] = sd[1];
root["result"][ii]["Date"] = sd[2];
?

It is less a risc to make a mistake. Same order you select, same order in code you work with.
I was debugging here, so I came upon this.
Otherwise, will undo it maybe tonight

gizmocuz:
Yes just revert the changes in that file

Olifant1990:
Done!

gizmocuz:
Thanks!

PR #3302, opened 2019-06-07
 Fix display of custom theme images when webroot is defined

davidlb:
If webroot is defined in Domoticz, custom images of themes are not correctly rendered.
Issue initialy reported on the Machinon theme EdddieN/machinon-domoticz_theme#256

To reproduce issue:
- run domoticz with webroot --> ./domoticz -webroot domoticz
- select a theme with custom images (for example [machinon](https://github.com/EdddieN/machinon-domoticz_theme))
- images used are the default one not the custom from the theme


webserver/cWebem.cpp
```diff
@@ -2021,11 +2026,12 @@ namespace http {
...
 				{
-					if (requestCopy.uri.find("/images/") == 0)
+					std::string uri = myWebem->ExtractRequestPath(requestCopy.uri);
+					if (uri.find("/images/") == 0)
 					{
-						std::string theme_images_path = myWebem->m_actTheme + requestCopy.uri;
+						std::string theme_images_path = myWebem->m_actTheme + uri;
 						if (file_exist((doc_root_ + theme_images_path).c_str()))
-							requestCopy.uri = theme_images_path;
+							requestCopy.uri = myWebem->GetWebRoot().c_str() + theme_images_path;
```
gizmocuz: As requestCopy.uri is a std::string, there is no need to call .c_str() (**myWebem->GetWebRoot().c_str()**)

gizmocuz:
@davidlb , did you noticed the review ?

davidlb:
@gizmocuz, yeah sorry, I'll take that in account tomorrow. I'm in a remote location with very limited connection

gizmocuz:
@davidlb , no problems... just double checking ;)
I just changed it online... going to test the patch and merge asap...

gizmocuz:
@davidlb , checking your PR and have some questions

First you are going to call ExtractRequestPath, which is a more expensive call then searching for "/images/", is that needed?

Then you are checking (file_exist) if this file exists: **doc_root_ + theme_images_path**
If it does, you are loading another file "**myWebem->GetWebRoot() + theme_images_path**"
The later file could not exist

This is why in the original code a variable 'theme_images_path' was searched for, and if existed, used

Could you also give an example how to reproduce where this goes wrong ? (custom theme with image example)
As this code was already created for custom images in a theme (maybe to with webroot)

davidlb:
Regarding the use of ExtractRequestPath, it's to be sure we don't catch another directory called "images" in a theme (ie theme/custom/images).

For the second point, you're absolutely right, my solution is not working in this case. Not sure how to deal with it :/

To reproduce the issue:
- install machinon theme
- run domoticz with webroot --> ./domoticz -webroot domoticz
- select machinon theme
- images used are the default one not the custom from the theme

If you try without webroot, it works fine, custom images are loaded.

gizmocuz:
Thanks! I will try to replicate this a.s.a.p.!

gizmocuz:
Thanks, i can replicate the issue

gizmocuz:
It is not the theme images that go wrong, but custom images the user selects for a switch are not loaded correct right?
Because i still see all default images being changed:

image

davidlb:
Oh wait, I applied a temporary fix to overwrite url for images in the theme (but it breaks custom images)/
Please edit file js/function.js and remove/comment lines 30 to 38 to reproduce the bug.

gizmocuz:
Perfect, when i removed those lines, also the custom selected icons work !
So we can close this patch ?

gizmocuz:
image

gizmocuz:
Ohhh wait... i see the original icons... nothing wrong with that ? ;) but i guess thats not what you want.... going to debug

gizmocuz:
Why is this theme calling 'checkauth()' every time ? setInterval(function () {
This is not needed, it seems to make the temperature widgets scrolling info to hickup as well...
The 'admin rights' is already in the rootScope which you can already access
Also please try to rewrite the $.ajax calls to $http it will make your theme much much faster/responsive

gizmocuz:
Your patch is working fine... but please have a look at the previous message...

davidlb:
Thanks for your remark, I'll check who wrote that and why.

PR #3306, opened 2019-06-09
Broken ozwcache filename

Ittiz:
The name of the ozwcfg file was changed to ozwcache but not updated everywhere.  This broken one of my scripts so I fixed it.  There may be more instances hidden in the code.


gizmocuz:
Thanks for reporting! This was the only instance

PR #3312, opened 2019-06-11
Allow changing icon for Lighting 5 types

rwaaren:
modified:   www/app/devices/deviceFactory.js

Based on forum topic / post [ Option to select icon for on/off switches has disappeared](https://www.domoticz.com/forum/viewtopic.php?f=6&t=27162#p216903)

### DIFF
--- a/www/app/devices/deviceFactory.js
+++ b/www/app/devices/deviceFactory.js
@@ -5,7 +5,7 @@ define(function () {
 
         function DeviceIcon(device) {
             this.isConfigurable = function() {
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thanks!

rwaaren:
You got to be kidding.... You merged so quickly that I was kind of worried, I pushed the wrong button  😁

gizmocuz:
Ha ! Yeah it was just a small addon ;) but actually i think we should check for "Lighting" here... so it will support 1,2,3,4 and 5

Issue #3314, opened 2019-06-12
Blockly events not working on latest beta


bennydente:
Hi, i'm on version Version: 4.10897 on ubuntu 18.04 but something is wrong on blockly events.
this simple test events never triggered.

![Capture](https://user-images.githubusercontent.com/22501791/59341525-4fd31880-8d08-11e9-99dd-a36a0b972286.JPG)

i have a cronjob that every minute:
- set clock60sec variable to 1
- sleep 30sec
- set clock60sec variable to 0

so, i aspect that the event is triggered every minute. but not.... :-(

rwaaren:
Better to post this on the forum first. 
Have you tried to change the user variable on the GUI ? It might just be the case that the Blockly will get fired then.

bennydente:
hi, yes, i tryed via GUI and i can set every integer value. also i see the variable change every 30 second according cronjob.

bennydente:
hi, i better test, and if I change the variable value by GUI the event is triggerd. if I change the value of varialbe from URL by json command via cronjob, the event is not triggered.

rwaaren:
Can you please close this issue here and continue discussion on [ the forum topic about this ](https://www.domoticz.com/forum/viewtopic.php?f=6&t=28333&p=217009) ?
as you write here in your last update: "the event is triggered."

bennydente:
hi, are you planning solve bug in next release? thakns for you job.

PR #3320, opened 2019-06-13
Allow variabletype in API call to be entered as word or as digit for AddUserVariable and UpdateUserVariable

rwaaren:
modified:   WebServer.cpp

Prevent entering text as variabletype as any text will be considered as 0 (vtype integer)
but this will not trigger an event because not equal to result[0][1] and bTypeNameChanged will therefore change to true 

Also added check for vtype  in 
`/json.htm?type=command&param=updateuservariable&vname=USERVARIABLENAME&
vtype=USERVARIABLETYPE&vvalue=USERVARIABLEVALUE`

 0 <= vtype <= 4

### DIFF
--- a/main/WebServer.cpp
+++ b/main/WebServer.cpp
@@ -1879,27 +1879,54 @@ namespace http {
 			if (session.rights != 2)
 			{
 				session.reply_status = reply::forbidden;
+				_log.Log(LOG_ERROR, "User: %s tried to add a uservariable!", session.username.c_str());
 				return; //Only admin user allowed
 			}
 			std::string variablename = request::findValue(&req, "vname");
 			std::string variablevalue = request::findValue(&req, "vvalue");
 			std::string variabletype = request::findValue(&req, "vtype");
+			
+			root["title"] = "AddUserVariable";
+			root["status"] = "ERR";
+			
+			if (!std::isdigit(variabletype[0])) 
+			{
+				stdlower(variabletype);
+				if (variabletype == "integer")
+					variabletype = "0";
+				else if (variabletype == "float")
+					variabletype = "1";
+				else if (variabletype == "string")
+					variabletype = "2";
+				else if (variabletype == "date")
+					variabletype = "3";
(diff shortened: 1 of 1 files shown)


gizmocuz:
@rwaaren , can you give an example where 'text' is entered as variable type ?
In the Wiki/JSON API it is stated that it should be numbers, if someone enters 'apple' instead of 2, then of course it is wrong.

But as it is only one number, you could do a more easy check:

`if (!isdigit(variabletype.c_str())) return false;`

rwaaren:
Example where the use of the word Integer caused confusion because it did update the variable but did not trigger an event; Yes this was a user error but not very obvious at first glance.

[#3314](https://github.com/domoticz/domoticz/issues/3314)
[Forum post ](https://www.domoticz.com/forum/viewtopic.php?f=6&t=28333&p=217082)

I thought about your way of simplifying the check but my consideration was to also check if the value is in the valid range [0..4] and because the value is now also in an integer the 2nd check was relatively easy. Please advise if there are more efficient or cleaner ways to do this.

gizmocuz:
Maybe we could translate the types as text to the correct enum ? ("Integer" => _eUsrVariableType::USERVARTYPE_INTEGER) , maybe first make it lowercase (with helper.cpp/stdlower) and check for "integer"

First check, is it digit, then check if it is in a valid range, if not, bail out (please add a errormessage)
If a string, try to match the type to enum, if not, bail out (please add a errormessage)

I know it is a bit more coding, but then we have both situations covered

Or we could say, the one created the link on the forum was not using the correct syntax

rwaaren:
Like your idea to allow both (valid) strings and (valid) integers. Will try to come up with something but expect that it might take me some time. (I am not fluent in C++)

rwaaren:
Added the discussed logic and resilience and messages for adduservariable and updateuservariable but do not know how to cast std::string to _eUsrVariableType so used hardcoded integers.

gizmocuz:
Thanks !!

PR #3334, opened 2019-06-18
Sending notifications to telegram through a proxy

devalex86:


gizmocuz:
Thanks, but this goes for all HTTP requests...
Can't you specify a proxy on your system?

devalex86:
This only affects telegram notification requests, since no other proxies will be specified.
Redirect requests only for telegram failed.
A proxy is needed only for telegrams and it would be wrong to send all system traffic through it.

devalex86:
I think I understand what the problem is
maybe here notifications/NotificationTelegram.cpp
```c++
#ifndef WIN32
	HTTPClient::SetSecurityOptions(true, true);
#endif
        if (_proxy.length() > 0) {
              HTTPClient::SetProxy(CURLEncode::URLDecode(_proxy));
	}
	bRet = HTTPClient::POST(sUrl,sPostData,ExtraHeaders,sResult);
        if (_proxy.length() > 0) {
              HTTPClient::SetProxy("");
	}
#ifndef WIN32
	HTTPClient::SetSecurityOptions(false, false);
#endif
```
but this does not solve the problem with parallel requests

gizmocuz:
@devalex86, let me get this correct....
Telegram uses the HTTP protocol to send messages.
You seem to be needing a 'proxy' to use outgoing HTTP traffic on your system, else why you only need this for Telegram and not for.. let's say... when you browse to a web page ?
On what hardware is Domoticz installed ?

devalex86:
Domoticz i have deployed on a Raspbian. This is true in countries where this service is blocked. I understand how to wrap all traffic through a proxy, but this is a bad idea in this case, since a third-party proxy with bad bandwidth and  is needed only for telegram.

gizmocuz:
I don't think you understand what I am saying or suggesting
But best use another notification system then
We have plenty

PR #3341, opened 2019-06-22
Fixes for "Unknown" in Name field. Fix Lux value scale error.

petergebruers:
Both issues are related to the update of OpenZWave Commit 4478eea2
by Justin Hammond, 06/17/2019 introducing improvements to the MultiSensor CC

I am going to open PR's to get the FGMS-001 ZW5 fully supported. That device is an excellent test case because it has Temperature, Illuminance, Acceleration and Seismic sensors.

This is what the device looks like before this PR:

FGMS-1

This PR fixes the "Unknown" Columns (for new devices). The first fix concerns the OZW config, from Commit 4478eea2 onwards OZW retrieves labels from an xml file and writes them to ozwcache. The xml file is missing in Domoticz. The second fix adds the label as a parameter instead of NULL in certain cases. I also add a log line in case OZW returns the "Unknown" label which should be rather exceptional. It is not and error because it does not really make the device inoperable.

The Lux versus % conversion logic was case sensitive and caused a 10x value increase. Fixed.

This is what the device looks like after this PR:

FGMS-2

The "Name" column makes sense know, but the device types for Acceleration and Seismic Intensity. I am working on a fix.


petergebruers:
If I am correct, CI fails because of OZW build failure:

```
make[1]: Entering directory `/home/travis/build/domoticz/domoticz/open-zwave-read-only/cpp/build'
Building OpenZWave Version 1.6-779-g057d2a28-dirty
(...)
make[1]: Entering directory `/home/travis/build/domoticz/domoticz/open-zwave-read-only/cpp/examples/MinOZW'
Building Main.cpp
In file included from /home/travis/build/domoticz/domoticz/open-zwave-read-only/cpp/src/Driver.h:38:0,
                 from /home/travis/build/domoticz/domoticz/open-zwave-read-only/cpp/src/Manager.h:39,
                 from Main.cpp:37:
/home/travis/build/domoticz/domoticz/open-zwave-read-only/cpp/src/Node.h:566:32: error: ‘std::shared_ptr’ has not been declared
    void SetProductDetails(std::shared_ptr<Internal::ProductDescriptor> product);
```

Not sure what is going on there, because <memory> is included (via Defs.h) and when I force an OZW build on my system with:

CC=c++-6 CXX=c++-6 BUILD=debug make -j

Then I get no errors and MinOZW is OK, so it's not because of the GCC version...

gizmocuz:
@petergebruers , there where other issues that needed to be adapted in Domoticz before a correct build could be done.  (sensor multilevel type changes like `ValueID_Index_SensorMultiLevel::Air_Temperature`). 

And there is currently a build issue in Travis (see the log), `shared_ptr` could not be found. A PR for this have been submitted (added the missing memory include)

https://github.com/OpenZWave/open-zwave/pull/1859

petergebruers:
I am not sure if I understand what you say about "build" and the temp IDs because locally everything builds and also there is no issue with temp sensors.

I see however an issue with multilevel sensors and device with multiple endpoints...

Please enlighten me :)

gizmocuz:
@petergebruers , about the maximum Lux percentage value, can we take this from the node's config file ?
For example here you can see that the maximum value is 1000

https://github.com/OpenZWave/open-zwave/blob/master/config/zooz/zse09.xml

I can't remember why I used the 1000, but it seemed good at that time

.. About the ID changes, there have been many 'ValueID_Index_SensorMultiLevel' changes recently, see

cpp/src/ValueIDIndexesDefines.def

and

https://github.com/OpenZWave/open-zwave/pull/1844/files

petergebruers:
I am not sure about that 1000, I'll try to find out. It has alsways been like that and if nobody complained...

Ah, you mean there are many more IDs. On my Mac I run a version which always creates custom sensors, so the "Accelleration" (see screenshots) no longer is a switch but reports proper value and units.

I have more issues with the implementation of the Notification class. I am working  on a document with some explanation...

gizmocuz:
@petergebruers , thanks, i will see it on Slack ;)
Currently no Travis build will succeed because of the missing include

petergebruers:
Fix for build was merged, OZW >= v1.6-792-g711a83af

petergebruers:
> For example here you can see that the maximum value is 1000
> 
> https://github.com/OpenZWave/open-zwave/blob/master/config/zooz/zse09.xml

Okay, in the parameter section they mention 1000 Lux - but I cannot deduce from the manual if this sensor reports Illumination as % or Lux... This seems a recent device and they always talk about "Lux" so I am guessing it reports in Lux. Unless of course someone tested it and confirms it is %.

I've had a look at a few older devices, like the HSP02 and ST814 but afaik they do not report Illumination at all, but they do have a % setting threshold do enable the motion reports.

Anyway... between 1.4 and 1.6 "lux" changed to "Lux" and so the test before this PR multiplied the Illumination level x10. That won't happen now because of the == "%" test...

petergebruers:
> Currently no Travis build will succeed because of the missing include

The addition of the memory header did not solve the build issue, Justin has pushed a change to OpenZWave master, it adds "-std=c++11" to the MinOZW compile.

hardware/OpenZWave.cpp
```diff
@@ -1510,6 +1510,10 @@ void COpenZWave::AddValue(const OpenZWave::ValueID& vID, const NodeInfo* pNodeIn
 	std::string vLabel = m_pManager->GetValueLabel(vID);
 	std::string vUnits = m_pManager->GetValueUnits(vID);
 
+	if (vLabel.compare("Unknown") == 0){
```
gizmocuz: Is it possible to make this compatible with our general coding style?
If (vlabel=="unknown")

petergebruers: Done! I plead guilty, I am lousy at C++ - I always expect the syntax to be long and difficult :)

gizmocuz:
Thanks!!

PR #3366, opened 2019-06-30
Activating a scene containing devices which are dimmable now correctly…

nhale25:
… sets the level parameter.


gizmocuz:
Thanks!!

jeroenkl:
WIth this change my devices in a scene which do not have a dimming function are not working anymore (not triggered anymore)

Issue #3367, opened 2019-07-01
XSS Vulnerability in ScenesController.js


Frige1:
## Description
The ScenesController.js is vulnerable against XSS. By adding a scene with some scripts in the scene name will lead to a XSS.
Obviously this is thankful for an attacker if he is privilege to adding or edit scenes. The attacker can execute arbitrary code.

## Affected Version
Current release 4.10717

## Steps to Reproduce
1. Clone and compile it. I follow the instructions from the wiki: https://www.domoticz.com/wiki/Raspberry_Pi_-_Build_Domoticz_from_source#Raspberry_Pi_-_Build_Domoticz_from_source
2. Run domoticz `./domoticz`
3. Open `http://localhost:8080/#/Scenes` in your browser.
4. Click button: Add Scene
5. Have fun to implement arbitrary scripts or even test it with `<script>alert("XSS");</script>`
![XSS-Vul](https://user-images.githubusercontent.com/21190223/60420534-468eea80-9be8-11e9-8c85-d6b68570aac9.png)

## Fix
Some `escape()` is used in the ScenesController.js, but this XSS is still present. Escaping the `item.name` fix the XSS. Furthermore use `encodeURI()` instead of `escape()`. The `escape()` function was deprecated in JavaScript v1.5.

![XSS-Fix](https://user-images.githubusercontent.com/21190223/60420799-daf94d00-9be8-11e9-92be-d7777b109e64.png)

See PR #3368

Frige1:
[https://github.com/domoticz/domoticz/pull/3368](url)

gizmocuz:
Thanks for the PR.
Your current PR causes all scene names that includes for example spaces to be displayed with %20 in the name
Also names containing non-latin characters to display wrongly.

It also seems not possible to change a name to a non-latin name

Is there another solution ?

Frige1:
Well, you're right. My focus was to (quick) fix the XSS-Vulnerability. 

Since now, i used encodeURI(), which escape not that much chars like encodeURIComponent(). Especially because (as you mentioned) some chars will be displayed as escaped chars.

Browsing on OWASP sites, i found a library for special XSS escaping:
https://www.owasp.org/index.php/Category:OWASP_Encoding_Project

Unfortunately I have not enough free-time space, to fix the XSS with propper presentation for the item.name . Maybe on a later point in time.

[...]

Issue #3329, opened 2019-06-17
RTL433 - Unhandlend sensor reading (Auriol IAN288624 / H13726A Weather Station)


find-a-good-nickname:
Hi!

I have this problem with the Auriol IAN288624 / H13726A type Weather Station. Got an RPi 3 with Domoticz v4.10717 and an RTL-SDR. This Auriol WS seems the same as Alecto V1, I think.

Please be so kind to help me to resolve this problem.
Thanks for your kindness

Domoticz log:

```
[...]

PR #3368, opened 2019-07-01
Fix XSS vulnerability

Frige1:
XXS in Line 935. The item.Name was not escaped.

See Issue #3367

### DIFF
--- a/www/app/ScenesController.js
+++ b/www/app/ScenesController.js
@@ -663,8 +663,8 @@ define(['app'], function (app) {
 				"sPaginationType": "full_numbers",
 				language: $.DataTableLanguage
 			});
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thanks for the PR.
Your current PR causes all scene names that includes for example spaces to be displayed with %20 in the name
Also names containing non-latin characters to display wrongly.

It also seems not possible to change a name to a non-latin name

Is there another solution ?

gizmocuz:
Maybe solved by using the encodeURIComponent/decodeuricomponent functions ?

gizmocuz:
Going to be solved differently (as discussed in the #3329

PR #3376, opened 2019-07-05
Adding support for reading consumption and net consumption counters from Enphase Envoy Metered

FrankvdAa:
Added code to Enphase API to read (net) consumption counters from production.json. The consumption counters are read all day in stead of only between sunrise and sunset (as is done for production readings).


FrankvdAa:
As I only have a Metered version of the Envoy, I haven't been able to test against non-Metered Envoy, but I reckon it shouldn't cause any problems.

gizmocuz:
Hi, i noticed you wanted to make a PR, then closed it, deleted it, restored it and deleted it...
Are you sure you want to delete it ?

gizmocuz:
I noticed the new PR
https://github.com/domoticz/domoticz/pull/3378

PR #3381, opened 2019-07-06
Added support for MySensors MQTT custom topic prefixes.

EtherFidelity:
I was thinking how nice it would be to use Domoticz with my current MQTT setup, without changing all my existing topic names to be Domoticz-centric when in fact my network is not. This patch enables custom topic prefixes with the MySensors MQTT gateway configuration.


gizmocuz:
@etherfi , thanks for the PR !

Some remarks:

Could you remove the do...while(0) part... that is not needed

It seems you are miss-using the CAFilename... Looking at the constructor of the class, i see there is also a 'Name' field, and a int Topics .... Could that be used instead ? Or could you use the 'Extra' field (see hardware table)
The CAFilename is used later in the class MQTT, this will cause conflicts.
Adding/Updating hardware via json supports the 'Extra' field, you also need to pass this to the constructor of MySensorsMQTT in MainWorker.cpp
(CWebServer::Cmd_AddHardware, field "extra")


Could you remove the leading space here data-i18n=" Enter
Before this span, add a html space (&nbsp;) please

Could you add the needed translations in www\i18n\domoticz-en.json.gz ?

EtherFidelity:
The do...while(0) is a programming technique that cleanly allows sections of code to be skipped without resorting to lengthy else-if blocks that can be more difficult to read.

http://edusagar.com/articles/view/6/useful-do-while0-trick

Actually, CAFilename is just for MQTT. There is no CAFilename variable outside of the MQTT system, and the variable is fully recovered upon loading, writing, etc.

The mode1 field (Topics) is useless because it's an int and would require a database schema change and massive refactoring to change.

Name is used all over the place.

The "Extra" field is for extra data relating to a piece of hardware, making it the perfect place to store this information.

I will add the translation information, but I don't see any reason to make any of these other changes. Am I wrong?

EtherFidelity:
By the way, the use of mode1, Mode1, model, and Model throughout the Hardware.js file is EXTREMELY confusing.

gizmocuz:
@etherfi , i don't think i ever mentioned mode1 ... i was talking about 'Extra' where you can store this information

As you already mention **"The "Extra" field is for extra data relating to a piece of hardware, making it the perfect place to store this information"**

It would be great if you can use the Extra field to store this information

EtherFidelity:
Ok, I see what you're saying. Actually change the name of the CAfilename parameter in the constructor for MySensorsMQTT so it reflects its content. (The data is actually stored in the Extras database field already.) 

I'll make that change as well.

EtherFidelity:
Hmmm turns out that the i18n translations for MySensors MQTT were never there in the first place. I added the relevant entries. My changes already store this data in the "Extra" field. Before this PR, the Extra field is used just to store the CAfilename information, and with this PR it is overloaded to additionally store the topic prefixes, using a # as a delimiter (a symbol which will never appear in a topic prefix).

gizmocuz:
@etherfi , Thanks for the language/translation addition... could you remove the extra text in the html files please,
when it is added in the i18n files, this could be written like:

`<span data-i18n="Topic Prefix Mode">Topic Prefix Mode</span>`

to

`<span data-i18n="Topic Prefix Mode"></span>`

(without the additional text from the data-i18n), also a great wat to find out if it is added in the translation file

EtherFidelity:
I just copied what I saw existing elsewhere in the same files. I don't understand what you want me to do.

gizmocuz:
@etherfi , seems part of my response was hidden because of some html tags, i updated the post above

EtherFidelity:
Extra text in the html files has been removed.

gizmocuz:
Thanks for all work and the PR!!

Issue #3357, opened 2019-06-28
lux of zwave device is muliplied by 10


NiekN:
im running now Version: 4.10935
I have 2 fibaro z-wave multisensors , and both LUX values are not correct anymore
When i look for the LUX value at the  control panel of the open zwave and compere this value which ade at domoticz the difference is a factor 10

rest of the values, like temperature looks ok

NiekN:
![image](https://user-images.githubusercontent.com/32562791/60316828-8ba6e880-996c-11e9-8049-0487e677d98a.png)

petergebruers:
Thanks. We're already working on this, fix is here:

https://github.com/domoticz/domoticz/pull/3341

NiekN:
OK thx for reply.
When can this be downloaded , so i can do an update for domoticz

petergebruers:
I'm not a Domoticz developer but I can tell you it very likely goes like this... @gizmocuz reviews/comments/accepts/rejects that PR. And after that, they can build a new beta version which can be downloaded from https://www.domoticz.com/downloads/

NiekN:
Issue is still not fixed at Version: 4.10949
or is #3341 not implementend yet in this version ????

gizmocuz:
@NiekN , 6 Lux is low... but which value is correct ?
Could you post here the node part of Config/ozwcache_xxx.xml ?
What is the min/max value ?

NiekN:
![lux](https://user-images.githubusercontent.com/32562791/60752244-609f4300-9fc3-11e9-9e0f-4d8a2ba2c82b.jpg)

petergebruers:
@gizmocuz there is a 2nd code path, give me a second

NiekN:
Sorry whcih node part should i send   , i cannot post the "ozwcache_0xe15eb478.xml" file

petergebruers:
```
case ZDTYPE_SENSOR_LIGHT:
		if (vType != OpenZWave::ValueID::ValueType_Decimal)
			return;
		if (vUnits != "lux")
		{
			//convert from % to Lux (where max is 1000 Lux)
			fValue = (1000.0f / 100.0f) * fValue;
		}
		pDevice->floatValue = fValue;
		break;
```

I would change that to ```if (vUnits == "%")``` like I did in "AddValue"

petergebruers:
@NiekN no need to send file

NiekN:
min max value of teh fibaro multi sensor ?
min = 0 
[...]

PR #3382, opened 2019-07-06
Fix Lux multiplied by 10 (Z-Wave Update Value)

petergebruers:
Should solve https://github.com/domoticz/domoticz/issues/3357


gizmocuz:
Thanks for the quick fix !

PR #3384, opened 2019-07-06
Revert "Activating a scene containing devices which are dimmable now correctly…"

gizmocuz:
Reverts domoticz/domoticz#3366


gizmocuz:
I reverted this PR... it should also not be necessary. You need to choose 'dimmer' type for the switch
As you can see in the code, it is considered a dimmer is the type is:

STYPE_Dimmer
STYPE_BlindsPercentage
STYPE_BlindsPercentageInverted
STYPE_Selector

AND the max dim level != 0

nhale25:
What type should I use for an RGB bulb? I was creating the device in a plugin like so:
Domoticz.Device("test", 1, Type=241, Subtype=4)
The plugin documentation implies that the Switchtype parameter is only valid if Type==244

gizmocuz:
@nhale25 , please use our forum.... if you add a dummy hardware and add a RGB light and look in your database backup then you will see the correct type

PR #3378, opened 2019-07-05
Added code to Enphase API to read (net) consumption counters from pro…

FrankvdAa:
…duction.json. The consumption counters are read all day in stead of only between sunrise and sunset (as is done for production readings).

As I only have a Metered version of the Envoy, I haven't been able to test against non-Metered Envoy, but I reckon it shouldn't cause any problems.

PR #3394, opened 2019-07-08
[EnphaseAPI] Bugfix on PR #3378: check if consumption data is present

FrankvdAa:
In case root["consumption"] is empty return and not write error entry to logs.

Addition to PR #3378.


gizmocuz:
Thanks!!

FireWizard52:
Thank you.
It works.
Problem Solved.

PR #3531, opened 2019-07-14
WebServerHelper: Fix compilation - deprecated APIs

neheb:
OpenSSL has initialization deprecated.


gizmocuz:
Thanks for the PR, should we not call 'OPENSSL_init_ssl' instead ?

neheb:
No need. OpenSSL does it internally.

gizmocuz:
Thanks !!!

PR #3547, opened 2019-07-26
Multi 1 wire busses

nielsvissers:
adding support for multiple 1wire busses by reading contents directory "/sys/bus/w1/devices"


gizmocuz:
Thanks !!!

Issue #3554, opened 2019-07-31
SolarEdge frequency reading displayed as percentage, instead of Hz


dheering:
SolarEdge inverter frequency is added as a percentage value with SendPercentageSensor(1 + iInverterNumber, SE_FREQ, 255, acFrequency, szTmp);.

Should be SendCustomSensor with "Hz" specified as unit?

https://github.com/domoticz/domoticz/blob/77a0b09a6d09881b573d4ec72171760fdf1baaf9/hardware/SolarEdgeAPI.cpp#L393

gizmocuz:
I think it was used like this because we did not had a custom sensor at that time
Sounds like a good idea, could you make a PR for this?
Please also use the database version number in sqlhelper.cpp to create a patch to convert the current percentage sensor to the custom sensor

dheering:
Never dabbled with this before, but I gave it a go. I forked the domoticz repository and cloned it to a Ubuntu 18.04 installation. Unfortunately, I ran into a lot of problems trying to get domoticz built. Especially the Boost stuff ;-(.

Also, changing SolarEdgeAPI.cpp is easy enough, but the SQLhelper.cpp looks quite daunting.

In short: I'd love to help, but this is too much of a learning curve for one evening ;-).

dheering:
After some RTFM, I got a Ubuntu VM with working domoticz now. Added a SolarEdge device and will allow it to run for some time. Then I will give modifying the code another go :-D

gizmocuz:
We have a wiki page about how to compile Domoticz.. 
It should be pretty straightforward...
You can also build and debug under windows or start a bash prompt there

dheering:
OK, I've been able to fix this issue, at least for my installation. Could not test for installations with multiple inverters, but I think it will work.

gizmocuz:
Thanks for this PR ! Merged!

Bertels999:
Hi, this issue also exists in the SBFSpot plugin. The frequency of my SMA invertor is dislayed as a percentage instead in Hz. See line 597 in SBFSpot.cpp : 

SendPercentageSensor((InvIdx * 10) + 2, 0, 255, percentage, "Hz"); 

![image](https://user-images.githubusercontent.com/122987321/217825795-d4931676-a175-4dd8-a879-ec6db7ff72c3.png)

gizmocuz:
I think because in that time of implementation we did not had custom sensors

You can change it to (pseudo code)

[...]

PR #3556, opened 2019-08-02
SolarEdge frequency reading displayed as percentage, instead of Hz #3554

dheering:
As reported in issue #3554: store Hz value from SolarEdge inverter as custom value. Update existing values by changing the corresponding DeviceStatus entry or entries.


gizmocuz:
Thanks you for this PR, looks great ! Great job!

PR #3558, opened 2019-08-03
Revert "Updating CMake configuration file to use version 3.14 as minimum"

ffontaine:
This partially reverts commit 4e4ee999c4b323514c0ba81ab1ff3afb10d2f1cd to avoid bumping cmake requirement version from 2.8.4 to 3.14. Indeed, it seems that domoticz builds fine with older cmake such as 3.8.2.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
-#set to minimum version that supports clean build
-cmake_minimum_required(VERSION 3.14.0)
+#set to minimum version that supports clean build on cygwin
+cmake_minimum_required(VERSION 2.8.4)
 
 project(domoticz)
 


gizmocuz:
@ffontaine , why would you work with an old CMake file ?
Did you test 3.8.2 against boost 1.70 ?

ffontaine:
I sent this PR because I'm the maintainer of domoticz (and boost) in buildroot (https://buildroot.org). Currently, we're enforcing a version of cmake greater or equal to 3.8 and I would like to avoid bumping this requirement to 3.14.
domoticz builds fine with cmake 3.8.2 and boost in version 1.70.0.

gizmocuz:
Thanks for the reply.
From what i know we need 3.14, also referenced here:

https://gitlab.kitware.com/cmake/cmake/issues/18865

Current CMake version is 3.15, i would bump your boost version to the most recent/stable version

ffontaine:
OK, thanks for your feedback, I better understand the issue. I disabled the cmake files provided by boost through the --no-cmake-config flag, that's why we don't need to bump our cmake version. I'll close this PR as it seems not relevant to merge it.

PR #3293, opened 2019-06-04
Quick fix for GCM not working anymore

Olifant1990:
GCM to FCM for android
https://github.com/domoticz/domoticz/issues/3292

PR #3592, opened 2019-08-30
Hotfix: Quick fix for GCM not working anymore (#3293) by Olifant1990

jsiegmund:
Requesting a hotfix on master for the GCM fix which is already included in develop. A number of people using the stable release are affected by broken GCM and imho this can be regarded as an active issue on the stable version, so it should be fixed for it to be stable again.

Also refer to https://www.domoticz.com/forum/viewtopic.php?t=28221

### DIFF
--- a/notifications/NotificationGCM.cpp
+++ b/notifications/NotificationGCM.cpp
@@ -5,7 +5,7 @@
 #include "../main/SQLHelper.h"
 #include "../json/json.h"
 
-#define GAPI_POST_URL "https://gcm-http.googleapis.com/gcm/send"
+#define GAPI_POST_URL "https://fcm.googleapis.com/fcm/send"
 #define GAPI "AIzaSyBnRMroiDaXCKbwPeOmoxkNiQfjWkGMre8"
 
 CNotificationGCM::CNotificationGCM() : CNotificationBase(std::string("gcm"), OPTIONS_NONE)


gizmocuz:
Can't you use the beta version?

gizmocuz:
If you really want this now, you could checkout the master branch and compile domoticz for yourself

jsiegmund:
I could use the beta version, but I rely heavily on zwave and as I've understood from the forums,  open zwave on the beta branch is not working great? And sure, I could compile myself but most users on the stable branch are probably there because compiling stuff themselves if not something they're comfortable with. So I thought I'd do some folks a favor by fixing this in stable.

But if this is not in line with your release policy then I guess it'll be broken for a while until there is a new stable release.

gizmocuz:
@jsiegmund , I think the OZW implementation in the beta should be pretty stable now, and when working, a lot better then the previous implementation.
Could you backup your whole domoticz folder and try this ?

odin-:
I've tried the beta version with this fix and it IS unstable. Crashes each time I try to turn on/off switches.

rwaaren:
I cannot replicate that behavior on my system (raspberry with debian stretch). for any of the Beta's created since the last stable Maybe best to create a topic for this on the forum and include your system , OS and version.

odin-:
OK well either way beta seems to be a bit unstable so the GCM hotfix could easily be merged into the stable baseline.

gizmocuz:
I can not replicate this issue. So hopefully @jsiegmund  could have a test as well ?

jsiegmund:
I've upgraded to beta. Not sure yet, I'll need to run it a bit longer to cast a serious vote. It hung on me once for now, had to restart the service to fix that. It also seems like a couple of zwave devices have lost all of their properties. I'm quite sure I had all kinds of settings like sensitivity and stuff for my PIR sensors, but now it only lists the wake up interval and nothing more than that. Maybe that'll take some time?

jsiegmund:
Sorry, but it's really not stable. I had it crash on me twice now. All of my double / triple click actions are dead, none of those work any more. Tried re-adding the trigger of one of the scenes but the triple click action doesn't come through. 

I also had to remove the battery from all of my window and door sensors for those to reconnect and a couple of built in wall dimmers have not returned to the network at all, so at the moment can only be operated using the hardwired switch (good to have those just in case). So I'll be going back to the back-up I made (is that a matter of stopping, putting it back in place and then restarting?).

GCM is working though... ;)

ameyniel:
Hi all.

Just giving a way to solve the problem on the STABLE version in a couple of minutes:
- backup the original "domoticz" binary to "domoticz-gcm" for example
- replace "https://gcm-http.googleapis.com/gcm/send" with "https://fcm.googleapis.com/fcm/send\0\0\0\0\0" (where each \0 is a byte value of zero)
- restart the Domoticz service
- and it just works!

You can do the binary replacement the way you want. I personally downloaded the file on Windows, used UltraEdit to make the change, and re-uploaded it to my Pi. Linux Gurus are welcome to give us the magic "sed" command ;-). What matters before everything is that the file is exactly the same size at the end of the process and a binary diff shows only the string changed.

Hope this helps.

jsiegmund:
Thanks @ameyniel! That works like a charm! 

For notepad++ users: to insert the byte value zero, open up the Character Panel under the "Edit" menu which allows for easy insert of byte values. Simply double click the value whilst your caret is in the place where you want it to go.

seansco:
@ameyniel You're a lifesaver!

annegerben:
thank you

PR #3602, opened 2019-09-04
add support for Belgian and (pre-decrypted) Luxembourgian P1 telegrams

gordonb3:
Following up on the prerequisite for Fluvius sharing their e-MUCS appendix on the DSMR P1 standard.


gizmocuz:
Thanks !!

PR #3603, opened 2019-09-04
missed Belgian gas meter in previous commit

gordonb3:
Gasmetering in Belgium uses an alternate 'tariff' indicator of 3. This is apparently related to the fact that Belgian law requires the meter's reading to be corrected for temperature whereas in The Netherlands the reading appears on the bill as is.


gizmocuz:
Thanks!

PR #3607, opened 2019-09-05
P1MeterBase: incorrect match on number value 1 instead of char value '1'

gordonb3:


gizmocuz:
Thanks! Speedy Gonzales 😁👍

PR #3614, opened 2019-09-08
OZW: Initial fan and operational state support

mooninite:
Fan state could be multiple values, but this initial support will only support
on/off states as I have only one hardware device to test against and it
only supports two states.

Operational state is always multiple values and is a new sub-type.

Tested against a Remotec ZTS-500US thermostat.


gizmocuz:
Great patch, great job !!!! Thank you for this !!!

gizmocuz:
Question, Should it be possible to also control the thermostat operating state or is it a read-only value ?
If it should be possible to operate, then it might be a idea so do a 'search all' for the string 'sTypeZWaveThermostatFanMode' in the code as this might help to implement it

mooninite:
Fan State and Operating State are read-only and report the active state of the thermostat. Domoticz already exposes the write values of Fan Mode and Operating Mode, which allow a user to set the desired mode of the thermostat.

Issue #3221, opened 2019-05-16
Openzwave cannot add values for slat en tilt position


Vinisz:
Although slat en tilt postitions can now be set in OZWCP, I'm unable to set them in domoticz.
Following is shown in the log when moving the blinds.

```
Error: OpenZWave: Value_Changed: Tried adding value, not succeeded!. Node: 72 (0x48), CommandClass: MANUFACTURER PROPRIETARY, Label: Venetian Blind slat position, Instance: 1, Index: 0
2019-05-16 09:56:34.148 Error: OpenZWave: Value_Changed: Tried adding value, not succeeded!. Node: 72 (0x48), CommandClass: MANUFACTURER PROPRIETARY, Label: Venetian blind tilt position, Instance: 1, Index: 1
```

Version: 4.10763
Build Hash: 792e9a60
Compile Date: 2019-05-16 09:00:25
dzVents Version: 2.4.20
Python Version: 3.5.3 (default, Sep 27 2018, 17:25:39) [GCC 6.3.0 20170516]

gizmocuz:
Could you show a screenshot from the control panel ?
The should also be a warning/error for an unknown command class before this ?
Could you post your cache file ? (config folder, zwcache_xxxx.xml)
Please Select the controller (node 1), enable debugging, and disable/enable the hardware
Set the position a few times like 100%, 0%, and 50%
Then copy the log file (Config/OZW_Log.xml) and post this as well
Then select the controller (node 1) and disable debug logging, and disable/enable the hardware again

Vinisz:
Ok, I can do that... but what exactly do you mean with “disable/enable the hardware” ?

gizmocuz:
@Vinisz , from the hardware page select the zwave hardware, disable it, press update, then enable it, press update
This forces a restart of the zwave hardware

Vinisz:
Ok, I think I am able to do this later today

Vinisz:
Hi, I put everything in a zip, hope this helps and is enough....
For in the log, I did move the blinds as well as the slats positions to some different values from the OZWCP. please note that I am able to control both from the OZWCP just fine, I am only missing the ability to control the slats position from domoticz itself, if you can add that it would be really awesome, for me, it is about the only thing I am missing now...

[Venetian blinds case.zip](https://github.com/domoticz/domoticz/files/3204333/Venetian.blinds.case.zip)

svde:
[...]

PR #3620, opened 2019-09-11
Add Logging to SendSwitchIfNotExists if devices with same DeviceID

petergebruers:
Example device is FGRM-222 having both "binary" and "multilevel" switch, log example:

Error: SendSwitchIfNotExists: Not adding '4.instance.1.index.0.commandClasses.38' because database already has DeviceID '00000401'. That ID matches '4.instance.1.index.0.commandClasses.37'

Helps diagnosing https://github.com/domoticz/domoticz/issues/3221

### DIFF
--- a/hardware/ZWaveBase.cpp
+++ b/hardware/ZWaveBase.cpp
@@ -234,7 +234,28 @@ void ZWaveBase::SendSwitchIfNotExists(const _tZWaveDevice* pDevice)
 		result = m_sql.safe_query("SELECT ID FROM DeviceStatus WHERE (HardwareID==%d) AND (Unit==%d) AND (Type==%d) AND (SubType==%d) AND (DeviceID=='%q')",
 			m_HwdID, int(unitcode), pTypeGeneralSwitch, sSwitchGeneralSwitch, szID);
 		if (!result.empty())
+		{
+			std::string first_non_matched_string_id;
+			std::map<std::string, _tZWaveDevice>::iterator itt;
+			for (itt = m_devices.begin(); itt != m_devices.end(); ++itt)
+			{
+				if (
+					(itt->second.nodeID == pDevice->nodeID) &&
+					(itt->second.instanceID == pDevice->instanceID) &&
+					(itt->second.string_id != pDevice->string_id))
+				{
+					first_non_matched_string_id = itt->second.string_id;
+				}
+			}
+			if (!first_non_matched_string_id.empty())
+
+				_log.Log(
+					LOG_ERROR,
+					"SendSwitchIfNotExists: Not adding '%s' because database already has DeviceID '%s'. That ID matches '%s'",
+					pDevice->string_id.c_str(), szID, first_non_matched_string_id.c_str());
+
 			return; //Already in the system
+		}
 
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thanks for the PR. I would use a 'status' type instead of a error, and also add a 'test' or something in the log string.
This because you will always experience this if you restart Domoticz, as when 'AddValue' is received when starting OpenZWave, all switches where previously added

petergebruers:
> This because you will always experience this if you restart Domoticz, as when 'AddValue' is received when starting OpenZWave, all switches where previously added

I took care of that. The code only logs if the device is not unique.

```
if (
    (itt->second.nodeID == pDevice->nodeID) &&
    (itt->second.instanceID == pDevice->instanceID) &&
    (itt->second.string_id != pDevice->string_id))
{
    first_non_matched_string_id = itt->second.string_id;
}
```

If the node & instance point to a single device, string_id is unique, no logging happens.

> Thanks for the PR. I would use a 'status' type instead of a error

I think you wanted me to change that to "status" because you suspected this code would print a bunch of lines after restart, but it does not...

On my 10 node network, with your FGRM-222 code active, this is really all it prints:

```
2019-09-11 16:42:40.895 OpenZWave: Value_Added: Node: 12 (0x0c), CommandClass: CENTRAL SCENE, Label: Scene 12, Instance: 1, Index: 12
2019-09-11 16:42:40.895 OpenZWave: Value_Added: Node: 12 (0x0c), CommandClass: BATTERY, Label: Battery Level, Instance: 1, Index: 0
2019-09-11 16:42:40.648 Status: OpenZWave: ManufacturerSpecificDB Ready
2019-09-11 16:42:40.874 Status: OpenZWave: Driver Ready
2019-09-11 16:42:40.879 Error: SendSwitchIfNotExists: Not adding '4.instance.1.index.0.commandClasses.38' because database already has DeviceID '00000401'. That ID matches '4.instance.1.index.0.commandClasses.37'
2019-09-11 16:42:40.882 Error: SendSwitchIfNotExists: Not adding '4.instance.1.index.1.commandClasses.145' because database already has DeviceID '00000401'. That ID matches '4.instance.1.index.2.commandClasses.50'
2019-09-11 16:42:43.102 (z) Thermostat (Heating 1)
2019-09-11 16:42:46.383 (z) Temp (Temperature)
2019-09-11 16:42:46.473 (z) General/kWh (kWh Meter)
2019-09-11 16:42:46.554 (z) Usage (Unknown)
```

It is 2 lines and those 2 really diagnose a fundamental issue: the first one tells the (normal) position (multilevel blind position) is not added. The second tells the "Tilt" value was not added.

gizmocuz:
Thanks, I see! 
The solution is also easy, but my problem is that current users will loose all there actual switches/timers/scripts 

As ID we (could/)should use <node_id><command_class><instance><index>

At the moment it is <00><YYYY><node_id><instance>

Where YYYY is actually always 0, but it;s actually (some bad/old code...) (node_id & 0xFF00) << 8 

Indexes > 255 are not used as a sensor type

I simplified the function a bit in the latest commit

petergebruers:
Thanks!

PR #3623, opened 2019-09-13
Update TeleinfoBase.cpp

darrepac:
Remove to not update the PAPP value if similar to previous (was causing some values to never update)


gizmocuz:
Thanks!! Beta building (11316) could take up 2 more then an hour

darrepac:
> Thanks!! Beta building (11316) could take up 2 more then an hour

I do not see the change in the changelog (https://releases.domoticz.com/releases/beta/history_linux_armv7l.txt). Is it in?

gizmocuz:
Check the github repository

darrepac:
I confirm that the bug is solved with this change.

PR #3624, opened 2019-09-13
Add Support for Fibaro FGRM-222 Roller Shutter

### LINKED ISSUE #3221: Openzwave cannot add values for slat en tilt position

petergebruers:
Should fix #3221

On an existing database, the first (and only) switch device changes to ON/OFF because the FGRM-222 exposes this first. This supports full open-close in all Roller Shutter Modes.

Three new devices are created, IDs end in 10, 11 and 12
10 Level -> Used to set position in Gate, Blind and Venetian Blind mode marked "with positioning" (see parameter 10 "Roller Shutter operating modes")
11 Venetian Blind slat position -> Similar to 10 Level
12 Venetian blind tilt position -> as the name says

For Tilt reporting, set Parameter 3 "Reports type" to 1 - Blind position reports sent to the main controller using Fibar
Command Class.

I checked my FGRM-222 and can control my FGRM in both Blinds and Venetian Blinds mode.

However, I had to check with ZNiffer: the device has its own mind about what to report back, sometimes it reports on ID ...10, sometimes ...11 or both. In Blinds mode, Tilt also changes reported position. This behaviour might be specific to my hardware revision and firmware.

The code is relatively simple but one warning, although I tested this, this line is semantically different and might have an impact.

Old:
if (GetValueByCommandClassIndex(pDevice->nodeID, instanceID, COMMAND_CLASS_SWITCH_MULTILEVEL, ValueID_Index_SwitchMultiLevel::Level, vID) == true)

New:
if (GetValueByCommandClassIndex(pDevice->nodeID, pDevice->orgInstanceID, pDevice->commandClassID, pDevice->orgIndexID, vID) == true)

The change is needed because the FGRM Multilevel Venetian Blind switch devices use CC "proprietary" and orgInstanceID encodes the real instance for OZW (while IndexID is 10, 11 or 12).

### DIFF
--- a/hardware/OpenZWave.cpp
+++ b/hardware/OpenZWave.cpp
@@ -1207,13 +1207,7 @@ bool COpenZWave::SwitchLight(_tZWaveDevice* pDevice, const int instanceID, const
 
 		if ((pDevice->devType == ZWaveBase::ZDTYPE_SWITCH_NORMAL) || (bHandleAsBinary))
 		{
-			//On/Off device
-			bool bFound = (GetValueByCommandClass(pDevice->nodeID, instanceID, COMMAND_CLASS_SWITCH_BINARY, vID) == true);
-			if (!bFound)
-				bFound = (GetValueByCommandClass(pDevice->nodeID, instanceID, COMMAND_CLASS_DOOR_LOCK, vID) == true);
-			if (!bFound)
-				bFound = (GetValueByCommandClassIndex(pDevice->nodeID, instanceID, COMMAND_CLASS_SWITCH_MULTILEVEL, ValueID_Index_SwitchMultiLevel::Level, vID) == true);
-			if (bFound)
+			if (GetValueByCommandClassIndex(pDevice->nodeID, pDevice->orgInstanceID, pDevice->commandClassID, pDevice->orgIndexID, vID))
 			{
 				OpenZWave::ValueID::ValueType vType = vID.GetType();
 				_log.Log(LOG_NORM, "OpenZWave: Domoticz has send a Switch command! NodeID: %d (0x%02x)", pDevice->nodeID, pDevice->nodeID);
@@ -1276,7 +1270,7 @@ bool COpenZWave::SwitchLight(_tZWaveDevice* pDevice, const int instanceID, const
 			//dimmable/color light  device
 			if ((svalue > 99) && (svalue != 255))
 				svalue = 99;
-			if (GetValueByCommandClassIndex(pDevice->nodeID, instanceID, COMMAND_CLASS_SWITCH_MULTILEVEL, ValueID_Index_SwitchMultiLevel::Level, vID) == true)
+			if (GetValueByCommandClassIndex(pDevice->nodeID, pDevice->orgInstanceID, pDevice->commandClassID, pDevice->orgIndexID, vID) == true)
 			{
 				std::string vLabel = m_pManager->GetValueLabel(vID);
 
@@ -1312,33 +1306,6 @@ bool COpenZWave::SwitchLight(_tZWaveDevice* pDevice, const int instanceID, const
 					}
 				}
 
-			}
-			else if (GetValueByCommandClassIndex(pDevice->nodeID, instanceID, COMMAND_CLASS_PROPRIETARY, ValueID_Index_SwitchMultiLevel::Level, vID) == true)
-			{
-				if (
-					(pNode->Manufacturer_id == 0x010F)
-					&& (pNode->Product_id == 0x1000)
-					&& (pNode->Product_type == 0x0302)
-					)
-				{
-					//Fibaro FGRM222 Roller Shutter Controller 2
-					//Slat/Tilt position
-					pDevice->intvalue = svalue;
-					_log.Log(LOG_NORM, "OpenZWave: Domoticz has send a Slat/Tilt command!, Level: %d, NodeID: %d (0x%02x)", svalue, pDevice->nodeID, pDevice->nodeID);
-
-					if (vID.GetType() == OpenZWave::ValueID::ValueType_Byte)
-					{
-						if (!m_pManager->SetValue(vID, (uint8_t)svalue))
-						{
-							_log.Log(LOG_ERROR, "OpenZWave: Error setting Slat/Tilt Value! NodeID: %d (0x%02x)", pDevice->nodeID, pDevice->nodeID);
-						}
-					}
-					else
-					{
-						_log.Log(LOG_ERROR, "OpenZWave: SwitchLight, Unhandled value type: %d, %s:%d, NodeID: %d (0x%02x)", vID.GetType(), std::string(__MYFUNCTION__).substr(std::string(__MYFUNCTION__).find_last_of("/\\") + 1).c_str(), __LINE__, pDevice->nodeID, pDevice->nodeID);
-						return false;
-					}
-				}
 			}
 			else
 			{
@@ -1679,6 +1646,15 @@ void COpenZWave::AddValue(NodeInfo* pNode, const OpenZWave::ValueID& vID)
 		{
 			_log.Log(LOG_ERROR, "OpenZWave: Unhandled value type: %d, %s:%d, NodeID: %d (0x%02x)", vType, std::string(__MYFUNCTION__).substr(std::string(__MYFUNCTION__).find_last_of("/\\") + 1).c_str(), __LINE__, NodeID, NodeID);
 			return;
+		}
+
+		if (IsFibaroFgrm222(_device))
+		{
+			// Handling Fibaro FGRM-222 Blind Binary Switch.
+			// Cannot use vOrgIndex or vOrginstanceID as instanceID because this overlaps with
+			// Domoticz device created for Switch Multilevel (in ZWaveBase::SendSwitchIfNotExists)
+			// Pick an instanceID, something that looks nice as a device ID in the browser and does not conflict...
+			_device.instanceID = 0x10;
 		}
 		InsertDevice(_device);
 	}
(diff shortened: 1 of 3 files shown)


gizmocuz:
Thanks for the PR!!
I think this will go wrong where nodes have both commandclass 37 and 38, where 37 is reported first (added first into the system), then 38 (which will fail), but the node is actually a dimmer.
This way the 'level' van not be set anymore
(There was one user reporting this error, still have to find out what node hardware/type he/she has)

Another way is to check if it is the Fibaro Fgrm222, then call GetValueByCommandClassIndex with the orgInstance/index, else use the original check (and against COMMAND_CLASS_SWITCH_MULTILEVEL first)

Hope that makes sense... 

I am able to test this probably next week (Thursday)

petergebruers:
> I think this will go wrong where nodes have both commandclass 37 and 38, where 37 is reported first (added first into the system), then 38 (which will fail), but the node is actually a dimmer.

In fact, My PR fixes the problem you describe for the FGRM-222 but does not introduce any changes for other devices.

I think in fact this workaround should be applied to all devices having commandclass 37 and 38.

> I am able to test this probably next week (Thursday)

Thanks. Testing is important!

It is related to this:

https://github.com/domoticz/domoticz/commit/e663f6f002f49ac89a953ab9de437ea0026345c9#commitcomment-35075479

I am going to look at that logging I introduced as promised, it probably false triggers in some cases like the "Alarm" we have seen. But this case is about an RGBW device and it probably exposes Dimmer(s) which may not get created (like Position wasn't created for a Roller Shutter because of the ON/OFF switch).

petergebruers:
@gizmocuz 

> I think this will go wrong where nodes have both commandclass 37 and 38, where 37 is reported first (added first into the system)

Well, that is an existing problem, and my PR tries to solve that for the FGRM specifically...

I doubled checked and my changes are really specific because of the "if (IsFibaroFgrm222(_device))"

```cpp
else if (commandclass == COMMAND_CLASS_SWITCH_MULTILEVEL)
{
    if (vOrgIndex == ValueID_Index_SwitchMultiLevel::Level)
    {
        if (vType == OpenZWave::ValueID::ValueType_Byte)
        {
            if (m_pManager->GetValueAsByte(vID, &byteValue) == true)
            {
                _device.devType = ZDTYPE_SWITCH_DIMMER;
                _device.intvalue = byteValue;

/***** UP TO HERE all code is existing code *****/

                if (IsFibaroFgrm222(_device))
                {
                    // Handling Fibaro FGRM-222 Blind Position.
                    // Cannot use vOrgIndex or vOrginstanceID as instanceID because this overlaps with
                    // Domoticz device created for Switch Binary (in ZWaveBase::SendSwitchIfNotExists)
                    // Pick an instanceID, something that looks nice as a device ID in the browser and does not conflict...
                    _device.instanceID = 0x10;
                }
```
Because of that test it does not affect existing code.

In fact, it might be interesting to apply that  _device.instanceID = 0x10; (or some counter if the device has multiple MultiLevel switches to maken them unique) to other devices types as well. To be tested...

petergebruers:
Ok, digging through te existing code after simplifying the void ZWaveBase::SendSwitchIfNotExists(const _tZWaveDevice* pDevice)

It is a bit hard to follow for my poor brain :)

My main question is: what makes a UNIQUE record in the database. Because that determines when a certain combination of node IC, CC, ZD_TYPE leads to conflicts.

There are 2 places to look at, we also have to look at how a Domoticz Device translates back to a Z-Wave device, that has to be kind of "the reverse operation of creating a database record). That is in bool ZWaveBase::WriteToHardware(const char* pdata, const unsigned char length)

But first the creation of a switch, code snippets + comment

This query should lead to one and only one Z-Wave ValueID:

```cpp
result = m_sql.safe_query("SELECT ID FROM DeviceStatus WHERE (HardwareID==%d) AND (Unit==%d) AND (Type==%d) AND (SubType==%d) AND (DeviceID=='%q')",
	m_HwdID, int(unitcode), devType, SubType, szID);
```

m_HwdID comes from the Hardware driver (in my case ID 2 pointing to OpenZWave driver)
int(unitcode) is hardcoded to be 1, it is never changed in SendSwitchIfNotExists

So the interesting parts:

devType
SubType
szID

It is possible to have devices with the same szID, as long as devType and SubType differ

Where do these come from?

We have been discussing szID, its definition cannot be changed easily to be "more unique" by adding for instance CC because that would lead to issues woth existing records. So it is: 

```cpp
unsigned char ID1 = 0;
unsigned char ID2 = (unsigned char)((pDevice->nodeID & 0xFF00) >> 8);
unsigned char ID3 = (unsigned char)pDevice->nodeID & 0xFF;
unsigned char ID4 = pDevice->instanceID;
```

That is quite OK to understand. It means a device with 1 instance ID but having a "Switch" and a "Dimmer" will have the same ID. But devType and SubType can make the database entry unique, this depends on the definition of those variable values:



```cpp
int devType = pTypeGeneralSwitch;
int SubType = sSwitchGeneralSwitch;
```

Those are defines in domoticz/hardware/hardwaretypes.h

```cpp
//General Switch
#define pTypeGeneralSwitch			0xF4
#define sSwitchTypeX10				0x00
(...)
#define sSwitchGeneralSwitch		0x49
(...)
```

So they start as a constant and the names suggest "a general switch"

Then:

if ((pDevice->devType == ZDTYPE_SWITCH_RGBW) || (pDevice->devType == ZDTYPE_SWITCH_COLOR))
{
	devType = pTypeColorSwitch;
	// TODO: For devices of type ZDTYPE_SWITCH_COLOR the supported color channels is reported in response to command SWITCH_COLOR_SUPPORTED_GET
	// This is already done by OpenZWave::Color class and used internally by the class, the information is needed here.
	SubType = (pDevice->devType == ZDTYPE_SWITCH_RGBW) ? sTypeColor_RGB_W_Z : sTypeColor_RGB_CW_WW_Z;
}
```

Specifically for ZDTYPE_SWITCH_RGBW and ZDTYPE_SWITCH_COLOR devType and SubType are modified. This means the same sZID can now refer to a "pTypeGeneralSwitch/sSwitchGeneralSwitch" and a "pTypeColorSwitch/sTypeColor_RGB_W_Z" and a "pTypeColorSwitch/sTypeColor_RGB_CW_WW_Z" - because that leads to unique devType, SubType, szID comination used in that SQL statement.

Interestingly enough, in this part of the code there is also:

```cpp

int SwitchType = (
	(pDevice->devType == ZDTYPE_SWITCH_DIMMER)
	|| (pDevice->devType == ZDTYPE_SWITCH_RGBW) 
	|| (pDevice->devType == ZDTYPE_SWITCH_COLOR)
	) ? STYPE_Dimmer : STYPE_OnOff;
```

This time, SwitchType uses an enum defined in domoticz/main/RFXNames.h

enum _eSwitchType
{
	STYPE_OnOff = 0,					//0
	STYPE_Doorbell,					//1
	STYPE_Contact,					//2
	STYPE_Blinds,					//3
	STYPE_X10Siren,					//4
	STYPE_SMOKEDETECTOR,			//5
	STYPE_BlindsInverted,			//6
(...)
	STYPE_DoorLockInverted,					//20
```

The sole purpose seems to be cosmetic:

```cpp
//Set SwitchType
m_sql.safe_query("UPDATE DeviceStatus SET SwitchType=%d WHERE (HardwareID==%d) AND (Unit==%d) AND (Type==%d) AND (SubType==%d) AND (DeviceID=='%q')",
	SwitchType, m_HwdID, int(unitcode), devType, SubType, szID);
```

The database field "SwitchType" is not used in the select query which determines if the device is in the database. If it did, we would have more possibilites to make a database record unique but there are probably good reasons for not doing that. I have only been studying the Domoticz source code recently so I do not understand everything and I certainly do no have any historical perspective.

Preliminary conclusions:

- It is not possible to have a dimmer and a switch with same instanceID, but it is possible to have for example a switch and a pTypeColorSwitch

- STYPE_Blinds is not used in OpenZWave, nor is STYPE_BlindsPercentage. I do find code in domoticz/hardware/XiaomiGateway.cpp using this when model == "curtain". I am not sure adding this to FGRM-222 would be easy because it can control blinds, venetian blinds or a gate depending on a parameter.

- My "duplicate detection" code will give false positives because I do not filter out non-switch like Z-Wave objects. I'll have to fix that.

- I really need to look at ZWaveBase::WriteToHardware to make sure data from the database is matched against the correct Z-Wave object (calling ZWaveBase::FindDevice and that one loops through m_devices).

petergebruers:
Digging through more code... The InstanceID trick idea predates this PR...

In AddValue line 1685:

```cpp
	else if (commandclass == COMMAND_CLASS_COLOR_CONTROL)
	{
		if (vOrgIndex == ValueID_Index_Color::Color)
		{
			if (vType == OpenZWave::ValueID::ValueType_String)
			{
				_device.label = "Color";
				_device.devType = ZDTYPE_SWITCH_COLOR;
				_device.instanceID = 101;
				InsertDevice(_device);
			}
		}
	}
```

Explains why Color CC device Is end in ...65

gizmocuz:
A unique device is HWID, DeviceID,Unit,DevType,SubType

petergebruers:
Exactly. That is how I understood it. The FGRM needs two (three if you count the on/of blind switch) extra switches to support VB mode. In this PR I chose to make DeviceID unique by manipulating InstanceID, like the color cc handling does. Correct me if I am wrong, but there are no 100% suitable DevType,SubType to represent slat pos and tilt. I did see some constant for blinds though, I did not try them. But if you know which ones could be used, that is an alternative aproach. I am unable to test this week, I do not have a device available.

gizmocuz:
@petergebruers , sorry about the delay, is this patch still needed? There is a git merge issue

petergebruers:
I see. I cannot answer that question right now because I do not have a test device at hand... To be honest, I do not need Venetian Blind mode, I wrote this patch to help the user that reported the issue... Maybe you can check with the OP in #3221 if it's OK now? If this patch is still needed, I can have a look at it next week but I cannot fix a date yet.

petergebruers:
Rebased, and builds on my Mac. Tested with a single device, a Neo NAS-SC01Z (Sim) - not an FGRM222. Please note subtle but important change on line 1269 of OpenZWave.cpp

petergebruers:
Don't merge yet, although it works as intended... Because I would like to try one more improvement.

I wrote: "On an existing database, the first (and only) switch device changes to ON/OFF because the FGRM-222 exposes this first. This supports full open-close in all Roller Shutter Modes."

It would be much nicer if I could keep the first device as "Position" (multilevel) for existing users, that would cause less confusion. It is a bit tricky though, I have to work with "orgInstanceID" and change the matching rules in `bool COpenZWave::SwitchLight(_tZWaveDevice* pDevice, const int instanceID, const int value)`. I did that for multilevel devices, but I have to add it for binary devices as well. The other option is to simply not add the ON/OFF switch at all. I'll have to investigate and test.

petergebruers:
I was able to keep the first device as "Position" but (unfortunately) that introduces a change in the handling of switches, see OpenZWave.cpp line 1210, similar to the one I already need for Venetian Blind mode (line 1279 old 1273 new).

I've tested this PR on a 10 node production network, that is the best I can do... I don't have a ton of test devices or Z-Wave networks...

I would like to have more people test it... I'd like to post a "call for help" on the forum but I have no experience with this... Do you think it stands any chance of attracting people willing to build and test my fgmr222 branch?

gizmocuz:
Maybe we should wait until the crash has been solved that i reported in a OZW issue and the query stages complete, then make a new stable release for Domoticz, and then i am going to rebuild the whole internal Domoticz device ID generation ?
This should also solve the problem...
But as that is a big change, and i want to try to have a painless migration, i think it is best to bring out a stable version first.

petergebruers:
Okay. I understand your priorities. I will keep a copy of this PR locally then close this one.

petergebruers:
BTW not related to this PR but

> Maybe we should wait until the crash has been solved that i reported in a OZW issue

For the record, that is about the "interview order" causing OZW to access non-existent ValueIDs and that is being worked on by Fishwaldo only... Sorry I cannot help with that one, it is kind of messy and the task cannot be easily split into parts.

gizmocuz:
Thank you for the PR and your time as well ! Hope we can merge this soon

petergebruers:
It was very educational, finding out how all the pieces of the puzzle fit together. :) I mean the glue between a OZW Value and a Domoticz device.

Issue #3627, opened 2019-09-14
The Alarm aka Notification Class needs work to get 1.6 compliant


petergebruers:
I recently introduced extra logging in the SendSwitchIfNotExists function. After this users posted interesting issues with alarm CC.

For example, see:

https://github.com/domoticz/domoticz/commit/e663f6f002f49ac89a953ab9de437ea0026345c9#commitcomment-35082930

The gist: "Not adding '11.instance.2.index.512.commandClasses.113' (i2 Alarm Type) "

So I added an FGMS-001 Fibaro Motion Sensor to my test network because it supports Notification CC...

It wants to add "7.instance.1.index.256.commandClasses.113"

```cpp
unsigned char ID1 = 0;
unsigned char ID2 = (unsigned char)((pDevice->nodeID & 0xFF00) >> 8); -----> 0
unsigned char ID3 = (unsigned char)pDevice->nodeID & 0xFF; -----> 7
unsigned char ID4 = pDevice->instanceID; -----> instanceID is 256 but ID4 is 0, NOK!
```

instanceID > 255 -----> BOOM!

Why 256? In AddValue

```cpp
int newInstance = GetIndexFromAlarm(vLabel);
if (newInstance == 0)
	newInstance = vOrgIndex;
_device.instanceID = newInstance;
```
Errrm...call to GetIndexFromAlarm is OZW 1.4 Stuff... Not good.

Index 256 means "Previous Event Cleared".

That code won't fly :wink:

I will submit a patch make the error message much clearer and at the same time reduce false positive logging.

```cpp
	if ((pDevice->instanceID > 255) || (pDevice->instanceID < 0))
	{
		_log.Log(
			LOG_ERROR,
			"SendSwitchIfNotExists: Not adding '%s' (%s) because instanceID %d does not fit in 8 bits",
			pDevice->string_id.c_str(), pDevice->label.c_str(), pDevice->instanceID);
		return;
	}
```

This will cause logging similar to this

```
Error: SendSwitchIfNotExists: Not adding '3.instance.1.index.256.commandClasses.113' (Previous Event Cleared) because instanceID 256 does not fit in byte
```

Of course you would rather want me to fix the handling of the Notification CC instead of logging something... I might do that, if I find the time.

gizmocuz:
Maybe we should drop all instances with value 256  ? That also solve the automagic clears?

petergebruers:
[...]

PR #3629, opened 2019-09-15
OZW, Update detection of duplicate/invalid DeviceID

petergebruers:
1) Improves message in case of instanceID > 255

See https://github.com/domoticz/domoticz/issues/3627

Example log "Error: SendSwitchIfNotExists: Not adding '3.instance.1.index.256.commandClasses.113' (Previous Event Cleared) because instanceID 256 does not fit in byte"

2) Change "detection logic" of duplicate devices, avoids possible false positives.

3) Change logging level from ERROR to STATUS as per request of @gizmocuz

Example of a real message, caused by FGRM-222

"SendSwitchIfNotExists: Device '4.instance.1.index.0.commandClasses.38' (Level) with DeviceID '00000401' matches '4.instance.1.index.0.commandClasses.37' (Switch). Domoticz will use the Dimmer (and hide the Switch)."

The rationale behind that message is... this is the "matching" code in Domoticz:

//First find dimmer
pDevice = FindDevice(nodeID, instanceID, indexID, ZDTYPE_SWITCH_DIMMER);

Then retry with ZDTYPE_SWITCH_NORMAL

### DIFF
--- a/hardware/ZWaveBase.cpp
+++ b/hardware/ZWaveBase.cpp
@@ -164,6 +164,15 @@ void ZWaveBase::SendSwitchIfNotExists(const _tZWaveDevice* pDevice)
 	//ID4 = pDevice->indexID&&0x00FF;
 	//but current users gets new devices in this case
 
+	if ((pDevice->instanceID > 255) || (pDevice->instanceID < 0))
+	{
+		_log.Log(
+			LOG_STATUS,
+			"SendSwitchIfNotExists: Not adding '%s' (%s) because instanceID %d does not fit in byte",
+			pDevice->string_id.c_str(), pDevice->label.c_str(), pDevice->instanceID);
+		return;
+	}
+
 	unsigned char ID1 = 0;
 	unsigned char ID2 = (unsigned char)((pDevice->nodeID & 0xFF00) >> 8);
 	unsigned char ID3 = (unsigned char)pDevice->nodeID & 0xFF;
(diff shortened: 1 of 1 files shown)


gizmocuz:
@petergebruers , going to checking some changes that will adopt the status change, but checking the instanceID >255 < 0, does not work anymore because it's a uint8_t
Keep you posted here

gizmocuz:
@petergebruers , I added the change in this commit

https://github.com/domoticz/domoticz/commit/3ace1a6c940eebdfb73751cf593f8907cf3b02a9

petergebruers:
Im not checking instanceID but pDevice->instanceID

That is an int and the debugger 100 % for sure tells me it is 256

Domitcz then does `unsigned char ID4 = pDevice->instanceID;`which turns 256 into 0

gizmocuz:
@petergebruers , have a look at the changes first please, 'instance' is 8 bit now (also in OZW)

petergebruers:
I'll have to add flags to compile C++14 standard because it does not compile on my Mac yet, error "'auto' not allowed in lambda parameter"

So this is your solution to handle Index=256?

/Users/peter/ozw/domoticz/hardware/OpenZWave.cpp line 2544
```
		instance = GetIndexFromAlarm(vLabel);
		if (instance == 0)
			instance = (uint8_t)vOrgIndex;
```

So 256 -> 0 ?

gizmocuz:
@petergebruers , most be the lambda calls... I will have a check for this, thanks for reporting.

No this is not a solution to handle index >=256 at all, this is part one of correcting the code, and in the near feature changing the DeviceID's for the ZWave switches

rkluwen:
If instance would be an unsigned char, then yes, this is a solution. Just me 0.5 cents.

gizmocuz:
@rkluwen , yes, but hmmm there is more then 256, see NotificationCCTypes.xml and ValueIDIndexesDefines.h

gizmocuz:
At the moment the Alarm is handled as both switch (on/off) and as pTypeGeneral/sTypeZWaveAlarm
Best to delete the support for the on/off switch, but this needs more work in the correct handling/logging/scripting of pTypeGeneral/sTypeZWaveAlarm

petergebruers:
Indeed,  I don't wonder about the cast. But the point is indeed vOrgIndex can be 256 or 512 or 513... You turn that into 8 bits sure but...

It is not "hypothetical" - an FGMS-001 does create such an index and I've seen similar posts by GravityRZ

eg 7.instance.1.index.256.commandClasses.113

petergebruers:
> At the moment the Alarm is handles as both switch (on/off) and as pTypeGeneral/sTypeZWaveAlarm
> Best to delete the support for the on/off switch, but this needs more work in the correct handling/logging/scripting of pTypeGeneral/sTypeZWaveAlarm

I have to check that, I have not checked that "pTypeGeneral/sTypeZWaveAlarm" you've mentioned. I need more time, I am a slow old man 😉

gizmocuz:
@petergebruers , so am I ;) ... project compiles again...

gizmocuz:
@petergebruers , I think these '256' messages are from the automagic clears correct?
Maybe the best way is to find out what needs to be cleared, and clear this (value/device), and not add a switch/device with 256 in the first place, as it is a 'clear' for something

petergebruers:
You are on the right track but it will become clearer after I have finished writing my report about handling this as promised in https://github.com/domoticz/domoticz/issues/3627

Very short answer: on the FGMS (and other devices as well but that is the one I use for testing) that ValueID tells you which one of the 2 possible alarms got clear.

The FGMS-001 can independently report Tamper (= shake the device) and Motion (= trigger the PIR). 

For example, OZW might log this when you pick up the device:

Class ALARM Instance 1 Index 7 Type list ValueAsString Motion Detected at Unknown Location
Class ALARM Instance 1 Index 7 Type list ValueAsString Tampering -  Cover Removed

Both alarms have independent "clear" signals. Eg you can shake the device, by default it will clear the tamper alarm after 30 s, but if you keep moving in front of it, the motion keeps "active" (does not clear). There are also parameters to set the timing of both alarms independently.

When you get a clear, you have to examine the contents of the Value with Index 256. Because the "Clear" packet tells you, with an extra byte, which alarm got cleared.

Class ALARM Instance 1 Index 256 Type byte ValueAsString 3 -> The three means "The Tamper alarm"

Class ALARM Instance 1 Index 256 Type byte ValueAsString 8 -> The eight means "Motion"

OpenZWave sets that index, prior to issuing a "clear", in the log that is:

Class ALARM Instance 1 Index 7 Type list ValueAsString Clear

This is not an example of "auto clear", quit the opposite, the device clearly signals "begin" and "end" of a an alarm condition. Some older device only signal a "begin". I'll try to find an example.

PR #3632, opened 2019-09-19
 Update to dzVents V2.4.29 (add setLevel, corrected typos and fixed small bugs)

rwaaren:


gizmocuz:
Thanks for the changes !
It might be an idea to not remove any failing checks in feature commits as the need to be solved.
If the PR was merged before the 'no trigger' fix, we would never have known that it was still not working.
If there are more checks that need to be solved, please enable them...

rwaaren:
@gizmocuz Thx for your comment. 
From the conversation we had on this issue, it was not completely clear to me if it was your intention to re-enable the option to use NOTRIGGER for sensor type devices. I agreed with Danny to comment that particular test until it was fixed or clarified what the future of this option was. On a next a next occasion I will include you into that conversation (I was not aware that you also use the dzVents busted test set)
As soon as I found out you fixed the issue, I included the test again and committed it to the same PR before merging. 
Hope that's OK

gizmocuz:
Sure that's OK ! ;) ... Yes I'm using the dzVents test to test if all is still working as the fix for the web interface update was for both dzVents and Lua.
Love the tests !!

PR #3633, opened 2019-09-19
Update EnphaseAPI.cpp

alfniev:
On my enphase metered , curves for Consumption and NetConsumption are identical.
Code for "getConsumptionDetail()" and  "getNetConsumptionDetail()" are copy-paste
Need to update index on "getNetConsumptionDetail()"

### DIFF
--- a/hardware/EnphaseAPI.cpp
+++ b/hardware/EnphaseAPI.cpp
@@ -298,13 +298,13 @@ void EnphaseAPI::getNetConsumptionDetail()
 	{
 		return;
 	}
-	if (root["consumption"][0].empty() == true)
+	if (root["consumption"][1].empty() == true)
 	{
 		_log.Log(LOG_ERROR, "EnphaseAPI: Invalid data received");
 		return;
 	}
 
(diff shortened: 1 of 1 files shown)


gizmocuz:
@alfniev , Thanks for the PR!
This code is also strange as it is calling the HTTP server 3 times... I will make some changes after the merge

gizmocuz:
@alfniev , i have made some changes here:

https://github.com/domoticz/domoticz/commit/c618c9d19e65880ac55f9f9495b40912d00401f1

Is it possible you could verify if all is still working ?

Is it possible you could post here a link to a https://pastebin.com/ containing the production.json output during daytime ?

Do you think we need to use the 'getSunRiseSunSetMinutes' function ? The json data is polled anyway

alfniev:
i've attached 2 json files.look like beta version of domoticz working fine on enphase now
For sunrise function , i see no information about it in my json files.for me , it is useless , but maybe present on non-metered version of enphase. 


Thanks for the update.

gizmocuz:
Thanks!

PR #3657, opened 2019-10-05
OpenWebNet: Update CenPlus and blinds percentage

Spugna85:
Corrected the type of CenPlus to be used as a contact
Fixed the calculation of the percentage to be sent to the blinds


gizmocuz:
Thanks !!

PR #3658, opened 2019-10-06
Update DashboardController.js and LightsController.js

Hollister-git:
To make the daylight logic consistent with the real life and the daylight sensors:
daylight = On means Sunny


gizmocuz:
Thanks!!

Issue #3660, opened 2019-10-07
Add Support for Hue LCG002 GU10 spot


ghost:
Philips has a new GU10 color spot. Smaller and with Gamut C. Device type is LCG002.

I'm already working on the [Hue Dimmer Switch](https://github.com/domoticz/domoticz/issues/3369). I will add the device, but give me a few days, its my first contribution on the Domoticz project.

Regards,
  Richard

ghost:
{"state":{"on":true,"bri":254,"hue":22464,"sat":246,"effect":"none","xy":[0.2385,0.6434],"ct":366,"alert":"select","colormode":"hs","mode":"homeautomation","reachable":false},"swupdate":{"state":"noupdates","lastinstall":"2019-10-07T19:15:37"},"type":"Extended color light","name":"Gang 1","modelid":"LCG002","manufacturername":"Philips","productname":"Hue color spot","capabilities":{"certified":true,"control":{"mindimlevel":200,"maxlumen":300,"colorgamuttype":"C","colorgamut":[[0.6915,0.3083],[0.1700,0.7000],[0.1532,0.0475]],"ct":{"min":153,"max":500}},"streaming":{"renderer":true,"proxy":true}},"config":{"archetype":"spotbulb","function":"mixed","direction":"downwards","startup":{"mode":"safety","configured":true}},"uniqueid":"00:17:88:01:06:8d:d4:61-0b","swversion":"1.55.7_r28193","swconfigid":"9AB97D21","productid":"Philips-LCG002-1-GU10ECLv2"}

PR #3662, opened 2019-10-07
Add Support for Hue LCG002 GU10 spot (#3660)

ghost:
Added the LCG002 bulb to the "GAMUT_C_BULBS_LIST_v". Compiled and tested with actual brand new G10 spot :)


gizmocuz:
Thank you very much for this !!!

Issue #3625, opened 2019-09-13
Incorrect rain_calendar calculation


krejkrejkrej:
There's a problem with how precipitation is transferred to Rain_Calendar from Rain. Most probably a corner case or possibly a problem with calculation order or so. 

This is all the precipitation that was recorded on two days:

> sqlite> select * from Rain where DeviceRowId=343 and Date < '2019-09-07 02:30:00' and Date > '2019-09-06 23:30:00' ;
> 343|507.0|0|2019-09-06 23:35:01
> 343|507.0|0|2019-09-06 23:40:00
> 343|507.0|0|2019-09-06 23:45:00
> 343|507.0|0|2019-09-06 23:50:00
> 343|507.0|0|2019-09-06 23:55:00
> 343|509.0|200|2019-09-07 00:00:00
> 343|509.0|200|2019-09-07 00:05:00
> 343|509.0|200|2019-09-07 00:10:00
> 343|509.0|200|2019-09-07 00:15:00
> 343|509.0|200|2019-09-07 00:20:00
> 343|509.0|200|2019-09-07 00:25:01
> 343|509.0|200|2019-09-07 00:30:00
> 343|509.0|200|2019-09-07 00:35:00
> 343|509.0|200|2019-09-07 00:40:00
> 343|509.0|200|2019-09-07 00:45:00
> 343|509.0|200|2019-09-07 00:50:00
> 343|509.0|200|2019-09-07 00:55:00
> 343|509.0|0|2019-09-07 01:00:00
> 343|509.0|0|2019-09-07 01:05:00
> 343|509.0|0|2019-09-07 01:10:00
> 343|510.0|100|2019-09-07 01:15:00
> 343|511.0|200|2019-09-07 01:20:00
> 343|511.0|200|2019-09-07 01:25:00
> 343|511.0|200|2019-09-07 01:30:00
> 343|511.0|200|2019-09-07 01:35:00
> 343|511.0|200|2019-09-07 01:40:00
> 343|511.0|200|2019-09-07 01:45:00
> 343|511.0|200|2019-09-07 01:50:00
> 343|511.0|200|2019-09-07 01:55:00
> 343|511.0|200|2019-09-07 02:00:00
> 343|511.0|200|2019-09-07 02:05:01
> 343|511.0|200|2019-09-07 02:10:00
> 343|511.0|100|2019-09-07 02:15:00
> 343|511.0|0|2019-09-07 02:20:00
[...]

Issue #3613, opened 2019-09-08
buienrader: Rain device suddenly shows negative numbers


GravityRZ:
currently running build 22290.

All was looking really good for 2 days (saturday and sunday)
all of a sudden i checked and the rainfall seems to have changed to negative.
this was still normal this afternoon

![image](https://user-images.githubusercontent.com/16853535/64492846-1e230c80-d279-11e9-980d-5f6e96774208.png)

gizmocuz:
Okey found the issue...
We are using the rain field 'rainFallLast24Hour'

This is the database:

![image](https://user-images.githubusercontent.com/2265299/64508402-a509d080-d2dd-11e9-905d-8ce9f3980971.png)

A 'normal' rain station (zwave/433/868Mhz) has a continues counter, this one wraps, at buienradars clock
Domoticz is using it's own clock

If you look in Buienradar.cpp, I tried to avoid issues here (~ line 434-452), but I guess it's not 100% safe

And to top it off, we poll every 10 minutes, so here is the issue

Made some modifications, please try beta 10295

Please note, if the rain toay is less then the first value of today (because of the loop, mine was 2.7), today's total will be 2.7

GravityRZ:
ok tested it with 11298 but the graphics still show negative
cleared cache. do i need to reset something else

![image](https://user-images.githubusercontent.com/16853535/64540783-fdae8d00-d320-11e9-8553-93414b08858f.png)

gizmocuz:
@GravityRZ , yes wait 24 hours, they will move out

GravityRZ:
ok. had some idea it would do that.

GravityRZ:
could this be coincidence

i updated to 11298 but now i see all the graphs are flat since 12:00.(updated around 17:00) could this be a side effect of the update.
devices are updated but not every 5 minutes but later (10 minutes or longer  i think)


![image](https://user-images.githubusercontent.com/16853535/64559872-8f30f580-d347-11e9-9878-f1e29c1d06fa.png)

GravityRZ:
checked the json link
i can access it and the values in there are the same as in domoticz.
This must mean that station Berkhout is sending out the wrong info. at least is does not change.
Maybee somebody should go up there and replace the batteries

[...]

Issue #3598, opened 2019-09-01
New Buienhardware does not update daily stats for rain device


akamming:
Very happy with the new native buienradar implementation. There seems to be a small bug however in the updating of the stats of the rain device: The daily counters are always zero. The reports also show no data.

details are in the forum thread [New Buienradar hardware does not update rain counters](https://www.domoticz.com/forum/viewtopic.php?f=6&t=29206&p=222772#p222772)

GravityRZ:
same here. also i notice that it looks like there is not a lot of rain reported (while there actually is a lot of rain falling)
maybee it is related

![image](https://user-images.githubusercontent.com/16853535/64362282-891cdb00-d00e-11e9-86cb-972053788d23.png)

gizmocuz:
@akamming , i was not able to test/debug with actual rain, but i suspect the rain counter is not a continues counter and always goes to zero when a new day begins.
Yep, just checked, it's zero today.
I am going to debug this

@GravityRZ , thats not related to this issue

gizmocuz:
@akamming , looks like i was using a wrong field for the total rain of the day...
Please try beta 11289, and test a few days... let's hope for some rain ;)

akamming:
@gizmocuz Tx for the fix! Just updated to beta (version shows 4.11290 now).  I am hopeful: I already see a change: The last day on the weekly, monthly and yearly graph now have positive values, which i didn't observe before (you can compare with the picture @GravityRZ attached in this issue, The last day is 0 in those graphs):

![buienradar_fix](https://user-images.githubusercontent.com/30364409/64412552-e311b500-d08f-11e9-8481-a37398a1c8cd.jpg)

I will keep monitoring the stats the upcoming days. Tx!

GravityRZ:
Yes with me it is also working.
appparently we both have 2.7mm total rain on saturday.
@akamming what city are You in?
 i suspect you are closeby. i am in Alkmaar which uses
 station Berkhout. Does the rainfall measurement/prediction add up to the reality?
I have the feeling that with the script there was much more going on then right now.
(eg there is a lot of rain falling but do not see it back in the charts)
maybee because of the fragmented clouds. if it starts raining all over in the netherlands we know for sure  me must see something.
@Gizmocuz thanks for the good work

![image](https://user-images.githubusercontent.com/16853535/64471344-e7ea6d80-d150-11e9-8216-122e182a498a.png)

akamming:
@gizmocus

[...]

PR #3663, opened 2019-10-10
Fix for Buienradar Rain sensor

akamming:
fix for issue #3598  (and possibly #3625 and #3613)

### DIFF
--- a/hardware/Buienradar.cpp
+++ b/hardware/Buienradar.cpp
@@ -422,14 +422,9 @@ void CBuienRadar::GetMeterDetails()
 		SendCustomSensor(2, 1, 255, sunpower, "Sun Power", "watt/m2");
 	}
 
-	float precipitation = -1;
-
 	if (!root["precipitationmm"].empty())
 	{
-		precipitation = root["precipitationmm"].asFloat();
-	}
-	if (precipitation != -1)
-	{
+		float precipitation = root["precipitationmm"].asFloat();
 		SendRainRateSensor(1, 255, precipitation,  "Rain");
 	}
 }
--- a/hardware/Buienradar.cpp
+++ b/hardware/Buienradar.cpp
@@ -154,7 +154,7 @@ void CBuienRadar::Do_Work()
 				//reset our rain counter
 				m_actDay = ltime.tm_mday;
 				m_lastRainCount = 0;
-				SendRainSensorWU(1, 255, 0, 0, "Rain");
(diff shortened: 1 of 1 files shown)


gizmocuz:
@akamming , thank you for the PR, looking good !
Just to check what has changed, you removed all extra checks and instead of SendRainSensorWU you are now issuing SendRainRateSensor ?

One small favor to ask could you rewrite the last part as

	if (!root["precipitationmm"].empty())
	{
		float precipitation = root["precipitationmm"].asFloat();
                SendRainRateSensor(1, 255, precipitation,  "Rain");
	}

Then it's like the other sensors

akamming:
> @akamming , thank you for the PR, looking good !
> Just to check what has changed, you removed all extra checks and instead of SendRainSensorWU you are now issuing SendRainRateSensor ?
> One small favor to ask could you rewrite the last part as
> if (!root["precipitationmm"].empty())
> {
> 	float precipitation = root["precipitationmm"].asFloat();
>             SendRainRateSensor(1, 255, precipitation,  "Rain");
> }
> 
> Then it's like the other sensors

Ok will do that tomorrow

akamming:
> Just to check what has changed, you removed all extra checks and instead of SendRainSensorWU you are now issuing SendRainRateSensor ?

Correct

akamming:
> @akamming , thank you for the PR, looking good !
> Just to check what has changed, you removed all extra checks and instead of SendRainSensorWU you are now issuing SendRainRateSensor ?
> One small favor to ask could you rewrite the last part as
> if (!root["precipitationmm"].empty())
> {
> 	float precipitation = root["precipitationmm"].asFloat();
>             SendRainRateSensor(1, 255, precipitation,  "Rain");
> }
> 
> Then it's like the other sensors

Done

gizmocuz:
Thanks !!!

rwaaren:
Many thanks for the PR !

Just one remark. Not very important but much obliged if you would want to consider.

The number of digits behind the decimal point gives the idea that the measurement / prediction is very accurate. Wouldn't it be sufficient to have only 1 digit behind the decimal point for Rain Intensity, Next Rainshower Avg Rainrate and Next Rainshower Max Rainrate ?

Buienradar

gizmocuz:
@rwaaren , done in commit 
https://github.com/domoticz/domoticz/commit/a4bb7f284c75fa7c2e13313d2f2a178516cbe36f

PR #3668, opened 2019-10-16
Added Next Rainshower Sensors in Buienradar hardware

akamming:
Several updates on the rainprediction as part of the new native buienradar implementation
- Next Rainshower is predicted in 4 new custom sensors (leadtime, duration, average rainrate and max rainrate)
- Added Possible Rain switch ('old' functionality restored)
- fixes an issue for handling 0.00 timestamp in rainprediction
- fixes an issue when the rain prediction is just a bit old (which happens quite often)

buienradar_new


gizmocuz:
Thank you !!

PR #3669, opened 2019-10-16
Fix include paths to not confuse local files with system files

danielwallner:
The build fails on my Arch Linux box without these changes

### DIFF
--- a/hardware/1Wire/1WireForWindows.cpp
+++ b/hardware/1Wire/1WireForWindows.cpp
@@ -4,7 +4,7 @@
 #include "1WireForWindows.h"
 #include "../../main/Logger.h"
 #include <boost/optional.hpp>
-#include "../json/json.h"
+#include "../../json/json.h"
 #include <WS2tcpip.h>
 
 #define _1WIRE_SERVICE_PORT "1664"
--- a/hardware/plugins/PluginManager.cpp
+++ b/hardware/plugins/PluginManager.cpp
@@ -14,14 +14,14 @@
 #include "PluginMessages.h"
 #include "PluginTransports.h"
 
-#include "../main/Helper.h"
-#include "../main/Logger.h"
-#include "../main/SQLHelper.h"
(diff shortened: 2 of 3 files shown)


gizmocuz:
Thanks, but could you post the output of your original build issues?
You are removing includes

gizmocuz:
Thanks!!

danielwallner:
For completeness, the build issue was a number of these:

PluginManager.cpp:(.text+0x3cd6): undefined reference to `Json::Value::operator=(Json::Value&&)'

Caused by binary incompatibility of system and local version of json/json.h

This was not found during compile due to this line in CMakeLists.txt:

include_directories(${CMAKE_SOURCE_DIR}/main)

Which makes ../json/json.h always be found even if the relative path is wrong, but unfortunately on my system a sibling to the system json include was in the include paths and this got picked up instead (if the relative paths are correct it won't be).

A different mechanism to find stdafx.h than adding main to the include dirs would make this more obvious.

I'm a bit surprised that this doesn't seem to be a problem for more users.

gizmocuz:
@danielwallner , seems your the first user that reports this issue.
I compiled on various systems and never had this issue.

Did you follow:

https://www.domoticz.com/wiki/Raspberry_Pi_-_Build_Domoticz_from_source

danielwallner:
I followed what seemed relevant to Arch as it isn't using APT package management.
I also had to specify ssl libraries or cmake wouldn't find them.

This problem is likely only happening on distros that default installs headers like Arch or Gentoo and have an incompatible jsoncpp installed. Maybe that's not so common then.

ffontaine:
For information, I got the same build failure with domoticz on buildroot: http://autobuild.buildroot.org/results/a73/a73406eb780a454369ea997654b6b4c6b3757a41/build-end.log.
It could be related to the latest versions of jsoncpp: we started to got build failures after bumping jsoncpp from 1.8.4 to 1.9.1.

PR #3670, opened 2019-10-16
separate monthly usage and return totals

laszlovandenhoek:
In the monthly report for energy, if you can return energy, the totals shown next to the usage (T1 + T2) and the return (R1 + R2) are both the same: the grand total ( (T1 + T2) - (R1 + R2) ) - not very useful IMO:

before

This patch makes it so that the T1 and T2 are totalled, and R1 and R2 too. Then, the grand total (which was previously shown after both) is displayed separately:

after


gizmocuz:
Thank you !!

jan-vandenberg:
The TOTAL number (Usage - Return), is not yet available/visible on the Year report page. Would be great to have it here too.

image

laszlovandenhoek:
@jan-vandenberg yes, that would be something to add above/below/next to the "Year Cost".

This is where the code should be changed: https://github.com/domoticz/domoticz/blob/development/www/app/report/EnergyMultiCounterReport.html#L20

Once you make the change, you should open a new Pull Request.

jan-vandenberg:
Done: https://github.com/domoticz/domoticz/pull/4272

PR #3675, opened 2019-10-20
add BuienRadar wiki link to hardware setup page

akamming:
buienradar_url


gizmocuz:
@akamming , we might want this for other hardware setup types as well.
It might be a good idea to make a static wiki/help link in each hardware class that is returned together with the hardware name/type in the hardware setup page.
If filled, then we add add an additional Wiki/Help URL at the bottom.

Instead of merging this, I think i will prepare the hardware classes for this ...

akamming:
@gizmocuz  yep that would be a good improvement!

akamming:
Closing this pull request since it's not being pulled into the code

PR #3686, opened 2019-10-24
Fix strings

xbeaudouin:
There is an unneeded space on %s in Tado.cpp, that fails to build under freebsd :  

```
/home/kiwi/domoticz.dev/hardware/Tado.cpp:292:72: warning: flag ' ' results in
      undefined behavior with 's' conversion specifier [-Wformat]
  ..."Tado: Failed to get information on zone '%s': % s", zone.Name.c_str(), ...
                                                    ~^~

```

Regards


gizmocuz:
Thanks for reporting!

PR #3690, opened 2019-10-25
SunRiseSet: Use round instead of std::round.

neheb:
The latter is not available with uClibc-ng.


gizmocuz:
Thanks !!

PR #3691, opened 2019-10-26
Correcting the handling of the free_fall state of the agara cube

rwaaren:
modified:   hardware/XiaomiGateway.cpp

as reported on the [forum](https://www.domoticz.com/forum/viewtopic.php?f=6&t=29886l)


gizmocuz:
Thanks !!

PR #3694, opened 2019-10-29
main/unzip_iterator.h: include zlib.h

ffontaine:
Include zlib.h to be able to build domoticz with minizip 2.9.0
(https://github.com/nmoinvaz/minizip) indeed minizip does not define
uLong (see https://github.com/nmoinvaz/minizip/issues/405)


gizmocuz:
Thanks !

PR #3696, opened 2019-10-29
OpenWebNet: rename all class members with m_

Spugna85:
rename all class members with m_


gizmocuz:
Thank you very much for this !!

PR #3698, opened 2019-10-29
Changed to BUIENRADAR_RAIN url as published by Buienradar. Include url in LOG_ERROR 

rwaaren:
in error message
	modified:   hardware/Buienradar.cpp

About 10% of the calls to raintext url fail.
Looked for options to increase timeout but 10 seconds should be more then enough.

Is it an idea to repeat the call after 1 minute when failed and try this x times before logging it as an error ?

### DIFF
--- a/hardware/Buienradar.cpp
+++ b/hardware/Buienradar.cpp
@@ -18,7 +18,7 @@
 #define BUIENRADAR_GRAFIEK_URL "https://www.buienradar.nl/nederland/weerbericht/weergrafieken/" //station_id
 //#define BUIENRARA_GRAFIEK_HISTORY_URL "https://graphdata.buienradar.nl/1.0/actualarchive/weatherstation/6370?startDate=2019-09-15"
 
-#define BUIENRADAR_RAIN "https://gadgets.buienradar.nl/data/raintext/?lat=" // + m_szMyLatitude + "&lon=" + m_szMyLongitude;
+#define BUIENRADAR_RAIN "https://gpsgadget.buienradar.nl/data/raintext?lat=" // + m_szMyLatitude + "&lon=" + m_szMyLongitude;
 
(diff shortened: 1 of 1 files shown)


gizmocuz:
I do not think that logging the URL will be helpful for any user as they do not know what it is, and they can not change it.
Maybe just change the URL in the top part of the code ?
Yes a retry could help indeed, you could also do this right away ? Like create a function FetchDataFromURL that will return the result, but tries it twice ?

rwaaren:
On the url in the message fair point but my thinking is that it will be useful for finding where the Error message come from. 
No need to include the full url but I will change it to identify if it is the call to raintext or one of the others.
On the function: Like the idea but my C++ skills are not to the level that I can do this immediate.
Let me study a bit more and I can come with something in a couple of days.

gizmocuz:
@rwaaren , maybe log "Can not reach Buienradar, please check your internet connection.." ?
I will have a look at this and will also change the fetch URL

gizmocuz:
@rwaaren , why did you change the URL ? I am not sure this is needed...
I checked my log... I don't get that many timeouts:

> 2019-10-24 09:36:06.848 Status: BuienRadar: Nearest station: Meetstation xxxx ID: yyyy
> 2019-10-30 07:15:21.696 Error: BuienRadar: Error getting http data! (Check your internet connection!)
> 
> 
Six days.... this morning i received a timeout...

Do you experience those timeouts ?

rwaaren:
Yes. About 10 % and at least two other forum members experience the same. 
To other url's I don't see this behavior and is my internet connection quite stable (fingers crossed)

gizmocuz:
@rwaaren , thanks, i have changed the url in both branches and added a retry
Could you have a look at the changes in .gitignore... if they where not meant, i have a new file ready to commit (added the new OZW cache file)

Issue #3369, opened 2019-07-02
[Request] Philips Hue Dimmer Switch 


Lyr3x:
It would be nice if domoticz can add a Hue Dimmer Switch (ZLLSwitch) to make it work with other domoticz devices. The workaround with the LUA script is not very handy.
I think the starting point is at the GetSensors function. We already have an empty else if condition for the switch types. 
```cpp
else if ((hsensor.m_type == SensorTypeZGPSwitch) || (hsensor.m_type == SensorTypeZLLSwitch))
{
}
```
Are there any rabbit holes I am not aware of?

Here is an example JSON
```json
"name": "wall switch",
    "type": "ZLLSwitch",
    "modelid": "RWL021",
    "manufacturername": "Philips",
    "productname": "Hue dimmer switch",
    "diversityid": "**cleared**",
    "swversion": "5.45.1.17846",
    "uniqueid": "**cleared**",
    "capabilities": {
      "certified": true,
      "primary": true,
      "inputs": [
        {
          "repeatintervals": [
            800
          ],
          "events": [
            {
              "buttonevent": 1000,
              "eventtype": "initial_press"
            },
            {
              "buttonevent": 1001,
              "eventtype": "repeat"
            },
            {
              "buttonevent": 1002,
              "eventtype": "short_release"
            },
            {
              "buttonevent": 1003,
              "eventtype": "long_release"
            }
          ]
        },
        {
          "repeatintervals": [
            800
          ],
          "events": [
            {
              "buttonevent": 2000,
              "eventtype": "initial_press"
            },
            {
              "buttonevent": 2001,
              "eventtype": "repeat"
            },
            {
              "buttonevent": 2002,
              "eventtype": "short_release"
            },
            {
              "buttonevent": 2003,
              "eventtype": "long_release"
            }
          ]
        },
```

gizmocuz:
Yes i think that's the place to add it....
But i miss the 'state' ... it seems your dimmer has two buttons, and you would like to know what button was pressed, and how it was pressed.
The 'how' (state) i do not see in your example JSON, is this the call to http://<ip:port/api/username ?
There should be a 'sensors' part in the JSON as well, could you post that ?

You want to convert 'short_release' and 'long_release' to be part of the 'id' for the switch so you get unique switches.
After this you should be able to use those switches in scenes/groups in domoticz

Could you point to the LUA script ?

[...]

PR #3700, opened 2019-10-30
#3369 Hue Switch Implementation

ghost:
Implementeing the ZLL switch, TAP and Friends Of Hue switches.


gizmocuz:
@RedDigital ,Thank you very much for this addon !
One request, could you change the member variable initialization to look like this

int myVar = 1;

instead of

int myVar {1};

This will make it more generic throughout the Domoticz code, and is clearer

gizmocuz:
Thanks!!

Issue #3666, opened 2019-10-13
TCP Hardware with Data timeout crashes


ghost:
I have P1 smart meter with LAN interface (its a simple rs232 to ethernet converter). Lately i have fairly unstable wifi network. Causing the PI loosing it connection sometimes. By itself it's not a domoticz issue.

But very often i find that domoticz is crashed after an reconnect. It generates a domoticz_crash.log (see below).

ghost:
Below the content of the domoticz_crash.log. Thread 32 has crashed.

```
[New LWP 24922]
[New LWP 24923]
[New LWP 24924]
[New LWP 24928]
[New LWP 24929]
[New LWP 24930]
[New LWP 24931]
[New LWP 24932]
[New LWP 24933]
[New LWP 24934]
[New LWP 24935]
[New LWP 24936]
[New LWP 24937]
[New LWP 24938]
[New LWP 24939]
[New LWP 24940]
[New LWP 24941]
[New LWP 24945]
[New LWP 24946]
[New LWP 24947]
[New LWP 24948]
[New LWP 24949]
[New LWP 24950]
[New LWP 24951]
[New LWP 24952]
[New LWP 24953]
[New LWP 24954]
[New LWP 24955]
[New LWP 24956]
[New LWP 24957]
[New LWP 28965]
[New LWP 28966]
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/arm-linux-gnueabihf/libthread_db.so.1".
0x76bdbd60 in nanosleep () at ../sysdeps/unix/syscall-template.S:84
84	../sysdeps/unix/syscall-template.S: No such file or directory.
  Id   Target Id         Frame 
* 1    Thread 0x76143180 (LWP 24921) "domoticz" 0x76bdbd60 in nanosleep () at ../sysdeps/unix/syscall-template.S:84
  2    Thread 0x75d0a430 (LWP 24922) "Watchdog" 0x76bdbd60 in nanosleep () at ../sysdeps/unix/syscall-template.S:84
  3    Thread 0x75501430 (LWP 24923) "SQLHelper" syscall () at ../sysdeps/unix/sysv/linux/arm/syscall.S:37
  4    Thread 0x74d00430 (LWP 24924) "PluginMgr" syscall () at ../sysdeps/unix/sysv/linux/arm/syscall.S:37
[...]

PR #3701, opened 2019-10-30
#3666 ASyncTCP fix for hanging in destructor

ghost:


gizmocuz:
Thank you very much for all work and this PR !!! Much appreciated, and hope to test this very soon!

Is it possible to change normal member variable initialization to look like

bool m_bMyBool = true;

instead of

bool m_bMyBool ( true };

This makes it more uniform throughout the application.

I would only use "uniform" initialization for structs/classes/arrays and so

ghost:
Yes no problem, i will do so.

ghost:
I pushed the changes for both the PR. 

Be critical at the code style. I think its important for readability to have the same style.

gizmocuz:
Thanks!!

Issue #3707, opened 2019-11-03
Build failed


pipiche38:
Last commit was:
```
commit 85b5bfab1327b818f3070775715b9f740c1a2cb2 (HEAD -> development, origin/development, origin/HEAD)
Date:   Sun Nov 3 17:02:34 2019 +0100

    Prevent error when using math.pow in user script but do give deprecate warning.
            modified:   dzVents/runtime/Utils.lua

            modified:   dzVents/runtime/Utils.lua
```

Here after are the last lines

```
                 from /usr/local/include/boost/asio/detail/regex_fwd.hpp:20,
                 from /usr/local/include/boost/asio/read_until.hpp:23,
                 from /usr/local/include/boost/asio.hpp:114,
                 from /usr/share/domoticz/beta-domoticz/hardware/ASyncTCP.cpp:3:
/usr/share/domoticz/beta-domoticz/hardware/ASyncTCP.cpp:32:23: note:   ‘std::shared_ptr<std::thread>’ is not derived from ‘const std::_List_iterator<_Tp>’
  assert(mTcpthread == false);
                       ^~~~~
In file included from /usr/include/c++/8/list:63,
                 from /usr/share/domoticz/beta-domoticz/hardware/../main/Logger.h:4,
                 from /usr/share/domoticz/beta-domoticz/hardware/ASyncTCP.cpp:6:
/usr/include/c++/8/bits/stl_list.h:1987:5: note: candidate: ‘template<class _Tp, class _Alloc> bool std::operator==(const std::__cxx11::list<_Tp, _Alloc>&, const std::__cxx11::list<_Tp, _Alloc>&)’
     operator==(const list<_Tp, _Alloc>& __x, const list<_Tp, _Alloc>& __y)
     ^~~~~~~~
/usr/include/c++/8/bits/stl_list.h:1987:5: note:   template argument deduction/substitution failed:
In file included from /usr/include/c++/8/cassert:44,
                 from /usr/local/include/boost/predef/detail/_cassert.h:12,
                 from /usr/local/include/boost/predef/library/c/_prefix.h:11,
                 from /usr/local/include/boost/predef/library/c.h:13,
                 from /usr/local/include/boost/predef/library.h:13,
                 from /usr/local/include/boost/predef.h:16,
                 from /usr/local/include/boost/regex/config.hpp:45,
                 from /usr/local/include/boost/regex_fwd.hpp:24,
[...]

PR #3708, opened 2019-11-03
#3707 Fixed the compile error in AsyncTCP.cpp

ghost:
On some compilers it didn't compile. the std::shared_ptr didn't use the bool operator.


gizmocuz:
Thank you!

PR #3711, opened 2019-11-05
Fix detection of LUA version if not linking with the provided sources.

xbeaudouin:
Trivial patch that allow linking Domoticz with LUA 5.3 if we don't use the provided LUA files.

Regards


gizmocuz:
Thanks!!

PR #3718, opened 2019-11-08
- AsyncTCP changes

rkluwen:
- AsyncTCP fixes
- proxyclient also connects when openssl is not compiled in
- PluginTransports compiles when openssl is not compiled in (but
probably won't work)


gizmocuz:
Thank you for this !!

Issue #3721, opened 2019-11-10
iterator out of bound


RDols:
First, I'm the same person as RedDigital. I'm cleaning up / merging various accounts i have on github.

During some testing i had a crash in CEventSystem::GetEventTrigger. I saw the iterator go out of bound. Just found a closed issue item on this too: #2644. But was not fixed at that time.

RDols:
Problem occurs when m_eventtrigger has only one element. The line

`itt = m_eventtrigger.erase(itt) - 1;`

wil get the iterator out of bound. Erase return the element after what is erased. That would be "end". since m_eventtrigger has 0 elements left, subtracting 1 will make the iterator go out of bound.

RDols:
I make a PR in few minutes.

gizmocuz:
Thanks for spotting! Merged the PR, thank you !

PR #3722, opened 2019-11-10
#3721 fix iterator out of bound

RDols:
if m_eventtrigger has only one element. After deleting it's in error because of the -1


gizmocuz:
Thanks !!

PR #3733, opened 2019-11-18
csocket: Replace sys/poll with poll

neheb:
Fixes warning when compiling under musl:

warning redirecting incorrect #include <sys/poll.h> to <poll.h>


gizmocuz:
Thank you !

PR #3734, opened 2019-11-18
serial/unix: Replace sys/signal with signal

neheb:
Fixes warning under musl:

warning redirecting incorrect #include <sys/signal.h> to <signal.h>


gizmocuz:
Thank you!

PR #3735, opened 2019-11-18
Tado: Fix printf format for 32-bit

neheb:
The .size method usually returns size_t, which is represented by %zu in printf.


gizmocuz:
Thank you !

PR #3738, opened 2019-11-20
ClearTimers only on ActiveTimerPlan

syrhus:
ClearTimers erase timers for ALL timerplan. Updated to clear timers for ActiveTimerPlan only


gizmocuz:
Thank you !

PR #3748, opened 2019-11-23
make database_version number available for admin users in getversion API call

rwaaren:
modified:   main/WebServer.cpp

### DIFF
--- a/main/WebServer.cpp
+++ b/main/WebServer.cpp
@@ -2584,6 +2584,7 @@ namespace http {
 		}
 
 		void CWebServer::Cmd_GetVersion(WebEmSession & session, const request& req, Json::Value &root)
+
 		{
 			root["status"] = "OK";
 			root["title"] = "GetVersion";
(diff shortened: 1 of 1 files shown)


gizmocuz:
Question, why would you need this "internal" value ?

rwaaren:
> Question, why would you need this "internal" value ?

I try to keep track of the database versions in combination with domoticz and dzVents versions Every once in a while I  see forum questions about mismatching domoticz - database versions after a recovery activity and it is helpful to include in my answer a (previous) domoticz version number that would be able to start with the database version restored.
I now use a popen(sqlite command)  in a script to get this information but my thinking was that it would be better to have all version information (domoticz,  Python,  dzVents and database ) combined in one API.

gizmocuz:
@rwaaren , that is because users go from a beta to a previous build.
Not that the database have many patches, but this is not done and explained.
There is also no relation to dzVents as it involves internal changes/updates

rwaaren:
I understand but users do make mistakes and do not always have recent backups.
I also know there is no other relation between dzVents vresion and DB_Version then that they are both versions.  
If you don't want this information to be added the getversion API call for admin users please feel free to close this PR or advise what might be a better solution to get this information.

gizmocuz:
Problem is, they should not go back to a previous version... i can not make it any better then it is
It is extreme dangerous !!
Okey closing this one. sorry... it should stay internal and you should not need it

PR #3749, opened 2019-11-23
#3661 Web socket crash (final?) fix

RDols:
It's a subset that has been tested last week. Also reverts 2 previous fixes.


gizmocuz:
Thank you ! And all for testing !!

PR #3754, opened 2019-11-24
Fix BleBox search issues with duplicate devices

gadgetmobile:
Cause:

During a search, temporary/stack-based `IPAddress` (std::string) objects were passed to long running threads, which means the `addNode()` threads were likely using already destroyed `IPAddress` objects.

During the `for` loop, those destroyed stack objects were replaced with  IPAddress objects from next iterations, which either caused the detection to fail (connecting to a different ip) or create duplicate devices (connecting to an address already added by another thread, or adding a record with an ip from another loop iteration).

Other changes:

- log an error if the pattern is invalid or unsupported
- minor renames ("pattern") and more helpful comments
- skip search if pattern results in no hosts (future-proof)

### DIFF
--- a/hardware/BleBox.cpp
+++ b/hardware/BleBox.cpp
@@ -1251,32 +1251,19 @@ void BleBox::UpdateFirmware()
 	}
 }
 
-void BleBox::SearchNodes(const std::string & ipmask)
+void BleBox::SearchNodes(const std::string & pattern)
 {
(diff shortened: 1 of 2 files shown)


gadgetmobile:
The diff layout is weird (due to code being refactored out into a separate method).

gizmocuz:
Thanks !!

PR #3757, opened 2019-11-26
remove unneccessary logging and add rkluwen fix for issue #3661

akamming:
Better Handling of connection_websocket_Closing state in connection::close (rkluwen fix)
Removed unneccessary loglines

fix was tested on rpi (i have no windows machine on  which i can test, but should be ok. simple fix)


gizmocuz:
Thanks!

Issue #3761, opened 2019-11-29
memory leak 


akamming:
Running on rpi3b+ stretch latest beta. 

The current beta release has a memory leak. I did several tests and on my system, on which i concluded
- 4.11505 is ok
- 4.11506 leaks memory (making commit Jsoncpp deprecate fix (#3741) a possible candidate)
- current production version (on 29/11/2019)  leaks more memory (So my suspicion is there might be more than one memory leak..) 

The analysis can be found in [this forumthread](https://www.domoticz.com/forum/viewtopic.php?f=6&t=30305&p=229607#p229607) as well as more users with the same issue

i already looked  at the code changes at commit #3741 , but i don't see where it leaks, hopefully more experienced developers can have a look at it...

akamming:
I'm not trying to report issues, but also try to fix them, so in the latest codebranch i reverted the commit #3741 and solved some merge issues, trying to prove this commit is really the issue. This morning i ran the test, and it looks like the memory leak is gone. With the memory leak, the memory usage immediately starts to rise. 

This effect is gone when this commit 3741 is no longer in the code, as you can see in the following graph : 
![result_versions2](https://user-images.githubusercontent.com/30364409/69899414-c53dae80-1365-11ea-99bd-cbee78a1acd5.png)
(4.11540 has the same code as 4.11539, just with commit 3741 reverted. The code can be found [here](https://github.com/akamming/domoticz/tree/revert3741) )

If someone tests this code also on windows (i cannot compile on windows), i could make a PR to fix the memory leak, but probably commit 3741 is in the code with a reason. So probably it's just a workaround till we fix the real problem. 

I already tried, but i cannot find the memory leak in the code of this commit (which probably says more about my coding skills than the complexity of the code ;-)). So I won't be able to fix the root cause.

@gizmocuz i just checked and found out you created all this code yourself. What is your advice here?

rwaaren:
Also seem to be a relation with hardware. I disabled Hue on my test system with latest Beta and no memory leak occurred after this. Might well be that all hardware using the new JSON stuff suffer from this but I don't have all hardware on my test system.
Now started a test run on my production system with disabled Hue. 
V4.11539 

```
Native hardware:
Domoticz Internal
RFtrx433
Aeotec  Zwave stick
RFLink
P1
Youless
Hue
Harmony
PI1 Motherboard
Evohome
Xiaomi
Yeelight
MQTT
DarkSky
Virtual
Buienradar

Plugins:
[...]

PR #3767, opened 2019-11-30
Deallocate CharReader after use

akamming:
Fixes issue #3761 (Memory leak at json_helper)

### DIFF
--- a/main/json_helper.cpp
+++ b/main/json_helper.cpp
@@ -3,12 +3,13 @@
 
 bool ParseJSon(const std::string& inStr, Json::Value& json_output, std::string *errstr)
 {
+
+	if (inStr.empty())
+		return false;	
+	
 	bool result;
 	Json::CharReaderBuilder builder;
 	Json::CharReader *reader = builder.newCharReader();
-
-	if (inStr.empty())
-		return false;
 	
 	result = reader->parse(
 		reinterpret_cast<const char*>(inStr.c_str()),
--- a/main/json_helper.cpp
+++ b/main/json_helper.cpp
@@ -3,13 +3,21 @@
 
 bool ParseJSon(const std::string& inStr, Json::Value& json_output, std::string *errstr)
 {
+	bool result;
+	Json::CharReaderBuilder builder;
+	Json::CharReader *reader = builder.newCharReader();
+
 	if (inStr.empty())
 		return false;
 	
-	return Json::CharReaderBuilder().newCharReader()->parse(
+	result = reader->parse(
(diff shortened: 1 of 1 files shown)


akamming:
> What happens if you have many cases of inStr.empty() ? Could this lead to memory leak ?

no c++ expert either, but I do not suspect this statement. the .empty() function is just a check whether the string is filled or not. If it it leaks i should still see a leak on my system. This is what i see on my system:
fixes_effect
16.14: 4.11539 is started, has a big memory leak. Memory keeps rising as long as it runs
08.47: 4.11540 is started, contains this PR, memory leak no longer leaks
11.40: 4.11540 is started, with now contains an alternate fix, no memory leak, but average memory usage went up by 5mb). 

In the alternate fix, the code ParseJSon function is change to
```
bool ParseJSon(const std::string& inStr, Json::Value& json_output, std::string *errstr)
{
        Json::CharReaderBuilder builder;
        const std::unique_ptr<Json::CharReader> reader(builder.newCharReader());

        if (inStr.empty())
                return false;

        return reader->parse(
                reinterpret_cast<const char*>(inStr.c_str()),
                reinterpret_cast<const char*>(inStr.c_str() + inStr.size()), &json_output,
                errstr);
}

```
@gizmocuz  This is cleaner c++ code, but because of the 5mb memory rise i did not change this PR. But if you think this would be better from a code quality perspective, i can change the PR if you like...

akamming:
@rwaaren  Sorry for the spam. I just thought of it. I would not remove the check on empty strings. but there is still another leak in this code indeed, which could explain why my fix makes it much better, but not fully fixes is it. In case of an empty string the "delete reader" is not called before the return!, leaking memory in case of empty strings! 


I don't have access to my system now (at work ;-)), so i cannot add this to the PR now.   In my alternate solution this would already be fixed. 

My suggestion is to have this PR pulled in asap, it would then be solved for many users. And then based on which solution of the code @gizmocuz  decides, we can make an additional PR to solve the other leak.

[EDIT] aha... you edited your remark; i understand now we are talking about the same thing ;-)
An additional remark: It might not a bad idea to debug why empty strings are sent.... has nothing to do with this leak (this function should be resistent to empty strings), but an empty string might be indication that something else is wrong in the code which sends the empty string.

gizmocuz:
Thanks for the PR. I have corrected the code and made it consistent with the ParseJSonStrict helper function.
Thank you for spotting this!

akamming:
Tx. I switched back to "normal" beta (4.11544) branch again this morning, now running for 4 hrs. I can confirm memory leak is gone

Issue #3781, opened 2019-12-06
Astronomical twilight start: 06:60:00 : generate error in Domoticz_API


jfsant29:
2019-12-06 00:00:23.950  Status: Domoticz V4.11062 (c)2012-2019 GizMoCuz
2019-12-06 00:00:23.950  Status: Build Hash: 77d4961e5, Date: 2019-07-29 12:39:01
2019-12-06 00:00:23.950  Status: Startup Path: /home/pi/domoticz/
2019-12-06 00:00:23.997  Sunrise: 08:53:00 SunSet: 17:24:00
2019-12-06 00:00:23.997  Day length: 08:30:00 Sun at south: 13:08:00
2019-12-06 00:00:23.997  Civil twilight start: 08:17:00 Civil twilight end: 17:**60**:00
2019-12-06 00:00:23.997  Nautical twilight start: 07:38:00 Nautical twilight end: 18:39:00
2019-12-06 00:00:23.997  Astronomical twilight start: 06:**60**:00 Astronomical twilight end: 19:17:00

rwaaren:
Thanks for reporting. Should be fixed when PR #3783 is approved and merged

rwaaren:
#3783 is merged into Beta V4.11553. Please check and report back #

gizmocuz:
@rwaaren , does it need to be revert?

rwaaren:
@gizmocuz sorry i meant report back

gizmocuz:
Closing issue as it is fixed by your PR, if anyone has an issue please reopen
Thanks for the patch!

PR #3782, opened 2019-12-06
Fix for issue #3781 Correct rounding errors in all Sunsets and daylength

rwaaren:
#3781


rwaaren:
I now just copy /pasted the solution @gizmocuz applied to sunset a couple of times and changed the variables (and this works) but if someone can show me how to make these similar checks into a c++ function (as I would use in Lua), I would learn something from it.

PR #3783, opened 2019-12-06
Fix for issue #3781 Correct rounding errors in all astronomical sunrise / sunsets and day length

rwaaren:


rwaaren:
Should fix #3781 and other issues where atronomical sunrises / sunsets are displayed as xx:60

gizmocuz:
Thanks!

PR #3792, opened 2019-12-12
Check on permission before json type commands

akamming:
@gizmocuz  there is a vulnerability  in the json interface. I sent you a DM on the forum for how you can exploit. This is a patch for that vulnerability. Can you check if this is the right fix? This code is called quite often and as you know i'm not an experienced programmer.

On my system the patch works and i see no side effects.

### DIFF
--- a/main/WebServer.cpp
+++ b/main/WebServer.cpp
@@ -11195,6 +11195,12 @@ namespace http {
 
 		void CWebServer::RType_CustomLightIcons(WebEmSession & session, const request& req, Json::Value &root)
 		{
+			if (session.rights < 0 )
+			{
+				session.reply_status = reply::unauthorized;
+				return; //Only access allowed when there are rights
+			}
+
 			int ii = 0;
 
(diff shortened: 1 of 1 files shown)


akamming:
hmmm… the build fails on something with mosquito (which i don't see on my system, so don't know how to fix). A colluege of mine also just tested the code and found out that this call:

http://domoticz-ip<:port>/json.htm?username=MkE=&password=OVM=&api-call

now give a 401 error, als with correct credentials, so not the right fix... closing the PR, will make a new one after this is fixed.

akamming:
Reopening this PR. I took a look at the finding of my colluegue, but apparantly he made a mistake in testing. He cannot reproduce his issue with this PR. Neither can i reproduce his finding. 

@gizmocuz  This is a fix for a vulnerability in the json interface. I tried to send a DM on the forum to explain how you can exploit this vulnerability, but for some unknow reasons my DM's stay in the outbox of the forum. Is there some other way i can send you this info. Because of the security risk I do not want to put this info on github...

gizmocuz:
@akamming , I have send you an email...

Issue #3793, opened 2019-12-12
Multipl MySensors Gateway with MQTT interface - same client ID - disconnection


darkManPi:
There is still problem with MQTT Mysensors connection.
With more than one  "MySensors Gateway with MQTT interface" hardwares connecting to same MQTT server, MQTT will keep on disconnecting/reconnecting. 

There was similar case #3750 
In fact commit b89ba87 help for case with single MQTT connection per Domoticz instance having  few independent domoticz instances, but not for above case.

**How to simulate problem:**

Create 2  "MySensors Gateway with MQTT interface" hardware connected to the same MQTT server. "Topic Prefix Mode" have not  influence.
After enabling them watch domoticz log.
Both hardware will keep on "fighting" for MQTT connection and MQTT server will be simply disconnecting one by one:

```1576152152: Client Domoticz-MySensorsb9bc2348-483f-4e3a-9b0f-f80129080239 already connected, closing old connection.
1576152152: Client Domoticz-MySensorsb9bc2348-483f-4e3a-9b0f-f80129080239 disconnected.
1576152152: New client connected from 192.168.221.5 as Domoticz-MySensorsb9bc2348-483f-4e3a-9b0f-f80129080239 (c1, k40).
1576152178: New connection from ....
```

**Solution:**

Generate unique MQTT client ID for each single Mys MQTT connection. At current moment MysMQTT client id is global for whole Domoticz instance as  introduced in commit b89ba87 / c55e29c.
"szRandomUUID = GenerateUUID();" could be set for example inside hardware/MySensorsMQTT.cpp. 
Same issue is exist for "hardware/TTNMQTT.cpp".

Greetings

gizmocuz:
@darkManPi , please do not create new issue topics... and make a PR with your change please ;)

rwaaren:
@darkManPi  Should be solved in Beta >= V4.11569 . Can you please check and report back

darkManPi:
> .. please do not create new issue topics... and make a PR with your change please ;)
I had the mod ready at time of writing the comment.. But the environment for domoticz compilation was not ready yet. I prefer not to push something not tested.

> ..Should be solved in Beta >= V4.11569 . Can you please check and report back
It is working well now. 
For me we can close the point.

gizmocuz:
Thanks all !!

Issue #3750, opened 2019-11-23
MQTT Client Gateway and MySensors Gateway with MQTT clash


kokangit:
In summary, the problem is that domoticz connects its mqtt client gateway and its mysensors gateway to the same mqtt broker with the same client id.

Domoticz will connect twice to the mqtt broker with the same client ID. My mosquitto server will report that Domoticz is already connected and close the old connection. This is then repeated approximately every 30 seconds. And on it goes.

I run Domoticz V4.11510 in a docker container on a debian 10 (buster) host. The mqtt broker runs on the debian host.

Can this be configured differently? Is my setup faulty?

gizmocuz:
Don't know, do you have a default MQTT installation ?

I just pushed some changes that will make sure each MQTT connection from Domoticz (main/mysensors/ttn) uses it's own ID

This has been added in beta 11527, which is currently building. (Could take up to 2 hours)

Could you test with this version ?

kokangit:
Thanks for the quick response, really appreciated!
Yes, I don't have any fancy MQTT changes. Just vanilla install on my debian host and nothing special regarding Domoticz.
Sure, I'll build the beta myself and test.

gizmocuz:
Just push a correction to my previous patch...
I indeed implemented the Client ID's last week... did not know it was so strict, but it makes sense ;)

kokangit:
I built the latest in development now and it works fine! Thank you so much!

rwaaren:
MQTT hardware gets disconnected every 30 seconds when connected from two or more domoticz instances to the same broker. On V4.11528

```
2019-11-24 16:35:41.299  Status: MQTT: Connecting to PI-5:1883
2019-11-24 16:35:41.501  Status: MQTT: re-connected to: PI-5:1883
2019-11-24 16:35:41.701  Status: MQTT: Subscribed
2019-11-24 16:36:08.646  Error: MQTT: disconnected, restarting (rc=7)
```

gizmocuz:
@rwaaren , could you check the MQTT broker log why this happens ? Could it be because of the duplicate client ID as well ? I mean two instances will use the same name.
Maybe we should include a unique ID as well.
But I hope your log will tell us (like client with same ID connected.. or something)

rwaaren:
In the broker log:
from 11/24/2019 @ 2:36pm (UTC) - 11/24/2019 @ 3:39pm (UTC)
141 times 
`Client Domoticz already connected, closing old connection. `

[...]

PR #3795, opened 2019-12-14
fix for issue #3750 / #3793; use unique MySensors and TTN MQTT call IDs

rwaaren:


gizmocuz (approved):
Looking good !!

gizmocuz:
Thanks for the PR!

cocooma:
I've came across this issue today. I'm running the latest build. I had to disable all mysensor devices and enable them.
I think it was probably due to the fact that they were created while the bug was present. 

After reenabling them the reconnect issue has gone 👍

PR #3798, opened 2019-12-14
workaround for Evohome v1 API session expiration

gordonb3:
Honeywell has stripped session renewal from the v1 API (used for better temperature resolution), which causes the interface to show all zones as Offline after 15 minutes, even though they can still be controlled as this is done using the v2 API.

This workaround performs an additional call to the portal, that unlike the regular call does return an error state when our session has expired so that we can respond by resending our credentials.

I have also included an icon for the "Offline" state as this was missing.


gizmocuz:
Thank you !!

Issue #3800, opened 2019-12-17
HTTPClient: why is this line here?


gordonb3:
https://github.com/domoticz/domoticz/blob/30eab4f5d72bef18894fed14824f65a97c7cc27c/httpclient/HTTPClient.cpp#L209

The result of this is that the calling process does not have access to the returned error from a GET operation, making automated responses and sensible logging impossible.

rwaaren:
Are you sure you are referring to the right commit (30eab4f ) here ? 
HTTPClient.cpp is not touched by this commit or by any other commit  in this PR

gordonb3:
It's not about the commit but the line. GitHub just added my highlighting this line to the most recent commit. My guess is that this line has been in there since the very first version of this file.

As stated in the opening post, the line causes curl to obscure whatever is returned and that makes it impossible to feed the user any information that the website gives about the cause of the problem. The main question here is: is it safe to remove this line, or will the return of unexpected data cause some hardware module to crash Domoticz?

gizmocuz:
@gordonb3 , please provide an example url where it goes wrong, and the expected result

gordonb3:
This is related to my latest PR #3798 concerning EvohomeWeb. A user raised an issue that turned out to be related to his Honeywell account, which we discovered when I walked him through manually executing the required steps to retrieve the info. What we find was that the manual method returned an error message whereas the error from within Domoticz indicated an empty result (and also no return header).

I have not gotten around to it yet, but it is very likely that #3798 may have been (mostly) obsolete if the existing method had had access to the returned error. The reason that the method in #3798 allows the proper response seems to be mostly that `CURLOPT_FAILONERROR` is only set for the GET operation and the API requires me to use a PUT operation for refreshing the session.

gordonb3:
PS per the forum post EvohomeWeb should have received:
 
> [
    {
        "code": "LatestEulaNotAccepted",
        "message": "Latest Eula is not accepted."
    }
]

Being able to log that message would make clear where the error originates from, but the current empty return only allows a generic failure to be reported, which may be either local or remote. That adds a huge amount of complexity into troubleshooting problems.

gizmocuz:
@gordonb3 , not to fast with the PR please *g*

The reason I put this line here is that it was always returning false (!CURLE_OK) when there was a HTTP error.
For example, in your PR when you remove the line(s) with 'CURLOPT_FAILONERROR', it will always return CURLE_OK, and we don't know if there was an error anymore.

[...]

PR #3801, opened 2019-12-18
allow HTTP-4xx returns to pass to caller and review EvohomeWeb error …

gordonb3:
EvohomeWeb error handling didn't actually work as expected because HTTP-4xx returns do not cause curl to return something other than `CURLE_OK`.

Removed a blockade in HTTPClient GET method that caused HTTP-4xx returns to be empty on return. See Issue #3800


gizmocuz:
Thanks!!

PR #3802, opened 2019-12-20
Further processing of 3FE0 messages for Evohome

brucemiranda:
This change reduces the errors due to commands not processed with an OpenTherm Bridge is within an Evohome system.


gizmocuz:
Thanks!!

PR #3807, opened 2019-12-20
Implementing timeouts in ASyncTCP class.

rkluwen:
After refactoring CProxyClient to use the ASyncTCP class, it lost its timeout capabilities.
This PR re-introduces timeouts for all ASyncTCP clients. By default. it is turned off in order to not break other current ASyncTCP derivates.


gizmocuz:
Thanks!

PR #3809, opened 2019-12-22
RoomPlan page update

stas-demydiuk:
- rewrite to use angular controllers and components
- extract new device selection into modal with ability to filter by hardware and by device name
- update checking permissions flow
- update uiBootstrap to fix issue with modal
- update modal styles

image

image


dannybloe:
Can't you get rid of those horrible up/down buttons or add drag-drop to reorder as well? 🙂

stas-demydiuk:
@dannybloe, basically do you know, why do we need this reorder at all?

stas-demydiuk:
@gizmocuz any chances to have that merged?

gizmocuz:
Thanks for this,  but i miss the idx in here... is it possible to prefix the device name with this?

gizmocuz:
@stas-demydiuk , the reorder is there because it's the order it will present itself on the page
For the IDX, i mean in the popup to select the device

stas-demydiuk:
image

stas-demydiuk:
image

gizmocuz:
Thanks!!

Issue #3039, opened 2019-02-12
[feature request] Netatmo 'boiler_status' binary switch


lboue:
Hello,

Could you add in the Netatmo plugin a new binary switch synchronised with the `boiler_status `variable?
I would be able to be notified when boiler is turning on/off..

Here is my getHomeStatus json result (check `type': 'NATherm1'`)
```
{   'body': {   'home': {   'id': '**************************',
                            'modules': [   {   'firmware_revision': 197,
                                               'id': '70:ee:50:xx:xx:xx',
                                               'rf_strength': 107,
                                               'type': 'NAPlug',
                                               'wifi_strength': 26},
                                           {   'anticipating': False,
                                               'battery_level': 3753,
                                               'battery_state': 'high',
                                               'boiler_status': False,
                                               'boiler_valve_comfort_boost': False,
                                               'bridge': '70:ee:50:ab:cd:ef',
                                               'firmware_revision': 71,
                                               'id': '04:00:00:xx:xx:xx',
                                               'reachable': True,
                                               'rf_strength': 39,
                                               'type': 'NATherm1'},
                                           {   'battery_level': 3213,
                                               'battery_state': 'full',
                                               'bridge': '70:ee:50:ab:cd:aa',
                                               'firmware_revision': 73,
                                               'id': '09:00:00:xx:xx:xx',
                                               'reachable': True,
                                               'rf_strength': 82,
                                               'type': 'NRV'},
                                           {   'battery_level': 3081,
                                               'battery_state': 'full',
                                               'bridge': '70:ee:50:ab:cd:bb',
                                               'firmware_revision': 73,
                                               'id': '09:00:00:xx:xx:xx',
                                               'reachable': True,
                                               'rf_strength': 55,
                                               'type': 'NRV'}],
```

Regards,

lboue:
It should be something like it is for `therm_setpoint_mode `to test `boiler_status` variable here: https://github.com/domoticz/domoticz/blob/80fcf90238d76a81b85ad1e077d8ac6cac935a94/hardware/Netatmo.cpp#L1494

Here is proposal at the end:

```
				m_thermostatDeviceID[roomID & 0xFFFFFF] = id;
				m_thermostatModuleID[roomID & 0xFFFFFF] = id;

				if (!room["therm_measured_temperature"].empty())
				{
					SendTempSensor(roomID, 255, room["therm_measured_temperature"].asFloat(), roomName);
				}
[...]

PR #3811, opened 2019-12-22
Added Netatmo Thermostat status as Utility/GeneralPercent

olivierberthomme:
Match the feature request #3039 Netatmo 'boiler_status' binary switch

Use the Netatmo public API to parse the actual "boiler status" and push it to Domoticz to new device "Status" true==100% ; false == 0%


olivierberthomme:
Commit log is not clear as I did not merge my dev to a "staging" branch. Anyway, the goal is write the current Netatmo Thermostat state. 
The current Netatmo Thermostat state should help my to log my convector's costs.

I first tried with a switch but as the Netatmo API do not allow to overide the thermostat so the switch should be "read only". I also faced issue retriving which switch is pressed from Domoticz interface (I was tring to identify Away switch from Status switch by their ChildID but did not found how to). Then I rewrited my dev and coded the state into a PercentageSensor (which is read-only by design).

Please advise if something's unclear / badly coded.

gizmocuz:
As this is a on/off status, please use a switch
You can later change the switch type in the user interface or set you own icon
It will also be much easier to use a switch in scripting

olivierberthomme:
> As this is a on/off status, please use a switch
> You can later change the switch type in the user interface or set you own icon
> It will also be much easier to use a switch in scripting

OK Thanks for you answer.
I agree to use a basic switch but I did not find a way to identify the pressed switch from WriteToHardware function... As the hardware will get 2 switches (Away + Status) how could I know the switch pressed from interface ?
Tried: pCmd->LIGHTING2.packettype ; pCmd->LIGHTING2.subtype

Could you help me?

gizmocuz:
@olivierberthomme , you give the ID/Unit to the switch, so it should be easy to find out what switch is pressed i think ?
The ID is in LIGHTING2.id1 till LIGHTING2.id4

olivierberthomme:
I finally managed to make it works with a On/Off switch. Many thanks for your help.
Played with bitwise and bit shift operand.

Sorry again for the messy commit log... forget to squash commit while merging. If it's an issue; let me know.

gizmocuz:
Thanks!!

PR #3817, opened 2019-12-27
Fix the Evohome external sensor temperature update.

cocooma:
It was not updating the Evohome controller V2 outside temperature gage
   when the UV sensor was not present.


gizmocuz:
Thank you!

PR #3824, opened 2019-12-31
Python, Add authorization header only if no autorization already set, not onl…

guillaumezin:
…y basic authorization

Add a space after colon for inserted basic authorization


gizmocuz:
Thanks!

PR #3829, opened 2020-01-03
update index.html comboyear weather

steph10200:
Add year 2019 and 2020 for comboyear for show weather function.
Maybe it would be interesting to make this selection dynamic, or configurable.

### DIFF
--- a/www/index.html
+++ b/www/index.html
@@ -1243,6 +1243,8 @@ $(document).ready(function() {
 						<option value="2016">2016</option>
 						<option value="2017">2017</option>
 						<option value="2018">2018</option>
+						<option value="2018">2019</option>
+						<option value="2018">2020</option>
 					</select>
 				</td>
 			</tr>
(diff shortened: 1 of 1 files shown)


gizmocuz:
Is it possible to compare this against the temperature and or the electic meter year combo’s ?
Maybe these are dynamic and you could adopt the code?

rwaaren:
@steph10200   

> Maybe it would be interesting to make this selection dynamic, or configurable.

Hope you are not offended by my  PR #3851 for the same issue.

steph10200:
> 
> 
> @steph10200
> 
> > Maybe it would be interesting to make this selection dynamic, or configurable.
> 
> Hope you are not offended by my PR #3851 for the same issue.

Indeed, I had not seen your patch. 
yours is better.

gizmocuz:
@StefanPuntNL , thank you for this PR !! Much appreciated!
I do prefer the dynamic solution, but please continue to submit feature PR's !

Issue #3822, opened 2019-12-30
HTTP connection crashes plugin framework depending on buffer contents


bloob00k:
CPluginProtocolHTTP::ProcessInbound assumes in many place that tokens are completely contained within a received buffer, particularly for chunked encoding.  If that is not the case
then the code may incorrectly parse the data stream, or believe it has  received the full response when actually there is still more data to come.   Depending on the circumstances this will cause the plugin framework to crash either immediately or on the next attempt to parse and collate an incoming message.

Examples:
1.  Terminating \r\n for a chunk is split across two receive buffers.  Code assumes either both chars are in the first buffer, or neither are.

```
2019-12-27 14:17:53.531  (dummywiser) Processing 'ReadEvent' message
2019-12-27 14:17:53.531  (dummywiser) Received 97 bytes of data
2019-12-27 14:17:53.531  (dummywiser)     48 54 54 50 2f 31 2e 31 20 32 30 30 20 4f 4b 0d 0a 43 6f 6e     HTTP/1.1.200.OK..Con
2019-12-27 14:17:53.531  (dummywiser)     6e 65 63 74 69 6f 6e 3a 20 63 6c 6f 73 65 0d 0a 43 6f 6e 74     nection:.close..Cont
2019-12-27 14:17:53.531  (dummywiser)     65 6e 74 2d 54 79 70 65 3a 20 61 70 70 6c 69 63 61 74 69 6f     ent-Type:.applicatio
2019-12-27 14:17:53.531  (dummywiser)     6e 2f 6a 73 6f 6e 0d 0a 54 72 61 6e 73 66 65 72 2d 45 6e 63     n/json..Transfer-Enc
2019-12-27 14:17:53.531  (dummywiser)     6f 64 69 6e 67 3a 20 63 68 75 6e 6b 65 64 0d 0a 0d .. .. ..     oding:.chunked...
2019-12-27 14:17:53.531  (dummywiser) Pushing 'onMessageCallback' on to queue
2019-12-27 14:17:53.531  (dummywiser) Processing 'ReadEvent' message
2019-12-27 14:17:53.531  (dummywiser) Received 6 bytes of data
2019-12-27 14:17:53.531  (dummywiser)     0a 30 0d 0a 0d 0a .. .. .. .. .. .. .. .. .. .. .. .. .. ..     .0....
2019-12-27 14:17:53.531  Error: PluginSystem: Exception processing message.
```

2.  Chunk size and its terminator are split across two buffers:
```
[...]

PR #3830, opened 2020-01-03
Fix multiple crashing bugs in plugin HTTP protocol handler (issue #3822)

bloob00k:
CPluginProtocolHTTP::ProcessInbound assumes in many place that tokens are
completely contained within a received buffer.  If that is not the case
then the code may read beyond the end of the buffer, or believe it has
received the full response when actually there is still more data to come.
Depending on the circumstances this will either cause the plugin framework
to crash immediately, or on the next attempt next to parse and collate
an incoming message.

Examples of receive buffers which cause these crashes are in the issue itself.  The changes here fix them.

### DIFF
--- a/hardware/plugins/PluginProtocols.cpp
+++ b/hardware/plugins/PluginProtocols.cpp
@@ -401,6 +401,13 @@ namespace Plugins {
 		m_Chunked = false;
 		m_RemainingChunk = 0;
 
+		// Need at least the whole of the first line before going any further; otherwise attempting to parse it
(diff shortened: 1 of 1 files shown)


bloob00k:
I thought you wanted PRs based on master, but everyone else seems to be doing them to development; maybe I misunderstood.  Let me know if you want it on development instead.

rwaaren:
>  Let me know if you want it on development instead.

Yes PR's should be done to development branch

PR #3837, opened 2020-01-06
Decode and display the Evohome OT messages

brucemiranda:
Currently they are displayed only on the Status log


gizmocuz:
Thanks!

PR #3841, opened 2020-01-09
All Evohome OT messages decoded and associated devices created

brucemiranda:


brucemiranda:
This PR now also creates devices for the decoded messages using the exact same naming and method that the OTGW Hardware uses. That way the implementation is kept the same. Two messages are not published to devices again because OTGW doesn't do it and there is not too much benefit in tracking the values of them.

brucemiranda:
Could someone check and tell me why my code compiles on Ubuntu and not on Windows?

gizmocuz:
Thank you for this great patch !!

PR #3851, opened 2020-01-16
Fix for missing years in combobox. Selectable years are now dynamic based on current year.…

rwaaren:
Could not find anywhere in the code for temperature or other places how this was done there but got this from other source; tested it extensively and it does work and does not seem to break anything else.
Having said that this is my first PR in domoticz in HTML / JS , so I could well be violating one or more design / style rules. If so please advise on what needs to change.


gizmocuz:
Thanks !! I think this was for the rain device only ?
In the end it would be great if it was done via the DeviceReport controller ;)

rwaaren:
Welcome back and yes; rain device only.
Will try to find (and understand :-) ) DeviceReport controller.

PR #3857, opened 2020-01-21
EvoHome: Decode Fault Logbook and print the messages on the Status Log

brucemiranda:
Also removed the conditional creation of the new devices so that a new value of 0 is updated on the devices.


gizmocuz:
Thanks!!

PR #3876, opened 2020-01-29
Reworked blebox integration

gadgetmobile:
Features of this PR:

### BleBox specific:
- 11 BleBox products supported
  * more features available (e.g. kick in airSensor to force measurement)
  * improved support for slider controls and buttons
  * immediate state updates in Domoticz when interacting with devices
  * newer/recommended Domoticz controls used   
- reworked hardware setup ui
  * UI is more intuitive (IP assigned to each product / "hardware", and "devices" are product features)
    - "features" instead of "nodes" which allow selecting device capabilities
    - device features can be selected conveniently for every BleBox product separately
    - "save" button which makes selected features immediately available
    - new (future) features can be automatically detected ("features" setup page)
  * very convenient and fast *network search* (as an asynchronous button)
    - auto detects best default network mask for searching
    - network search can fix IP problems (e.g. update IPs correctly after a router change)     
  * thorough and comprehensive error handling (helping immediately identify the cause with sufficient info from logs to fix it permanently, without much hassle from the user)
 
### Domoticz improvements (demonstrated using BleBox integration as an example)
- integration tests (end-to-end through Domoticz mainworker) using Google Test/Mock
  - any breakage in Domoticz internals affecting BleBox integration will be detected automatically
  - tests take only milliseconds to run
  - tests are separate from build (less risk of breaking builds)
  - possible to simulate commands send from UI
  - lots of conveniences for automated testing, e.g. comparing Domoticz db state against JSON
  - tests are thorough enough for development, refactoring and testing without physical access to
 BleBox devices
  - tests are not BleBox specific (they would be almost identical for any HTTP-based devices)
  - in-memory DB with full Domoticz tables schema loaded and reset before every test
  - extensive list of stubs (so Domoticz parts included in tests can be selected without requiring Domoticz to become modular)
  - tests compile on modern compilers (c++17 on Clang and Gnu G++) - with specific tweakable options for each
  - tests support unity/jumbo builds (currently not set up as options, but unity through Cotire worked)
* separate, organized CMake modules used by tests (to avoid a large, unmaintainable CMakeLists.txt file)
* stubs for 99% of Domoticz, which allow selectively testing, stubbing and monitoring of almost any Domoticz feature during tests 
* generic handling of HTTP/JSON-based integrations, even with complex GET/POST requests and responses (including schema checking and field extraction helpers)
* rfx value compatibility checking (prevents user for messing up device state by switching to incompatible devices)
* convenience wrappers for Domoticz specifics
* automatically selecting the best Domoticz widget types for given function (e.g. "0-100% slider", "push button")
* elegant and modern architecture to make development, maintenance, testing and understanding easier (everything is designed to be self-explanatory)
* BleBox integration as a separate cmake subdirectory (with it's own set of warnings and options)
* non-invasive changes (almost no changes in Domoticz itself, despite heavy reliance on matching new records with complex Domoticz internals and specifics)
* improved handling of web-originated commands (less code and with tracing)
* annotated DB queries (to know which query came from exactly which method - and to distinguish between internal Domoticz SQL queries from those run by integrations)
* easy to track and translate exceptions, carefully placed and organized into a type hierarchy (for better handling)
* mock'able session class tied to an IP
* separate type to handle IP values (uses boost)
* carefully typed usage of RFX values (including uint8_t, std::byte and appropriate warnings enabled - all to avoid mistakes that cost lots of debugging time)
* precompiled header support that can be specific to integrations (CMake 3.16 required)
* integrations do not require to include old, Microsoft-specific stdafx.h (and have their headers ordered appropriately) to build

### Changes in requirements
- gcc-7 is a minimal requirement (since it's the first version to support C++17 for std::byte)
- cmake 3.16 (for precompiled headers and unity/jumbo builds)

Please feel free to comment, ask questions and share suggestions.


gadgetmobile:
@szczukot - you are the original author of the BleBox integration (which the above is based on), so any suggestions from you would mean a lot.

Thanks!

gizmocuz:
@gadgetmobile ,
Thanks for all your work!!! Here some feedback from me:

- No need for a new subtype for AirQuality, just forget the naming of the subdevice, this might soon drop
- RFXtrx,h, not allowed to edit this file, and sure we already have enough temp sub types you can choose from

- SQLHelper.cpp, no special include of blebox migration, maybe a quick delete of all blebox related things in the database and start again ?

- Tests, great they are there, but as they are hardware specific (not Domoticz specific) please do not include any in the Domoticz project.
It is making the project also larger
Maybe we could make blebox a sub module/project of Domoticz in the repository or it's own repository, and use a (C)make file flag to include it ? (Or when it is found, include it)

- Hardware Setup, Could you post a small screenshot what has been changed ? (Not talking about BleBox changes, but you made some global changes)
Could you move more BleBox code from Hardware.js to www\app\hardware ?
This will make it structured... (yes i/we need to do this for lots of other hardware too)
(BleBoxAddUserHardware/BleBoxDetectHardware/BleBoxUpgradeFirmware/BleBoxInit...)

- CMake file, i see some separate cmake files now, but i do not see them used by looking at the changes in CMakeList.txt
Are they needed?

Once again, much appreciate all the work !!

gadgetmobile:
Thanks!

First of all, I have lot of respect for maintainers of Domoticz.

I respect your time (maintaining hundreds of integrations is tough) and so I wanted to create a high quality integration.

This wasn't possible with just a few lines of code.

And with lots of code it would take a lot of time to review.

So my next best option was to keep things as non-intrusive and separate as possible.

But that then takes even more code. (Although it's separate and generic).

The risk there is: maintainers could look at the amount of code and misjudge (and think e.g. "this guy wrote too much code for me to read, so I'll just say we don't want so much hardware-specific stuff in our project").

It turns out that BleBox integration is trivial (simple HTTP/JSON). It's Domoticz integration that's hard.

That's why 99% of the code here is Domoticz specific. (Convenience wrappers and Domoticz workarounds).

You could swap the tests with any other HTTP-based integration - and most of this PR would be exactly the same.

This PR is actually a gigantic, hardware-independent add-on for Domoticz. (With BleBox integration just as a bonus usage example).

It's mostly a set of Domoticz-specific development tools useful to ANYONE developing or maintaining Domoticz integrations.

Yes, there's probably a lot to look at, but this PR is 99% for the benefit of Domoticz, not the integration.

Even if I added tests for another integration, 99% of the code here would be the same.

Here are points of interest to demonstrate this:

1. `CMakeLists.txt` -> the new integration no longer fits in just `BleBox.cpp`, thus a separate `hardware/blebox/CMakeLists.txt`. No changes otherwise. (Tests manage their own dependencies using refactored cmake/*.cmake files, which is how Domoticz CMakeLists.txt itself should be refactored).

1. `hardware/blebox/CMakeLists.txt` -> any generic integration should have it's own files and warning options, etc.

1. `hardware/blebox/rfx.cpp` -> generic convenience wrapper for Domoticz dev/stype/swtype ids

1. `hardware/blebox/widgets.cpp` -> generic wrapper mapping Domoticz UI commands/values for generic HTTP/JSON responses

1. `hardware/precompiled.hxx` -> each hardware can maintain it's own precompiled headers

1. `hardware/blebox/db/device_status.cpp` -> a simple generic ORM for Domoticz devicestatus table

1. `hardware/blebox/db/hardware.cpp` -> mostly a generic ORM for Domoticz hardware table

1. `hardware/blebox/db/migration.cpp` -> how to avoid polluting SQLHelper with migration commands (and how to test migrations)

1. `hardware/blebox/db/annotate.cpp` -> to decorate SQL commands, so you can tell which ones come from a specific integration (and which method)

1. `hardware/blebox/version.cpp` -> simple generic SemVer version compatibility checker (to avoid more dependencies in project)

1. `hardware/blebox/session.cpp` -> a generic, mock'able HTTP/JSON/TCP session wrapper tied to an IP

1. `hardware/blebox/errors.cpp` -> separate list of used exceptions to avoid polluting sources with untranslated text and errors and for easy review

1. `hardware/blebox/web.cpp` -> generic way to map pass Web-originated to comands to worker methods

1. `hardware/blebox/widgets/widget.cpp` -> a generic way to map JSON fields to Domoticz device fields

1. `hardware/blebox/widgets.cpp` -> a generic way to map "universal types/values" (e.g. "a 100% slider") to actual Domoticz recommended types for UI

1. `hardware/blebox/box.cpp` -> first of the ONLY TWO BleBox-specific files in the whole PR (and yet most of it is generic JSON response processing)

1. `hardware/blebox/products.cpp` -> second of the ONLY TWO BleBox-specific files in the whole PR - pretty much like an XML file describing a product's "features" in terms of "widgets" (which widget.cpp maps to Domoticz recommended widgets for that use case)

So only 2 small files are actually hardware-specific.


I think the most understandings come from the tests.

Let me repeat myself:

This PR is actually a gigantic (explained in a moment), hardware-independent add-on for Domoticz - with BleBox integration just as a bonus usage example.

Here's the list of what's worth noting in tests/

1. `tests/README.md` -> simple instructions to build tests (separate from main build to avoid breakage risks)

1. `tests/CMakeLists.txt` -> generic way to build an executable test file (BleBox specifics are added with `add_directory`)

1. `tests/extern` -> a checkout of googletest/googlemock (this is the only recommended and practical way of using googletest/googlemock)

1. `tests/memdb.cpp` -> generic, creates a replica of Domoticz database in memory, and resets the database to clean state before every test

1. `tests/worker_helper.cpp` -> a generic helper to manage the Domoticz main worker and simulate calls from Web UI (mostly `MainWorker::SwitchLight`)

1. `tests/test_helper.cpp` -> a generic helper to make it easier to compare Domoticz database state with expected results

1. `tests/stubs.cpp` -> a gigantic, 145kB file stubbing all the devices - just to Domoticz MainWorker can be used in integration tests without code modifications (almost everything HAD to be stubbed, except things like `main()` in main/domoticz.cpp of course). If Domoticz were more modular, this file wouldn't be needed at all. Otherwise, MainWorker pulls in 99% of the project (including e.g. IFTTT) just to be used. Domoticz is 100% monolithic and non-reusable. (That's why this file exists and is so big).

1. `tests/hardware/blebox/products/*.cpp` -> supposedly BleBox specific tests, but they're all almost identical (except for product names and Domoticz ids). To add a tests for any HTTP/JSON based integration, the tests would be almost exactly the same.

1. `tests/hardware/blebox/products/fixtures/*.cpp` -> sample responses from real BleBox devices - thanks to these, this integration can be developed without needing access to real BleBox devices.

1. `tests/hardware/blebox/db/migration_test.cpp` -> shows how to test a complex migration - the setup (using old code to create the records) and comparing results using the test helpers shows how migration edge cases can be easily guaranteed to work in practice


Just look at e.g. `tests/hardware/blebox/products/airSensor_test.cpp`

Most of it is checking if the Domoticz-specific device_status values are correct, so the test is Domoticz specific.

Making the tests separate would make no sense at all. (You'd need a copy for `main/mainworker.cpp`, `main/SQLHelper.cpp`, `main/RFXNames.cpp`, etc, to get the integration tests to even compile! It makes no sense to copy those files to a new project - you'd need a git submodule of the whole Domoticz project!).

As you can also see, you can easily replace the test with any HTTP/JSON based integration and the code would be almost the same.

So the REAL thing being tested here is Domoticz itself. (`MainWorker` and `CSQLHelper` mostly).

Also, tests help confirm a developer's understanding of Domoticz internals (which aren't usefully documented otherwise - from a integration developer's point of view).

I hope the above summary I especially prepared will help you appreciate how much this PR helps revive Domoticz as an exciting project more than the "bonus" hardware integration it uses as an example.

I'm sure finding really good maintainers (who have experience with Domoticz internals and C++) is challenging, and rejecting such improvements is probably the most discouraging thing you can do. (Since experienced developers often look for quality unit tests as a signal for getting involved or not).

So these tests are Domoticz tests, and are not "hardware specific" in any way.

If you care about Domoticz being a viable development/maintenance platform, please don't carelessly ask me to remove things that benefit such development and Domoticz itself. (If you have concerns, avoid quick judgement and just ask instead).

I understand you may have many angry users who don't understand technical reality of such projects. Just make sure you don't accidentally treat thoughtful and careful developers as you would people just learning to code.

Instead, please ask me questions about how I handled your concerns (instead of asking me to drop stuff that "seems too hardware specific", which never even was intended that way).

I understand this is a big PR. But it's also potentially an enormous benefit to Domoticz.

So I'd say investing time to look into this will pay off quickly.

Also, the PR is so non-invasive, that merging it would help many users with no downside at all.

(As for the code size, 145Kb is for stubbing out Domoticz functionality simply because Domoticz is 100% monolithic and can't be made modular without major code changes - it's proof of bad code architecture in Domoticz, and it's a non-invasive workaround, which can be considered almost magic how the MainWorker is used from the monolithic app without code changes).

gadgetmobile:
## New types (airSensor/tempSensor)

Just to clear things up where needed:

I REALLY wanted to change as little as possible (e.g. nothing) and reuse existing types (which I did, initially).

It was frustrating being constantly asking "Why is it Voltcraft and not BleBox? What is Voltcraft? Is that some kind of bug? Can you fix it?".

I didn't want users to think Domoticz had such poor quality that not even a device name can be changed. (I'm technical so I understand how hard it is to support hundreds of devices, etc. - but explaning that to users doesn't work, because they can't believe that changing a description is hard).

So I added those types afterwards - just to avoid constantly trying to answer in a way that doesn't suggest poor quality somewhere.

Ideally, I wanted to use `pTypeGeneral` with `sTypeTemperature` (for tempSensor), but this didn't seem implemented.

I don't mind the shared logic (I'd prefer it actually), so here's what I'd suggest:
 - adding an enum `sTypeAirQuality` with the same value as `sTypeVoltcraft` and changing name+description in `RFXNames.cpp` (since it only affects UI)
 - same for sTypeTemperature enum alias for sTypeTEMP5 ("LaCrosse TX3" -> "Temperature")

This doesn't affect the logic, Voltcraft/LaCrosse branding isn't hurt and users aren't confused.

(I can add separate PRs for both - just let me know if you're ok with this).

gadgetmobile:
## Migration

> SQLHelper.cpp, no special include of blebox migration, maybe a quick delete of all blebox related things in the database and start again ?

That's what I did in development mode - delete every time. I realized that the previous integration is so different, that it could cause weird errors and maybe crashes in Domoticz on startup due to unforeseeable edge cases. So a migration must remove the old records before the workers start anyway (to avoid possible crashes and unnecessary bug reports and frustrated users). So since a migration is the best solution (for users and maintainers), I created one.

It removes records (less risky) and creates new records with old device names (less confusing for the user). The actual migration is in a separate file `blebox/db/migration.cpp`, where further fixes/changes can be added without changing SQLHelper.cpp.

In short, I didn't want to change anything in SQLHelper either. It's just that a migration creates a much smaller risk of maintainer/user time/frustration. (I've designed the rest of the integration to make sure more migrations won't be needed in the future).

I considered "deleting records" to be a disregard for both users and maintainers. (That's why e.g. this new integration fixes bad IP addresses - when e.g. the router is changed or something).

Even in case of migration errors, details are logged and transactions roll back the changes, so a Domoticz update can be released to retry the migration - otherwise more uses could have problems and experience unnecessary frustrations with Domoticz, which they would vent on maintainers.

It makes no sense to purposely revert to no migration - unless you want to frustrate users on purpose.

I'm sure users aren't complaining "I'm not using Domoticz because there's a migration for a specific hardware in SQLHelper!". And I'm sure people aren't discouraged to become contributors for that same reason. (I'd say a lack of unit tests is a bigger source of discouragement, especially when it comes to professional developers).

It's also not a maintenance burden - it's a necessary change because of unfortunate decisions I didn't make. (Which this PR tries hard to undo without breaking anything for anyone).

The correct way to handle any migration is to rollback on errors, so a fix can be released quickly and the migration restarted. And prior migration tests are such a bonus to a project.

In any case, the schema should be in a separate file (e.g. `Schema.cpp`) and the migrations too (e.g. `Migrations.cpp`). Sure, more files would e.g. increase build time a little, but that's because precompiled headers are broken in Domoticz (a different issue) and e.g. unity builds aren't possible without major code changes. (Which would likely be reject just because they're ... "major code changes").

And I understand that a large codebase is harder to maintain with more code, while cleanup can break things. But then more unit tests and coverage fixes that problem.

No one will want to cleanup Domoticz (lots of work) if there's a high chance you'll reject the changes.

Also adding changes requested in reviews is a horrible experience if you have to test everything manually all over again after changing a simple line.

Development in Domoticz has been an absolutely horrible experience, and instead of complaining I invested time to make it better. Telling me to drop what really helped me make that experience better - that's really discouraging, especially since it's a huge red flag suggesting a lack of professionalism.

I'm just saying this for the future if you want to attract careful maintainers who want to help (but not be expected to waste enormous chunks of their time doing work that can be automated or preventable).

gadgetmobile:
> Maybe we could make blebox a sub module/project of Domoticz in the repository or it's own repository, and use a (C)make file flag to include it ? (Or when it is found, include it)

That should be the case with every integration.

Ironically, it doesn't make sense here, because of these reasons:
1. Most of this PR is generic (except `products.cpp` and parts of `box.cpp`), so extracting "generic Domoticz wrappers" to a new project makes no sense. (Why not e.g. put Domoticz's `Config/*` files in another repository?)
2. Some integrations deserve their own projects a lot more than this. BleBox is simple HTTP/JSON comunication, where even the HTTP and JSON is handled by Domoticz provided services.
3. The unit tests are pretty much just checking if `Domoticz` calls `UpdateIntValue` with the correct parameters (to prevent duplicate records) and if device values a properly propagated through Domoticz into the database.
4. Any HTTP/JSON based integration would share/require most of the same code. And that code is Domoticz specific. (At some point, BleBox integration could be reduced to an XML file - though currently the work required to do this won't pay off - see `box.cpp`, where this is partially done).
5. The inclusion of hardware workers in Domoticz isn't generic or "plug and play" - BleBox would be the exception and it's the least viable candidate to be an exception.
6. It would make more sense to extract out existing integrations the way you want. (And then I could reduce that giant `stubs.cpp` file.) Then, I'd know exactly what you'd want and I'd adapt this PR to match.
7. Currently there's a huge risk on my side that I'd put in lots of time to do this properly and you'd reject it for reason not well explained. I'm not immortal with that much time on my hands. `Stubs.cpp` is an example of me bending over backwards to side-step the issues of all the current integrations unconditionally hardcoded into Domoticz.

I understand you don't want "more integrations", and this PR does lots of things to reduce the maintenance burden of future integrations.

Just notice that `products.cpp` sufficiently covers the integration of 11 (!) different devices with different features.

You can't ask for LESS code for handling the integration of 11 devices. (Even the XML's in Config are much bigger and more complex - and not covered by unit tests).

I think the core issue is that this isn't a standard PR where someone is "adding support for a device".

Instead, you'd likely expect to merge more ideas/workarounds/approaches from this PR into Domoticz itself.

gizmocuz:
Okey, maybe let's start by splitting this PR in two different PR's
- Domoticz specific
- Blebox specific

About the notation of the temp/air sensor, I was thinking about removing it in the feature
A temp sensor is a temp sensor, and there is no need to know if it was Model-Y
(Else we end up with 1000 temp sensors in Domoticz in the feature, we might make the sub-type text based to avoid this in a later stadium)

gadgetmobile:
> Okey, maybe let's start by splitting this PR in two different PR's
> 
> * Domoticz specific
> * Blebox specific

I don't see the point. What would that accomplish?

If you pretty much ignore `products.cpp` and `box.cpp` (and www/BleBox stuff obviously), most of the code (and the way tests are done) can be reused by any other integration (and it's tests). Until there are more tests, the BleBox example is the only one using them, so there's no point in refactoring.

99% of the code is to help develop integrations with tests just like BleBox - and where it's useful to put e.g. JSON field validators into Domoticz makes more sense only when more integrations would use them.

The point is: if any of the generic conveniences in this PR are worth introducing into Domoticz helpers, that can be done by maintainers later - with no risk of breaking BleBox functionality. (Thanks to the tests).

For example, to rfx.cpp class is useful when you want a specific, consistent behavior in Domoticz, even if recommendations changes (e.g. pLighting2 -> pGeneralSwitch), without having to change  anything in the integrations using it.

So if BleBox is using a specific convenience, it makes sense to keep it in the BleBox folder. (Moving files from one directory into another is not a risky repository change).

Meanwhile, currently, SQLHelper contains methods used by only one integration. (I don't remember which), which makes the file huge, unreadable and the methods are hard to reuse if they need tweaking.

The idea: extracting conveniences from this BleBox example is easy, tests assure BleBox functionality isn't broken by such refactorings, and you can change whatever you're extracting in any way you want. (Without requiring PRs from me, asking me to blindly make distinctions and separations which make no sense to me).

As long as the tests pass (reflecting what happens in reality), I don't really plan to change anything this PR delivers. I really only plan to add stuff to `products.cpp` and `box.cpp` to handle more features/products as users demand them. (And tests obviously, since they're affected by adding features).

> About the notation of the temp/air sensor, I was thinking about removing it in the feature

I don't understand what you mean by "in the feature". There's a generic `_tAirQualityMeter` class I'm using (for `pTypeAirQuality`) and I don't care about the subtype, as long as it works and I don't get user questions about the weird device name. So can I make the enum alias as I described as a compromise?

> A temp sensor is a temp sensor, and there is no need to know if it was Model-Y

I agree. It's just that currently during setup it displays "LaCrosse TX3", which is wrong, misleading, distracting and wastes life in pointless explanations. Changing that requires changes anyway, so we might as well add a generic temperature subtype called e.g. "generic".

Pretty much the other temperature types are used for debugging output only anyway, so I can remove that too if you want. (`mainworker.cpp`, which I almost know by heart at this point).

gadgetmobile:
> CMake file, i see some separate cmake files now, but i do not see them used by looking at the changes in CMakeList.txt
> Are they needed?

`add_subdirectory` in CMake is how those `CMakeLists.txt` in subdirectories are loaded.

I used `add_subdirectory` both in the main `CMakeLists.txt` (to separate BleBox integration and the convenience wrappers only used by that integration for now) and in the `tests` one (which includes `tests/hardware/blebox` tests just like it could include any other implementation-specific tests).

Some dependencies also use this feature to build build-in dependencies, e.g. MQTT.

The problem with building BleBox integration separately and then "optionally" would mean you'd need hundreds of CMAKE build options (like `USE_BUILTIN_MINIZIP`) - one for every integration you want to build or not. It wouldn't make sense for users to turn integrations on or off (faulty ones can be disabled anyway) because there'd be too much to configure. You could have groups, e.g. "serial hardware", "tcp hardware" and "usb hardware" which could make more sense, but it's quite a lot of work and build testing regardless. (With not much payoff).

gadgetmobile:
> Hardware Setup, Could you post a small screenshot what has been changed ?

These are the extra controls in "BleBox mode" (available only either when BleBox is selected from dropdown combo, or when a table record is selected for a BleBox device).

blebox_detect

I just reverted the change with the "update" button (I'll explain in a moment). Meanwhile, the above is just a div that's shown/hidden based on that "BleBox mode", so it doesn't interfere. (It doesn't look great, but I'm not a web designer).

As for the commit I reverted, it fixes this UI issue:

normal_add

The problem is the "Update" button at the top (under the table) should be beside the "Add" button at the bottom. Because the user can select something in the table, change using the inputs below and then go click the "Update" button.

That's because both "Update" and "Add" are like form submission buttons. It's confusing. (That's what the reverted commit does - puts them back together).

Ideally, the user should have an "add" and "edit" button which can pop up input fields to edit.

Domoticz has almost no usage of angular features and no asset pipeline (and no development improvements like live reload). It also doesn't use a CSS framework that helps with layouting - way too many things to fix and beyond the scope of this PR.

I considered integration tests using the browser, but this makes no sense without a proper asset pipeline (and a convenient way to stub out the complex ajax calls). Also the UI doesn't have a decent architecture, which makes adding tests a nightmare to create and maintain.

> Could you move more BleBox code from Hardware.js to www\app\hardware ?

If you mean to move that code to a separate JS file, it depends if you like the solution of loading a separate JS via the browser at runtime. (Normally, you'd use `require()` or `import()` which would compile, minify and concat JS assets for the browser). This is beyond the scope of this PR.

And it's the same with this PR: it lays out the groundwork necessary for many improvements. Once the groundwork is there, no one else has to start from scratch.

So the point is to pull in something that can be used as a base - and not try to clean up first without having the tools in place (that help with the cleanup).

Cleaning up C++ code quickly and reliably requires unit tests, modular code, tools and an example to test on (like BleBox implementation). I achieved that without requiring Domoticz to be modular, which is a technical miracle. (That's why the idea of throwing away the miracle and getting rid of useful tools "to reduce the code" is such an incomprehensible idea).

I can't do the same for angular I'm afraid. It's probably better to start a fresh modern React/Angular project/pipeline and add the existing features in a clearer way (so we can have separate JS files for each hardware/component/whatever and develop with live-reload changes and all the other conveniences lacking here).

Again, that's way beyond the scope of this PR. The best I can do is have the functions well-names and easy to extract later (once things are in place). I tried to add "hooks" where needed (where those functions are called).

gadgetmobile:
## Web UI changes

I just pushed a new set of commits so that the changes in `www/app/hardware` are easier to see (along with some fixes and undoing unnecessary whitespace changes):

1. `www/app/hardware/Hardware.html` - added 2 `div`s, hidden on startup, activated only during "BleBox" mode (as described above). Once div to add input+button for network search and button for  upgrading firmware. These changes do not affect any other integrations.
1. `www/app/hardware/Hardware.js` - hiding the above 2 divs on init, and 6 places for "hooks" to use blebox-specific functions (that only activate in BleBox mode). Functions for these hooks are all in one place at the end of the file. I wish I could easily add to a separate file, but the angular setup in Domoticz and caching is something I'm not comfortable touching. This also doesn't affect anything other than BleBox device setups. 
1. `www/app/hardware/HardwareSetup.js` - just to add a `Save >` button on the opposite far right side opposed to the `< Back`. This button makes the user-selected BleBox features as `used` in the database. (For convenience since BleBox devices often have a lot of features, which may require many separate BleBox widgets to represent). This also appears only on the BleBox setup page and nowhere else. (Button is hidden during init in `BleBox.js`).

I believe I've addressed all the issues.

Let me know if you want me to proceed with the air/temp sensor text/enum changes as described earlier.

Again, as for further refactorings, they can probably be done after the PR, since they way I've organized things, it should only result in easy to fix cmake/compiler errors.

gadgetmobile:
After a successful Travis build, I reorganized the commits for clarity.

I can add the tests (build + running) to Travis also in the PR if you want. (Pretty much the same steps as in `tests/README.md`).

The AppVeyor build requires changes to the Visual Studio solution. I don't use Windows, so I believe the following changes are needed:

1. Adding the missing files from `hardware/blebox` to the solution.
2. Setting the options for those files to not use precompiled headers (or to forcefully including them with some magical Microsoft C++ option, which should work also).

I believe cmake can generate a Visual Studio solution - if that helps.

For now I've ran out of things necessary to fix in this PR.

If you want to see only the changes in Domoticz (and not in the integration), you can do so on a local git checkout with:

`git diff --diff-filter=M upstream/development.. -- ':(exclude)*/BleBox.*'`

.travis.yml
```diff
@@ -73,7 +73,7 @@ matrix:
       - cd ..
       - (cd open-zwave-read-only; make)
       - (cd open-zwave-read-only; git checkout 76e21d80a140c321413a46a2b62ea6c11dd6ace0; make)
-      - cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_LIBRARY_PATH=open-zwave-read-only CMakeLists.txt
+      - cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_VERBOSE_MAKEFILE:BOOL=ON -DCMAKE_LIBRARY_PATH=open-zwave-read-only CMakeLists.txt
```
gizmocuz: Can you undo the 'CMAKE_VERBOSE_MAKEFILE' option here

.travis.yml
```diff
@@ -39,15 +39,15 @@ matrix:
     compiler: gcc
     env:
           - TARGET_ARCHITECTURE=x86_64
-          - MATRIX_EVAL="CC=gcc-6 && CXX=g++-6"
+          - MATRIX_EVAL="CC=gcc-7 && CXX=g++-7"
```
gizmocuz: We still have users on Stretch where the current GCC is version 6
Also the Arm-x86 build system has GCC-6 as there are still a lot of raspbian users running on stretch
I am (also) not to happy about this, but after a new stable release we could upgrade to 8

hardware/blebox/CMakeLists.txt
```diff
@@ -0,0 +1,152 @@
...
+    -Wall
+    -Wextra
+    -Wno-unused-parameter
+    -Wno-missing-field-initializers
+    #-Wno-parentheses
+    )
+endif()
+
+# use cxx-17 for more help from clang-tools
+target_compile_features(blebox PRIVATE cxx_std_17)
```
gizmocuz, line 81: Not sure if this would work on all systems.... we just upped to c++11

main/SQLHelper.cpp
```diff
@@ -13,6 +13,8 @@
 #include "../sqlite/sqlite3.h"
 #endif
 #include "../hardware/hardwaretypes.h"
+#include "../hardware/blebox/db/migration.h"
```
gizmocuz: As the 'migration' is only done once and there is not much code, please do it in this file.
No need for a separate migration ...

hardware/blebox/workarounds/colorswitch.h
```diff
@@ -0,0 +1,4 @@
+#pragma once
```
gizmocuz: Just drop all workarounds, use the normal functions in DomoticzHardaware.h when possible
This is crazy

main/RFXtrx.h
```diff
@@ -1062,6 +1062,7 @@ SDK version 4.9
 #define sTypeTEMP9 0x9  //RUBiCSON
 #define sTypeTEMP10 0xA  //TFA 30.3133
 #define sTypeTEMP11 0xB  //WT0122
+#define sTypeTEMPBleBox 0xC //BleBox tempSensor
```
gizmocuz: As mentioned, do not add new types, there are enough
And you can not touch this file as it is not owned by us

tests/CMakeLists.txt
```diff
@@ -0,0 +1,248 @@
+cmake_minimum_required(VERSION 3.16.0)
```
gizmocuz, line 1: Please remove all tests

gizmocuz (changes requested):
Still have some comments here...
Also try to avoid using boost as much as possible and use the native functions
As the 'setup' probably will not change that much, I still think it could be a good idea to make this a sub-module ?

.travis.yml
gadgetmobile: reverted in 381f64aee

(Will be squashed/rebased later to avoid messing up review process).

.travis.yml
gadgetmobile: - reverted in 850a2bbb5
- gcc-6 support fixed in fe60bf1ea
- workaround header can be removed once gcc is upgraded

(will be squashed/rebased later)

hardware/blebox/workarounds/colorswitch.h
gadgetmobile: - missing includes added in a3772e3c9 (will be kept as separate commit)
- added proper cereal handling (allows using external cereal) in e34b0c69f (as separate commit)
- workarounds removed in 95f0236e7 (will be squashed/rebased later)

> use the normal functions in DomoticzHardaware.h 

I don't know what you mean. I tried to use as much as possible from existing codebase, unless changes were needed that could potentially interfere/break other integrations.

The `Send*` methods were especially problematic, with weird/inconsistent handling of fields/values and DeviceID/Unit values, that caused duplicate records to appear (which was a big problem even in the previous integration here) - this lead to lots of partially initialized records in the database, which was a horror to debug and fix. Of course, any necessary fixes could have led to breaking other integrations, which obviously wouldn't be acceptable. (Even the previous integration reimplemented one of the Send* methods for the same reason).

Reimplementing those methods to match the `decode_*` methods was safer, more robust and easier to track and test.

Overall, the Send_* methods are a bad idea overall, and the whole API should be based rather on structures for type safety (like the `_tGeneralSwitch`) to avoid mismatching Send_* methods upon switching widget types. (Which was a problem when migrating the old implementation away form pTypeLighting2).

There's a strong reason for everything I did - my primary approach was to reuse as much as possible (using "the Domoticz way" where possible) and change Domoticz internals as little as possible (ideally nothing, since other integrations could be relying on broken functionality).

If it's possible to reuse Domoticz functionality without breaking BleBox integration, that will be obvious by running the tests - so as maintainers, you can change anything you want with the confidence that if tests work, actual devices will work too.

main/RFXtrx.h
gadgetmobile: - reverted in 1b3801519 (will be squashed later so that there will be no changes to this file)
- reverted new names in RFXNames.cpp in d31896716 (will be squashed later)
- added HW type checking for naming to solve naming issues in a generic way in 5a18ae465
- alias blebox enums to existing subtypes in 8e383e5e8 and c2cf6cfa6

Also, if RFXtrx.h is not to be modified, it should be in a separate repository with a git submodule here. That's the proper way of assuring it isn't modified accidentally.

Please review 5a18ae465 carefully, since there are no unit tests I could run to verify the changes are correct. (I did manually check the modified SQL queries as much as I could).
I did make sure the changes are minimal, careful and easy to review (which is a poor substitute for unit tests and good coverage). I believe it's an excellent solution to avoiding misleading names without lots of vendor-specific product names added to the database.

main/SQLHelper.cpp
gadgetmobile: - I moved the migration to an existing file (BleBox.cpp) in 2216906be

Of course, putting the whole migration inline wouldn't make sense, because SQLHelper.cpp would have to require a lot more headers.

Ideally, migrations should be separate from "SQL helpers" anyway, along with the schema. And since migrations can get messy in production code, one-file-per-migration is a good approach, especially since older migrations should be periodically deprecated/removed anyway. (And simply removing files from a repository is the best way to remove functionality).

I'm getting this sens of "allergy" towards having "lots of files", which makes no sense in 2020, especially with git, modern filesystems and the need to reduce maintenance burdens. (Instead of having gigantic files with enormous switch/if statements).

So I really don't understand what you mean here and what benefit outweighs the costs.

hardware/blebox/CMakeLists.txt
gadgetmobile: Works on gcc-6 on travis, which - along with cmake - is almost guaranteed to work on any other system. (You can assume that all modern C++ compilers have decent support of c++-17).

Also, the "target_compile_features" is a lot less useful than it sounds. Basically, it's best to support c++-17, unless there's a really, really strong reason to support only c++-14.

That's because there are almost no migration issues/work. (Mostly just the need to upgrade project dependencies from very old pre-c++-11 versions).

There are pretty much only benefits and almost no incompatibilities going from c++-11 to c++17 for this codebase. Besides, the unit/integration tests cover the important parts of MainWorker and SQLHelper, so there's little danger beyond the danger already lurking in any changes in the core codebase. There's also a chance that changes in semantics and strictness could expose existing bugs in Domoticz that no one has found yet.

gadgetmobile:
> Still have some comments here...

Sure.

> Also try to avoid using boost as much as possible and use the native functions

I don't know what you mean. Could you give examples?

I've been burned before by using e.g. `StringSplit`, which incorrectly handles "foo:" (where ":" is the separater) - which is why I switch to `boost::split` (instead of adding messy workarounds in the invocation). Same with `boost::format` which is currently the best way of formatting strings in c++ in a type-safe way, that works the same on every compiler. Boost functions also have 2 things missing here: documentation and unit tests.

`boost::asio::ip` was also very useful in reducing the code and making it more robust/portable.

I'm also not adding any extra boost dependencies beyond the existing minimum of boost::thread and boost::system. (Even the tests work on travis with this minimal boost build).

If it's something else, tell me which methods I'm not using and which ones I should be using. (Because I tried to reuse all the ones I could find that worked without modifications).

> As the 'setup' probably will not change that much, I still think it could be a good idea to make this a sub-module ?

I don't know what you mean by 'setup' or what would be a 'sub-module' here.

This integration is heavily Domoticz-specific and Domoticz is a monolithic app - which makes the code useless unless it's included in a given Domoticz version.

I did separate "blebox-specific" code from "domoticz-specific" code as much as I could - and the "blebox-specific" code is really just 2-4 files, which doesn't make sense to add to a separate repository, because it would just create lots of extra work with no purpose.

tests/CMakeLists.txt
gadgetmobile: This is probably the most ridiculous request I've ever heard. (Given the context).

I can't think of any single benefit and the downside is just huge.

The only reason I can think of to remove the tests is something stupid and unprofessional like "we don't want too many files in our project", which makes no sense given a git repository and a modern filesystem.

I can only think of 2 categories of answers:
1. stupid / unprofessional / incompetent
2. malicious / hidden agenda

The tests don't affect anyone negatively or break anything or interfere with anything. They only bring in benefits. (`stubs.cpp` is the exception, but the size of the file is the fault of Domoticz being 100% monolithic).

So I assume there's some serious misunderstanding here.

All other objections have been handled:
- they don't interfere with anything (separate folder / 100% independent build step)
- they test mostly the MainWorker and SQLHelper (so they're 100% dependent on Domoticz)
- they assure a level of quality (so you'd have to hate users and developers to have a reason to remove them)
- they're well written, well organized and generic (so they're useful for other JSON/HTTP based integrations)
- it's an industry standard to have at least basic unit tests included in quality open source projects
- the maintenance cost is ZERO - if they stop working, you can ignore them until someone sends a PR to fix them
- they build pretty fast (given that precompiled headers and poorly organized in Domoticz itself, which is why the main build is so slow to begin with)
- they run under 300 ms on Travis

I could go on...

Here is an example of them running on Travis (after a main build):

https://travis-ci.org/domoticz/domoticz/builds/647552213?utm_source=github_status&utm_medium=notification

Rejecting tests like these in this situation seems like something someone posted on: https://thedailywtf.com/

Especially since you have the option to include the tests and then just ignore them. (So whoever wants to use them can).

And you can always ping me to help out anyone who'd want to use (or fix) these tests.

gadgetmobile:
## Rebase + summary

I've rebased to upstream/development and reorganized the commits for the final merge:

Domoticz related commits:
- 85adf294c allow using external cereal headers
- cf1721aca add missing includes
- f5fa8afc8 use HW type to determine correct non-RFX device names - **handles generic naming properly**
- f23ee642a travis: update cmake to 3.16.3
- db754d5d5 move Update button next to Add (HW setup) - **makes UI more intuitive**

BleBox integration related:
- 246fc583e add googletest submodule for tests - **git submodule**
- 3edd6c495 blebox: add new reworked implementation with tests - **all related commits squashed**

Travis:
- 4e64bafac travis: run tests after build (safely, without breaking build) - [**test results are here**](https://travis-ci.org/domoticz/domoticz/builds/647589571?utm_source=github_status&utm_medium=notification)

tests/CMakeLists.txt
dannybloe: Have you looked at the dzVents tests? There are almost 1000 tests including many integration tests. Imo it's always good to have tests and I strongly believe Domoticz should have more of them. Without these dzVents tests it wouldn't be possible for me to maintain the project (dzVents). It happened more than once that someone else changed something in Domoticz core causing essential parts to stop functioning. The tests revealed this.

gadgetmobile: Yes, the dzVents tests were the first thing I tried to use.

While these seem useful for maintainers, they have 2 major flaws:
1. The unit tests obviously are pretty much useless for integration developers
2. The integration tests require a running Domoticz instance

The second is particularly problematic for developing c++ integrations:
- extremely long compile times (due to broken precompiled header support, which this PR allows fixing) to rerun tests
- difficult to debug segfaults and trace Domoticz internals - inconvenient to wait for build, server startup, and doing all those steps manually every time
- difficult to setup edge cases in database
- no way to drive/control the main worker (needed for edge cases in integration tests)
- not suitable for TDD (small changes in code + immediate feedback), which is crucial for development and refactoring in Domoticz (otherwise development is extremely discouraging)
- difficult to reliably run server instances for tests on Travis (which is probably why the existing dzVents are failing - 4 errors broken in upstream/development)
- testing some edge cases may require additional test-specific functionality exposed via Lua bindings (which would mean more changes to the core, which are beyond the scope of this PR)
- the tests are already really slow (even when run against a compiled domoticz) and that means the more tests are added (for edge cases), the slower the tests become (until no one will want to run them)

Meanwhile, these tests:
- are easy to add to Travis
- they bulid really fast (and can build even faster if tweaked), even on Travis
- they run under 300 milliseconds, even on Travis
- they don't require any new dependencies (except Google test/mock as a submodule, which is the recommended way and has no downside)
- they assure the tested integrations work, even without requiring actual devices for tests
- they don't require any changes in the core code to work (no extra bridges or layers)
- Google mock allows conveniently stubbing/mocking out any functionality (e.g. HTTP requests - which isn't possible with Lua integration tests)
- the tests allowed me to quickly migrate away from pTypeLighting2 to pGeneralDevice without debugging (which was necessary to work around a pTypeLighting2 bug)
- the tests were a great way to understand how Domoticz MainWorker handles each device type, without having to debug through the code full of gigantic switch statements and vendor-specific workarounds and hacks
- the tests support 11 new, very different devices through integration tests (meaning, using the database and MainWorker just as it does normally), and these would be very difficult to port to the Lua tests (which would make the tests likely umaintainable and would require lots of code changes in Domoticz). This alone is a huge achievement that would help a lot of device integrators while allowing the project owners to change anything they want in Domoticz internals and release confidently knowing they didn't break anything. Also, those 11 devices have almost identical tests (except for actual values and naming), which shows how generic and reusable they are. (And how difficult it is to develop reliable integrations for Domoticz - which is why these tests are crucial).
- these tests are heavily tied to Domoticz (calling the worker through the same API the UI ends up calling, and comparing Domoticz-specific database states with expectations), so they they can't even be run without a Domoticz source tree - and it makes zero sense to clone Domoticz, patch it, add tests and run this in a separate github/travis project ... just to "not have it in Domoticz". And those tests are really just testing to make sure Domoticz isn't broken - while testing on Dummy devices doesn't expose all the real-life edge cases that happen with real devices.
- code coverage is available to show device integration coverage (which is really, really high right now) and it highlights the paths takes in MainWorker/SQLHelper code. Adding code coverage to the whole codebase while running lua integration tests would be extremely slow and add a lot of noise (lots of paths triggered during test setup and app startup). Also, the reports would be huge.
- the tests are compatible with unity builds (I used Cotire until switching to recent cmake that superseded it), so going from a clean checkout to having test results and coverage reports is really fast and automatic. 
- many popular and respected projects EXPECT unit tests to be added along with changes - precisely to reduce testing and maintenance burden
- Lua tests don't have compile-time type checking, which is extremely useful for a C++ project like Domoticz, where defensive coding and type safety can prevent bugs from being introduce in code to begin with
- Lua tests aren't testing any specific integration (except for dummy devices), so they're not good, proven examples to use as a base for testing new/existing integrations, while the coverage in these tests show that ANYONE can make changes in Domoticz or the integrations covered by these tests and have confidence they didn't break anything BEFORE even submitting a PR. Again, the automation, convenience and speed of such tests is crucial to get people using them. (And adding tests themselves).
- the tests are very well designed, with the separation of hardware and "discoverable" features, and it makes the best usage of Domoticz architecture, so any new integrations that fit well with these exact tests with have a decent level of quality to start of with. This also means new devices can be added - along with their tests - mostly by copy&paste (or making things generic and introducing them into Domoticz). Otherwise, other developers new to Domoticz will copy whatever they pick, which means lots of bad practices, using outdated APIs and approaches, lots of confusion and lots of rework (and often after they've already invested lots of time manually testing their work for a PR). A single, maintained unit test in C++ like these would've saved me a lot of time and confusion.
- it's a separate folder requiring no changes to Domoticz itself (which is a technical miracle, considering how monolithic Domoticz is - you cannot otherwise build a binary for testing without including EVERYTHING except `main()` from `main/domoticz.cpp`). 

Also, Domoticz has an incredibly steep learning curve - which means lots of human lives wasted (instead of used for more valuable work, like adding useful features) just to avoid "having too many files in the project". Unit tests at least show HOW to do such things in a RECOMMENDED way. Throwing this away is like deliberately making development hard and inconvenient to frustrate and discourage people on purpose.

I've addressed all the concerns I could think of (most addressed above).

If you really don't want "too many files", I can just glue the test files into one giant file - which would be just as stupid to do. (And more inconvenient for everyone with no upside).

Since I've explained and covered so much about these tests (and why the lua tests aren't sufficient at all), I assume there's no reasonable reason to reject THESE tests in THIS situation.

gadgetmobile:
A huge amount of work and care went into this PR.

So please let me know if you need anything else for this to get merged.

hardware/blebox/workarounds/colorswitch.h
gizmocuz: If you have problems with the DeviceID/Unit fields in the database (as maybe it does not fit in a 32bit value), you can also use a temporary workaround for this and use the table 'WOLNodes' for now.
This table will be renamed in a feature release to reflect a better name.
Some other hardware types are also using this table to convert a large ID to a Usable Domoticz DeviceID/Unit, with this you should be able to create enough devices
I do not like so much code for 1 device type, hope you can understand
And as mentioned, please drop the whole test folder

gizmocuz:
@gadgetmobile , I added some comments

gadgetmobile:
@gizmocuz - I don't see any new comments.

I think your comments may be still "pending", because you probably haven't submitted your review.

gizmocuz:
Don't think so. You might find my comments 'ridiculous' but I don't
Please double check all my comments and requests to change

gadgetmobile:
The status is "Awaiting requested review from gizmocuz".

Other than that I addressed all the previous issues (which are closed), so I have no idea what you're talking about.

I found your request about removing tests to be ridiculous (given the context) and for the sake of ethics and professionalism it requires at least an explanation.

I thoroughly explained why I thought removing the tests was ridiculous and addressed every concern I could think of.

Almost all reputable open source projects **require** tests for every change/feature. (Or at least, tests are always extremely welcome).

Ask any professional senior software developer - your requests to remove tests will at best seem weird.

Other than that, I found a decent workaround for handling device names, etc.

You can also remove the tests AFTER merging this PR (which at least leaves the tests available for cherry-picking for others), so I completely don't understand the benefit.

(Except maybe something unethical/evil, like deliberately trying to sabotage the project or make life harder for everyone else).

gizmocuz:
Dear GadgetMobile,

Please do note that I do not have to accept any PR, but I am very glad that people contribute to the project.
The reason why I do not like the tests have been explained as well. It takes to much space.
That is why I asked if you think it would be an idea to make this hardware a sub project of Domoticz
I do not want to let everybody check out any more code/files then needed

What's with workarounds/gcc_6.h / byte ? (uint8_t ? should be there in GCC6)

I see you also removed the first line in BleBox.cpp
#include "stdafx.h"

As you noticed this introduced a lot of changes to header files, that is why we have this file
I would like to see those undo as well

gadgetmobile:
> The reason why I do not like the tests have been explained as well. It takes to much space.

You can move the `Manual` directory to a new repository and you'll gain almost 20MB. You haven't explained why space is an issue. (There are much better workarounds than removing tests).

As I've explained, the `stubs.cpp` is big, but that's because Domoticz isn't modular. (And this file allows integration tests). This allows gradually adding tests (e.g. for other integrations) and gradually increasing test coverage.

If you really cared about space, you'd filter out gigantic files (with e.g. `git-filter-branch`, `bfg`) and create a new repo/branch, which could be 200 MB smaller without any effort. (You don't have many PRs left to worry about rebase issues).

> That is why I asked if you think it would be an idea to make this hardware a sub project of Domoticz

It technically already is - on the CMake level.

Making it a submodule doesn't make sense - Domoticz integrations are heavily tied to Domoticz and make no sense outside Domoticz. And based on how hard it is to reason with you and understand your requirements, it makes no sense for me to make such big changes outside the scope of this PR. You can merge this request and then organize anything the way you want, while still giving people access to the tests in this PR. A git submodule is extra work for no good reason. (I've explained better ways of recovering "size" above).

> I do not want to let everybody check out any more code/files then needed

Well, currently, ALL Domoticz integrations are complied in and they're hardcoded. If there's no prior work in moving modules to separate repositories (or whatever), it doesn't make sense to ask a contributor to make such drastic repository changes. I also can't create repositories within the github `Domoticz` group, because I'm not a maintainer.

So it's beyond the scope of this PR, it's not a technical requirement to get this PR merged, and there are better ways to reduce the repo size by a few hundred megabytes.

(Also, tests are useful to compile and run on various platforms).

> What's with workarounds/gcc_6.h / byte ? (uint8_t ? should be there in GCC6)

It's a workaround for partial cxx-17 support in gcc6. `std::byte` is more restrictive than `uint8_t`, which helps detect potential bugs during compilation time. This was very useful during development (and I wish I switched to cxx-17 sooner during the development).

Ideally, you should have separate types for various values instead of "int", especially with values in ranges 0..15 or 0..100, etc. (Otherwise debugging values out of ranges in MainWorker is a nightmare).

If you want to remove this, you might as well remove the `typedef unsigned char BYTE` in `stdafx.h`.

> I see you also removed the first line in BleBox.cpp
> #include "stdafx.h"
>
> As you noticed this introduced a lot of changes to header files, that is why we have this file
> I would like to see those undo as well

The current header usage is botched. (Makes precompiled headers useless).

In general, the idea of including `stdafx.h` is obsolete, because you can force-include the file anyway (in MSVC and GCC/Clang, etc.).

> Please do note that I do not have to accept any PR, but I am very glad that people contribute to the project.

The goal of this was to create a high-quality integration that other integrations could benefit from.

The tests prevent the breakage that already exists in the current BleBox integration - so it prevents people in the future from making the same mistakes. Also, the current BleBox integration can't be easily fixed - and especially not without these unit tests.

You don't have to accept the PR and I don't have to contribute, sure.

I also can refuse to do things I consider unethical or unprofessional.

I've invested a lot of time to answer your objections.

I refuse the remove the tests (I can improve them if they're lacking in some way) because it deprives users and contributors of tools to assure a decent level of quality. (The lack of which already caused the current, broken BleBox integration to have been accepted - accepted by you, actually).

Summary: you accepted a broken PR (current BleBox integration) and you're rejecting the very tools that could've prevented this.

Again, I've presented better alternatives as a workaroudn for the "size" issue you mentioned.

gizmocuz:
Sorry you feel this way! Please feel free to make a new PR with the wanted changes
Maybe you should split this PR in multiple parts
Your absolutely right about the manual folder, nobody seems to want to update this, and the wiki has more information.
As soon as the content from this old manual is migrated to the wiki this can be removed

PR #3877, opened 2020-01-30
Add managedmultimeter device

guillaumezin:
This pull request creates a new Managed Multi Counter device similar to P1 smart meter, that exposes a way to directly update data history. It is useful for scripts and plugins that grabs metering data from the cloud, to insert it in Domoticz DB. It was already the case with Managed Counter, Managed Multi Counter adds the possibility to manage 2 costs and energy consumption and delivry


www/app/hardware/Hardware.html
```diff
@@ -1348,6 +1348,7 @@ <h2><span data-i18n="Hardware">Hardware</span>:</h2>
 						<option value="0xF31B" data-i18n="Distance">Distance</option>
 						<option value="0xF31D" data-i18n="Electric (Instant+Counter)">Electric (Instant+Counter)</option>
 						<option value="0xF321" data-i18n="Managed counter">Managed counter</option>
+						<option value="0xF322" data-i18n="Managed multi counter">Managed multi counter</option>
```
gizmocuz: Please try to use Camel Case here ... (Also for the previous managed counter)
Managed Counter
Managed Multi Counter

This makes it consistent throughout the application

gizmocuz:
Thank you for the PR!!
This looks indeed almost identical to the P1 Meter... Can't your script update a Dummy P1 Meter ?

guillaumezin:
Nope, you will see in sqlhelper.cpp that these particular meters (Managed and Managed multi) can directly insert data in Meter and MultiMeter_Calendar if you pass the date in sValue. And P1 Smart Meter updates Meter and MultiMeter_Calendar every day, I don't want this with these meters. There are a lot of particular cases where I think it is cleaner to deal with a separate device.

I will have a look to the code to use "Camel Case" everywhere.

www/app/hardware/Hardware.html
guillaumezin: Cf. 1991c520ce9ccc5c88184ee8fefd23e379fd5250

gizmocuz:
This is the way Domoticz works internally. Please use the P1 meter alternative

Issue #3872, opened 2020-01-28
RTL433 PCR800 mm and inch confusion ?


dgrostoto:
This night i had my first rain reported by my PCR800 (i hadn't rain since days...), and  i found values reported in domoticz rather low, i investigated quickly and i think there confusion between inches and mm (the old battle royal Britanic system vs international metric system which already made a plane crash...)
rtl_433 software seems to report rain in inches and inches per hour.
But domoticz believe it is mm and mm per hour.
see below my data in the same hour
rtl433 - classic format (rounded 0.1%f)
time      : 2020-01-28 02:47:02                    brand     : OS
model     : PCR800       House Code: 134
Channel   : 0            Battery   : OK            Rain Rate : 0.3 in/hr     Total Rain: 1.5 in

rtl433 - csv format (without round by default)
[...]

Issue #3834, opened 2020-01-04
RTl433 Smoke detector GS


mojso:
please insert it into devices
Smoke detector GS 558 - vibration sensore
does not recognize it as a device

Json:
{"time" : "2020-01-04 14:19:30", "model" : "Smoke detector GS 558", "id" : 16624, "unit" : 5, "learn" : 0, "code" : "c81e05"}

[...]

Issue #3272, opened 2019-05-26
RTL433: rid is ignored, cannot distinguish sensors


taismi:
RTL433 uses id, rid and channel to distinguish sensors but Domoticz's Rtl433.cpp use only id and channel. 

`unsigned int sensoridx = (id & 0xff) | ((channel & 0xff) << 8);`
https://github.com/domoticz/domoticz/blob/development/hardware/Rtl433.cpp#L341
https://github.com/domoticz/domoticz/blob/4.10717/hardware/Rtl433.cpp#L331

This causes sensors with same id an channel to be seen as one device in Domoticz. RFLink sees sensors as separate devices.

gizmocuz:
True, I would have xorred it maybe or shifted it
Not something that can be changed without all current users getting new id's
But how big are the odds that this is happening and can this be avoided by the user as well ?

taismi:
I have sensors with IDs like 0109, 0209 and 0309. This is easy to use if you have only three sensors. Sensors support three channels. This doesn't work anymore with higher sensors count. Maybe an option to combine id and rid would solve things so by default users would not get new IDs. I have rids 28, 116, 30, 46, 113, 116, 124, 232. 

In the transmission first octet is 9. After that there are two octets which are seen in RFLink's ID but no it RTL433. After that there's an octet which's last two bits are channel. First two bits are unknown for me.

-------
I found from the RTL433 source code that  first octet could also be 5 .
> Prologue sensor protocol
> 4 bit fixed 1001 (9) or 0110 (5)

I don't know why RTL433 set fixed value to ID and rest to RID. Some sensors don't show RID, only ID which contains all.

gizmocuz:
Seems like you can control those id's ? So you can also assign them a different id ... 1,2,3 etc

taismi:
Only three different id's are possible and I have 8 sensors. All sensors have unique rid's.

bousqi:
> RTL433 uses id, rid and channel to distinguish sensors but Domoticz's Rtl433.cpp use only id and channel.
> 
> `unsigned int sensoridx = (id & 0xff) | ((channel & 0xff) << 8);`
> https://github.com/domoticz/domoticz/blob/development/hardware/Rtl433.cpp#L341
> https://github.com/domoticz/domoticz/blob/4.10717/hardware/Rtl433.cpp#L331
> 
> This causes sensors with same id an channel to be seen as one device in Domoticz. RFLink sees sensors as separate devices.

[...]

Issue #3144, opened 2019-04-09
Rtl433: Unhandled sensor reading  akhan / X10


alex1313c:
Hello
Sorry for my english translated
I have two problems  i use Realtek RTL2832 based DVB dongle to receive 433Mhz I use the latest version RTL, and domoticz version: 4.10602 
the first is a remote control key 3 buttons ("beeper X6 R-TX2") which is very well
 recognized by the command rtl__433 -G ( or protocol 48)

```
time      : 2019-04-09 20:27:00
model     : Akhan 100F14 remote keyless entry      ID (20bit): 0x6e890
Data (4bit): 0x4 (Mute)
_ _ _ _ _ _ _ _ _ _ _ _

time      : 2019-04-09 20:27:07
model     : Akhan 100F14 remote keyless entry      ID (20bit): 0x6e890
Data (4bit): 0x1 (Lock)
_ _ _ _ _ _ _ _ _ _ _ _

time      : 2019-04-09 20:27:08
model     : Akhan 100F14 remote keyless entry      ID (20bit): 0x6e890
Data (4bit): 0x2 (Unlock)
_ _ _ _ _ _ _ _ _ _ _ _

```
and on domoticz:

```
2019-04-09 20:28:57.341 Status: Rtl433: Unhandled sensor reading, please report: (time,msg,codes,model,id,channel,state,unit,group,battery,
temperature_C,humidity,brand,sid,data,sensor_id,battery_low,battery_ok,sequence_num,message_type,wind_speed_mph,wind_speed_kph,wind_avg_mi_h,
wind_avg_km_h,wind_dir_deg,rain_inch,rain_in,temperature_F
2019-04-09 20:28:57.341 |2019-04-09 20:28:56,,,Akhan 100F14 remote keyless entry,452752,,,,,,,,,,0x4 (Mute),,,,,,,,,,,,,
2019-04-09 20:28:57.341 ) 

-------------------

2019-04-09 20:29:05.254 Status: Rtl433: Unhandled sensor reading, please report: (time,msg,codes,model,id,channel,state,unit,group,battery,
temperature_C,humidity,brand,sid,data,sensor_id,battery_low,battery_ok,sequence_num,message_type,wind_speed_mph,wind_speed_kph,wind_avg_mi_h,
wind_avg_km_h,wind_dir_deg,rain_inch,rain_in,temperature_F
[...]

PR #3902, opened 2020-02-10
compliance to rtl_433 newmodel (normalized keys)

dgrostoto:
works like a charm at home. Now should check that units are the same on domoticz side.
Units to be checked :
wind_avg in km/h
wind_dir in degre
wind_max in km/h
depth in cm
other units should be OK.

### DIFF
--- a/hardware/Rtl433.cpp
+++ b/hardware/Rtl433.cpp
@@ -30,10 +30,16 @@ CRtl433::CRtl433(const int ID, const std::string &cmdline) :
 	m_time_last_received = 0;
 /*
 	#ifdef _DEBUG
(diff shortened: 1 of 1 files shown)


MrHobbes74:
Your changes are not backwards compatible. Old devices that do not have the new API installed, will not work after this change. Am i correct? In that case, you can't just change the code to work with the new API, while forgetting about the old API.

dgrostoto:
Not at all! every device support the new "API" (this is not really a new API but a new set of normalized keys but let's call it an API). It has been worked on rtl_433 side for more than a year now. And i pointed the remaining bugs i've found on rtl_433 and they 've been corrected. So all of the 201 devices supported by rtl_433 are functional with domoticz with the new API. I've checked every key of every device used in domoticz (that's quite a big work for me). I can also say that (moreover i did not left on side any device) there are other devices that did not correctly or at all that are now working correctly thanks to those normalized keys.
Bonus, the code is now simpler on domoticz side (no more trick depending of device type).
Let me know is the coding style is OK for you or if I could left some bug (i'm human after all :) )

gizmocuz:
@dgrostoto , thanks for the explanation.
@MrHobbes74 , are you able to test this ? Else I merge the PR

dgrostoto:
The point i'd like some additional checks concerns the units quoted above, to be sure they are interpreted in same unit that rtl_433 send or if they are additional conversion to be done.

gizmocuz:
@dgrostoto , Domoticz works internally in m/s , see mainworker.cpp, decode_wind,  with 1 decimal.
3.6 mtr/sec should be stored as 36, same for temperature 5.1 should be 51

MrHobbes74:
> @dgrostoto , thanks for the explanation.
> @MrHobbes74 , are you able to test this ? Else I merge the PR

@gizmocuz I cannot test this, I don’t have this hardware, I’m also not familiar with it, but it was just an observation. But it may mean all devices that users have and didn’t upgrade for a year, will need an upgrade? All devices may support the new API but not all may have this API installed. That was more my concern.

dgrostoto:
@gizmocuz , OK i'll integrate the appropriate conversions. The metric system and unit used by the newmodel have a great advantage : they have chosen the smallest units, so precision is better and even more some events may be detected whereas they may be not (due to rounds) with bigger sized units (i had the concern with my PCR800 raining sensor device).
@MrHobbes74 , let me explain futhermore : the "API" (as i said, just a set of normalized keys called a model, not an API strictly speaking) is not on the device side (the hardware) but on rtl_433 software side. So every hardware device is compliant with this new model since rtl_433 is handling the mapping of the raw data to the corresponding key in the new model (temperature, wind direction, id, etc...). So there is no need to have any driver nor something specific on device side to be compliant for this model, which is great isn't it ? :)
For the users who have not upgrade for a year, the change will be transparent, provided of course they upgrade domoticz AND rtl_433. To have the things done right, i will tell which minimum version of rtl_433 must be used and we could add this information in the documentation.

MrHobbes74:
Totally clear to me! Thanks for explaining!

dgrostoto:
Conversions are done. it is OK for merging from my side.

dgrostoto:
rtl_433 project  tagged its latest version. So i can tell now that version with flag "20.02" (2020-02-17 branch master) is the version minimal need for this PR (#3902 ) and also the recommended version.

gizmocuz:
Thank you for this !!

dgrostoto:
found some issue that should be solved by this PR
https://github.com/domoticz/domoticz/issues/3872
https://github.com/domoticz/domoticz/issues/3834
https://github.com/domoticz/domoticz/issues/3272
https://github.com/domoticz/domoticz/issues/3104
https://github.com/domoticz/domoticz/issues/2966
There is some device support demand that cannot be satisfied as is 
https://github.com/domoticz/domoticz/issues/3144

PR #3904, opened 2020-02-11
minor fixes to EvohomeWeb logging

gordonb3:
* parsing of html content incorrectly showed opening angle brackets
   for entities that are not part of the resulting log line.

 * notification that the frequency of login attemps is being trottled
   down should only be displayed when the interval is actually being
   increased.


gizmocuz:
Thank you !

Issue #3896, opened 2020-02-07
Calculation of Counter T1 of P1 Smart meter is not correct


FireWizard52:
It looks that the calculation of the Counter T1 (Usage) is not correct.
All other counters (T2, R1 and R2) calculate exactly the correct values.
The value of the day should be the sum of the value of the previous day and the usage of the day.

Example:
January 1: Counter T1 (7551,166)  + Usage T1 20,448 should result in:
January 2: Counter T1 (7571,614) instead of 7571,796.

For more information: See https://www.domoticz.com/forum/viewtopic.php?f=6&t=31192
Posts of erem and FireWizard, February 6, 2020

Running: Domoticz 4.11670
Rasbian Buster on RPi2B

gizmocuz:
That is odd as there is no 'calculation' being made , we use the direct counter values

It is stored in the database every (5) minutes here:

void CSQLHelper::UpdateMultiMeter()

and at the end of the day:

void CSQLHelper::AddCalendarUpdateMultiMeter()

you should be able to verify this in the database tables

FireWizard52:
Thanks for the reply.
I will check tomorow and report.

gordonb3:
This is a long known shortcoming, I wouldn't call it a defect perse. A day in Domoticz is not 24 hours because every 24 hour period you can think of contains at least one time value that has a different date, causing it to fall outside the query range used in these calculations. Because the values are stored in 5 minute intervals a day therefore becomes a period of 23 hours and 55 minutes. Typically this will result in a deviation of around 10Wh but of course it might also be close to a full kWh if you're charging an EV.

FireWizard52:
After the forum reaction of gordonb3 and the fact that it is obviously a long known issue, it doesn't make sense for me to look into the database, so I wont do that.
As it is clear to me that it will only happen with Usage T1, it is the question, do we want to solve it?

gordonb3:
Need to note that I have not looked at this particular implementation for quite some time. As such I'm not completely sure whether only data from the shortlog is used to calculate these daily values or if it uses the device state value as the last known value. Either way there appears to be a time difference between the time that this calculation is based on and the creation of the first shortlog entry for the next day. Whether that's 5 minutes, 10 seconds (DSMR standard meter) or 1 second (ESMR 5.x meter) doesn't really matter as the difference will always be notable.

FireWizard52:
Please see the recent posts:
[...]

PR #3905, opened 2020-02-11
correct query range for meter reports

gordonb3:
As discussed in issue #3896 the daily totals in meter reports do not include the last 5 minutes of the day due to the fact that the following midnight belongs to the next day. This patch extends the range for the various updates to `(multi)meter_calendar` to include the midnight value.

Additionally I reviewed the code for creating the line for `today` in the P1 report and added the missing `MIN(counter)` values to it.


gizmocuz:
Thanks !!

Issue #3909, opened 2020-02-13
Update to most recent beta gives: mainworker.cpp error: conversion from ‘time_t {aka long int}’ to ‘Json::Value’


HoekieN:
I am trying to upgrade from current version:
Version: 4.11672
Build Hash: ba48297aa-modified
Compile Date: 2020-02-06 00:46:09
dzVents Version: 2.5.7
Python Version: 3.6.9 (default, Nov 7 2019, 10:44:02) [GCC 8.3.0]
To the latest beta and i am getting below error

ubuntu@ubuntu:~/dev-domoticz$ cat /etc/*-release
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=18.04
DISTRIB_CODENAME=bionic
DISTRIB_DESCRIPTION="Ubuntu 18.04.4 LTS"
NAME="Ubuntu"
VERSION="18.04.4 LTS (Bionic Beaver)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 18.04.4 LTS"
VERSION_ID="18.04"

[ 29%] Building CXX object CMakeFiles/domoticz.dir/main/mainworker.cpp.o
/home/ubuntu/dev-domoticz/main/mainworker.cpp: In member function ‘void MainWorker::HandleAutomaticBackups()’:
/home/ubuntu/dev-domoticz/main/mainworker.cpp:1433:44: error: conversion from ‘time_t {aka long int}’ to ‘Json::Value’ is ambiguous
    backupInfo["duration"] = mytime(NULL) - now;
                                            ^~~
In file included from /home/ubuntu/dev-domoticz/main/../main/../json/reader.h:11:0,
                 from /home/ubuntu/dev-domoticz/main/../main/../json/json.h:11,
                 from /home/ubuntu/dev-domoticz/main/../main/json_helper.h:3,
                 from /home/ubuntu/dev-domoticz/main/mainworker.cpp:17:
/home/ubuntu/dev-domoticz/main/../main/../json/value.h:326:3: note: candidate: Json::Value::Value(bool)
   Value(bool value);
   ^~~~~
/home/ubuntu/dev-domoticz/main/../main/../json/value.h:304:3: note: candidate: Json::Value::Value(double)
   Value(double value);
   ^~~~~
/home/ubuntu/dev-domoticz/main/../main/../json/value.h:302:3: note: candidate: Json::Value::Value(Json::Value::UInt64)
   Value(UInt64 value);
[...]

PR #3910, opened 2020-02-14
Changed backup duration calc to use difftime. Fix for compile issue on stretch (#3909) 

rwaaren:
duration changed to float (number in Lua)

	modified:   main/WebServer.cpp
	modified:   main/dzVents.cpp
	modified:   main/mainworker.cpp


gizmocuz:
Thanks!!

PR #3914, opened 2020-02-15
Change WebSockets support on Devices page

stas-demydiuk:
- Refactor websockets service
- Extracted browser notifications to own module
- Fix websocket subscriptions on existed pages

### DIFF
--- a/www/app/DashboardController.js
+++ b/www/app/DashboardController.js
@@ -3,8 +3,6 @@ define(['app', 'livesocket'], function (app) {
 		var $element = $('#main-view #dashcontent').last();
 
 		$scope.LastUpdateTime = parseInt(0);
-		$scope.broadcast_unsubscribe_devices = undefined;
-		$scope.broadcast_unsubscribe_scenes = undefined;
 
 		//Evohome...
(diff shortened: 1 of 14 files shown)


gizmocuz:
Thank you again for this great PR!!
Small note, please take care of the formatting... "function (bla)" vs "function(bla)" , it is still a mix,
I prefer the one without the extra space, as most editors will also do.
Maybe you use a editor that does put a space here and this could be changed ?

https://jsonformatter.org/jsbeautifier

rwaaren:
@stas-demydiuk Stanislav Demydiuk

Hi can you please have a look at the discussion on the [forum](https://www.domoticz.com/forum/viewtopic.php?f=6&t=31379) And the suggestion made there to add a selection on the left pane to activate / deactivate the real time updates ?
Maybe also try to prevent the page jumping back to 1 at every update.

Thx !

rwaaren:
@stas-demydiuk Stanislav Demydiuk

Hi can you please have a look at the issue of replacing the group / scenes by the device with the same id when the device is updated. [Forum post P1 Meter - duplicate devices](https://www.domoticz.com/forum/viewtopic.php?f=14&t=31659
)

Issue #3913, opened 2020-02-15
High memory use (100%) on 4.11690


remcom:
On version Version: 4.11690 Build Hash: f1c50073d domoticz is using 100% memory (~7,5 gb) within a hour of the start. After a few hours the high memory usage the system freeze up.

This is on a Debian 10 system.

This issue wasn't present in commit 91c3475ac3c629ed28171e8fa5c8fcc5b40d974b (9 Feb).

cronyx:
Same problem...

Version: 4.11690
Build Hash: f1c50073d
Compile Date: 2020-02-14 14:59:38
dzVents Version: 3.0.0
Python Version: 3.6.9 (default, Nov 7 2019, 10:44:02) [GCC 8.3.0]

```
Feb 15 05:15:26 domoticz kernel: [36749.999537] Out of memory: Kill process 17526 (domoticz) score 912 or sacrifice child
Feb 15 05:15:26 domoticz kernel: [36749.999646] Killed process 17526 (domoticz) total-vm:6235420kB, anon-rss:3810188kB, file-rss:0kB, shmem-rss:0kB
Feb 15 05:15:26 domoticz kernel: [36750.179617] oom_reaper: reaped process 17526 (domoticz), now anon-rss:0kB, file-rss:0kB, shmem-rss:0kB
```
![Выделение_125](https://user-images.githubusercontent.com/2557102/74590380-e33eb500-501e-11ea-8bd6-7d6a0d79efde.png)

HoekieN:
![image](https://user-images.githubusercontent.com/28273279/74589400-5a674f80-5005-11ea-8a4a-7b3689e44f44.png)
Same here.

Version: 4.11690
Build Hash: f1c50073d-modified
Compile Date: 2020-02-14 12:59:38
dzVents Version: 3.0.0
Python Version: 3.6.9 (default, Nov 7 2019, 10:44:02) [GCC 8.3.0]

rwaaren:
Thx for reporting. It is being worked now so no further need for extra reports

GravityRZ:
just noticed the same
this time it is a massive memory leak.
installed V4.11690 at 15:13
at 20:30 it hit 100% memory usage

a second problem i noticed you can not roll back backups

i installed V4.11605 but i could not restore backup from v4.11690
i got a mesage that downgrade is not permitted
lucky i also made a 605 backup.

when did things change that we can not roll back?

MrHobbes74:
[...]

PR #3915, opened 2020-02-15
solve memory leak in usage of LuaTable

MrHobbes74:
Introduction of the new LuaTable class unfortunately introduced memory leaks (issue #3913 ). This PR solves the leak.


rwaaren:
Tested this for 7 hours and no memory leak after I applied the changes from this PR. 
Without the changes my memory usage increased by 4% / hour.

gizmocuz:
Thanks!!

Issue #3922, opened 2020-02-18
devices deselected automatically


felix63:
Running Version: 4.11696 on Rasberry.

When selecting devices in the device list they get deselected automatically after a couple of seconds.

rwaaren:
#3924 fixed it in such a way that now only devices that received a real time update while being selected are de-selected. Other selected devices will keep the select mark active. 
Can you check and when OK close this issue ?

felix63:
Version: 4.11707 does resolve the problem of deselected devices. However when deleting or changing names of devices these updates are not shown unless the screen is refreshed.

rwaaren:
> Version: 4.11707 does resolve the problem of deselected devices. However when deleting or changing names of devices these updates are not shown unless the screen is refreshed.

That should be fixed in later Beta's (> V4.11711)

gizmocuz:
Solved

Issue #3918, opened 2020-02-17
Device page is reacting slow after PR #3914


rwaaren:
Not sure if @stas-demydiuk gets notified about my remark on the closed #3914 
If it does I will close this issue 

Please see the discussion on the [forum](https://www.domoticz.com/forum/viewtopic.php?f=6&t=31379) about the behavior of the devices tab after #3914 
 
If an extra selection could be added to device tab (similar to the Used = (No / Yes) one ), to activate / deactivate it, I would see the real time updates on the devices tab as a welcome extra option. Especial together with the selections you can make on the left pane.
obviously the jumping back to page 1 should be prevented.

stas-demydiuk:
Hi @rwaaren, thanks for reporting, will take a look!

rwaaren:
I guess [issue #3922 ](https://github.com/domoticz/domoticz/issues/3922)  might also be related ?

PR #3924, opened 2020-02-18
#3918 #3922 Handle websocket messages by device table itself

stas-demydiuk:
Should address #3922 and probably #3918 issues as well as issue posted on the [forum](https://www.domoticz.com/forum/viewtopic.php?f=6&t=31379) with resetting current page

### DIFF
--- a/www/app/devices/Devices.js
+++ b/www/app/devices/Devices.js
@@ -232,6 +232,16 @@ define(['app', 'livesocket'], function (app) {
                     $scope.$apply();
                 });
 
+                $scope.$on('device_update', function (event, deviceData) {
+                    table.api().rows().every(function() {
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thanks!!

gizmocuz:
@stas-demydiuk , i seem to have a minor issue (maybe unrelated), when I add a new device to the system (green arrow) it is added, but the list is not refreshed and it 'seems' the device was not added
refreshing the browser solves it

stas-demydiuk:
That's because there is no websocket event from server side about device update after you add or remove device

rwaaren:
Same is true when you rename a device

gizmocuz:
Maybe do a complete refresh of the list ?

PR #3926, opened 2020-02-19
Get rid of annoying "Async Secure Read Exception: 1, stream truncated"

guillaumezin:
It happens with plugins that use https connection, when the third party closes the connection (meaning most of the time)

### DIFF
--- a/hardware/plugins/PluginTransports.cpp
+++ b/hardware/plugins/PluginTransports.cpp
@@ -481,6 +481,7 @@ namespace Plugins {
 			else
 			{
 				if ((e.value() != boost::asio::error::eof) &&
+					(e.value() != 1) &&	// Stream truncated, this exception occurs when third party closes the connection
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thanks !!

skarab22:
I'm novice on github using. How to include your changes in installed domoticz ?

kiddigital:
> I'm novice on github using. How to include your changes in installed domoticz ? 

This is a question for on the forum.

But in short: install the beta. Look at the wiki to check what is available in the beta that is not yet in the stable.

skarab22:
> > I'm novice on github using. How to include your changes in installed domoticz ?
> 
> This is a question for on the forum.
> 
> But in short: install the beta. Look at the wiki to check what is available in the beta that is not yet in the stable.

Ok thanks. Excuse me for the disturb

PR #3931, opened 2020-02-20
add missing includes

gadgetmobile:
Extracted from closed PR:
https://github.com/domoticz/domoticz/pull/3876

- allows the given files to be complied in different contexts
- required by new blebox integration (mitigates precompiled header issues)
- reduces dependency on obsolete Microsoft/stdafx precompiled header setup

### DIFF
--- a/main/Logger.h
+++ b/main/Logger.h
@@ -4,6 +4,9 @@
 #include <list>
 #include <string>
 #include <fstream>
+#include <mutex>
+#include <map>
+#include <sstream>
 
 enum _eLogLevel : uint32_t
 {
--- a/main/RFXNames.h
+++ b/main/RFXNames.h
@@ -1,6 +1,8 @@
 #pragma once
 
 #include <string>
+#include <map>
+#include <vector>
 
 #define sTypeTH_LC_TC 0xA0   //La Cross Temp_Hum combined
 #define sTypeTEMP_SYSTEM 0xA0  //Internal sensor
--- a/main/SQLHelper.h
+++ b/main/SQLHelper.h
@@ -1,6 +1,11 @@
 #pragma once
 
 #include <string>
+#include <iomanip>
+#include <map>
+#include <thread>
+#include <vector>
+
 #include "RFXNames.h"
 #include "../hardware/hardwaretypes.h"
 #include "Helper.h"


gizmocuz:
These includes are not needed as they are already inside the stdafx.h header which is/should be included at the top of each .cpp file before including anything else.
Source compiles also on all platforms
Seems to be you are not including this file in your code ?

gadgetmobile:
As I've mentioned before:
- stdafx is obsolete (even from Microsoft's perspective)
- including a header in every source file is a poor and inflexible way to set compile options (especially since this file can be included automatically with a compiler option)
- if you ever want to suggest moving files into separate submodules, how do you propose deal with the fact that the stdafx.h is in this project? It makes every cpp module highly dependent on Domoticz
- the precompiled header artifacts are huge and usage is broken in Domoticz - allowing subprojects (like blebox and tests) to use their own headers drastically speeds up compile time and shows they're not necessary
- stdafx usage gets in the way of CMake precompiled header functionality (and the cotire project, which is supercedes)
- files should compile regardless what's in stdafx.h - so headers and definitions should be duplicated in every cpp file anyway (since it's up to a given project/subproject what to put in stdafx.h)
- a BYTE typedef in a header is bad practice
- TDD requires fast develop/compile/test cycles, and the current Domoticz build is slowed down by bad precompiled header handling. (This requires an overhaul for the whole project to benefit, but I'm discouraged by the friction and unnecessary effort required to get such trivial and highly beneficial changes merged).

I could go on.

It's valuable to have per-subproject precompiled header support. (Otherwise you'd be either slowing the build either through huge precompiled headers, or lots of non-precompiled headers being included in every build).

I didn't create this PR because I'm bored or that I'm so stupid I didn't realize what stdafx.h does or I forgot to include it.

Domoticz is incredibly monolithic (by any standards) and this is a non-invasive way to help change that gradually.

Do you have very strong counter-arguments to reject this PR?

gizmocuz:
I don't agree Microsoft is saying not to use precompiled headers, where did you read that ?
For example, try installing for example VS2019, and create a dialog application...

Anyway, here is a document from Oct. 2019

https://docs.microsoft.com/en-us/cpp/build/creating-precompiled-header-files?view=vs-2019

If you want to contribute to this project you have to follow the projects current standards.

If you think that removing the precompiled header from this project will increase build time,
(and have of course existing prove for both windows/unix based platforms), sure I would not say no to this, but make a PR for this
Currently you have many PR's, so I suggest if you want to have some fixes for BleBox that you first make them in the current Domoticz/Project way, then continue making 'a' new PR
If the last BleBox PR had wrong code, i can also revert this if that is easier ?

gadgetmobile:
> I don't agree Microsoft is saying not to use precompiled headers, where did you read that ?

Nowadays, precompiled headers are **generated**.

For example, the [cotire project](https://github.com/sakra/cotire) does this - it's a CMake module where `stdafx.h` is [not needed](https://github.com/sakra/cotire/tree/master/src)

And if they're not generated, they can be [**force included**](https://docs.microsoft.com/en-us/cpp/build/reference/fi-name-forced-include-file?view=vs-2019).

Also, Domoticz is **supposed** to compile **without** including stdafx.h everywhere, because it's set up (incorrectly?) to use this strategy on MSVC: https://github.com/domoticz/domoticz/blob/development/CMakeLists.txt#L228

So if it doesn't work without `stdafx.h` **the Windows builds have a broken CMake config**.

So there's NO reason to include `stdafx.h` AT ALL. Especially in **portable** projects (which Microsoft historically never ever cared about), where each major compiler (GCC/Clang/MSVC) has a **different** approach to precompiled headers.

Without fixing the precompiled header situation (which requires lots of PRs like this), building is unnecessarily **slow** on non-Windows platforms (and also slower than necessary on Windows too).

> For example, try installing for example VS2019, and create a dialog application...
 
The `afx` in `stdafx` comes from 1994 and Microsoft never was a source of good portible software development practices. Microsoft always **preferred** to have their project non-portable. That's also why many windows-only software developers are so dangerously incompetent - they rely on Microsoft defaults without even using their brains.

I haven't developed for Windows since 2007, and I've never since seen `stdafx.h` anywhere except in old, unportable Windows projects.

> Anyway, here is a document from Oct. 2019

Since this project uses CMake and targets mostly Raspberry, there's little point in using Microsoft-specific features and tools at all. (Except maybe for Microsoft-specific API if it's unavoidable - but there are #ifdef's for those). CMake even generates VC++ solutions by itself automatically.

Different compilers have different precompiled header strategies: https://en.wikipedia.org/wiki/Precompiled_header

And that's why you shouldn't have project-specific defines or typedefs in precompiled headers (like you do).

> If you want to contribute to this project you have to follow the projects current standards.

The standards aren't documented and the CMake precompiled-header handling config suggests the `stdafx.h` is unnecessary.

Also, precompiled headers are for **increasing build time**. Their purpose is not to "include commonly used headers to avoid including them". You're also heavily polluting the namespace, because every compilation unit includes lots of headers it doesn't need. 

E.g. What does `tinyxml` need sockets for? Or `stdio.h`, which is an c header (that should be replace with `cstdio.h` at least. Isn't that ridiculous?

For portability purposes, an empty stdafx.h file should work just as well - because precompiled headers are a "project **build** setting" not a list of **include dependencies**. 

> If you think that removing the precompiled header from this project will increase build time,
> (and have of course existing prove for both windows/unix based platforms), sure I would not say no to this, but make a PR for this

This PR fixes the BUILD. You can't test differences in build time if the **build is broken**.

Look again at the changes and example to me **what am I breaking?**

Nothing.

I'm including missing **standard c++ headers** in the compilation units **which require them**.

I'm not even **removing stdafx here** so why the refusal to merge this?

You can keep the unnecessary stdafx includes where they are, and if you accept this PR it breaks nothing and prevents build breakage for other people down the road.

Clang/GCC use a **completely different** precompiled header strategy. Using `stdafx` gets in the way of fast builds on non-windows platforms.

Ideally, even in VC++, I should be able to setup **any** set of includes to precompiled for the sources I need - and changing them should require changes to source files. 

> Currently you have many PR's, so I suggest if you want to have some fixes for BleBox that you first make them in the current Domoticz/Project way, then continue making 'a' new PR

1. This PR doesn't break anything. It's neutral on Windows builds. (Unless you don't understand programing or something). It's a trivial merge that doesn't require any comments.

2. This is necessary for BleBox to `build`. If this is merged, the BleBox PR should be a no-op with regards to this. If you later revert this, you'll likely break the BleBox build.

3. If something breaks, the unit tests in BleBox will help discover this much sooner

4. The worst thing this PR can do is break the build (which I can't imagine a scenario where it can). It doesn't break the build. (See appveyor build). So what's the fuss all about here?

> If the last BleBox PR had wrong code, i can also revert this if that is easier ?

I don't understand.

If you're scared of breaking things, you need more tests.

You're creating unnecessary obstacles and wasting my time. You're also not paying me to explain things to you.

Maybe you should step down as a maintainer and ask more competent/fluent people that you trust to help you maintain this project?

Because otherwise I'm wasting my time explaining things a seasoned open-source developer would have no objections to - instead of actually contributing to the project.

It seems this project is not worth contributing to - too much time wasted on fake obstacles and groundless opinions instead of helping users enjoy better quality software. This is sad.

gizmocuz:
Sorry if you see it this way.

PR #3935, opened 2020-02-20
travis: update cmake to 3.16.3

gadgetmobile:
Extracted from closed PR: #3876

Updates CMake to 3.16, which enables useful feature like built-in precompiled header support, more and better default modules, and lots of bug fixes.

Distros often contain very old versions of CMake, so it usually has to be downloaded and installed anyway - and so, it makes no sense to support older versions of CMake.


gizmocuz:
Thanks!!

PR #3938, opened 2020-02-21
Introduction of tesla hardware in domoticz

MrHobbes74:
I created a domoticz native Tesla hardware module.
The module connects to a Tesla, and the first version provides:
- a battery level indicator
- a alert indicator giving you the status of the car 
(offline, not home, home, cable disconnected, charging, charging stopped)
- a temperature meter indicating the temperature in the tesla
- a switch to turn on / off charging
- a switch to turn on / off climate control
- a switch to turn on / off defrost mode

For safety, switches cannot be set when the car is not at home/around the house.

### DIFF
--- a/hardware/Tesla.cpp
+++ b/hardware/Tesla.cpp
@@ -289,13 +289,13 @@ bool CTesla::WakeUp()
 		Log(LOG_STATUS, "Waking up car.");
 
 	if (m_trycounter > 0 && m_car.is_home)
-		SendAlertSensor(TESLA_ALERT_STATUS, 255, WakingUp, "Waking up", m_Name + " State");	
+		SendAlertSensor(TESLA_ALERT_STATUS, 255, Offline, "At home, Waking up", m_Name + " State");	
 
 	if (m_api.SendCommand(CTeslaApi::Wake_Up))
 	{
 		m_car.awake = true;
 		if (m_trycounter > 0 && m_car.is_home)
-			SendAlertSensor(TESLA_ALERT_STATUS, 255, Disconnected, "Online", m_Name + " State");
+			SendAlertSensor(TESLA_ALERT_STATUS, 255, Home, "At home", m_Name + " State");
 		m_trycounter = 0;
 		return true;
 	}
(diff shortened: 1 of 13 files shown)


hardware/Tesla.cpp
```diff
@@ -0,0 +1,502 @@
...
+{
+	int sec_counter = 593;
+	int sec_counter_after_active = 0;
+	bool waking_up = false;
+	int interval = 1000;
+	Log(LOG_STATUS, "Worker started...");
+
+	while (!IsStopRequested(interval))
+	{
+		if (IsStopRequested(0))
```
gizmocuz: The second IsStopRequested is not needed.
When it's time to quit the worker, it stops directly

hardware/Tesla.h
```diff
@@ -0,0 +1,69 @@
+#pragma once
+
+#include "DomoticzHardware.h"
+#include "TeslaApi.h"
+#include "../main/json_helper.h"
```
gizmocuz: Not sure if you need this include here... (json_helper.h)

gizmocuz (commented):
Thanks for this hardware addon ! Some minor questions/comments

hardware/Tesla.h
MrHobbes74: indeed, not needed. Remainder of a previous change. Will remove.

MrHobbes74:
Note that I have separated the API from the worker. Except the result parameters (in json) the tesla details are all in TeslaApi. A next version I plan to fully decouple the worker from the car specifics, such that the worker can become an Electrical Vehicle, that can support various API's.

gizmocuz:
Thanks for the changes !!!
Just a question before i merge your code.... ;) ... do you plan to create more source files when you are going to decouple ?
If yes, maybe you could move all into a Tesla folder ? Keeps everything in one place

MrHobbes74:
Not planning, but new apis for other vehicles may result in new source files. Let me promise that in a next PR, after decoupling I will create an EV folder for all api’s and a generic EV worker.

gizmocuz:
Thanks!!

MrHobbes74:
Thanks for integrating!

ebretteville:
Hye @MrHobbes74 , just updated today to the latest beta and discovered by the official stable announcement that a tesla hardware plugin were created so I configured it right away. Everything worked fine. Only the lert indicator is missing... Any ideas why?

MrHobbes74:
Could you please bring this to the forum? There is a tesla topic. If it is missing, you did not activate it in the devices.

ebretteville:
> 
> 
> Could you please bring this to the forum? There is a tesla topic. If it is missing, you did not activate it in the devices.

I found the problem. Sentrymode wasn't activted, so the state device wasn't discovered.

MrHobbes74:
Sentrymode has nothing to do with the state device. But waking up the car could have been helpful ;-)

PR #3940, opened 2020-02-21
Tesla: fix merge error in vcxproj.filters and flippering alertstatus when charging

MrHobbes74:
Solve wrongly merged vcxproj filter file. Also small fix for flippering alertstatus when charging.


gizmocuz:
Thanks!!

Issue #3939, opened 2020-02-21
Add device to system in device list.


kovainfo:
When I add a new device to the system, the list looks like I didn’t add the device (the green arrow remains). And vice versa.
But if you refresh the page, you can see that the device has been added.

felix63:
related to #3922

kovainfo:
Thank you

PR #3944, opened 2020-02-22
#3939 Manually refresh devices table after changes to device

stas-demydiuk:
As a follow up of discussion in #3924


gizmocuz:
Thanks!!

Issue #3943, opened 2020-02-22
New OTGW devices


svkeulen:
There are OpenTherm GateWays that output a different format over serial. See the post:
https://domoticz.com/forum/viewtopic.php?f=26&t=23042
A time ago I wrote code that I am still using today (https://domoticz.com/forum/viewtopic.php?f=26&t=23042&start=20#p181457). 

Recently some other user was asking on the forum if the code ended up in the main developers branch. I replied that I will contact the developers about this.

Maybe it is a wise idea to add an extra device to the list. I don't know if there are any users using the current OTGW source.

gizmocuz:
@svkeulen , thanks for the post and links!
Yes I am using the current OTGW implementation, and i think some other users as well.
The reason why the new code (your code) is not being added to the development branch is that I haven't seen a Git Pull Request for this ...
Could you make a PR ? Then we can also see the changes
Do the old OTGW hardware devices still work with your code ? Or is it a matter of upgrading the OTGW firmware ?

svkeulen:
Old hardware doesn't work with my code. I think it has to do with the firmware version. I'm running version 4.2.5. The code changed quite a lot since the output differs a lot. I will make a PR

gizmocuz:
This was not a bug

PR #3945, opened 2020-02-23
Changes for OTGW with firmware version 4.2.5

svkeulen:
See https://github.com/domoticz/domoticz/issues/3943 and https://domoticz.com/forum/viewtopic.php?f=26&t=23042


gizmocuz:
I think you have made a pull request with very old code...
Please have a look at the changes, and also please make the appropriate changes (and reverts)
Why did you choose for different idx numbers ? yes, this will make sure that old setpoints are not working anymore. If you kept the original values then it was hopefully backwards compatible
Is it possible to do so ?

svkeulen:
Oh I forgot to branch when doing some compiling stuff on travis-ci... I reverted the pull request.
I'm afraid that it is just not a matter of some idxs that changed. The device outputs completely different text that needed to be parsed (see forum post for examples):

- previous firmware reports (I think, because I dont have it): 00000010/00000000,10.00,00000000/00000000,100.00,0/0,15.00,0.00,0.00,23.03,60.69,0.00,0.00,0.00,0/0,0/0,65.00,0.00,0,0,0,0,0,0,0,0
- frmware 4.2.5. only reports messages of length 9 like BC0090000

This code will NOT be backwards compatible.  Maybe we can make a NEW device in domoticz so you can choice between the old en new firmware type?
BTW, what firmware do you use on de OTGW?

gizmocuz:
I am using firmware  4.2.5
I know it is old, maybe two years old or even more.

I suspect more users have a old firmware, and therefor it is best if it is backwards compatible,
we have many log information, and also timers, scripts that control the thermostat
It has been quite some time since i worked on this code... I do remember that the old firmware also reported in the 'short' way, but you could enable the 'long' way by sending 'PS=1'
I can not believe that they made a new firmware where this is not backwards compatible ?
Could you try sending 'PS=1' with the otmonitor application ?
Could you find something related to this command and the new firmware ?

svkeulen:
Hmm we are using the same firmware version... I'm looking into the source of the OTGW assembler file now (https://otgw.tclcode.com/download/otgw-4.2.5.zip). It says:

> Messages on the OpenTherm link are reported as a letter follwed by 8 hex digits. The letter indicates if the message was received from the boiler (B), or from the thermostat (T). If the gateway modifies the message to the boiler in any way, the modified request is also reported (R), as is a modified answer back to the thermostat (A). The hex digits represent the message contents.

And is also says something about the commands it can handle:

> PS=1		Select whether to only print a summary line (1) on request, or every received/transmitted opentherm message (0)

If i remember corectly PS=1 or 0 both didn't work with the original domoticz code. I capture the "raw" text from the usb serial device and do some test to see what happens.  Can you send me some of your  output so i can compare them with mine?

svkeulen:
HA! Nailed it! When I issue a PS=1 it outputs the "old" output:

```
B500E6400
T00000300
B40000302
T900155E6
B500155E6
PS: 1    <--- My command
00000011/00000010,85.90,00000011/00000011,100.00,19/27,20.50,0.00,0.00,18.57,53.85,0.00,4.90,0.00,60/40,80/20,60.00,90.00,28656,20696,9192,12296,2076,3060,12934,156
OT: 4.90
SC: 16:39/3
PR: G=00
PR: I=00
PR: O=N
OT: 4.90
```

I'm almost sure I tried it over and over in the past and ending up with garbage.  Maybe it is an issue with the serial connection on a raspberry. The other person on the forum with the same problem is also using a raspi.

I also notice that for PS=1 you actively need to poll the OTGW as for PS=0 it reports every second or so.  So this code IS backwards compatible. 

I will do some testing on the device see if it is raspi related.

gizmocuz:
Maybe you are also user other applications/software with OTGW ?
You should not ....

gizmocuz:
And it is not OLD format , it is a different format!

svkeulen:
Sorry, I meant with "old" ps=1, since I wrote new code for ps=0. Do you mean with different that my ps=1 output  differs from yours?  

I tried it on windows as well with domoticz and ot monitor. With PS=1 mode  on the rpi the otgw resets  frequently (due to some watchdog-timer?). It echos commands in the middle of the summery, so domoticz is not able to parse the data. 
And third: it is just a summery. PS=0 contains all information on the bus. (Easier to debug or extend OpenTherm stuff)

So we could test if your device works with my code after giving a ps=0 manually. 

Do you use a USB (FTDI) or a lan OTGW? Someone mentioned it.

gizmocuz:
I suspect that you run another application as well that communicates with the OTGW that resets the PS status.
I use a old TP Link router that runs debian and use "ser2net" to communicate with the OTGW device
But any device that runs ser2net can do this
So we can confirm that there is no actual issue here in the code.
You could try to send the PS=1 more regular
We do not need more information, so PS=0 is not needed

svkeulen:
The devices crashes often with PS=1. And not only with my setup. At least two people experience the same. It is up to you now....

gizmocuz:
Using it for years without issues same as lots of others
Closing this as there is no big

PR #3946, opened 2020-02-23
Evohome demand, show percentage instead of On/Off only in bigtext element

Olifant1990:
Show % for Evohome heat demand devices as bigtext, just like dimmers


gizmocuz:
Thanks!

PR #3948, opened 2020-02-24
Tesla, Added Outside temperature sensor.

xbeaudouin:
Tesla cars have outside temperature sensor. Added it into Tesla hardware.


gizmocuz:
Thanks!!

PR #3953, opened 2020-02-27
make sure already installed tesla hardware deals well with new parameters

MrHobbes74:
make sure already installed tesla hardware deals well with new parameters (they will be 0)
also solve some compile warnings on linux


gizmocuz:
Thanks!!

PR #3955, opened 2020-02-28
Add DisableLogAutoUpdate and AddDBLogEntry device options 

guillaumezin:
Theses options are for counters, to be enabled from plugins

DisableLogAutoUpdate prevents Domoticz to update Meters and MultiMeters DB on a regular basis
AddDBLogEntry permits to insert data directly inside Meter / Meter_Calendar / MultiMeter / MultiMeter_Calendar tables

This is a simpler way to achieve what was intended in PR #3877 

Maybe we could alse remove the whole sTypeManagedCounter logic after this PR, it shouldn't be needed anymore


gizmocuz:
Thank you for the PR, some questions:
- can you use std::stoll ? makes the code also more readable
- UpdateCalendarMultiMeterLong vs Short, the difference is the addition of the counter values ? Could those not be made optional in the function headers? I would like it to be still one function, or even better, keep the 'AddCalendarUpdateMeter' function as is ?

guillaumezin:
Done !

gizmocuz:
Thank you for the changes !
Currently I have a email conversation with someone trying to get the Managed Meter to work...
I have NO IDEA how it works... it seems not documented, also not on the WIKI/JSON page which makes it very hard for everyone to implement.
Do you have any idea ? And if yes, would it be possible you could update the Wiki/JSON page to add this type of counter ?

Now in this code/PR, it confuses me what ShortLog is ? Why is this needed ?
For example, (if i understand correctly), A managed counter consists of a counter value (semicolon) date/time ?
When this is updated via JSON, why not store this directly in the database ?
ShortLog means 'DateOnly' ? or DateTime ? The ShortLog name is confusing...
Is the Managed Counter updated once a day ? If no, what is the difference with the current kWh counter ?

guillaumezin:
Hello,

I agree, the Managed Counter is mentionned on Wiki/JSON but it is very light, I must improve that.

Managed Counters are counters "in the cloud". You don't have real time data, the data are collected by a third party which gives you a website or an API to access you consumption data afterwards. The idea is to inject data in Domoticz DB to let user easily know their consumption. But usually, you don't have the absolute counter value, only usage difference day to day and hour to hour.

The goal of the different patches I made is to permit to insert data in Meter or Meter_Calendar table or MultiMeter or MultiMeter_Calendar table. What I call ShortLog = true is the Meter / MultiMeter table that shows last 7 days of usage in counter log, ShortLog = false is the Meter_Calendar / MultiMeter_table that shows last week / month / year usage in counter log. EnableLogDBAccess option and sTypeManagedCounter allow that).

Because the data are managed by an external script, from not real time data, I don't want Domoticz to manipulate automatically data between Meter and Meter_Calendar tables for instance, every 24 hours (DisableLogAutoUpdate option and sTypeManagedCounter block that).

To know insert data in theses tables with JSON or script, you give the value1;value2;...;datetime as svalue, if the datetime is date only, it is for Meter_Calendar / MultiMeter_table (ShortLog = false), it datetime is date AND time, it is for Meter / MultiMeter table (ShortLog = true)

So Managed Counters is very close to kWh counter, but it permits to interact directly with last 7 days / week / month / year tables, and it hides absolute counter value. But as I said earlier, if you accept the DisableLogAutoUpdate and EnableLogDBAccess options approach, which are more generic, sTypeManagedCounter could be obsoleted.

In my plugin (https://github.com/guillaumezin/DomoticzLinky), I grab data every 24hours, but each time, I insert a few days of data, because the third party that collects data I want to inject is not very reliable, some day some data are missing.

Maybe with this patch, we won't keep or encourage the usage of sTypeManagedCounter but only of DisableLogAutoUpdate and EnableLogDBAccess options, I will improve the Wiki depending on that, let me know.

gizmocuz:
Thank you for the explanation!

What do you mean by "value1;value2;...;datetime"
Does that mean you can have multiple counters ? If yes, why not use only the multimeter table ?
It has room for 6 counters
If you have a time, it should use the multimeter table, if it has only a date it should use the multimeter_calendar table ?

guillaumezin:
Sorry, I was giving a generic example.

For sTypeManagedCounter and simple counters with EnableLogDBAccess enabled:

sValue must be a single value to update Dashboard, for instance "1234"
sValue must be a value followed by a space and a date ("%Y-%m-%d" format) to update last week/month/year history, for instance "123456;78;2019-09-24" -> Meter_Calendar table, 123456 is counter value and 78 usage
sValue must be a value followed by a space a date a space and a time ("%Y-%m-%d %H:%M:%S" format) to update last days history, for instance "123456;78;2019-10-03 14:55:00" --> Meter

For P1 smart meters or other multi meters with EnableLogDBAccess enabled:

sValue must be a value to update Dashboard: "USAGE1;USAGE2;RETURN1;RETURN2;CONS;PROD"
sValue must be a value followed by a space and a date ("%Y-%m-%d" format) to update last week/month/year history: "USAGE1;USAGE2;RETURN1;RETURN2;CONS;PROD;2019-09-24" -> MultiMeter_Calendar table
sValue must be a value followed by a space a date a space and a time ("%Y-%m-%d %H:%M:%S" format) to update last days history: "USAGE1;USAGE2;RETURN1;RETURN2;CONS;PROD;2019-10-03 14:55:00" --> MultiMeter
Optionnaly, absolute counter values can be appended:

USAGE1;USAGE2;RETURN1;RETURN2;CONS;PROD;COUNTER1;COUNTER2;COUNTER3;COUNTER4;2019-09-24" -> MultiMeter table

USAGE1;USAGE2;RETURN1;RETURN2;CONS;PROD;COUNTER1;COUNTER2;COUNTER3;COUNTER4;2019-10-03 14:55:00" -> MultiMeter_Calendar table

And I was wrong, there is a brief explanation of Managed Counter in https://www.domoticz.com/wiki/Developing_a_Python_plugin, not in JSON Wiki page, and it is outdated, I will update it.

gizmocuz:
Thanks again,

I think you need to swap the tables, the short-log tables are without '_Calendar' and contains the time
The calendar tables are for daily logs and do not contain time
Because they are daily values, they are used for day/month/week graphs
The shortlog for the last xx days

guillaumezin:
You're right, I swaped in my explanations, but the code is right, and the Wiki too (I updated it). I edited my comment.

gizmocuz:
Thanks! Is it possible to also update the JSON wiki ? As i think this is where most users will start programming against

guillaumezin:
Yes, but with sTypeManagedCounter or DisableLogAutoUpdate and EnableLogDBAccess options explanations?

gizmocuz:
For me the variable name is unclear ... "EnableLogDBAccess" ? It means AddDBLogEntry ?

guillaumezin:
It means "authorize to insert data in Meter/MultiMeter/Meter_Calendar/MultiMeterCalendar" db.

I can change to AddDBLogEntry if you want

gizmocuz:
Authorize ? Yes you mean to add a database log entry ... so yes, if possible, please name the variable like that ;)
I think when it is good documented in the JSON Wiki it would be clear...

guillaumezin:
I documented Managed counter in JSON wiki:
https://www.domoticz.com/wiki/Domoticz_API/JSON_URL%27s#Managed_counter

I changed EnableLogDBAccess options to AddDBLogEntry

gizmocuz:
Thank you !!

s0nik42:
Hello guys, sorry to jump in but I'm using "Managed Counter" not from a plugin  but from external scripts.

A great feature of the sTypeManagedCounter  is "historical DATA integrity"

By timestamping the data you send to domoticz from your node, It allows you to implement cache on nodes : 
- To manage disruption like : (a node not available for a certain period of time because of network failure, battery issue, server crash, etc...). . your nodes can keep the data locally and send it to the server once the communication is back. Then it push the data to domoticz with the right Timestamp. 
- To save battery on nodes. it's not connected all day, but stay off most of the time and bulk send the update once per  day or once per week. 

I don't think any other device allow that as they timestamp the data at the time domoticz receive it. 

I think it will be great to allow it for any type of devices, but it will produce a lot of side effect...  I'm fine removing sTypeManagedCounter as long as : 
- the options listed above will be accessible from the API
- we can migrate the existing ManagedCounter to another type of Counter.

Also @guillaumezin , regarding the documentation, I've got a bunch of questions for you. I'm going to open a post on the forum. 

Have a nice day guys. This software is great !

PR #3956, opened 2020-02-28
Allow compiling with external JSonCPP

xbeaudouin:
This small patch permit to compile if the user / maintainer need / want Domoticz with an external JSonCPP library. 
Avoiding to have code that is not shared by the project.
This patch works on my FreeBSD port for days, without any pain.
Please consider adding this into Domoticz since it does not change the original way it is compiled (like for example zlib / minizip dependency).
Regards,

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -126,6 +126,28 @@ ELSE(USE_BUILTIN_SQLITE)
   ENDIF(SQLite_FOUND)
 ENDIF(USE_BUILTIN_SQLITE)
 
+option(USE_BUILTIN_JSONCPP "Use builtin JSonCPP" YES)
+IF(USE_BUILTIN_JSONCPP)
+  MESSAGE(STATUS "Using builtin JSonCPP")
+  set (domoticz_jsoncpp_SRCS
+       json/json_reader.cpp
+       json/json_value.cpp
+       json/json_writer.cpp
+  )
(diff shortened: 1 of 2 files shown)


gizmocuz:
how many extra bytes will this actually save ? and how many extra CPU cycles will this cost by adding a pre-defined value that is passed to every source/header file during compile time ?

xbeaudouin:
Well I cannot answer to this question. The library on FreeBSD /AMD64 takes ~260k.
The CPU cycle taken depends also of the compiler. Since the precompilator is mostly fast I dunno how many cpu cycle it will take.
Notice I have set this to OFF per default like :
- lua
- zlib 
- openzwave
- telldus
- minizip 
- mqtt
- sqlite

Adding this simple check, tested once (since the header status will be somewhere is compliler cache), will not double or more the time to compile domoticz.
I understand your statement : the lib is small, so add the whole code into domoticz doesn't care especially on arm platform and really you are right static code is faster than dynamic code.
But since it will touch only one header file, and the CMake system, I don't see why you seem reluctant to add some flexibility for package maintainers.
Regards

gizmocuz:
FreeBSD is quite up to date, but if i look at for example at Debian it seems the libjsoncpp is quite old. (1.7.4-3)
But after more reading it seems indeed to have very little impact.
Maybe someday the complete Domoticz project structure should be redefined...

xbeaudouin:
Thanks !

PR #3961, opened 2020-03-02
Add HTTPClient PUT and DELETE support

dwmw2:
The code for these is here but wasn't wired up. For my cat flap I need to use PUT support.

### DIFF
--- a/httpclient/HTTPClient.cpp
+++ b/httpclient/HTTPClient.cpp
@@ -386,6 +386,8 @@ std::vector<std::string> &vHeaderData, const long TimeOut)
 		if (TimeOut != -1)
 			curl_easy_setopt(curl, CURLOPT_TIMEOUT, TimeOut);
 
+		curl_easy_setopt(curl, CURLOPT_HEADERFUNCTION, write_curl_headerdata);
+		curl_easy_setopt(curl, CURLOPT_HEADERDATA, &vHeaderData);
 		curl_easy_setopt(curl, CURLOPT_WRITEDATA, (void *)&response);
 		curl_easy_setopt(curl, CURLOPT_URL, url.c_str());
 		//curl_easy_setopt(curl, CURLOPT_PUT, 1);
@@ -454,6 +456,8 @@ std::vector<std::string> &vHeaderData, const long TimeOut)
 		if (TimeOut != -1)
 			curl_easy_setopt(curl, CURLOPT_TIMEOUT, TimeOut);
 
+		curl_easy_setopt(curl, CURLOPT_HEADERFUNCTION, write_curl_headerdata);
+		curl_easy_setopt(curl, CURLOPT_HEADERDATA, &vHeaderData);
 		curl_easy_setopt(curl, CURLOPT_WRITEDATA, (void *)&response);
 		curl_easy_setopt(curl, CURLOPT_URL, url.c_str());
 		//curl_easy_setopt(curl, CURLOPT_PUT, 1);
--- a/httpclient/HTTPClient.h
+++ b/httpclient/HTTPClient.h
@@ -10,7 +10,9 @@ public:
 	enum _eHTTPmethod
 	{
 		HTTP_METHOD_GET,
-		HTTP_METHOD_POST
+		HTTP_METHOD_POST,
+		HTTP_METHOD_PUT,
+		HTTP_METHOD_DELETE
 	};
 
 
--- a/main/SQLHelper.cpp
+++ b/main/SQLHelper.cpp
@@ -3410,6 +3410,14 @@ void CSQLHelper::Do_Work()
 				{
 					ret = HTTPClient::POST(itt->_sValue, postData, extraHeaders, response, headerData, true, true);
 				}
+				else if (method == HTTPClient::HTTP_METHOD_PUT)
+				{
+					ret = HTTPClient::PUT(itt->_sValue, postData, extraHeaders, response, headerData, true);
+				}
+				else if (method == HTTPClient::HTTP_METHOD_DELETE)
+				{
+					ret = HTTPClient::Delete(itt->_sValue, postData, extraHeaders, response, headerData, true);
+				}
 				else
 					return; //unsupported method
 


gizmocuz:
Thanks !! Could you post a link to your cat flap ? ;)

dwmw2:
Cat flap based on https://github.com/alextoft/sureflap

Looks a bit like this, based on virtual switches. I am currently trying to debug why I need to set the level for them numerically (10, 20, etc.) because otherwise, they get set to zero (Kept out) immediately after setting them. I suspect a Domoticz bug but I'm not sure yet.

For this kind of thing we'd really like a 'Set the value but don't do triggers' option, FWIW. Which might be related...

```
local apibase='https://app.api.surehub.io/api/'

local function login(domoticz, callback)
  return 'foo'
  end

local tags = {
	['Truckle'] = 102331;
	['Frank'] = 102333;
	['Daisy'] = 102427;
};
local tag_cats = {
	[102331] = 'Truckle';
	[102333] = 'Frank';
	[102427] = 'Daisy';
};

local profiles = {
	['Kept out'] = 5;
	['Allowed out'] = 2;
	['Kept in'] = 3;
};

local profile_levels = {
	[5] = 'Kept out';
	[2] = 'Allowed out';
	[3] = 'Kept in';
};

local level_numbers = {
      ['Kept out'] = 0;
      ['Allowed out'] = 10;
      ['Kept in'] = 20;
};

local function refresh(domoticz)
	local options = {
	  	url = apibase .. 'device/422826/tag',
		method = 'GET',
		headers = {
		    ['Authorization'] = 'Bearer ' .. domoticz.data.token
		},
		callback = 'Poll'
	    }
	domoticz.openURL(options)
end

local function update(domoticz, pet, profile)
	local tag = tags[pet]
	local options = {
	    	url = apibase .. 'device/422826/tag/' .. tag,
		method = 'PUT',
		postData = '{"profile":' .. profile .. '}',
		headers = {
		    ['Content-Type'] = 'application/json',
		    ['Authorization'] = 'Bearer ' .. domoticz.data.token
		},
		callback = pet
	    }
	domoticz.openURL(options)
end

local function setlocal(domoticz, cat, profile)
	local level = profile_levels[profile]
	if (cat ~= nil and level ~= nil) then
		local catdev = domoticz.devices(cat)
		if (catdev ~= nil and catdev.state ~= level) then
		   	domoticz.log("Set " .. cat .. " to " .. level_numbers[level], domoticz.LOG_INFO)
			domoticz.data.netstatus[cat] = level
			catdev.switchSelector(level_numbers[level])
		end
	end
end

return {
        on = {
                devices = {
                        'Truckle',
                        'Frank',
                        'Daisy'
                },
		httpResponses = {
			'Truckle',
			'Frank',
			'Daisy',
			'Poll',
			'Truckle-login',
			'Frank-login',
			'Daisy-login',
			'Poll-login'
		},
		timer = {'every minute'}
        },
	data = {
		'token',
		'etags',
		'set',
		'netstatus'
	},
        execute = function(domoticz, device)
		if (domoticz.data.etags == nil) then
		   domoticz.data.etags = {}
		end
		if (domoticz.data.set == nil) then
		   domoticz.data.set = {}
		end
		if (domoticz.data.netstatus == nil) then
		   domoticz.data.netstatus = {}
		end
		if (domoticz.data.token == nil) then
                        -- FIXME: proper login support
			domoticz.data.token = '.....secret.....'
		end

		if (device.isTimer) then
			domoticz.log('Cat flap poll', domoticz.LOG_INFO)
			refresh(domoticz)
			return
		elseif (device.isHTTPResponse) then
	    		domoticz.log(device.callback .. ' Response code ' .. device.statusCode .. ' data: ' .. device.data, domoticz.LOG_INFO)
			if (device.callback == 'Poll') then
				for i,rec in pairs(device.json.data) do
					local cat = tag_cats[rec.id]
					if (cat ~= nil and domoticz.data.set[cat] == nil) then
						print('ID ' .. rec.id .. ' Cat ' .. cat .. ' profile ' .. rec.profile)
						setlocal(domoticz, cat, rec.profile)
					end
				end
				return
			elseif (tags[device.callback] ~= nil) then
			       domoticz.data.set[device.callback] = nil
			       setlocal(domoticz, device.callback, device.json.data.profile)
			end
--			for h in pairs(device.headers) do
--				 domoticz.log(h .. ': ' .. device.headers[h], domoticz.LOG_INFO)
--			end
			return
		end

		if (domoticz.data.netstatus[device.name] == device.state) then
			domoticz.log('Cat ' .. device.name .. ' incoming change to ' .. device.state , domoticz.LOG_INFO)
		else
			domoticz.log('Cat ' .. device.name .. ' was changed locally to ' .. device.state , domoticz.LOG_INFO)

			local profile = profiles[device.state]

			domoticz.data.set[device.name] = profile
			update(domoticz, device.name, profile)
		end
		domoticz.data.netstatus[device.name] = nil
        end
}


```

gizmocuz:
Thanks! going to look at this project ;)

What do you mean with

"For this kind of thing we'd really like a 'Set the value but don't do triggers' option"

because this is already supported in the JSON api, and I think also with dzVents

Another option is to use a ESP8266/MQTT maybe ?

dwmw2:
> What do you mean with 
> "For this kind of thing we'd really like a 'Set the value but don't do triggers' option"
> 
> because this is already supported in the JSON api, and I think also with dzVents

As noted on Slack, what happens is..

 My script above polls the catflap web API, notices a device has changed state.
 It calls  (e.g.) `domoticz.devices([NAME]).switchSelector('Kept in')`
 My script is immediately invoked again by the trigger on that device, and writes the new state *back* to the web API where it just came from.

I've worked around this by using the `domoticz.data.netstatus[]` array and not triggering the outbound setting if it's equal to what we've just 'received' but it's a bit of a hack.

gizmocuz:
Have you tried this:

 - Added `.silent()` option to commands like `switchOn().afterSec(4).silent()` causing no follow-up events to be triggered. This also works when updating non-switch-like devices, scenes, groups and user variables. If you do not call `silent()``, a follow-up event is *always* triggered (for devices, scenes, groups and user variables).

dwmw2:
Aha, nice. Will play with that; thanks!

Issue #3960, opened 2020-03-01
Decimal places - data transmitted to influx


ravsit-activum:
https://github.com/domoticz/domoticz/blob/ab401616e317170fc31bf8396c5ce10befa52784/push/BasePush.cpp#L515

If this is not a problem, I suggest increasing the accuracy of the data transmitted to influx.

In: BasePush.cpp
1. for Temperature - line 504:
sprintf (szData, "**% .2f**", tvalue);

2. for humidity - line 515:
sprintf (szData, "**% .2f", atof (rawsendValue.c_str ())**);

gizmocuz:
How about using %g ? This will use a integer for integer and has no rounding issues
Do you have such a accurate temp sensor ? You need to calibrate the sensor as well if you want to have a correct value
Else 1 decimal should be sufficient

gizmocuz:
I changed the code to use %g now, If you compile yourself, could you test this ?
Else it will be available in beta 11761 which should be there for x86/arm systems within the hour.
Currently there is a download issue on the boost website that prevents automatic compilation for travis (ubuntu) builds

ravsit-activum:
Domoticz save data with 2 decimals - so it is better to send it to influx too.
I have sensor with such accuracy.

from devicestatus table:

ID | HardwareID | DeviceID | Unit | Name | Used | Type | SubType | SignalLevel | BatteryLevel | nValue | sValue | LastUpdate
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
30 | 10 | 82030 | 1 | Temp1 | 1 | 80 | 5 | 3 | 100 | 0 | 23.81 | 27.02.2020 20:40
32 | 10 | 82032 | 1 | Temp3 | 1 | 84 | 1 | 7 | 100 | 0 | 21.65;64.89;0;972;0 | 27.02.2020 20:40
65 | 16 | 82065 | 1 | Temp4 | 1 | 82 | 1 | 3 | 100 | 0 | 22.48;30.81;0 | 30.10.2019 13:01
81 | 19 | 130007 | 7 | Airly - Ciśnienie powietrza | 1 | 243 | 26 | 12 | 255 | 0 | 1001.61;0 | 27.02.2020 20:35

for Barometer - line 523:
		sprintf(szData, "%.2f", atof(rawsendValue.c_str()));

I prefere %.2f because of %g can go to string with e+, for example: 5.6e+4

[...]

PR #3964, opened 2020-03-02
Pushers, Humidity as float, not integer

ravsit-activum:
Related discussion:
https://github.com/domoticz/domoticz/issues/3960


gizmocuz:
Thanks!

PR #3968, opened 2020-03-06
Allow domoticz to build with external cereal

xbeaudouin:
Hello,
On FreeBSD there is cereal 1.3.0 as ports (on devel/cereal), this patch allow building domoticz with cereal from FreeBSD port.
Per default builtin cereal is used so it should not change anything for other build.
I am open to any suggestions about this kind of patch.

Regards

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -209,6 +209,21 @@ IF (INCLUDE_SPI)
 	ENDIF (HAVE_LINUX_SPI_H)
 ENDIF (INCLUDE_SPI)
 
+option(USE_BUILTIN_CEREAL "Build with built-in cereal" YES)
+include(CheckIncludeFileCXX)
+if(USE_BUILTIN_CEREAL)
+  find_path(cereal_INCLUDE_DIR cereal/cereal.hpp PATHS ${CMAKE_CURRENT_LIST_DIR})
+else(USE_BUILDIN_CEREAL)
+  find_path(cereal_INCLUDE_DIR cereal/cereal.hpp)
+endif(USE_BUILDIN_CEREAL)
+mark_as_advanced(cereal_INCLUDE_DIR)
+include_directories(${cereal_INCLUDE_DIR})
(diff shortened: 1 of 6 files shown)


gizmocuz:
Thanks for the PR, please have a look at the appveyor build log

xbeaudouin:
Ok seems to have in issue with windows build... Will try to fix that....

gizmocuz:
I am trying to see what the benefit of the PR is....
I mean when you install the cereal package on your system with apt you get a lot of files/binaries that you will not need on the system

https://packages.debian.org/buster/all/cereal/filelist

When you install the -dev package on your system, you do get the same headers that are included in Domoticz, or maybe they can even brake compilation of domoticz
It is not that there is a library installed on your system

We might change the external dependencies from Domoticz in a next stable version (not the next)
Removing all third party libraries/code from the project and be totally depending on external installed packages by the developer (and user), or we link them statically by default where possible so the end-user does not need to install this package while updating

xbeaudouin:
Well I totaly agree with you. On FreeBSD the cereal dependency can be added "on build", eg for the package manager (poudriere), the files are just added when compiling, but the binary package does not depend on it on installation.
About compilation I agree we have to remove the third party projects that are not part of domoticz, like for example on the build process ask for -dev pkg as a main dependcy.
For binary distribution we may build 2 kind of domoticz :
- one static
- one dynamic
So I am really open to see how we can make this better. 
The benefit of this PR in my point of vue : less warnings FreeBSD. Since clang 8 is really picky about C++ this can be positive.
Regards

gizmocuz:
We have to see if a Dynamic binary release will work on most systems in the near feature
Closing this PR for now as I hope to remove all third party libraries in the project folder

xbeaudouin:
Ok. Will add this to my FreeBSD port per default until third party libs are removed from the project folder (by the way TinyXPath is really old and this seems to not really maintained now )...

PR #3972, opened 2020-03-08
OpenThermGateway hardware: Fixed Setpoint and OpenThermMonitor compatibility

bramvreugd:
The two changes in this fix of the OTGW (OpenTherm Gateway) are:
1 Changed SetPoint from readonly to read/write. 
The setpoint in the OTGW implementation was only able to look at the setpoint set by the thermostat.
With this addition/you can also overwrite the setpoint. Setting a value of zero will cancel the overwrite and the boiler will use the thermostat value again.  
This is very useful in case of controlling the boiler in case of many restarts.

2 Changed PS behaviour PS=1 and PS=0 to be able to use OpenThermMonitor and Domoticz paralel.

### DIFF
--- a/dzVents/runtime/device-adapters/opentherm_gateway_device.lua
+++ b/dzVents/runtime/device-adapters/opentherm_gateway_device.lua
@@ -19,7 +19,7 @@ return {
 		function device.updateSetPoint(setPoint)
 			-- send the command using openURL otherwise, due to a bug in Domoticz, you will get a timeout on the script
 			local url = domoticz.settings['Domoticz url'] ..
-					'/json.htm?param=udevice&type=command&idx=' .. device.id .. '&nvalue=0&svalue=' .. setPoint
+					'/json.htm?param=udevice&type=command&idx=' .. device.id .. '&nvalue=' .. toNumber(setPoint) .. '&svalue=' .. setPoint
 			return domoticz.openURL(url)
 		end
 
--- a/hardware/OTGWBase.cpp
+++ b/hardware/OTGWBase.cpp
@@ -27,6 +27,7 @@ OTGWBase::OTGWBase(void) :
 	m_OutsideTemperatureIdx=0;//use build in
 	m_bufferpos = 0;
 	m_OverrideTemperature = 0.0f;
+    m_OverrideControlTemperature = 0.0f;
 	m_bRequestVersion = true;
 }
 
@@ -120,9 +121,7 @@ void OTGWBase::UpdateSetPointSensor(const unsigned char Idx, const float Temp, c
 	thermos.id3=0;
 	thermos.id4=Idx;
 	thermos.dunit=0;
-
 	thermos.temp=Temp;
-
 	sDecodeRXMessage(this, (const unsigned char *)&thermos, defaultname.c_str(), 255);
 }
 
@@ -251,6 +250,14 @@ void OTGWBase::GetGatewayDetails()
 	WriteInt((const unsigned char*)&szCmd, (const unsigned char)strlen(szCmd));
 }
 
+void OTGWBase::SwitchToPS0()
+{
+	char szCmd[30];
+	strcpy(szCmd, "PS=0\r\n");
+	WriteInt((const unsigned char*)&szCmd, (const unsigned char)strlen(szCmd));
+}
+
+
 void OTGWBase::SendTime()
 {
 	time_t atime = mytime(NULL);
@@ -287,6 +294,8 @@ void OTGWBase::SetSetpoint(const int idx, const float temp)
 		_log.Log(LOG_STATUS, "OTGW: Setting Control SetPoint to: %.1f", temp);
 		sprintf(szCmd, "CS=%.1f\r\n", temp);
 		WriteInt((const unsigned char*)&szCmd, (const unsigned char)strlen(szCmd));
+		UpdateSetPointSensor((uint8_t)idx, temp, "Control Setpoint");
+		m_OverrideControlTemperature = temp;
 	}
 	else if (idx == 5)
 	{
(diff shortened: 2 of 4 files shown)


hardware/OTGWTCP.cpp
```diff
@@ -64,6 +64,7 @@ void OTGWTCP::Do_Work()
 {
 	int sec_counter = 25;
 	connect(m_szIPAddress,m_usIPPort);
+	int ps;
```
gizmocuz, line 67: Could you Initialize this variable

hardware/OTGWBase.cpp
```diff
@@ -395,19 +410,24 @@ void OTGWBase::ParseLine()
 			SendPressureSensor(0, idx - 1, 255, _status.CH_water_pressure, "CH Water Pressure");
 		}
 
-		_status.Room_temperature = static_cast<float>(atof(results[idx++].c_str()));						SendTempSensor(idx - 1, 255, _status.Room_temperature, "Room Temperature");
-		_status.Boiler_water_temperature = static_cast<float>(atof(results[idx++].c_str()));				SendTempSensor(idx - 1, 255, _status.Boiler_water_temperature, "Boiler Water Temperature");
-		_status.DHW_temperature = static_cast<float>(atof(results[idx++].c_str()));							SendTempSensor(idx - 1, 255, _status.DHW_temperature, "DHW Temperature");
-		_status.Outside_temperature = static_cast<float>(atof(results[idx++].c_str()));						SendTempSensor(idx - 1, 255, _status.Outside_temperature, "Outside Temperature");
-		_status.Return_water_temperature = static_cast<float>(atof(results[idx++].c_str()));				SendTempSensor(idx - 1, 255, _status.Return_water_temperature, "Return Water Temperature");
+		_status.Room_temperature = static_cast<float>(atof(results[idx++].c_str()));
```
gizmocuz, line 413: Could you undo all these new lines ? I don't think any code has changed on this part ?

gizmocuz (commented):
Thank you for your PR, some comments made.
Also some additional questions, SetPoint always worked, what was not working ?
Why do you want to keep using otmonitor ? Not to happy with PS=0, works without issues with PS=1

hardware/OTGWBase.cpp
bramvreugd: You're right. I will change it back. I accidentally changed it when I was trying to understand the code. The formating looked somewhat strange . The second statement on the last part of the line was not visible in my editor.

hardware/OTGWTCP.cpp
bramvreugd: Thanks for the comment.

bramvreugd:
I checked if the control setpoint is working on version Version: 4.11780 Build Hash: 10a4759be Compile Date: 2020-03-10 09:15:03
Domoticz reads the control setpoint from the OTGW device correctly. But the OTGW firmware has the functionality to override the control setpoint.  Version: 4.11780 can't do that. There is only an adjustment possibility which does not sends its value to OTGW.
Overriding can be useful for situation where you want 
 -  To save energy by lowering the setpoint. Some boiler allow that on there control panel. Some not.
- Change the water temperature of the heating system depending on the weather temperature. If it's cold a higher temperature is needed to warm your home.
- Prevent excessive restarting of burner by temporary lowering the control setpoint.

Testing this again showed a downside of this fix: Domoticz creates a new device in the utility tab. The old device in the temperature tab stops working. I don't know what to do about that. I think it should be a device like a Thermostat with override and cancel override. I don't know how to do that.

The pull request title is misleading This is not a bug with a fix instead  it is missing functionality and a feature enhancement.

About the ps=0.
I can see you point. I always try to leave a device in a state that other programs/systems can use it too. 
There are quite a few discussions on forums of people who wants to use Opentherm monitor in parallel. I use it sometimes to monitor the burner restarts and the return water temperature.  For that I do needs very frequent updates. Opentherm monitor shows a value every 5 seconds.
It is not a must have feature for me. I always can compile it for the few occasions I do need it.

gizmocuz:
If you use the same IDX, then no new device should be created
Maybe i am talking about the wrong setpoint here, but there is no problem setting the temperature to a new value with Domoticz, you can do this from the Utility tab or you can add timers to your temperature setpoint
This has always worked ?

bramvreugd:
I will check again.

bramvreugd:
The device "control setpoint" is on the temperature tab. There is no button for timers nor set button.For other devices like thermostats it is working well. I'm very happy with Domoticz.

gizmocuz:
But the SetPoint device is on the Utility tab

gizmocuz:
image

bramvreugd:
This is how "control setpoint" looks like with the normal Domoticz.
 
image
Just named it two to keep them apart during testing.

And this is how it look after my changes are applied.

image

gizmocuz:
This is the normal setpoint to adjust the temperature on the thermostat ?
As you can see in my screenshot, this is already there (On the Utility tab)

bramvreugd:
In januari 2014 you wrote the following: [https://www.domoticz.com/forum/viewtopic.php?t=211&start=60] 

gizmocuz wrote:
I am quite new to opentherm/thermostats/boilers....

Besides the Room SetPoint i see three more:
- Control Setpoint (MsgID=1)
- DHW Setpoint (MsgID=56)
- MAX_CH Water Setpoint (MsgID=57)

It seems the above three SetPoints are automatically adjusted by someone (therm/boiler?) Do we(Can we) want to control this?

It also seems we can set two more setpoints (besides the room):
- Setpoint Heating (It is not clear for me which one this is, MsgID1 or MsgID=57)
- Setpoint Water (this is the DHW i think? MsgID=56)

If they are controlled automatically by the boiler (because of a room setpoint), do we want to control the above two? And why?_
"

I did no found that discussion before.  I want to change the  Control Setpoint (MsgID=1)  for the reasons mentioned before.

gizmocuz:
Okey clear, so it is NOT the room setpoint but another

Okey, then of course we get a new device, as you now report/handle it as setpoint.

No problem, but please have a look in SQLHelper.cpp how we can perform database patches.
It would be great if you can add one that changes the temperature device to a setpoint device and keep the log/data

bramvreugd:
Yes. I will look into the sqlhelper. You probably did the same way back in 2014 with the other setpoint.

gizmocuz:
Did you manage to have a look ?

bramvreugd:
It wasn't easy and I have done already most things. I'm struggling and
learning to work with git. Unfortunately if been in hospital for 8 weeks
with covid,corona . Now I am learning to walk again and such things. It
will take some time before I am able to continue with git because it will
take some weeks before I can go home.
I'm sorry for the delay

Best regards,
Bram

gizmocuz:
@bramvreugd , hope you're OK again.

dzVents/runtime/device-adapters/opentherm_gateway_device.lua
```diff
@@ -19,7 +19,7 @@ return {
 		function device.updateSetPoint(setPoint)
 			-- send the command using openURL otherwise, due to a bug in Domoticz, you will get a timeout on the script
 			local url = domoticz.settings['Domoticz url'] ..
-					'/json.htm?param=udevice&type=command&idx=' .. device.id .. '&nvalue=0&svalue=' .. setPoint
+					'/json.htm?param=udevice&type=command&idx=' .. device.id .. '&nvalue=' .. toNumber(setPoint) .. '&svalue=' .. setPoint
```
gizmocuz, line 22: A setpoint device stores its temperature in svalue (like 21.5) , not in the nvalue (which is a integer)

gizmocuz:
@bramvreugd , I still don't know what to do with this PR... the OTGW is functioning for everyone for years.
It seems you want to control it from two applications and that could/should not be done (otmonitor),
but you can send JSON API requests to Domoticz to control it.
If you miss a setpoint type to control, please make a new PR based on the current code
Going to leave the PS Mode as it currently is

Issue #3978, opened 2020-03-13
[GUI devices tab] Group / scenes rows are replaced by a copy of the row of the device with the same id when the device is updated. 


rwaaren:
@stas-demydiuk Stanislav Demydiuk

Hi can you please have a look at the issue of replacing the group / scenes by the device with the same id when the device is updated. [Forum post P1 Meter - duplicate devices](https://www.domoticz.com/forum/viewtopic.php?f=14&t=31659) 
I expect it is related to #3914

rwaaren:
Thx for the quick fix. An extra clear browsercache was required to activate it.

PR #3980, opened 2020-03-13
#3978 Differentiate devices and scenes update

stas-demydiuk:


gizmocuz:
Thank you for the quick fix !!

PR #3981, opened 2020-03-14
Tesla: Introduced optional keep-asleep mode

MrHobbes74:
This will prevent the car to be woken up by domoticz.

This prevents battery drain. In this mode, car updates are only retrieved at the configured
interval when the car is awake. In case the car wakes up by itself for any reason, an update is
also retrieved. User commands sent to the car (e.g. start charge) will always wake up the car.

Also reconfigured the state device, it now provides only a new value when it's different from the
previous value.

### DIFF
--- a/hardware/eVehicles/TeslaApi.cpp
+++ b/hardware/eVehicles/TeslaApi.cpp
@@ -68,6 +68,11 @@ bool CTeslaApi::RefreshLogin()
 	return false;
 }
 
+int CTeslaApi::GetSleepInterval()
+{
+	return 20;
+}
+
 bool CTeslaApi::GetAllData(tAllCarData& data)
 {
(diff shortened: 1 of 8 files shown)


Nougreche:
labview

MrHobbes74:
@gizmocuz can you integrate this PR? There are no issues found after running for a week :-)

gizmocuz:
Thanks for the work again !!

MrHobbes74:
Thanks a lot!

gohamstergo:
@MrHobbes74 Great work on this module! I've started some suggestions on the forums here: https://www.domoticz.com/forum/viewtopic.php?f=31&t=31781&p=240402

Sorry to semi-hijack these comments, couldn't find another way to get in touch.

PR #3992, opened 2020-03-19
- websocket patch

rkluwen:
- Fixing WS_Write() in case data is sent whilst the websocket connection is not completely set up yet
- Making connecion_type an enum class


gizmocuz:
Thank you !

PR #3993, opened 2020-03-20
Correct report headers for energy generation

rklomp:
The headers when having a generation type are currently showing Usage and Cost. This change modifies this to Generation and Earnings


gizmocuz:
Thank you !!

PR #4015, opened 2020-03-26
Fix sign extension and overflow problems with fault log date handling

dwmw2:
If we cast payload[11]<<24 to a signed long, then when it gets implicitly
cast to uint64_t as it's masked in with the others, it gets sign-extended
and overwrites payload[10] with all ones. This causes the month to appear
as 15 whenever the day number is odd. Fix it by casting to (long long)
instead. Although switching everything to unsigned times would probably
have worked too.

Also, when calculating the day we find that (0b11111UL << 31) is
truncated to 32 bits, giving 0x80000000 not 0xf80000000. Thus the
top bits of the day are lost. Fix all the mask-and-shift to shift
*first* then mask, which is simpler and doesn't suffer the same
problem.


gizmocuz:
Thanks!!

Issue #3975, opened 2020-03-09
dzVents stops running, HTTPS server stops responding.


dwmw2:
I'm using the dzVents script I showed in https://github.com/domoticz/domoticz/pull/3961 to poll a REST API for the cat flap every minute. 

Every few days, it stops running. And Domoticz stops responding on its HTTP sockets too.

This didn't happen before the cat flap script was added.

Here's an example, starting from the last successful catflap poll at 10:37. At 10:38 it makes the request but the httpResponse handler is never invoked (did the request fail?). There's nothing at 10:39.
<details>
  <summary>Click to show log </summary>

```
Mon Mar  9 10:37:00 2020 daemon.info domoticz: SQLH UpdateValueInt Gas Temperature (right) HwID:2  DevID:82439 Type:80  sType:5 nValue:0 sValue:6.63
Mon Mar  9 10:37:00 2020 daemon.notice domoticz: dzVents: Info:  ------ Start external script: catflap.lua:, trigger: every minute
Mon Mar  9 10:37:00 2020 daemon.notice domoticz: dzVents: Info:  Cat flap poll
Mon Mar  9 10:37:00 2020 daemon.notice domoticz: dzVents: Info:  ------ Finished catflap.lua
Mon Mar  9 10:37:00 2020 daemon.info domoticz: SQLH AddTask: Request to add task: idx=7600494473754144256, DelayTime=0.000000, Command='', Level=1994070744, Color='{INVALID}', RelatedEvent='!#_trigger: Poll!#method: GET!#URL: https://app.api.surehub.io/api/device/422826/tag!#Authorization: Bearer ...token...'
Mon Mar  9 10:37:00 2020 daemon.notice domoticz: EventSystem: Script event triggered: /usr/share/domoticz/dzVents/runtime/dzVents.lua
Mon Mar  9 10:37:00 2020 daemon.info domoticz: SQLH: Do Task ItemType:6 Cmd: Value:https://app.api.surehub.io/api/device/422826/tag
Mon Mar  9 10:37:02 2020 daemon.notice domoticz: dzVents: Info:  Handling httpResponse-events for: "Poll
Mon Mar  9 10:37:02 2020 daemon.notice domoticz: dzVents: Info:  ------ Start external script: catflap.lua: HTTPResponse: "Poll"
[...]

PR #4018, opened 2020-03-27
Increase camera fetch timeout to 10s

dwmw2:
I use a CGI script to fetch a snapshot from an RTSP stream:

```
 #!/bin/sh
 cat <<EOF
 Content-Type: image/jpeg

 EOF
 ffmpeg -skip_frame nokey -i rtsp://${CAMERA}:554/ch01.264 \
      -qscale:v 4 -frames:v 1 -f image2pipe -
```
It takes a little over 5 seconds, every time. I had to bump the
timeout up to 10s. This could be configurable, maybe, and perhaps
we should even teach Domoticz to spawn ffmpeg for itself for an
RTSP camera URI. But a ten-second timeout, and doing the image
processing on a more powerful box, works well in the short term.

### DIFF
--- a/main/Camera.cpp
+++ b/main/Camera.cpp
@@ -263,7 +263,7 @@ bool CCameraHandler::TakeSnapshot(const uint64_t CamID, std::vector<unsigned cha
 		return TakeUVCSnapshot(pCamera->Username, camimage);
 
 	std::vector<std::string> ExtraHeaders;
-	return HTTPClient::GETBinary(szURL, ExtraHeaders, camimage, 5);
+	return HTTPClient::GETBinary(szURL, ExtraHeaders, camimage, 10);
 }
 
 std::string WrapBase64(const std::string &szSource, const size_t lsize = 72)


gizmocuz:
Thank you for the PR, would this not slow it down ? 5 seconds to get a image is quite long, maybe for the first 1 or 2 images, but then it should be ok ?
It would be great if we could support RTSP native 
Most camera's have a snapshot url ... is that not fast ? Does it only have RTSP ?

dwmw2:
Note that the default timeout for HTTP requests (see httpclient/HTTPClient.cpp) is 90 seconds. And yes, that's quite a long time given that we run curl synchronously in the SQLHelper thread. We should use the `curl_multi` API and run it asynchronously so that we aren't locking up the Domoticz thread for so long at a time. I'm going to look at that.

But *this* change from 5 to 10 seconds doesn't make a lot of difference there, when the default is 90s anyway.

It takes that time for each fetch, FWIW. It's got to wait for a "key frame" in the video stream, which is a standalone image in its own right rather than just a delta from the previous. My camera appears not to have a snapshot facility.

gizmocuz:
The default here was 5 not 90
What is your camera brand/model ?

dwmw2:
The default for an all other curl fetches is 90. If I removed that final `, 5)` argument and let it use the default, it'd be 90.

The camera is a crappy Windows/ActiveX only thing but thankfully does have RTSP support that works.

gizmocuz:
@dwmw2, there is a connection timeout, and a download timeout
In your case we are still talking about connection timeout I suppose.
Anyway this PR will increase it to 10 seconds, which is NOT ok if you are testing/adding your camera and enter a wrong IP address (for example), or when the camera is offline
I am also not sure if we should do this patch because of your crappy camera ;) ... (no offense !)
You can buy a decent/good camera on Aliexpress for < $20

dwmw2:
There's a connection timeout, and an overall timeout for the request. The fiormer is 90 seconds, but the latter is 10 seconds by default so we'd never actually wait 90 seconds. Using s shorter *connect* timeout would make sense,, but 10 seconds is the default overall timeout we use for all other fetches including anything from dzVents; I'm not sure I understand why it would be such a problem to use that for cameras too.

I'm more concerned by the fact that even after switching curl to the threaded resolver, I'm still seeing the lockups reported in #3975. My debugging in the HTTPClient code get this far...
```
Mar 28 03:39:01 wndr3800.infradead.org domoticz: HTTP GET: curl init...
Mar 28 03:39:01 wndr3800.infradead.org domoticz: HTTP: curl global timeout 10, connection timeout 90
Mar 28 03:39:01 wndr3800.infradead.org domoticz: HTTP GET: curl perform...
```

... and never to the 'done' that it would print after `curl_easy_perform()` returns.

I think we need to fix that to use curl_multi and not to lock up the thread for the whole length of the HTTP operation *anyway*, regardless of trivial changes to timeouts. I'll look at doing that.

gizmocuz:
The reason is because it is a camera, we would like to have quick updates or a quick stop if it can not be reached
I have read your comment/ideas about curl_multi, it would be great, but scripts are like a state machine, everything has to be executed in order, If we request a HTTP GET, we have to wait until it is done, and continue for maybe the next HTTP GET
But that is of course not related to this issue...
I would suggest to have a browse on Aliexpress for a nice cheap camera if you do not mind ?

ghost (commented):
delete

gizmocuz:
Closing for now, maybe this would work better with the curl_multi method

PR #4030, opened 2020-03-30
Cleanup Xiaomi Gateway code and prepare for 2 new devices

Huupke:
Cleaned up Xiaomi Gateway implementation, mainly be moving strings to reusable definitions in a new separate header file to make it more readable and less error prone. This should also enable a more easy way to add support for new devices.


gizmocuz:
Thank you!!

PR #4034, opened 2020-04-01
Fix the time zone offset bug in the Temperature panel on the web interface.

cocooma:
The bug manifests when the overwrite temperature expire time is being set for 
  the heating zone/radiator TVR.
  In this case I used Honeywell Evohome hardware with HGI80.
  The previously the toISOString() function was used for converting time format
  which does not take into account the Time Zone.
  
  As a consequences the overwrite expire time on the Web interface was set with
  the correct local time zone but the api request sent to the backend was sent with
  the incorrect UTC time.
  
  I've implemented a new function called ConvertTimeWithTimeZoneOffset() which
  takes into account the Time Zone when converts the time format.
  ex.: from "Wed Apr 01 2020 17:00:00 GMT+0100 (British Summer Time)" to "2020-04-01T17:00:00.000Z".
  As well as moved up the time conversion inside the code base so it is taking place earlier
  accommodating both the backend and the frontend.

### DIFF
--- a/www/app/TemperatureController.js
+++ b/www/app/TemperatureController.js
@@ -189,6 +189,14 @@ define(['app', 'livesocket'], function (app) {
 			});
 		};
 
+		// Convert time format taking account the time zone offset. Improved version of toISOString() function.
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thank you !!

Issue #4036, opened 2020-04-01
Device delete does not propagate to fibarolink, httplink and pushlink


rwaaren:
Triggered by this  [forum post ](https://www.domoticz.com/forum/viewtopic.php?f=6&t=31983#p241859) I looked in SQLHelper.cpp and did not found the expected delete statements in CSQLHelper::DeleteDevices.  I then tested a device in httplink and it remained in the database after deletion from the device tab.
I will create a PR to add the delete statements.

PR #4037, opened 2020-04-01
Propagate device deletes to *Link tables

rwaaren:
Add delete device from 

PushLink
HttpLink
FibaroLink
GooglePubSubLink

tables when device is deleted.


rwaaren:
Fix for #4036

gizmocuz:
Thank you for this ! Maybe it is a idea to migrate them all to one database in the feature
They seem all the same except the PushLink as an additional 'PushType' field which we could set to 0 as default
No idea why it is now in different tables ... it has been so since quite some time
We could use the 'PushType' for the type of pusher, like Fibaro, Influx etc..
Currently it is always set to 1
Let me create a quick PR for this and maybe you could share your idea about this...

rwaaren:
If you are going to merge these tables into one new, could you please also change the field deviceID to deviceRowID to make it consistent with all other tables.
(That is; if there is no specific reason to have a different name in the *Link tables)

gizmocuz:
Pff that was not an easy patch... verified all database table migrations .. Should be OK

gizmocuz:
https://github.com/domoticz/domoticz/pull/4042

gizmocuz:
Thanks for the PR, closing it because it's inside PR 4042

PR #4046, opened 2020-04-03
Fix LUA detection on FreeBSD as external library

xbeaudouin:
Fix the detection of LUA on FreeBSD using find_package() to find right paths on FreeBSD.

Kind regards


gizmocuz:
Thanks for the PR, but why does 'find_library' not work as it should ?

xbeaudouin:
Well find_library() does not work on FreeBSD because libraries are called liblua-5.3 instead of liblua5.3 (same for directories : lua-5.3 instead of lua5.3).
Since find_package(Lua "5.3" REQUIRED) is low level CMake why not using it as well ?

gizmocuz:
@xbeaudouin , it is very important we link to the static library, if you know a way to use find_package/required that would be great!

xbeaudouin:
I think we have to fix some stuff on the CMakefiles... Will close this PR and make a better one with find_package if possible for numerous parts.
Because you are right we have to fix that, and also I have to fix the compilation under FreeBSD as well that blown now :) (Thanks for the external stuff by the way).

xbeaudouin:
Closing.: )

gizmocuz:
Yep, but all external libraries are removed from the Domoticz code which is good !
Looking at the FindLua.cmake i can't see a way that it finds a static lib, I do see some static reference but not sure if we can use it

PR #4058, opened 2020-04-06
Refactor CMakeList.txt

xbeaudouin:
This hard work has been done a Adrian who is a CMake Wizard that fixed lots of the CMakeLists.txt on some FreeBSD ports... and ... know few strange things about CMakefiles and... pkg config stuff.

This should help a lot the project to be compiled everywhere.


gizmocuz:
@xbeaudouin , @adriaandegroot, Thank you for refactoring the CMake project file !
It looks very clear now ! It got quite messy after 6 years...

@xbeaudouin , could you display me the output of:

message(STATUS "OpenZWave includes found at: ${OPENZWAVE_INCLUDE_DIRS}/openzwave)

and a

'ls' of that folder ?

On every system I tested the first is displaying

/usr/local/include/openzwave/

and a ls:

> rob@ubuntuhyperv:~/dev-domoticz$ ls /usr/local/include/openzwave/
> Bitfield.h             DoxygenMain.h  Localization.h            Node.h                 Options.h                  Utils.h                  aes
> CompatOptionManager.h  Driver.h       Manager.h                 Notification.h         Scene.h                    ValueIDIndexes.h         command_classes
> DNSThread.h            Group.h        ManufacturerSpecificDB.h  NotificationCCTypes.h  SensorMultiLevelCCTypes.h  ValueIDIndexesDefines.h  platform
> Defs.h                 Http.h         Msg.h                     OZWException.h         TimerThread.h              ZWSecurity.h             value_classes
> 

Are you compiling with OpenZWave support ?

The reason I ask is that '/usr/local/include/openzwave/value_classes/ValueID.h:' is including 'ValueIDIndexes.h' that sits inside the 'openzwave' folder, and that is why it needs to be added to the include paths
A fix would be for this file to include instead '../ValueIDIndexes.h' or '<openzwave/ValueIDIndexes.h>'

As it is possible to use USE_BUILTIN_MQTT and USE_BUILTIN_SQLITE (and disable minizip/jsoncpp), would it be a good idea to also check for these git submodule folders ?

xbeaudouin:
Well about OZW, the CMake find it at : /usr/local/include/openzwave/ (variable ${OPENZWAVE_INCLUDE_DIRS}), 

```
-- Checking for module 'libopenzwave'
--   Found libopenzwave, version 1.6
-- ==== OpenZWave package found!
-- OpenZWavelibraryfoundat:/usr/local/lib/libopenzwave.so
-- OpenZWave library found at: /usr/local/lib/libopenzwave.so
-- OpenZWave includes found at: /usr/local/include/openzwave/

```

The directory here :  
```

[kiwi@portbuild ~]$ ls /usr/local/include/openzwave/
Bitfield.h                      NotificationCCTypes.h
CompatOptionManager.h           OZWException.h
DNSThread.h                     Options.h
Defs.h                          Scene.h
DoxygenMain.h                   SensorMultiLevelCCTypes.h
Driver.h                        TimerThread.h
Group.h                         Utils.h
Http.h                          ValueIDIndexes.h
Localization.h                  ValueIDIndexesDefines.h
Manager.h                       ZWSecurity.h
ManufacturerSpecificDB.h        aes
Msg.h                           command_classes
Node.h                          platform
Notification.h                  value_classes

```

if you look at /usr/local/include/openzwave//openzwave (eg: ${OPENZWAVE_INCLUDE_DIRS}/openzwave)

```
ls /usr/local/include/openzwave//openzwave
ls: /usr/local/include/openzwave//openzwave: No such file or directory
```

And yes I compile with OpenZWave support see the Cmake command : (on my testflight) : 
```

cmake -DCMAKE_BUILD_TYPE=Release -DUSE_BUILTIN_SQLITE="NO" \
                -DUSE_LUA_STATIC="NO" \
                -DUSE_STATIC_BOOST="NO" \
                -DUSE_STATIC_OPENZWAVE="NO" \
                -DUSE_OPENSSL_STATIC="NO" \
                -DUSE_PYTHON="YES" \
                -DUSE_BUILTIN_MQTT="NO"  \
                -DUSE_BUILTIN_JSONCPP="NO" \
                -DUSE_BUILTIN_MINIZIP="NO" \
                -DUSE_PRECOMPILED_HEADER="NO" \
                -DGIT_SUBMODULE="NO" 

```

About USE_BUILTIN_MQTT and USE_BUILTIN_SQLITE, I think this is possible as well...

gizmocuz:
@xbeaudouin , thanks for the addition and the log.

That is so strange, because we ask "where can we find **openzwave/Manager.h**" and not "where can we find **Manager.h**"
It should report back "/usr/local/include" as the full path is /usr/local/include/**openzwave/Manager.h**

I just hope this is only different for FreeBSD ...

dwmw2:
Hm, the change in commit 0bf7fdb5b03554bd8d5fcedbfa25f2e919460c43 seems to be needed for systems other than FreeBSD. My builds on native Fedora Linux and in OpenWRT are breaking because we add `/usr/include/openzwave/openwave/` to the include path, while that doesn't exist (everything's under `/usr/include/openzwave/`.

When using pkg-config, surely we should be using the include path that pkg-config gives us without mangling it in any way? Which systems *do* need this? Can we fix them, and if all else fails lets make *those* the special case instead?

I think our own source code should probably be including `Manager.h` not `openzwave/Manager.h` since that file is directly in the include directory that libopenzwave.pc adds to our include path.

dwmw2:
See #4206 which should fix this for all cases.

gizmocuz:
Imagine we have a header file "Manager.h" in our code ?
That is why it should be openzwave/Manager.h

gizmocuz:
When you issue 'sudo make install' in the openzwave library those header files are inside the openzwave folder
PR fails to build
It's the same for minizip, it's included as 

#include <minizip/unzip.h>

And it's quite common to do so

dwmw2:
> Imagine we have a header file "Manager.h" in our code ?

That would work about as well as having a header file named "stdio.h".

We use these libraries; we need to avoid making header files which conflict with them — and functions and classes which conflict with them, in our own code.

The OpenZWave pkg-config file is its contract with us. It might install its header files anywhere. Perhaps `/usr/include/openzwave`, perhaps `/usr/local/include/openzwave` perhaps `/opt/OpenZWave/include`. What it promises is that the correct path to add to the compiler search path is the one in the pkg-config file. 

The OpenZWave header files include each other without the `openzwave/` prefix. When we fail to add the correct path to the compiler, it already breaks. And if we added our own header files that conflicted, like `stdio.h`, that would already break.

Much better to follow the programming contract and make things work reliably by using the header files as intended.

dwmw2:
This also allows us to use the headers as-is from the `open-zwave-read-only` directory instead of requiring root access and `sudo make install`

gizmocuz:
This will only work when openzwave can be found in our build folder (or known location), that dependency (and also the inclusion of other external libraries in the domoticz repo) has been removed.
Indeed the includes and library can be installed anywhere in the system and CMake should search for it's location
Same as minizip

if CMake find the headers in

/my/opt/include/openzwave

then it should add /my/opt/include/ to the includes

then we can use <openzwave/someheader.h> again

dwmw2:
Cmake will use the path explicitly provided to it by the pkg-config file, as $DEITY intended. OpenZWave doesn't build unless the correct directory is on the compiler's include path because it includes its *own* header files as e.g. `ValueIDIndexes.h` and the build fails — for all systems I tested on, not just FreeBSD which got a partial fix.

dwmw2:
Minizip has similar problems, FWIW. Upstream minizip installs its headers to `${prefix}/include` not `${prefix}/include/minizip`

It's just that some distributions force it to go to subdir in their packaging.

But because the contract is "you use the directory that pkg-config says", it should all work out fine. Except for Domoticz, which doesn't do it correctly. I'll follow up with a patch for that one too, although I had worked around it in OpenWRT by putting the minizip files in a subdir anyway, just to reduce pollution.

dwmw2:
> This will only work when openzwave can be found in our build folder (or known location), that dependency (and also the inclusion of other external libraries in the domoticz repo) has been removed.

Hm? The support for `../open-zwave-read-only` is still there, and it looks like it's still used in the Travis build. I tested that in *addition* to the proper system installation, but the reason for #4206 is because Domoticz doesn't build with the system installation. It does now, after fixing the include paths.

I tested both configurations.

> if CMake find the headers in
> 
> /my/opt/include/openzwave
> 
> then it should add /my/opt/include/ to the includes
> 
> then we can use <openzwave/someheader.h> again

No, that doesn't work. Before my patch in #4206: 
```
$ cmake . -DBUILD_SHARED_LIBS=yes -DBoost_NO_BOOST_CMAKE=yes -DCMAKE_BUILD_TYPE=Release -DUSE_BUILTIN_MQTT=no -DUSE_BUILTIN_SQLITE=no -DUSE_BUILTIN_JSONCPP=no -DUSE_BUILTIN_MINIZIP=yes -DUSE_STATIC_BOOST=no -DUSE_STATIC_LIBSTDCXX=no -DUSE_STATIC_OPENZWAVE=no -DUSE_OPENSSL_STATIC=no -DUSE_PYTHON=yes -DWITH_LIBUSB=no
…
-- Checking for module 'libopenzwave'
--   Found libopenzwave, version 1.6
-- ==== OpenZWave package found!
-- OpenZWavelibraryfoundat:/usr/lib64/libopenzwave.so
-- OpenZWave library found at: /usr/lib64/libopenzwave.so
-- OpenZWave includes found at: /usr/include/openzwave/
…
```
```
$ make
…
In file included from /usr/include/openzwave/Driver.h:37,
                 from /home/dwmw2/git/domoticz/main/../hardware/openzwave/control_panel/ozwcp.h:41,
                 from /home/dwmw2/git/domoticz/main/../hardware/OpenZWave.h:10,
                 from /home/dwmw2/git/domoticz/main/mainworker.cpp:39:
/usr/include/openzwave/value_classes/ValueID.h:34:10: fatal error: ValueIDIndexes.h: No such file or directory
   34 | #include "ValueIDIndexes.h"
      |          ^~~~~~~~~~~~~~~~~~
compilation terminated.
```

Because OpenZWave expects its header directory (in this case `/usr/include/openzwave`) to be on the compiler include path, and it includes its own header files directly by their name, which is the API that OpenZWave has defined for its header files.

If you look at the OpenZWave documentation, for example at http://www.openzwave.com/dev/classOpenZWave_1_1Manager.html you see it explicitly tells you which header files to include. In that case `#include <Manager.h>`.

If the distribution has installed OpenZWave such that the last element of the directory path is actually `/openzwave/` and we happen to have the parent directory on the compiler's include path, then *our* code might get lucky and be able to violate the OpenZWave defined API by including `openzwave/Manager.h` for example.

But we *do* still have to put the correct directory into the compiler search path anyway because OpenZWave uses its own header files correctly according to its own documentation and defined API. Which in turn means we *can't* create our own files called (e.g.) `Manager.h` anyway.

I do agree that it would be nicer if they put their headers into a subdirectory of the compiler include path that they define. GLib does that, for example:
```
$ pkg-config --cflags glib-2.0
-I/usr/include/glib-2.0 -I/usr/lib64/glib-2.0/include 
 ls -l /usr/include/glib-2.0 /usr/lib64/glib-2.0/include
/usr/include/glib-2.0:
total 36
drwxr-xr-x. 2 root root 4096 May 24 20:23 gio
drwxr-xr-x. 3 root root 4096 May 24 20:23 glib
-rw-r--r--. 1 root root 3381 May 20 11:46 glib.h
-rw-r--r--. 1 root root 1462 May 20 11:46 glib-object.h
-rw-r--r--. 1 root root 4461 May 20 11:46 glib-unix.h
-rw-r--r--. 1 root root 4318 May 20 11:46 gmodule.h
drwxr-xr-x. 2 root root 4096 May 24 20:23 gobject

/usr/lib64/glib-2.0/include:
total 8
-rw-r--r--. 1 root root 5678 May 20 12:18 glibconfig.h
```

Although if you look at e.g. https://developer.gnome.org/glib/stable/glib-Memory-Allocation.html#glib-Memory-Allocation.includes you'll note that they do insist that you include `<glib.h>`. It's just that all the other "internal" headers included from that are in the subdirectory avoiding pollution.

But really, we just need to follow the API defined by the library. If we want to talk to OpenZWave about polluting the include path a bit less by using a subdirectory, we could do that. But the API is what it is, right now. And our build is broken because we're not using it correctly.

dwmw2:
Ah, I've spotted the issue with the `../open-zwave-read-only` build in the PR; it isn't actually `../open-zwave-read-only` at all; it's `./open-zwave-read-only` because it's overridden on the command line by setting `-DCMAKE_LIBRARY_PATH=open-zwave-read-only` and the hints in `CMakeLists.txt` are redundant. Fixed by https://github.com/domoticz/domoticz/pull/4206/commits/63d825d9afc90d95a9153845776e94a46a92a171 it to find the includes for the static build in the same place as the library itself was found, instead of only looking in `../open-zwave-read-only`.

gizmocuz:
Travis is a build system that has it's own compile script

On my build systems I don't have a openzwave readonly folder in my domoticz source folder, or the folder before that
The library is installed (make && make install)

Do note that 'The API is what it is' , is that you need to compile and install the library
After that, third party applications have to search for it, and use it's path

The current PR is just reverting to the old build setup, it would be great if CMake could 'find' the library anywhere on the system and use it.
Try doing a make install, and remove the openzwave build folder, and then compile Domoticz

dwmw2:
For the installed library, CMake does "find" it through the proper channels — asking pkg-config where the include path is, and doing what it's told. That's what my PR fixes.

I'm just working out the details for the non-standard installation now. I fixed the static open-zwave-read-only build and now looking at making the Windows build work too when we follow the OpenZWave API as it was intended.

dwmw2:
.... and it looks like the latest CI builds in the PR are now working, having completed while I was on the phone...

gizmocuz:
Still it feels like going back to the old build system...
I would have likes a PR for OpenZWave more... (Everything nicely via <openzwave/...>

Issue #4076, opened 2020-04-13
Wrong wind speed and wind gusts with WUnderground


mantoz:
Hi,
  I'm experiencing wrong wind speed (speed and gusts) when getting data from Wunderground.
After a short investigation, it seems that Domoticz in Wunderground.cpp thinks that Wunderground is returning data in m/s, while it's in km/h.
I think the issue arose when Wunderground switched to API v2. 

Using 'metric' (... &units=m ...) in the query string now returns wind speed (and gusts) in km/h as specified at [https://ibm.co/APICom](https://ibm.co/APICom) and [https://ibm.co/v2PWSCC](https://ibm.co/v2PWSCC), so that the conversion in Domoticz is wrong.

To get data in m/s, you now have to specify 's' (... &units=s ...) in order to get metric_si.
Of course also the returned JSON string changes from 'metric' to 'metric_si'.
According to that documentation, also 'distance' changes from km to m, but I don't think it's used anywhere in the code. Others units remain the same.

A fixed Wunderground.cpp is in the attachment (not tested, yet!), with the query changed to 'units=s' and the modified "metric" to "metric_si" strings in the appropriate JSON sections.

Thanks,
mantoz

[Wunderground.cpp.txt](https://github.com/domoticz/domoticz/files/4471555/Wunderground.cpp.txt)

Edit reason: forgot attachment :)

gizmocuz:
Thank you for your finding. Could you make a pull request with the changes ?
Please test it first...
To compile Domoticz you can follow this guide:
https://www.domoticz.com/wiki/Installing_and_running_Domoticz_on_a_Raspberry_PI

mantoz:
I tested the fix this afternoon and I had no problems, so I made the pull request #4079

gizmocuz:
Thanks !! PR applied !

PR #4079, opened 2020-04-14
Fix for WUnderground service

mantoz:
According to WUnderground API v2 documentation, in order to get wind speed (and gusts) data in m/s the query string needs to be 's', not 'm' (... &units=s ...). In addition, JSON data changes from 'metric' to 'metric_si'.

This fixes issue #4076


gizmocuz:
Thank you for this !!

Issue #4083, opened 2020-04-16
Buienradar


rwaaren:
Starting april 16th around midnight, Buienradar does return 0,0 for lat,lon location of the weatherstation when called using [the URL used in Buienradar.cpp](https://data.buienradar.nl/2.0/feed/json) 

See this [forum topic](https://www.domoticz.com/forum/viewtopic.php?f=6&t=32214) for more details.

It's reported to buienradar but until they respond to that we don't know if this is a temporary issue on there side or a changed policy that will not be changed back.
If the latter we need to change the way we get the weatherstation closest to the calling systems location. Hopefully it is not needed but if so below table might be of some use. Stations on this list are verified and operational except Hoofdplaat which seems to be down
 
```
Station	lat,	lon	location
===============================
6215	52.12,04.43 Voorschoten
6225	52.47,04.57 IJmuiden
6229	53.00,04.75 Texelhors
6235	52.92,04.78 Den Helder
6239	54.85,04.73 Zeeplatform F-3"
6240	52.30,04.77 Schiphol
6242	53.25,04.92 Vlieland
6248	52.63,05.17 Wijdenes
6249	52.65,04.98 Berkhout
6251	53.38,05.35 Hoorn Terschelling
6252	53.22,03.22 Zeeplatform K13
6257	52.50,04.60 Wijk aan Zee
6258	52.65,05.40 Houtribdijk
6260	52.10,05.18 De Bilt
6267	52.88,05.38 Stavoren
6269	52.45,05.53 Lelystad
6270	53.22,05.77 Leeuwarden
6273	52.70,05.88 Marknesse
6275	52.07,05.88 Arnhem
6277	53.42,06.20 Lauwersoog
6278	52.43,06.27 Heino
6279	52.73,06.52 Hoogeveen
6280	53.13,06.58 Groningen
6283	52.07,06.65 Groenlo-Hupsel
6285	53.57,06.40 Huibertgat
6286	53.20,07.15 Beerta
6290	52.27,06.90 Twente
6308	51.38,03.38 Cadzand
6310	51.45,03.60 Vlissingen
6311	51.37,03.66 Hoofdplaat     -- Down
6312	51.46,03.62 Oosterschelde
6313	51.50,03.25 Vlakte aan de Raa
6315	51.45,04.00 Hansweert
6316	51.65,03.70 Schaar
6319	51.23,03.83 Westdorpe
6320	51.93,03.67 LE Goeree
6321	52.00,03.28 Euro platform
[...]

PR #4084, opened 2020-04-17
Add Configuration Parameter for StationID to BuienRadar

akamming:
This PR allows a custom station ID for the BuienRadar hardware. If the parameter is 0, the Buienradar hardware will have the same behaviour as without this PR. But if it is filled you can get weather data for a different station.

this allows to have multiple stations in your domoticz instance. (e.g. if you want to monitor the weather of different sites)

It is also a workaround for [this issue](https://www.domoticz.com/forum/viewtopic.php?f=6&t=32214) which is now solved, but might happen again.

Edit:
I just  found out that the rain prediction is not derived from the station id, but from the long/lattitude coordinates in the domoticz settings. So another config setting should be added here as well. I propose to merge this PR 1st. 

And i will create another one for the long/lattitude config param (i am thinking of something like that if long/lattitude is provided as config param that will be used, if not, depending on config param station id. If provided, then get the long/latt from buienradar. If  not, use long/latt from domoticz config).


akamming:
Since not yet merged i just added another commit to the PR which retrieves the latitude and longitude from Buienradar if a custom id is provided in the config

akamming:
and since this one is not yet merged, i now also included the last changes i wanted to make in also in this PR. The location is now also customizable. You can now ignore the update remark, because that's fixed in this same PR now.

gizmocuz:
Is there somewhere a page where the Station ID's can be found ?

akamming:
@rwaaren made a list here: https://github.com/domoticz/domoticz/issues/4083. I don't know what is the origin

And there is also a list on this location (with slightly  different longitude/latitude) : http://projects.knmi.nl/datacentrum/catalogus/catalogus/content/nl-obs-surf-stationslijst.htm

gizmocuz:
You can get it by calling

https://observations.buienradar.nl/1.0/actual/weatherstation/all

But there will be questions like 'what is this station ID, how do i know my nearest station id, how to convert the output of the above to readable format etc...'

I am thinking about added a 'Help' getter function for each Hardware class that can be displayed when you select a hardware type in domoticz (and clicking on a blue questionmark for example)

akamming:
That's a good idea. I would not let it block the merge of this item. In hardware classes like weather underground and darksky you can also configure stationid or location without any help function. 

Basically this commit brings buienradar on par with these hardware classes for this capability. 

And on the forum there are users requesting this capability.

rwaaren:
@akamming 
I created this list by visiting the individual links of these weatherstations

gizmocuz:
That's correct. Going to make some small changes to be more compatible with WU and coding style...
Will be merged soon!

gizmocuz:
Thanks for the PR!

I merged it with some small adjustments (tested of course):

Using same Gui text as in WU, that saves us new translation strings
Merged Station ID/Lat,Long in one field (same as WU)
Removed extra log lines (I always use breakpoints and use my mouse to hover over variables, saved these debug lines)
Changed to internal log function (we should do that for all hardware one day)
Small code style adjustments

akamming:
Nice!

akamming:
@gizmocuz  i just checked. looks good!  

although I do have have 1 finding. although the combining of stationid and location into 1 config makes is more consistent with other weather services. I coded it to be able to configure both, so you can also use it as a workaround for the issues with buienradar that it does not always provide the longitude/lattitude (giving errors on the rain prediction).   If you can configure both, you're not dependent of Buienradar to provide this info anymore.  

So i think splitting of the fields is better for Buienradar.  What do you think?

gizmocuz:
I think we are fine as it is now. The rain prediction is only depended on a coordinate.
If a station does not report it, don't use it... And what happened earlier was a bug, should not be the default....
It there will be a protocol change we can always split it

akamming:
ok clear...

PR #4095, opened 2020-04-23
Null pointer in MySensorBase.cpp

bmellink:
This should be the good one


gizmocuz:
Thank you for the PR, but I think in the latest code in the repository it is OK now ?

gizmocuz:
https://github.com/domoticz/domoticz/blob/4d558c1cf90a1425a4e11826fa9240db5bcf5b39/hardware/MySensorsBase.cpp#L994

PR #4100, opened 2020-04-25
TTNMQTT: GPS data was not stored (now PercentageSensor and UserVariable)

kiddigital:
This PR fixes a small omission that was introduced in the previous change. GPS was not stored anymore. Now it is stored again in the same way as before.

### DIFF
--- a/hardware/TTNMQTT.cpp
+++ b/hardware/TTNMQTT.cpp
@@ -515,6 +515,11 @@ void CTTNMQTT::on_message(const struct mosquitto_message *message)
 							bBaro = true;
 						}
 						else if (type == "gps") {
+							std::stringstream sstr;
+							sstr << vSensor["lat"].asFloat() << "," << vSensor["lon"].asFloat() << "," << vSensor["alt"].asFloat();
+
+							SendPercentageSensor(DeviceID, channel, BatteryLevel, vSensor["alt"].asFloat(), DeviceName + " Altitude");
+							UpdateUserVariable(DeviceName, sstr.str());
 							bGps = true;
 						}
 						else if (type == "digital_input") {


gizmocuz:
Thank you !!

jeroenkl:
Yes indeed a percentage sensor is seen in domoticz, but only for altitude.
Lora sensor data:
{
  "analog_in_8": 3.8,
  "gps_1": {
    "altitude": 0.1,
    "latitude": 52.xxxx,
    "longitude": 5.1xxx
  }
}

In domoticz received:

Idx | Laatste Aanpassing | Naam | Status | Waarde
1257 | 2020-02-07 07:04:36 | rak7200_xxxx Altitude | 2.30 | 0/2.30

Tested with 3 sensors:
2x RAK7200 and 1 TTGO-beam,  behavior is exactly the same.
- Alt is seen as percentage and incorrect number
- Lon and Lat is not seen

BR,

Jeroen

jeroenkl:
If it helps I'm happy to provide for investigations purposes,  a test TTN key and app name.

kiddigital:
Jeroen, all the GPS data is also stored in a UserVariable (see Settings->More Options->UserVariables). I agree that it is not optimal.

Thanks for the offer to help testing :) I have some sensors here as well so no direct need. But thx.

What would be a better ‘device’ in Domoticz for the GPS data? I don’t know what devicetype would be better suited. Maybe there is a need now for a new devicetype GPS? How/where should it be shown ? Suggestions?

jeroenkl:
Hi kiddigital,

Great to see the GPS lan/lon/alt as UderVariables. I'm using it to calculate the distance (line of sight) to my home and trigger a notification when heading 1km distance.

- Is it possible to change the 'altitude sensor' as a custum sensor like you did with the 'analogue_input'? As a user we can change it to our previlege, like a batt sensor in V or 'hight' in meters, instead of currently a 'percentage'.

- GPS data, lan/lon, difficult to determine how to show it in domoticz. Best think is to have it seen as a 'location sensor', showing the data and when opening a 'location map?' 

great help, thanks! And if you need me, just give me a 'shout'.

BR,

Jeroen

kiddigital:
@jeroenkl , I have been looking into your question and tinkering with adding the 'concept of Location' to Domoticz as there isn't any thing like that yet (as far as I could find).

Already I made some changes to introduce a new devicetype for Location (Lon, Lat, Alt but also timestamp, horizontal- and vertical accuracy, bearing and speed for more advanced use of the location concept) and some initial processing for it.

Once [PR 1403](https://github.com/domoticz/domoticz/pull/4103) has been approved and merged, I will see if I can push a beta.

The UI part isn't easy :P but I was even considering that 'Location' devices should get their own TAB in Domoticz as they are a class of their own and probably need other ways of representing them on screen (mini map or combined, etc.). But I could not find how to add a TAB yet (@gizmocuz , any pointers? If you consider it a good idea at least.)

So instead of changing it from a Percentage Sensor into a Custom Sensor, it becomes a 'Location'  Sensor :)

For now it assumes Digital Degrees (DD) as input and stores it that way, but I am considering adding a DD to DMS conversion function so you can display Degrees, Minutes and Seconds.

Also, I incorporated 2 subtypes:
1) The actual devices
2) 'Gateways' like the LoRa gateways, but could be any other gateway or 'home' Location
And if possible link the 2 together, meaning an actual devices can have 1 'nearest/last seen' gateway. This opens up the possibility to do some things like trying to calculate the distance between the 2 and/or make GeoFencing around a gateway possible. But I probably can use some help in the actual math behind it all, so if you have any suggestions :) ?

tbc...

kiddigital:
@jeroenkl , see [this PR](https://github.com/domoticz/domoticz/pull/4123) for the Location devicetype as mentioned above.

PR #4103, opened 2020-04-26
TTNMQTT: Improvement on calculating and storing signallevels

kiddigital:
This PR introduces a better processing of the LoRa RSSI and SNR and translates it to a SignalLevel within a range that Domoticz is able to work with. Quite some assumptions are done to guestimate a SignalLevel between 1 and 9.

The used 'CustomSensor' didn't handle the SignalLevel and BatteryLevel correctly, so some under the hood changes were needed to make sure the measured values are correctly passed downstream. These changes could also be beneficial for other hardware and devices.

Could not find any clear documentation on Domoticz SignalLevel, but the working assumption was that it originates from the RFX433 'SignalLevels' and uses somekind of translation instead of more exact measurement units like dBm or dB.

### DIFF
--- a/hardware/Buienradar.cpp
+++ b/hardware/Buienradar.cpp
@@ -395,7 +395,7 @@ void CBuienRadar::GetMeterDetails()
 	int stationID = root["stationid"].asInt();
 	if (stationID != m_iStationID)
 	{
-		Log(LOG_ERROR, "Invalid data received (invalid stationid) or no data returned!", stationID, m_iStationID);
+		Log(LOG_ERROR, "Invalid data received (invalid stationid %i / %i ) or no data returned!", stationID, m_iStationID);
 		return;
 	}
 
--- a/hardware/Buienradar.cpp
+++ b/hardware/Buienradar.cpp
@@ -395,7 +395,7 @@ void CBuienRadar::GetMeterDetails()
 	int stationID = root["stationid"].asInt();
 	if (stationID != m_iStationID)
 	{
-		Log(LOG_ERROR, "Invalid data received (invalid stationid %i / %i ) or no data returned!", stationID, m_iStationID);
+		Log(LOG_ERROR, "Invalid data received (invalid stationid %d / %d ) or no data returned!", stationID, m_iStationID);
 		return;
 	}
 
--- a/hardware/DomoticzHardware.cpp
+++ b/hardware/DomoticzHardware.cpp
@@ -824,12 +824,14 @@ void CDomoticzHardwareBase::SendVisibilitySensor(const int NodeID, const int Chi
 	sDecodeRXMessage(this, (const unsigned char*)& gDevice, defaultname.c_str(), BatteryLevel);
 }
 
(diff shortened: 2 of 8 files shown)


gizmocuz:
@kiddigital , thank you for the PR
It's looking good, but as it also contains changes to internal functions could you:
- Leave the SignalLevel variable name as is

In decode_general, just issue:

uint8_t SignalLevel = pMeter->rssi;

This way it is consistent with all existing code

kiddigital:
@gizmocuz , due to const declaration of these variables in the function declaration, a little different approach has been taken but the usage of SignalLevel is now consistent again :) Hopefully this is satisfactory?

gizmocuz:
@kiddigital , much better thanks, but i like your approach, so i think you can just delete this from the function declaration 'const uint8_t pSignalLevel, const uint8_t pBatteryLevel' and always take this from the structure
Just make sure default it is set to  12/255 in the constructor of the structure like you do now

gizmocuz:
Thanks for all changes, only one request left:

	uint8_t	SignalLevel = 12;
uint8_t	BatteryLevel = 255;


if (pMeter->rssi != 12)
  SignalLevel = pMeter->rssi;
if (pMeter->battery_level != 255)
  BatteryLevel = pMeter->battery_level;

That does not comply... you can remove the whole if then checking because above you set it to 12/255, and if it is 12, you may set it to 12

I would change this to:

uint8_t SignalLevel = pMeter->rssi;
uint8_t BatteryLevel = pMeter->battery_level;

kiddigital:
@gizmocuz , done :)

btw... I thought it would be a good idea to update the Wiki and add some info on the TTNMQTT and LoRa in general, but I don't see a way to update/add to the Wiki?

rwaaren:
> @gizmocuz , done :)
> 
> btw... I thought it would be a good idea to update the Wiki and add some info on the TTNMQTT and LoRa in general, but I don't see a way to update/add to the Wiki?

Please send me (waaren) a PM on the forum with your Email  adres and preferred username and I will create your wiki User ID.
This procedure became necessary after the recent massive spam attacks.
Sorry for the inconvenience.

jeroenkl:
happy to test as soon as it is inplemented into the new beta

jeroenkl:
> Thanks for all changes, only one request left:
> 
> ```
> uint8_t	SignalLevel = 12;
> ```
> 
> uint8_t BatteryLevel = 255;
> 
> if (pMeter->rssi != 12)
> SignalLevel = pMeter->rssi;
> if (pMeter->battery_level != 255)
> BatteryLevel = pMeter->battery_level;
> 
> That does not comply... you can remove the whole if then checking because above you set it to 12/255, and if it is 12, you may set it to 12
> 
> I would change this to:
> 
> uint8_t SignalLevel = pMeter->rssi;
> uint8_t BatteryLevel = pMeter->battery_level;

Hmmm, can't wait to test it but I'll be patient ;-), not in beta yet

gizmocuz:
Thank you very much ! Beta's are building

kiddigital:
@gizmocuz , @jeroenkl you are welcome :)

I have some other changes pending (see my fork) but before I make a PR for those, I like to make sure it works with this PR and other changes that have been added in the meantime.

Thx and till the next PR ;)

Issue #4112, opened 2020-04-27
The value of otherdevices_lastupdate is set BEFORE starting eventSystem scripts in recent versions


rwaaren:
Also influence behavior of lastUpdate and lastLevel attributes of devices in dzVents.

 See forum posts:

[2020.2 version has changed the otherdevices_lastupdate](https://www.domoticz.com/forum/viewtopic.php?f=6&t=32388) and [Help for dimming time script](https://www.domoticz.com/forum/viewtopic.php?p=245291#p245291)

build 11886 shows new behavior
build 11885 shows old behavior

So seem to be a side effect of https://github.com/domoticz/domoticz/commit/1bce2493c

lorenzo-deluca:
I also noticed this change in behavior.  
I have a script with which I manage the double press of the buttons to activate scenarios or dimming light. 
Since the last update it no longer works causing loops so I had to temporarily disable this script.
It would be very useful if it returned to behave as before the last update.

Thanks for your great work!

PR #4115, opened 2020-04-28
Fix for #4112; lastLevel and lastUpdate set after eventSystem trigger

rwaaren:
Was a side effect of https://github.com/domoticz/domoticz/commit/1bce2493c

#4112


gizmocuz:
Thank you !

PR #4117, opened 2020-04-28
Updated eVehicles module

MrHobbes74:
The following new features are added:
- ODO meter
- unlock/open alert
- maximum charge level switch

Also superfluous logging related to checking if car is awake is removed.


gizmocuz:
Thanks!!

PR #4118, opened 2020-04-29
Updated dzVents documentation to correct data store location

joro75:
Updated the indentation of the location where the dzVents data is stored. In the previous version it seemed that the dzVents directory was a sibling of the scripts directory, but in reality it is a child.

### DIFF
--- a/dzVents/documentation/README.md
+++ b/dzVents/documentation/README.md
@@ -2005,17 +2005,17 @@ For every script file that defines persisted variables (using the `data={ … }`
 
 	domoticz/
 		scripts/
- 		dzVents/
-			data/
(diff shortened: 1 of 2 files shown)


rwaaren (approved):
Thanks for the modification. It makes it more clear how the directory for persistent data files is structured.

The README.md is the source for the README.wiki 
Howto create the wiki file is described in the file pandoc.md  (both in in the same directory) 

For this PR I have generated the wiki and added it as a commit to your PR.

If you have more modifications to the documentation / wiki, please combine all of them before creating a new PR as every PR does generate a full compile task for Linux and for Windows and bumps the build number.

PR #4123, opened 2020-05-01
Concept of Locations, first used by TTNMQTT 

kiddigital:
This PR adds a new hardware type Location, ability to store and display Location data and notify on distance between Domoticz and the Location device. 

AND the TTNMQTT modules makes use of these new Location possibilities as the GPS coordinates send by a LoRa device, are now processed and stored using this new hardware type. Which was the first reason for introducing this new Location concept to begin with.

The concept of Location was not available in Domoticz, as far as I could tell at least. The Location hwType can hold information like:
- Latitude, Longitude and Altitude
- Bearing and Speed
- Horizontal- and Vertical Accuracy
- Timestamp
(and also battery_level, rssi, etc. offcourse)

Two subTypes, GPS and Gateway, are available but only GPS is implemented and has all the required functions to store, display and notify on it.

Using the 'earthDistance' calculation from the Buienradar hardware, which is now moved over to the more generic Helper functions, notifications can be configured that fire-off based on the distance between the GPS device location and Domoticz. Make sure that the 'home' locations of Domoticz is configured in the Settings.

Known limitations:
- The Location devices are only shown at the Utility tab and cannot be shown on the dashboard (nicely prevented by disabling the Favorite button)

Although I did quite some testing, still I think this should be considered Beta mainly because a large number of changes are in crucial parts of Domoticz. Yes I am confident that they are done right, otherwise it would be silly to create this PR, but still. So it would be very useful if this gets merged into the Beta so many othes can test this as well.

On the wanted/to-do list:
- I think 'Locations' should get their own tab in the UI where Location specific view can get implemented (someone with UI/UX knowledge and skills willing to do/assist here?)
- At the moment only the GPS subtype is implemented and only support Lat, Lon and Alt. Storing additional measurements like timestamp, velocity, heading, etc. is already foreseen but not implemented
- Implementing the Gateway subType
- Make it possible to link 1 or more GPS subtype devices to a Gateway subtype device and calculate the distance between these two. Instead of only the distance between the GPS subtype and the 'home' location of Domoticz.

Looking forward to the responses and feedback (and less to the Issues, but I will do my best to tackle them asap) :)


notifications/NotificationHelper.cpp
```diff
@@ -447,6 +447,53 @@ bool CNotificationHelper::CheckAndHandleNotification(const uint64_t DevRowIdx, c
...
+				std::string Latitude;
+				std::string Longitude;
+
+				uint64_t nDistance = 0;	// Just for testing, set the distance at 1000 meters
+				double fLat = atof(strarray[0].c_str());
+				double fLon = atof(strarray[1].c_str());
+				double fAlt = atof(strarray[2].c_str());
+
+				// Let's get the 'home' Location of this Domoticz instance from the Preferences
+				if (!m_sql.GetPreferencesVar("Location", nValue, sValue))
```
kiddigital, line 467: @gizmocuz , is there an easier/better way of getting the Location data from Domoticz setting instead of going to the database each time? Aren't these variables in memory somewhere?

jeroenkl:
Great work @kiddigital, again happy to test and letting you know my findings.

jeroenkl:
I see it isn't merged in Version: 2020.2 (build 12012) yet.

gizmocuz:
@jeroenkl , this patch needs more time to check because there is a new device type introduced

@kiddigital , could you not use a Custom Sensor for this ? Specially because the other parameters are not available yet, and you might want to perform actions based on the other parameters, so they might need to end up in another (custom) sensor

kiddigital:
@gizmocuz , I considered that as well, but as custom sensors are 'limited' to 2 ints and 2 floats (and 1 text), I thought it would make more sense to have a dedicated Location type. And yes, I didn't use all variables right away but this way I think the structure is future proof and can be used for all kind of Location usages.

Also for visualization purposes, Locations can get their own specific visuals (or even tab maybe?) and use all available properties (if populated ofcourse).

I see more and more extensions of CayenneLPP that support additional readings for Location usage like bearing, timestamp, etc. These readings could all be independent sensors but I suspect that many future operations are all in relation to each other. For example, based on position, bearing & speed calculation can be done to estimate ETA, etc. Don't know how that would work if the values are 'spreaded' over multiple sensors?

Do you think the implementations is too complicated and risky?

I have some more work coming up for the TTNMQTT module that builds upon this.

And I am looking into some more advanced scenario's/use-cases.

Happy to discuss ideas and open for suggestions and review comments.

Thanks already for taking the time :)

notifications/NotificationHelper.cpp
kiddigital: @gizmocuz , any suggestions for improvement I could try to implemented for the above 'issue'?

jeroenkl:
Need help to test?

kiddigital:
> Need help to test?

Yes please, when this PR is merged into the Beta.

jeroenkl:
> > Need help to test?
> 
> Yes please, when this PR is merged into the Beta.

No problem, curious when it is merged ;-)

jeroenkl:
@gizmocuz , is it possible to have this added into beta? Happy to test as soon as it is available.

Greetz, Jeroen

gizmocuz:
I am not happy with the new sensor type...
I don't think this is needed.
It could be put inside a text sensor, a custom sensor ... or if this is really needed... a new subtype for the pTypeGeneral

jeroenkl:
> I am not happy with the new sensor type...
> I don't think this is needed.
> It could be put inside a text sensor, a custom sensor ... or if this is really needed... a new subtype for the pTypeGeneral

@gizmocuz, ok clear. Is it possible to add it as a custom sensor?
@kiddigital can investigate any other option after this addition.

BR, Jeroen

kiddigital:
@gizmocuz 

> I am not happy with the new sensor type...

Isn’t implemented correctly? Or are Sensor Types used differently?
> 
> I don't think this is needed.

For just Lon, Lat, Alt I agree that it can be done differently/easier. But more advanced for moving objects, geofencing, etc. it is limited.

> 
> It could be put inside a text sensor, a custom sensor ... 

True.. but doesn’t that mean when wanting to use these reading, you have to ‘decode’ them over and over again and hope/check/rely that the sensor values are populated correctly? A sensor (sub)type forces this, doesn’t it?

> or if this is really needed... a new subtype for the pTypeGeneral

Ok, why should one prefer a new subtype of pTypeGeneral over a new dedicated Type? Maybe something I am not seeing or understand?

Thanks anyway for looking at the PR.

@jeroenkl, what is your intended use-case and how would it work if we put the readings into a Custom sensor?

jeroenkl:
> @gizmocuz
> 
> > I am not happy with the new sensor type...
> 
> Isn’t implemented correctly? Or are Sensor Types used differently?
> 
> > I don't think this is needed.
> 
> For just Lon, Lat, Alt I agree that it can be done differently/easier. But more advanced for moving objects, geofencing, etc. it is limited.
> 
> > It could be put inside a text sensor, a custom sensor ...
> 
> True.. but doesn’t that mean when wanting to use these reading, you have to ‘decode’ them over and over again and hope/check/rely that the sensor values are populated correctly? A sensor (sub)type forces this, doesn’t it?
> 
> > or if this is really needed... a new subtype for the pTypeGeneral
> 
> Ok, why should one prefer a new subtype of pTypeGeneral over a new dedicated Type? Maybe something I am not seeing or understand?
> 
> Thanks anyway for looking at the PR.
> 
> @jeroenkl, what is your intended use-case and how would it work if we put the readings into a Custom sensor?

Hi @kiddigital, 

Great you're helping this way, appreciated.
The ultiem final solution is a GPS sensor and to add more features like geofencing.
More and more sensors will be moving arround, so this is beneficial according to me.

But, having already a custom sensor (no text sensor), it is also nicer to see on the Dashboard.
Scripting with use of the custom sensor data is my next step (Geofencing, etc)

BR,

Jeroen

kiddigital:
@jeroenkl  : I am preparing a little different version of this PR with Geofencing :) using a generic device as suggested by @gizmocuz . Probably in a day or two.

gizmocuz:
Thanks for the other PR, closing this one

Issue #3541, opened 2019-07-22
Custom Icon Yeelight Colour Bulb not showing


sincze:
### Issue Guidelines
Domoticz V4.10988

I have 2 types of yeelight. The Mono/WW and RGBW

For the Mono I can change the custom Icon.... for the Colour I can change it in the device properties but it does not show the icon on the dashboard as shown in the image.
![image](https://user-images.githubusercontent.com/5776333/61657005-38e9f380-acc2-11e9-804d-3ba7da602500.png)

gizmocuz:
Yes this is the RGB(W/WW) color picker icon

Harald777:
Would love to see that it is possible to change the RGB Icon to a custom icon, currently not possible.

waltervl:
Solved with pr

PR #4125, opened 2020-05-01
Custom icon for RBGW YeeLight's

jordanalex04:
Changing the javascript files to allow users to use custom icons for the YeeLight RGBW bulbs. If the user still has the default icon selected then the icon will be the usual RGB cubes (www/images/RGB48...). 

This fixes issue #3541


gizmocuz:
Thanks! This patch not only affects YeeLight RGBW types but all RGBW types.
The only issue is that it is not really consistent as some third party developers will still receive a JSON with a wrong image
And the Custom Icon selector has a 'dimmer' icon at the top instead of the RGB icon

PR #4134, opened 2020-05-04
Document the app root command line parameter in usage

nuxeh:
Update the usage string for the main Domoticz binary to include the `-approot` parameter.


nuxeh:
I think this parameter is a very useful thing to have, to add flexibility in install location. Having it present in the usage string would have saved me time, so it would be nice to have this in. Thanks!

gizmocuz:
Thank you for the PR. The PR was not made against the active branch (development) and could therefor not be merged via Github. I have done this via the command line against the development branch.

PR #4138, opened 2020-05-05
smtp client improved error logging

pieterh:
since it is sometimes difficult to find the cause smtp email troubles, I have added some more detailed error logging. 

Example:
2020-05-05 15:20:10.863 Error: SMTP Mailer: Error sending Email to: <xxx@xxxx> !
2020-05-05 15:20:10.863 Error: libcurl: (67)
2020-05-05 15:20:10.863 Error: Login denied


gizmocuz:
Thank you !!

Issue #2435, opened 2018-05-27
[Teleinfo ] Linky - PAPP not used, only IINST


manolo75:
Hi,
I moved a few weeks ago and got a brand new linky. My teleinfo was working like a charm before my move.
I noticed that here, it only increase/decrease by 230 steps, so I guess there is a trouble.
I checked what is going on with my USB device and noticed I received correctly the informations, but it seems there is a bug. 
Here is an exemple today : 

ISOUSC 60 <
HCHC 000256932 !
HCHP 000368014 )
PTEC HP..
**IINST 005** \
IMAX 060 E
**PAPP 01140 '**
HHPHC A ,
MOTDETAT 000000 B
ADCO 031764199281 J
OPTARIF HC.. <
ISOUSC 60 <
HCHC 000256932 !
HCHP 000368014 )
PTEC HP..
**IINST 005** \
IMAX 060 E
**PAPP 01140 '**
HHPHC A ,
MOTDETAT 000000 B
ADCO 031764199281 J
OPTARIF HC.. <
ISOUSC 60 <
HCHC 000256932 !
HCHP 000368015 *
PTEC HP..
**IINST 005** \

So PAPP is not used. It is only using the IINST value to calculate the current consommation. 
May be it is due to the fact I have cheap hours and normal hours (heures pleines et heures creuses), and that case was never tested before (PTEC value).

You can contact me if you need me for testing.
My version of Domoticz is the latest beta (3.9501) on a Raspberry 2 with Raspbian.

Regards.

Manu

Martial83:
Hi, 
I agree with that: i have an ecodevices, connected on a linky, and production PV.
When i have production with PAPP=0, but IINST1,IINST2 and IINST3 not 0, the value in domoticz is cranky !
from ecodevices:
<T1_IINST>2</T1_IINST>
<T1_IINST1>2</T1_IINST1>
<T1_IINST2>4</T1_IINST2>
<T1_IINST3>2</T1_IINST3>
<T1_PPAP>0</T1_PPAP>
<T1_PAPP>0</T1_PAPP>
and from domoticz (json):
Usage  "1840 Watt"

I can't use the value Usage to known when i am self-sufficient.
Thank you in advance for your help.

waltervl:
Can be closed as fix is merged.

PR #4142, opened 2020-05-06
Issue/2435

laoranz:
In the EcoDevice class, the parameter withPAPP from Teleinfo struct is never set at true.
Also when Teleinfo struct is used in ProcessTeleinfo function, PAPP value is estimate by this calcul :
`teleinfo.PAPP = (teleinfo.triphase ? (teleinfo.IINST1 + teleinfo.IINST2 + teleinfo.IINST3) : (teleinfo.IINST)) * 230;`

This patch set withPAPP at true when PAPP > 0

This fix should correct this issue #2435


gizmocuz:
Thanks!!

PR #4145, opened 2020-05-07
Do not install dependencies binaries on make install

danielhoperez:
After 2020.2, due to the switch from external dependencies to git submodules, make install tries to install these dependencies when they are actually not needed and bloats the installation a bit.
On top of that, mosquitto make install is broken (and apparently has been broken basically forever) when using cmake instead of regular Makefiles on linux.  It fails to build documentation and then fails to install these doc files.
As far as I know adding EXCLUDE_FROM_ALL to add_subdirectory prevents make install from diving into these submodules and thus installing only domoticz own files.


gizmocuz:
Thank you for the PR. Do we have to add this for MINIZIP as well ? I am missing this addition here

danielhoperez:
The main reason I did not add it for minizip is that I did not see any files being installed. However, for the sake of correctness, it might actually make sense to add it as well.

danielhoperez:
I can see the windows building failure. I am not much of a windows guy myself. But it looks like it is unrelated to the patch and is actually because of innosetup seems to have released a new version and the building scripts are trying to invoke 6.0.3 when the version being downloaded is 6.0.4

gizmocuz:
Thanks!!!

PR #4151, opened 2020-05-09
OpenWebNet: gateway info and time synchronization

Spugna85:
- added 'gateway interface management' (WHO = 13)
- added Date / time check and synchronization (resolution 1 minute, minimum 5 minutes)
NOTE: if the OWN gateway used is configured as master for date / time synchronization,
synchronization will be propagated to all devices configured as slaves.


gizmocuz:
tm_gmtoff does not exist under Windows

gizmocuz:
Thanks!!

PR #4153, opened 2020-05-10
Strip innosetup version from path 

danielhoperez:
This will prevent ci from breaking on every innosetup new version.
Another option would be to move the directory to strip the version.
I think it depends if you want to keep control over innosetup version or not.


gizmocuz:
Thank you for the PR! If you could make it that it is not depended on any version that would be great!

danielhoperez:
I think this one should do it

gizmocuz:
Yes!! Great, thank you very much for this one ! One check less ;)

PR #4158, opened 2020-05-12
Split buienradar sensors ability

akamming:
code change on request of some users in [this forum thread](https://www.domoticz.com/forum/viewtopic.php?f=31&t=31853&p=247212#p247212). This adds an option to the buienradar plugin to split the humtemp sensor into seperate hum and temp sensors, just like this option was available in the python plugin. allowing users to keep their current lua scripts if they want to switch to the native buienradar plugin

note: in theory the baro sensor would now also become seperate if the parameter is set to "yes". i was unable to test this part of the code , because i was unable to find a buienradar station which has this barometric reading.


gizmocuz:
@akamming , sorry I would like to keep it consistent with all hardware/plugins
It should be very easy to change their scripts I suppose

PR #4161, opened 2020-05-14
Logging,  added optional parm level to addlogmessage JSON

rwaaren:


gizmocuz:
I am not sure if this should be part of the Domoticz code or part of some script/system
Because this is just to log a message, maybe this needs to done upstream somewhere ?

rwaaren:
> I am not sure if this should be part of the Domoticz code or part of some script/system
> Because this is just to log a message, maybe this needs to done upstream somewhere ?

It is a [Feature request](https://www.domoticz.com/forum/viewtopic.php?f=28&t=32670) and my thinking was that we already provide the option to use this to log a message but now only as status.
This PR only adds the option to choose the type of the logmessage which makes it easier to filter.

Tsjippy:
I need this as I have a python script running which posts sensor data to domoticz using json. If the script catches an error I want to sent that error to domoticz as an error not as an status

gizmocuz:
That's perfectly fine. Just before you call this function add your status text.

rwaaren:
@gizmocuz  Do you have enough information to either merge or reject (close) this PR?

gizmocuz:
@rwaaren , yes, I think it is up to the caller to build the string to be logged

Tsjippy:
I don’t understand, do you mean this is already possible? how does this work?

gizmocuz:
@Tsjippy , easy, the JSON addlogmessage JSON call expects one parameter, a string that needs to be added in the log.
Whatever you put in the string is up to you, you have to create/format the string, then call this function

MrHobbes74:
I guess the point is that this function currently logs everything as level ‘status’ while he wants to log an ‘error’.

gizmocuz:
Oeps sorry, totally missed that one indeed ! I implemented it here:

https://github.com/domoticz/domoticz/commit/a29cc48a20c9e5ba49f0b0ba802a939341f9c214
(and just added the missing OK status)

Tsjippy:
Thanks!

rwaaren:
> Oeps sorry, totally missed that one indeed ! I implemented it here:
> 
> [a29cc48](https://github.com/domoticz/domoticz/commit/a29cc48a20c9e5ba49f0b0ba802a939341f9c214)
> (and just added the missing OK status)

Thx!

PR #4163, opened 2020-05-15
Feature/meteorologisk hardware integration

jeje42:
Hi,

Integration of meteorologisk:
      - implementation on backend and frontend the same way as darksky. The api docs can be found here: https://api.met.no/.

This api is free and does not require any api key. It is provided by the Norwegian Meteorological Institute and has a lot of potential for forecast. The only requirement that comes from this organisation is to not query too much the api, and I think that should not be a problem with domoticz.

I implemented a new hardware to integrate this api. The data may not directly correspond to the current time but should be very close, most of the time the next hour. For example, if domoticz queries the api at 19h30, the closest data that can be extracted is valid at 20h.
    
If no location is provided on the hardware, the global location set in the global settings is taken.

Like said previously, the api would be very useful to build a forecast component as Darsky support is going to end (if I understand it right according to their website).

I tried to comply with the code style, just tell me if more configuration is needed. Also, thanks for building such an amazing software :)


gizmocuz:
Thank you very much for your PR, all looking perfect, except one thing in hardware.js

It looks you removed a bit to much ?

image

jeje42:
I replaced this if condition because It was duplicated from the previous if (current lines 3673-3676 if I am not mistaken). Inside the condition, I remove the apikey related line because no apikey is necessary with meteorologisk.

I do not know precisely why one of the build fails in the PR. It looks related to windows but I could not yet setup a windows dev environnment :).

gizmocuz:
There is indeed a duplication here, thanks for spotting.
Question, instead of the new html block, could you also use an existing block like 'divlocation' ?
If not, we keep the additional block of course

Please add this also to the visual studio project, see the AppVeyor build log/error

gizmocuz:
Thanks for the change, now only the windows project needs to be adjusted, are you able to do so ?

RicardP:
@jeje42 Sorry for hijacking but this plugin sounds great, I know NMI is very good in prediction quality :) Could the NMI-API even provide forecast for a requested time, say every full hour 24 hours  ahead = one hour resolutions? I am really missing a weather forecast "future graph" in Domoticz, for different weather sensors like the present historical data. Historical data is mainly curiosities, forecast can be used a lot more... :) My idea is to control my house 5 hours slow reacting concrete Floor Heating "in advance", to coming hours wind-sun-temp outdoor properties, as a variable for predictive energy need "now" to influence room temperature hours later... :)

jeje42:
Hi @RicardP and thanks for your suggestion :) I think it can provide such need, some example can be found on the norwegian website using this api: https://www.yr.no/place/Norway/Oslo/Oslo/Oslo/hour_by_hour.html (I tried Oslo and several cities in Europe, looks like the data is quite complete). 
Also I think the default call response fetched by the hardware is quite complete, for example: https://api.met.no/weatherapi/locationforecast/2.0/.json?lat=59.930161&lon=10.71872 . So I agree on the fact that met.no has a lot of potential on prediction usage.
The complete api docs can be found here: https://api.met.no/

@gizmocuz I tried to setup a windows dev environnment following the page: https://www.domoticz.com/wiki/Windows . However I get errors trying to build master as a first step. Are there any further VS configuration required?
I am not very used to VS, so if anyone wants to do it faster than me no problem :)

gizmocuz:
@jeje42 , thank you for this !
I now also see another issue, you made your PR not against the active branch (development)
Is it possible to do this ?

jeje42:
Sure, I read somewhere the PR needs to be done against master, I will rebase on development these days.

gizmocuz:
Thank you for the changes. Is it possible you could have a look at our build systems output ? (https://ci.appveyor.com/project/gizmocuz/domoticz/builds/32987304)
There are some compile errors

jeje42:
Thanks, I am still trying to get my windows environment working, but the build still fails on development branch. I followed the instructions in INSTALL.md about compiling on Windows.
The main error is about stdafx.h not found but the precompiled headers section of the project looks fine. Are there other things I need to be aware of to get the project to build on Visual Studio 2019 ?

gizmocuz:
@jeje42 , you just need to install Visual Studio 2019 Community edition, download the win32 package (see install.md), unpack this in the msbuild folder
Launch visual studio, the first time it will ask you how to configure the keyboard shortcuts, choose c++
From the msbuild folder, open the solution file (sln) and start compiling (F7)

jeje42:
Thanks, I could finally build but I realized that strptime is not available on Windows. I have to rewrite the implementation a bit before this pull request is ready for reviewing.

gizmocuz:
Depending on the string, you maybe could use the function ParseSQLdatetime (build into domoticz)

jeje42:
Thank you for your suggestion, I could reuse ParseSQLdatetime to parse an iso date, which is quite close to sql date format. Both builds are passing right now :)

main/localtime_r.cpp
```diff
@@ -64,6 +65,27 @@ bool ParseSQLdatetime(time_t &time, struct tm &result, const std::string &szSQLd
...
+ */
+bool ParseISOdatetime(time_t &time, struct tm &result, const std::string &sISOdate) {
+	if (sISOdate.length() < 20)
+	{
+		return false;
+	}
+
+	std::string sDateWithoutZ = sISOdate.substr(0, 19);
+	std::vector<std::string> splittedDate;
+	boost::split(splittedDate, sDateWithoutZ, boost::is_any_of("T"));
```
gizmocuz: Is it possible to use the Domoticz StringSplit function from Helper.h ?

hardware/Meteorologisk.cpp
```diff
@@ -0,0 +1,446 @@
...
+#include "../httpclient/HTTPClient.h"
+#include "../main/json_helper.h"
+#include "../main/RFXtrx.h"
+#include "../main/mainworker.h"
+#include "../main/SQLHelper.h"
+
+#define round(a) ( int ) ( a + .5 )
+
+#ifdef _DEBUG
+#define DEBUG_MeteorologiskR
```
gizmocuz: Please comment the two debug lines as some are building in Debug mode

hardware/Meteorologisk.cpp
```diff
@@ -0,0 +1,446 @@
...
+	{
+		windspeed_ms = static_cast<float>(atof(instantData["wind_speed"].asString().c_str()));
+	}
+	if (instantData["wind_speed_of_gust"].empty()==false)
+	{
+		windgust_ms = static_cast<float>(atof(instantData["wind_speed_of_gust"].asString().c_str()));
+	}
+	if (wind_degrees!=-1)
+	{
+		RBUF tsen;
```
gizmocuz: Please use CDomoticzHardwareBase::SendWind

hardware/Meteorologisk.cpp
```diff
@@ -0,0 +1,446 @@
...
+
+	//Visibility
+	if (root["currently"]["visibility"].empty()==false)
+	{
+		if ((root["currently"]["visibility"] != "N/A") && (root["currently"]["visibility"] != "--"))
+		{
+			float visibility = static_cast<float>(atof(root["currently"]["visibility"].asString().c_str()))*1.60934f; //miles to km
+			if (visibility>=0)
+			{
+				_tGeneralDevice gdevice;
```
gizmocuz: Please use CDomoticzHardwareBase::SendVisibilitySensor

gizmocuz (commented):
Are you sure the radiation is in DU units ?
I tried the API but did not receive any radiantion value...

main/localtime_r.cpp
jeje42: Yes, it should be fixed now

hardware/Meteorologisk.cpp
jeje42: Done

hardware/Meteorologisk.cpp
jeje42: Done, based on the Buienrader implementation

hardware/Meteorologisk.cpp
jeje42: Just removed it as I do not see it in Locationforecast either, same for radiation

jeje42:
@gizmocuz thanks for your feedback, I refactored a bit in Meteorologisk.cpp and the issues there should be fixed.

ejantzen:
hello
good job, 
but some time, the "wind_speed_of_gust" is not in the result. 
https://api.met.no/weatherapi/locationforecast/2.0/.json?lat=45.767201&lon=4.875536
We have the 2 other wind values.
Can you suppress the test on wind_speed_of_gust not found.
Thank's
Eric J, France

waltervl:
Nice work, but still some usability issues:

1. What to fill in in the location field? People are having difficulties to fill it in. Should it be something like lat=45.767201&lon=4.875536 ? If the do somthing wrong here domoticz crashes (see https://www.domoticz.com/forum/viewtopic.php?f=12&t=34975&p=264480) Please add some instruction to the Add Hardware page.

2. If you leave the location field empty It will take the default location of Domoticz (Setup-Settings). But then you get an ERROR in the log everytime the devices are updated: "Error: Meteorologisk: No location fount for this hardware. Falling back to global domoticz location settings". A normal log entry should be enough or even not needed, the error  for sure is overkill.

gizmocuz:
@waltervl , if there is missing a small help line in the gui, please add it. Indeed when empty it should take the default location
2, yes it should log it once (info), you could use a Boolean for this.
If you want, please make a pull request with the needed changes

eirikman:
@jeje42 , Just testing the new hardware for the first time in domoticz 2021.1 Great addition!
I am located Norway, and is currently two hours ahead of GMT with daylight saving, and this seems to causing an issue on the timeseries used for data.

It appers that the local time on my system is compared to the UTZ reported from Meteorologisk, and hence retrieved values are currently 2 hours ahead.

Any chance that this could be fixed?

gizmocuz:
@eirikman , It's best to look at the source file (c++ code) that handles the meteorologisk handling even if you don't know much about c++
There is no date/time stamping involved here and values can not be retreived 2 hours ahead or behind.
It could be that your geo 'location' in your Domoticz settings is not correct ? Or the location you provided as plugin parameter is not correct ?
Also check your time zone on the hardware you installed Domoticz on... If this is a raspberry pi, issue 'sudo raspi-config' else issue 'sudo dpkg-reconfigure tzdata'

eirikman:
@gizmocuz, I am currently running on Ubuntu 20.04 LTS on WSL.
Edit: Addet result from date comand
`eirik@DESKTOP-TL9M4N9:~$ date
Sun Apr 25 21:17:57 CEST 2021`

I know nothing about c++, but looking at the code I find the following:
```c++
	Json::Value timeseries = root["properties"]["timeseries"];

	Json::Value selectedTimeserie = 0;
	time_t now = time(nullptr);

	for(int i = 0; i < (int)timeseries.size(); i++)
	{
		Json::Value timeserie = timeseries[i];
		std::string stime = timeserie["time"].asString();

		time_t parsedDate;
		struct tm parsedDateTm;

		if(ParseISOdatetime(parsedDate, parsedDateTm, stime))
		{
			double diff = difftime(mktime(&parsedDateTm), now);
			if(diff > 0)
			{
				selectedTimeserie = timeserie;
				i = timeseries.size();
			}
		}
	}
```
To me it looks like the program search trough the timeseries retrieved by the API, and select the first timeserie with a time greater than `now`. This is where I believe the problem occurs: `now` appears to be local time, and timeseries retrieved from the API are in UTC.

An example from my system:
image

Data is recuested 15:05 local time. The data corrsponds to the forecast for 16:00 UTC returned by the API, and not 14:00 UTC as expected (data below retieved from [api.met.no][L1])
```
        "time": "2021-04-25T16:00:00Z",
        "data": {
          "instant": {
            "details": {
              "air_pressure_at_sea_level": 1028.1,
              "air_temperature": 7,
              "cloud_area_fraction": 0,
              "cloud_area_fraction_high": 0,
              "cloud_area_fraction_low": 0,
              "cloud_area_fraction_medium": 0,
              "dew_point_temperature": -4.8,
              "fog_area_fraction": 0,
              "relative_humidity": 46.3,
              "ultraviolet_index_clear_sky": 0.8,
              "wind_from_direction": 331.1,
              "wind_speed": 7.2,
              "wind_speed_of_gust": 12.8
```
```
        "time": "2021-04-25T14:00:00Z",
        "data": {
          "instant": {
            "details": {
              "air_pressure_at_sea_level": 1028.6,
              "air_temperature": 8.3,
              "cloud_area_fraction": 0.7,
              "cloud_area_fraction_high": 0.8,
              "cloud_area_fraction_low": 0,
              "cloud_area_fraction_medium": 0,
              "dew_point_temperature": -5.5,
              "fog_area_fraction": 0,
              "relative_humidity": 40.5,
              "ultraviolet_index_clear_sky": 2,
              "wind_from_direction": 324,
              "wind_speed": 6.5,
              "wind_speed_of_gust": 11.7
```

[L1]: https://api.met.no/weatherapi/locationforecast/2.0/documentation#!/data/get_complete

gizmocuz:
Yes your correct about that one !!

I debugged the code/results returned from the metrorologisk site.
It seems you get a lot of data returned, data from the past, and data up to about 10 days in the feature.
The hardware class tries to determine the closest date/time compared to the systems date/time

As the data returned from the website is in blocks of one hour, and your system date is 15:05 and the compare function in the code want to find a time block with a time higher then the actual system date/time, it is using the block from 16:00

For example it's 08:02 am here, and now it wants to use the result from 09:00am

We should indeed take the data from 08:00am...

But how accurate is this data ? Maybe this data is 'guessed' a week ago ?

I made some changes and in the current beta's (13226) the correct time block is used.

eirikman:
The forecast is updated at iregular intervals.

The time for last update of the forecast is found here as `"updated_at"`:
```
{
  "type": "Feature",
  "geometry": {
    "type": "Point",
    "coordinates": [
      xxx,
      yyy,
      zzz
    ]
  },
  "properties": {
    "meta": {
      "updated_at": "2021-04-26T05:58:39Z",
      "units": {
```
And in the responce header you can find the `expires` that tell you the the expected time for the next update (Idealy you should not send a new request before this)
```
{
  "accept-ranges": "bytes",
  "access-control-allow-headers": "Origin",
  "access-control-allow-methods": "GET",
  "access-control-allow-origin": "*",
  "age": "0",
  "connection": "keep-alive",
  "content-encoding": "gzip",
  "content-length": "5014",
  "content-type": "application/json",
  "date": "Mon, 26 Apr 2021 06:43:12 GMT",
  "expires": "Mon, 26 Apr 2021 07:14:03 GMT",
  "last-modified": "Mon, 26 Apr 2021 06:43:12 GMT",
  "server": "nginx/1.10.3 (Ubuntu)",
  "vary": "Accept, Accept-Encoding",
  "via": "1.1 varnish (Varnish/6.2)",
  "x-backend-host": "b_ybs_api3_b4_api_met_no_locationforecast",
  "x-varnish": "236697991"
}
```

gizmocuz:
Okey but for now you should get better results

Please feel free to make a pull request (PR) if you want to use the updated_at/expires

PR #4168, opened 2020-05-18
Closes #4167 outside temp was filled with inside temp data

### LINKED ISSUE #4167: Tesla temperature and Outside temperature are the same
Hi,

Since May 1st (About) the Tesla "Temparature" and "Outside temperature" are the same. Looking at the numbers, I suspect that Outside Temp. is given the values for Temparature. 

Would be great if this can be fixed!

MrHobbes74:


gizmocuz:
Thanks!!

PR #4176, opened 2020-05-23
mqtt internal for domoticz tasmota

megamarco833:
hi could be introduced a feature, maybe as hardware, where we can use topic to be listened and topic to publish, in order to be integrated for example devices like tasmotas or energy meter that publish on their own topics?
some examples:
1) energy meter:
i have device that publish power usage in topic:
`casa/energia/consumi/P`
and has a number payload that rapresent the enegry value like:
`254`
or for example:
`casa/energia/produzione/P 2850`
where topic is =` casa/energia/produzione/P`
payload = `2850`
or again
`casa/energia/consumi/IE 541021`
where topic is =` `casa/energia/consumi/IE`
payload = `541021`
that rapresent the power usage of today.

that could be fantastic have a customaize mqtt topic to listen so it can be addressed for example to a dummy device type energy that will be updated by mqtt with for example actual power usage and today tota energy usage.
immagine

2) another good example could be tasmota, where many devices with esp8266 are supported now
tasmota for lights publish the command in:
cmnd/<topic>/power<x> <payload>
example
cmnd/tasmota_156/power1 1
will be set relay1 on
cmnd/tasmota_156/power7 0
will be set relay7 off

to listen:
stat/tasmota_156/POWER16 = ON
say that relay16 is on

using a dummy device type light we can update the idx with status.

last example for roller shutter:
topic= `cmnd/tasmota_blind/shutterposition1`
mesage: `67`
this will move shutter1 at 67%

`cmnd/tasmota_blind/shutterposition3` `24`
this will move shutter3 at 24%

related to position...tasmota use SHUTTER to define a position...
`stat/tasmota_blind/RESULT = {"Shutter1":{"Position":44,"Direction":0,"Target":46},"Shutter2":{"Position":51,"Direction":0,"Target":51},"Shutter3":{"Position":53,"Direction":0,"Target":53},"Shutter4":{"Position":51,"Direction":0,"Target":51}}`
so looking at position an target domoticz will update dummy device type blind.

i think that i will require a big effort, but i think that having a mqtt internal where it will be possibile to customize topic to listen and idx to menage it will be fantastic to a wordwide integration with every kind of system
what do you think? it will be possible?
thanks


rwaaren:
Please don't use git pull request for feature requests.
Pull request are for prepared code changes. 

The domoticz forum has a subforum suggestions. All feature requests should be posted there.

PR #4178, opened 2020-05-23
RTL433: added wind_avg_m_s and wind_max_m_s for AlectoV1-Wind

annegerben:
added wind_avg_m_s and wind_max_m_s for AlectoV1-Wind to also show average and gust windspeed


gizmocuz:
Thank you for the PR, Please have a look at your 'moisture' code adjustments
In the original code, did you not get any wind speed measurements ?

annegerben:
In both stable and dev version windspeed is not shown, direction is.
Saw a few other user with the same issue: https://www.domoticz.com/forum/viewtopic.php?f=12&t=27847

I fixed the moisture mistake I made.

gizmocuz:
Thank you !

PR #4179, opened 2020-05-24
main: Allow both On and on in switchcmd parm of switchScene JSON

rwaaren:
see  [forum post](https://www.domoticz.com/forum/viewtopic.php?f=6&t=32804)

### DIFF
--- a/main/mainworker.cpp
+++ b/main/mainworker.cpp
@@ -12994,7 +12994,7 @@ bool MainWorker::SwitchScene(const std::string& idx, const std::string& switchcm
 bool MainWorker::SwitchScene(const uint64_t idx, std::string switchcmd, const std::string& User)
 {
 	std::vector<std::vector<std::string> > result;
-	int nValue = (switchcmd == "On") ? 1 : 0;
+	int nValue = (switchcmd == "On" || switchcmd == "on" ) ? 1 : 0;
 
 	std::string Name = "Unknown?";
 	_eSceneGroupType scenetype = SGTYPE_SCENE;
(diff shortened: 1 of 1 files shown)


kiddigital (commented):
Isn’t it easier and more readable to use the ‘stdlower’ function (Helper.cpp) so only checking the lowercase value instead of the different writing styles? 

They all mean the same in this situation.

gizmocuz:
Yes, but people should use the correct Case .... this is not a Domoticz issue but a user issue, should not be solved in code

rwaaren:
> Yes, but people should use the correct Case .... this is not a Domoticz issue but a user issue, should not be solved in code

I agree but in this case the confusion is understandable (if my understanding of the code is correct) because 'Off', 'off', 'OFF' are all considered as valid commands. (in fact anything else as 'On' is interpreted as 'Off'

gizmocuz:
@rwaaren , not according out JSON documentation on the wiki ?

rwaaren:
> not according out JSON documentation on the wiki ?

@gizmocuz; wiki documentation is correct on this. I was referring to the code. 

Fact that using 'on' is handled as 'Off' triggered me to create this PR

Do you want me to modify the code to return an error when this JSON is called with `switchcmd=off `  or `switchcmd=on` ?

kiddigital:
> Yes, but people should use the correct Case .... this is not a Domoticz issue but a user issue, should not be solved in code

Agree, would be better/cleaner if everything was case-sensitive and invalid values should be an error (and not a ‘default’ value/action if not an exact match).

gizmocuz:
Yes, for example for 'switchlight' this is checked/handled, because of a call to 'GetLightCommand', if the command is not valid for that type of hardware it returns false, and the JSON returns a ERROR
We could add a 'GetSceneCommand' and check for valid scene commands and bail out if there is a error (webserver.cpp, ~ line 6860)
Then again, i trust people that are able to call JSON API commands to also read the documentation

rwaaren:
> Yes, for example for 'switchlight' this is checked/handled, because of a call to 'GetLightCommand', 
> We could add a 'GetSceneCommand' 

Like that idea. Will have a look somewhere today

rwaaren:
Had a look and a couple of tries to implement it but my C++ knowledge is not enough for this yet.

PR #4184, opened 2020-05-26
TTNMQTT: Better GPS/Locations handling, distance calculation (Geofencing) and more

kiddigital:
This PR is an alternative for PR [#4123 ](https://github.com/domoticz/domoticz/pull/4123).

It processes GPS data from a CayenneLPP payload now as a custom sensor showing the Altitude AND puts the Lat;Lon;Alt data into a User Variable.

The TTN MetaData coming with each LoRa message is now also fully processed. This means when the Geo Location of a (usually static) sensor is set in the settings of a LoRa device through the TTN Console, this data becomes part of the MetaData and will be used as if the sensor has send GPS data.

Also when the Location of Domoticz is set in the generic settings, the distance between the Geo Location of the sensor and the 'home' location of the Domoticz instance is calculated (in meters) and stored into a separate device.

Using the standard notification functionality, this distance device can be used as a 'Geofence' to trigger an action if a sensor comes within reach or gets outside of reach. 

The distance calculations are done using the (now generic) DistanceEarth function originally found in the BuienRadar module.

Now also the MetaData of all LoRa gateways that have seen the sensor is being processed (instead of only the first one) and based on the 'best' reception of the signal, the one 'closest' to the sensor is selected. If the gateway has Geo Location data than this location is also stored as a separate sensor and the distance between the sensor and this nearest gateway gets calculated and stored as well.
  
Finally, the RSSI calculation has been improved a little.

### DIFF
--- a/hardware/Buienradar.cpp
+++ b/hardware/Buienradar.cpp
@@ -54,37 +54,6 @@ std::string ReadFile(std::string filename)
 }
 #endif
 
-//Haversine formula to calculate distance between two points
-
-#define earthRadiusKm 6371.0
-
-// This function converts decimal degrees to radians
-double deg2rad(double deg)
-{
-	return (deg * 3.14159265358979323846264338327950288 / 180.0);
-}
-
-/**
- * Returns the distance between two points on the Earth.
- * Direct translation from http://en.wikipedia.org/wiki/Haversine_formula
- * @param lat1d Latitude of the first point in degrees
- * @param lon1d Longitude of the first point in degrees
- * @param lat2d Latitude of the second point in degrees
- * @param lon2d Longitude of the second point in degrees
- * @return The distance between the two points in kilometers
- */
-double distanceEarth(double lat1d, double lon1d, double lat2d, double lon2d)
-{
-	double lat1r, lon1r, lat2r, lon2r, u, v;
-	lat1r = deg2rad(lat1d);
-	lon1r = deg2rad(lon1d);
-	lat2r = deg2rad(lat2d);
-	lon2r = deg2rad(lon2d);
-	u = sin((lat2r - lat1r) / 2);
-	v = sin((lon2r - lon1r) / 2);
-	return 2.0 * earthRadiusKm * asin(sqrt(u * u + cos(lat1r) * cos(lat2r) * v * v));
-}
-
 CBuienRadar::CBuienRadar(const int ID, const int iForecast, const int iThreshold, const std::string& Location)
 {
 	m_HwdID = ID;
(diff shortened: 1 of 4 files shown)


gizmocuz:
Thank you for the PR! Could you check the windows build systems output ? (AppVeyor)

kiddigital:
Will do :) Noticed the error and warnings, will fix them asap.

gizmocuz:
Thanks !!

jeroenkl:
Hi @kiddigital ,

Scripts works! Thanks.
I've got 3 custum sensors added automatically:
- **Geo**: Works fine! When adding a notification it should be 'afstand' instead of 'verbruik' as trigger

- **Home Distance:**  Not working fine, is this the distance to my Home location? What's the difference with "Geo" as the data of this sensor is correct (approx 11m, the Home Distance sensor is currently arround 4275m)

- **Gateway Distance**: Works fine, distance to my home gateway is currently approx 7m. 

I've got no other sensors like RSSI or Batt, should I get this also?
Can I get in contact with you directly to discuss offline further?

BR,

Jeroen

kiddigital:
> 
> Hi @kiddigital ,
> 
> Scripts works! Thanks.

Great :) You are welcome.

> I've got 3 custum sensors added automatically:
> 
>     * **Geo**: Works fine! When adding a notification it should be 'afstand' instead of 'verbruik' as trigger

Will take a look to see if/how to improve this.

> 
>     * **Home Distance:**  Not working fine, is this the distance to my Home location? What's the difference with "Geo" as the data of this sensor is correct (approx 11m, the Home Distance sensor is currently arround 4275m)

This is the distance between the reported Location of the device (so probably coming from the GPS sensor) and the Lat/Lon coordinates as defined under 'Setup -> Settings' top right 'Location' settings. Give it a try to calculate the distance manually and see if it check out.
> 
>     * **Gateway Distance**: Works fine, distance to my home gateway is currently approx 7m.
> 


Finally; Gateway distance is the distance between the device GPS coordinates and the known coordinates of the nearest gateway.

> 
> I've got no other sensors like RSSI or Batt, should I get this also?

RSSI/Signal strength is not a separate sensor but can be seen on the 'Device' tab.

For battery level, there is no real support yet. The LoRa specification has some direction but these don't seem to be implemented at least not in TTN as far a I am aware. You could send the battery level (or voltage) as a measurement from your device (as an analog input for example) and this measurement will become a separate Domoticz device.

> Can I get in contact with you directly to discuss offline further?

Feel free to mail me (see my updated profile).

> 
> BR,
> 
> Jeroen

jeroenkl:
@kiddigital 👍 
Yes, the Home Distance sensor is now also working fine after the geo change in domoticz (I took the same lan/lon as my LoRa gateway.
Also the RSSI level, see it in the device tab (sorry I missed it ;-) )

thanks and I'll contact you

PR #4194, opened 2020-05-31
RTL433, added RF Signal Strength in device tab based on reported SNR

FredrikFornstad:
This PR add support to show the Signal to Noise (SNR) information in the RF Signal Strength field in the Device tab when using rtl_433 and a RTL-SDR USB stick. Since the RF Signal Strength RSSI-field is 4 bit only (0-15) in Domoticz, I have done some mapping of the reported SNR figures (in dB) to the RSSI-variable. 

To enable SNR reporting from rtl_433, I have added the "-M level" switch to the rtl_433 command.

During my testing it seems that rtl_433 does not manage to detect any sensors if the SNR is 4dB or less, and communication starts to be reliable at 7-9dB. So in order not to waste any positions and have higher granularity for lower snr, I mapped 
`snr<4 to rssi=0 ;
snr=4-11 to rssi=1-8  (1 to 1) ; 
snr=12-25 to rssi=8-15  (2 snr steps per rssi step) ;
snr>25 to rssi=15`

For some reason, and I do not have enouogh Domoticz experience to understand why, not all device types support reporting of RF Signal Strength in DomoticzHardware.h, why I have skipped those in this PR. The functions that are missing this capability in DomoticzHardware.h are:
SendDistanceSensor()
SendMoistureSensor()
SendWattMeter()
SendKwhMeter()
If these functions were declared to accept the rssi-parameter as the other SendXXXX() functions, then this PR can be expanded to include those too.

I have tested it and it seems to work fine. I have not however tested the full range and assume that it is OK to report the full range (0-15) as "rssi" information.


gizmocuz:
@FredrikFornstad , thanks for the PR! I will do my best to add the missing RSSI levels for those type of sensors...

pipiche38:
I read that  with attention, and I'm surprised on 

> Since the RF Signal Strength RSSI-field is 4 bit only (0-15) in Domoticz,

I'm much more on the impression than Domoticz rssi (as displayed on the web is 0 to 11 and 12 stands for unknown !

Am I wrong ?

gizmocuz:
@pipiche38 , did you do even a code review before given such a comment?

pipiche38:
I'm not a C++ guru. But I did check the domoticz code and that is where I found 
12 is for unknown
0-11 is for rssi

FredrikFornstad:
I can see that rssi = 12 is set as default value in many places and that can be an indication of what you are saying (or just that if nothing else assume 75% of max rssi...).
I did look at the rssi variable and could see that in main/RFXtrx.h  it is defined as 4 bits. I did not find anywhere how it is actually used... So I assumed 0-15 was OK as I wrote in my PR text above. I am happy to remap to 0-11 if needed, but would like to use as much of the available range as possible. When I tested I did not have better RF condition than 10... Unless you know where in the code (beyond DomoticzHardware.cpp) this is used I will see if I can repeat my testing with better RF conditions.

gizmocuz:
We assume 12 for devices that have a direct connection to Domoticz and do not go via RF
I am nit saying your conversion is not correct but the above statement that all rssi Kerberos are wrong is stupid
Especially when that user is not able to read c++ code
Just ignore it, very happy with your PR! I was wondering how to get the snr but with my rfx version it was not mentioned in the help

pipiche38:
@gizmocuz thanks very much for such nice word, this make me very happy to participate to such community.
My comment was a question, isn't ? And I think that was on an acceptable tone .

@FredrikFornstad , I just commenting here from the user view, and I never see the RSSI above 11 displayed on the Web user Interface. I might be wrong and do apologize.

gizmocuz:
@pipiche38 , did not look like a question was it? Next time maybe change the tone!

FredrikFornstad:
@pipiche38 Thanks for proof-reading. I ended up making a test-compile with a different mapping so I could test the whole range, which was easier than trying to find my way in the code... For values above 12 the device tab shows '-', so effective range is 0-11 (actually I have not tested '0' either... but if that shows '-' too it is OK I think).

@gizmocuz I have made a new mapping to RSSI 0-11. Not sure how to amend to a closed and merged PR. Is it OK if I open a new PR?

gizmocuz:
@FredrikFornstad , Sure you can open a PR any time ;)

PR #4197, opened 2020-06-02
RTL433: Corrected reported RF Quality to match RSSI range of 0-11

FredrikFornstad:
Adjusted the SNR to RSSI mapping so that it uses RSSI range of 0-11. I have tried to make it as usable as possible with higher granularity for SNR=4-10 where a change in RF signal strength matters the most.


gizmocuz:
Thank you again !

dgrostoto:
Hello, can you tell a bit more about his ?
In particular, is this a breaking change about the database ? can we use an "old" populated database with this new code ?

PR #4198, opened 2020-06-02
Clang tidy 

xbeaudouin:
Hello,
This is work in progress, some clang tidy stuff... (Needs to see if the windows build is ok or not).


xbeaudouin:
By the way I see lots of : 
`#define round(a) ( int ) ( a + .5 )`
In the code. Is there any good reason about that to not use : std::lrint() instead?

xbeaudouin:
The WIP is done, except if you want I remove #define round by some internal cmath stuff?

gizmocuz:
Thank you for this PR and your work !
Two questions:
1. Why did you change the order of some includes ? Like putting logger.h behind another include, most of them where alphabetical and this easier to see if one was used or not
2. You are using the cmath include a lot, please put this in stdafx.h so we do not need to include it in the source files and makes compiling faster

There are some compile casting errors as you can see from the build systems:

Meteostick.cpp(230,9): warning C4244: 'initializing': conversion from 'float' to 'int', possible loss of data
TTNMQTT.cpp(532,46): warning C4244: 'argument': conversion from 'float' to 'const int', possible loss of data
main\dzVents.cpp(341,41): warning C4244: 'argument': conversion from 'const float' to 'const int', possible loss of data
main\dzVents.cpp(341,37): warning C4244: '=': conversion from 'int' to 'float', possible loss of data 
main\dzVents.cpp(415,41): warning C4244: 'argument': conversion from 'const float' to 'const int', possible loss of data
main\dzVents.cpp(415,37): warning C4244: '=': conversion from 'int' to 'float', possible loss of data 
main\dzVents.cpp(443,14): warning C4244: '=': conversion from 'const float' to 'int', possible loss of data
dzVents.cpp(445,17): warning C4244: '=': conversion from 'const float' to 'int', possible loss of data 
main\dzVents.cpp(447,20): warning C4244: '=': conversion from 'const float' to 'int', possible loss of data
main\dzVents.cpp(449,40): warning C4244: 'argument': conversion from 'const float' to 'const int', possible loss of data
main\dzVents.cpp(449,36): warning C4244: '=': conversion from 'int' to 'float', possible loss of data 
main\dzVents.cpp(484,41): warning C4244: 'argument': conversion from 'const float' to 'const int', 
main\dzVents.cpp(484,37): warning C4244: '=': conversion from 'int' to 'float', possible loss of data 
main\dzVents.cpp(534,14): warning C4244: '=': conversion from 'const float' to 'int', possible loss of 
main\dzVents.cpp(536,41): warning C4244: 'argument': conversion from 'const float' to 'const int', 
main\dzVents.cpp(536,37): warning C4244: '=': conversion from 'int' to 'float', possible loss of data 
main\dzVents.cpp(565,13): warning C4244: '=': conversion from 'const float' to 'int', possible loss of 
main\dzVents.cpp(642,19): warning C4244: '=': conversion from 'lua_Number' to 'float', possible loss 

Previously when compiling with GCC/msbuild we did not experience any compile issues nor warnings, why are the ctidy changes needed?
Sure I do support const iteration changes a big plus (for loops), but round-> std::round rint->std::rint, pow->std::pow ,  these are common I suppose,  making my type much more characters each time ;)

xbeaudouin:
About adding cmath into stdafx.h -> good idea.
What about my remark about the #define round() ? Do you think it is better for remove that and use std::lrint().
About warning in dzvents, this is strange I didn't touch it... AFAIK, will look anyway why there is this kind of warnings.
Will have a double check about the new warnings anyway.
Clang is really picky, very more than GCC. (I dunno about that on Windows, this is why I use the build system to test and check). There is also some other warnings on Clang, but if we can use cmath stuff this can be better I think for the project.

gizmocuz:
Please feel free to change the defined round function to lrint ... going to have to get used to this ;)
You can configure visual studio to use clang as well... community edition is free if you want to try it ;)

gizmocuz:
You are now using old style casting (not c++ style), I think you should also change BYTE to uint8_t
It seems like quite some more work to use a rounding function then the define of round
It also involves casting ....
What do you think ? Reverting all rounding calls to use the define again ?

xbeaudouin:
Well BYTE casting was done because windows build was complaining about that... 
I am not really fan of defines in general because of casting issues in general).
About old style casting, what are the "new" style ? What would you prefer ?
In fact some of define round was not used anymore in the code...

gizmocuz:
Not sure why a msbuild complained, but these days we use uint8_t

old style var a=(int)b, new style var a = static_cast<int>(b)

I think we better keep the define, this will simplify the code , else we get somthing like

var a = static_cast<int>std::rlint(b)

vs

var a = round(b)

xbeaudouin:
Why not create a function instead of define?

gizmocuz:
@xbeaudouin , defines seems faster. But you just define it in stdafx.h, and you don't have to declare it in any source file anymore

xbeaudouin:
Yeah defines seems faster, but this also growing the code. Modern processors (even RPI 1) have decent caching systems, so for a small function, this should be ok with less duplicate code in the software.
Are you ok I define this into stdafx.h, and use it only for the ones I had to cast ?

gizmocuz:
If you want to put it in a function, I think it is best to put it in Helper.cpp/h and include this file in the code where it is needed

xbeaudouin:
Yeah Helper.cpp/h is better (and since it is added on most part, this will include the function). Thanks :)

xbeaudouin:
@gizmocuz okay I have reverted the round stuff... You are totaly right, keep it simple and stupid, this is better for all of us. 
Regards

gizmocuz:
@xbeaudouin , thank you for this.
Does that also mean that we can directly use the 'round' function/define in other classes like BleBox.cpp ?

xbeaudouin:
Well didn't do that but, I can create a define into Helper.h ? What do you think of that?

gizmocuz:
I would keep and use the 'round' function (as define) as we had, but just move that define to helper.h
There is no need for dzRound , the define will work good as round

xbeaudouin:
Let's do it then :p

gizmocuz:
@xbeaudouin , are you going to revert some parts of this fix and use the 'round' define (again)?

xbeaudouin:
Well... I think I am out of time for that. What can we do ? forget this pr or doing something else?

gizmocuz:
@xbeaudouin , If possible, please revert the changes you made regarding the 'round' function/call

xbeaudouin:
Well let's give a try ;D

xbeaudouin:
@gizmocuz is it good for you ?

gizmocuz:
@xbeaudouin , yes thanks!!

Two questions:
stdafx.h, do we need the cmath header include ?

in:
hardware/eHouse/EhouseUdpListener.cpp, you have converted one round -> std::round

xbeaudouin:
cmath in stdafx.h is usefull because we use in some random places the cmath primitives.
Sorry about lost std::round in hardware/eHouse/EhouseUdpListener.cpp, this is fixed now

gizmocuz:
Thanks for this PR and all the work !!

PR #4200, opened 2020-06-02
TTNMQTT: Possibility to use TTN custom decoder for decoding (incl. batterylevel)

kiddigital:
This PR add the possibility of processing custom payload fields to the TTNMQTT module.

Within the The Things Network not only CayenneLPP payloads can be used, but there is also an option to use a custom 'external' decoder (custom Javascript decoder running on the TTN backend manageable through the TTN Console) to decode received payloads on the backend BEFORE they are send to MQTT subscribers.

These custom decoded fields become part of the MQTT payload (payload_fields) so can be used.

As the TTNMQTT module is one of those MQTT subscribers, if the structure of the decoded data is not too complex and similar to the defined CayenneLPP fields, the module tries to map the 'payload_fields' and processes them.

This means, within a customer decoder the fields with measurement should be called similar to the CayenneLPP field types:
- temp (or temperature)
- hum (or humidity)
- baro
- gps
- digital_input
- digital_output
- analog_input
- analog_output
- presence
- luminocity

and a special one called 'batterylevel' (percentage 0-100) which is processed as the 'battery level' of the device.

For GPS, the measurement value should be a single string like  '[lat];[lon];[alt]' where lat/lon in digital degrees and alt in meters.

So using the 'custom encoder' option of TTN and this processing, it becomes possible to process data from LoRa devices that do not send there data in CayenneLPP format.

(I will document this more properly on the Domoticz Wiki when part of the Beta and more testing has been done).

### DIFF
--- a/hardware/TTNMQTT.cpp
+++ b/hardware/TTNMQTT.cpp
@@ -64,9 +64,9 @@ CTTNMQTT::CTTNMQTT(const int ID, const std::string &IPAddress, const unsigned sh
 	m_TopicIn = Username + "/devices/+/up";
 
 #ifdef DEBUG_TTN_R
-	std::string sResult = ReadFile("ttn_mqtt.json");
+	std::string sResult = ReadFile("ttn_mqtt_stringfields.json");
 	mosquitto_message mqtt_msg;
-	mqtt_msg.topic = "domo_rob/devices/lopy4/up";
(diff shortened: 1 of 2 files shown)


jeroenkl:
Happy test as soon as it is merged.

gizmocuz:
Thanks!

PR #4206, opened 2020-06-04
Fix up OpenZWave include path handling

dwmw2:
The path specified by the pkg-config file will be, for example,
/usr/include/openzwave.

That directory needs to be on the compiler's include path, because
OpenZWave's own include files assume they can include each other
simply as (e.g.) "ValueIDIndexes.h"; not "openzwave/ValueIDIndexes.h"

Our own files do include <openzwave/Foo.h> though, which means that
the *parent* directory needs to be on the compilers's include path
too. We generally get lucky because /usr/include is automatically
included, so we find /usr/include/openzwave/Foo.h anyway.

Fix our C files to rely on the correct include path discovered from
pkg-config, and to include OpenZWave headers by name without the
erroneous openzwave/ prefix.

That means we can fix the ../open-zwave-read-only static build to use
the header files directly from there just like it does the static
library .a file, without requiring the 'sudo make install' step — and
without suffering a mismatch of static openzwave build vs. headers of
a different version that were installed on the system, which could
previously happen.

Tested with both static and dynamic builds of OpenZWave.


gizmocuz:
Thanks!!

svde:
Finally! Thanks!

Issue #4207, opened 2020-06-04
RFLink over TCP - version not detected


bilbolodz:
Domoticz 2020.2, RFlink with TCP over ESP8266 module.
After startup/restart RFLink is not detected (Setup->Hardware->"RFLink Version": empty).
Temporary solution: sending command to RFLink: "10;VERSION;" (Created RFLink Devices->Command: 10;VERSION;) version appear and exist to next restart.
Probably Domoticz "actively" no try to detect version. When connection is over USB, connecting to port causes device reset (Android booloader feature) and presents it's version.
On TCP connection there is no reset so there is no information about RFLink version.
Permanent solution: Domoticz after successful connection over TCP should ask (by command: "10;VERSION;") for RFLink version.

bilbolodz:
I've made a pull request to fix it: https://github.com/domoticz/domoticz/pull/4209

PR #4209, opened 2020-06-04
Fix for RFLink version reading using TCP connection

bilbolodz:
Fix for RFLink version reading using TCP connection.

### DIFF
--- a/hardware/RFLinkTCP.cpp
+++ b/hardware/RFLinkTCP.cpp
@@ -54,6 +54,7 @@ void CRFLinkTCP::OnConnect()
 	m_LastReceivedTime = mytime(NULL);
 	sOnConnected(this);
 	write("10;PING;\n");
+	write("10;VERSION;\n");
 }
 
 void CRFLinkTCP::OnDisconnect()


gizmocuz:
Thanks for the PR! If this a normal RFLink that is serial and you use (for example) ser2net to make it available over TCP ?
Please have a look in the code here:
RFLinkBase.cpp , line 679:

`if ((Name_ID.find("Nodo RadioFrequencyLink") != std::string::npos) || (Name_ID.find("RFLink Gateway") != std::string::npos))
`

This should request the version.

Maybe your Name_ID is different ?

I think the patch should be made in the base class ?

bilbolodz:
My RFlink is "normal RFLink" with connected ESP8266 module with  "serial bridge" software. I've described it in these issue: https://github.com/domoticz/domoticz/issues/4207
Problem occurs ONLY for TCP: when connection is made over USB Arduino is automatically reset and at start it presents firmware revision. When connection is over TCP NO Arduino reset is done so RFLink doesn't know nothing about "new connection" and NOT present version. Solution is a) modify Serial Bridge to reset Arduino on every incoming new TCP connection (require hardware modification) b) add to request version at start. So I think modification is in right place.

gizmocuz:
Yes in this case there is no other option... Got a link to the sketch ? Always interested in this ;)

bilbolodz:
I'm using ESPeasy with Ser2Net module: https://www.letscontrolit.com/wiki/index.php/Ser2Net
In case you have more RFLinks (for other frequencies) you can use ESP32 with 3 serial ports and these sketch: https://github.com/AlphaLima/ESP32-Serial-Bridge (Not yet tested by me).

gizmocuz:
Thanks ! Going to look at the ESP32-Serial-Bridge as the first link displays that it can not work with large data (like a P1 meter)

bilbolodz:
Yes my friend said that ESP-01 as serial server for RFLink sometimes is loosing some data (in case of some massive data flow). I'm going to build new version of my RFLink devices and integrate 3 RFLink (315,432,868MHz) on separate Arduino Mega 2560 modules into one using 3x Arduino Mega 2560 mini and ESP32 as serial bridge.

PR #4210, opened 2020-06-04
RTL433: Fixed bug so that user parameters can be used again

FredrikFornstad:
Realized I omitted a space when adding the -M level parameter to the command line. If the user adds any user-parameter in the webinterface it will be appended to "level" without any space in between… Fixed now. Sorry!

I also added snr to the two SendSwitch commands. I have not tested it, but think it should work. 

Finally removed unnecessary variable haveSnr and set default value to 12 instead of 0, so that the device tab will show "-" if no snr is reported.

### DIFF
--- a/hardware/Rtl433.cpp
+++ b/hardware/Rtl433.cpp
@@ -149,8 +149,7 @@ bool CRtl433::ParseData(std::map<std::string, std::string>& data)
 	bool haveUV = false;
 	float uvi = 0;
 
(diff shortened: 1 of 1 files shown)


FredrikFornstad:
I have now tested the with switches (that trigger the SendSwitch command) with this PR and snr is reported nicely in the device tab the same way sensors are.

gizmocuz:
Thanks !!

geschriever:
Is working again, thanks! Issue can be closed

PR #4213, opened 2020-06-05
Fix use of external minizip

dwmw2:
Fix an API breakage caused by https://github.com/nmoinvaz/minizip/issues/490 and correct the include paths we use for minizip header files.


gizmocuz:
I'm using vcpkg and we have to use it the way it is intended
Please no more PR for external libraries

dwmw2:
Not sure what you're saying there. This is absolutely not using minizip the way is it intended. The upstream minizip doesn't even work this way at all.

If you don't want to support properly using minizip via pkg-config perhaps we should remove that option and only support the internal one?

PR #4216, opened 2020-06-05
Update ZWaveCommands.h

matjoker:
Sound Switch Class not defined


gizmocuz:
Thanks, this needs to be implemented on multiple locations, are you able to assist with this ?
I don't have a node to test with

matjoker:
Sure, I can!
Thanks

Matthieu

PR #4223, opened 2020-06-09
Update Honeywell.cpp

deligoz:
WriteToHardware function pulls the nodeID from wrong id number. it should be nodeID= pCmd->LIGHTING2.id3;
current implementation only updates the first thermostat, when you try to update second thermostat it updates the first one


gizmocuz:
Thanks !!

PR #4225, opened 2020-06-09
Buienradar stability fix

akamming:
handles several situations in buienradar json, which occur every now and then(e.g. look at [this thread](https://www.domoticz.com/forum/viewtopic.php?f=12&t=32999&hilit=buienradar), i've seen this behaviour several times now) so buienradar gives less errors and updates the sensors more often:
- handle situation when lat/lon are reported as 0 by buienradar  (and recover from that if that was the case during initialisation of the hardware)
- handle situation when there is no rain prediction on serverside
- handle the stationid if it is reported as stationId (difference = capital I) 

"unfortunately"  buienradar is now stable, so i could not test all extra code. I have had this code running for this afternoon and do see a  lot less errors on buienradar. But a code review would be nice before merging.

### DIFF
--- a/hardware/Buienradar.cpp
+++ b/hardware/Buienradar.cpp
@@ -500,7 +500,7 @@ void CBuienRadar::GetMeterDetails()
 
 void CBuienRadar::GetRainPrediction()
 {
-	if (m_szMyLatitude.empty() or m_szMyLongitude.empty() or std::stoi(m_szMyLatitude)<1 or std::stoi(m_szMyLongitude)<1)
+	if (m_szMyLatitude.empty() || m_szMyLongitude.empty() ||  std::stoi(m_szMyLatitude)<1 || std::stoi(m_szMyLongitude)<1)
 		return;
 
 	std::string sResult;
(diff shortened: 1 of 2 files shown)


akamming:
hmmmm … the automated build gives a syntax error. On my system compilation is OK. I don't see what's wrong with the code....

gizmocuz:
@akamming , you can just click on 'details' next to the build system that has errors, then search for the word error

akamming:
> 
> 
> @akamming , you can just click on 'details' next to the build system that has errors, then search for the word error

i understand, but i get a syntax error on this code (which compiles ok on unix, and gives an error on windows apparrently)

and i don't see what's wrong with this code, and since i dont' have a compiler in windows, i cannot reproduce to fix... 

so hopefully someone can tell me what's different in the syntax on windows....

```
 if (m_szMyLatitude.empty() or m_szMyLongitude.empty() or std::stoi(m_szMyLatitude)<1 or std::stoi(m_szMyLongitude)<1)
                return;
```

UPDATE: I noticed a difference in the  if statement with other if statements, which use '||' for 'or', so changed this. Hopefully it now also compiles on windows. It is building now...

akamming:
and now the appveyor build gives me "Build execution time has reached the maximum allowed time for your plan (60 minutes)." This is something i cannot fix. When i look at the output the process did not yet reach the changed code. 

when i look at the output it got stuck at compiling ASyncTCP.cpp which was not touched by this PR (only buienradar.h and buienradar.cpp), so i would assume something goes wrong at the build server.

gizmocuz:
Thanks, it's compiling now.

PR #4226, opened 2020-06-10
Update OpenZWave.cpp

matjoker:
Create new switch devices for the Aeotec DoorBell 6 (Sound Switch class).
Just removed the comments as the test was successfull on my side


gizmocuz:
Thanks!!

PR #4228, opened 2020-06-10
Update OpenZWave.cpp

matjoker:
Valueupdate (On / Off) for an Aeotec DoorBell 6 device, using the SOUND_SWITCH Command Class


gizmocuz:
Thanks!

PR #4232, opened 2020-06-12
Update Honeywell.h

deligoz:
Honeywell.h for my other pull request


deligoz:
added to the other pull request

PR #4237, opened 2020-06-15
Opens domoticz window instead of stack overflow on desktop notification

mikewoudenberg:
I fixed the behaviour that every time I click on a desktop notification it goes to the stackoverflow post where this code came from. It will now open the domoticz window instead.


gizmocuz:
;) Thanks!

PR #4238, opened 2020-06-15
fix build with external minizip

ffontaine:
When building with the most active fork of minizip (i.e.
https://github.com/nmoinvaz/minizip), build fails because unzip.h is not
installed under minizip directory so remove this path from domoticz's
includes and add it to INCLUDE_DIRECTORIES when using the built-in
minizip


gizmocuz:
Using VCPKG here (try it), unzip IS installed in the minizip folder
When using the BUILD_IN minizip, the we do not use the above forked repo, but the domoticz repository where this is also installed inside the minizip folder
The repo you are pointing to (forked), does not include a unzip.h
So just use the BUILD_IN version...
Thinking about removing all external possibilities as it adds just a view kilobytes to the final executable.
When we add a new image it is even more

ffontaine:
The repo I'm pointing out is installing unzip.h if MZ_COMPAT is set to ON (which is the default), see https://github.com/nmoinvaz/minizip/blob/master/CMakeLists.txt.

PR #4257, opened 2020-07-06
Add "Has Node Failed" and "Replace Failed Node" to ZWave GUI

akamming:
I used the OZW CP a lot in case of failed nodes. which can be fixed in the combination of  "refresh node info", "has node failed""  and "replace failed node". Refresh node is already present in the ZWave screen. But Has node Failed and Replace Failed Node are not. resulting in the workaround to use to OZWCP. But i found out that domoticz becomes quite unstable when using the ozwcp. (segfaults and signal 11) 

so i made a PR to add these 2 functions to the OZW setup screen. (buttons are located next to the "refresh node"  button).

I tested a lot. Results:
1. The functionality works! 
2. Unstability (when switching to ozwcp) is gone.
3. but also found out that the "replace node" is not working as expected on OZW side.   I reported an [issue upstream](https://github.com/OpenZWave/open-zwave/issues/2293) for that problem. The code works around this issue, so from domoticz perspective is is OK now.

### DIFF
--- a/hardware/OpenZWave.cpp
+++ b/hardware/OpenZWave.cpp
@@ -814,10 +814,10 @@ void COpenZWave::OnZWaveNotification(OpenZWave::Notification const* _notificatio
 	case OpenZWave::Notification::Type_ControllerCommand:
 	{
 #ifndef _DEBUG
-		if (!m_bControllerCommandInProgress)
+		 if (!m_bControllerCommandInProgress)
 			return;
 #endif
-		_log.Log(LOG_STATUS, "OpenZWave: Type_ControllerCommand %s", _notification->GetAsString().c_str());
+		// _log.Log(LOG_STATUS, "OpenZWave: Type_ControllerCommand %s", _notification->GetAsString().c_str());
 
 		uint8_t controller_command = _notification->GetCommand();
 		if (controller_command == OpenZWave::Driver::ControllerCommand_RemoveDevice)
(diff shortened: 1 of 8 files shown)


gizmocuz:
@akamming , thank you for the PR. Maybe we should move this to the Node dropdown menu ?
could you also use the data-i18n for the string "<span>Press the include button on your new device...</span>"

The button/option "Has Node Failed" is sending the request to the controller, but it does not present to the user if it failed or not
Is it possible to add this as well ?

akamming:
@gizmocuz 
>Maybe we should move this to the Node dropdown menu ?

Sure, we could do that,  but to me that is not logical, cause now in the node menu we have generic commands for which you do not have have selected a node.   Now these buttons only light up when a node is selected.  Basically i reused the CJ as was created for the refresh node info. If you move these buttons to the node menu then it is not logical that the refresh info stays at is is now.

What i think would be the best CJ (but i did not want to change too much, so did not implement this, i now reused what was already there...)  is that we make a clear split between generic OZW functions which are in the node menu or below the table and specific functions for nodes appear in the table as icons, just like the recover node button icon is in every line on the right of table.  The result then would be:
- For an alive node there are icons for "refresh node info" (in case of not all sensors being reported) and "has node failed" (to force a check for a dead node) and "delete node" (button now on lower left).
- Far a dead node the GUI could show the icons "refresh node info" (sometimes recovers the node, "delete node" (the button now on the lower left) and "replace failed node" (to replace the device with a new one)
- the "recover node" icon which is now every line can be  removed. I think this is the same function as network heal, which is already in the node menu and is a generic function (if not, then ofcourse it should stay were it is).

> could you also use the data-i18n for the string "Press the include button on your new device..."

--> Done

> The button/option "Has Node Failed" is sending the request to the controller, but it does not present to the user if it failed or not.  Is it possible to add this as well ?

--> Done

akamming:
your 2nd and 3rd remark are done and added to the PR. 

About my comment on your 1st remark: I made a small change to see what it would look like:

zwaveactionicons

if you place  [these 2 files](https://github.com/domoticz/domoticz/files/4888756/Zwave.zip) in the www/app/hardware/setup dir you can play around with it yourself. (code not finished and tested yet, contains only the change to see what it would look like).

Let me know what you think of it.

gizmocuz:
@akamming , Thanks for the adjustments
I will have a play with the new status icons ;)

PR #4258, opened 2020-07-06
Database guard table preferences against double keys

rwaaren:
see issue on [forum post](https://www.domoticz.com/forum/viewtopic.php?f=6&t=33279#p251776)


gizmocuz:
Looking good! But is something in Domoticz causing a duplicate entry in this table ?

rwaaren:
> is something in Domoticz causing a duplicate entry in this table ?

I will try to pinpoint what caused the duplicate entries in this table with the OP of the forum post. I suspect a manual / recovery type of action.

PR #4262, opened 2020-07-11
RTL433: Added X-10 Security support

FredrikFornstad:
Added support for X-10 Security using the RTL-SDR usb stick. Tested OK with DS18 (Door/Window sensor), KR18 (Key Fob) and MS18 (Motion Sensor).

Please check my code so it matches your standard. I am fairly new to Domoticz and might have missunderstood or overlooked something. 

NodeID in X-10 Security is two bytes only, why ID1 will always be 0x00. However, X-10 Security is not the only type using Security1, so I tried to keep the code as generic as possible.


gizmocuz:

PR #4263, opened 2020-07-13
Better look in notifications tab

DewGew:
notifications

### DIFF
--- a/www/views/setup.html
+++ b/www/views/setup.html
@@ -386,11 +386,24 @@
 									<br>
 								</div>
 								<div class="span6">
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thanks !! Much cleaner indeed! Maybe we can/should make a kind of accordion style for the boxes like:

https://mdbootstrap.com/snippets/jquery/mdbootstrap/887912

DewGew:
Something like this:
notifications

gizmocuz:
That looks great! Personally i would put the arrow on the 'right' side ;)
Does this make sense ? It looks quite clean !

PR #4271, opened 2020-07-16
Added accordion to notifications tab

DewGew:
notifications

### DIFF
--- a/www/views/setup.html
+++ b/www/views/setup.html
@@ -26,6 +26,13 @@
 		    </table>
     </form>
 </div>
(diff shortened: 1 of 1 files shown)


gizmocuz:
Looking great, thank you !!

DewGew:
Maybe we can add this in more tabs..

gizmocuz:
The notification page was getting to big because of all the options.
The rest of the pages are looking OK, or do you think some good benefit from this as well ? (Which ones?)

Maybe one feature request in the notification page, when a notification system is enabled, maybe give a different color to the dropdown arrow ? Something to indicate that this system is active, maybe the dark blue background could be made dark green or something.

image

(#0e416d)

or

image

(#286554)

PR #4281, opened 2020-07-30
Feature: Mercedes Me connected Vehicle

kiddigital:
This is a PR for the beta implementation for the (Connected) Mercedes Me.

It utilizes the Daimler/Mercedes API's to retrieve data from your own Mercedes and make it available in Domoticz.

As there is/was nothing available in Domoticz that represented Vehicles except the 'eVehicle' implementation by @MrHobbes74, I implemented the Mercedes Me as an eVehicle (although mine is neither an Electric or Hybrid Mercedes).

It is hard to test if everything is still working for the Tesla implementation (unless someone wants to donate a Tesla for testing purposes ? :) ).

At the moment because the Mercedes Me API's only support the standard OAuth2 grant type using redirection, it is a little cumber sum to get the hardware up-and-running because you need to supply a valid OAuth2 Refresh-token (next to the client_id and client_secret).

To register you own Mercedes, you have to register with your existing Mercedes Me account at 'https://developer.mercedes-benz.com/' and request access for all 5 BYOCAR API's. Each of those API's has its own scope, so within the request for an access- and refresh token, all 5 scopes have to be supplied (space separated).

Hopefully others can test this (beta) implementation and also validate if the Tesla one is still working as it should.

Question, suggestions, additions, etc, please put them in the comments below.

### DIFF
--- a/hardware/eVehicles/MercApi.cpp
+++ b/hardware/eVehicles/MercApi.cpp
@@ -16,6 +16,7 @@ License: Public domain
 #include "../../httpclient/HTTPClient.h"
 #include "../../main/json_helper.h"
 #include "../../main/Helper.h"
+#include "../../main/SQLHelper.h"
 #include "../../webserver/Base64.h"
 #include <sstream>
 #include <iomanip>
@@ -42,14 +43,19 @@ CMercApi::CMercApi(const std::string username, const std::string password, const
 	m_username = username;
 	m_password = base64_encode(username);
 	m_VIN = vinnr;
-	m_authtoken = "";
-	m_refreshtoken = password;
+
+	m_authtoken = password;
+	m_accesstoken = "";
+	m_refreshtoken = "";
+	m_uservar_refreshtoken = "mercedesme_refreshtoken";
+	m_uservar_refreshtoken_idx = 0;
+	m_authenticating = false;
+
 	m_carid = 0;
 	m_config.car_name = "";
 	m_config.unit_miles = false;
 	m_config.distance_unit = "km";
 
-	m_authenticating = false;
 	m_crc = 0;
 	m_fields = "";
 
@@ -64,6 +70,19 @@ CMercApi::CMercApi(const std::string username, const std::string password, const
 	m_capabilities.has_charge_limit = false;
 	m_capabilities.has_custom_data = true;
 	m_capabilities.sleep_interval = 0;
+
+	std::vector<std::vector<std::string> > result;
+	result = m_sql.safe_query("SELECT ID, Value FROM UserVariables WHERE (Name=='%q')", m_uservar_refreshtoken.c_str());
+	if (result.empty())
+	{
+		m_sql.safe_query("INSERT INTO UserVariables (Name, ValueType, Value) VALUES ('%q',%d,'%q')", m_uservar_refreshtoken.c_str(), USERVARTYPE_STRING, m_refreshtoken.c_str());
+		result = m_sql.safe_query("SELECT ID, Value FROM UserVariables WHERE (Name=='%q')", m_uservar_refreshtoken.c_str());
+		if (result.empty())
+			return;
+	}
+
+	m_uservar_refreshtoken_idx = atoi(result[0][0].c_str());
+	m_refreshtoken = result[0][1].c_str();
 }
 
 CMercApi::~CMercApi()
@@ -82,13 +101,16 @@ bool CMercApi::RefreshLogin()
 	m_authenticating = true;
 	if (GetAuthToken("", "", true))
 	{	
+		std::string szLastUpdate = TimeToString(NULL, TF_DateTime);
+		m_sql.safe_query("UPDATE UserVariables SET Value='%q', LastUpdate='%q' WHERE (ID==%d)", m_refreshtoken.c_str(), szLastUpdate.c_str(), m_uservar_refreshtoken_idx);
+
 		_log.Log(LOG_NORM, "MercApi: Refresh successful.");
 		m_authenticating = false;
 		return true;
 	}
 
 	_log.Log(LOG_ERROR, "MercApi: Failed to refresh login credentials.");
-	m_authtoken = "";
+	m_accesstoken = "";
 	m_refreshtoken = "";
 	m_authenticating = false;
 	return false;
@@ -666,8 +688,8 @@ bool CMercApi::GetAuthToken(const std::string username, const std::string passwo
 		return false;
 	}
 
-	m_authtoken = _jsRoot["access_token"].asString();
-	if (m_authtoken.size() == 0)
+	m_accesstoken = _jsRoot["access_token"].asString();
+	if (m_accesstoken.size() == 0)
 	{
 		_log.Log(LOG_ERROR, "MercApi: Received token is zero length.");
 		return false;
@@ -698,7 +720,7 @@ bool CMercApi::SendToApi(const eApiMethod eMethod, const std::string& sUrl, cons
 	{
 		// If there is no token stored then there is no point in doing a request. Unless we specifically
 		// decide not to do authentication.
-		if (m_authtoken.size() == 0 && bSendAuthHeaders) 
+		if (m_accesstoken.size() == 0 && bSendAuthHeaders) 
 		{
 			_log.Log(LOG_ERROR, "MercApi: No access token available.");
 			return false;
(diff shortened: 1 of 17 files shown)


gizmocuz:
@MrHobbes74 , Is it maybe possible you could have a look to see if this PR is correct ?

kiddigital:
@MrHobbes74 , I have extended the base eVehicle class to allow for custom data that does not (yet) have a structured place.

Therefor I had to update the Tesla class as well, so please could you (re)test?

Shall we extend the VehicleData structure to support much more items like, individual doors, gas- or trunk opening, tire-pressure, (indoor)lights, etc.? Or maybe create a separate vehicleLock structure with all the possible doors and lids from which can be deduced if the whole vehicle is safely locked?

13az13:
Hi Guys,
Firstly, massive thank you for all the work here @kiddigital ! Any idea when this may get pushed live so I can update and hopefully utilise it’s features?
Many thanks

gizmocuz:
@kiddigital , 

Thanks for this PR!!, All code looks fine, I do have a few comments:

- If possible do not return true/false like this

`return (true); or return (false);`

just use

`return true; or return false;`

When printing a decimal number, please use %d, do not use %i
(I know that there are some parts in the Domoticz code that also use %i, and I hope to rewrite them as well)

kiddigital:
> 
> @kiddigital ,
> 
> Thanks for this PR!!, All code looks fine, I do have a few comments:
> 
>     * If possible do not return true/false like this
> 
> 
> `return (true); or return (false);`
> 
> just use
> 
> `return true; or return false;`
> 
> When printing a decimal number, please use %d, do not use %i
> (I know that there are some parts in the Domoticz code that also use %i, and I hope to rewrite them as well)

@gizmocuz, thx for reviewing. Your suggestions are applied. And some other small fixes :)

@MrHobbes74, hope you get the change to test this PR/beta as well to make sure the generic pieces are still working as they should for the Tesla implementation. Both here and on the forum I already noticed interest from potential users, so I have good hope for some Mercedes testers 👍 

The user-experience of 'configuring' the API is far from perfect, but not easy to improve yet. For this beta I think it is ok as you 'only' have to go trough this procedure once to get it running. I have mentioned this 'problem' to the API developers and they did respond that they would take a look. We have to see and find out if/how/when this indeed gets addressed. 

Looking forward to suggestions, bugreports, etc.

MrHobbes74:
Hi @kiddigital @gizmocuz, I have been away for a while, just discovered this PR. I will have a look at it after the weekend. Is that ok?

gizmocuz:
Sure! Take it easy 👍😁

kiddigital:
For those in need for more info, etc. see [this Domoticz forum topic](https://www.domoticz.com/forum/viewtopic.php?f=56&t=33734)

www/app/hardware/Hardware.html
```diff
@@ -1127,6 +1127,56 @@ <h2><span data-i18n="Hardware">Hardware</span>:</h2>
 			</tr>
 		</table>
 	</div>
+	<div id="divmercedes">
```
MrHobbes74, line 1130: I would suggest to leave options that are not supported out. And then fill in defaults in the constructor (e.g. specify one interval, and in the constructor use that same interval in both parameters etc)

hardware/eVehicles/eVehicle.cpp
```diff
@@ -310,6 +327,17 @@ void CeVehicle::Do_Work()
 		if (!m_loggedin || (sec_counter % 68400 == 0))
 		{
 			Login();
+			if(!m_loggedin)
```
MrHobbes74, line 330: interesting addition, thanks!

hardware/eVehicles/MercApi.cpp
```diff
@@ -0,0 +1,845 @@
...
+				_log.Debug(DEBUG_NORM, "MercApi: Found non empty field %s", id.c_str());
+
+				if (id == "doorlockstatusvehicle")
+				{
+					Json::Value iter2;
+					iter2 = iter[id];
+					if(!iter2["value"].empty())
+					{
+						_log.Debug(DEBUG_NORM, "MercApi: DoorLockStatusVehicle has value %s", iter2["value"].asString().c_str());
+						data.car_open = (iter2["value"].asString() == "1" || iter2["value"].asString() == "2" ? false : true);
```
MrHobbes74, line 385: What if a window is open? The tesla api considers everything that can be used to access the vehicle as an unlocked status (even though the car is 'locked', the window might be open hence the car is not locked). I would check all possible openings here ;-)

hardware/eVehicles/eVehicle.cpp
```diff
@@ -34,6 +36,7 @@ License: Public domain
 #define VEHICLE_ALERT_LOCK 2
 #define VEHICLE_LEVEL_BATTERY 1
 #define VEHICLE_COUNTER_ODO 1
+#define VEHICLE_CUSTOM 9
```
MrHobbes74, line 39: You add two custom types (custom sensor and custom text). I would make two definitions to separate the different device types (e.g. VEHICLE_CUSTOM_SENSOR and VEHICLE_CUSTOM_TEXT.

MrHobbes74 (changes requested):

MrHobbes74:
Hi @kiddigital , I have reviewed your changes. Of course have some comments ;-), please have a look. But I think it is a great use of the eVehicle module, as intended (even though it's not a electric vehicle :-D )! 
Regarding the custom data, I think it's fine to leave it, but the way you implement it now, don't you get double items for those you did categorize, like odo and doorlockstatus? 
We could add a separate VehicleLockData structure + get functions to display each locked item in a device + detect lock status in the eVehicle class, but I would not put specific members in, but a json array with integers/floats like you made in custom (but then only for lockable items).

www/app/hardware/Hardware.html
kiddigital: Agree.

Only challenge is that all (both the Mercedes and Tesla) hardware screens using the eVehicle framework have to put their data into the same constructor of the eVehicle class. And I think that all implementation classes used by the eVehicle class should also share a single constructor definition which means that some flexibility has to come from a more flexibel additional parameter in the constructor or custom properties that are set after construction. But that leads to car/brand specific code in the eVehicle class, which I think we should try to avoid.

I can at least make not used parameters hidden so the user is not bothered with them.

hardware/eVehicles/MercApi.cpp
kiddigital: Yes, I was thinking about that too. At least adjust the message if something is open. I look into it.

hardware/eVehicles/eVehicle.cpp
kiddigital: Hm.. I interpreted the defines as referring to what they measure/mean and not the the type of Domoticz sensor.

As these defines are used for the ID of the domoticz sensors (see also the screenshot I put on the forum topic for this PR) all custom sensors can be easily identified as they all share the same (parent) ID.

I was actually looking for a way to let the user decide if they want custom sensors to be created at all or not.

kiddigital:
Thx @MrHobbes74 😃

Have you been able to build this PR version and test if everything this works for the Tesla implementation?

I like the idea of creating a Lock structure with flexible properties and predefined Getters to extract the Lock state which can be individually implemented for each vehicle and be read generic. But I prefer to work on this after the PR made it into the Beta so other Mercedes owners can start betatesting this implementation.

I am pleased to hear that my attempt is at least in line with your ideas of the framework. And I saw that Mercedes just extended its offering of Hybrid engines for many models and the API's offer EV metrics so I will also attempt to map these metrics into the available EV structure and hope some can use/test that later for real.

To be continued... 😃

kiddigital:
@MrHobbes74 @gizmocuz I think the PR is in even better shape now (thx for the reviews) and ready for Beta testing. Some review changes and some improvements have been made. Also I added the processing of EV data but that is not easy to really test as my car in not electrical or hybrid. Hopefully there are some Mercedes drivers out there with either EV or Hybrid cars that can also test the Beta (and Tesla drivers as well).

gizmocuz:
@kiddigital , looking great ! I was only waiting for the answer to your question

"Have you been able to build this PR version and test if everything this works for the Tesla implementation?"

Sure fine by me to merge this PR

MrHobbes74:
I haven’t tested it, but from code I don’t expect it will fail. I suggest to merge and I will test in beta.

gizmocuz:
Thank you !!

PR #4284, opened 2020-08-02
Improvement/openweathermap

kiddigital:
This PR fixes the 'broken' Open Weather Map integration.

As Open Weather Map also released a new and improved version of the API, this PR also switches to this new API.

Multiple new Domoticz sensor have been added and improved. Now also Wind Gust, Chill temperature, UV Index, Weather description and 'feel' temperature are available as well. Now 9 sensors for the current weather situation.

Daily forecast added, showing the forecast for Today and 7 more days. Each day has 10 sensors (so 80 in total). Each day has for example min- max temperature and probability of Precipitation (changes of something coming down from the clouds in any shape or form like rain, snow, drizzle, hail. etc.). Also some textual sensor to build a dashboard, etc.

Also hourly forecast data processed if available for 48 hours. With 8 sensors per hour a total of 384 sensor are generated!

Multiple choices can be made to (not) use certain parts like current weather and/or forecast data and the polling interval can be provided (600 seconds by default).

Main reason to get the Open Weather Map integration back up and running is the disappearance of many services for actual free weather data.
- DarkSky has been acquired by Apple and will stop (API at least, Android already since 1-8-2020) by the end of 2021
- Wunderground (as part of The Weather Company, owned by IBM since 2016) also stopped there API's
- AccuWeather is focussed on US mainly.
- PWSWeather has a lot of free data from Personal Weather Stations, but is also US owned (AerisWeather)

Future ideas/improvements:
- You can also feed data from your own weather station, so the idea is to have different local Domoticz sensor become the input sensors for your local OWM weather station 
- Support for a lot of great weather Icons that could be utilized
- Generate local weather maps (not sure if animated) to display in Domoticz
- and more ...?? 

Comments/suggestions/bugs/etc. feel free to put in the comments...

### DIFF
--- a/main/mainworker.cpp
+++ b/main/mainworker.cpp
@@ -1043,7 +1043,7 @@ bool MainWorker::AddHardwareFromParams(
 		pHardware = new BleBox(ID, Mode1);
 		break;
 	case HTYPE_OpenWeatherMap:
-		pHardware = new COpenWeatherMap(ID, Username, Password);
+		pHardware = new COpenWeatherMap(ID, Username, Password, Mode1, Mode2, Mode3, Mode4);
 		break;
 	case HTYPE_Ec3kMeterTCP:
 		pHardware = new Ec3kMeterTCP(ID, Address, Port);
--- a/www/app/hardware/Hardware.html
+++ b/www/app/hardware/Hardware.html
@@ -979,6 +979,23 @@
 			</tr>
 		</table>
 	</div>
+	<div id="divopenweathermap">
+		<br>
+		<table class="display" id="hardwareparamsopenweathermap" border="0" cellpadding="0" cellspacing="20">
+			<tr>
+				<td align="right" style="width:110px"><label><span data-i18n="API Key"></span>:</label></td>
+				<td><input type="text" id="apikey" style="width: 250px; padding: .2em;" class="text ui-widget-content ui-corner-all" /></td>
+			</tr>
+			<tr>
+				<td align="right" style="width:110px"><label><span data-i18n="Location"></span>:</label></td>
+				<td>
+					<input type="text" id="location" style="width: 250px; padding: .2em;" class="text ui-widget-content ui-corner-all" /><br />
+					<span data-i18n="LatLong"></span><br />
+					<span data-i18n="(0 = Application default)"></span>
+				</td>
+			</tr>
+		</table>
+	</div>
 	<div id="divbuienradar">
 		<br>
 		<table class="display" id="hardwareparamsbuienradar" border="0" cellpadding="0" cellspacing="20">
--- a/www/app/hardware/Hardware.html
+++ b/www/app/hardware/Hardware.html
@@ -994,6 +994,22 @@
 					<span data-i18n="(0 = Application default)"></span>
 				</td>
 			</tr>
+			<tr>
+				<td align="right" style="width:110px"><label for="addcurrent"><span data-i18n="Current Weather">Current Weather</span>:</label></td>
+				<td><input type="checkbox" id="addcurrent" checked="checked" />&nbsp;<label for="addcurrent">&nbsp;</label></td>
+			</tr>
+			<tr>
+				<td align="right" style="width:110px"><label for="adddayforecast"><span data-i18n="Daily forecast">Daily forecast</span>:</label></td>
+				<td><input type="checkbox" id="adddayforecast" />&nbsp;<label for="adddayforecast">&nbsp;Forecast for today and coming 7 days  (if available).</label></td>
+			</tr>
+			<tr>
+				<td align="right" style="width:110px"><label for="addhourforecast"><span data-i18n="Hourly forecast">Hourly forecast</span>:</label></td>
+				<td><input type="checkbox" id="addhourforecast" />&nbsp;<label for="addhourforecast">&nbsp;Forecast for each hour for 48 hours (if available).</label></td>
+			</tr>
+			<tr>
+				<td align="right" style="width:110px"><label for="pollinterval"><span data-i18n="Poll Interval"></span>:</label></td>
+				<td><input type="text" id="intervalseconds" style="width: 80px; padding: .2em;" class="text ui-widget-content ui-corner-all" />&nbsp;(<span data-i18n="Seconds">Seconds</span>)</td>
+			</tr>
 		</table>
 	</div>
 	<div id="divbuienradar">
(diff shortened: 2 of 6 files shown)


hardware/DomoticzHardware.cpp
```diff
@@ -792,7 +792,8 @@ void CDomoticzHardwareBase::SendPercentageSensor(const int NodeID, const uint8_t
 	_tGeneralDevice gDevice;
 	gDevice.subtype = sTypePercentage;
 	gDevice.id = ChildID;
-	gDevice.intval1 = NodeID;
+	gDevice.battery_level = BatteryLevel;
+	gDevice.intval1 = (NodeID << 8) | ChildID;
```
gizmocuz: This will cause a lot of new Node ID's for various hardware. Please undo this change

gizmocuz: Battery Percentage is passed along with sDecodeRXMessage, not needed here

hardware/DomoticzHardware.cpp
```diff
@@ -822,6 +823,7 @@ void CDomoticzHardwareBase::SendVisibilitySensor(const int NodeID, const int Chi
 	_tGeneralDevice gDevice;
 	gDevice.subtype = sTypeVisibility;
 	gDevice.id = ChildID;
+	gDevice.battery_level = BatteryLevel;
```
gizmocuz: Battery Percentage is passed along with sDecodeRXMessage, not needed here

main/mainworker.cpp
```diff
@@ -10095,6 +10095,7 @@ void MainWorker::decode_General(const CDomoticzHardwareBase* pHardware, const tR
 		(subType == sTypePercentage) ||
 		(subType == sTypeWaterflow) ||
 		(subType == sTypePressure) ||
+		(subType == sTypeVisibility) ||
```
gizmocuz: This change will cause other hardware to have new ID's, please undo

www/app/hardware/Hardware.html
```diff
@@ -979,6 +979,39 @@ <h2><span data-i18n="Hardware">Hardware</span>:</h2>
...
+			<tr>
+				<td align="right" style="width:110px"><label for="adddayforecast"><span data-i18n="Daily forecast">Daily forecast</span>:</label></td>
+				<td><input type="checkbox" id="adddayforecast" />&nbsp;<label for="adddayforecast">&nbsp;Forecast for today and coming 7 days  (if available).</label></td>
+			</tr>
+			<tr>
+				<td align="right" style="width:110px"><label for="addhourforecast"><span data-i18n="Hourly forecast">Hourly forecast</span>:</label></td>
+				<td><input type="checkbox" id="addhourforecast" />&nbsp;<label for="addhourforecast">&nbsp;Forecast for each hour for 48 hours (if available).</label></td>
+			</tr>
+			<tr>
+				<td align="right" style="width:110px"><label for="pollinterval"><span data-i18n="Poll Interval"></span>:</label></td>
```
gizmocuz, line 1010: I don't think this should be a user defined option here. It might cause hammering and could result in a ban.
It also does not make any sense to poll faster then once every 5 minutes

www/app/hardware/Hardware.html
```diff
@@ -979,6 +979,39 @@ <h2><span data-i18n="Hardware">Hardware</span>:</h2>
...
+					<span data-i18n="(0 = Application default)"></span>
+				</td>
+			</tr>
+			<tr>
+				<td align="right" style="width:110px"><label for="addcurrent"><span data-i18n="Current Weather">Current Weather</span>:</label></td>
+				<td><input type="checkbox" id="addcurrent" checked="checked" />&nbsp;<label for="addcurrent">&nbsp;</label></td>
+			</tr>
+			<tr>
+				<td align="right" style="width:110px"><label for="adddayforecast"><span data-i18n="Daily forecast">Daily forecast</span>:</label></td>
+				<td><input type="checkbox" id="adddayforecast" />&nbsp;<label for="adddayforecast">&nbsp;Forecast for today and coming 7 days  (if available).</label></td>
```
gizmocuz: Noticed extra space before (if avail...)

www/app/hardware/Hardware.js
```diff
@@ -2055,7 +2089,37 @@ define(['app'], function (app) {
 					}
 				});
 			}
-			else if (text.indexOf("Buienradar") >= 0) {
```
gizmocuz, line 2058: Just to be sure, Buienradar settings are still working ?

gizmocuz (commented):
Thank you for this PR and work !
Please review my review

hardware/DomoticzHardware.cpp
kiddigital: > 
> 
> This will cause a lot of new Node ID's for various hardware. Please undo this change

True, will revert.

The behavior of the generated NodeID for the Percentage Sensor is different from a lot of other sensors because the ChildID is not part of the generated ID. Could be considered a 'bug'?

main/mainworker.cpp
kiddigital: For the Visibility sensor the NodeID is also not using the ChildID, sort of same as for Percentage. Maybe for consistency somehow make these sensors behave the same?

Will revert for this PR.

www/app/hardware/Hardware.html
kiddigital: Good point, Did considered it also and there is some sanity checking the values in the constructor (around line 80) with a minimum of 10 seconds and max of 86400 (a day). OWM has a limit of 60 per second! and 1mln per day :) But agree that some sensible values are needed.

I will set the minimum to 60 seconds and default remains 600 seconds.

www/app/hardware/Hardware.js
kiddigital: Everything still works as far as I could see and test :)

kiddigital:
@gizmocuz , review comments have been addressed. All should be fine now (I hope :) )

hardware/DomoticzHardware.cpp
gizmocuz: It is not a 'bug' but it is how it was designed to work. There are more sensors. I guess at that time i thought 255 percentage sensors should be enough per hardware ;)

gizmocuz:
Thanks for this PR and your hard work !!!

xztraz:
Awesome work!

I read someone is missing the poll intervall setting since he got a warning on too many calls via api.
https://www.domoticz.com/forum/viewtopic.php?f=12&t=35153

is every sensor having its own api call now? could the calls be bunched together?

kiddigital:
@xztraz , all sensors are updated using just 1 single API call.

The ability to change the poll frequency has been left out after discussing this with @gizmocuz 

Maybe we could think about a settings 'lower update frequency' that updates the sensors at half the standard frequency? That means an OWM module only performs half of the amount of API calls it does now.

gizmocuz:
Or just create another account ?

xztraz:
i've not got that error myself(one installation). maybe the guy in the forum is using same api on many installations. if only one call per update there should not be any problem.

PR #4293, opened 2020-08-17
Tune auto db backup (and prevent crash)

akamming:
makes sure now 1mb (256 DB pages) are copied per step in the db backup process.


gizmocuz:
Thanks!!

PR #4294, opened 2020-08-18
OpenWeatherMap: Fix for textual description of the hour of the forecast

kiddigital:
The text sensor 'Weather Description Hour _X_ Name' got a value describing the day (Today, Thu, Day 27) instead of the hour.

So when building a dashboard or something from these description, a wrong textual value was presented. Now it shows the correct hour (00..23).


gizmocuz:
Thanks again !

PR #4299, opened 2020-08-23
Hardware Monitor: Add Clock Speeds

akamming:
image

I could only test on Pi 3b+ and PI 4.

### DIFF
--- a/hardware/HardwareMonitor.cpp
+++ b/hardware/HardwareMonitor.cpp
@@ -63,6 +63,11 @@
 extern bool bHasInternalTemperature;
 extern std::string szInternalTemperatureCommand;
 
+extern bool bHasInternalClockSpeeds;
+extern std::string szInternalARMSpeedCommand;
+extern std::string szInternalV3DSpeedCommand;
+extern std::string szInternalCoreSpeedCommand;
+
 extern bool bHasInternalVoltage;
 extern std::string szInternalVoltageCommand;
 
@@ -177,6 +182,7 @@ void CHardwareMonitor::Do_Work()
 				try
 				{
 					FetchUnixCPU();
+					FetchClockSpeeds();
 				}
 				catch (...)
 				{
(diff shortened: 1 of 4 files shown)


gizmocuz:
Thank you for this PR !!

Question, is there a difference (in Chart) between the Clock Speed and the CPU Usage ? (Core) Isn't that a bit the same ?

akamming:
It's not the same.

E.g. when your system is at 100% cpu load with not enough cooling facilities. If the temperature then hits the throttling treshold  (i think it's around 85 celcius) your cpu usage will remain 100%, but your clock speed will be throttled down.

gizmocuz:
Great! Is it possible to also add support for the BCM2836 and BCM2837 chipset ?

akamming:
Are those also pi's with a vcgencmd installed? And which can be recognized with cat /proc/cpuinfo?

If so i can add the lines. But i cannot test...

gizmocuz:
I was browsing here:

https://www.raspberrypi.org/documentation/hardware/raspberrypi/

I would expect that they support vcgencmd ? If you don't think so, lets leave it and merge this PR ;)

akamming:
aha i now see these are all raspberry pi's with raspbian, so i think it's safe to assume they all have the vcgencmd command.

But what's weird is that according to this list my pi 3b+ should report as a 2837. 

But /proc/cpuinfo gives me 
```
Hardware	: BCM2835
Revision	: a020d3
Serial		: 00000000d63e391c
Model		: Raspberry Pi 3 Model B Plus Rev 1.3

```
Where BCM2835 is a pi zero according to this list.

If the vcgencmd does not exist, there will be no errors (the code checks for the wordt "frequency"  in the output), so they could be added. I personally would feel more confident if i  have conformation that these models would report the correct chips, but this is something i cannot do, cause i do not own these models.

i think it's up to you (maybe you own one of these models, then you can just do a 'cat /proc/cpuinfo' ?) To add them you only need to add the model numbers as of line  488 in main/domoticz.cpp.

gizmocuz:
Unfortunately I don't have access to those models anymore ;)

gizmocuz:
Thanks!

akamming:
Tx for merging.

Just had a brainwave. Instwad of checking on the model, probably better to check on the presence of the vcgencmd command. Then all (including future) models which support the capability will work. 

 Will make the pr somewhere later this week...

PR #4300, opened 2020-08-23
Wind speed multiply

hvbommel:
Most wind anemometers require calibration to report correct windspeed. This PR adds a "Multiply" field to the weather edit screen to allow increase or decrease of reported windspeeds and its graphs. The default value is 1. In that case no adjustment and wind readings will be as in previous versions. If you set Multiply to 1.35 for example it will increase readings accordingly. Setting Multiply to 0.8 would lower the windspeed value.

### DIFF
--- a/main/WebServer.cpp
+++ b/main/WebServer.cpp
@@ -10038,6 +10038,7 @@ namespace http {
 							if (dSubType != sTypeWIND5)
 							{
 								int intSpeed = atoi(strarray[2].c_str());
+								intSpeed *= AddjMulti;
 								if (m_sql.m_windunit != WINDUNIT_Beaufort)
 								{
 									sprintf(szTmp, "%.1f", float(intSpeed) * m_sql.m_windscale);
@@ -10053,6 +10054,7 @@ namespace http {
 							//if (dSubType!=sTypeWIND6) //problem in RFXCOM firmware? gust=speed?
 							{
 								int intGust = atoi(strarray[3].c_str());
+								intGust *= AddjMulti;
 								if (m_sql.m_windunit != WINDUNIT_Beaufort)
 								{
 									sprintf(szTmp, "%.1f", float(intGust) *m_sql.m_windscale);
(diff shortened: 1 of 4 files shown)


gizmocuz:
Thank you for the PR.
How do you calibrate a wind sensor ? It is not like watching the weather channel and adjusting your wind sensor to match it, as it very different on any position.
The multiplication should not be done in the webserver code, but in the mainworker::decode_Wind code (see decode_Temp for an example)
This way also all notifications will be correct

hvbommel:
The way I calibrate it is use the 4 official airfield reports (Metars) in my neighborhood (in my case EHEH, EBBL, EHVK and EDLV) . That way I get close. I have a brandnew TFA Dostmann wind station. Altough mounted at 10 meters height the wind it is reporting is 30% too low. With the Muliply I get values that match the Metar reports. I will have a look at mainworker::decode_Wind and see if I can put it in there. Then I will test it again. Thanks for your feedback.

gizmocuz:
Yes but even in your area the wind can be totally different...
I have no problem with the multiply ... but 30% is a lot! ;)

hvbommel:
I know 30% is a lot. But I happen to be a balloonpilot for 20 years now. I can judge wind and can tell the difference between 6 and 8 knots. And you may be right, over time I may find out its not 30% but 20% ... but still worth to compensate for. It also could be that my new windstation has not been calibrated very well. All I know is that the anemometer cups turn with no friction. 

I moved part of the codechanges to mainworker.cpp now. I have (quick) verified that both wind and temperature adjustments still work in both the regular display and in the graphs.

gizmocuz:
Thanks!!

sincze:
This option is excellent. RFXCOM and RFLINK are reporting windmeter values differently. 
Domoticz shows them both at km/h but with different values.
1 have one windmeter on the roof and both the receiver-devices. 
With this option I am now able to correct the RFLINK values by multiplying them x1.55
Many thanks.

PR #4302, opened 2020-08-23
Hardwaremonitor: Improve PI hardware detection

akamming:
Pi hardware detection is now based on the succesful execution of vcgencmd command instead of checking for specific models.

### DIFF
--- a/main/domoticz.cpp
+++ b/main/domoticz.cpp
@@ -123,13 +123,18 @@ std::string szWWWFolder;
 std::string szWebRoot;
 std::string dbasefile;
 
+#define VCGENCMDTEMPCOMMAND "vcgencmd measure_temp"
+#define VCGENCMDARMSPEEDCOMMAND "vcgencmd measure_clock arm"
+#define VCGENCMDV3DSPEEDCOMMAND "vcgencmd measure_clock v3d"
+#define VCGENCMDCORESPEEDCOMMAND "vcgencmd measure_clock core"
+
 bool bHasInternalTemperature=false;
-std::string szInternalTemperatureCommand = "vcgencmd measure_temp";
+std::string szInternalTemperatureCommand = "";
 
 bool bHasInternalClockSpeeds=false;
-std::string szInternalARMSpeedCommand = "vcgencmd measure_clock arm";
-std::string szInternalV3DSpeedCommand = "vcgencmd measure_clock v3d";
-std::string szInternalCoreSpeedCommand = "vcgencmd measure_clock core";
+std::string szInternalARMSpeedCommand = "";
+std::string szInternalV3DSpeedCommand = "";
+std::string szInternalCoreSpeedCommand = "";
 
 bool bHasInternalVoltage=false;
 std::string szInternalVoltageCommand = "";
@@ -471,7 +476,7 @@ void CheckForOnboardSensors()
 	//Check if we have vcgencmd (are running on a RaspberryPi)
 	//
 	int returncode=0;
-	std::vector<std::string> ret = ExecuteCommandAndReturn ("vcgencmd measure_temp",returncode);
+	std::vector<std::string> ret = ExecuteCommandAndReturn (VCGENCMDTEMPCOMMAND,returncode);
 
 	if (ret.empty()) {
 		// _log.Log(LOG_STATUS,"No vcgencmd detected (empty string)");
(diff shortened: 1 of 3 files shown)


kofec:
Hi,
What if vcgencmd is in different localization ? is It possible to do it more generic ?
```
which vcgencmd 
/usr/bin/vcgencmd
```

akamming:
> Hi,
> What if vcgencmd is in different localization ? is It possible to do it more generic ?
> which vcgencmd 
> /usr/bin/vcgencmd

Wat would cause vcgencmd to be in a different location? I think in Raspbian it is always in the same place….

kofec:
I think there is more than one Linux distribution :-)
In my case it is openwrt

akamming:
made it more generic.  now detects is vcgencmd is in the search path.

Still works on pi 4 and pi 3b+

@kofec can/will you test on openwrt?

gizmocuz:
@akamming , maybe for safety reasons don't use std::atof if you expect the value to be a non decimal but use
	float voltage = static_cast<float>(atof(tmpline.c_str()));

This also eliminates the try/catch warning
else for the catch use (...) to solve a compile warning

HardwareMonitor.cpp(284,34): warning C4101: 'e': unreferenced local variable
HardwareMonitor.cpp(321,34): warning C4101: 'e': unreferenced local variable
HardwareMonitor.cpp(358,34): warning C4101: 'e': unreferenced local variable

akamming:
@gizmocuz  I don't have this warning when i compile.  is this on windows?

Anyway: I change the code in latest commit to avoid use of std::atof..

gizmocuz:
@akamming , thanks! Yes this is on Windows.... It's great to compile on multiple platforms to eliminate compile issues ;)

main/domoticz.cpp
```diff
@@ -124,12 +124,12 @@ std::string szWebRoot;
 std::string dbasefile;
 
 bool bHasInternalTemperature=false;
-std::string szInternalTemperatureCommand = "/opt/vc/bin/vcgencmd measure_temp";
+std::string szInternalTemperatureCommand = "vcgencmd measure_temp";
```
kiddigital: @akamming , maybe separate the command from the arguments? Have a define at the top for the command and use this define here.

main/domoticz.cpp
```diff
@@ -124,12 +124,12 @@ std::string szWebRoot;
...
 
 bool bHasInternalClockSpeeds=false;
-std::string szInternalARMSpeedCommand = "";
-std::string szInternalV3DSpeedCommand = "";
-std::string szInternalCoreSpeedCommand = "";
+std::string szInternalARMSpeedCommand = "vcgencmd measure_clock arm";
+std::string szInternalV3DSpeedCommand = "vcgencmd measure_clock v3d";
+std::string szInternalCoreSpeedCommand = "vcgencmd measure_clock core";
```
kiddigital: same here.. use define for command

main/domoticz.cpp
```diff
@@ -468,46 +468,31 @@ void GetAppVersion()
...
-				szInternalV3DSpeedCommand = "/opt/vc/bin/vcgencmd measure_clock v3d";
-				szInternalCoreSpeedCommand = "/opt/vc/bin/vcgencmd measure_clock core";
-				bHasInternalClockSpeeds=true;
-
-				break;
-			}
+	//Check if we have vcgencmd (are running on a RaspberryPi)
+	//
+	int returncode=0;
+	std::vector<std::string> ret = ExecuteCommandAndReturn ("vcgencmd measure_temp",returncode);
```
kiddigital: @akamming , here you have the command hard-coded and are not using the defined variables. Could go wrong now if this works but variables are different.

main/domoticz.cpp
```diff
@@ -468,46 +468,31 @@ void GetAppVersion()
...
+			_log.Log(LOG_STATUS,"Hardware Monitor: Raspberry Pi detected");
+			//Core temperature of BCM2835 SoC
+			szInternalTemperatureCommand = "vcgencmd measure_temp";
+			bHasInternalTemperature = true;
 
+			//PI Clock speeds	
+			szInternalARMSpeedCommand = "vcgencmd measure_clock arm";
+			szInternalV3DSpeedCommand = "vcgencmd measure_clock v3d";
+			szInternalCoreSpeedCommand = "vcgencmd measure_clock core";
+			bHasInternalClockSpeeds=true;
```
kiddigital, line 497: Above are hard-coded again... use the defined variables I would think

kiddigital (commented):
@akamming , some suggestions :)

akamming:
@kiddigital sure no problem....just added the commit

kofec:
Hi, It is working on openwrt:
image

akamming:
@kofec  Tx! 
@gizmocuz all work is done then. i think it can be merged...

gizmocuz:
Thanks!!

PR #4307, opened 2020-08-28
Adding MoceanSMS as alternative notification SMS carrier

tzhongyan:
As titled, this PR allows user to send notification through MoceanSMS as an alternative SMS carrier.

Disclosure: I work for MoceanSMS

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -347,6 +347,7 @@ notifications/NotificationPushover.cpp
 notifications/NotificationPushsafer.cpp
 notifications/NotificationPushalot.cpp
 notifications/NotificationSMS.cpp
+notifications/NotificationMoceanSMS.cpp
 notifications/NotificationTelegram.cpp
 smtpclient/SMTPClient.cpp
 tcpserver/TCPClient.cpp
--- a/msbuild/domoticz.vcxproj
+++ b/msbuild/domoticz.vcxproj
@@ -753,6 +753,7 @@
     <ClInclude Include="..\notifications\NotificationPushover.h" />
     <ClInclude Include="..\notifications\NotificationPushsafer.h" />
     <ClInclude Include="..\notifications\NotificationSMS.h" />
+    <ClInclude Include="..\notifications\NotificationMoceanSMS.h" />
     <ClInclude Include="..\notifications\NotificationTelegram.h" />
     <ClInclude Include="..\push\FibaroPush.h" />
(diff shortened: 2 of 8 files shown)


gizmocuz:
@tzhongyan , Thank you for the PR!
I am not sure if we should merge this PR, sure it is working great, but I mean, I'm struggling with the fact that it is 'another' notification type, and we already have many.

The only benefit for sending out SMS messages is when the end-user does not have a 'data' connection....
(Could be because he/she is oversees, or can not effort it)
For the other 99% where end-users are connected, a push notification seems a better solution, and we support quite a few already (like PushOver)

If you would like to use MoceanSMS, you could use the HTTP Notification method and enter your details in the API call URL

tzhongyan:
Thanks for the review! The main reason this PR was made is because I was already using HTTP notification method for less important notification that can be sent over the internet/data connection; and SMS for higher priority incidents that needs my attention, in case I am in places with limited mobile internet connection. I thought we might be able to make use of this feature without needing a customised webhook receiver that HTTP notification method sends request to.

But now that you mentioned it, I realised that I might be able to write a script for differentiating between the two and call it using HTTP notification method. Whilst I respect your decision to not merge the PR, I do wish that I don't have to write another script for this =D

So, uhm... I guess I have to leave the decision back to you guys, please close this PR if this addon is not really applicable for the whole system

kiddigital:
Although it could be a nice addition (thx @tzhongyan ), I agree with @gizmocuz that it adds another notification system.

I think SMS is indeed a value for those situations where either the end-user does not always have a good data-connection (only 2G or 3G) or Domoticz does not have Internet connectivity. It would be great if there was a hardware module (with simcard) that could be connected locally to Domoticz so there is no need for Domoticz to be connected to the Internet and still be able to send important notifications.

The second take-away of your PR is the need to 'prioritize'. I think for Domoticz it would be great if there is a way to prioritize notifications and use a different notification channel/system per priority.

Maybe a 'simplification' of the notification system with priority options could be useful as most notification system depend on a HTTP push to an API (like your SMS provider) or Pub(/Sub) via MQTT or the like. The 'only' thing needed would be some kind of 'payload encoder' that encodes the Domoticz data into the payload for the specified target provider. (With a 'string replacement encoder' being the simplest form).

Is there an idea of the demand for this specific provider? Maybe ask on the forum?

gizmocuz:
Closing PR, thank you @tzhongyan for the PR. But one SMS is okey in the system.

PR #4311, opened 2020-08-30
 Patch making Plugwise Annatherm preset accept/synch with remote changes and allow events to react II

p-hagen:
Making sure the Anna Preset selector switch now notifies the Event system, allowing Events to react to these changes as well
using m_mainworker.PushAndWaitRxMessage.
Fixed general event handling
Added Logs line to better trace what happens
Added Usage log line to help new user with the Preset handling

- tabs vs  spaces etc

### DIFF
--- a/History.txt
+++ b/History.txt
@@ -16,7 +16,7 @@ Version 2020.xxxx (xxxx)
 - Implemented: RTL433, Switched to Json input
 - Implemented: The following sensor types now support RSSI: Distance Sensor, Moisture Sensor, Watt Meter, kWh Meter
 - Implemented: Websocket notification for secondary/sub devices
-- Implemented: ZWave, Legend and tooltips in Neighbor’s overview
+- Implemented: ZWave, Legend and tooltips in Neighborï¿½s overview
 - Updated: dzVents version 3.0.11 ( https://www.domoticz.com/wiki/DzVents:_next_generation_LUA_scripting#History )
 - Updated: eVehicles, added ODO meter, unlock/open alert, max charge level to Tesla module
 - Updated: OpenZWave configuration files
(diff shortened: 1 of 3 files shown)


gizmocuz:
@p-hagen , thank you for this PR and your work.
Question, instead of using PushAndWaitRxMessage and querying the database, why not use the function 

SendSwitch(const int NodeID, const uint8_t ChildID, const int BatteryLevel, const bool bOn, const double Level, const std::string& defaultname, const int RssiLevel /* =12 */)

p-hagen:
@gizmocuz, I have followed examples of coding for an selector switch in Domoticz code (onkyo, Daikin,PhilipsHue as fe examples)
 
As far as i could tell they all followed the PushAndWaitRxMessage then query the database.
Initially to update the switch with the correct options string as it gets created by the PushAndWaitRxMessage , then secondly to update SValue if the switch changed.

I know in the past i tried to work with SendSwitch when i first implemented the preset selector switch , but the only way i managed to get the selector switch to work reliable was with  updating the Database.

If you can point me to a working example of an selector switch using SendSwitch, I will happily change the code.


For all other switches i've added in the past, I use the regular helper functions as they are super easy

gizmocuz:
@p-hagen , Sorry about this, totally missed the 'sector switch' .... thought it was a normal switch ;)
Maybe in the feature we should make a helper function for this switch as well ;)

gizmocuz:
Thanks!!

p-hagen:
No problem...!  Thanks for validating I still haven't lost my code stealing..ehhh.. borrowing skills ;)
a helper function switch would indeed be handy making is easier for adding them to  ccp plugins..

awalsum:
Could any of this also have changed the function of "flame_state"? What I see happening now is that flame_state is switched on as soon as the heat pump is started, while in the anna.plugwise.net portal you see that the boiler is off. If I'm correct, flame_state used to be the command to also start the boiler (should the Elga not be capable of heating on its own).

If flame_state is now used to signal if the Elga heat pump is operational, is it possible to also signal if the boiler is required?
I had a look through the code (for as much as I can decipher it) and see that there is also a parameter called "boiler_state". I don't however see this created as a device when I add the Anna as new hardware. Modulation level of the heat pump would also be useful.

PR #4313, opened 2020-09-02
Made the selector switch inside able to be updated if the option string is different (plugwise plugin)

p-hagen:
Added code that checks the old option string vs the one currently used in the code for the selector switched in the plugwise plugin.
If they are different the new one will be applied, causing tit to adapt the new look.
This is done  to allow people to migrate to the new version without having to reload the plugin  messing up scripts etc.
Also opens the selector switch  to be "user configurable in a later stage - drop done vs button row)


gizmocuz:
Thank you for the PR.
Maybe you could write a global SendSelectorSwitch function (like CDomoticzHardwareBase::SendSwitch) that:
- Checks if the switch does not exist and create it
- If it exists, check if the value (option) is different, if yes, set it (like SendSwitch)

p-hagen:
Hey @gizmocuz,  I had the same idea to see if i could do that, and a quick review looks its best to follow sendCustomSensor  as that  looks to work identical as  the current code i have.
I raise a seperate pr for that change and then an other when update the plugwise plug  to use the new function?

gizmocuz:
@p-hagen , yes perfect, thanks !! But check the SendSwitch function to see how it detects the same value, and if so, skip it

PR #4315, opened 2020-09-02
create a new lua table otherdevices_idx_svalues

badzz:
This is to be able to deal with device that have the same names

### DIFF
--- a/main/EventSystem.cpp
+++ b/main/EventSystem.cpp
@@ -2654,7 +2654,15 @@ void CEventSystem::ExportDeviceStatesToLua(lua_State *lua_state, const _tEventQu
 			item.sValue : iterator->second.sValue);
 	}
 	luaTable.Publish();
-	
+
+        luaTable.InitTable(lua_state, "otherdevices_idx_svalues", (int)m_devicestates.size(), 0);
+        for (iterator = m_devicestates.begin(); iterator != m_devicestates.end(); ++iterator)
+        {
+                luaTable.AddString(iterator->second.ID, (iterator->first == item.id && item.reason == REASON_DEVICE) ?
+                        item.sValue : iterator->second.sValue);
+        }
+        luaTable.Publish();
+
 	luaTable.InitTable(lua_state, "otherdevices_idx", (int)m_devicestates.size(), 0);
 	for (iterator = m_devicestates.begin(); iterator != m_devicestates.end(); ++iterator)
 	{


gizmocuz:
@badzz , thank you for this PR!
Before merging this code that creates a new table (and could impact system performance if you have many devices),
have you also looked at our dzVents scripting language ?

rwaaren:
@badzz, 
Using the dzVents framework it is already possible to create a Lua script addressing same name devices by using their idx

```
return 
{
	on = 
	{
		devices = 
		{
			518, 519,
		},
	},
	
	
	execute = function(dz, device)
		dz.log('Device ' .. device.idx ..'(' .. device.name .. ') was updated', dz.LOG_INFO)
	end
} 
```

badzz:
Hi,
I did have a look but I do lua scripting for other projects and have a
quite a lot of library on it.
This table should not be more expensive than the other ones to generate.
Thanks

rwaaren:
> Hi, I did have a look but I do lua scripting for other projects and have a quite a lot of library on it. This table should not be more expensive than the other ones to generate. 

Because dzVents is 100% Lua there should be no reason why your libraries would not work with dzVents.
Probably you are right that the generation of this extra table is not more expensive then generating the other ones but it is still an extra table to solve an issue that is already addressed.

gizmocuz:
@badzz , could you work with the dzVents solution ?

Hedda:
Any more feedback on this?

badzz:
Hi,
I am testing .. it is just a shame that 

> @badzz,
> Using the dzVents framework it is already possible to create a Lua script addressing same name devices by using their idx
> 
> ```
> return 
> {
> 	on = 
> 	{
> 		devices = 
> 		{
> 			518, 519,
> 		},
> 	},
> 	
> 	
> 	execute = function(dz, device)
> 		dz.log('Device ' .. device.idx ..'(' .. device.name .. ') was updated', dz.LOG_INFO)
> 	end
> } 
> ```

Hi,
Thanks I am redoing my whole script with dzvent.
However your example does not suit my need, I want to get an another device than the one that triggered the event.
I am looking at the doc to find out
Thx
Edit : ok found it : domoticz.devices(idx) .. let's see

gizmocuz:
@badzz , Is it working out for you ? Can this PR be closed?

rrozema:
Getting a device by it's idx is done easily in dzVents.local myDevice =
domoticz.devices(20) returns you the device object for the device where idx
equals 20. Please come over to the forum where multiple users will be able
to assist you in using dzVents.

PR #4316, opened 2020-09-02
Mercedes Me: Made authentication process easier

kiddigital:
This Pull Request is for improving the user experience configuring the Mercedes Me module.

The previous process to get the module authenticated was cumbersome although it was needed only once. This patch should make it easier and less error prone.

### DIFF
--- a/hardware/eVehicles/MercApi.cpp
+++ b/hardware/eVehicles/MercApi.cpp
@@ -36,9 +36,9 @@ License: Public domain
 #define MERC_URL "https://api.mercedes-benz.com"
 #define MERC_API "/vehicledata/v1/vehicles"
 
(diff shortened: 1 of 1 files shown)


hardware/eVehicles/MercApi.cpp
```diff
@@ -89,17 +89,40 @@ CMercApi::CMercApi(const std::string username, const std::string password, const
 
 CMercApi::~CMercApi()
 {
+	m_fields = "";
```
gizmocuz: I don't think this is needed in the destructor of a class

kiddigital: removed

gizmocuz:
Thanks!

PR #4317, opened 2020-09-03
Created a new Helper function for the Selector switch SendSelectorSwitch

p-hagen:
new helper function  for a selector switch 
function void SendSelectorSwitch(const int NodeID, const uint8_t ChildID, const int sValue, const std::string& defaultname, const int customImage , const bool Dropdown, const std::string& LevelNames, const std::string& LevelActions, const bool nooff)
 
 _Dropdown_ controles the look of the selector switch - True for dropdown false for buttons
_LevelNames_ should contain the names of the selector in a string  separated by | example : "Off|Label1|label2|Label3"
_LevelActions_ should contain the corresponding values in a sting  separated by | example :  "00|10|20|30" 
_nooff_ controls showing the Off level in the selector - True hides the Off level  

sample usage 
SendSelectorSwitch(sAnnaPresets,  1, std::stoi(sPreset), PresetName.c_str(), 16, false, "Off|Home|Away|Night|Vacation|Frost","00|10|20|30|40|50", true);
		
 files updated DomoticzHardware.cpp +  DomoticzHardware.h

The change to Annathermostat.cpp is just the first implementation of this function

### DIFF
--- a/hardware/AnnaThermostat.cpp
+++ b/hardware/AnnaThermostat.cpp
@@ -15,7 +15,6 @@
 
 // Plugwise Anna Thermostat
 // Anna Sensors
-
 // Anna Switches
 
 #define sAnneBoilerState 8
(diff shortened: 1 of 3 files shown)


p-hagen:
Closing and will replace with version with comments.
Have one use case I think of adding. ( beter upgrade path)

PR #4318, opened 2020-09-03
eVehicle: Performance optimisation. Query home location only during startup

kiddigital:
Small PR to optimise SQLite query load. Instead of querying the settings every time, now only during Hardware setup the Home Location data of Domoticz is retrieved. This way saving a lot of query executions.

@MrHobbes74 , hopefully you are happy with this as well and can really test this better with the Tesla (I still don't have one ;) )


gizmocuz:
Thanks!

PR #4325, opened 2020-09-06
fix vcgencmd error on some distributions

akamming:
on some distributions which don't have the vcgencmd an error occurs "sh: 1: vcgencmd: not found". 

tx to @rwaaren  for noticing..

### DIFF
--- a/main/domoticz.cpp
+++ b/main/domoticz.cpp
@@ -123,6 +123,7 @@ std::string szWWWFolder;
 std::string szWebRoot;
 std::string dbasefile;
 
+#define VCGENCMDDETECTCOMMAND "which vcgencmd"
 #define VCGENCMDTEMPCOMMAND "vcgencmd measure_temp"
 #define VCGENCMDARMSPEEDCOMMAND "vcgencmd measure_clock arm"
 #define VCGENCMDV3DSPEEDCOMMAND "vcgencmd measure_clock v3d"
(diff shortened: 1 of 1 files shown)


gizmocuz:
@akamming , thank you for the PR.
This patch indeed silences the warning, but it might not be bad to notify a user that this command is missing ?

rwaaren:
> , but it might not be bad to notify a user that this command is missing ?

The vcgencmd seems only available for Raspberry systems. Would there be any value for other systems to show this warning?

akamming:
i agree with @rwaaren : issue was found on an intel nuc , so there would be no use to warn the user...

alternative would be to design an alternate way to detect for pi hardware and only then do a check on vcgencmd. but this introduces a lot of complexity only for the use case that you have a messed up os: normally vcgencmd is always available on a pi. and if you have a messed up os, then you can expect more problems. I would not adapt the code to check for those conditions...

gizmocuz:
@akamming , yes that was initially also my thought, why does that not work anymore...

In the past we had some code before this check:

```
#if defined(__FreeBSD__)
	infile.open("/compat/linux/proc/cpuinfo");
#else
	infile.open("/proc/cpuinfo");
#endif
	if (infile.is_open())
	{
		while (!infile.eof())
		{
			getline(infile, sLine);
			if (
				(sLine.find("BCM2708") != std::string::npos) ||
				(sLine.find("BCM2709") != std::string::npos) ||
				(sLine.find("BCM2711") != std::string::npos) ||
				(sLine.find("BCM2835") != std::string::npos) 

				)
			{
 ... Check for raspberry stuff

```

I added that code back in:

https://github.com/domoticz/domoticz/commit/2ca3a805d73daa27b5a7ff3769ba2e4b00a3402c

Hope that this will solve the issue ?

akamming:
@gizmocuz but then we are back on the original issue that only model 2708,2709,2711&2835 are supported, while there are more pi models.  For my use case it's fine (i have a 3b+ and a 4b, which are covered) ..  but if in the future a new model is launched it might also not work.

maybe better to check if  "/proc/device-tree/model" or "/sys/firmware/devicetree/base/model" contains the text "Raspberry Pi"?

gizmocuz:
@akamming , that is correct, but then we can add the new ID of a new raspberry Pi.
Because to be notified that vcgencmd is (for me) a Plus... because when you enable the hardware monitor, and you are on a raspberry pi, you would like to know why it does not work...
Silencing the error would raise some confused questions

akamming:
ok understand what you are saying. Closing this PR.  You might want to consider  to uncomment lines 505 and 510 and give a (admin)user friendly error there, that would then be even better...

gizmocuz:
Done

akamming:
Well done!

PR #4326, opened 2020-09-06
Fix PHP

lokonli:
PHP did not work with query strings.

The fix in this PR probably only works on Linux systems, because of setting the environment variable:

   fullexecmd = "export QUERY_STRING='"+szQueryString + "' && " + fullexecmd;

I can't test on Windows.

I also don't know whether this is the cleanest way to set the PHP variables, but it works on my system.

### DIFF
--- a/webserver/fastcgi.cpp
+++ b/webserver/fastcgi.cpp
@@ -347,7 +347,7 @@ bool fastcgi_parser::handlePHP(const server_settings &settings, const std::strin
 		fcgi_params[rName] = CURLEncode::URLEncode(ittHeader->value);
 	}
 #ifdef WIN32
-	fullexecmd = "SET QUERY_STRING="+szQueryString + " & " + fullexecmd;
+	fullexecmd = "SET QUERY_STRING=\""+szQueryString + "\" & " + fullexecmd;
 #else
 	fullexecmd = "export QUERY_STRING='"+szQueryString + "' && " + fullexecmd;
 #endif
--- a/webserver/fastcgi.cpp
+++ b/webserver/fastcgi.cpp
@@ -346,8 +346,12 @@ bool fastcgi_parser::handlePHP(const server_settings &settings, const std::strin
 
 		fcgi_params[rName] = CURLEncode::URLEncode(ittHeader->value);
 	}
-
+#ifdef WIN32
+	fullexecmd = "SET QUERY_STRING="+szQueryString + " & " + fullexecmd;
+#else
 	fullexecmd = "export QUERY_STRING='"+szQueryString + "' && " + fullexecmd;
+#endif
+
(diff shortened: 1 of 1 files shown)


gizmocuz:
@lokonli , thank you for the PR. Yes this command does not work for Windows.
You could add a #ifdef to differentiate between windows/non-windows systems.
But could you give an example of a query string that does not work, i could try to simulate this for Windows

lokonli:
To test: Create the following scripts in the www folder with name test.php:

```
<?php
print( $_SERVER["QUERY_STRING"]);
print("<br>");
print_r($_GET);
exit(0)
?>
```


Then request:

```
http://<domoticz-ip:port/test.php?test=123&moretexthere
```

This should give output like:
```
test=123&moretexthere
Array ( [/home/pi/domoticz/www/test_php] => [test] => 123 [SERVER_SOFTWARE] => Domoticz [SERVER_NAME] => localhost [SERVER_ADDR] => :: [SERVER_PORT] => 8080 [REMOTE_ADDR] => 192.168.178.11 [QUERY_STRING] => test=123 [moretexthere] => [HTTP_HOST] => test.pi.home:8080 [HTTP_CONNECTION] => keep-alive [HTTP_PRAGMA] => no-cache [HTTP_CACHE_CONTROL] => no-cache [HTTP_UPGRADE_INSECURE_REQUESTS] => 1 [HTTP_USER_AGENT] => Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/85.0.4183.83 Safari/537.36 [HTTP_ACCEPT] => text/html,application/xhtml xml,application/xml;q=0.9,image/avif,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3;q=0.9 [HTTP_ACCEPT_ENCODING] => gzip, deflate [HTTP_ACCEPT_LANGUAGE] => en-NL,en;q=0.9,nl-NL;q=0.8,nl;q=0.7,en-US;q=0.6 [HTTP_COOKIE] => _ga=GA1.2.846285750.1591806061 )
```

The first line in the response echos the complete query string.
After that a dump of the $_GET variable follows.
This should contain the key 'test' with value '123'

lokonli:
On windows it probably should be something like:
```
fullexecmd = "SET QUERY_STRING="+szQueryString + " & " + fullexecmd;
```

gizmocuz:
@lokonli , thank you for the addition for WIN32, but please test the patch

For WIN32 could you change the patch to include "quotes" like:

			fullexecmd = "SET QUERY_STRING=\"" + szQueryString + "\" & " + fullexecmd;

Not sure if this is needed by Non-WIN32, but I assume you are able to test this

lokonli:
Tested on Debian [ok]
Also now Dashticz works with the Domoticz webserver, which in fact was my goal of this action.

gizmocuz:
Thanks!!

PR #4327, opened 2020-09-06
Added Helper functions for selector switch

p-hagen:
Added two functions :
SendSelectorSwitch:   Use case:  Simplify the way to create and update a selector switch similar to other existing helper functions.  For detail see comments for the function

MigrateSelectorSwitch: function to test and migrate the names and actions of a selector switch. 
Usecase: allow developers to update or add/remove level actions and names on existing   selector switches
This is useful if your plugin  needs to add or remove actions from a selector switch, leaving the look and feel as the current version.   Main reason is to stop people to  require to  remove and add  hardware plugins to get the new selector switch

Intended to be run  once at init or at begin of do work() for usage  see comments for the function
The changes in the Plugwise  plugwise Annatherm.ccp   are the  to make use of the above functions.

This wil allow people to change from stable or earlier beta's to the latest beta without selector switch issues

### DIFF
--- a/hardware/AnnaThermostat.cpp
+++ b/hardware/AnnaThermostat.cpp
@@ -15,7 +15,6 @@
 
 // Plugwise Anna Thermostat
 // Anna Sensors
-
 // Anna Switches
 
 #define sAnneBoilerState 8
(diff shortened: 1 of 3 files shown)


gizmocuz:
Thanks!!!

Issue #3764, opened 2019-11-29
Tado hardware mixes up thermostats setpoints


bxlouis:
Hi,

I am reporting an issue with the Tado Hardware plugin which behaves abnormally with the Thermostats SetPoints.

I am running Domoticz v4.10717 on a Raspberry Pi 2 (Raspbian Stretch). 
The issue appeared after I added a 4th thermostat to the Tado application.
I tried to disable/delete/add the plugin several times, remove and add again the thermostat from the Tado application. The issue is persisting.

The issue is that when changing the SetPoint of one particular thermostat, it changes the SetPoint of another thermostat instead. Please find a video of the issue here: https://youtu.be/Igpc_2yKhD4

Along with the logs from Domoticz at the same time:

> 2019-11-29 23:18:11.271 (Tado) Thermostat (Home Cuisine Setpoint)
> 2019-11-29 23:18:11.276 (Tado) Temp + Humidity (Home Cuisine TempHum)
> 2019-11-29 23:18:11.281 (Tado) General/Percentage (Home Cuisine Heating Power)
> 2019-11-29 23:18:11.263 Debug: SQLH UpdateValueInt Home Cuisine Setpoint HwID:9 DevID:000002E Type:242 sType:1 nValue:0 sValue:20.00
> 2019-11-29 23:18:11.275 Debug: SQLH UpdateValueInt Home Cuisine TempHum HwID:9 DevID:303 Type:82 sType:5 nValue:0 sValue:20.2;51;1
> 2019-11-29 23:18:11.280 Debug: SQLH UpdateValueInt Home Cuisine Heating Power HwID:9 DevID:00000133 Type:243 sType:6 nValue:0 sValue:0.00
> **2019-11-29 23:18:27.541 Tado: SetSetpoint() called with idx=46, temp=17.000000
> 2019-11-29 23:18:27.541 Tado: CreateOverlay() called with idx=46, temp=17.000000, termination type=TADO_MODE**
> 2019-11-29 23:18:29.330 (Tado) Thermostat (Home Chambre Setpoint)
> 2019-11-29 23:18:29.335 (Tado) Temp + Humidity (Home Chambre TempHum)
> 2019-11-29 23:18:29.346 (Tado) General/Percentage (Home Chambre Heating Power)
> **2019-11-29 23:18:29.325 Debug: SQLH UpdateValueInt Home Chambre Setpoint HwID:9 DevID:0000002 Type:242 sType:1 nValue:0 sValue:17.00**
[...]

PR #4331, opened 2020-09-08
Fix Tado Home/Zone limit and Setpoint mix-up

LarsenDX:
Fixes Tado.cpp's current limitation of 1 home and 3 zones at most. (It pushed IDs > 255 into unsigned char.) Moreover fixes the side-effect of setpoint mix-ups reported in issue #3764 . How to upgrade: It is recommend to delete all Tado switch and setpoint devices before enabling the Tado hardware.

This code was successfully tested with 1 home and 8 zones.

### DIFF
--- a/hardware/Tado.cpp
+++ b/hardware/Tado.cpp
@@ -90,16 +90,19 @@ bool CTado::WriteToHardware(const char * pdata, const unsigned char length)
 	if (pCmd->LIGHTING2.packettype != pTypeLighting2)
 		return false;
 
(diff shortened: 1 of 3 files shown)


gizmocuz:
@LarsenDX , thank you for the PR. Is it possible to undo all changes to the 'extern' folder, I think you checked in to much

gizmocuz:
@LarsenDX , You could leave a PR open, make changes locally and commit them. Then they will be updated in the PR as well ;)

PR #4332, opened 2020-09-09
Fix for Comm5 MA-4200 hardware

correa:
Sometimes a CRC error can occur due to problem with signed char calculation.

### DIFF
--- a/hardware/Comm5Serial.cpp
+++ b/hardware/Comm5Serial.cpp
@@ -262,9 +262,11 @@ void Comm5Serial::ParseData(const unsigned char* data, const size_t len)
 		case STFRAME_CRC2:
 			frame.push_back(data[i]);
 			frameCRC = crc16_update(frameCRC, 0);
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thank you for the PR, does this mean the line should read now:

`readCRC =  (uint16_t)(frame.at(frame.size() - 2) << 8)  | data[i];`

or

`readCRC =  (uint16_t)(data[i - 1] << 8)  | data[i];`

correa:
I think readCRC = `(static_cast<uint16_t>(frame.at(frame.size() - 2)) << 8) | data[I]` should do, but I will run tests on the device to be sure.

gizmocuz:
@correa , please either choose frame.size()-x , or i or i-1 , now it is a mix

correa:
Updated and tested with hardware.

gizmocuz:
Thanks!!

PR #4333, opened 2020-09-09
fix PHP handling of binary content

lokonli:
The current PHP implementation only supported text responses. Binary content, like images, didn't work.

I've replaced the string buffer with a dynamic resizing vector to read the binary data.

To test:

Create the following php file domoticz/www/load.php:

```
<?php
function parseHeaders( $headers )
{
    $head = array();
    foreach( $headers as $k=>$v )
    {
        $t = explode( ':', $v, 2 );
        if( isset( $t[1] ) )
            $head[ trim($t[0]) ] = trim( $t[1] );
        else
        {
            $head[] = $v;
            if( preg_match( "#HTTP/[0-9\.]+\s+([0-9]+)#",$v, $out ) )
                $head['reponse_code'] = intval($out[1]);
        }
    }
    return $head;
}


header("Access-Control-Allow-Origin: *");
header("Cache-Control: no-store, no-cache, must-revalidate, max-age=0");
header("Cache-Control: post-check=0, pre-check=0", false);
header("Pragma: no-cache");
$reply = file_get_contents($_SERVER["QUERY_STRING"]);
$head = parseHeaders($http_response_header);
if (isset($head["Content-Type"])) {
    header("Content-Type: ".$head["Content-Type"]);
}
echo $reply;
exit();
?>

```

Then load an image via load.php in your browser:
http://domoticz-ip:port/load.php?https://www.domoticz.com/wiki/images/8/82/Domoticz.png

The Domotics logo should be visible.

### DIFF
--- a/webserver/fastcgi.cpp
+++ b/webserver/fastcgi.cpp
@@ -120,10 +120,9 @@ struct _tFCGI_UnknownTypeRecord {
 	_tFCGI_UnknownTypeBody body;
 };
 
-std::string ExecuteCommandAndReturnRaw(const std::string &szCommand)
+std::vector<char> ExecuteCommandAndReturnRaw(const std::string &szCommand)
 {
-	std::string ret;
-
+	std::vector<char> myData;
 	try
 	{
 		FILE *fp;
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thank you for the patch, but... by default this is not possible:

Warning: file_get_contents("https://www.domoticz.com/wiki/images/8/82/Domoticz.png" ): failed to open stream: Permission denied in S:\Domoticz\www\load.php on line 25

in php.ini:
allow_url_fopen = On

By default (Windows PHP FastCGI) there is no php.ini, you have to copy the example production/development to php.ini

Is it possible to test this under Windows ? (I have not tested this on Linux)

lokonli:
I've tested on Linux, and it's working.

I'll try to setup a Windows test environment as well...

PR #4334, opened 2020-09-09
Fix Tado Zone/Home limit, Setpoint mix-up

LarsenDX:
Fixes Tado.cpp's current limitation of 1 home and 3 zones at most. (It pushed IDs > 255 into unsigned char.) Moreover fixes the side-effect of setpoint mix-ups reported in issue #3764 . How to upgrade: It is recommend to delete all Tado switch and setpoint devices before enabling the Tado hardware.

This code was successfully tested with 1 home and 8 zones.


LarsenDX:
Cleaned up PR. Sorry for the back and forth. Still pretty new to github.

gizmocuz:
Thanks!

Issue #4336, opened 2020-09-10
Tado New Feature: Open Window Detection


LarsenDX:
This issue tracks the addition of the feature Open Window Detection to domoticz/hardware/Tado. 

The Tado Smart Thermostat can detect an open window in a zone and alert the user. The Domoticz Tado hardware, at start-up, checks each zone and records if the zone supports Open Window Detection. For each supported zone a new switch is created. ex:  Home Kitchen Open Window Detected. 
The switch gets pulled high when the API reports an open window. The switch get pulled low when the API no longer reports an open window. This is a read-only switch.

gizmocuz:
@LarsenDX , thank you for the PR. It's been merged!

PR #4337, opened 2020-09-10
Tado Feature. Add support for per-zone Open Window Detection

LarsenDX:
Adds support for Open Window Detection to Tado.cpp . See also issue #4336


gizmocuz:
Thanks!!

PR #4341, opened 2020-09-11
TTNMQTT: Optimise reading location at startup instead of each cycle

kiddigital:
And some code (style) cleanup.

The location settings of Domoticz (settings screen) are now only read at startup instead of each cycle. Saving a lot on unneeded SQL (SQLite) queries.


gizmocuz:
Thanks!!

PR #4342, opened 2020-09-11
GraphLog autorefresh feature

michahagg:
The GraphLog now supports autorefresh. Refresh is triggered by the 'device_update' event.

In normal view (no zoom), upon each update event, it accepts and presents all datapoints available. When new datapoints are added, they are added to the view.

In zoom view, two modes are available for both left and right zoom edges: sticky and relative-to-chart-edge.
Sticky: the zoom edge is kept on the initial date and time; the chart in the zoom window does not change.
Relative-to-chart-edge: the zoom edge is kept the same distance from the chart edge; the chart _slides_ through the zoom window.
The relative-to-chart-edge mode is default. Hold down the CTRL-key when starting and/or finishing a zoom window selection to activate sticky mode.


gizmocuz:
@michahagg , thank you for this great patch !!

gizmocuz:
@michahagg , using the default theme I cannot see the checkbox at the bottom. I can see the text (translation), not the checkbox
(Seems the opacity is 0?, if I disable this I see the checkbox):
image

But maybe it should be on by default ? (Not having this option?)

Could you also make this work for the Temperature (other) charts ?

michahagg:
@gizmocuz thanks for merging 😊

I'll fix the checkbox issue soon. I'm already in the code to update the other charts as well. Except those are implemented differently and I'd prefer to fix that first.

gizmocuz:
@michahagg , thank you, much appreciated.
Could you have a look at this issue:

https://github.com/domoticz/domoticz/issues/4356

I think it is related to your chart modification code...

PR #4343, opened 2020-09-12
MercApi: Request one resource per cycle to make module more responsive.

kiddigital:
PR to make the module more responsive. Previously multiple resources were requested using multiple API calls. Each API call can takes seconds (or many seconds before timing out) adding up to total processing time of tens of seconds leading to Domoticz (or at least the eVehicle/MercApi module) being less responsive.


gizmocuz:
Thanks!

PR #4344, opened 2020-09-13
Tado: fix compiler warning. Switched logger to HardwareBase

kiddigital:
Fixed compiler warning. 

Also (as I was editing anyway) switched logger from logger class to DomoticzHardwareBase class as that seems to be the (new?) preferred way of logging now.


kiddigital:
@gizmocuz , should/could we also remove the dependency (include) to the logger class? Tried it and it compiles fine, but wasn't sure if that is a goal and how to test if everything still works as it should.

gizmocuz:
@kiddigital , thanks! Yes this also logs the hardware class name... ;)

gizmocuz:
@kiddigital , just did ;) thanks!

PR #4346, opened 2020-09-14
Enable preset migration even on previous versions

p-hagen:
Found there version out there where the preset selector switch has unit 0 and some that have 1.   Since there is only one selector switch in this plugin. they will be migrated to have  unit 0 as value.
This will enable the migrations to complete successfully.   
( run once per start up and migration should happen only once)

Added version - plug in now reports the version. (1.0.1)


gizmocuz:
0 is ... nothing... In almost all plugins we use 1 as initial unit... I would leave it like this ?

p-hagen:
Understood... will close  the PR...  and change the migration  and change  the ones that are 0 to 1 so they are something instead of.. nothing :) and then migrate.
then fire the PR again

PR #4347, opened 2020-09-14
 fixing unit  for preset selector... setting them to 1 for those  that where set to 0.

p-hagen:
this version  makes sure  the selector  will have a unit  # 1 from now on and migrate them

also has a version  that we van check again


p-hagen:
Note: this completes the upgrade of the plugin, should now also migrate older stable clients as well as the latest beta.

PR #4348, opened 2020-09-15
update to Plugwise scene selector and migration process for all stable and beta versions

p-hagen:
Replacement for previous PR that had a typo.

This patch will correctly update the Scene preset switch to the latest version that is in used by the plugwise product suite.
It will make sure all unit are  having a correct value   which is a requirement for further planned upgrades.

Removed some old code that was commented out to get a cleaner code base.
Will run a migration check at initial start of the hardware,  and only migrate once
Has run for 18 hour on my mac and pi test platfrom


gizmocuz:
@p-hagen , once you create a PR, you can keep pushing changes to it (git commit/push) like normal. the PR will update
No need for a close and a new PR

gizmocuz:
Thanks!

PR #4350, opened 2020-09-17
Add "from" field for Clickatell SMS notification

ojozef:
Hello

The only way to use Clicaktell in USA is two-way communication. This is restricted by law. So in this pull request I'm adding optional "from" field. 
Another change in this commit is to use GET method instead of POST. I don't know why but I was not able to make it work with POST (I think in general this functionality currently is not working). Also fixed copy to send texts to only one number. Currently used method (http) supports only one:
See https://docs.clickatell.com/channels/sms-channels/sms-api-reference/#operation/sendMessageHTTP

There are couple more things to improve:
- use REST api to send bulk messages
- fix error handling. Right now we have positive message regardless of result.

Please consider this changes. I will be happy to fix above things as well. 

Thanks!


gizmocuz:
Thank you for the PR!!
Is it possible you could have a look here:

https://docs.clickatell.com/channels/sms-channels/sms-api-reference/#operation/sendMessageREST_1

I would prefer to keep using a HTTP Post when possible

ojozef:
Switched implementation to use REST api. 
Now it would be possible to send texts to multiple recipients!

gizmocuz:
Thanks !!!!

Issue #4277, opened 2020-07-29
ZWave powerplug sometimes reports wrong values


RDols:
I have quite some powerplugs, most of them are Neo Coolcam plugs. Yes the cheap china stuff. And a few Fibaro plugs. Start using them about 3 years ago, a graduadly got more of them. I always had problems that sometimes i had usages of MWh. Even with todays beta this happens. Not only with the Neo plugs, also with the Fibaro plugs. In the past i just deleted those values. But with the increasing number of plugs this is not fun anymore. So i decided to see if i can find out why. Got the latest Domoticz and OpenZWave source from github and made a windows debug build.  (ow, normally i run on a raspberry pi). Until now i identified 3 problems.

1. One of my more recent plugs is broken. It simpy reports wrong values sometimes.
2. During startup and "refresh node info" a value of "0" sometimes occur.
3. During startup and "refresh node info" the scaleMultiply is sometimes not set yet (still has the default value of 1) when a (valid) value is received.

I will try to fix 2 and 3. but will take a little time, i'm limited in time.

I pretty sure there is more, but not sure if it's domoticz or the OpenZWave stack.

Don't know if more people have the same issues?

SargonofAssyria:
I have sometimes that issue too, but only (as far as know) with the newer Coolcam plugs.

RDols:
I guess you have extreme negative values (e.g. -21474832.82)?
(hint: Add 21474836.48 and probably you get the correct value).

That's looks like a issue with the Newer CoolCam plugs (0x1027 / 0x0200).

https://github.com/OpenZWave/open-zwave/issues/2333

SargonofAssyria:
Tnx, I look into that.

gizmocuz:
Maybe we could write a sanity check for this in the code ?

sincze:
Does this illustrate the issue? 
![image](https://user-images.githubusercontent.com/5776333/89915353-88b19a80-dbf6-11ea-9663-d5ec612cbfc0.png)

RDols:
Yes thats typically the new neo plug. You see extreme high day to day 
value. Also the spikes.

Look in the openzwave issue i posted earlier. I put some more information 
there.

Still thinking how to solve this on a acceptable way.

Buy in short:
The plug has about a 2.7x too high kwh value. 50% of the power reports are 
erroneous. RF is very poor.

Op 11 augustus 2020 17:18:16 schreef sincze <notifications@github.com>:

RDols:
[...]

PR #4355, opened 2020-09-20
#4277 Kwh filter for ZWave devices

RDols:
PR addresses the folowing issues:
- General filter for KWh devices.
- Fixed a startup glitch with scaleMultiply
- Specific filter for the new Neo coolcam plugs.


gizmocuz:
I think the only check needs to be in 'SendKwhMeter' if the power reported is above the threashold (4000) then we skip this value
If your 'new' neocoolcam always report 3 times to high, you better throw it out!

RDols:
I agree on the neocam specific filters. These plugs have multiple showstopper bugs, including full hang-ups, and spontaneously output switches. I personally already have these put them in my square permanent archive.

But faulty usage reports are not only at NeoCam plugs. These are also present on the Fibaro plugs, fibaro double relais and on Aeon nano. Looking in the data table of domoticz i see all power reporting devices having faulty values. I traced back faulty values in domoticz and openzwave, but it's already in the serialport report. I even have seen values from a different plug inside other reports. I suspect the Aeon zwave stick has some bugs.

I suggest also keep the usage filtering in.

gizmocuz:
@RDols , Is it OK for you if I have a go with a Fix ? I would like to make the Max. Power a configuration option as well
I think the counter values are also sometimes wrong

RDols:
@gizmocuz Yes that's ok with me.

gizmocuz:
I made a new commit, where you can now specify the maximum power in the settings:

image

For most households this will be ~3600, so 4000 should be OK, but I have to see a appliance that can measure above 3600 Watt yes.
New beta's are building and should be available within the hour

gizmocuz:
https://github.com/domoticz/domoticz/commit/f79cf66245cc27de722cf81cbe49481e0ee14490

https://github.com/domoticz/domoticz/commit/bee66dffacf7690a425cc37d93bcf69dc6930de9

gizmocuz:
@RDols , Did you found some time to test the new version ? Is it working and can we close this issue ?

RDols:
@gizmocuz I'm currently start trying this out. Takes a one or two days to have results.

RDols:
The new beta crashes at startup in a plugin. I didn't update my beta for a while now. I will take a detailed look later today.

gizmocuz:
@RDols , did you manage to have a look if this issue is solved?

Issue #4301, opened 2020-08-23
AirconWithMe implementation


RDols:
@gizmocuz 
Recently i had two airco units installed. These are equiped with an AirconWithMe wifi module. There are some scripts available, but doens't do the job for me. And some requests on the forum. It has a (relativly) well documented local web interface. 

I implemented this in my (local) fork as a hardware type. After a little clealup of the code i would like to create a pull request for this.

gizmocuz:
Great ! Hoping for more sunny days here ;)

RDols:
@gizmocuz 

I could be there is a simular request for this: #4193 
Depending on the outcome there if it's indeed the same, and if the WMP protocol implementation is better i will create a pull request, or change it (e.g. renamening it to "Intesis Airco Control".

LarsenDX:
@RDols Thank you! I have AirconWithMe devices attached to my Mitsubishi Heavy (MHI) AC in the house. I can help test the code. Is a pull request planned for this? About WMP, as far as I can tell, AirConWithMe devices only offer the JSON API (HTTP port 80) and do not implement the WMP protocol. I port scanned my devices and only find port 80 to be open.

RDols:
@LarsenDX There will be a PR. Don't known when. It's hanging on 4 minor points:
1. Simularity with Intesis devices. (by now i known that they are not the same, so this bullet point can be checked.
2. need to fix a minor issue with the configuration. 
3. Want to look if its possible to add default values for for new devies (username and password)
4. Me. I'm a little limited in time.

Code is a branch on my account, but you need to compile it yourself.

LarsenDX:
@RDols Good stuff. I'll check out your branch and give it a spin.

LarsenDX:
>     1. Simularity with Intesis devices. (by now i known that they are not the same, so this bullet point can be checked.
Does your code not account for that by calling and parsing the result of getavailabledatapoints?

RDols:
The getavailabledatapoints is called once (for every instance). The description is hard coded though. I'm not using "http://IpAddress/js/data/data.json" for that. I found this call just recently, and would not add much. 

It looks like the Intesis Wifi modules have the HTTP port blocked. And the Intesis WMP modules don't support this API at all.

LarsenDX:
Sure. I read through vendor comments in #4193. Looks to me like it makes sense to have 3 different hardwares in domo.
[...]

PR #4357, opened 2020-09-20
#4301 added AirconWithMe

RDols:


LarsenDX:
Successfully built and tested here on a MHI AC equipped with the AirconWithMe Wifi module.

gizmocuz:
@RDols , thank you for this great addon!

Is it possible to move all structs/enums in the header file underneath the class declaration, so they are private to the class:

```
class CAirconWithMe: public CDomoticzHardwareBase
{
```

Do we need to let the user specify a poll internval ? This will only confuse people, does it need to be polled that frequently ?
Could we define the poll interval in the .cpp file ?

LarsenDX:
The default poll interval in the UI of the device is 30s.
Screenshot 2020-10-02 at 11 07 22

gizmocuz:
@RDols , great, is it possible to hard code this with a 'define' in the .cpp file ?
I just merged a very large PR, and there seems to be a conflict in the Hardware.html file, sorry

LarsenDX:
> @RDols , thank you for this great addon!
> 
> Is it possible to move all structs/enums in the header file underneath the class declaration, so they are private to the class:
> 
> ```
> class CAirconWithMe: public CDomoticzHardwareBase
> {
> ```
> 
> Do we need to let the user specify a poll internval ? This will only confuse people, does it need to be polled that frequently ?
> Could we define the poll interval in the .cpp file ?

@gizmocuz I believe @RDols is really busy elsewhere at the moment so let me see if I can button this up.

I propose to move the declaration inside the class and leave the definition outside for readability.  Or would you rather have it all inside the class? Whatever fits the style of  C++ in domoticz best is fine with me.

class Container {
    struct Inner; // Declare, but not define, the struct.
};

struct Container::Inner {
   /* Define the struct here. */
};

gizmocuz:
@LarsenDX , thanks for picking this up.
No everything inside the struct (also define it), as it is private to this class
I can also do it, takes me about 5 minutes.... let me know what you prefer

LarsenDX:
@gizmocuz Alright, I've moved all remaining structs and enums into CAirconWithMe. I can't push those changes as I don't have write permissions. What am I missing?

gizmocuz:
@LarsenDX , maybe you could make a new branch and create a PR ? I think the changes of RDols will be in there as well in that case

LarsenDX:
Thanks, @gizmocuz . Please see #4383

Issue #4353, opened 2020-09-19
eVehicle: (Connected) Mercedes Me new OAuth endpoint


kiddigital:
The (connected) Mercedes Me API's and the OAuth endpoint have been changed.

The old version is still operational, but the new (more secure?) OAuth service is available and also the new endpoints (v2) of the different (BYOCAR) API's that use this new OAuth service as well. See [this link](https://developer.mercedes-benz.com/oauth_migration) for more info.

Unfortunately, a lot of the actual API's are still down (see the [status page](https://developer.mercedes-benz.com/status)) so both the old and new situation do not function properly yet.

I am in the process of creating the update, but it is hard to complete when most pieces are down.

kiddigital:
See PR #4362 

Received response from Mercedes that most issues should be resolved and indeed the (new) API's and OAuth2 endpoint seem to work 9 out of 10 times (still some random 401 errors but that is a known issue).

PR #4362, opened 2020-09-23
eVehicle: (Connected) Mercedes Me new OAuth endpoint (#4353)

kiddigital:
This PR makes the switch to the new OAuth2 endpoint and corresponding (BYOCAR) API endpoints (v2).

See also issue #4353


gizmocuz:
Thanks!!

PR #4366, opened 2020-09-28
Update OpenWeatherMap.cpp

Olifant1990:
Renamed rain (snow) to precipitation and added amount of snow together with rain.


kiddigital (approved):
Hi @Olifant1990 , I like your improvement. Indeed better to accumulate rain and snow in the same way as done with the forecasts. Thx.

Olifant1990:
Also changed behaviour of wind sensor type and wind_chill temperature.
Chill temperature was same as current temperature and sensor type could have changed between requests.

gizmocuz:
Thanks!!

PR #4371, opened 2020-10-02
Additional changes to previous PR

michahagg:
@gizmocuz Thanks for merging! No problem to add the fixes in this PR.

Contains:
- The necessary newlines.
- Instructions for the unit tests.
- Unit test folder moved to test folder. (I saw you did exactly the same ;-) )


gizmocuz:
Thank you again !!

gizmocuz:
@michahagg , I do have some issues with my Temp+Hum+Baro And Temp+Baro temperature log, the day values are gone:

image

gizmocuz:
The reason seems to be that it is calling the JSON API for a 'counter' instead of a temp:

`http://127.0.0.1:8080/json.htm?idx=1&range=day&sensor=counter&type=graph`

the month/year are called like:

`http://127.0.0.1:8080/json.htm?idx=1&range=year&sensor=temp&type=graph`

gizmocuz:
I think the function 'sensorTypeForDevice' is not correct.
A device has a Type and SubType that we could use for this.

I made a small patch that solves it for the temperature (and maybe feature) type charts, hope it is done OK?

Patch 1:
https://github.com/domoticz/domoticz/commit/70c3809cd0009c7ed2e418112ee364b32da22a12

Patch 2:
https://github.com/domoticz/domoticz/commit/12113bc63265fbf270e132939871884b30e6cc11

gizmocuz:
@michahagg , I do see some Y Axis issue...

Initially the chart looks great:

image

I see 3 API calls being fired:

http://172.16.0.117:8080/json.htm?idx=1&range=day&sensor=temp&type=graph
http://172.16.0.117:8080/json.htm?idx=2&range=month&sensor=temp&type=graph
http://172.16.0.117:8080/json.htm?idx=3&range=year&sensor=temp&type=graph

A second later, another API call is made:
http://172.16.0.117:8080/json.htm?idx=4&range=day&sensor=temp&type=graph

Now the chart looks like:
image

michahagg:
@gizmocuz thanks for testing the change! I'm sorry I didn't find these issues myself. Maybe in the future it would be better to test on a separate branch first before merging with development.

And thanks for the fixes! I will look into it later today or tomorrow. Hope you don't mind.

gizmocuz:
@michahagg , no worries, no problem at all!
I do notice some double checkboxes, for example in the settings:

image

Maybe this has to do with the upgrade to a newer JQuery or perhaps AngularJS

Normally those first checkboxes should be replaced with the one next to it

michahagg:
hi @gizmocuz , about the checkboxes, that problem is solved. It should work like this. In the default theme, the label is automatically replaced with the fancy toggle switch and the checkbox is hidden by setting opacity to 0 in the css. It only works when a label exists that is bound to the checkbox by id, otherwise you won't see the fancy checkbox or the regular checkbox. Remember this issue before? ;-)

The final fix was setting the opacity back to 0 (so the regular checkboxes won't show, just the fancy ones) and putting a proper label after every checkbox that didn't have one (to show the fancy toggle).

I found that adding class 'noscheck' to a checkbox, disables this fancy toggle feature. In other words, the regular checkbox is shown instead of the fancy one. An example of that you can find on the custom temp log.

I'll put this in a PR later on.

gizmocuz:
@michahagg , I am working on a solution that does not involve calling 'sensorTypeForDevice', It's working already for the temperature chart. Hope to commit this soon
From the original chart controller, I set a variable 'sensorType' in the function 'chartParams'

`sensorType: ctrl.sensorType`

'ctrl.sensorType' is in `app.directive('temperatureShortChart', function () {`

```
                self.range = 'day';
                self.sensorType = 'temp';
```

michahagg:
@gizmocuz about the graph being redrawn differently after a refresh. I have reproduced that issue in an isolated highchart. Please see here on jsfiddle: https://jsfiddle.net/miicha/wzufn0pk/2/.  The fiddle loads the chart when opened. If you then press the button, it'll load the same data and the chart shows the issue.

I'll look in highcharts to see what causes this issue. I think it's good to know it's not a particular Domoticz issue.

michahagg:
> @michahagg , I am working on a solution that does not involve calling 'sensorTypeForDevice', It's working already for the temperature chart. Hope to commit this soon
> From the original chart controller, I set a variable 'sensorType' in the function 'chartParams'
> 
> `sensorType: ctrl.sensorType`
> 
> 'ctrl.sensorType' is in `app.directive('temperatureShortChart', function () {`
> 
> ```
>                 self.range = 'day';
>                 self.sensorType = 'temp';
> ```

That sounds very good. That way even the controller is more reusable than it was! :-)

gizmocuz:
@michahagg , I made an attempt:

https://github.com/domoticz/domoticz/commit/74f4d57bc925fab2a7f15f825aeb6d89cac6f921

It is maybe far from ideal, but now the temperature charts work... I still use the old function (sensorTypeForDevice) in GraphLog.js

michahagg:
> @michahagg , I made an attempt:
> 
> [74f4d57](https://github.com/domoticz/domoticz/commit/74f4d57bc925fab2a7f15f825aeb6d89cac6f921)
> 
> It is maybe far from ideal, but now the temperature charts work... I still use the old function (sensorTypeForDevice) in GraphLog.js

This works very well! 👍 

BTW no need to pass 'this.sensorType' into refreshChartData() ;-)

gizmocuz:
Correct, that was left over from the previous attempt ;) Removed now
Only thing left is the Y-Axis issue

michahagg:
> Only thing left is the Y-Axis issue

I submitted a bug report at Highcharts. Let's hope it gets picked up!
https://github.com/highcharts/highcharts/issues/14297

PR #4375, opened 2020-10-03
Fix fancy checkboxes in default theme

michahagg:
How checkboxes work in the gui. All checkboxes have (or should have) an id-bound label. Like:

`<input type="checkbox" id="yesorno"><label for="yesorno">Yes or no?</label>`

When using the default theme, the label is then automatically extended with a fancy toggle switch in front of the text and the checkbox is made invisible. I'm not sure how the toggle switch is made but the checkbox is made invisible by setting the css opacity to 0.

It is possible to make an exception for a particular checkbox and show the regular checkbox instead of the toggle switch. This is done by adding css class 'noscheck' to the input-tag. In that case, an id-bound label is no longer required. An example of this can be found on the page CustomTempLog.

This PR contains:

- Added label to checkbox where missing.
- Set css opacity back to 0 for checkboxes in the default theme.

### DIFF
--- a/www/app/log/GraphLog.html
+++ b/www/app/log/GraphLog.html
@@ -17,4 +17,4 @@
         chart-title="Last Year"
         range="year"></long-chart>
 
-<label><input type="checkbox" ng-model="$ctrl.autoRefresh"> {{ ::'Auto refresh on device update' | translate }}</label>
+<input type="checkbox" id="AutoRefresh" ng-model="$ctrl.autoRefresh">&nbsp;<label for="AutoRefresh" data-i18n="Auto refresh on device update"></label>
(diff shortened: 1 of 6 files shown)


gizmocuz:
Thank you for this PR/Fix !
Maybe we should remove the option for the auto refresh and make this default ? Would there be a reason not to auto refresh ?

michahagg:
I'm for it! :-) I just copied it from TextLog.html.
On 31-3-2020 @stas-demydiuk chose to put the autorefresh checkbox there. I'm curious as to what his reasoning was for having an option to disable auto refresh.

gizmocuz:
I'm going to remove those options as they will appear on every chart soon...
If there is a good valid reason for it, we could make it globally ... define it once ;)

PR #4383, opened 2020-10-07
Aircon with me

LarsenDX:
Clean up AirconWithMe, see PR #4357

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -182,6 +182,7 @@ hardware/1Wire/1WireByKernel.cpp
 hardware/1Wire/1WireCommon.cpp
 hardware/1Wire/1WireForWindows.cpp
 hardware/AccuWeather.cpp
+hardware/AirconWithMe.cpp
 hardware/AnnaThermostat.cpp
 hardware/Arilux.cpp
 hardware/ASyncSerial.cpp
--- a/History.txt
+++ b/History.txt
@@ -19,6 +19,7 @@ Version 2020.xxxx (xxxx)
 - Implemented: The following sensor types now support RSSI: Distance Sensor, Moisture Sensor, Watt Meter, kWh Meter
 - Implemented: Websocket notification for secondary/sub devices
 - Implemented: ZWave, Legend and tooltips in Neighbor�s overview
+- Implemented: AirconWithMe wifi module
 - Updated: dzVents version 3.0.12 ( https://www.domoticz.com/wiki/DzVents:_next_generation_LUA_scripting#History )
 - Updated: eVehicles, added ODO meter, unlock/open alert, max charge level to Tesla module
 - Updated: OpenZWave configuration files
--- a/hardware/AirconWithMe.cpp
+++ b/hardware/AirconWithMe.cpp
@@ -1,0 +1,474 @@
+//
+#include "stdafx.h"
+#include "AirconWithMe.h"
+
+#include <regex>
+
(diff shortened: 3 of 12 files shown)


gizmocuz:
@LarsenDX , thank you for the PR.

Is it possible to:

- declare the Poll interval as a '#define' in the source file
- remove the poll interval in Hardware.html (this will also resolve the merge conflict, just replace the Hardware.html file with the current file from the repository)
- remove the poll interval code in Hardware.js (related to AirconWithMe of course)

LarsenDX:
> @LarsenDX , thank you for the PR.
> 
> Is it possible to:
> 
>     * declare the Poll interval as a '#define' in the source file
> 
>     * remove the poll interval in Hardware.html (this will also resolve the merge conflict, just replace the Hardware.html file with the current file from the repository)
> 
>     * remove the poll interval code in Hardware.js (related to AirconWithMe of course)

@gizmocuz, poll interval code replaced with #define . Thank you for the Hardware.html tip, saved me some work. The AirconWithMe hardware should now be good to go.

gizmocuz:
@LarsenDX , thank you (and @RDols ) very much for all the work and PR!

mcmikev:
Hi, Thanks for this great addon. I use it with my Airco and can create schedules and triggers (windows open etc)

But there is one thing that is an issue for me. When I change the IP of the airconwithme module, and change that in domoticz then the worker stops but fails to start after that.
 2020-11-06 18:44:58.003 Error: AirconWithMe: Login failed : request failed!
2020-11-06 18:44:58.003 Error: AirconWithMe: GetAvailableDataPoints failed : 

Only possible fix then is to remove hardware completly in domoticz and re-add it with new ip. That is some sort of a hassle to do so maybe this bug/error can be fixed?

And yes, the username and password are 100% correct.

RDols:
Hmmm. Im able to reproduce the error. I will look into this.

mcmikev:
> 
> 
> Hmmm. Im able to reproduce the error. I will look into this.

Great that it is reproducable.. Hope that It's a quick and simpel fix for you

RDols:
@mcmikev I tried to reproduce the error again. The only way i can reproduce this as you described is with a wrong username/password combination. Since you 100% confident you had entered the correct username and password, three additional questions.

1. Did this happen more then once?
2. Does the error persist after you restarted domoticz?
3. do you use the default username password (operator / operator) ?

LarsenDX:
Hi, I have 3 AirconWithMe modules in my house. I ran some tests and can confirm the behaviour as follows:
* In the hardware list click on the AirconWithMe hardware (already configured and enabled)
* Don't change any value, click Update
* or re-enter operator/operator in the Username/Password fields, click Update

It throws the errors in the log:

2020-11-09 13:26:25.987 Error: AirconWithMe: Login failed : request failed!
2020-11-09 13:26:25.988 Error: AirconWithMe: GetAvailableDataPoints failed

LarsenDX:
If I had to guess I'd say there's something wrong with the Update code. If I look at the DB following a click on Update, I see that the password is not set for the hardware. Instead, it reads "undefined".
Screenshot 2020-11-09 at 14 13 45

LarsenDX:
Found it! hardware.js is missing 
 				var password = encodeURIComponent($("#hardwarecontent #divlogin #password").val()); 
I dropped the line when I was cleaning up the code for the PR. Mea culpa.

PR #4385, opened 2020-10-09
Improvement Weather forecast screen and functionality

kiddigital:
This PR tries to improve the currently very limited Weather forecast screen in the Domoticz UI. It was limited to just show the DarkSky 7 day forecast of the Domoticz home location (from the settings).

Most of the Weather hardware modules (AccuWeather, DarkSky, OpenWeatherMap, BuienRadar, etc.) have a 'GetForecastURL' method implemented but that isn't used (anymore? Maybe once was, but now it isn't as far as I could find).

This PR covers the following:

- UI: Forecast controller tries to make smart use of the Forecast information from the Webserver

- UI: Forecast view shows a back button and supports different divs per hardware module and has a fallback (still DarkSky)

- Webserver: Now support a new (JSON command) call 'getforecastconfig' which return a JSON object with weather forecast data from a configured active hardware module

- Webserver: The 'getforecastconfig' command now looks for an optional UserVariable called 'forecasthardware' (integer) which should hold the hardware index number of a Weather hardware module. If found, the Weather hardware that this variable points to, will be used to get the Forecast information from.

- Webserver: For the hardware modules Buienradar and OpenWeatherMap, the new 'getforecastconfig' command retrieves the available forecast data. For not (yet) supported modules, it defaults to only basic Domoticz home location data as before. (Used by the currently default DarkSky fallback).

- Webserver: Fixed _broken_ 'getlocation' command. It did work, but also returned status ERROR instead of OK. Now it works again, but it seemed only used by the 'forecast controller' which now uses the new function. So maybe this function/command can be removed from the Webserver as I did not find any other usages of this command? (@gizmocuz any ideas?)

- OpenWeatherMap: Implemented a new public function 'getForecastData' that returns not only a possible URL, but more details like the location as configured by the module (which could be different than the Domoticz home location), cityid if available, location name, hardwaretype (OpenWeatherMap), etc. for use by the UI (forecastcontroller)

NOTE: this PR can be seen as Beta as not everything works as it should. As I am not a UI/AngularJS wizard, I definitely could use some help on that part (@michahagg, @gizmocuz ?).
BUT: it does at least work as good (or bad ;)) as the existing functionality so it is not a step back :)

New NOTE (19-10-2020): **out-of-Beta** :) OpenWeatherMap widget or Buienradar widget shown when hardware module is selected. 

To-do (or should/could):

- _Get the OpenWeatherMap widget working by having Angular pass the variables in a correct way_ -> **Fixed**

- Instead of using a UserVariable, make the selection of a Weather hardware device to use for Forecast data, something that can be done in the settings tab (and make it a PreferencesVariable).

### DIFF
--- a/hardware/OpenWeatherMap.cpp
+++ b/hardware/OpenWeatherMap.cpp
@@ -71,7 +71,7 @@ COpenWeatherMap::~COpenWeatherMap(void)
 {
 }
 
-bool COpenWeatherMap::ResolveLocation(const std::string& Location, double& latitude, double& longitude, const bool IsCityName)
+bool COpenWeatherMap::ResolveLocation(const std::string& Location, double& latitude, double& longitude, uint32_t& cityid, const bool IsCityName)
 {
 	std::string sResult;
 	std::stringstream sURL;
@@ -118,15 +118,22 @@ bool COpenWeatherMap::ResolveLocation(const std::string& Location, double& latit
 		return false;
 	}
 
-	// Process current
-	if (root["coord"].empty())
+	if (root["cod"].empty() || root["id"].empty() || root["coord"].empty())
 	{
-		Log(LOG_ERROR, "Invalid data received, could not find current weather data!");
+		Log(LOG_ERROR, "Invalid data received, missing relevant fields (cod, id, coord)!");
 		return false;
 	}
 
+	if (root["cod"].asString() != "200")
+	{
+		Log(LOG_ERROR, "Error code received (%s)!", root["cod"].asString().c_str());
+		return false;
+	}
+
+	// Process current
 	latitude = root["coord"]["lat"].asDouble();
 	longitude = root["coord"]["lon"].asDouble();
+	cityid = root["id"].asUInt();
 
 	return true;
 }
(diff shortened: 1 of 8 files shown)


michahagg:
Great addition! I'll have a look and a try sometime this weekend.

kiddigital:
Thx @michahagg that would be great. The OpenWeatherMap widget does not work as it should as somehow I can't get the view to work with the variables from the controller correctly :(

Easiest way to test the functionality is
- create the UserVariable 'forecasthardware'  (Integer with value 0)
- make sure Domoticz has a home location (settings)
- add for example an OpenWeatherMap hardware module with location '0' (Domoticz home)
- add another OpenWeathermap hardware module with location 'Kathmandu,NP' (or some other nice location :))
- add for example Buienradar hardware module

Now play around with these hardware modules and the forecast screen and switch between these hardware modules (and the default of 0) by changing the UserVariable using the UI.

I also notices that the button/tabs somehow don't always respond as expected when or after visiting the Forecast screen, not sure why.

kiddigital:
Finally now also the OpenWeatherMap widget is working :)
Instead of using the word 'device' now the term 'hardware' is used to be consistent with Domoticz
@gizmocuz , I think this is ready to be merged. It fixes some issues and improves a lot on the forecast screen
@michahagg , still I have the idea that sometimes the UI doesn't properly respond due to some 'state' issues somewhere? Can you have a look?

michahagg:
I am eager to look at your improvements and will do that soon. Unfortunately I'm a little busy for work and almost finished with an update to the temperature and power/energy charts adding autorefresh. I'll let you know!

kiddigital:
@gizmocuz , I have rebased this PR with the latest changes in Dev including the UI changes from @michahagg . 

Testing showed that everything still works as before, so I think this PR is ready to be merged.

gizmocuz:
Thank you again for your work !

kiddigital:
@michahagg , could you have a look at forecast screen? Somehow it looks like the menu or back button does not respond directly... thx

Issue #4386, opened 2020-10-09
Hardware Monitor: Possible code cleanup, detecting Pi hardware


kiddigital:
At the moment, the 'Montherboard Sensors" hardware (HTYPE_system) in case of Pi hardware, is using variables/values populated outside of the hardware module itself.

It depends on the 'CheckForOnboardSensors' function which is a classless method defined in the core code of 'domoticz.cpp'.

Shouldn't this function be moved to the internals of the 'hardwaremonitor' module?

This could cleanup the core code and gets rid of some global variables and defines that are only relevant for this hardwaretype.

Did I miss something in my analysis why this would not be possible?

@gizmocuz , @rwaaren , @akamming , I noticed a PR from a little time ago where some changes has been made in this module and core. Any thought?

If feasible, I will give it a try somewhere in the coming days.

gizmocuz:
@kiddigital , yes you're right, this could be moved to the hardware monitor class, thanks in advance!

PR #4389, opened 2020-10-10
Hardware Monitor: Moved code from main to specific hardware module (private)

kiddigital:
This PR is for Issue #4386 

I think that the code has been moved correctly from the main 'domoticz.cpp' to the correct locations in the hardware/hardwaremonitor.cpp (.h).

Difficulty is that it is hard to test the different hardware types. It seems to run fine on my (virtual) Linux machine, but that does not mean it runs on all others..

Hopefully some people can help testing?


gizmocuz:
Thanks for the awesome work!

PR #4392, opened 2020-10-11
Added new Evohome Fault log message

brucemiranda:


gizmocuz:
@brucemiranda , thank you for this PR!.
Is it possible to use tabs for idents instead of spaces, and if possible keep the 'if/else' code using the checks starting with the lowest number to the highest ?

PR #4393, opened 2020-10-11
Hardware Monitor, more code cleanup and less system/compiletime specific code

kiddigital:
This PR is a followup on #4389 

- Removed (commented out) code from main domoticz and helper (left over from previous PR) 

- HardwareMonitor code has been cleaned further. Moved global variables and functions to be private.

- HardwareMonitor generic module code (like Do_work, etc) does not have any system/compiletime specific code anymore. Moved to system specific functions.

- Some small improvement like better debugging messages, quicker to first run of reading (and adding) sensors, etc.

- There now is a public method 'GetOSType' which can be used to retrieve the detected OS (system) type.

Hopefully, the code cleanup and refactoring makes the module code better readable and easier to maintain and/or extend

NOTE: I am not able to test this module on all the different hardware/OS platform it could/should support.
Based on the code, it should work on:
* Linux -> _tested_
* Linux on Windows Subsystem for Linux (WSL)
* Raspberry Pi (Linux)
* FreeBSD
* OpenBSD
* Cygwin (sort of WSL but older)
* Apple
* Windows (Using openhardwaremonitor.org module and WMI)

Hopefully others can test and confirm it works

### DIFF
--- a/hardware/HardwareMonitor.cpp
+++ b/hardware/HardwareMonitor.cpp
@@ -157,7 +157,7 @@ void CHardwareMonitor::Do_Work()
 	Log(LOG_STATUS, "Hardware Monitor: Started");
 
 	int msec_counter = 0;
-	int64_t sec_counter = POLL_INTERVAL_CPU - 5;
+	int64_t sec_counter = 140 - 2;	// Start at a moment that is close to most devicecheck intervals
 	while (!IsStopRequested(500))
 	{
 		msec_counter++;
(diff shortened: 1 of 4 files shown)


gizmocuz:
Thank you for this PR!

Could you restore this change:

--	int64_t sec_counter = POLL_INTERVAL_CPU - 5;
	int64_t sec_counter = 140 - 2;	// Start at a moment that is close to most devicecheck intervals

This is causing the first poll to be after a very long time

kiddigital:
> 
> 
> Thank you for this PR!
> 
> Could you restore this change:
> 
> -- int64_t sec_counter = POLL_INTERVAL_CPU - 5;
> int64_t sec_counter = 140 - 2; // Start at a moment that is close to most devicecheck intervals
> 
> This is causing the first poll to be after a very long time

That might seem the case at a glance but actually in the way it was implemented here, the first poll is actually earlier and foremost all other first polls for certain sensor are also much earlier.

Before it took about 120+ seconds before for example the Disk check was performed, so there was a windows of more than 2 minutes between the first and last check for the first time. Now all checks are performed within a 35 second window and still keeping the differently timed checking intervals with 170 seconds for disk for example.

 Hope it makes sense :)

gizmocuz:
Thanks for the explanation !
And thanks for the PR!

kiddigital:
No problem and your welcome.

Always nice if things get easier 😁

PR #4394, opened 2020-10-12
RequestStop instead of ugly break

kiddigital:
PR with small cleanup.

Goal is/was to get rid of the heartbeat checking message 'Error: XXX hardware (X) thread seems to have ended unexpectedly' as the worker exits in a clean way now.

Although the Worker (and thus the thread) now exits nicely, the hardware thread remains on the threadstack that is used to check for heartbeat updates, etc.

As the thread has ended, the heartbeat is not updated anymore offcourse, but Domoticz keeps checking this thread.

@gizmocuz , any suggestions how to remove the thread from the threadstack somehow? This can not be done in the 'Do_Work' method itself because it will crash Domoticz (makes sense, you can not remove 'yourself' while you are active).

### DIFF
--- a/hardware/eVehicles/eVehicle.cpp
+++ b/hardware/eVehicles/eVehicle.cpp
@@ -337,6 +337,7 @@ void CeVehicle::Do_Work()
 	int fail_counter = 0;
 	int interval = 1000;
 	bool initial_check = true;
+	bool bIsAborted = false;
 	Log(LOG_STATUS, "Worker started...");
 
 	while (!IsStopRequested(interval))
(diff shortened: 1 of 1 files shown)


gizmocuz:
@kiddigital , there is nothing wrong with the break.
It causes the while loop to abort and the thread to end

In this particular case, you could also set a boolean (bThreadAborted) and if this is the case, do a continue

```
	bool bIsAborted = false;

	while (!IsStopRequested(interval))
	{
		interval = 1000;
		sec_counter++;
		time_t now = mytime(0);
		m_LastHeartbeat = now;

		if (bIsAborted)
			continue;

```

kiddigital:
That does the thread to keep running, just doing nothing? I thought of that as well, but it seemed a waste of resources (true that it is very limited).

So there is no clean way (yet) for a Hardware thread to de-activate itself?

If not, I will go for your suggestion.

gizmocuz:
That is a clean way. Of course, what should happen maybe is that it does nothing after 3 retries and then tries again in an hour ?
This will cause a (3) log messages every hour to get the users attention.
Currently the authentication could fail yesterday, and nobody knows by looking in the log

kiddigital:
Good point... I will implement the loop and a warning once every few hours.

kiddigital:
Implemented the IsAborted check. Now the worker keeps heartbeating but does nothing after 3 failed login attempts. Once every 2 hours it logs an Error message and once every day (24hrs) it tries to authenticate again with the same (wrong?) credentials for max 3 times. And than we are back to reporting every 2 hours, and so on...

gizmocuz:
Thanks!!

PR #4397, opened 2020-10-13
Fix FreeBSD HardwareMonitor compilation under FreeBSD

xbeaudouin:
Two fixes for FreeBSD HardwareMonitor compilation under FreeBSD... (and one to say to CLang to stop complaining...)

### DIFF
--- a/hardware/HardwareMonitor.h
+++ b/hardware/HardwareMonitor.h
@@ -96,6 +96,9 @@ private:
 	float GetProcessMemUsage();
 	float GetMemUsageLinux();
 #endif
+#if defined (__FreeBSD__)
+	float GetMemUsageLinux();
+#endif
 #if defined (__FreeBSD__) || defined(__OpenBSD__)
 	float GetMemUsageOpenBSD();
 #endif
--- a/hardware/HardwareMonitor.cpp
+++ b/hardware/HardwareMonitor.cpp
@@ -465,7 +465,7 @@ bool CHardwareMonitor::GetOSType(nOSType &OStype)
 
 std::string CHardwareMonitor::TranslateOSTypeToString(nOSType OSType)
 {
-	std::string sOSType = "Unknown";
+	std::string sOSType = "";
 
 	switch (OSType)
 	{
@@ -493,7 +493,9 @@ std::string CHardwareMonitor::TranslateOSTypeToString(nOSType OSType)
 		case OStype_Apple:
 			sOSType = "Apple";
 			break;
-		default: break;
+		default:
+			sOSType = "Unknown";
+			break;
 	}
 	return sOSType;
 }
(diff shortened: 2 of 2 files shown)


hardware/HardwareMonitor.h
```diff
@@ -96,6 +96,9 @@ class CHardwareMonitor : public CDomoticzHardwareBase
 	float GetProcessMemUsage();
 	float GetMemUsageLinux();
 #endif
+#if defined (__FreeBSD__)
+	float GetMemUsageLinux();
+#endif
 #if defined (__FreeBSD__) || defined(__OpenBSD__)
```
kiddigital, line 102: @xbeaudouin , how does this relate to this OR ? We compile 2 memory function for FreeBSD (GetMemUsageLinux and GetMemUsgaeOpenBSD)?

kiddigital (commented):
Good catch... maybe here explicit set the sOStype variable to Unknown (again and/or make the variable at the top empty to start with) to make things even more readable.

hardware/HardwareMonitor.h
xbeaudouin: Well the original compilation looks for GetMemUsageLinux() as you see in the following error while compiling.
```
hardware/HardwareMonitor.cpp:743:26: error: out-of-line definition of 'GetMemUsageLinux' does not match any declaration in 'CHardwareMonitor'
        float CHardwareMonitor::GetMemUsageLinux()
                                ^~~~~~~~~~~~~~~~
1 warning and 1 error generated.
```

Aslo on this code there is also specific FreeBSD stuff. So FreeBSD seems to need this function.

xbeaudouin:
Well Clang 8.x is very picky about switch().
It needs a default.
Maybe we should do

```
case default : 
  sOStype= "Unknown';
  break; 
```

What do you think ?

kiddigital:
> Well Clang 8.x is very picky about switch().
> 
> It needs a default.
> 
> Maybe we should do
> 
> 
> 
> ```
> 
> case default : 
> 
>   sOStype= "Unknown';
> 
>   break; 
> 
> ```
> 
> 
> 
> What do you think ?
> 
> 

Sounds good 👍

xbeaudouin:
This is done :D

hardware/HardwareMonitor.h
kiddigital: @xbeaudouin ok clear. Maybe than it is more clear to make it something like
`#if defined (__linux__) || defined (__FreeBSD__)`
and remove the function declaration 'float GetMemUsageLinux()' from the lines above (97)

This way it is only referenced once in the header file and it is visible at once for which systems it is needed.

kiddigital:
BTW, thx for testing and helping out to make sure my previous changes from #4393 work on FreeBSD :)

gizmocuz:
Thank you both !!

Issue #4379, opened 2020-10-05
Zwave blind open issue


nonolk:
Dear Gizmocuz,

In complement to this [4285](https://github.com/domoticz/domoticz/pull/4285), it's working perfectly when using the slider, but Open button is still sending 255 which is wrong see logs below:
When pushing the open button:
_2020-10-05 14:53:21.046 Status: User: Admin initiated a switch command (1029/Volet Baie vitrée 1/On)
2020-10-05 14:53:21.047 OpenZWave: Domoticz has send a Switch command!, Level: 255, NodeID: 55 (0x37)
2020-10-05 14:53:21.056 (Zwave) Light/Switch (Volet Baie vitrée 1)_

When using the slider:
_2020-10-05 14:54:52.917 Status: User: Admin initiated a switch command (1029/Volet Baie vitrée 1/Set Level)
2020-10-05 14:54:52.918 OpenZWave: Domoticz has send a Switch command!, Level: 99, NodeID: 55 (0x37)
2020-10-05 14:54:52.926 (Zwave) Light/Switch (Volet Baie vitrée 1)_

The device type is Blind percentage inverted as you can see below:
![Screen Shot 2020-10-05 at 16 56 17](https://user-images.githubusercontent.com/26922736/95096404-81a59500-072c-11eb-8a88-47abfb5ef491.png)

And the micromodule is a Qubino ZMNHCDx:
![Screen Shot 2020-10-05 at 17 02 54](https://user-images.githubusercontent.com/26922736/95096511-a437ae00-072c-11eb-8391-7045706629b7.png)

I'm able to compile domoticz from sources and to test debug code if needed.

Best regards,

gizmocuz:
@nonolk , when you set the slider to 50%, next press the 'close' button.
Next press the 'open' button, where does your blind go to ?

nonolk:
@gizmocuz the blind goes back to 50% ;-)
Here are the logs:
_ 2020-10-10 10:02:40.570 OpenZWave: Domoticz has send a Switch command!, Level: 50, NodeID: 54 (0x36)
2020-10-10 10:03:11.166 OpenZWave: Domoticz has send a Switch command!, Level: 0, NodeID: 54 (0x36)
2020-10-10 10:03:27.819 OpenZWave: Domoticz has send a Switch command!, Level: 255, NodeID: 54 (0x36)_
[...]

PR #4398, opened 2020-10-13
Updated Mainworker.cpp for better handling of Zwave blinds

nonolk:
As described in issue #4379, Zwave blind were not handled in a proper way,  as when pushing the open button, Domoticz was sending level 255, which means go back the last position. Together with Gizmocuz, we found a clever way to solve this.


gizmocuz:
Thanks!!

PR #4401, opened 2020-10-14
HardwareMonitor: More clear code. No real change, just better readability

kiddigital:
Small PR to improve code readability.

Instead of 2 times declaring the same method for each OS separately, now combined nicely.


gizmocuz:
Thanks!

PR #4406, opened 2020-10-19
Removed obsolete code as function has moved to HardwareMonitor module

kiddigital:
Found some leftover code in main/Helper.h that should not be there anymore as this function is now a private method of the HardwareMonitor module.


gizmocuz:
Thank you ! Nice find!

Issue #3861, opened 2020-01-23
Evohome Hotwater sensor graphs are showing incorrect setpoint


brucemiranda:
I changed my Hot Water setpoint from 60C to 55C today on the Evohome Controller. However the Domoticz graphs for the Hot Water device are still showing the setpoint as 60C. Which means that the set point is fixed in Domoticz somewhere. But where is that set?
I'd like the code to be able to change that set point value based on what the user does on Evohome.

brucemiranda:
I've now submitted a PR myself to fix this.

Olifant1990:
Please refer to it with a link ;)
https://github.com/domoticz/domoticz/pull/4414

brucemiranda:
This will stay open until the PR 4414 is merged.
https://github.com/domoticz/domoticz/pull/4414

Olifant1990:
That is wrong in my and GitHubs way ;)
You can close this as this problem is fixed in a PR.
As it is linked by naming it, everyone can see the PR is still open, but the bug itself is fixed!

brucemiranda:
Fair enough. Closed.

PR #4414, opened 2020-10-22
Decode, display and store Evohome DHW Setpoint

brucemiranda:
The only bit that I cannot figure out how to do is when the EditState window is called, how to default the State to "On" when the Set point is set to a numeric value other than 60C

### DIFF
--- a/hardware/EvohomeRadio.cpp
+++ b/hardware/EvohomeRadio.cpp
@@ -41,11 +41,12 @@ enum evoCommands
 	cmdSetpointOverride = 0x2349,
 	cmdDHWState = 0x1F41,
 	cmdDHWTemp = 0x1260,
+        cmdDHWSettings = 0x10A0,//DHW settings sent between controller and DHW sensor can also be requested by the gateway
 	cmdControllerMode = 0x2E04,
 	cmdControllerHeatDemand = 0x0008,//Heat demand sent by the controller for CH / DHW / Boiler  (F9/FA/FC)
 	cmdOpenThermBridge = 0x3220,//OT Bridge Status messages
 	cmdOpenThermSetpoint = 0x22D9,//OT Bridge Control Setpoint
-	cmdActuatorState = 0x3EF0,
+ 	cmdActuatorState = 0x3EF0,
 	cmdActuatorCheck = 0x3B00,
 	cmdBinding = 0x1FC9,
 	cmdExternalSensor = 0x0002,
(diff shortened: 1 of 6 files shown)


brucemiranda:
OK this PR can be merged. The lack of a Default doesn't affect the functionality.

Olifant1990:
This fixed the bug in https://github.com/domoticz/domoticz/issues/3861

brucemiranda:
@gizmocuz any reason why this PR hasn't been merged yet?
It's a feature that quite a lot of Evohome users have asked for.

gizmocuz:
@brucemiranda , Not sure, but it seems fixed by the comment of Olifant1990 ?

What is the deal in

SQLHelper::UpdateTemperatureLog()

If On/Off then 0 or 60 ?
For example on my heating system the DHW value is 75, so a SetPoint above 60 is valid
Can't see in the code where there is a check for 60 or 0, is it in the Gui/Graph code ?
Another way is to use a Switch for the On/Off state ?

Happy to merge, but just need to be sure if it is fixed already

Olifant1990:
@gizmocuz sorry, with "this fixed the bug in 3861" I meant to refer to the originating bug ticket.

I did not test it, but wrote it to connect the dots/history ;)

brucemiranda:
> @brucemiranda , Not sure, but it seems fixed by the comment of Olifant1990 ?
> 
> What is the deal in
> 
> SQLHelper::UpdateTemperatureLog()
> 
> If On/Off then 0 or 60 ?
> For example on my heating system the DHW value is 75, so a SetPoint above 60 is valid
> Can't see in the code where there is a check for 60 or 0, is it in the Gui/Graph code ?
> Another way is to use a Switch for the On/Off state ?
> 
> Happy to merge, but just need to be sure if it is fixed already

@gizmocuz Before this change Evohome DHW used to only use ON and OFF states. However within the db, ON was being stored as 60C and OFF was 0C. What the change now does is if there is a value besides ON or OFF, then it uses the actual value to update the database. So in your example if the DHW setpoint is 75C it would store 75C. The code just keeps the old functionality still working, where ON and OFF were set to hard coded values. BTW : OFF is still stored as 0C, which is correct.

gizmocuz:
@brucemiranda , thank you for the explanation. Do we still need the old functionality ? It seems you rebuild it and it is now obsolete?

brucemiranda:
> @brucemiranda , thank you for the explanation. Do we still need the old functionality ? It seems you rebuild it and it is now obsolete?

Probably best to keep the old functionality for a while. We can phase it out in time.

gizmocuz:
Thanks!!

PR #4416, opened 2020-10-24
clang-tidy fixes

neheb:
Besides the last one, this is cosmetic.


gizmocuz:
@neheb , thank you very much for this PR and all the work !! A lot of work it must have been !

A few comments:

- I am not a fan of changing everything to 'Auto' specially when on some places it makes the code hard to read and debug
Sure for (for example) iterators it makes perfect sence

This for example:
const tRBUF *pSen = reinterpret_cast<const tRBUF*>(pdata);
const auto pSen = reinterpret_cast<const tRBUF*>(pdata);

There is no need to use the auto keyword, it's perfectly clear was is meant here
 
Declaring std::strings as function parameters, keep them the way there are, there is no need to remove 'const std::string &' and should also not be done.

For example:

`	CAnnaThermostat(const int ID, std::string IPAddress, const unsigned short usIPPort, const std::string &Username, const std::string &Password);
`
change this back to: const std::string &IPAddress

Or code like:

before:
`CAirconWithMe::CAirconWithMe(const int id, const std::string& ipaddress, const unsigned short ipport, const std::string& users, const std::string& password)
`

after:
`CAirconWithMe::CAirconWithMe(const int id, std::string  ipaddress, const unsigned short ipport, std::string  users, std::string  password)
`	

A lot has been done by search/replace, but this leaves double spaces like:

`Comm5Serial(const int ID, std::string  devname`


I would have split this PR into multiple PR's with different changes or first a small discussion

neheb:
Well, initially this PR has the first three commits. Then I added on it. I can get rid of the extra ones and place them as separate PRs.

I'll look at the function thing.

This codebase is a bit of a mess. Normally when running clang-tidy, I also run git clang-format to get rid of those space issues you mentioned. But with this codebase, that would result in a lot of changes. I have no idea which coding style to specify either.

gizmocuz:
Yep, well that's because not every code is written by one person...

neheb:
> Yep, well that's because not every code is written by one person...

Right. I assume it would be a good idea to set up a clang-format file so that moving forward, all commits can be formatted to a standard.

neheb:
Removed some of the problematic commits.

kiddigital:
> > Yep, well that's because not every code is written by one person...
> 
> 
> 
> Right. I assume it would be a good idea to set up a clang-format file so that moving forward, all commits can be formatted to a standard.

I am in favor of such a format-style (and other clang-tidy options that are considered relevant for this project) so it becomes possible to assess the code 'quality' and per PR use this as guidance.

I don't think it is possible (yet) to enforce this as this could mean a 'small' change somewhere could mean that the whole source file has to be cleaned-up. That will scare off potential/current committers to push changes.

neheb:
eh no.

git clang-format limits itself to the actual changes that are commited. It doesn't modify the whole file. You need clang-format for that.

Note that all of these commits have git clang-format applied to them. The style I used is based on LLVM with it modified to use tabs instead of spaces.

neheb:
Rebased. I removed several changes having to do with tinyxpath, which should not really be modified.

I also applied clang-format to it. Please review.

gizmocuz:
@neheb , thanks for all the work !! There seems to be some compile issues, for example:

eHouseTCP.cpp(827,56): error C2065: 'i': undeclared identifier
eHouseTCP.cpp(829,29): error C2065: 'i': undeclared identifier
eHouseTCP.cpp(831,13): error C2065: 'i': undeclared identifier
eHouseTCP.cpp(842,60): error C2065: 'i': undeclared identifier
eHouseTCP.cpp(844,29): error C2065: 'i': undeclared identifier
eHouseTCP.cpp(846,13): error C2065: 'i': undeclared identifier

I am going to revert this patch, Please test this patch also when you compile with Visual Studio 2019 (community edition)

neheb:
The problem is only the last commit. Interesting things happen when doing things manually :).

gizmocuz:
@neheb , no worries, but how to get this PR to be back again, or are you making a new PR?

neheb:
I'll make a new one. Maybe separate the changes so they can be reviewed easier.

gizmocuz:
That would be great... I was just saying on our slack channel that done PR's are so large that it is hard to test
Breaking them in smaller pieces will be great!

PR #4417, opened 2020-10-26
clang-tidy: replace bind calls with lambdas

neheb:
Found with modernize-avoid-bind

The motivation to remove boost::bind is the change in boost 1.73 where the placeholders moved namespaces.

All of these were manually done. The actual check relies on C++14 as it uses auto&& whereas here I've removed such things to keep it compatible with C++11.

From the description page:

std::bind can be hard to read and can result in larger object files and binaries due to type information that will not be produced by equivalent lambdas.

### DIFF
--- a/hardware/ASyncSerial.cpp
+++ b/hardware/ASyncSerial.cpp
@@ -33,7 +33,6 @@
 #include <algorithm>
 #include <iostream>
 #include <boost/asio.hpp>
-#include <boost/bind/bind.hpp>
 #include <boost/function.hpp>
 #include <boost/thread.hpp>
 #include <boost/smart_ptr/shared_array.hpp>  // for shared_array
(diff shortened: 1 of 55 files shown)


neheb:
This is finished.

neheb:
I'll rebase after everything else goes through.

gizmocuz:
@neheb , Not sure if this PR was ready to be merged?

neheb:
I have a newer version that gets rid of all bind calls. Will update soon.

neheb:
Done. The only exception is all of the binds in main/WebServer.cpp. For some reason, replacing those bind calls results in larger compiled size.

neheb:
Meh. replaced with std::bind for those.

gizmocuz:
I think this could be because the bind takes a pointer, and the lambda is compiled inline
The speed of a Lambda is faster than the bind, but in this case/project it would not matter much ;)

gizmocuz:
Thank you for all the work again !!!

neheb:
Right but the lamba conversions elsewhere result in lower compiled size. Maybe lambdas being const default allows the optimizer to do some work.

Anyway, whatever.

kofec:
After all clang cleaning the size grow about 1,5MB for arm (Rasp PI) on Openwrt:
```
10 098 871 domoticz_2020.2.12279-1_arm_arm1176jzf-s_vfp.ipk
11 744 169 domoticz_2020.2.12686-1_arm_arm1176jzf-s_vfp.ipk
```

neheb:
Hmm maybe it's a clang thing. I compile with clang instead of GCC.

gizmocuz:
I have a crash when using the web interface (http)

connection.cpp
Line 210

```
			else {
				// Perform plain read
				socket_->async_read_some(buf, [this](const boost::system::error_code &err, size_t bytes) { shared_from_this()->handle_read(err, bytes); });
			}
```

I have to revert this PR...

gizmocuz:
It's working again.
I suspect the issue is because the whole webserver code heavily depends on boost ... 
I was not able to test the other changes as connecting to my Domoticz debug instance via HTTP already caused the above crash
Maybe changing the boost usage to std would be more standard c++11

PR #4418, opened 2020-10-26
add clang-format file

neheb:

Note that this will replace tabs with spaces. The codebase currently uses a mix of tabs and spaces.

This is basically just clang-tidy --dump-config with column limit bumped to 100.


gizmocuz:
I think (hope) dat 99,9% the code is using tabs instead of spaces.
Could you adjust this ?

neheb:
Do you mean to change https://github.com/domoticz/domoticz/pull/4418/commits/8679c58078b821ee116605b4c549a30e8311551c#diff-1026e0038b722990204a42bed8a6f7c0ec2302aa79e3fad1959d62ba968edfa2R135 ?

gizmocuz:
@neheb , Yes I think thank field would do the trick. I do have to get myself some clang-format lessons

neheb:
Reworked the file so that it's closer to the style used here.

gizmocuz:
Thank you !!

PR #4419, opened 2020-10-26
Fix SQLHelper UpdateValueInt to check DeviceCreation status (fixes #4411)

### LINKED ISSUE #4411: Phantom records in LightingLog with DeviceRowID of 18446744073709551615 (Hex: 0xFFFFFFFFFFFFFFFF)
I am on Linux raspberry 5.4.51-v7l+ #1333 SMP Mon Aug 10 16:51:40 BST 2020 armv7l GNU/Linux with Domoticz version 2020.2 build 12467
I have 2 systems running with different hardware attached. Both system have the above phantom records. The slave has a lot less phantom records then the other main system.
Looking at the date and time of those records and compare them with the logging I could not find a devices what could be responsible for those records.
I also tried with logging: 
`DAEMON_ARGS="$DAEMON_ARGS -loglevel normal,status,error,debug -debuglevel normal,hardware,received,webserver,eventsyst$
DAEMON_ARGS="$DAEMON_ARGS -loglevel info"`
But there was no SQLH record in the log file that gave a hint to which device might be responsible.
For the main system with approx. 230 active devices I had to set the Short Log Sensors to 1 day, because of the number of phantom records.

Where are they come from, and what can be done about it?

kiddigital:
Small PR to fix incorrect handling of the -1 return code (actually being an uint64_t) in the SQLHelper::UpdateValueInt method.

This should fix issue #4411


gizmocuz:
Thanks!!

PR #4420, opened 2020-10-27
Open Weather Map: Add a delay toggle to lower the update interval to 30 minutes (instead of default 5 minutes)

kiddigital:
This PR adds an additional toggle to the Open Weather Map hardware module that delays the update frequency.
With this toggle 'active' the normal 5 minute interval will be delayed to a 30 minute interval (so 2x per hour).

Normally an instance of the OWM module refreshes its data every 5 minutes, which is fine for most use-cases.
But in some cases this update interval is to fast. 

Some people like to monitor the weather for multiple locations and therefor run multiple instances of the OWM module.
At the moment, the Free OWM API licence key has a limit of 1000 calls per day (30.000 per month).
When running at an interval of 5 minutes, which translates to 288 calls per day, no more than 3 OWM instances can run at the same time. Otherwise the modules will stop receiving data near the end of the day once the day limit is hit.
By activating the Delay toggle, an OWM module instance will only update once every 30 minutes resulting in 48 calls per day. With 48 calls per day for a single OWM module instance, one could run 20 OWM instances when all the instances have the Delay activated.

(And yes, one could request multiple Free API keys as a work-around.)

See also the 'suggestions' section on the forum for more details about the needs for a lower update frequency.

Any comments, suggestions, etc. always welcome.

### DIFF
--- a/hardware/OpenWeatherMap.cpp
+++ b/hardware/OpenWeatherMap.cpp
@@ -51,14 +51,18 @@ std::string ReadFile(std::string filename)
 #define OWM_icon_URL "https://openweathermap.org/img/wn/"	// for example 10d@4x.png
 #define OWM_forecast_URL "https://openweathermap.org/city/"
 
-COpenWeatherMap::COpenWeatherMap(const int ID, const std::string &APIKey, const std::string &Location, const int adddayforecast, const int addhourforecast) :
+#define OWM_Poll_Interval 300
+
+COpenWeatherMap::COpenWeatherMap(const int ID, const std::string &APIKey, const std::string &Location, const int adddayforecast, const int addhourforecast, const int intervaldelay) :
 	m_APIKey(APIKey),
 	m_Location(Location),
 	m_Language("en"),
 	m_add_dayforecast(adddayforecast),
-	m_add_hourforecast(addhourforecast)
+	m_add_hourforecast(addhourforecast),
+	m_intervaldelay(intervaldelay)
 {
 	m_HwdID=ID;
+	m_interval = OWM_Poll_Interval;
 
 	std::string sValue;
 	if (m_sql.GetPreferencesVar("Language", sValue))
@@ -67,7 +71,7 @@ COpenWeatherMap::COpenWeatherMap(const int ID, const std::string &APIKey, const 
 	}
 }
 
-COpenWeatherMap::~COpenWeatherMap(void)
+COpenWeatherMap::~COpenWeatherMap()
 {
 }
 
@@ -136,7 +140,7 @@ bool COpenWeatherMap::StartHardware()
 	std::string sValue, sLatitude, sLongitude;
 	std::vector<std::string> strarray;
 	Debug(DEBUG_NORM, "Got location parameter %s", m_Location.c_str());
-	Debug(DEBUG_NORM, "Starting with setting %d, %d", m_add_dayforecast, m_add_hourforecast);
+	Debug(DEBUG_NORM, "Starting with setting %d, %d, %d", m_add_dayforecast, m_add_hourforecast, m_intervaldelay);
 
 	if (m_Location.empty())
 	{
@@ -253,6 +257,11 @@ bool COpenWeatherMap::StartHardware()
 		}
 	}
 
+	if (m_intervaldelay > 1)
+	{
+		m_interval = OWM_Poll_Interval * m_intervaldelay;
+	}
+
 	RequestStart();
 
 	m_thread = std::make_shared<std::thread>(&COpenWeatherMap::Do_Work, this);
@@ -275,18 +284,16 @@ bool COpenWeatherMap::StopHardware()
 	return true;
 }
 
-#define OpenWeatherMap_Poll_Interval 300
-
 void COpenWeatherMap::Do_Work()
 {
-	int sec_counter = OpenWeatherMap_Poll_Interval - 3;
+	uint64_t sec_counter = m_interval - 3;
 	while (!IsStopRequested(1000))
 	{
 		sec_counter++;
 		if (sec_counter % 12 == 0) {
-			m_LastHeartbeat = mytime(NULL);
+			m_LastHeartbeat = mytime(nullptr);
 		}
-		if (sec_counter % OpenWeatherMap_Poll_Interval == 0)
+		if (sec_counter % m_interval == 0)
 		{
 			try
 			{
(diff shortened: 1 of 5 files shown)


gizmocuz:
Yep, my suggestion is to get yourself another free account, then you can monitor 6
I don't know why people would like to monitor that much...

jefft4:
@gizmocuz , one origin of this request was me, and it was nothing to do with having a lot of forecasts (though I see that angle); it's just that 5 minutes is way more frequent than is necessary if one is only looking for a forecast (those change perhaps every hour in reality) rather than current data.  As there is no way to filter what data is retrieved, pulling data every 5 minutes creates unnecessary processing, network traffic, log entries, etc.
Thanks, @kiddigital for adding that :-)

gizmocuz:
@jefft4 , there is more information that just a forecast (temp/wind/...)
And it's perfectly fine polling every 5 minutes with one account. If you are short of requests (more then 3 different locations), you can create another account.
It's also very little data
Closing this one...

eaglefred:
i'm sorry to read that the update timer of 5 minutes could'nt be personalized. In fact it is far more than necessary for my use (just an update every 2 or 4 hours is enough for me).
The most annoying fact of this is that every update (of about 50/60 sensors ?), the MQTT topic domoticz/out is filled by all those updated informations. And lot of my sensors which don't have interest into Weather infos must flush all of the domoticz/out messages before retrieving something interesting for them.

gizmocuz:
@eaglefred , Yeah and also 433Mhz temperature sensors that transmit every 30 seconds if maybe to much but can't be changed
Not sure why you use MQTT, but you can also look at the Domoticz client/server setup if you have multiple locations

jefft4:
@gizmocuz , you're kind of missing the point here.  OWM hits the log and the message queue every 5 minutes with loads of updates, many of which are of no use to many users.  Personally, I use it just to get 3 hours of temperature forecast to help my heating control algorithm, but it posts something like 50 items that I couldn't care less about.  That's not the same thing as one message from a sensor that's there because someone cares what it says.
I've never been clear why you object to making the poll frequency configurable, or the solution that @kiddigital created.

eaglefred:
Yes, I think i was misunderstood. OWM throws 50 messages in MQTT domoticz/out every 5 minutes. But i just need some UV index forecast.
So all my sensors have to pass all this MQTT messages before getting what they want, and this is very time/bandwidth/battery consuming.
My sensors could be ok with just an 1 or 2 seconds MQTT.loop() but i have to loop every 500ms to be able to flush the MQTT queue of the 50 messages every 5 minutes.
Solution, i think would be:
- solution by @kiddigital 
- or being able to delete some OWM sensors, and only have those really wanted by the user

gizmocuz:
Because some users want to have the poll interval in 10 seconds which also does not make sense
Domoticz is of a broad kind of users with all their different wishes. And most of all the concept is 'keep it simple'
Adding tons of configuration options will not make things easier

What happens when you set the sensors you do not want to not used? (From the devices tab)
Do you still get MQTT updates for them, because this is something we can disable for unused devices

eaglefred:
You're right, everybody have their own usage of Domoticz but I  think some kind of genericity is desirable.

In order to have less MQTT messages, i tried to delete all the OWM sensors i don't use, but  OWM recreate them on the next update.
I just have the "Used" flag for
UV Index
UV Index 0
UV Index 1

but get MQTT messages for all.

gizmocuz:
Did you disable accept new hardware in settings?

eaglefred:
No i haven't
Then i can delete the unused OWM sensors ?

gizmocuz:
Okey, disable accept new hardware in the settings, next delete your unwanted devices

eaglefred:
Yeah, it's ok now. OWM Sensors aren't recreated and I just have 3 OWM MQTT messages.
So 3 messages to skip every 5 minutes by my "non-weather sensors" is pretty quick to do.
But what if somedays I need to enable again new hardware detection in settings ? Risk is high that i have to delete again all the unwanted OWM sensors...

gizmocuz:
That is why I asked what happened when you set the devices to 'not-used'. This is not the same as deleting.
(Press the little blue circle with the white error to the left)
But how often do you add new hardware?

eaglefred:
When some OWM devices are unused, i still get their MQTT messages.

You're right, I don't add new hardware very often, 2 or 3 by year, but often with try and retry. And all in all, this behavior isnt' really user friendly, isn't it ?
Despite my complain, be sure that I find OWM extremely useful and pleasant to use. And this is just my REX and user point of view for improvement.

kiddigital:
My 2 cents..

Depending on the hardware module, it can generate a few or a lot of devices, some relevant to the user some not.

The update frequency can be high or low again depending on the device and could or could not be useful to the user...

The 'problem' discussed here about MQTT is about 'which messages from which devices are relevant to whom when and how often'. This 'problem' is independent of the amount of devices and/or update frequency, it is a side effect.

MQTT (and other pub/sub models) are designed to help with this 'problem', enter [Topics](https://www.hivemq.com/blog/mqtt-essentials-part-5-mqtt-topics-best-practices/)

Unless you want to know everything, you should not subscribe to the `out` topic (as this topic is meant to broadcast _everything_) but only to relevant topics.

Primarily Domoticz is about receiving (and processing) incoming measurements and if needed act on them. Currently _redistributing_ (or _broadcasting_) is not core and bare functionality. It just publish everything.

If there is a need to be better/smarter about what/how/when to publish, than the focus should be on that.

Not changing the behavior of the hardware modules.

eaglefred:
Some comments about your post @kiddigital 
- the fact that some MQTT messages are broadcasted (and more generally there's some processing) on OWM devices that are flagged "unused" isn't judicious. It is unnecesseraly time/cpu/network consuming
- impose a frequency update to everybody  isn't UX friendly. When it's not just a personal dev module, and you want to deliver your work to community, you would have to be as generic as possible. It's as if Domoticz didn't offer a retention period of his logs, or a frequency notification, or expiration delay of sensors,... 
- when some IoT devices is waiting for info from domoticz, the most efficient way to do this is with MQTT. And unfortunately, Domoticz have juste one topic to send info: domoticz/out. This is a major problem of Domoticz, need some improvements, and it causes some side effects to OWM.

Despite all of this, I repeat that OWM is a great module, and i'm very happy with all the info it brings to me. So thank's for that work. Thank you very much.

PR #4425, opened 2020-11-02
P1 Chart: don't show 'return' graphs if not applicable

michahagg:
Hi @gizmocuz, this PR contains the fix for the issue that you reported. When the P1 data does not contain returned-energy-data, as per boolean 'delivered' in the data response, the corresponding series will not be added to the charts. And they won't show up in the legend as well.

Your concern was about how the charts appeared with the extra (empty) series added. I think that the chart itself appears unchanged, but the extra item in the legend might confuse some people.

This PR also contains some refactorings that are unrelated to the fix, but I had already committed those. As I tested the fix, the other changes appeared not to have any influence on the charts. The refactorings are my preparation for transforming the regular counter charts (Gas, Water, etc) to the new chart functionality.


gizmocuz:
@michahagg , It's just that there are currently large 'caps' between the columns:

image

gizmocuz:
@michahagg , Maybe it's a good idea to place the 'Past usage/Trendline' on top of the columns?
It is currently behind
... Not sure if we need to zoom in the week chart, as it's a very small chart with 7 items

image

michahagg:
Hi @gizmocuz thanks for the issue reports. I've been working on the fixes and I hope I got it right this time. I'm a little confused about your week chart showing 'Return 1' and 'Return 2' without any actual data (no columns, that appear to be gaps but are just 0-height columns). 

Could you please try again? Thanks!

If you find more issues, please keep sending them, as I know you will ;-)

This is what your week chart should look like, without return:
image


I don't have data of more than one year ago, but I'm confident the trendline will show up in front of the columns. As is shown in below image.
image

michahagg:
@gizmocuz BTW I've been thinking about the layout of the charts page. I think it could improve by rearranging the week, month and year charts. Now the week, month and year chart actually contain the same information, on different zoom levels. My idea is to replace the current week chart with the current year chart (in which one can easily zoom to week and month level) and replace the current month chart with a chart that shows the amount per month instead of day and the year chart to show a column for each year.

Do you think that would be an improvement? What would you do differently?

Does this idea relate to your comment in any way? 
> ... Not sure if we need to zoom in the week chart, as it's a very small chart with 7 items

gizmocuz:
@michahagg , thanks for this !!!

The trend-line should also work for the month graphs (see the temperature month chart)

What I mean by zooming in the week chart, because it's a chart displaying the last week (in columns) , there is no reason why we can/have to zoom in this chart. Now I can use the mouse to zoom.

I think a week/month chart is nice to have, specially if you have a year of data

Are you on Slack ? Or can you give me your Forum username ?
I can send you a database containing more then a year of data



For the power charts, it could be interesting to see in the month charts the values from last month or year maybe ?

michahagg:
Hi @gizmocuz a database with real data could be very helpful! Of course I can generate simulation data, but real life data is always correct. I'm on the Forum, my username is 'michah'. I just made a Slack account as well, but I've never used it before.

rwaaren:
>a database with real data could be very helpful! 

I have sent you a PM on the forum.

gizmocuz:
@michahagg , Send you a PM/Email with P1 Meter data for multiple years

gizmocuz:
@michahagg , I'm not sure if this PR is ready to be merged ?

kiddigital:
@michahagg , would be great if this PR is stable to get merged. Is there something left to do/checked?

gizmocuz:
@michahagg , I tested the PR and made one small change (removed the option to zoom in week charts)
Unfortunately I'm not that good with Git and it created a new PR here:
https://github.com/domoticz/domoticz/pull/4553

I have merged the PR, thank you for all the work again !!

michahagg:
Thanks @gizmocuz sorry about the delay. I've not been able to spend time on this. My plan is to make some time soon.

PR #4431, opened 2020-11-04
run through clang-tidy

neheb:
With no options it found two problems.

### DIFF
--- a/hardware/GoodweAPI.cpp
+++ b/hardware/GoodweAPI.cpp
@@ -289,7 +289,8 @@ std::string getStatusString(const int status)
 	case STATUS_WAITING: return "Waiting for the Sun";
 	case STATUS_NORMAL: return "Normal/ Working";
 	case STATUS_OFFLINE: return "Offline";
-	default: return ("Unkown status value " + status);
+	default:
+		return "Unkown status value " + std::to_string(status);
 	}
 }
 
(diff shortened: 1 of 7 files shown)


gizmocuz:
@neheb , thank you for these changes !!

One question/remark, I'm a bit worried about the std::rint vs std::lrint usage
Most code is using the #define which is declared in many classes and that we might could move to stdafx.h so that every class can use it

`#define round(a) (int ) ( a + .5 )`

This functions rounds differently then rint/rlint and I suspect this is also what you would like to have in most cases
It's rounding upward (but not to compare against setting **std::fesetround(FE_UPWARD);**, which does strange things when you want to round 0.1 (to 0))
If you search in the Domoticz code, you see extreme limited use of rint / std::rint)

The actual correct way is also to c style casting and use

static_cast<int>(std::lrint(0.1))

What do you think about using the define in these cases ?

neheb:
clang-tidy did not take that define into consideration. It thought it was using the C function and added the std:: namespace.

If you want it to round upward, why not use floor and ceil?

gizmocuz:
As these functions also do not round the way it should 

floor(+2.7) = 2.000000
floor(-2.7) = -3.000000

ceil(+2.4) = 3.000000
ceil(-2.4) = -2.000000

My round define

    float tempf = 2.4f;
    int tempi = round(tempf);

prints 2, which makes more sense

And you have to cast to a interger as well, not with the define

neheb:
Removed those math changes and only fixed abs. This was detected with no clang-tidy options.

gizmocuz:
Thanks!

PR #4434, opened 2020-11-05
links to new tab

flopp999:
added  target="_blank" so links are opened in new tab


gizmocuz:
@flopp999 , thank you for your patch. Could you make the patch against our default (development) branch ?

kiddigital:
> added  target="_blank" so links are opened in new tab

Why do you want this? 

The UI is a single-page webapp. Opening new tabs creates a mess of open tabs where the content of these tabs can become out-of-sync or even conflicting...

PR #4435, opened 2020-11-05
clang-format: don't put functions on the same line

neheb:
Codebase seems to not put the function body on the same line as the
declaration.


neheb:
Hmm that's inconsistent. Some parts of the code place it on the same line.

gizmocuz:
Thank you !!!

PR #4439, opened 2020-11-07
removal of void and defaulted destructors

neheb:


gizmocuz:
@neheb, it's a bit hard to read for me (destructor = default) ... but it's correct code of course
Could you make the .clang-format file to have idents/tabs of 4 please
VStudio 2019 can also work with this file !... But... It is expecting this in the 'msbuild' folder... (because the project file resides there)
I have not found a way to specify the location sadly.... one way is to duplicate the file to this location, or to move the project file in the root folder...

neheb:
an ident siize of 4 would use spaces instead of tabs. Interestingly enough, it uses an indent size of 4 by default.

I assume a symlink can be created.

edit: yeah just tested it. First indent uses spaces, second one tabs.

gizmocuz:
@neheb , then we better leave it like it is now, it was just that in VStudio the below are tabs and is spaced on every 4th character
(In reality it is a real tab, just displayed differently, hoped that would also be the case by changing it to 4 in the config file, but also here it does not change it: https://zed0.co.uk/clang-format-configurator)

image

gizmocuz:
Thank you again !! Maybe we should also have a file with the clang-tidy options used in the project ?

neheb:
I don't use Visual Studio so no idea there.

gizmocuz:
@neheb , It's free (Community Edition)
https://visualstudio.microsoft.com/vs/

And a good editor as well ! If not, the best there is at the moment.

neheb:
@gizmocuz note that clang-tidy includes a microsoft style

```
clang-format --dump-config --style=Microsoft
```

It does not use tabs.

PR #4441, opened 2020-11-07
new links in new tab

flopp999:
when follow a link in a Hardware plugin description it will open in a new Tab, because something you want to read about the plugin when you fill-in the config. To open the link in same Tab you may lost some already filled-in config


rwaaren:
@flopp999  Maybe I miss the point but in the browsers I use it is already an option to open a link in a new tab / new window.

flopp999:
> @flopp999 Maybe I miss the point but in the browsers I use it is already an option to open a link in a new tab / new window.

the point is that the link will open in a new tab for all people that doesn't have that option

rwaaren:
@flopp999 

Searched for it but could not find a browser that does not have the option to open a page on a new tab.
Which browsers do not have this option ?

flopp999:
so you mean that you don't want to add my change to master?

When I search through the code I found lots of placed where you used target="_blank"

rwaaren:
> so you mean that you don't want to add my change to master?
That is not up to me. 

I just asked the question to help me understand what the reasoning behind this PR is. What I see in the sources that this construct is used in places where webpages in external domains are called and I do agree that it makes sense to use the same construct for external wiki or github pages.

gizmocuz:
Thank you all!

PR #4442, opened 2020-11-08
for range loop conversions

neheb:
Some of the manually done ones have a size reduction as there is no need for a copy.

### DIFF
--- a/hardware/1Wire.cpp
+++ b/hardware/1Wire.cpp
@@ -160,10 +160,10 @@ void C1Wire::SensorThread()
 		}
 
 		// Parse our devices
-		std::set<_t1WireDevice>::const_iterator itt;
-		for (itt = m_sensors.begin(); itt != m_sensors.end() && !IsStopRequested(0); ++itt)
+		for (const auto &device : m_sensors)
 		{
-			const _t1WireDevice& device = *itt;
+			if (IsStopRequested(0))
+				break;
 
 			// Manage families specificities
 			switch (device.family)
@@ -308,7 +308,7 @@ void C1Wire::BuildSensorList() {
 	m_sensors.clear();
 	m_system->GetDevices(devices);
 
-	for (const auto & device : devices)
+	for (const auto &device : devices)
 	{
 		switch (device.family)
 		{
@@ -327,7 +327,6 @@ void C1Wire::BuildSensorList() {
 		default:
 			break;
 		}
-
 	}
 	devices.clear();
 }
@@ -343,7 +342,7 @@ void C1Wire::BuildSwitchList() {
 	m_switches.clear();
 	m_system->GetDevices(devices);
 
-	for (const auto & device : devices)
+	for (const auto &device : devices)
 	{
 		switch (device.family)
 		{
@@ -360,7 +359,6 @@ void C1Wire::BuildSwitchList() {
 		default:
 			break;
 		}
-
 	}
 	devices.clear();
 }
@@ -372,10 +370,8 @@ void C1Wire::PollSwitches()
 		return;
 
 	// Parse our devices (have to test if m_TaskSwitches.IsStopRequested because it can take some time in case of big networks)
-	for (const auto & itt : m_switches)
+	for (const auto &device : m_switches)
 	{
-		const _t1WireDevice& device = itt;
-
 		// Manage families specificities
 		switch (device.family)
 		{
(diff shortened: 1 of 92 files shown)


gizmocuz:
@neheb , sure not the right place here, but 'SortIncludes' in the clang-format configuration file is giving issues as 'stdafx.h' needs to be included as first include.
Sure we can add everywhere '// clang-format off/on' but that does not seem right
Sometimes you want includes in a logical order then alphabetical order...
Maybe set it to false?

neheb:
I think there's a different way to accomplish that. Setting it to false seems correct for now.

gizmocuz:
Not sure why you choose to call it 'm' instead of 'itt' (from iterator)

for (auto& itt : m_devices)
	for (auto &m : m_devices)

'itt' seems more logical as you know it's from the iterator loop

... actually, because of my spelling error, maybe it should be 'it' ?

neheb:
Because it's not an iterator at all. It's a reference to what the iterator points to.

edit: maybe d or device is better...

gizmocuz:
Sure if it's related to devices, 'd' would also be good
Its a iterator ... "An Iterator is an object that can be used to loop through collections" does not matter how its used ;)
The thing is, if you keep renaming it, it's not really consistent and you don't know what it is... like 'm' could be a member
like 'i' is used (a lot) for indexes

neheb:
I believe m_devices is an std::map. const auto& m evaluates to std::pair<x, y>. In other words, a reference to the map member. That’s why -> is replaced with . Also note some of the other changes where *itt is completely removed as it’s not needed.

edit: and results in a smaller binary as a copy is not needed.

gizmocuz:
Yes I know that this is the result, still it's a Iteration you do over the members of a map.
calling it 'm' is not clear, specially when 'm' ... 'm_' is used for members
'**It**' (Iterator) seems more clear in some cases then '**m**'

gizmocuz:
For example:
			for (const auto &val : values)

... later in the code you can't see the for loop... and you might think like... val ... was a structure? 
when it was called 'it' (itt), you should think, ahhh it's from the for loop

kiddigital:
As the goal is to have readable code, I would say that variablenames should reflect/explain what the variable is about.

When iteration over a list, the object you are working with is the actual object (and not the list) and for the functions you are applying to it, it is usually not relevant that the object comes from a list.

So I would argue that the variablename should at least indicate the object(type).

But knowing that the object comes from a list that is iterated, can be valuable as well in understanding the purpose of the processing.

So maybe combine that?
d_it instead of it(t) or just d?

neheb:
I'll get rid of the m when I get a chance.

C++17 has structured binding declarations, which avoid having to set a variable to x.first and another to x.second. From reading the CMake file, this project is C++11 with no plans to move to a higher version.

neheb:
Renamed all &m

gizmocuz:
@neheb , we just moved C++11, C++17 is a bit to soon and could introduce some compile issues on various hardware platforms

gizmocuz:
Thank you for all the work !!

gizmocuz:
@neheb , there was a problem with at least eHouseTCP.cpp:

1>eHouseTCP.cpp
1>hardware\eHouseTCP.cpp(908): warning C4700: uninitialized local variable 'i' used

I used to be a index and was not used anymore.
VS2019 did notice this, so please if possible, use VC2019 and GCC to test the changes

gizmocuz:
@neheb , there was also a problem reported in void CEventSystem::WWWGetItemStates, this was caused by the usage of 'unsigned long' , instead of 'uint64_t)

neheb:
Those are warnings to be fair. I’ll look into them when I can.

gizmocuz:
Not really, the eHouseTCP code used a uninitialized variable as a index in a function
(Do note I am not pointing fingers!)

I just solved a crash where a iterator was still being used but this was replaced by a for-loop conversion
The iterator was not used/assigned before, but was parsed as a pointer to another function (CEventSystem::ParseBlocklyLua)
VS2019 did also not warn... because the pointer was parsed as 'const' and not assigned yet, it could/should have given a warning

https://github.com/domoticz/domoticz/commit/c301f6632e95b8cf40dbc37bb2656b8cbff61546

neheb:
:). Happens with manual conversions unfortunately. To add to that, I should probably change my editor. I'm still using nano...

gizmocuz:
@neheb , I know... it was a lot of work for sure !!!

Yes, please use VStudio 2019!, It's very good

Found another left over:
https://github.com/domoticz/domoticz/commit/d4bfea473be6ec0a5acb1226b9b372a232bcad73

Ps. If you search in all source files for '::const_iterator' ... there are a few for loop left overs ;)
(And a few when searching for '::iterator'

PR #4444, opened 2020-11-09
clang-format: base file on Microsoft style

neheb:
Seems to work better than the previous custom one.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,19 +1,22 @@
 ---
 Language:        Cpp
+# BasedOnStyle:  Microsoft
 AccessModifierOffset: -2
 AlignAfterOpenBracket: Align
 AlignConsecutiveMacros: false
 AlignConsecutiveAssignments: false
+AlignConsecutiveBitFields: false
 AlignConsecutiveDeclarations: false
 AlignEscapedNewlines: Right
-AlignOperands:   true
+AlignOperands:   Align
 AlignTrailingComments: true
 AllowAllArgumentsOnNextLine: true
 AllowAllConstructorInitializersOnNextLine: true
 AllowAllParametersOfDeclarationOnNextLine: true
+AllowShortEnumsOnASingleLine: false
 AllowShortBlocksOnASingleLine: Never
 AllowShortCaseLabelsOnASingleLine: false
-AllowShortFunctionsOnASingleLine: Empty
+AllowShortFunctionsOnASingleLine: None
 AllowShortLambdasOnASingleLine: All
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thanks!

gizmocuz:
@neheb , I just push a commit removing some unknown keys that VS2019 does not recognize
It seems now VS2019 picks up the clang-format file in the root folder.
When you press control-k and then control-d, it will format the code
Please also test with VS2019, it's free, and there should be no reason not to test against it

gizmocuz:
Question, Is it possible to tweak the file so it will not modify this:

```
result=m_sql.safe_query("INSERT INTO DeviceToPlansMap (DeviceRowID, DevSceneType, PlanID) VALUES ('%d', '%d', '%d')",
					i, 0, m_PlanID);

```

to

```
result = m_sql.safe_query(
					"INSERT INTO DeviceToPlansMap (DeviceRowID, DevSceneType, PlanID) VALUES ('%d', '%d', '%d')", i, 0,
					m_PlanID);
```

The space behind 'result' is perfect, but I see no problems putting the fields on the second line, actually this could be a good thing

Or:
`void eHouseTCP::UpdateSQLStatus(int devh, int devl, int /*devtype*/, int code, int nr, char /*signal*/, int nValue, const char  *sValue, int /*battery*/)
`
to
```
void eHouseTCP::UpdateSQLStatus(int devh, int devl, int /*devtype*/, int code, int nr, char /*signal*/, int nValue, const char *sValue,
				int /*battery*/)
```

(The first is on one line)

neheb:
That first one looks strange...

BinPackParameters either fits as many as possible on one line or each parameter gets its own line.

Adding a // NOLINT comment will disable specific sections from being formatted.

gizmocuz:
Maybe there is a parameter that says you can only have a maximum of xx characters on one line ?
I believe we can do

ColumnLimit: 0

It is currently set to 140

neheb:
Right. The idea is to have all the lines visible without needing to resort to vertical scrolling.

Linux kernel for example still uses an 80 character max...

In my workflow I typically have two windows side by side on one monitor which works out to ~157 characters visible per line.

gizmocuz:
140 is way to low these days
My normal workflow involves a 44'' Dell monitor with a 4K resolution
Linus want the 80 character limit gone

How about we set it to 200 ? This works perfectly in a IDE with a screen size of 1600x1200

neheb:
I have no objection.

PR #4448, opened 2020-11-09
AirconWithMe Update Hardware.js fixes bug reported in #4383

LarsenDX:
AirconWithMe Update Hardware.js fixes bug reported in #4383


gizmocuz:
Thanks !!

PR #4451, opened 2020-11-10
clang-tidy readability cleanups + some warnings

neheb:


gizmocuz:
What parameters are you using for clang-tidy ? Should we put them in a file in the repo as well ?

neheb:
I've listed them all in each commit message.

I normally run

```
/usr/share/clang/run-clang-tidy.py -checks=\* > err.txt
```

and grep the output.

edit: I also have a .clang-tidy file with some modifications. Nothing too big.

gizmocuz:
Pffff a lot of changes again

Some I would keep, like:

```
if (silent == false)
{
	throw;
}
```

Instead of 'Forcing':
```
if (!silent)
	throw;
```

Pretty readable with the first, and maybe someone wants to add something later between those lines

The below also is good:
```
if (strcmp(tmpstr.c_str(), "on") == 0)
{
	bSwitch = true;
}
else
{
	bSwitch = false;
}
```

Yes this as well, but should not be forced?

`bSwitch = strcmp(tmpstr.c_str(), "on") == 0;`

Or the below:

```
if (
	(hwid == "") ||
	(mode1 == "") ||
	(mode2 == "")
	)
	return;
```	

Instead of putting everything on one line like you did now:

```
if ((hwid.empty()) || (mode1.empty()) || (mode2.empty()))
	return;
```

Actually this is better, as you can easily add new lines (compare operation infront) 
```
if (
	(hwid == "")
	|| (mode1 == "")
	|| (mode2 == "")
	)
	return;
```

neheb:
The first actually has a check: http://clang.llvm.org/extra/clang-tidy/checks/readability-braces-around-statements.html

Maybe should make a specific commit that adds that.

The strcmp example is just a simplification. It's clearer that way I think.

The || in front or back is a clang-format setting.

gizmocuz:
Do you know which setting it is ?
Do you also know if there is a setting that enforces braces after the first 'if' line here:

```
		if (trecv != 0x16)
			while ((rbuffer[0x00] == 0x88) && (rbuffer[0x07] == 0x01))
			{
				totRetries++;
				trecv = recvfrom(m_RemoteSocket, (char*)&rbuffer, sizeof(rbuffer), 0, (struct sockaddr*) & si_other, &slen);
				continue;
			}
```

neheb:
Enforce no. But that clang-tidy fix I pointed to will fix that.

gizmocuz:
Hmmm I don't see that flag in any online editor (ShortStatementLines) and not in our file ?

neheb:
clang-tidy, not clang-format.

gizmocuz:
It would be great if you added a .clang-tidy file

neheb:
I'll be breaking this up.

PR #4456, opened 2020-11-12
mostly clang compiler warning fixes

neheb:
Split off from https://github.com/domoticz/domoticz/pull/4451

Should be easier to read.

### DIFF
--- a/hardware/Netatmo.cpp
+++ b/hardware/Netatmo.cpp
@@ -1358,7 +1358,7 @@ bool CNetatmo::ParseHomeStatus(const std::string &sResult)
 				{
 					std::string aName = "Status";
 					bool bIsActive = (module["boiler_status"].asString() == "true");
-					SendSwitch(moduleID & 0x00FFFFFF | 0x10000000, 1, 255, (bIsActive == true), 0, aName);
+					SendSwitch((moduleID & 0x00FFFFFF) | 0x10000000, 1, 255, bIsActive, 0, aName);
 				}
 				if (!module["battery_level"].empty())
 				{
@@ -1419,18 +1419,19 @@ bool CNetatmo::ParseHomeStatus(const std::string &sResult)
 
 				if (!room["therm_measured_temperature"].empty())
 				{
-					SendTempSensor(roomID & 0x00FFFFFF | 0x03000000, 255, room["therm_measured_temperature"].asFloat(), roomName);
+					SendTempSensor((roomID & 0x00FFFFFF) | 0x03000000, 255, room["therm_measured_temperature"].asFloat(), roomName);
 				}
 				if (!room["therm_setpoint_temperature"].empty())
 				{
-					SendSetPointSensor((uint8_t)((roomID & 0X00FF0000 | 0x02000000) >> 16), (roomID & 0XFF00) >> 8, roomID & 0XFF, room["therm_setpoint_temperature"].asFloat(), roomName);
+					SendSetPointSensor((uint8_t)(((roomID & 0x00FF0000) | 0x02000000) >> 16), (roomID & 0XFF00) >> 8, roomID & 0XFF, room["therm_setpoint_temperature"].asFloat(),
+							   roomName);
 				}
 				if (!room["therm_setpoint_mode"].empty())
 				{
 					std::string setpoint_mode = room["therm_setpoint_mode"].asString();
 					bool bIsAway = (setpoint_mode == "away");
 					std::string aName = "Away " + roomName;
-					SendSwitch(roomID & 0x00FFFFFF | 0x01000000, 1, 255, bIsAway, 0, aName);
+					SendSwitch((roomID & 0x00FFFFFF) | 0x01000000, 1, 255, bIsAway, 0, aName);
 				}
 			}
 			iDevIndex++;
--- a/hardware/OpenZWave.cpp
+++ b/hardware/OpenZWave.cpp
@@ -2300,11 +2300,7 @@ void COpenZWave::AddValue(NodeInfo* pNode, const OpenZWave::ValueID& vID)
 			else if (vOrgIndex == ValueID_Index_Clock::Minute)
 			{
 				pNode->tClockMinute = byteValue;
-				if (
-					(pNode->tClockDay != -1) &&
-					(pNode->tClockHour != -1) &&
-					(pNode->tClockMinute != -1)
-					)
+				if ((pNode->tClockDay != UINT8_MAX) && (pNode->tClockHour != UINT8_MAX) && (pNode->tClockMinute != UINT8_MAX))
 				{
 					_log.Log(LOG_NORM, "OpenZWave: NodeID: %d (0x%02x), Thermostat Clock: %s %02d:%02d", NodeID, NodeID, ZWave_Clock_Days(pNode->tClockDay), pNode->tClockHour, pNode->tClockMinute);
 					_device.intvalue = (pNode->tClockDay * (24 * 60)) + (pNode->tClockHour * 60) + pNode->tClockMinute;
@@ -2866,11 +2862,7 @@ void COpenZWave::UpdateValue(NodeInfo* pNode, const OpenZWave::ValueID& vID)
 			else if (vOrgIndex == ValueID_Index_Clock::Minute)
 			{
 				pNode->tClockMinute = byteValue;
-				if (
-					(pNode->tClockDay != -1) &&
-					(pNode->tClockHour != -1) &&
-					(pNode->tClockMinute != -1)
-					)
+				if ((pNode->tClockDay != UINT8_MAX) && (pNode->tClockHour != UINT8_MAX) && (pNode->tClockMinute != UINT8_MAX))
 				{
 					_log.Log(LOG_NORM, "OpenZWave: NodeID: %d (0x%02x), Thermostat Clock: %s %02d:%02d", NodeID, NodeID, ZWave_Clock_Days(pNode->tClockDay), pNode->tClockHour, pNode->tClockMinute);
 				}
--- a/hardware/XiaomiGateway.cpp
+++ b/hardware/XiaomiGateway.cpp
@@ -195,7 +195,9 @@ bool XiaomiGateway::WriteToHardware(const char * pdata, const unsigned char leng
 			xcmd->unitcode == XiaomiUnitCode::SELECTOR_WIRED_WALL_DUAL_CHANNEL_1 || ((xcmd->subtype == sSwitchGeneralSwitch) && (xcmd->unitcode == XiaomiUnitCode::ACT_ONOFF_PLUG))) {
 			message = "{\"cmd\":\"write\",\"model\":\"" + cmddevice + "\",\"sid\":\"158d00" + sid + "\",\"short_id\":0,\"data\":\"{\\\"" + cmdchannel + "\\\":\\\"" + cmdcommand + "\\\",\\\"key\\\":\\\"@gatewaykey\\\"}\" }";
 		}
-		else if ((xcmd->subtype == sSwitchTypeSelector) && (xcmd->unitcode >= XiaomiUnitCode::GATEWAY_SOUND_ALARM_RINGTONE && xcmd->unitcode <= XiaomiUnitCode::GATEWAY_SOUND_DOORBELL) || (xcmd->subtype == sSwitchGeneralSwitch) && (xcmd->unitcode == XiaomiUnitCode::GATEWAY_SOUND_MP3)) {
+		else if (((xcmd->subtype == sSwitchTypeSelector) && (xcmd->unitcode >= XiaomiUnitCode::GATEWAY_SOUND_ALARM_RINGTONE && xcmd->unitcode <= XiaomiUnitCode::GATEWAY_SOUND_DOORBELL)) ||
+			 ((xcmd->subtype == sSwitchGeneralSwitch) && (xcmd->unitcode == XiaomiUnitCode::GATEWAY_SOUND_MP3)))
+		{
 			std::stringstream ss;
 			if (xcmd->unitcode == XiaomiUnitCode::GATEWAY_SOUND_MP3) {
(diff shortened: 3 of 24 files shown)


hardware/Netatmo.cpp
```diff
@@ -1358,7 +1358,7 @@ bool CNetatmo::ParseHomeStatus(const std::string &sResult)
 				{
 					std::string aName = "Status";
 					bool bIsActive = (module["boiler_status"].asString() == "true");
-					SendSwitch(moduleID & 0x00FFFFFF | 0x10000000, 1, 255, (bIsActive == true), 0, aName);
+					SendSwitch((moduleID & 0x00FFFFFF) | 0x10000000, 1, 255, (bIsActive), 0, aName);
```
gizmocuz: I think we can also remove the ( ) characters here?

hardware/OpenZWave.cpp
```diff
@@ -2300,11 +2300,7 @@ void COpenZWave::AddValue(NodeInfo* pNode, const OpenZWave::ValueID& vID)
 			else if (vOrgIndex == ValueID_Index_Clock::Minute)
 			{
 				pNode->tClockMinute = byteValue;
-				if (
-					(pNode->tClockDay != -1) &&
-					(pNode->tClockHour != -1) &&
-					(pNode->tClockMinute != -1)
-					)
+				if ((pNode->tClockDay != UINT8_MAX) && (pNode->tClockHour != UINT8_MAX) && (pNode->tClockMinute != UINT8_MAX))
```
gizmocuz, line 2303: Not a fan of this presentation, the original is much clearer, but would be better if it was
if (
   (a == 1)
   || (b==1)
   || (c==1)
  )

hardware/OpenZWave.cpp
```diff
@@ -2866,11 +2862,7 @@ void COpenZWave::UpdateValue(NodeInfo* pNode, const OpenZWave::ValueID& vID)
...
```
gizmocuz, line 2865: Not a fan of this presentation, the original is much clearer, but would be better if it was
if (
(a == 1)
|| (b==1)
|| (c==1)
)

hardware/XiaomiGateway.cpp
```diff
@@ -195,7 +195,9 @@ bool XiaomiGateway::WriteToHardware(const char * pdata, const unsigned char leng
 			xcmd->unitcode == XiaomiUnitCode::SELECTOR_WIRED_WALL_DUAL_CHANNEL_1 || ((xcmd->subtype == sSwitchGeneralSwitch) && (xcmd->unitcode == XiaomiUnitCode::ACT_ONOFF_PLUG))) {
 			message = "{\"cmd\":\"write\",\"model\":\"" + cmddevice + "\",\"sid\":\"158d00" + sid + "\",\"short_id\":0,\"data\":\"{\\\"" + cmdchannel + "\\\":\\\"" + cmdcommand + "\\\",\\\"key\\\":\\\"@gatewaykey\\\"}\" }";
 		}
-		else if ((xcmd->subtype == sSwitchTypeSelector) && (xcmd->unitcode >= XiaomiUnitCode::GATEWAY_SOUND_ALARM_RINGTONE && xcmd->unitcode <= XiaomiUnitCode::GATEWAY_SOUND_DOORBELL) || (xcmd->subtype == sSwitchGeneralSwitch) && (xcmd->unitcode == XiaomiUnitCode::GATEWAY_SOUND_MP3)) {
+		else if (((xcmd->subtype == sSwitchTypeSelector) && (xcmd->unitcode >= XiaomiUnitCode::GATEWAY_SOUND_ALARM_RINGTONE && xcmd->unitcode <= XiaomiUnitCode::GATEWAY_SOUND_DOORBELL)) ||
+			 ((xcmd->subtype == sSwitchGeneralSwitch) && (xcmd->unitcode == XiaomiUnitCode::GATEWAY_SOUND_MP3)))
+		{
 			std::stringstream ss;
```
gizmocuz, line 201: would be better if it was
if (
(a == 1)
|| (b==1)
|| (c==1)
)

hardware/openwebnet/bt_openwebnet.cpp
```diff
@@ -896,27 +896,19 @@ bool bt_openwebnet::IsEqual(const bt_openwebnet& msg_to_compare)
...
-        (msg_to_compare.Extract_what().compare(m_what) == 0) &&
-        (msg_to_compare.Extract_where().compare(m_where) == 0) &&
-        (msg_to_compare.Extract_when().compare(m_when) == 0) &&
-        (msg_to_compare.Extract_dimension().compare(m_dimension) == 0))
-      return true;
-    else
-      return false;
+	  if ((msg_to_compare.Extract_who() == m_who) && (msg_to_compare.Extract_what() == m_what) && (msg_to_compare.Extract_where() == m_where) && (msg_to_compare.Extract_when() == m_when) &&
+	      (msg_to_compare.Extract_dimension() == m_dimension))
+		  return true;
```
gizmocuz, line 901: would be better if it was
if (
(a == 1)
|| (b==1)
|| (c==1)
)

hardware/openwebnet/bt_openwebnet.cpp
```diff
@@ -896,27 +896,19 @@ bool bt_openwebnet::IsEqual(const bt_openwebnet& msg_to_compare)
...
   else
   {
-    if ((msg_to_compare.Extract_who().compare(m_who) == 0) &&
-        (msg_to_compare.Extract_what().compare(m_what) == 0) &&
-        (msg_to_compare.Extract_where().compare(m_where) == 0) &&
-        (msg_to_compare.Extract_level().compare(m_level) == 0) &&
-        (msg_to_compare.Extract_interface().compare(m_sInterface) == 0) &&
-        (msg_to_compare.Extract_when().compare(m_when) == 0) &&
-        (msg_to_compare.Extract_dimension().compare(m_dimension) == 0))
-      return true;
```
gizmocuz, line 917: if (
(a == 1)
|| (b==1)
|| (c==1)
)

hardware/openwebnet/bt_openwebnet.cpp
```diff
@@ -896,27 +896,19 @@ bool bt_openwebnet::IsEqual(const bt_openwebnet& msg_to_compare)
...
-    else
-      return false;
+	  if ((msg_to_compare.Extract_who() == m_who) && (msg_to_compare.Extract_what() == m_what) && (msg_to_compare.Extract_where() == m_where) && (msg_to_compare.Extract_level() == m_level) &&
+	      (msg_to_compare.Extract_interface() == m_sInterface) && (msg_to_compare.Extract_when() == m_when) && (msg_to_compare.Extract_dimension() == m_dimension))
+		  return true;
```
gizmocuz, line 909: would be better if it was
if (
(a == 1)
|| (b==1)
|| (c==1)
)

hardware/plugins/Plugins.cpp
```diff
@@ -833,8 +833,7 @@ namespace Plugins {
 
 	void CPlugin::Notifier(std::string Notifier)
 	{
-		if (m_Notifier)
-			delete m_Notifier;
+		delete m_Notifier;
```
gizmocuz, line 836: Is this safe ? Maybe StopHardware was called before...
Not knowing what the feature of code will bring, It's safer to check if this pointer is set
(If possible, I would also rename this variable to m_pNotifier)

main/NotificationSystem.cpp
```diff
@@ -135,14 +135,8 @@ bool CNotificationSystem::Unregister(CNotificationObserver* pNotifier)
...
 		boost::unique_lock<boost::shared_mutex> lock(m_mutex);
-		for (size_t i = 0; i < m_notifiers.size(); i++)
-		{
-			if (m_notifiers[i] == pNotifier)
-			{
-				m_notifiers.erase(m_notifiers.begin() + i);
-				return true;
-			}
-		}
+		m_notifiers.erase(std::remove_if(m_notifiers.begin(), m_notifiers.end(), [=](CNotificationObserver *notifier) { return notifier == pNotifier; }), m_notifiers.end());
```
gizmocuz: Not sure what is clearer to read ?
Maybe better to not write everything on a single line

main/NotificationSystem.cpp
```diff
@@ -135,14 +135,8 @@ bool CNotificationSystem::Unregister(CNotificationObserver* pNotifier)
...
 	}
-	return false;
+
+	return std::any_of(m_notifiers.begin(), m_notifiers.end(), [=](CNotificationObserver *notifier) { return notifier == pNotifier; });
 }
```
gizmocuz: I think with the two changes it is actually more code, and not very readable for beginners

main/WebServer.cpp
```diff
@@ -6784,7 +6784,7 @@ namespace http {
 						root["status"] = "OK";
 						root["title"] = "LearnSW";
 						root["ID"] = m_sql.m_LastSwitchID;
-						root["idx"] = m_sql.m_LastSwitchRowID;
+						root["idx"] = Json::Value::UInt64(m_sql.m_LastSwitchRowID);
```
gizmocuz, line 6787: I don't have any warnings here...
It's already a uint64_t 
Is it the JsonCPP version maybe ? As mine Json::Value::UInt64 points to a unsigned __int64 -> uint64_t

tcpserver/TCPClient.cpp
```diff
@@ -17,7 +17,7 @@ CTCPClientBase::CTCPClientBase(CTCPServerIntBase *pManager)
 
 CTCPClientBase::~CTCPClientBase()
 {
-	if (socket_) delete socket_;
```
gizmocuz, line 20: Not sure what the feature will bring... better to check if the pointer is set

webserver/connection.cpp
```diff
@@ -303,8 +303,8 @@ namespace http {
 
 			if (sendfile_.is_open())
 				sendfile_.close();
-			if (send_buffer_)
```
gizmocuz, line 306: Also here, just check if the pointer is set (and in the below code

gizmocuz (commented):
Some small comments and questions

gizmocuz:
Just some small comments/questions
Yes I know it should be safe to just delete a nullptr, but what about when using a memory management tool to check for memory leaks.
This overwrites the new/delete functions
But I know it should be safe... It's maybe just the idea

main/WebServer.cpp
neheb: I think so. Either that or GCC8 is too stupid to figure it out.

hardware/OpenZWave.cpp
neheb: Ehhh that can sort of be done with //, but I don't see the reason for it. The conditions are pretty short.

neheb:
That's a problem with those tools.

delete is not like free in C. it already checks for nullptr.

main/NotificationSystem.cpp
neheb: It reduces it to https://en.wikipedia.org/wiki/Erase%E2%80%93remove_idiom

which I think is easier to reason about.

main/NotificationSystem.cpp
neheb: There's no size difference that I can see in the binary. I think both GCC and clang are smart enough to figure out that the algos are over the same range.

hardware/Netatmo.cpp
neheb: Yeah I'll do that when I can.

main/NotificationSystem.cpp
gizmocuz: True, but for readability I would say it does not make things clearer
The for loop could be rewritten to 'auto', but the erase is a- not clear to read, b-does not return true if erased
First the possible item is erased, then you check if the item is in the array

main/NotificationSystem.cpp
neheb: Removed anyway.

hardware/Netatmo.cpp
neheb: Done.

hardware/XiaomiGateway.cpp
gizmocuz: Is there a clang-format option that we can use to accomplish the above 'if' check ?
It's way more cleaner and you have a much better overview
Also adding checks if much easier

neheb: No.

I've never seen this format before domoticz.

clang-format can be selectively disabled with //NOLINT.

hardware/openwebnet/bt_openwebnet.cpp
gizmocuz: The original check looks much cleaner, except it used string.compare, and i would place && at the start

hardware/plugins/PluginTransports.cpp
```diff
@@ -527,7 +527,7 @@ namespace Plugins {
 
 				m_Socket = new boost::asio::ip::udp::socket(ios, boost::asio::ip::udp::endpoint(boost::asio::ip::udp::v4(), iPort));
 				m_Socket->set_option(boost::asio::ip::udp::socket::reuse_address(true));
-				if ((m_IP.substr(0, 4) >= "224.") && (m_IP.substr(0, 4) <= "239.") || (m_IP.substr(0, 4) == "255."))
+				if (((m_IP.substr(0, 4) >= "224.") && (m_IP.substr(0, 4) <= "239.")) || (m_IP.substr(0, 4) == "255."))
 				{
```
gizmocuz, line 531: If we had a forces format of:

```
if (
	(
		(m_IP.substr(0, 4) >= "224.")
		&& (m_IP.substr(0, 4) <= "239.")
	)
	|| (m_IP.substr(0, 4) == "255.")
   )
   {
   ......
   }

```  

The original author might have seen this coming ;)

hardware/XiaomiGateway.cpp
gizmocuz: Shoot... would have been a great option

gizmocuz:
Thanks again for all hard work !!

PR #4464, opened 2020-11-15
fix compilation with newer Python

neheb:
error: cannot initialize a member subobject of type 'Py_ssize_t'
(aka 'long') with an rvalue of type 'nullptr_t'


gizmocuz:
Thanks!!

PR #4465, opened 2020-11-15
clang-tidy: mostly readability stuff

neheb:


hardware/CurrentCostMeterTCP.h
```diff
@@ -2,13 +2,13 @@
 
 #include "CurrentCostMeterBase.h"
 
-class CurrentCostMeterTCP: public CurrentCostMeterBase
+class CurrentCostMeterTCP : public CurrentCostMeterBase
 {
-public:
-	CurrentCostMeterTCP(const int ID, const std::string &IPAddress, const unsigned short usIPPort);
+      public:
+	CurrentCostMeterTCP(int ID, const std::string &IPAddress, unsigned short usIPPort);
```
gizmocuz, line 8: Maybe I am missing something, but what is wrong declaring a function parameter with const ?

neheb: From https://clang.llvm.org/extra/clang-tidy/checks/readability-const-return-type.html

Such use of const is usually superfluous, and can prevent valuable compiler optimizations.

gizmocuz:
@neheb , I think you can also take it a bit to far....?
Could you undo:
- simplify booleans, actually when you code/debug for a long period, you might overlook the exclamation mark
- use emplace_back
- use std::move

hardware/CurrentCostMeterTCP.h
gizmocuz: Thats for "const returns", I'm referring to the use of 'const' in the function parameters

CurrentCostMeterTCP(const int ID, const std::string &IPAddress, const unsigned short usIPPort);

why could we not declare 'int ID' as const ? or 'usIPPort' ?

neheb: Ah. http://clang.llvm.org/extra/clang-tidy/checks/readability-avoid-const-params-in-decls.html

Looks like there's no real need for it.

neheb:
Removed those.

gizmocuz:
Thank you again !!

PR #4473, opened 2020-11-18
Log button fix

Olifant1990:
Evohome temperature device has a log too, but button was not clickable.
Also remove log action button if there is no log link


gizmocuz:
Thank you !!

PR #4476, opened 2020-11-19
Revert "clang-tidy: replace bind calls with lambdas"

gizmocuz:
Reverts domoticz/domoticz#4417

Needed because it crashed in

connection.cpp
line 210

```
			else {
				// Perform plain read
				socket_->async_read_some(buf, [this](const boost::system::error_code &err, size_t bytes) { shared_from_this()->handle_read(err, bytes); });
			}

```

### DIFF
--- a/hardware/ASyncSerial.cpp
+++ b/hardware/ASyncSerial.cpp
@@ -33,6 +33,7 @@
 #include <algorithm>
 #include <iostream>
 #include <boost/asio.hpp>
+#include <boost/bind/bind.hpp>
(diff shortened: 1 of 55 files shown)


wijagels:
The reason that code is crashing is because `shared_from_this()` needs to be in the capture list, not inside the lambda.

`socket_->async_read_some(buf, [this, self = shared_from_this()](const boost::system::error_code &err, size_t bytes) { handle_read(err, bytes); });`

PR #4495, opened 2020-11-29
Revert 4476 and fix shared_from_this() usage

wijagels:
This fixes some bugs in #4417 and restores the changes to replace bind with lambdas

### DIFF
--- a/hardware/ASyncTCP.cpp
+++ b/hardware/ASyncTCP.cpp
@@ -1,11 +1,8 @@
 #include "stdafx.h"
 #include "ASyncTCP.h"
 #include <boost/asio.hpp>
-#include <boost/bind/bind.hpp>
 #include <boost/system/error_code.hpp>     // for error_code
 #include "../main/Logger.h"
-
-//using namespace boost::placeholders;
 
 struct hostent;
 
@@ -65,14 +62,14 @@ void ASyncTCP::connect(const std::string& ip, uint16_t port)
 	// RK: After the reset, we need to provide it work anew
 	mTcpwork = std::make_shared<boost::asio::io_service::work>(mIos);
 	if (!mTcpthread)
-		mTcpthread = std::make_shared<std::thread>(boost::bind(&boost::asio::io_service::run, &mIos));
+		mTcpthread = std::make_shared<std::thread>([this] { mIos.run(); });
 
 	mIp = ip;
 	mPort = port;
 	std::string port_str = std::to_string(port);
 	boost::asio::ip::tcp::resolver::query query(ip, port_str);
 	timeout_start_timer();
-	mResolver.async_resolve(query, boost::bind(&ASyncTCP::cb_resolve_done, this, boost::asio::placeholders::error, boost::asio::placeholders::iterator));
+	mResolver.async_resolve(query, [this](const boost::system::error_code &err, boost::asio::ip::tcp::resolver::iterator iter) { cb_resolve_done(err, iter); });
 }
 
 void ASyncTCP::cb_resolve_done(const boost::system::error_code& error, boost::asio::ip::tcp::resolver::iterator endpoint_iterator)
@@ -101,13 +98,12 @@ void ASyncTCP::connect_start(boost::asio::ip::tcp::resolver::iterator& endpoint_
 	{
 		// we reset the ssl socket, because the ssl context needs to be reinitialized after a reconnect
 		mSslSocket.reset(new boost::asio::ssl::stream<boost::asio::ip::tcp::socket>(mIos, mContext));
-		mSslSocket->lowest_layer().async_connect(mEndPoint,
-			boost::bind(&ASyncTCP::cb_connect_done, this, boost::asio::placeholders::error, endpoint_iterator));
+		mSslSocket->lowest_layer().async_connect(mEndPoint, [this, &endpoint_iterator](const boost::system::error_code &err) mutable { cb_connect_done(err, endpoint_iterator); });
 	}
 	else
 #endif
 	{
-		mSocket.async_connect(mEndPoint, boost::bind(&ASyncTCP::cb_connect_done, this, boost::asio::placeholders::error, endpoint_iterator));
+		mSocket.async_connect(mEndPoint, [this, &endpoint_iterator](const boost::system::error_code &err) mutable { cb_connect_done(err, endpoint_iterator); });
 	}
 }
 
@@ -121,9 +117,7 @@ void ASyncTCP::cb_connect_done(const boost::system::error_code& error, boost::as
 		if (mSecure) 
 		{
 			timeout_start_timer();
-			mSslSocket->async_handshake(boost::asio::ssl::stream_base::client,
-				boost::bind(&ASyncTCP::cb_handshake_done, this,
-					boost::asio::placeholders::error));
+			mSslSocket->async_handshake(boost::asio::ssl::stream_base::client, [this, &error](const boost::system::error_code &) { cb_handshake_done(error); });
 		}
 		else
 #endif
@@ -168,7 +162,7 @@ void ASyncTCP::reconnect_start_timer()
 		mIsReconnecting = true;
 
 		mReconnectTimer.expires_from_now(boost::posix_time::seconds(mReconnectDelay));
-		mReconnectTimer.async_wait(boost::bind(&ASyncTCP::cb_reconnect_start, this, boost::asio::placeholders::error));
+		mReconnectTimer.async_wait([this](const boost::system::error_code &err) { cb_reconnect_start(err); });
 	}
 }
 
@@ -211,7 +205,7 @@ void ASyncTCP::disconnect(const bool silent)
 
 	try
 	{
-		mIos.post(boost::bind(&ASyncTCP::do_close, this));
+		mIos.post([this] { do_close(); });
 	}
 	catch (...)
 	{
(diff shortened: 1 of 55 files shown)


neheb:
Interesting...

I think this is why the check is limited to C++14.

gizmocuz:
Thank you both for the patches !

kofec:
Can you check if you have issue with stoping domoticz (mine is in Openwrt):
```
2020-12-03 14:25:30.148 Status: Domoticz V2020.2 (build 12719) (c)2012-2020 GizMoCuz
2020-12-03 14:25:30.150 Status: Build Hash: 126d4d59, Date: 2020-12-03 11:19:29):
```

```
Thu Dec  3 14:09:29 2020 daemon.info domoticz: Domoticz is exiting...
Thu Dec  3 14:09:30 2020 daemon.notice domoticz: Closing application!...
Thu Dec  3 14:09:30 2020 daemon.notice domoticz: Stopping worker...
Thu Dec  3 14:09:30 2020 daemon.notice domoticz: RxQueue: queue worker stopped...
Thu Dec  3 14:09:30 2020 daemon.notice domoticz: WebServer(HTTP) stopped
Thu Dec  3 14:09:30 2020 daemon.notice domoticz: TCPServer: shared server stopped
Thu Dec  3 14:09:30 2020 daemon.notice domoticz: Stopping all hardware...
Thu Dec  3 14:09:30 2020 daemon.notice domoticz: HardwareMonitor: Hardware Monitor: Stopped...
Thu Dec  3 14:09:30 2020 daemon.notice domoticz: MySensors: Worker stopped...
Thu Dec  3 14:09:30 2020 daemon.notice domoticz: MQTT: Worker stopped...
Thu Dec  3 14:09:30 2020 daemon.notice mosquitto[1962]: Socket error on client Domoticzc09f027d-b6fd-43a3-bf87-cb16979f1e52, disconnecting.
Thu Dec  3 14:09:30 2020 daemon.notice domoticz: Scheduler stopped...
Thu Dec  3 14:09:30 2020 daemon.notice domoticz: EventSystem: Stopped...
Thu Dec  3 14:09:30 2020 daemon.notice domoticz: EventSystem: Queue thread stopped...
Thu Dec  3 14:09:30 2020 daemon.err domoticz: Domoticz(pid:3588, tid:3588('-')) received fatal signal 11 (Segmentation fault)
Thu Dec  3 14:09:30 2020 daemon.err domoticz: siginfo address=0x8, address=0xb6d66110
Thu Dec  3 14:09:30 2020 daemon.err domoticz: Failed to start gdb, will use backtrace() for printing stack frame
```

gizmocuz:
Also crashes here in 1 minute:

```
2020-12-03 16:42:54.903  Error: Domoticz(pid:24967, tid:24988('Domoticz Remote')) received fatal signal 6 (Aborted)
2020-12-03 16:42:54.903  Error: siginfo address=0x6187, address=0xffffffff
2020-12-03 16:42:54.909  Status: EventSystem: reset all device statuses...
2020-12-03 16:42:55.095  Status: Python EventSystem: Initalizing event module.
2020-12-03 16:42:55.096  Status: EventSystem: Started
2020-12-03 16:42:55.096  Status: EventSystem: Queue thread started...
2020-12-03 16:42:55.312  Status: PluginSystem: Entering work loop.
2020-12-03 16:42:57.443  Error: Thread 17 (Thread 0xac9fe220 (LWP 24988)):
2020-12-03 16:42:57.444  Error: #0  __waitpid (options=0, stat_loc=0xac9fd150, pid=24998) at ../sysdeps/unix/sysv/linux/waitpid.c:30
2020-12-03 16:42:57.444  Error: #1  __waitpid (pid=24998, stat_loc=0xac9fd150, options=0) at ../sysdeps/unix/sysv/linux/waitpid.c:25
2020-12-03 16:42:57.444  Error: #2  0x002b2494 in dumpstack_gdb(bool) ()
2020-12-03 16:42:57.444  Error: #3  0x002b2a60 in signal_handler(int, siginfo_t*, void*) ()
2020-12-03 16:42:57.444  Error: #4  <signal handler called>
2020-12-03 16:42:57.445  Error: #5  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
2020-12-03 16:42:57.445  Error: #6  0xb6978230 in __GI_abort () at abort.c:79
2020-12-03 16:42:57.445  Error: #7  0x009b5d08 in __gnu_cxx::__verbose_terminate_handler() ()
2020-12-03 16:42:57.445  Error: #8  0x009190bc in __cxxabiv1::__terminate(void (*)()) ()
2020-12-03 16:42:57.445  Error: #9  0x00919130 in std::terminate() ()
2020-12-03 16:42:57.445  Error: #10 0x00923be0 in __cxa_throw ()
2020-12-03 16:42:57.446  Error: #11 0x009513b0 in std::__throw_logic_error(char const*) ()
2020-12-03 16:42:57.446  Error: #12 0x003e27d8 in void std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::_M_construct<char*>(char*, char*, std::forward_iterator_tag) [clone .constprop.1143] ()
2020-12-03 16:42:57.446  Error: #13 0x003e8128 in ASyncTCP::cb_write_queue(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) ()
2020-12-03 16:42:57.446  Error: #14 0x003e8378 in boost::asio::detail::completion_handler<ASyncTCP::write(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)::{lambda()#1}>::do_complete(void*, boost::asio::detail::scheduler_operation*, boost::system::error_code const&, unsigned int) ()
2020-12-03 16:42:57.446  Error: #15 0x003eeddc in boost::asio::detail::strand_service::do_complete(void*, boost::asio::detail::scheduler_operation*, boost::system::error_code const&, unsigned int) ()
2020-12-03 16:42:57.447  Error: #16 0x003da934 in boost::asio::detail::scheduler::run(boost::system::error_code&) ()
2020-12-03 16:42:57.447  Error: #17 0x003e3838 in std::thread::_State_impl<std::thread::_Invoker<std::tuple<ASyncTCP::connect(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, unsigned short)::{lambda()#1}> > >::_M_run() ()
2020-12-03 16:42:57.447  Error: #18 0x0098c20c in execute_native_thread_routine ()
2020-12-03 16:42:57.447  Error: #19 0xb6b85494 in start_thread (arg=0xac9fe220) at pthread_create.c:486
2020-12-03 16:42:57.447  Error: #20 0xb6a38578 in ?? () at ../sysdeps/unix/sysv/linux/arm/clone.S:73 from /lib/arm-linux-gnueabihf/libc.so.6
```
Reverting PR and checking if that solves the issue. Seems that these changes need to be tested and tested

wijagels:
Huh... Guess that wasn't the only issue then, apologies for introducing bugs.

gizmocuz:
No worries ! Maybe this PR is to huge and needs to be split into multi PR's (that have been tested)

hardware/ASyncTCP.cpp
```diff
@@ -299,7 +285,7 @@ void ASyncTCP::write(const std::string& msg)
 {
 	if (!mTcpthread) return;
 
-	mSendStrand.post(boost::bind(&ASyncTCP::cb_write_queue, this, msg));
+	mSendStrand.post([&]() { cb_write_queue(msg); });
```
wijagels, line 288: This is invalid asio, `msg` will go out of scope before the lambda is executed.

wijagels:
I'll let someone else fix up this change since I have no good way of testing, basically the idea is anywhere there's a lambda passed to asio you must ensure anything captured by the lambda is valid until the lambda is executed.  So anywhere there's a capture list of `[&]` is almost certainly wrong.  A simple fix would be to change that to `[=]` to capture by value.

gizmocuz:
I can confirm that all is running OK now with this PR reverted.
I'm afraid I am not that deep into lamda's and bindings.... using lambda's mainly for sorting

kofec:
I hope that someone will continue these changes

neheb:
@wijagels I may have done that because otherwise it would not compile.

Anyway, capturing references by reference (eg. const std::string&) should be fine. I'll do some investigation when I get the time.

edit: hmmm [&] is equivalent to [this, &msg] for that lambda.

wijagels:
Here's a quick demonstration of why `[&]` is problematic for asio: https://gcc.godbolt.org/z/z53Y5G

The old boost::bind code behaves pretty much like `[=]` would, since everything is passed by value (unless you use boost::ref()).  This is very important because calls to `post()` or `async_*()` return before the lambda is ever executed, so whatever string you passed in could be long gone by the time the asio worker gets around to running your code.  Even if it's capturing a reference by reference, there's no way to keep that reference alive until the lambda is called.

So pretty much in all cases with asio, using capture by reference or capturing `this` without using `shared_from_this()` is unsafe.  An exception to that rule is any cases where `this` or whatever you're capturing is guaranteed to be alive for longer than the io operation (for example in hardware/ASyncTCP.cpp ASyncTCP owns the io service).

neheb:
Nasty. Thanks for helping me understand what's going wrong. It's unfortunate that this is C++11 and we don't have string_view where we can just pass everything by value.

You may have noticed that I tend to favour capturing this explicitly instead of using a capture default. Well, the latter is deprecated in C++20. And using [this, =] fails in earlier standards.

As for capturing this without using shared_from_this(), I believe this is possible with C++14 lambdas. But because I tried to make it compatible with C++11 this happened. No auto in lambdas is also quite annoying.

wijagels:
Well string_view actually makes the problem even worse!  If you wrote a function like this, it would work okay:
```
void doStuff(const std::string& m) {
    io.post([=] () { doThing(m); });
}
```

But change it to use string_view and suddenly you're capturing a type that doesn't own any memory, so this would crash (or at least be undefined) only on the string_view version:
```
{
    std::string s = "hello, world!";
    doStuff(s);
} // s goes out of scope
```

C++14 does make life a bit easier since you can write your lambdas with initializers, so you don't need to declare a `self` variable at all: `[this, _ = shared_from_this()] () { /* do something with this */ }`

neheb:
No I mean even using string_view for the function parameter.

Yeah that's what I meant. clang-tidy does exactly that. I should probably show an example...

gizmocuz:
@neheb , actually I don't mind lifting this to C++14, as now all major distributions have a GCC version supporting this.

neheb:
This is correct. Even CentOS 7 (supported until 2024) supports C++14. I have no idea what the situation is with MSVC.

gizmocuz:
VS2019 supports C++17, Raspbian distribution supports C++14 since stretch. Not sure about OpenWRT, but I assume it must be
So don't stick to c++11 coding if you could improve with c++14. We probably have to adjust the CMakeList.txt project file as well

neheb:
OpenWrt uses GCC7, so yes. Older OpenWrt uses GCC5 so also yes.

PR #4498, opened 2020-11-30
Make CustomImage of Utility devices visible to javascript and use it …

ultrasuperpingu:
…in javascript

### DIFF
--- a/main/WebServer.cpp
+++ b/main/WebServer.cpp
@@ -11088,7 +11088,7 @@ namespace http {
 							sprintf(szData, "%g%%", atof(sValue.c_str()));
 							root["result"][ii]["Data"] = szData;
 							root["result"][ii]["HaveTimeout"] = bHaveTimeout;
-							root["result"][ii]["Image"] = "Computer";
+							//root["result"][ii]["Image"] = "Computer";
 							root["result"][ii]["TypeImg"] = "hardware";
 						}
 						else if (dSubType == sTypeWaterflow)
@@ -11452,6 +11452,14 @@ namespace http {
 						}
 						break;
 						}
+					}
+					if(CustomImage != 0 && !root["result"][ii].isMember("Image"))
+					{
+							auto ittIcon = m_custom_light_icons_lookup.find(CustomImage);
+							if (ittIcon != m_custom_light_icons_lookup.end())
+							{
+								root["result"][ii]["Image"] = m_custom_light_icons[ittIcon->second].RootFile;
+							}
 					}
 #ifdef ENABLE_PYTHON
 					if (pHardware != nullptr)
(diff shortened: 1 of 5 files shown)


gizmocuz:
@ultrasuperpingu , thank you for this PR!
Quick comment, and this has nothing todo with your code, but could maybe be great to apply as well...
In the javascript code there is currently a lot of duplicate code like

```
if(item.CustomImage == 0)
      imagehtml = '<a href="#/Devices/' + item.idx + '/Log"><img src="images/Gas48.png" class="lcursor" height="40" width="40"></a></td>\n';
else
      imagehtml = '<a href="#/Devices/' + item.idx + '/Log"><img src="images/' + item.Image +'48_On.png" class="lcursor" height="40" width="40"></a></td>\n';
```

Could you do instead a short if/then like: (pseudo code, hope it's error free!)

```
item.Image = (item.CustomImage == 0)  ? 'Gas48.png' : item.Image + '48_On.png';
imagehtml = '<a href="#/Devices/' + item.idx + '/Log"><img src="images/' + item.Image +'" class="lcursor" height="40" width="40"></a></td>\n';

```

ultrasuperpingu:
Ok, it's done

gizmocuz:
You're fast ! Hope to test this PR shortly !

gizmocuz:
@ultrasuperpingu , Is it correct that I don't see the code where a user can select a custom image ? (Like you can when you edit a Switch)

ultrasuperpingu:
Yes, it's correct. The idea is to allow plugins to change the image of their devices. Users still can't.

ultrasuperpingu:
Do you want me to modify something ? Or to provide you something to be able to test it easyly ? (in fact, to test, I just added on a dummy db file all the utility sensors availables in Dummy hardware and added and removed CustomImage in db via sql)

gizmocuz:
@ultrasuperpingu , no I just need to test this. The only pity here is that this is only for plugin developers and a normal user is not able to change to a custom image.
Hope to test this soon

ultrasuperpingu:
You're right, it should be easy to add the UI to be able to change the custom image. But this commit is part of the job. I guess I'll do it later, it shouldn't be a huge work (just some ui and update the ajax request).

jeroenkl:
> You're right, it should be easy to add the UI to be able to change the custom image. But this commit is part of the job. I guess I'll do it later, it shouldn't be a huge work (just some ui and update the ajax request).

@ultrasuperpingu, many users will like this feature, if you have time, could you add it also?
thxs!

ultrasuperpingu:
@jeroenkl, yes, as I said, I'll probably do it later, when I'll have some time and motivation. But this could be done in another PR and if possible, I'd like this PR to be reviewed before I do it.

gizmocuz:
Sorry about the delay! Thanks for the PR all looking good!

PR #4499, opened 2020-12-01
Add daikin energy usage

jgaalen:
Add Daikin Energy usage as kWh meter


gizmocuz:
Thank you for this addon !!

PR #4515, opened 2020-12-10
manual constexpr conversions

neheb:

Small size change:

```
Before
10138120
8251920
After
10136272
8250096
```

### DIFF
--- a/hardware/1Wire.cpp
+++ b/hardware/1Wire.cpp
@@ -506,7 +506,7 @@ void C1Wire::ReportCounter(const std::string& deviceId, const int unit, const un
 	unsigned char deviceIdByteArray[DEVICE_ID_SIZE] = { 0 };
 	DeviceIdToByteArray(deviceId, deviceIdByteArray);
 
-	SendMeterSensor(deviceIdByteArray[0], deviceIdByteArray[1] + unit, 255, (const float)counter / 1000.0f, "Counter");
+	SendMeterSensor(deviceIdByteArray[0], deviceIdByteArray[1] + unit, 255, (const float)counter / 1000.0F, "Counter");
 }
 
 void C1Wire::ReportVoltage(const std::string& /*deviceId*/, const int unit, const int voltage)
--- a/hardware/1Wire/1WireByKernel.cpp
+++ b/hardware/1Wire/1WireByKernel.cpp
@@ -280,7 +280,7 @@ float C1WireByKernel::GetTemperature(const _t1WireDevice& device) const
 	case programmable_resolution_digital_thermometer:
 	{
 		DeviceCollection::const_iterator devIt = m_Devices.find(device.devid);
-		return (devIt == m_Devices.end()) ? -1000.0f : (*devIt).second->m_Temperature;
+		return (devIt == m_Devices.end()) ? -1000.0F : (*devIt).second->m_Temperature;
 	}
 	default: // Device not supported in kernel mode (maybe later...), use OWFS solution.
 		return -1000.0;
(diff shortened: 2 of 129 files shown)


neheb:
Getting rid of boost::assign brought it down to

```
10130496
8246000
```

neheb:
Got it to

```
10120712
8236560
```

I think I'll stop here.

gizmocuz:
Is it possible to check the build logs, specifically this one:
https://ci.appveyor.com/project/gizmocuz/domoticz/builds/36994785

There are some errors caused by this commit

neheb:
I think I know what the problem is. auto x = std::array and std::array x are not the same in C++11... What a mess.

neheb:
https://youtu.be/7DTlWPgX6zs?t=1176

Nasty.

neheb:
github actions failure is unrelated to this PR.

gizmocuz:
That is correct, but I just restarted the AppVeyor build and there are some errors:

https://ci.appveyor.com/project/gizmocuz/domoticz/builds/37392136

Mochad.cpp
quite a few of these:
1>MochadTCP.cpp(256,10): error C2065: 't': undeclared identifier
1>MochadTCP.cpp(259,7): error C2065: 't': undeclared identifier

Maybe that block needs to be in the for-loop ?

1>OnkyoAVTCP.cpp(582): warning C4700: uninitialized local variable 'i' used

The rest is perfect.

Could you try to compile under Visual Studio 2019 as well ? (community edition is free)

gizmocuz:
Looks like t needs to be declared outside the for-loop

neheb:
It’s what happens when force pushing and not compiling :). In any case, my recent changes exposed some problematic things with the original changes. I’ll fix in a day or two:

gizmocuz:
No worries ! Glad your helping improving the code !

neheb:
k. all relevant warnings fixed.

gizmocuz:
Thank you again for this PR and your time!!

xbeaudouin:
Seems I have some issue on FreeBSD 12.2 👍 
```
/home/kiwi/domoticz.dev/main/domoticz.cpp:104:57: error: constexpr variable 'facilities' must be initialized by a constant expression
        constexpr std::array<std::pair<const char *, int>, 10> facilities{
                                                               ^         ~
/home/kiwi/domoticz.dev/main/domoticz.cpp:106:4: note: non-constexpr constructor 'pair<char const (&)[7], int, false>' cannot be used in a constant expression
                        { "daemon", LOG_DAEMON }, //
                        ^
/usr/include/c++/v1/utility:444:5: note: declared here
    pair(_U1&& __u1, _U2&& __u2)
    ^

```

neheb:
oh. pair is constexpr in C++14...

xbeaudouin:
Well :  clang -v
FreeBSD clang version 10.0.1 (git@github.com:llvm/llvm-project.git llvmorg-10.0.1-0-gef32c611aa2)
Target: x86_64-unknown-freebsd12.2
Thread model: posix
InstalledDir: /usr/bin

neheb:
2 solutions for right now.

Change the C++ standard to 14, or change every constexpr pair to const.

@gizmocuz can we target C++14?

gizmocuz:
I see no problems in moving to C++14
@xbeaudouin , are you able to compile c++14 code on FreeBSD or do you forsee a issue ?

xbeaudouin:
According to https://clang.llvm.org/cxx_status.html it should works. Maybe add a check on CMakeFile to check if the compiler can hande C++14 ?

```
Clang 3.4 and later implement all of the ISO C++ 2014 standard.

You can use Clang in C++14 mode with the -std=c++14 option (use -std=c++1y in Clang 3.4 and earlier).
```

gizmocuz:
Sure, we already have a C==11 check
We use CMake 3.19 now these says

@neheb , could you also make a patch for the CMakeFile.txt ?

neheb:
Funny. CXX_STANDARD is only set for clang compilers. Otherwise what gets used is whatever the compiler defaults to. GCC since version 7 defaults to C++14 I believe. MSVC is similar. Which explains why the buildbot never failed.

PR #4518, opened 2020-12-13
Python plugin : make python plugins thread capable without blocked GIL

badzz:
When doing multi thread python module, if one python thread updated a domoticz object, it will take the GIL and if the domoticz is blocked (let say by a database backup), then all the other module thread stops running as the GIL is locked.

This fix will allow python other threads to use the GIL when we do a potentially blocking action like accessing sqllite.


gizmocuz:
Thank you for this PR!!

PR #4531, opened 2020-12-22
Custom panasonic remote buttons and Custom URLs in menu

rpeyron:
Hello,

This PR includes two modifications:

### Panasonic remote buttons

It adds customizable buttons to the default remote for Panasonic TVs:

domoticz_pana_remote_custombuttons

The buttons are customizable in the Panasonic hardware settings:

domoticz_pana_remote_settings

To be able to use codes that are not known of the plugin there is an option to allow sending unknown commands. There is basic sanitization but as it may cause a security issue there is also a setting to allow or not this behavior.

### URL in Custom menu

It adds the possibility to include URLs in the custom template menu:
domoticz_externalurl_menu

To do so, just create a file per URL in the www/templates folder with:
- 'url' extension 
- the title of the link as filename
- the URL as file contents

domoticz_externalurl_folder

Hope you will find this PR useful and thanks for domoticz!

Best regards

### DIFF
--- a/hardware/PanasonicTV.cpp
+++ b/hardware/PanasonicTV.cpp
@@ -142,6 +142,7 @@ private:
   std::string buildXMLStringRendCtl(const std::string &, const std::string &);
   std::string buildXMLStringRendCtl(const std::string &, const std::string &, const std::string &);
   std::string buildXMLStringNetCtl(const std::string &);
+  std::string buildXMLStringAppCtl(const std::string &);
 
   int m_HwdID;
   char m_szDevID[40];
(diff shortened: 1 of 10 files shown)


gizmocuz:
@rpeyron , thank you for the PR.
I think you included a bit to much in the PR (like the extern folder)
This PR also consists of 2 patches, maybe you can split them in two PR's ?
Not sure about the custom icons... Domoticz is not intended to replace a Panasonic remote control, it is easier to pick up the remote and operate.
But did the remote not had icons before ? How did it originally looked ? Maybe changing the original icons will do if there are not OK ?

rpeyron (commented):
@gizmocuz thanks for your feedback.
Sorry for the bit too much in the PR, I am not experienced with the PR function, here is a change to clean it and split in two. I keep this one for the remote buttons and will create a new one for custom URLs. Hope it is OK.

About the custom buttons, there is some use cases where in my opinion it is very useful to have the buttons you use in domoticz in addition to the physical remote:
- you can add it in scene and automate for instance the input or app on your TV (in a 'back to home' scene for instance), and add also voice automation with voice assistants
- the original remote does not have the corresponding buttons and involves a lot of buttons pressed (that is my case for direct HDMI selection, and app running that I use most of the time)
- these buttons may not work any more on the original physical remote and not included in cheap replacement (if you don't want to spend the ridiculously high price for an original remote)

I fully agree that maybe not everybody need this, add those needing this will also have different needs, that is why I made this fully customizable in settings. Having the buttons included in the original domoticz remote will in my opinion overload unnecessarily users that are happy with the current remote (but it is definitely easier to do ;-) ; it is the case in fact in the first commits) 

To answer you question about the original look of the remote, below are screenshots of the full remote:

Original remote:
Capture d’écran 2020-12-23 113938

Example of heavily customized remote:
Capture d’écran 2020-12-23 114134

And example of a lighter customization:
Capture d’écran 2020-12-23 114218

And without customization (default value), you have exactly the original remote above.

Hope that you will have a more clear idea about these modifications and that will be OK for you.
Thanks

rpeyron:
Hum sorry this obviously did not change anything in the PR... will try something else.

rpeyron:
Replaced by new PRs  #4535  and #4536

PR #4535, opened 2020-12-23
Custom URLs in menu

rpeyron:
Hello,

This is the custom URLs in menu part of my previous PR #4531 

It adds the possibility to include URLs in the custom template menu:
domoticz_externalurl_menu

To do so, just create a file per URL in the www/templates folder with:
- 'url' extension 
- the title of the link as filename
- the URL as file contents

domoticz_externalurl_folder

Hope you will find this PR useful and thanks for domoticz!

### DIFF
--- a/www/app/app.js
+++ b/www/app/app.js
@@ -439,23 +439,25 @@ define(['angularAMD', 'app.routes', 'app.constants', 'app.notifications', 'app.p
 
 						$rootScope.MakeGlobalConfig();
 
-						if ( (typeof data.result.templates != 'undefined') || (typeof data.result.urls != 'undefined') ) {
-							var customHTML = "";
+						var customHTML = "";
+						if (typeof data.result.templates != 'undefined')  {
 							$.each(data.result.templates, function (i, item) {
 								var cFile = item;
 								var cName = cFile.charAt(0).toUpperCase() + cFile.slice(1);
 								var cURL = "templates/" + cFile;
 								customHTML += '<li><a href="javascript:SwitchLayout(\'' + cURL + '\')">' + cName + '</a></li>';
 							});
+						}
+						if (typeof data.result.urls != 'undefined')  {
 							$.each(data.result.urls, function (name, url) {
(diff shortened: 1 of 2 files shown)


www/app/app.js
```diff
@@ -439,14 +439,19 @@ define(['angularAMD', 'app.routes', 'app.constants', 'app.notifications', 'app.p
 
 						$rootScope.MakeGlobalConfig();
 
-						if (typeof data.result.templates != 'undefined') {
+						if ( (typeof data.result.templates != 'undefined') || (typeof data.result.urls != 'undefined') ) {
```
gizmocuz: Maybe it's best to make a second IF block only for the urls (when undefined) ?

```
if (typeof data.result.urls != 'undefined')  {
...
}

```

rpeyron: @gizmocuz   I am not sure to understand what you mean: this would require to duplicate the code to properly set the $rootScope.config.EnableTabCustom?

gizmocuz: I mean two seperate $.each loops depending if the result templates  or urls is undefined

```
if (typeof data.result.templates != 'undefined') {
	$.each(data.result.templates, function (i, item) {
		var cFile = item;
		var cName = cFile.charAt(0).toUpperCase() + cFile.slice(1);
		var cURL = "templates/" + cFile;
		customHTML += '<li><a href="javascript:SwitchLayout(\'' + cURL + '\')">' + cName + '</a></li>';
	});
}
```

```
if (typeof data.result.urls != 'undefined')  {
	$.each(data.result.urls, function (name, url) {
		var cName = name.charAt(0).toUpperCase() + name.slice(1);
		var cURL = url;
		customHTML += '<li><a target="_blank" href="' + cURL + '">' + cName + '</a></li>';
	});
}
```

rpeyron: @gizmocuz ok sorry I missed that, even surprising it didn't crash...  changes done.

gizmocuz:
Thanks!!

PR #4536, opened 2020-12-23
Custom panasonic remote buttons

rpeyron:
Hello again,

This is the custom panasonic buttons part of my previous PR #4531 (while struggling with splitting the two it seems it closed the previous one, sorry about that).

It adds customizable buttons to the default remote for Panasonic TVs:

domoticz_pana_remote_custombuttons

The buttons are customizable in the Panasonic hardware settings:

domoticz_pana_remote_settings

To be able to use codes that are not known of the plugin there is an option to allow sending unknown commands. There is basic sanitization but as it may cause a security issue there is also a setting to allow or not this behavior.

**To answer the questions you made on PR #4531:** 

About the custom buttons, there is some use cases where in my opinion it is very useful to have the buttons you use in domoticz in addition to the physical remote:
- you can add it in scene and automate for instance the input or app on your TV (in a 'back to home' scene for instance), and add also voice automation with voice assistants
- the original remote does not have the corresponding buttons and involves a lot of buttons pressed (that is my case for direct HDMI selection, and app running that I use most of the time)
- these buttons may not work any more on the original physical remote and not included in cheap replacement (if you don't want to spend the ridiculously high price for an original remote)

I fully agree that maybe not everybody need this and those needing this will also have different needs. That is why I made this fully customizable in settings and will default to original look. Having the buttons included in the original domoticz remote will in my opinion overload unnecessarily users that are happy with the current remote (but it is definitely easier to do ;-) ; it is the case in fact in the first commits) 

To answer your questions about the original look of the remote, below are screenshots of the full remote:

Original remote:
Capture d’écran 2020-12-23 113938

Example of heavily customized remote:
Capture d’écran 2020-12-23 114134

And example of a lighter customization:
Capture d’écran 2020-12-23 114218

And without customization (default value), you have exactly the original remote above.

Hope that you will have a more clear idea about these modifications and that will be OK for you.
Thanks

### DIFF
--- a/hardware/PanasonicTV.cpp
+++ b/hardware/PanasonicTV.cpp
@@ -96,9 +96,9 @@ class CPanasonicNode : public StoppableTask
 		void			Clear();
 		std::string		LogMessage();
 		std::string		StatusMessage();
-		bool			LogRequired(CPanasonicStatus&);
-		bool			UpdateRequired(CPanasonicStatus&);
-		bool			OnOffRequired(CPanasonicStatus&);
+		bool			LogRequired(CPanasonicStatus& pPrevious);
+		bool			UpdateRequired(CPanasonicStatus& pPrevious);
+		bool			OnOffRequired(CPanasonicStatus& pPrevious);
 		bool			IsOn() { return (m_nStatus != MSTAT_OFF); };
 		void			Volume(int pVolume) { m_VolumeLevel = pVolume; };
 		void			Muted(bool pMuted) { m_Muted = pMuted; };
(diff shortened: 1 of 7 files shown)


gizmocuz:
Thank you for the PR and the explanation, the PR is perfect and can be merged without issues!
But... since you are making modifications, could you also make a medication that the original author should have done (and I have overlooked)
In the header file (.h), it would be best if you also have the function parameters as names , not like this:

before (current):
`  std::string buildXMLStringAppCtl(const std::string &);
`
If possible change them to:
`  std::string buildXMLStringAppCtl(const std::string &productId);
`
Otherwise when looking in the header, you have to go back/forwarded to the header and cpp file

(I think you know, but if you make changes you can just commit/push it github and it will be updated here...)

rpeyron:
@gizmocuz Hello, function parameters names have been added  for all functions (if I have not missed one :) )

gizmocuz:
@rpeyron , thanks for this !! And the PR!

PR #4542, opened 2020-12-27
Fix the sanity checking of Z-Wave power meter counter values. Fixes #4541.

### LINKED ISSUE #4541: Sanity checking of Z-Wave power meter counter values is faulty
I'm running a fresh build of the development branch on NetBSD/amd64-current:

Version: 2020.2 (build 12785)
Build Hash: c3a1521d0-modified
Compile Date: 2020-12-26 15:03:31
dzVents Version: 3.0.19
Python Version: None

I've discovered a problem with commit 7933f659d668a24d6e8eaf8a54998672f5ca63d9, added on December 6th, to do some sanity checking of Z-Wave power meter counter values before using them to update the database. (This sort of checking is welcome, by the way, as I've had to clean up bad entries from time to time.)

The problem is that some power meters will stop working after they count up a certain amount of power used, because there is code in the above commit that mistakenly compares a counter (the total accumulated consumption, before scaling) to a limit on the maximum acceptable power consumption per interval. Further, the ignoring of what is judged to be erroneous data happens without logging anything.  In my case, I have a NorthQ Q-Power NQ-92021 meter monitoring the main power consumption meter in my home, and reporting the accumulated consumption as a counter (i.e. in Wh, which are then scaled to kWh by Domoticz) every 15 minutes.

I have a working patch to correct the code in question, and will create a pull request.

tih:
As described in #4541, commit 7933f659d668a24d6e8eaf8a54998672f5ca63d9, while adding various sanity checks to catch erroneous counter values from power meters (caused by e.g. counter rollovers or meter resets), got one of these checks wrong, by comparing an unscaled counter representing the total accumulated power usage to a limit on the maximum acceptable rate of change of said counter. An attempt was made to allow a seeming discrepancy to be accepted if it persisted over time, but this would not always work, either.

This small change modifies the code in question to do what I believe to be the original author's actual intent. It also adds logging of the (temporary) suppression of what is judged to be possibly erroneous data.

I'm currently running with this modification locally, and have observed it to work correctly.

### DIFF
--- a/hardware/ZWaveBase.cpp
+++ b/hardware/ZWaveBase.cpp
@@ -477,16 +477,20 @@ void ZWaveBase::SendDevice2Domoticz(_tZWaveDevice* pDevice)
 		|| (pDevice->devType == ZDTYPE_SENSOR_KVARH)
 		)
 	{
-		// check for possible counter / usage issues
-		if ((pDevice->floatValue < -m_sql.m_max_kwh_usage) || (pDevice->floatValue > m_sql.m_max_kwh_usage))
-			return;
+		// check for possible counter issues
 		if (pDevice->lastSendValue != 0)
 		{
-			if (pDevice->floatValue < pDevice->prevFloatValue)
+			float floatChange = (pDevice->floatValue - pDevice->prevFloatValue) / pDevice->scaleMultiply;
+			if ((floatChange < 0) || (floatChange > m_sql.m_max_kwh_usage))
 			{
 				if (pDevice->lastreceived - pDevice->lastSendValue < 3600)
 				{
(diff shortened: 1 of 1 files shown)


gizmocuz:
@tih , thank you for the patch!
If I remember correctly, some meter counters decrease (if you have a meter that registers both usage as return power, same as an old analog meter)
For that reason I checked against -m_sql.m_max_kwh_usage
If your convenient enough that this situation does not exist I am happy to merge the PR, else please modify the PR to use -m_sql.m_max_kwh_usage

gizmocuz:
Thanks !!

PR #4544, opened 2020-12-28
MercedesMe: Fix to handle HTTP/2 (and other) response header correctly

kiddigital:
The Mercedes API's now gives both HTTP/1.1 and HTTP/2 resultheaders and this was not handled correctly. This caused some correct responses not being handled as correct because the Status-Code was not extracted correctly and therefor the results where presumed to be incorrect.

The HTTP responsecode is extracted from this Status-Line response header (RFC2616) but that was not done in the correct way. As the HTTP versions can differ and have a different length, the position of the HTTP Status-Code is not fixed but can vary. Between the HTTP-Version and the Status-Code a SPACE character is used as the separator.

Now the parsing of the Status-Line is done correctly using this separator resulting in the correct extraction of the Status-Code.


gizmocuz:
Thank you for this PR. Is it possible to keep the extension of the header file .h ? All our files are .h (except a tiny few)

kiddigital:
> 
> 
> Thank you for this PR. Is it possible to keep the extension of the header file .h ? All our files are .h (except a tiny few)

Sure, but don't we want to move over to full C++ anyway? This way tools, github, etc. recognize the code as C++ and not as plain C anymore.

gizmocuz:
@kiddigital , not because of a header I hope... 
It is (or I agree with that), that it is good practice to use .hpp for header only libraries and not for class headers

gizmocuz:
Thank you, but I think you have to revert it on more locations in your PR

kiddigital:
@gizmocuz , reversed the header back to .h and I can find myself in the reasoning you gave.

Looked at some/my IDE and it is fine with both as there seems no difference (I could find at least) in how the IDE supports the developer with these files.

Added a few directives for Github's linguist feature so it recognizes the .h headers as C++ and excludes some external directories.

gizmocuz:
Thank you again for your work!!

PR #4549, opened 2020-12-30
eVehicle: Handle wrong or empty battery readings

kiddigital:
When no- or bad readings of the batterystatus was received, the value become random due to not initialized variables. This is now fixed by a) initializing the variable and b) checking/correcting incoming values before storing


gizmocuz:
Thanks!

PR #4550, opened 2020-12-30
eVehicles: Create switches instead of textsensor for boolean custom values

kiddigital:
When retrieving custom values from a vehicle's API, most returned values where converted to either a text sensor (containing text) or a generic sensor (containing a number). But when the API returned a 'boolean' value, these were converted to text (in case of 'true'  or 'false') or a generic sensor (in case of 0 or 1 or real booleans) as well.

Now additional checking is done trying to convert a boolean value into a real boolean and putting it into a Switchtype sensor. This way proper Switches become available for boolean type variables like 'left frontdoor open/closed' or 'carlights on/off', etc.


gizmocuz:
Thanks!!

PR #4551, opened 2020-12-31
dzVents Shell Command: Removed some more logging

akamming:
This logging should be visibile when debugging only


kiddigital:
@akamming , why use compiletime debug (#ifdef) and not runtime debug (_log.Debug)? Makes it much easier to just (re)start domoticz with Debug flags enabled.

gizmocuz:
Yep, I agree with that, It's what it is designed for ;)

akamming:
ok... corrected...

gizmocuz:
Thank you!

PR #4552, opened 2020-12-31
Custom Icon selection for Utilities

gizmocuz:


megamarco833:
hi, it is possible to add also the possibility to change icon of blind percentage / blind percentage inverted?
in this type of dummy switch it is not possible to change icon.

gizmocuz:
@megamarco833 , this patch if for Utilities, the icons of the blinds are just fine

PR #4554, opened 2020-12-31
MercApi: Added Status message for non-standard HTTP results

kiddigital:
Sometimes the Mercedes API's do respond but with a non-standard HTTP result-code. Made this visible in Status logging so no need to restart Domoticz in debug mode to find out anymore.

### DIFF
--- a/hardware/eVehicles/MercApi.cpp
+++ b/hardware/eVehicles/MercApi.cpp
@@ -903,7 +903,7 @@ uint16_t CMercApi::ExtractHTTPResultCode(const std::string& sResponseHeaderLine0
 				iHttpCode = (uint16_t)std::stoi(sResponseHeaderLine0.substr(iHttpCodeStartPos, 3));
 				if (iHttpCode < 100 || iHttpCode > 599)		// Check valid resultcode range
 				{
-					iHttpCode = 9999;
+					_log.Log(LOG_STATUS, "Found non-standard resultcode (%d) in HTTP response statusline: %s", iHttpCode, sResponseHeaderLine0.c_str());
 				}
 			}
 		}


gizmocuz:
Thanks! I wonder what the code will be ;)

kiddigital:
`Status: Found non-standard resultcode (28) in HTTP response statusline: HTTP/1.1 28 Timeout was reached`

After just a few seconds after the request... probably something internal at the API service infrastructure that gets exposed outside incorrectly. I guess a 504 (Gateway Timeout) or 502 (Bad Gateway) would be more fitting. And a 408 (Request Timeout) doesn't make sense as there is no additional information from the client required.

And thanks for merging so quickly :)

gizmocuz:
Ahhh good you logged this, the error is actually a curl error:

Errors 0-99 match curl errors: 28 is “Operation timeout. The specified time-out period was reached according to the conditions.”

kiddigital:
Hmm.. so it might actually be coming from Domoticz internally (using the curl lib)? 
I also see Buienradar module telling me 'Problem Connecting...' a lot, might be similar? Maybe I should take a look there as well, see if it might be the same problem? I think both (and most modules doing API calls) are using the HTTPClient class?

kiddigital:
Just looked at the HTTPClient class and indeed it fooled me :)
```
		else
		{
			if (res != CURLE_HTTP_RETURNED_ERROR)
			{
				//Need to generate a header
				std::stringstream ss;
				ss << "HTTP/1.1 " << res << " " << curl_easy_strerror(res);
				vHeaderData.push_back(ss.str());
			}
		}
```
a standard HTTP Statusline is generated here containing the CURL error code.
Maybe concat the word '(DomoticzCURLlib)' in this header as part of the description (curl_easy_strerror(res)) so it is clear(er) that it comes from Domoticz and not from the serverside of the called party?

gizmocuz:
If you want to add a special header tag that is OK with me, It's not a 'Domoticz' Curl thing, it's just a Curl Error
In your case it's a timeout... so you might have a connection issue, if you google for error 28 you find some hinds

PR #4559, opened 2021-01-03
Support of partitions > 2Tb in FetchUnixDisk()

tofbouf:
Use of long long variables and %lld format in sscanf.
The former implementation (with long) was OK only up to 2Tb partitions.


gizmocuz:
Thank you for the PR and this patch!

Issue #4568, opened 2021-01-06
domoticz crashes every 5 minutes if a non numeric value is in the sValue of a multimeter type device


rwaaren:
See [Forum post](https://www.domoticz.com/forum/viewtopic.php?p=265462#p265462)

During tests why domoticz crashed on  a wrong value introduced via MQTT when these values are propagated to the short term history tables., a crash occurred  and was reproducible  on build 12825 when the wrong value was injected into a multimeter device with MQTT. 
A similar value in a meter type device did not cause a crash but resulted in an error message to the log

`2021-01-06 19:50:00.638  Error: UpdateMeter: Error converting sValue/sUsage! (IDX: 780, sValue: '-x', sUsage: '0', dType: 113, sType: 0)`

rwaaren:
I will try to create a PR for this

gizmocuz:
I am not able to quickly replicate this. It is for a RFXMeter type as the devType = 113
in **CSQLHelper::UpdateMultiMeter**, it is checked if the sValue has 6 fields (separated by a semicolum)
Sure it is still a good idea to surround the '**std::stoull**' with a try/catch here as well
Or it's a counter incremental ?
I made a small patch for the pTypeP1Power

rwaaren:
It's a P1 meter. 
I am close to have a PR ready that wil protect the updates to multimeter table.

gizmocuz:
I think it is already covered by my previous commit.
But in the Forum post it displays devType 113? (Also in your initial post), a P1 is not 113 ?
Edit: P1 = 250

rwaaren:
@gizmocuz; yes, I tested with P1 (devicetype 250) and you fixed that one already.
.
Created #4569 to also protect other (multimeter) devicetypes

rwaaren:
fixed by @gizmocuz ( commit https://github.com/domoticz/domoticz/commit/7a7c9645cf814b06e127a03848d07a1e85722919 )

PR #4569, opened 2021-01-07
Protect multimeter sValue values conversions (fix for #4568)

rwaaren:


gizmocuz:
Thank you for the PR.
I like my commit better as it is consistent with the way the 'meter' catches the exception (it will use 0 for the values)
functions like atof/atoi/atol do not throw exceptions but returns 0 in case of invalid data

rwaaren:
The reason why I did not use 0 was that the code then converts an invalid value into a valid value ( 0 ) Probably a matter of opinion but in this case I would prefer to have no value at all. 
Anyway; most important is that a potential crash is prevented and I learned something new today.

Maybe I overlook something but in your commit the other devicetypes ( pTypeCURRENT and pTypeCURRENTENERGY)
are not protected. Is that with a specific reason?

gizmocuz:
The default behavior is to return 0, so this is simulating the default atol/atoi/atof commands now (same as the meter)
 That is correct, as mentioned atoi/atof/atol do not throw execptions

rwaaren:
OK. Understand now. Thx for explaining !

PR #4576, opened 2021-01-11
Implemented multiline parameters in python plugins

ultrasuperpingu:
Python plugins only have a limited number of parameters and sometimes, ones need more. As a solution, I propose to add multiline parameters, so parameters can contains json, xml, or whatever parsable format and still being editable and readable in hardware configuration GUI.

### DIFF
--- a/www/app/hardware/Hardware.js
+++ b/www/app/hardware/Hardware.js
@@ -4793,23 +4793,29 @@ define(['app'], function (app) {
 											});
 											PluginParams += '</select></td>';
 										} else {
-											var multiline=(typeof (param.multiline) != "undefined") && (param.multiline == "true")
+											var nbRows=-1;
+											if (typeof (param.rows) != "undefined") {
+												var n = Math.floor(Number(param.rows));
+												if(n !== Infinity && String(n) === param.rows && n >= 0) {
+													nbRows = n;
+												}
+											}
 											PluginParams += '<td>'
-											if ((typeof (param.password) != "undefined") && (param.password == "true"))
-												PluginParams += '<input type="password" ';
-											else if(multiline)
-												PluginParams += '<textarea ';
-											else
-												PluginParams += '<input type="text" ';
-											PluginParams += 'id="' + param.field + '" style="width:' + param.width + '; padding: .2em;" class="text ui-widget-content ui-corner-all" '
-											if (typeof (param.default) != "undefined" && !multiline) PluginParams += 'value="' + param.default + '"';
-											if ((typeof (param.required) != "undefined") && (param.required == "true")) PluginParams += ' required';
-											if(multiline) {
-												PluginParams += ' >';
+											if (nbRows >= 0) {
+												PluginParams += '<textarea id="' + param.field + '" style="width:' + param.width + '; padding: .2em;" class="text ui-widget-content ui-corner-all" rows="' + nbRows + '" ';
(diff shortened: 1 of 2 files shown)


gizmocuz:
Thank you for the PR. I do not understand what this PR is for based on this title and looking at the changed files.
Is this for a 'Extra' parameter option ? Our internal hardware sensors also have a 'Extra' field, could you use this ?
If not, could you name it 'Extra Parameters' or Additional Parameters ? That sounds better then multiline

ultrasuperpingu:
In python plugin, you can define a `<param field="Mode1" label="Test" width="150px" multiline="true"/>`. It will be display with a textarea instead of a `<input type=text>`

ultrasuperpingu:
Sorry, I pushed it too fast, it's not working well. I'll try to fix it this evening.

ultrasuperpingu:
I think it's ok now. Tested with something like:
```
<param field="Mode1" label="Test" width="240px" required="false" default="my
multiline
field" multiline="true"/>
```

gizmocuz:
I think you need to refactor the changes, make a whole block where you render the multiline textbox and use a 'else' for the rest
Maybe change 'multiline' to 'rows' like rows="4". How many rows does the text container contain now ?

ultrasuperpingu:
Ok, it's done. Thanks for the suggestion

www/app/hardware/Hardware.js
```diff
@@ -4793,23 +4793,29 @@ define(['app'], function (app) {
 											});
 											PluginParams += '</select></td>';
 										} else {
-											var multiline=(typeof (param.multiline) != "undefined") && (param.multiline == "true")
+											var nbRows=-1;
+											if (typeof (param.rows) != "undefined") {
+												var n = Math.floor(Number(param.rows));
```
gizmocuz: How about just use

`nbRows = parseInt(param.rows)`

ultrasuperpingu: Just to be sure to ignore wrong values like 10px or -10

ultrasuperpingu: But it's maybe not the best way to do it.

gizmocuz: Then after the parseInt just check if it's smaller <1 or >10 and set a default number like 5 (or 1)

ultrasuperpingu: I'm pretty sure parseInt("10e3") for example returns 10 for example. But it's probably not a big deal, I can do it if you want.

gizmocuz: That's fine as I wont expect any plugin author is going to enter this value... or any arbitrary value, even if he/she does, they would know when testing their plugin of course ;)

ultrasuperpingu: Ok, it's done

gizmocuz:
Thank you for your work and the PR!!!

ultrasuperpingu:
Just a thought, I don't know how the documentation can be updated and when it should be...

rwaaren:
@ultrasuperpingu Most documentation is on the wiki and everyone with a user-id can update the information on the pages.
If you send me a PM with your preferred user-id (not the same as in use on the forum due to technical restrictions) and the Email address where you want to receive your temporary password, I will create a user on the domoticz wiki system for you.

(my forum name is waaren)

PR #4577, opened 2021-01-12
Expand mqtt command send notification to include more variables.

brjhaverkamp:
This allows for sending notifications to a single subsystem or device
(My usecase is sending a message over FCM to only my mobile)

### DIFF
--- a/hardware/MQTT.cpp
+++ b/hardware/MQTT.cpp
@@ -483,7 +483,8 @@ void MQTT::on_message(const struct mosquitto_message* message)
 		else if (szCommand == "sendnotification")
 		{
 			uint64_t idx=0;
-			std::string name=std::string(""), subsystems=NOTIFYALL, subject, body, extradata=std::string(""), sound;
+			std::string name, subject, body, extradata, sound;
+			std::string subsystems=NOTIFYALL;
 			bool bfromnotification=true;
 			int priority = 0;
 			if (!root["idx"].empty())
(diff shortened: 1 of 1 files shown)


hardware/MQTT.cpp
```diff
@@ -482,8 +482,22 @@ void MQTT::on_message(const struct mosquitto_message* message)
 		}
 		else if (szCommand == "sendnotification")
 		{
-			std::string subject, body, sound;
+			uint64_t idx=0;
+			std::string name=std::string(""), subsystems=NOTIFYALL, subject, body, extradata=std::string(""), sound;
```
gizmocuz: Could you place subsystem on a separate line as you initialize this
You do not have to use std::string(""), a string is initialized empty by default.

brjhaverkamp:
I did the update, but made bit of a mess on commiting and reverting  
(first time I worked with a github pull request, learning on the fly)
Let me know if this is now OK, or what I should do to clean it up.

Bert

gizmocuz:
Looking great!! Thanks for this !!!

rwaaren:
@brjhaverkamp

Below is what is in [the wiki MQTT to domoticz ](https://www.domoticz.com/wiki/MQTT#MQTT_to_Domoticz ) now. 

{"command": "sendnotification", "subject": "My Subject", "body": "My Body", "priority": 0, "sound": "mysound" }

Can you reply with what is should be now? I will update the Wiki accordingly.

Or even better update the wiki yourself (You can send me a PM on the forum to request a userid to do that)

brjhaverkamp:
Hi and thanks for reaching out. It was on my ToDo-list to find how I could update the wiki on this.
I'll PM you for the id.

rwaaren:
@brjhaverkamp You have a reply with a request on the forum.

PR #4578, opened 2021-01-14
Added option to Panasonic hardware to allow to send command even if off

rpeyron:
Very small PR, to add an option to Panasonic hardware to allow sending commands even if the domoticz status of the device is off.

This is useful when you want to send some commands just after powering up the TV  (by a script by example) and the status has not yet been updated (the notification system takes up to 60 seconds to update the state, but the TV is already able to receive commands)

This option is not enabled by default.

### DIFF
--- a/hardware/PanasonicTV.cpp
+++ b/hardware/PanasonicTV.cpp
@@ -617,10 +617,20 @@ void CPanasonicNode::SendCommand(const std::string &command)
 
 	if (m_CurrentStatus.Status() == MSTAT_OFF && !m_PowerOnSupported)
 	{
-		// no point trying to send a command if we know the device is off
-		// if we get a 400 response when TV is off then Power toggle can be sent
-		_log.Log(LOG_ERROR, "Panasonic Plugin: (%s) Device is Off, cannot send command.", m_Name.c_str());
-		return;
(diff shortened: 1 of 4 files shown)


gizmocuz:
What happens when the TV is actually off and you are trying to send a command, will this cause a delay as the HTTP query is blocking ? Or does it return quickly ? 
I haven't used this myself, but can you turn on the TV when it's off ? That would imply that the HTTP service is always listening
Maybe if so, we could make this the default and there is no need for a configuration parameter

rpeyron:
There is a lot of Panasonic TV with different behavior. I can only test with mine, so I cannot say if following answers are valid for other TVs. If the TV is actually off and the new option activated, the connection is refused quickly and there is no delay. The command will just be ignored, and I haven't seen any nasty behavior caused by that.

My Panasonic TV does not allow to be turned on by the network.  I know there is some Panasonic TV that allows that, and it seems in the domoticz code that the Panasonic hardware code handles that pretty well (the m_PowerOnSupported  feature is detected and used).  My TV does not support power on by network, but I used another device connected with CEC to turn the TV on (a FireTV that I can contol within domoticz scripts).  I guess my use case is not the most common use case... and that's why  I added this as a configuration parameter in order not to break something with other models of Panasonic TVs.

gizmocuz:
Thank you !!

PR #4580, opened 2021-01-15
Fix GetSupportedThermostatModes 

rrozema:
Fix for zwave thermostat mode devices showing an empty options list.
(@gizmocuz: this was an actual intended pull request, hope I did it right this time :-) )

### DIFF
--- a/hardware/MQTT.cpp
+++ b/hardware/MQTT.cpp
@@ -169,7 +169,7 @@ void MQTT::on_message(const struct mosquitto_message* message)
 	std::string topic = message->topic;
 	std::string qMessage = std::string((char*)message->payload, (char*)message->payload + message->payloadlen);
 
-	_log.Log(LOG_NORM, "MQTT: Topic: %s, Message: %s", topic.c_str(), qMessage.c_str());
+	//_log.Log(LOG_NORM, "MQTT: Topic: %s, Message: %s", topic.c_str(), qMessage.c_str());
 
 	if (qMessage.empty())
 		return;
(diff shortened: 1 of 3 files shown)


rwaaren:
Is there a relation between the MQTT message and the zwave thermostat fix?

rrozema:
> Is there a relation between the MQTT message and the zwave thermostat fix?

No, that's my inexperience with using git. Still, it's a very annoying message, so if we can get rid of it, please?

gizmocuz:
@rrozema , Thank you for the PR.
There is to much in this PR that is unrelated to the problem..
A lot of changes in the 'extern' folder that should not be in this PR
A log change in the MQTT class, the log was fine
A lot of logging in the OZW class now that should not be nececery
A check that is not needed (for example if (tMode >= 0 && tMode <) that is not needed

So if I look at the actual fix it's this:

**Before:**
`const _tZWaveDevice* pDevice = FindDevice((uint8_t)nodeID, instanceID, indexID, ZDTYPE_SENSOR_THERMOSTAT_MODE);`
**After:**
`	const _tZWaveDevice* pDevice = FindDevice((uint8_t)nodeID, instanceID, COMMAND_CLASS_THERMOSTAT_MODE, ZDTYPE_SENSOR_THERMOSTAT_MODE);
`

and

**Before:**
`const _tZWaveDevice* pDevice = FindDevice((uint8_t)nodeID, instanceID, indexID, ZDTYPE_SENSOR_THERMOSTAT_FAN_MODE);`
**After:**
`const _tZWaveDevice* pDevice = FindDevice((uint8_t)nodeID, instanceID, COMMAND_CLASS_THERMOSTAT_FAN_MODE, ZDTYPE_SENSOR_THERMOSTAT_FAN_MODE);`

I know you are having a hard time with Git/Github, so I have made the changes for you (And of course referenced you)
Thank you very much for finding/fixing this!!

[Commit](https://github.com/domoticz/domoticz/commit/24b1f259f1d3c19d3f49ae7262c46d4c8d76a9dc)

Should be solved in upcoming beta version (12843)

@rwaaren , this could solve it as well

PR #4581, opened 2021-01-15
Control log level output per hardware device

gizmocuz:
This patch can enable/disable log out per hardware based on Info/Status or Error log levels
Default all is enabled

### DIFF
--- a/hardware/DomoticzHardware.h
+++ b/hardware/DomoticzHardware.h
@@ -70,7 +70,7 @@ class CDomoticzHardwareBase : public StoppableTask
 		__attribute__((format(printf, 3, 4)))
 #endif
 		;
-
+	uint32_t m_LogLevelEnabled; //bitwise _eLogLevel 7 = LOG_NORM | LOG_STATUS | LOG_ERROR
       protected:
 	virtual bool StartHardware() = 0;
 	virtual bool StopHardware() = 0;
--- a/main/SQLHelper.cpp
+++ b/main/SQLHelper.cpp
@@ -38,7 +38,7 @@
 #define __STDC_FORMAT_MACROS
 #include <inttypes.h>
 
-#define DB_VERSION 147
+#define DB_VERSION 148
 
 extern http::server::CWebServerHelper m_webservers;
 extern std::string szWWWFolder;
@@ -233,6 +233,7 @@ const char* sqlCreateHardware =
 "[Name] VARCHAR(200) NOT NULL, "
 "[Enabled] INTEGER DEFAULT 1, "
 "[Type] INTEGER NOT NULL, "
+"[LogLevel] INTEGER default 7, " //LOG_NORM + LOG_STATUS + LOG_ERROR
 "[Address] VARCHAR(200), "
 "[Port] INTEGER, "
 "[SerialPort] TEXT DEFAULT (''), "
@@ -2769,6 +2770,10 @@ bool CSQLHelper::OpenDatabase()
 		{
 			//Pushlink is not zero based anymore
 			safe_exec_no_return("UPDATE PushLink SET DelimitedValue=1 WHERE (DelimitedValue == 0)");
+		}
+		if (dbversion < 148)
+		{
+			query("ALTER TABLE Hardware ADD COLUMN [LogLevel] INTEGER DEFAULT 7"); // LOG_NORM + LOG_STATUS + LOG_ERROR
 		}
 	}
 	else if (bNewInstall)
--- a/main/WebServer.cpp
+++ b/main/WebServer.cpp
@@ -1045,6 +1045,7 @@ namespace http {
 			std::string name = HTMLSanitizer::Sanitize(CURLEncode::URLDecode(request::findValue(&req, "name")));
 			std::string senabled = request::findValue(&req, "enabled");
 			std::string shtype = request::findValue(&req, "htype");
+			std::string loglevel = request::findValue(&req, "loglevel");
 			std::string address = HTMLSanitizer::Sanitize(request::findValue(&req, "address"));
 			std::string sport = request::findValue(&req, "port");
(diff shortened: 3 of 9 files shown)


hardware/DomoticzHardware.h
```diff
@@ -70,7 +70,7 @@ class CDomoticzHardwareBase : public StoppableTask
 		__attribute__((format(printf, 3, 4)))
 #endif
 		;
-
+	uint32_t m_LogLevelEnabled; //bitwise _eLogLevel 7 = LOG_NORM | LOG_STATUS | LOG_ERROR
```
kiddigital, line 73: As this is a unsigned integer used as a bit-wise logical structure, shouldn't we make it a little smaller? We only use 3 bits now and reserving 32 is quite some room to grow :) We are not using these bits for hardware- or devicetypes (of which there are many). Maybe uint16_t is more than enough?

main/SQLHelper.cpp
```diff
@@ -233,6 +233,7 @@ const char* sqlCreateHardware =
 "[Name] VARCHAR(200) NOT NULL, "
 "[Enabled] INTEGER DEFAULT 1, "
 "[Type] INTEGER NOT NULL, "
+"[LogLevel] INTEGER default 7, " //LOG_NORM + LOG_STATUS + LOG_ERROR
```
kiddigital, line 236: My suggestion would be to turn all bits on by default. Including the future once! So if we use LogLevel as a 16 bit (or 32, see previous comment on size) bit-wise structure, should we set the default value to 65535 (xFFFF)? This way all bits are **On**

main/SQLHelper.cpp
```diff
@@ -2770,6 +2771,10 @@ bool CSQLHelper::OpenDatabase()
 			//Pushlink is not zero based anymore
 			safe_exec_no_return("UPDATE PushLink SET DelimitedValue=1 WHERE (DelimitedValue == 0)");
 		}
+		if (dbversion < 148)
+		{
+			query("ALTER TABLE Hardware ADD COLUMN [LogLevel] INTEGER DEFAULT 7"); // LOG_NORM + LOG_STATUS + LOG_ERROR
```
kiddigital, line 2776: See comment on default

main/WebServer.cpp
```diff
@@ -1373,10 +1375,11 @@ namespace http {
 
 			if (htype == HTYPE_HTTPPOLLER) {
 				m_sql.safe_query(
-					"INSERT INTO Hardware (Name, Enabled, Type, Address, Port, SerialPort, Username, Password, Extra, Mode1, Mode2, Mode3, Mode4, Mode5, Mode6, DataTimeout) VALUES ('%q',%d, %d,'%q',%d,'%q','%q','%q','%q','%q','%q', '%q', '%q', '%q', '%q', %d)",
+					"INSERT INTO Hardware (Name, Enabled, Type, LogLevel, Address, Port, SerialPort, Username, Password, Extra, Mode1, Mode2, Mode3, Mode4, Mode5, Mode6, DataTimeout) VALUES ('%q',%d, %d, %d,'%q',%d,'%q','%q','%q','%q','%q','%q', '%q', '%q', '%q', '%q', %d)",
```
kiddigital, line 1378: Maybe move both the insert and update query to a (SQL)helper function instead of having the same query multiple times in the code? Less code and just 1 implementation of the query :)

main/WebServer.cpp
```diff
@@ -1818,10 +1830,11 @@ namespace http {
 				}
 				else {
 					m_sql.safe_query(
-						"UPDATE Hardware SET Name='%q', Enabled=%d, Type=%d, Address='%q', Port=%d, SerialPort='%q', Username='%q', Password='%q', Extra='%q', Mode1=%d, Mode2=%d, Mode3=%d, Mode4=%d, Mode5=%d, Mode6=%d, DataTimeout=%d WHERE (ID == '%q')",
+						"UPDATE Hardware SET Name='%q', Enabled=%d, Type=%d, LogLevel=%d, Address='%q', Port=%d, SerialPort='%q', Username='%q', Password='%q', Extra='%q', Mode1=%d, Mode2=%d, Mode3=%d, Mode4=%d, Mode5=%d, Mode6=%d, DataTimeout=%d WHERE (ID == '%q')",
```
kiddigital, line 1833: See above suggestion to move these to a single function instead of copied code

hardware/DomoticzHardware.h
gizmocuz: Internally we already used a uint32 so I continued with it.. 😁
But there a possible issue with this PR that I need to check..  some error in hardware.js

kiddigital (commented):
Nice feature, gives much more fine grained control over the output as some hardware modules are more chatty than others (although useful, sometimes just a lot when things are stable).

main/SQLHelper.cpp
gizmocuz: I also want to do more work on the log window

rwaaren:
Last commit before merge did change the behavior. You now also have to untick error to prevent info and status messages

gizmocuz:
I think this is corrected by now

Mark-Sellwood:
Should this work for Python Plugins?

gizmocuz:
@Mark-Sellwood , I suspect so, did you try ?

Mark-Sellwood:
Yes, an example is the Zigbee2MQTT plugin, if I select just Error then apply, when I come back all 3 are enabled.
I also tried removing the hardware & adding it back with the same result.
To me it looks as if it only works with native hardware.

Mark-Sellwood:
Looks as if you have fixed it in build 12892 | cb924ffe3
Thanks

gizmocuz:
@Mark-Sellwood , Yes sorry, forgot to mention this ! Thanks for the feedback !

PR #4585, opened 2021-01-18
dzVents openURL optimisation

akamming:
dzVents openURL processes http requests sequencially in the SQLHelper worker thread, blocking that thread in case of long running http requests or timeouts (which can take up to 120 sec). During that periode the SQL helper thread is blocked.

This PR fixes that by executing http requests on seperate threads

@rwaaren  ofcourse i already tested myself, but i cannot get the  the dzvents regression testscripts to run. Could you run those on this PR?

### DIFF
--- a/dzVents/runtime/integration-tests/stage1.lua
+++ b/dzVents/runtime/integration-tests/stage1.lua
@@ -691,7 +691,7 @@ local testRGBW = function(name)
 		["timedOut"] = false;
 	})
 	dev.setRGB(15, 30, 60)
-	dev.dimTo(15)
+	dev.dimTo(15).afterSec(5)
 	tstMsg('Test RGBW device', res)
 	return res
 end
--- a/main/SQLHelper.cpp
+++ b/main/SQLHelper.cpp
@@ -3583,7 +3583,6 @@ void CSQLHelper::PerformThreadedAction(const _tTaskItem itt)
 		std::string callback = itt._ID;
 		std::string url = itt._sValue;
 		int method = itt._switchtype;
-		_log.Log(LOG_ERROR, "opening url: %s", url.c_str());
 
 		if (!itt._relatedEvent.empty())
 			StringSplit(itt._relatedEvent, "!#", extraHeaders);
--- a/main/SQLHelper.cpp
+++ b/main/SQLHelper.cpp
@@ -3333,285 +3333,354 @@ void CSQLHelper::ManageExecuteScriptTimeout(int pid, int timeout, bool *stillRun
 }
 #endif
 
-void CSQLHelper::ExecuteScriptThreaded(std::string command, std::string callback, int timeout, std::string path)
+void CSQLHelper::PerformThreadedAction(const _tTaskItem itt)
 {
-	std::ifstream infile;
-	std::string sLine;
-	std::string filename;
-	std::string filenamestderr;
-	std::string scriptoutput;
-	std::string scriptstderr;
-	bool commmandExecutedSuccesfully = false;
-	int exitcode = 0;
+	if (itt._ItemType == TITEM_EXECUTESHELLCOMMAND)
+	{
+		std::string command = itt._sValue;
+		std::string callback = itt._ID;
+		std::string path = itt._sUser;
+		int timeout = itt._nValue;
+
+		std::ifstream infile;
+		std::string sLine;
+		std::string filename;
+		std::string filenamestderr;
+		std::string scriptoutput;
+		std::string scriptstderr;
+		bool commmandExecutedSuccesfully = false;
+		int exitcode = 0;
 #ifndef WIN32
-	int pid;
-	bool stillRunning = true;
-	std::shared_ptr<std::thread> T;
+		int pid;
+		bool stillRunning = true;
+		std::shared_ptr<std::thread> T;
 #endif
-	bool timeoutOccurred = false;
+		bool timeoutOccurred = false;
 
-	// make sure we have unique filenames
-	scriptoutputindex++;
-	if (scriptoutputindex > 10000) // should be a big number, to prevent parallel scripts having the same output files. 250 concurrent will probably never be reached
-	{
-		scriptoutputindex = 1;
-	}
-	std::string scriptoutputindextext = std::to_string(scriptoutputindex);
+		// make sure we have unique filenames
+		scriptoutputindex++;
+		if (scriptoutputindex > 10000) // should be a big number, to prevent parallel scripts having the same output files. 250 concurrent will probably never be reached
+		{
+			scriptoutputindex = 1;
+		}
+		std::string scriptoutputindextext = std::to_string(scriptoutputindex);
 
-	// create filenames for stderr and stdout  ("path+domscript+index+<.out|.err>")
-	filename = path;
-	filename.append("domscript");
-	filename.append(scriptoutputindextext);
-	filenamestderr = filename;
-	filename.append(".out");       // stdout
-	filenamestderr.append(".err"); // stderr
+		// create filenames for stderr and stdout  ("path+domscript+index+<.out|.err>")
+		filename = path;
+		filename.append("domscript");
+		filename.append(scriptoutputindextext);
+		filenamestderr = filename;
+		filename.append(".out");       // stdout
+		filenamestderr.append(".err"); // stderr
 
-	// Remove temp file if they exist
-	if (!remove(filename.c_str()))
-		_log.Log(LOG_ERROR, "old temp file (%s) was still there, removing...", filename.c_str());
-	if (!remove(filenamestderr.c_str()))
-		_log.Log(LOG_ERROR, "old temp file (%s) was still there, removing...", filenamestderr.c_str());
+		// Remove temp file if they exist
+		if (!remove(filename.c_str()))
+			_log.Log(LOG_ERROR, "old temp file (%s) was still there, removing...", filename.c_str());
+		if (!remove(filenamestderr.c_str()))
+			_log.Log(LOG_ERROR, "old temp file (%s) was still there, removing...", filenamestderr.c_str());
 
-	if (timeout > 0)
-	{
-		_log.Debug(DEBUG_NORM, "dzVents: Executing shellcommmand %s for max %d seconds", command.c_str(), timeout);
-	}
-	else
-	{
-		_log.Debug(DEBUG_NORM, "dzVents: Executing shellcommmand %s", command.c_str());
-	}
-
-#ifdef WIN32
-	SECURITY_ATTRIBUTES sa;
-	sa.nLength = sizeof(sa);
-	sa.lpSecurityDescriptor = NULL;
-	sa.bInheritHandle = TRUE;
-
-	HANDLE hstdout = CreateFile(_T(filename.c_str()), FILE_APPEND_DATA, FILE_SHARE_WRITE | FILE_SHARE_READ, &sa, OPEN_ALWAYS, FILE_ATTRIBUTE_NORMAL, NULL);
-	HANDLE hstderr = CreateFile(_T(filenamestderr.c_str()), FILE_APPEND_DATA, FILE_SHARE_WRITE | FILE_SHARE_READ, &sa, OPEN_ALWAYS, FILE_ATTRIBUTE_NORMAL, NULL);
-
-	PROCESS_INFORMATION pi;
-	STARTUPINFO si;
-	BOOL ret = FALSE;
-	DWORD flags = CREATE_NO_WINDOW;
-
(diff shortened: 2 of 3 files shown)


rwaaren:
Will do; compiling now. Just for my understanding are dzVents openURL calls handled different then ones coming from other event sub-systems?

rwaaren:
OK for dzVents.  (after fixing one test in integration tests)
Only thing noticed is that the setRGB command now took precedence over the dimTo. Before this change the calculated brightness value  of 24 was "overruled" by a dimTo(15) send directly after the setRGB() Now that only works when the dimTo() is delayed. 
Could also be that this means the colors are now directly available and that we no longer have to force them to become available as we used to do to overcome this. 
So for me All good







```
======== domoticz V2020.2 (12852), dzVents V3.1.1 +========================== Tests ==============================+
  time  | test-script                              | expected | tests | result  | successful |  failed  | seconds  |
===================================================+===============================================================+
00:00:02  Device                                        1        118      OK            118         0      0.3157
00:00:02  Domoticz                                      1         82      OK             82         0      0.3284
00:00:02  EventHelpers                                  1         33      OK             33         0      0.2858
00:00:03  EventHelpersStorage                           1         50      OK             50         0      0.2369
00:00:03  HTTPResponse                                  1          6      OK              6         0      0.0156
00:00:03  Lodash                                        1        100      OK            100         0      0.1722
00:00:03  ScriptdzVentsDispatching                      1          2      OK              2         0      0.0700
00:00:04  TimedCommand                                  1         46      OK             46         0      0.1064
00:00:04  Time                                          3        369      OK            369         0      0.8071
00:00:05  Utils                                         1         36      OK             36         0      0.0708
00:00:05  Variable                                      1         15      OK             15         0      0.0344
00:00:05  ContactDoorLockInvertedSwitch                 3          2      OK              2         0      3.0661
00:00:08  DelayedVariableScene                          8          2      OK              2         0      7.1800
00:00:15  EventState                                   19          2      OK              2         0     18.2048
00:00:34  Integration                                  34        222      OK            222         0     29.9441
00:01:04  SelectorSwitch                               10          2      OK              2         0      9.6135
00:01:23  SystemAndCustomEvents                         1          7      OK              7         0      0.0941

Total tests: 1094
   Mon Jan 18 13:15:57 CET 2021; dzVents version 3.1.1 tested without errors after 00:01:32
```

akamming:
Tx a lot!

About your questions:

> Only thing noticed is that the setRGB command now took precedence over the dimTo. Before this change the calculated brightness value of 24 was "overruled" by a dimTo(15) send directly after the setRGB() Now that only works when the dimTo() is delayed.

setRGB is using openURL "under water" to set RGB levels.  Both calls use the SQLHelper thread. So in the old code dimTo would be called after setRGB() is finished. In the new code dimTo called rightaway after the HTTP request thread is created, and they are called in parallel which is why you have to change the testscenario. 

> Could also be that this means the colors are now directly available and that we no longer have to force them to become available as we used to do to overcome this.

i don't exactly understand what you are saying here, but i would not expect any functional changes (apart from some a timing thing like the one above). This change only ensures that asynchronous requests really become asynchronous by not blocking anything during execution. If you are dependent of the anwser you should make sure you should wait for the callback.

> Just for my understanding are dzVents openURL calls handled different then ones coming from other event sub-systems?


That's a good question. i assumed this was the case, but i Just checked and it is not. The same open URL is also called by:
- Kodi, PanasonicTV and LogitechMediaServer  (which in the code call  "HandleOnOffAction", which itself calls this same mechanism). I don't expect any different behaviour since this is one-way traffic. These calls do not give a callback argument waiting for an answer
- Any script which has an openURL command in the command array at the end.

dzVents is the ony subsystem which uses the "callback", So i don't  expect any different functional behaviour other than the timing being a little bit different, which should not effect these components (thery are not awaiting the anwser)

I could add code to make this only effect dzVents, but i think it is better to leave it like this. Keeps the code simpler and if for whatever reason a openURL call takes a long time (e.g. timeout) it freezes the whole event system in the old code, and apparently without knowing, i accidentally fixed that as well ;-)

rwaaren:
@gizmocuz could you please have a look at this and merge if OK for you. I tested it from a dzVents perspective and that was successful.
Thx!

gizmocuz:
There no issues with this PR, it should work well in this case, but maybe @akamming can extend/tweak this a little bit

CSQLHelper::GetUrlThreaded ---> CSQLHelper::GetURLThreaded (Capital URL)

Do this as well for :

TITEM_EXECUTE_SCRIPT
TITEM_EMAIL_CAMERA_SNAPSHOT
TITEM_SEND_EMAIL
TITEM_SEND_EMAIL_TO
TITEM_SEND_SMS

As 'TITEM_EXECUTESHELLCOMMAND' already does this, maybe instead of 100 separated functions,
create one new function:

PerformThreadedAction(const _tTaskItem item)

and call/use this like

			else if (itt->_ItemType == TITEM_EXECUTE_SCRIPT)
			{
                              PerformThreadedAction(itt);
			}

And inside PerformThreadedAction the if/then/else and code logic from the previous functions

Hope I explained myself clear here ;) ...

akamming:
> Hope I explained myself clear here ;) ...

I think i understand, but i have some questions
- what is the reaons of doing the other ones as well: Do you expect they can take a long duration as well and therefor block the helper thread?
- the EXECUTESCRIPT still uses pipes in the underlying logic, this will still block stuff if performed simultaneously.... so change that logic as well?
- i think then the function call should then be something like below (right?) .. otherwise you would still have a lot of function calls and then i don't understand what you are referring to
```
else if (itt->_ItemType == TITEM_EXECUTE_SCRIPT)
		{
                          std::thread ActionThread(&CSQLHelper::PerformThreadedAction,this,itt);
                          ActionThread.detach();
		}

```

kiddigital:
Interesting approach and I like the improvement idea of making things asynchronous to keep Domoticz and all things it interact with as snappy as possible.

But I wonder if the async HTTP(S) calls are handled correctly by most modules as I think that currently most modules expect a synchronous/blocking call? So far I haven't seen any callback functions (maybe I just did not find it).

Anyway, would it be a suggestion to make a synchronous 'async' method? Meaning, it is synchronous/blocking from the callers perspective (so it can be a drop-in replacement for most current calls) but perform the actual call asynchronous allowing other threads to do stuff in the mean-time?

Or am I way off here and missing something?

(And also move the HTTP (a)synchandling, etc. out of the 'SQL' helper thread as it was not intended to do this? And as there are more and more HTTP based interaction, even in a local Domoticz netwerk, it would qualify for a thread of its own?)

akamming:
@kiddigital for the HTTP requests in the code which are handled by the SQL helper I already checked the code. 

i cannot find any occurences (apart from dzvents which uses the callback, so that's handled and OK)  where the code is awaiting the answer,

So in all those cases the caller is not awaiting the answer, so it doesn't matter when the answer arrives.   I tested with dzvents and the command array in python/lua en doing http calls using blockly. All work OK. 

The only other 3 occurences in the code (Kodi, PanasonicTV and LogitechMediaServer) i cannot test, but i checked the code: They perform on/off functions in domoticz using the json interface to domoticz.  if you want my 2 cents: If we were to optimize here, it would not be how to call the http function but to make that code way more efficient by directly calling the on/off event (why make httprequest to a json function, which leads to an event on the event queue, which when it's handled switches a switch on or off, if you can switch it on or off directly ;-)) 

so in my opinion a simulated synchronous call which is executed asynchronously in the background does not add anything. If it is needed in the future, the callback functionality can be used for this new code to await the answer..

akamming:
UPDATE: the requested change by gizmocus in this PR turned out not to be a lot of work (just moving code from one function to the other)
so i made the code changes for TITEM_EMAIL_CAMERA_SNAPSHOT, TITEM_SEND_EMAIL, TITEM_SEND_EMAIL_TO
 and TITEM_SEND_SMS as well using a generic function.

i did not do it for TITEM_EXECUTESCRIPT. Reaons: I know the underlying pipes can lead to deadlock (seen this while test the PR for EXECUTESHELLCOMMAND). So this function should be rewritten 1st.

is there some easy way (maybe an automated test?) to do a regression test, would otherwise be a lot of work...

kiddigital:
> 
> 
> @kiddigital for the HTTP requests in the code which are handled by the SQL helper I already checked the code.
> 
> i cannot find any occurences (apart from dzvents which uses the callback, so that's handled and OK) where the code is awaiting the answer,

Ok, clear. Maybe add a check to verify if the callback is specified and exists? That way other calls without a callback don't get 'stuck'?

> 
> So in all those cases the caller is not awaiting the answer, so it doesn't matter when the answer arrives. I tested with dzvents and the command array in python/lua en doing http calls using blockly. All work OK.

Sounds weird, that the caller is not awaiting an answer. At least some confirmation of succesfull or unsuccesful calling (HTTP resultcode for example?) should be expected. Otherwise it is a 'fire-and-forget-and-hope-it-works'. Seems weird if an action actually changes a state (from on to off for example) without knowing if it actually succeeded.

> 
> The only other 3 occurences in the code (Kodi, PanasonicTV and LogitechMediaServer) i cannot test, but i checked the code: They perform on/off functions in domoticz using the json interface to domoticz. if you want my 2 cents: If we were to optimize here, it would not be how to call the http function but to make that code way more efficient by directly calling the on/off event (why make httprequest to a json function, which leads to an event on the event queue, which when it's handled switches a switch on or off, if you can switch it on or off directly ;-))

Totally true... maybe we need to refactor them...
> 
> so in my opinion a simulated synchronous call which is executed asynchronously in the background does not add anything. If it is needed in the future, the callback functionality can be used for this new code to await the answer..

The gain I see is that currently a lot of modules (including some where I have worked on) use 'synchronous' calls so that worker thread is being blocked while waiting for a response (or time-out). If those synchronous calls could benefit from something async, those workers would become more responsive as well. Although when thinking about it some more, I could also imagine an issue if the worker gets called again in the mean-time... hm... needs more thinking I guess :)

akamming:
> Ok, clear. Maybe add a check to verify if the callback is specified and exists? That way other calls without a callback don't get 'stuck'?

already tested this with this PR: if no callback is specified, no event is generated for dzvents to act upon
if the callback has an invalid name, the answer never gets back to the calling dvents script but nothing  is 'stuck'. The script is just not called for the answer. This is no change since this PR. So nothing can get 'stuck'. All other calls to this function don't use callbacks (see below)

> Sounds weird, that the caller is not awaiting an answer. At least some confirmation of succesfull or unsuccesful calling (HTTP resultcode for example?) should be expected. Otherwise it is a 'fire-and-forget-and-hope-it-works'. Seems weird if an action actually changes a state (from on to off for example) without knowing if it actually succeeded.

Agree, but this is how it currently works, scripts in the event system build up a command array and then the command array commands are executed after the scripts are finished as a fire and forget. The only execption is the callback mechanism in dzvents, where scripts are re-executed with the callback as trigger.

> The gain I see is that currently a lot of modules (including some where I have worked on) use 'synchronous' calls so that worker thread is being blocked while waiting for a response (or time-out). If those synchronous calls could benefit from something async, those workers would become more responsive as well. Although when thinking about it some more, I could also imagine an issue if the worker gets called again in the mean-time... hm... needs more thinking I guess :)

Again agree, but also agree this needs some more rethinking. All code which is made async should be checked if it is threadsafe and should be thoroughly tested.

gizmocuz:
@akamming , the reason is that there are other actions that could block like scripts, email, sms, URL's
And to avoid having 100 new functions just for the threading part, so no function like:

PerformThreadedAction
GetURLThreaded 
SendSMSThreaded
SendEmailThreaded

But just one function that is called with the taskitem structure so it knows what to perform:

PerformThreadedAction

gizmocuz:
@akamming , Sorry this issue thread got flooded quite quickly with all kinds of replies. (Thanks kiddigital *g*)
I think what you currently made looks perfect! Is it possible to verify the workings ? (I expect them to work, but just in case)

You know that there is also another change in this PR ?

dzVents/runtime/integration-tests/stage1.lua

Is this correct?

akamming:
@gizmocuz 
>  Is it possible to verify the workings

- i verified sendmail, sendemailto, openurl and executeshellcommand.
- execute script i did not change (this needs to refactored 1st before we can make it to be executed in parallel. Now not threadsafe due to underlying pipes)
- sendsms en email camersnapshot i cannot verify (don't have a camera and no account a supported sms service)

> You know that there is also another change in this PR ?
>
>dzVents/runtime/integration-tests/stage1.lua

yes, was made by @rwaaren (needed to change a testcase because http requests are now send in parallel)

gizmocuz:
Thank you for the changes. I am currently debugging some issues with Github Actions before I can merge this

gizmocuz:
Thank you for this PR!

PR #4592, opened 2021-01-24
MercedesMe: Improve handling of API responses (partial 403's)

kiddigital:
The Mercedes API's seem sometimes to give a 403 - Access denied error even when Authentication and Authorization for these API's have been succesful.

This is unexpected and previously stopped further processing of data. Although unexpected, other API's can be available and so we like to process them anyway. So now when an API returns a 403, this API is skipped but processing of others is continued.

These unexpected 403 - Access Denied errors are now reported in the Status logging.


gizmocuz:
Thank you !!

PR #4594, opened 2021-01-25
explicitly set C++ version to 14

neheb:
This was already the case with MSVC which defaults to C++14. In the
case of clang, this was causing build failures as C++14 features are
used in the codebase. In the case of gcc, the libc++ used there is
non-standards conformant and exposes C++14 features to C++11.

This fixes compilation under FreeBSD which uses clang and libcxx as the
codebase introduced C++14 stuff recently.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -45,11 +45,9 @@ option(GIT_SUBMODULE "Check submodules during build" ON)
 ### COMPILER SETTINGS
 #
 #
-IF(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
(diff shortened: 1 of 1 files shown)


neheb:
ping @xbeaudouin

gizmocuz:
Thank you for this !!!

gizmocuz:
@xbeaudouin , sorry I was a bit to enthusiastic, hope it's working out for you as well ?

xbeaudouin:
@gizmocuz Sorry for the late reply : yeah this is working well ! Thanks both of you :)

PR #4595, opened 2021-01-25
treewide: remove most uses of static_cast

neheb:
static_cast is verbose for no reason. Just replace with normal cast.

### DIFF
--- a/hardware/1Wire.cpp
+++ b/hardware/1Wire.cpp
@@ -440,7 +440,7 @@ void C1Wire::ReportWiper(const std::string& deviceId, const int wiper)
 	DeviceIdToByteArray(deviceId, deviceIdByteArray);
 
 	int NodeID = (deviceIdByteArray[0] << 24) | (deviceIdByteArray[1] << 16) | (deviceIdByteArray[2] << 8) | (deviceIdByteArray[3]);
-	unsigned int value = static_cast<int>(wiper * (100.0 / 255.0));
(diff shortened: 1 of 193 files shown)


gizmocuz:
Thank you for this PR. Is it possible to solve the merge conflicts ? The code in OTGW has been updated recently

neheb:
It's rebased.

gizmocuz:
@neheb , the reason why I changed a lot of casts to use c++ style casting is because my compiler (VStudio) complained about the use of old style casting...

neheb:
Note that (char) is a C cast. I changed it to char(), which should be fine with C++.

gizmocuz:
Most warnings I received where about integers and floats

neheb:
Hillarious. One issue I found with AppVeyor is long == 32-bits under windows. 64-bits under Linux. Joy...

gizmocuz:
Still I think we should use c++ style casting

https://stackoverflow.com/questions/332030/when-should-static-cast-dynamic-cast-const-cast-and-reinterpret-cast-be-used/332086#332086

neheb:
That link calls it a function style cast. It's C++ only. But your call.

gizmocuz:
Having some merge conflicts here, and it's also not a desired patch.

 return float(atof(readValue.c_str()));

should be

 return static_cast<float>(atof(readValue.c_str()));

as this is the preferred way of c++ style casting

neheb:
Alright.

PR #4602, opened 2021-01-29
add missing m_isTemperature fill

neheb:
Mistakenly removed with the last commit.


rwaaren:
Thx for the quick fix.

PR #4604, opened 2021-01-30
Updated github action, added artifact upload and excluded some actions for forked repos

kiddigital:
Small PR to update the Github actions so when using a GitHub fork of **domoticz/domoticz**, the Github action workflow for the development branch within the fork does not fail all the time (as it cannot upload certain artifacts as this is private to the main domoticz repo).

Also, the create package is now also made available as a Github artifact with a 7 day retenion instead of the default 90, but that seems way to long because of the update frequency of the development branch. As there would be a huge amount of Artifacts created, one for each update that triggers the workflow.

### DIFF
--- a/.github/workflows/development.yml
+++ b/.github/workflows/development.yml
@@ -74,9 +74,18 @@ jobs:
           shasum -a 256 package/domoticz_linux_x86_64.tgz > package/domoticz_linux_x86_64.tgz.sha256sum
           cp appversion.h.txt package/version_linux_x86_64.h
           cp History.txt package/history_linux_x86_64.txt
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thank you for this PR!! One question, the ' Artifact upload', where does it upload to ?

kiddigital:
@gizmocuz 

To the standard 'Artifact' area for the build run, see for example this action triggerd by updating my development branch:
https://github.com/kiddigital/domoticz/actions/runs/523618811

gizmocuz:
Thank you for this PR!

PR #4605, opened 2021-01-31
OTGW: Fix to handle the new v5 firmware

kiddigital:
This PR is make the existing OTGW code handle the modified PS=1 output line from the new v5 firmware.

The v5 firmware adds new fields and also changes the order of the fields (new fields are inserted not at the end, but in between at certain places).

This PR does not do anything yet with the 9 new fields (mostly about support for a 2nd zone and ventilation support, both things that most boilers do not support/have yet), it just makes sure that the existing values are handled correctly.

See this forum thread for more info:
https://www.domoticz.com/forum/viewtopic.php?f=6&t=35384

As I don't have an OTGW I couldn't test it myself so it has been tested by forum user **TheSpanishInq** who runs an OTGW with firmware version 5.

_NOTE: I don't think this PR should be merged right now, but first also tested by someone with an OTGW that runs an older firmware version to be sure that these older versions still work properly. Not something I can test either._

### DIFF
--- a/hardware/OTGWBase.cpp
+++ b/hardware/OTGWBase.cpp
@@ -214,7 +214,7 @@ bool OTGWBase::SwitchLight(const int idx, const std::string &LCmd, const int /*s
 		}
 		else
 		{
-			Log(LOG_ERROR, "OTGW: Invalid switch command received!");
+			Log(LOG_ERROR, "Invalid switch command received!");
 			return false;
 		}
 		WriteInt((const unsigned char*)&szCmd, (const unsigned char)strlen(szCmd));
@@ -257,7 +257,7 @@ bool OTGWBase::WriteToHardware(const char *pdata, const unsigned char /*length*/
 	}
 	else
 	{
-		Log(LOG_STATUS, "OTGW: Skipping writing to Hardware for type: %02X, subType: %02X", packettype, subtype);
+		Log(LOG_STATUS, "Skipping writing to Hardware for type: %02X, subType: %02X", packettype, subtype);
 	}
 	return true;
 }
@@ -315,7 +315,7 @@ void OTGWBase::SetSetpoint(const int idx, const float temp)
 	if (idx == 1)
 	{
 		//Control Set Point (MsgID=1)
-		Log(LOG_STATUS, "OTGW: Setting Control SetPoint to: %.1f", temp);
+		Log(LOG_STATUS, "Setting Control SetPoint to: %.1f", temp);
 		sprintf(szCmd, "CS=%.1f\r\n", temp);
 		WriteInt((const unsigned char*)&szCmd, (const unsigned char)strlen(szCmd));
 	}
@@ -323,7 +323,7 @@ void OTGWBase::SetSetpoint(const int idx, const float temp)
 	{
 		//Room Set Point
 		//Make this a temporarily Set Point, this will be overridden when the thermostat changes/applying it's program
-		Log(LOG_STATUS, "OTGW: Setting Room SetPoint to: %.1f", temp);
+		Log(LOG_STATUS, "Setting Room SetPoint to: %.1f", temp);
 		sprintf(szCmd, "TT=%.1f\r\n", temp);
 		WriteInt((const unsigned char*)&szCmd, (const unsigned char)strlen(szCmd));
 		UpdateSetPointSensor((uint8_t)idx, temp, "Room Setpoint");
@@ -331,7 +331,7 @@ void OTGWBase::SetSetpoint(const int idx, const float temp)
 	else if (idx == 15)
 	{
 		//DHW setpoint (MsgID=56)
-		Log(LOG_STATUS, "OTGW: Setting Heating SetPoint to: %.1f", temp);
+		Log(LOG_STATUS, "Setting Heating SetPoint to: %.1f", temp);
 		sprintf(szCmd, "SW=%.1f\r\n", temp);
 		WriteInt((const unsigned char*)&szCmd, (const unsigned char)strlen(szCmd));
 		UpdateSetPointSensor((uint8_t)idx, temp, "DHW Setpoint");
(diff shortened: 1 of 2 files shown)


gizmocuz:
Thank you for the patch/PR. Maybe instead of using idx++ (which was a easy way), it might be better to use #define's for this

		_status.Room_temperature = static_cast<float>(atof(results[OTGW_ROOM_TEMPERATURE_ID].c_str()));						SendTempSensor(OTGW_ROOM_TEMPERATURE_ID - 1, 255, _status.Room_temperature, "Room Temperature");
		_status.Boiler_water_temperature = static_cast<float>(atof(results[OTGW_WATER_TEMPERATURE_ID].c_str()));				SendTempSensor(OTGW_WATER_TEMPERATURE_ID - 1, 255, _status.Boiler_water_temperature, "Boiler Water Temperature");

Will this make it also backwards compatible ?

kiddigital:
@gizmocuz , was indeed thinking about something like that to keep it more backwards compatible.

Only 'issue' I see is that the 'old' idx numbering was based on the 25 fields and their order. It would be more future proof if we would switch to using the OpenTherm IDs so when new items become available, they will be using new ID's anyway. But that would break compatibility with the current situation.

So it is a dilemma :( We could introduce a 'OTGW v2' hardware module? Which works with both old and new version and uses OpenTherm ID's so we can move forward for a long(er) time? People with the old OTGW firmware can keep using the current one. Would that be a possible way-out?

gizmocuz:
How about making a array/lookup table ? ID verses IDX ?

kiddigital:
@gizmocuz I added a simple define list to keep things compatible.

Some more testing/testers are needed to confirm everything works as expected both for the new firmware as for the old.

kiddigital:
For those that want to know more about the v5 Firmware, check out [this](https://www.domoticaforum.eu/viewtopic.php?f=75&t=13054) forum thread at the domoticaforum (where most of the OTGW support is done).

gizmocuz:
Are there new sensors in the V5 firmware ? Is it possible to add them instead of doing

```
		if(m_bFirmware5)
		{
			idx+=2;
		}
```

(If they are of interest of course that is ;) )

kiddigital:
Yes there are 9 new 'sensors' of which I think a few at least are of interest. Especially the 2 (or 3) related to a second zone (CH2). But not sure if/how we should implement those? Should we use a setting to enable a second zone in the hardware settings before adding them? How else to know if a second zone is available? Not sure if it can be determined from the values? (Does a value of 0 mean not available, off or just 0). And currently, of the 'old' 25 values only some are translated into sensors. Maybe some should be added as well?

Problem/challenge for me is that I am not a user/owner of an OTGW module so hard to determine what is usable.

The concern addressed with this PR is that an owner of an OTGW module with a new v5 firmware still can use the Domoticz OTGW module as it was before.

But yes, if someone can suggest what/why/which of the new values to add as sensors, I will give it a go 😁

gizmocuz:
Thank you for the PR and all the help!

gizmocuz:
New beta version available (also added DHW Flow Rate) (version 12913)

hardware/OTGWBase.cpp
```diff
@@ -396,11 +396,6 @@ void OTGWBase::ParseLine()
 		int idx=0;
 		bool bIsFirmware5 = (results.size()==34);
 
-		std::string sFirmwareMsg = "pre version 5 firmware";
-		if (bIsFirmware5)
-			sFirmwareMsg = "firmware 5 or higher";
-		Debug(DEBUG_HARDWARE, "OTGW: Assuming %s! Parsing the following PS=1 input line : .%s.", sFirmwareMsg.c_str(), sLine.c_str());
-
```
kiddigital, line 351: @gizmocuz , Is there a specific reason to remove this? As this message uses a specific Debug loglevel (hardware) I thought it makes sense to have something in there as it helps OTGW users/owners to get specific insights in the PS=1 messages from their OTGW module if they really wanted to. Something they otherwise can not see unless hooking up for example the specific OTmonitor software.

These message do not appear in the normal (debug) logging anyway.

(Just to understand the reasoning as I thought the extra debuglevels where there to make it both easier to debug but also much more specific and not flooding normal (debug)logging?)

gizmocuz: @kiddigital , I run in hardware debug mode, and after seeing this message 100 times, I know it's v4 or v5.
If you want to display this information for a specific reason, I would add a boolean (m_bDisplayedVersion), default false, and display the firmware version 1 time if the boolean is false, and set the boolean to true

gizmocuz: Also, although it ain't logged, the message is still created and send to the log function

kiddigital: Ok clear. Agree that the versioning info is not relevant to keep repeating. Doing it once at start is indeed sufficient. But now also the PS=1 line info is not displayed when in (certain) debug mode. Isn't that the relevant info that you would like to see during debugging (as every reading contains the actual readings)? So you can see/find odd readings, find out if things have changes, etc.

For me it is fine right now, let's see if there is demand for additional debug info from Domoticz and than we can always add something :)

(The desire to buying an OTGW myself has been bumped up a lot on my wish list because of working on this ;) )

gizmocuz: Let's see if someone needs this information.
Love my OTGW (connected via ser2net on a TPLink nano

sincze: > Let's see if someone needs this information.
> Love my OTGW (connected via ser2net on a TPLink nano

Same kind of setup here. What should/can I do to assist ? 
Received a new boiler yesterday.

gizmocuz: @sincze , I think all is fine and you should be able to use the latest beta version with your OTGW and also when you upgrade your OTGW to firmware v5

PR #4614, opened 2021-02-04
Tesla quick fix for authorisation: added support for manual API key

MrHobbes74:
The tesla authorisation has changed, hence the tesla module does not work anymore.
Implementing the new authorisation procedure is going to take a while.
To allow users to keep using the tesla module for now, I have added the option to add a generated API key (instead of retrieving it via the authorisation procedure).


gizmocuz:
Thank you !!

PR #4615, opened 2021-02-05
clang-tidy: use bind

neheb:
Found with modernize-use-bind

fix shared_from_this() use in asio calls

Uses C++14 auto lamba parameters

### DIFF
--- a/hardware/1Wire.cpp
+++ b/hardware/1Wire.cpp
@@ -72,11 +72,11 @@ bool C1Wire::StartHardware()
 
 	// Start worker thread
 	if (m_sensorThreadPeriod != 0) {
-		m_threadSensors = std::make_shared<std::thread>(&C1Wire::SensorThread, this);
+		m_threadSensors = std::make_shared<std::thread>([this] { SensorThread(); });
 		SetThreadName(m_threadSensors->native_handle(), "1WireSensors");
 	}
 	if (m_switchThreadPeriod != 0) {
-		m_threadSwitches = std::make_shared<std::thread>(&C1Wire::SwitchThread, this);
+		m_threadSwitches = std::make_shared<std::thread>([this] { SwitchThread(); });
 		SetThreadName(m_threadSwitches->native_handle(), "1WireSwitches");
 	}
 	m_bIsStarted = true;
--- a/hardware/1Wire/1WireByKernel.cpp
+++ b/hardware/1Wire/1WireByKernel.cpp
@@ -31,7 +31,7 @@
 
 C1WireByKernel::C1WireByKernel()
 {
-	m_thread = std::make_shared<std::thread>(&C1WireByKernel::ThreadFunction, this);
+	m_thread = std::make_shared<std::thread>([this] { ThreadFunction(); });
 	SetThreadName(m_thread->native_handle(), "1WireByKernel");
 	_log.Log(LOG_STATUS, "Using 1-Wire support (kernel W1 module)...");
 }
--- a/hardware/ASyncSerial.cpp
+++ b/hardware/ASyncSerial.cpp
@@ -33,7 +33,6 @@
 #include <algorithm>
 #include <iostream>
 #include <boost/asio.hpp>
-#include <boost/bind/bind.hpp>
 #include <boost/function.hpp>
 #include <boost/thread.hpp>
 #include <boost/smart_ptr/shared_array.hpp>  // for shared_array
@@ -100,10 +99,11 @@ void AsyncSerial::open(const std::string& devname, unsigned int baud_rate,
         boost::asio::serial_port_base::flow_control opt_flow,
         boost::asio::serial_port_base::stop_bits opt_stop)
 {
-    if(isOpen()) close();
+	if (isOpen())
+		close();
 
-    setErrorStatus(true);//If an exception is thrown, error_ remains true
-    pimpl->port.open(devname);
+	setErrorStatus(true); // If an exception is thrown, error_ remains true
+	pimpl->port.open(devname);
 	try {
 		pimpl->port.set_option(boost::asio::serial_port_base::baud_rate(baud_rate));
 		pimpl->port.set_option(opt_parity);
@@ -120,13 +120,13 @@ void AsyncSerial::open(const std::string& devname, unsigned int baud_rate,
 
 	pimpl->io.reset();
 
-    //This gives some work to the io_service before it is started
-    pimpl->io.post(boost::bind(&AsyncSerial::doRead, this));
+	// This gives some work to the io_service before it is started
+	pimpl->io.post([this] { return doRead(); });
 
-    boost::thread t(boost::bind(&boost::asio::io_service::run, &pimpl->io));
-    pimpl->backgroundThread.swap(t);
-    setErrorStatus(false);//If we get here, no error
-    pimpl->open=true; //Port is now open
+	boost::thread t([this] { pimpl->io.run(); });
+	pimpl->backgroundThread.swap(t);
+	setErrorStatus(false); // If we get here, no error
+	pimpl->open = true;    // Port is now open
 }
 
 void AsyncSerial::openOnlyBaud(const std::string& devname, unsigned int baud_rate,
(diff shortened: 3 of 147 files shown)


neheb:
@wijagels please check. Since this is C++14 now, I added auto everywhere.

gizmocuz:
I found no issues with this (tested web interface, TCP, Serial). Performance stays the same

wijagels:
> @wijagels please check. Since this is C++14 now, I added auto everywhere.

Looks good, change was just to add auto for the lambda parameters right?

neheb:
Yeah. I also changed a bunch of Do_Work() and whatnot calls to use lambdas. Simpler and saves size.

gizmocuz:
@neheb , tell me when you want this to be merged

neheb:
I have nothing to add to this PR. Might as well.

gizmocuz:
Thanks again for the many work and refactoring !

xbeaudouin:
Oh... It seems I have some... issues again on FreeBSD : 

```
/home/kiwi/domoticz.dev/main/WebServer.cpp:315:103: error: reference to '_1' is ambiguous
                        m_pWebEm->RegisterIncludeCode("switchtypes", std::bind(&CWebServer::DisplaySwitchTypesCombo, this, _1));
                                                                                                                           ^
/usr/local/include/boost/bind/placeholders.hpp:46:38: note: candidate found by name lookup is 'boost::placeholders::_1'
BOOST_STATIC_CONSTEXPR boost::arg<1> _1;
                                     ^
/usr/include/c++/v1/functional:2574:46: note: candidate found by name lookup is 'std::__1::placeholders::_1'
/* _LIBCPP_INLINE_VAR */ constexpr __ph<1>   _1{};
                                             ^
/home/kiwi/domoticz.dev/main/WebServer.cpp:316:101: error: reference to '_1' is ambiguous
                        m_pWebEm->RegisterIncludeCode("metertypes", std::bind(&CWebServer::DisplayMeterTypesCombo, this, _1));
                                                                                                                         ^
/usr/local/include/boost/bind/placeholders.hpp:46:38: note: candidate found by name lookup is 'boost::placeholders::_1'
BOOST_STATIC_CONSTEXPR boost::arg<1> _1;
                                     ^

```

neheb:
Yeah I noticed the same on Ubuntu. I'll make a PR.

gizmocuz:
@neheb , could this be related to this PR?

https://github.com/domoticz/domoticz/issues/4621

neheb:
No idea.

gizmocuz:
@neheb, following the discussion in the other issue, this PR seems to have caused a problem (besides compilation)
Could you think of a possible cause (python plugin related in this case), it stops after a certain time

pipiche38:
May be to help to narrow the issue, what seems to create an issue is that within the plugin, 

1/ create/open a TCP/IP connection with Protocol HTTP.
2/ listen on the connection to receive messages and process them.


In parallel, the plugin I'm maintaining does the same on Serial line and I have no issue. Did the same test with TCP/IP protocol None and seems to work as well.

So it looks like the issue is around TCP/IP with HTTP protocol

Happy to help and do tests (as most of my community will be down, as we rely on this ),

pipiche38:
I just recompile the all domoticz without the changes into the 3 plugin related files for which I used the previous version.
And the plugin behave has before without issue!

gizmocuz:
Does this mean there is no issue?

pipiche38:
seems so ... I'll investigate more ... maybe a mistake in the code there ... it looks like it was done automatically as comment have been updated

gizmocuz:
That could be great news

pipiche38:
@gizmocuz , @neheb , so I do confirm that the changes on the 3 files under hardware/plugins are creating an issue.

After recompiling the all domoticz without a change for those 3 files, the plugin was working again. In order to get a conclusive result, I then rebuild domoticz after having done git reset -hard (to get the changes). The end result is non-working plugin and stale thread.

So there is change somewhere which seems break something. My C++ knowledge is limited and I cannot really do more

PluginTransports.cpp
PluginManager.cpp
Plugin.cpp

neheb:
Change for Plugins.cpp:

```
			m_thread = std::make_shared<std::thread>(&CPlugin::Do_Work, this);
			m_thread = std::make_shared<std::thread>([this] { Do_Work(); });
```

Change for PluginManager.cpp:

```
		m_thread = std::make_shared<std::thread>(&CPluginSystem::Do_Work, this);
		m_thread = std::make_shared<std::thread>([this] { Do_Work(); });
```

Would be great if you could revert just those and see if you still get the error.

I think the actual issue is in PluginTransports.cpp , which has a lot of changes.

neheb:
The other potential issue I see is:

```
--- a/hardware/plugins/PluginTransports.cpp
+++ b/hardware/plugins/PluginTransports.cpp
@@ -213,7 +213,7 @@ namespace Plugins {
 			}
 
 			pTcpTransport->m_Socket->async_read_some(boost::asio::buffer(pTcpTransport->m_Buffer, sizeof pTcpTransport->m_Buffer),
-								 [this](const auto &err, auto bytes) { handleRead(err, bytes); });
+								 boost::bind(&CPluginTransportTCP::handleRead, pTcpTransport, boost::asio::placeholders::error, boost::asio::placeholders::bytes_transferred));
 
 			// Requeue listener
 			if (m_Acceptor)

```

handleRead takes two parameters, not three. but the bind has three parameters.

neheb:
Interesting...

clang-tidy's autogenerated version of that bind call is:

```
[pTcpTransport] { pTcpTransport->handleRead(boost::asio::placeholders::error, boost::asio::placeholders::bytes_transferred); });
```

edit: modifying it a bit to get it to compile:

```
[pTcpTransport, this] (auto &&err, auto bytes) { pTcpTransport->handleRead(err, bytes); });
```

Looks like the correct solution.

nipsgur:
Hello neheb,

I'm running domoticz still on Stretch (I know, it is not supported anymore, but I could still compile it so far and was happy about it).
Since this changes to mainworker.cpp I get compiler errors and it does not finish.
The following changes seem to be related to my problem.

`pHardware->sDecodeRXMessage.connect(boost::bind(&MainWorker::DecodeRXMessage, this, _1, _2, _3, _4, _5));
	pHardware->sOnConnected.connect(boost::bind(&MainWorker::OnHardwareConnected, this, _1));
	pHardware->sDecodeRXMessage.connect([this](auto hw, auto rx, auto name, auto battery, auto userName) { DecodeRXMessage(hw, rx, name, battery, userName); });
	pHardware->sOnConnected.connect([this](auto hw) { OnHardwareConnected(hw); });`

But I'm not really much into c++, maybe you could help me out what this is about?
The error message is lengthy but I think this is the problem.

`mainworker.cpp:1189:110: error: cannot call member function ‘void MainWorker::DecodeRXMessage(const CDomoticzHardwareBase*, const uint8_t*, const char*, int, const char*)’ without object
 Message.connect([this](auto hw, auto rx, auto name, auto battery, auto userName) { DecodeRXMessage(hw, rx, name, battery, userName); });`

I would be glad about any hints, how to fix this and still compile on Stretch.

neheb:
I’ll whip up a VM to see what’s happening. I actually have a reason to have GCC 6 working.

nipsgur:
cool, thanks in advance!
Updating libboost to 1.75 and cmake to 3.19.4 didn't do the job for me. And I don't understand enough of the c++ libraries to get a clue, what might be the cause.

neheb:
It's definitely a GCC bug: https://stackoverflow.com/questions/32097759/calling-this-member-function-from-generic-lambda-clang-vs-gcc

I solved the issue on Debian 9 by installing clang-6.0 and setting -DCMAKE_CXX_COMPILER=clang-6.0

edit: sorry -DCMAKE_C_COMPILER=clang-6.0 -DCMAKE_CXX_COMPILER

edit2: also requires https://github.com/domoticz/domoticz/pull/4624 because of boost.

nipsgur:
cool, changing to clang and it compiles now with

`cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_COMPILER=clang-6.0 -DCMAKE_CXX_COMPILER=clang++-6.0 CMakeLists.txt`

thanks! Although I didn't understand edit2, but I guess it should be taken into account in the latest release.

Only, it throws quite a number of warning like:

`dev-domoticz/tinyxpath/xpath_stream.h:73:18: warning: enumeration value 's_end' not handled in switch [-Wswitch]
         switch (state)`

neheb:
I was using boost1.67 since that's what stretch-backports comes with. Seems you're using 1.75 which has that fixed.

The switch warning are because clang is more verbose when it comes to warnings than gcc.

edit: btw it's possible to work around this bug by making the pointer explicit. I don't know if it's worth it though...

nipsgur:
> edit: btw it's possible to work around this bug by making the pointer explicit. I don't know if it's worth it though...

Not because of only me.
The LTS support for Stretch is on the horizon, so sooner or later I have move to buster anyhow.

PR #4619, opened 2021-02-06
get device from database with ID as searchkey

rwaaren:
This will prevent an error message when device is just renamed with API call

See this [forum post](https://www.domoticz.com/forum/viewtopic.php?f=59&t=35451)

### DIFF
--- a/hardware/Kodi.cpp
+++ b/hardware/Kodi.cpp
@@ -567,7 +567,7 @@ void CKodiNode::UpdateStatus()
 	if (m_CurrentStatus.Status() != m_PreviousStatus.Status())
 	{
 		m_notifications.CheckAndHandleNotification(m_ID, m_Name, m_CurrentStatus.NotificationType(), sLogText);
-		m_mainworker.m_eventsystem.ProcessDevice(m_HwdID, m_ID, 1, int(pTypeLighting2), int(sTypeAC), 12, 100, int(m_CurrentStatus.Status()), m_CurrentStatus.StatusMessage().c_str(), m_Name);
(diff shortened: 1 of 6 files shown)


gizmocuz:
Basically what you do is not using the devname that ProcessDevice has as function argument.
Maybe m_eventsystem.GetCurrentStates() needs to be called when renaming a device ?

rwaaren:
> Basically what you do is not using the devname that ProcessDevice has as function argument.
> Maybe m_eventsystem.GetCurrentStates() needs to be called when renaming a device ?

Thx. Will try to do that. 

btw. Is there a reason why this function works with the "none unique" Name ?
I would expect it would use the unique ID.  Do I overlook something?

gizmocuz:
No idea, I haven't started this class ;)... If you think it could be better, please go ahead.
(The less database queries the better for speed)

rwaaren:
@gizmocuz

tried to solve it with your proposed change in CWebServer::Cmd_RenameDevice but then the timing issue persists. In some cases (on my system in about 20% of the renames) a subsequent call to the device immediately following the rename fails with. 

**2021-02-07 11:47:28.333  Error: EventSystem: Could not find device in system:**

I guess this is a side effect of #4585
 
In this PR I already use the unique ID and no extra database call is done. 
Is it OK with you to merge this PR as it is now?

gizmocuz:
I find it very odd that a device name is provided for/by the function, and inside the function we need to query the database for the name.
Something is not called in the correct order....

rwaaren:
> I find it very odd that a device name is provided for/by the function, and inside the function we need to query the database for the name.

The name is now only used in the error message. It is no longer needed for the functionality as the database query is done based on the ID. 
Let me remove the devicename from the function calls.

PR #4620, opened 2021-02-06
boost to std conversions

neheb:


gizmocuz:
Do you think this can also be included ?

https://think-async.com/Asio/

It can work in non-boost mode as well

neheb:
:). That's gonna take a while. I still can't figure out CMake and how it does dependencies.

neheb:
Rebased.

neheb:
Rebased.

webserver/cWebem.cpp
```diff
@@ -770,7 +770,7 @@ namespace http {
 				std::string attachment;
 				for (const auto &header : rep.headers)
 				{
-					if (boost::iequals(header.name, "Content-Disposition"))
+					if (header.name == "Content-Disposition")
```
gizmocuz: This wont work as it is a case insensitive compare...
You could add the below two functions to helper.h/cpp and use those:

bool iequals(const std::string &Input, const std::string &Test)
{
	return Input.size() == Test.size()
	       && equal(Input.cbegin(), Input.cend(), Test.cbegin(),
			[](std::string::value_type l1, std::string::value_type r1) {
		       return toupper(l1) == toupper(r1);
	       });
}

bool iequals(const std::wstring &Input, const std::wstring &Test)
{
	return Input.size() == Test.size()
	       && equal(Input.cbegin(), Input.cend(), Test.cbegin(),
			[](std::wstring::value_type l1, std::wstring::value_type r1) { return towupper(l1) == towupper(r1); });
}

gizmocuz (changes requested):

webserver/cWebem.cpp
neheb: This will take a while. I need to rebuild my dev machine.

neheb: I got rid of this change. Maybe will tackle at another time...

hardware/1Wire/1WireByKernel.cpp
```diff
@@ -551,7 +544,8 @@ void C1WireByKernel::GetDevice(const std::string& deviceName, /*out*/_t1WireDevi
 	device.family = ToFamily(deviceName.substr(0, 2));
 
 	// Device Id (6 chars after '.')
-	std_to_upper(deviceName.substr(3, 3 + 6 * 2), device.devid);
+	auto substr = deviceName.substr(3, 3 + 6 * 2);
+	std::transform(substr.begin(), substr.end(), device.devid.begin(), ::toupper);
```
gizmocuz: In our helper.h/cpp we have a function 'stdupper'

hardware/NestOAuthAPI.cpp
```diff
@@ -562,7 +562,7 @@ void CNestOAuthAPI::GetMeterDetails()
 
 			// Find out if we're using C or F.
 			std::string temperatureScale = ndevice["temperature_scale"].asString();
-			boost::to_lower(temperatureScale);
+			std::transform(temperatureScale.begin(), temperatureScale.end(), temperatureScale.begin(), ::tolower);
```
gizmocuz: You could use the helper.c/h function 'stdlower'

gizmocuz:
All looks perfect except the isequal and maybe use the helper.c/h functions

hardware/1Wire/1WireByKernel.cpp
neheb: Funny. The last commit converts it to be identical.

hardware/NestOAuthAPI.cpp
neheb: Will do.

neheb:
Rebased. From 10069656 to 9918104 locally.

gizmocuz:
I think this PR is split into multiple PR's, could this PR be closed ?

PR #4624, opened 2021-02-08
revert to using boost::bind instead of std::bind

neheb:
Fixes compilation issues with earlier versions of boost where the
placeholders are "ambiguous".

### DIFF
--- a/main/WebServer.cpp
+++ b/main/WebServer.cpp
@@ -66,8 +66,6 @@
 #define __STDC_FORMAT_MACROS
 #include <inttypes.h>
 
-using namespace std::placeholders;
-
 #define round(a) ( int ) ( a + .5 )
 
 extern std::string szStartupFolder;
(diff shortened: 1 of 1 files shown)


gizmocuz:
@neheb , I think this might be caused by the definition of (for example) webem_include_function in cWebem.h (as it is using boost::function)
Maybe if this was changed to non-boost, it will work/compile directly ? Would be a shame if we need to depend on boost for this...

neheb:
Weird. In this PR: https://github.com/domoticz/domoticz/pull/4620 , I also have a similar commit as it wouldn't build. One of the commits there changes boost::function to std::function but same problem.

I'll look some more...

gizmocuz:
Maybe if you also change the declaration in cWebem.h (replace the boost::function) it will work ?

neheb:
No. The actual problem was fixed with boost 1.73. Internally in many boost headers, there's a using namespace boost::placeholders.

neheb:
Yeah, the problem is in the asio headers, which cannot simply be removed: https://www.boost.org/users/history/version_1_73_0.html

gizmocuz:
True, but if we could remove the boost bind dependency, we do not need this...

neheb:
I can convert to lambdas. My original reasoning for not doing so is the size increase. But that can be offset by other changes.

neheb:
Removed bind. These changes were done automatically.

neheb:
Closing in favor of https://github.com/domoticz/domoticz/pull/4646

Issue #4623, opened 2021-02-08
EnOcean ESP3: adding profile F6-05-02 (smoke detector)


alexis-via:
I have an Enocean smoke detector Ubiwizz UBILD001-QM which uses the Enocean profile F6-05-02 and an Enocean dongle USB310.

I know that this profile is **not** supported yet by Domoticz, and I'm working on implementing it. I studied the Enocean Serial Protocol 3 and the details of the profile F6-05-02 (a very simple profile !). And I started to dive in domoticz/hardware/EnOceanESP3.cpp

When I press the test button of my smoke detector, it enters the method ParseRadioDatagram https://github.com/domoticz/domoticz/blob/development/hardware/EnOceanESP3.cpp#L1117 and I get the following logs :

2021-02-08 21:20:19.110  EnOcean: destination: 0xffffffff RSSI: 30
2021-02-08 21:20:19.110  EnOcean: RPS data: Sender id: 0x058e8fbd Status: 00 Data: 00
2021-02-08 21:20:20.134  Status: EnOcean: 01 (07/07) F6 00 05 8E 8F BD 00 00 FF FF FF FF 4C 00

In this method ParseRadioDatagram, when I look at "case RORG_RPS" https://github.com/domoticz/domoticz/blob/development/hardware/EnOceanESP3.cpp#L1680 (the smoke detector uses an RPS telegram), the code doesn't seem to be organized to support several profiles: it seems that it only implements profile F6-02-01 (Rocker switch, 2 Rocker), because it uses variables T21 and NU which are specific to that profile AFAIK. I would need to write code in there to add support for profile F6-05-02, but I feel the code is not ready to support a new profile. On the opposite, when I look at the code of "case RORG_4BS" https://github.com/domoticz/domoticz/blob/development/hardware/EnOceanESP3.cpp#L1169 I have the feeling that the code is organized to support several profiles.

How can we implement several profiles in case RORG_RPS ? 
How can we make the distinction in the code between F6-02-01 and F6-05-02 ? I must say that, after studying the Enocean telegram structure for packet type = Radio (0x01) and RORG = RPS, I don't know how we are supposed to differenciate the different profiles.

Thanks in advance for your help!

alexis-via:
[...]

PR #4629, opened 2021-02-10
Enocean: Add support for F6-05-xx profiles (smoke, wind and liquid detectors) + RSSI reporting

alexis-via:
Following discussion in issue #4623 
Screencast showing setup and use with my Enocean Ubiwizz UBILD001-QM smoke detector... with a real fire ! https://www.youtube.com/watch?v=s0fu1bFiu2Y

@gizmocuz I converted the log lines in the new code: instead of sprintf() + _log.Log(), I use Log() (without preceding sprintf). If you are OK with this, I'll change it everywhere in EnOceanESP3.cpp as you suggested.

If you agree, I propose to convert #ifdef ENOCEAN_BUTTON_DEBUG + _log.Log() by Debug(DEBUG_HARDWARE, "") without ifdef.

### DIFF
--- a/hardware/EnOceanESP3.cpp
+++ b/hardware/EnOceanESP3.cpp
@@ -1704,15 +1704,30 @@ void CEnOceanESP3::ParseRadioDatagram()
 				long id = (ID_BYTE3 << 24) + (ID_BYTE2 << 16) + (ID_BYTE1 << 8) + ID_BYTE0;
 				char szDeviceID[20];
 				sprintf(szDeviceID,"%08X",(unsigned int)id);
+				int Profile;
+				int iType;
 
 				// if a button is attached to a module, we should ignore it else its datagram will conflict with status reported by the module using VLD datagram
 				std::vector<std::vector<std::string> > result;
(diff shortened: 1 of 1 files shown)


alexis-via:
Two things I wanted to say:
1) Most of the diff is just adding one indentation for the code that handle the profile F6-02-01, to move that code inside an "IF profile = 2 and itype = 1". But the code itself hasn't been changed.
2) It's the first time I contribute to Domoticz and it's also the first time I write C++... so review carefully :)

gizmocuz:
Thank you for the PR !!
For the switch, could you use the function

SendSwitch(int NodeID, uint8_t ChildID, int BatteryLevel, bool bOn, double Level, const std::string &defaultname, const std::string &userName, int RssiLevel = 12);

That is the new method for new switches , so the old switches have to be like it is now for backwards compatibility

You have one line

if (result.empty())

and further

else if (result.size()==1)

I think you can just use 'else' (without result.size......)

alexis-via:
@gizmocuz  I took into account your 2 suggestions in my last commit. And I tested again with a real fire -> it works.

gizmocuz:
Thank you very much for this PR!!

PR #4631, opened 2021-02-11
some more lambdas

neheb:
Smaller size

### DIFF
--- a/hardware/plugins/PluginTransports.cpp
+++ b/hardware/plugins/PluginTransports.cpp
@@ -67,7 +67,7 @@ namespace Plugins {
 				//
 				//	Async resolve/connect based on http://www.boost.org/doc/libs/1_45_0/doc/html/boost_asio/example/http/client/async_client.cpp
 				//
-				m_Resolver.async_resolve(query, [this](const auto &err, auto end) { handleAsyncResolve(err, end); });
+				m_Resolver.async_resolve(query, [this](auto &&err, auto end) { handleAsyncResolve(err, end); });
 			}
 		}
 		catch (std::exception& e)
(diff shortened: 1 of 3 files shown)


hardware/plugins/PluginTransports.cpp
```diff
@@ -67,7 +67,7 @@ namespace Plugins {
 				//
 				//	Async resolve/connect based on http://www.boost.org/doc/libs/1_45_0/doc/html/boost_asio/example/http/client/async_client.cpp
 				//
-				m_Resolver.async_resolve(query, [this](const auto &err, auto end) { handleAsyncResolve(err, end); });
+				m_Resolver.async_resolve(query, [this](auto &&err, auto end) { handleAsyncResolve(err, end); });
```
gizmocuz, line 70: &&err ? Are you sure ? (Just double checking)

neheb: Yeah. It's shorter and implicitly convertable to const auto ref.

neheb: Actually passing auto&& for a lambda parameter doesn't work right for a function accepting && as a parameter. You still have to call std::move.

gizmocuz: Is && not the same as a pointer ? It's just that I never seen ormused this before.... & / * / ** But never &&

neheb: It's an rvalue reference. Also known as a universal reference since it implicitly converts to many things.

Note that the clang-tidy modernize-avoid-bind check uses then by default. I didn't originally use them for C++11 compatibility (no auto in lambdas)

neheb:
Closing in favor of https://github.com/domoticz/domoticz/pull/4646

PR #4633, opened 2021-02-11
dzVents update to 3.2.0

rwaaren:
```
======== domoticz V2020.2 (12936), dzVents V3.2.0 +========================== Tests ==============================+
  time  | test-script                              | expected | tests | result  | successful |  failed  | seconds  |
===================================================+===============================================================+
00:00:02  Device                                        1        119      OK            119         0      0.3014
00:00:03  Domoticz                                      1         85      OK             85         0      0.3233
00:00:03  EventHelpers                                  1         34      OK             34         0      0.1943
00:00:03  EventHelpersStorage                           1         50      OK             50         0      0.2170
00:00:03  HTTPResponse                                  1          6      OK              6         0      0.0148
00:00:04  Lodash                                        1        100      OK            100         0      0.1622
00:00:04  ScriptdzVentsDispatching                      1          2      OK              2         0      0.6345
00:00:04  TimedCommand                                  1         46      OK             46         0      0.1114
00:00:05  Time                                          3        369      OK            369         0      0.8017
00:00:06  Utils                                         1         36      OK             36         0      0.0683
00:00:06  Variable                                      1         15      OK             15         0      0.0333
00:00:06  ContactDoorLockInvertedSwitch                 3          2      OK              2         0      3.0608
00:00:09  DelayedVariableScene                          8          2      OK              2         0      7.0622
00:00:16  EventState                                   19          2      OK              2         0     18.0730
00:00:34  Integration                                  34        222      OK            222         0     29.2089
00:01:03  SelectorSwitch                               10          2      OK              2         0      9.1826
00:01:22  SystemAndCustomEvents                         1          9      OK              9         0      0.0409

Total tests: 1101
Thu 11 Feb 2021 03:53:31 PM CET; dzVents version 3.2.0 tested without errors after 00:01:22

```

- Add attribute user to switchable devices, groups and scenes (identifies who / what last updated the item)
- Add resetAllDevicestatus and resetAllEvents to type system trigger

thx to @MrHobbes74 for the system notifications

### DIFF
--- a/History.txt
+++ b/History.txt
@@ -6,6 +6,7 @@ Version 2020.xxxx (xxxx)
 - Implemented: Allow custom icons for thermostat setpoints on floorplan
 - Implemented: Allow custom icons for utility sensors
 - Implemented: Build systems for linux based on github actions
+- Implemented: Build system for docker
 - Implemented: Counter meter type now supports a divider
 - Implemented: Custom icons for RGB/W
 - Implemented: Hardware Monitor, Add clock speeds for Raspberry Pi
@@ -27,19 +28,22 @@ Version 2020.xxxx (xxxx)
 - Implemented: Websocket notification for secondary/sub devices
 - Implemented: ZWave, Legend and tooltips in Neighbor�s overview
 - Updated: Blebox; use of the latest SwitchBox API
-- Updated: dzVents; version 3.1.4 ( https://www.domoticz.com/wiki/DzVents:_next_generation_LUA_scripting )
+- Updated: dzVents; version 3.2.0 ( https://www.domoticz.com/wiki/DzVents:_next_generation_LUA_scripting )
 - Updated: eVehicles; added ODO meter, unlock/open alert, max charge level to Tesla module
+- Updated: eVehicles; added option to manual set API key
 - Updated: EvoHome; decode, display and store hotwater setpoint
 - Updated: Lay-out of notifications tab
 - Updated: MQTT; added option to choose notification subsystem
 - Updated: MQTT; added publish schemes: device index, device name and implemented custom topics for in- and outbound messages
 - Updated: OpenZWave; configuration files
 - Updated: Openweathermap; use latest API and overal improvements
+- Updated: OTGW; ready for firmware 5.0 and added domestic hot water flow rate
 - Updated: Plugwise; update of scene selector and added a migration process
 - Updated: SolarEdge; add support for 3 Phase inverter
 - Updated: Tado; add support for per-zone Open Window Detection
 - Updated: TTNMQTT; Better GPS/Locations handling, distance calculation (Geofencing)
 - Updated: TTNMQTT; Improved calculation and storage of signal-levels
+- Updated: Xiaomi; added new devicetypes
 - Changed: During a database backup we now 'sleep' when the status != OK
 - Changed: Differentiate kWh, kVah, kVar and kVarh
 - Changed: Energy devices allow negative values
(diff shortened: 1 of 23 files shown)


gizmocuz:
I'm not happy with the SQL table changes. Why is this needed ?
If you look in the switch log, you already see the last(current) user that switched the device

rwaaren:
I tried to get this information from the lightinglog en scenelog in an earlier PR but to get it for all devices and scenes extra database calls were needed and it was not possible to do this with an acceptable performance so I had to revert that PR. 
Using the method in this PR there is no extra database call and no performance penalty.

gizmocuz:
I think you can easly query this information in one call (where you also get all device states) (also select the last switch user from the light/scene, if it's not in this table it's empty
But with this patch you are creating extra (not often needed... only for dzVents?) fields , resulting in larger databases / storage
Better to query all device states (together with last switch user), and keep this central somewhere in Domoticz.
And keep this up 2 date always when adjusting values/fields

Quick workaround is to query this initially in dzVents, and form that moment keep this up2date

But extra calls just for dzVents... The information is there in the light/scene table

rwaaren:
I already have a couple of workarounds available and posted them on the forum for users looking for this type of information but this PR is meant to come to a solution. 

The user information is with this PR also directly available in the  /json.htm?type=devices and /json.htm?type=scenes API calls so not limited to dzVents.

the PR does not use extra database calls and I do think the increase in size of the database is quite limited but it is your call and if you know there are better ways to do this then it should be done using that other approach but I don't think my knowledge of domoticz is sufficient to do that.

Maybe someone else with more in depth knowledge of domoticz / c++ can have a go at it?

rwaaren:
@gizmocuz 

To workaround the need for an extra field 'user' in devicestatus
 
Would it be OK to create a view and  use that in CEventSystem::GetCurrentStates() ?

like:

```
CREATE VIEW LastUser AS SELECT DeviceRowID ID, User, MAX(Date) FROM LightingLog GROUP BY ID;

SELECT ds.HardwareID, ds.ID, ds.Name, ds.nValue, ds.sValue, ds.Type, ds.SubType, ds.SwitchType, ds.LastUpdate, ds.LastLevel, ds.Options, ds.Description, ds.BatteryLevel, ds.SignalLevel, ds.Unit, ds.DeviceID, ds.Protected, ds.AddjValue, ds.AddjMulti, ds.AddjValue2, ds.AddjMulti2, lu.User
	FROM DeviceStatus ds
	LEFT JOIN Hardware hw ON hw.ID = ds.HardwareID
	LEFT JOIN LastUser lu ON ds.ID = lu.ID
	WHERE hw.Enabled = 1;
```

although this could work It can only show information from records in lightinglog that are not cleaned yet and will probably have a higher impact on performance than the solution with the extra fields in devicestatus / scenes.

gizmocuz:
I think you can make a SQL Query in GetCurrentStates that also retreives the last user in the light/scene log in one query
Let me try this, otherwise there this functionality will not be available to dzVents, why is this needed ?

rwaaren:
> I think you can make a SQL Query in GetCurrentStates that also retreives the last user in the light/scene log in one query
> Let me try this, otherwise there this functionality will not be available to dzVents, why is this needed ?

"Needed" is a big word but the request to get this information readily available in dzVents pops up every couple of months in the forum.

gizmocuz:
True... But if users want to know who switched last they just need to look in the log.
I don't think it is a good idea to make the concisions for a few users.
Simplest solution (inpact on startup), is to get all devices status as it is now, and add two addition calls like:
SELECT A.ID, B.User FROM DeviceStatus AS A, LightingLog as B WHERE B.DeviceRowID = A.ID
SELECT A.ID, B.User FROM DeviceStatus AS A, SceneLog as B WHERE B.DeviceRowID = A.ID

And update the device status table

gizmocuz:
Better solution (other PR), is to have 1 current deviceStates table in memory that is always up 2 date, and used globally instead of always doing a sql query for a device state
This could also be a in-memory SQLite database table

rwaaren:
OK Back to the drawing board. Will close this one and come with a new stripped down version without the User information.

PR #4637, opened 2021-02-14
manual std::array conversions

neheb:


gizmocuz:
Sorry , this is making the code way less readable

meters[no_meters].id = id;
meters[no_meters].last_update = time(nullptr);

to:
	meters[no_meters].first = id;
	meters[no_meters].second = time(nullptr);

first, second...

And I think arrays not a big issue

neheb:
I'll look into it. This is not C++17 where there are structured bindings.

neheb:
I couldn't undo first/second in hardware/PhilipsHueHelper.cpp . The original struct is not usable in constexpr context.

neheb:
Found the issue. direct initialization was used for some reason...

gizmocuz:
Sorry, this is making TT he code unreadable, and probably slower as well

(m_Rego6XXTemp.ID.data()

neheb:
.data() is equivalent to const char *. There's no copying involved. std::array is a bit special in that sense.

Nevertheless, I'll split these up into smaller commits.

gizmocuz:
Yes, but I (we) don't want to call .data()
And using buffer[10] is also straight forward

neheb:
What's wrong with data() ?

gizmocuz:
Extra calls, extra typing, different declaration of buffers etc

neheb:
Doesn't result in extra assembly from my experience.

Anyway, I'll look into avoiding data().

gizmocuz:
@neheb , it is not only that

uint8_t m_varbuffer[8]

I see no problem with this

m_varbuffer[3] = 128;

I also see no problem with this

And yes, the one writing this code should check for boundary issues, but that's whats coding is about

std::string _szMyString="Hello world";
_szMyString[3] = _szMyString[5];

I see no issue with this, but yes, _szMyString.at(5) does the same

neheb:
The issue is that as written, some of these arrays are not constexpr (or even const). Making them so reduces compiled size as the compiler can more thoroughly optimize it.

There's also the issue of some uses of [] without a constant expression. std::array has at() for this, but that's not available with normal C arrays.

gizmocuz:
The code should also be pleasant to read/write... How many bytes are saved ?

neheb:
A quick look with this PR shows:

8313184 vs.
8296448

Not too big but this is a big project.

PR #4638, opened 2021-02-14
Warning text for openweatherMap forecast feature

jmleglise:
Add a warning text to this surprising feature... that creates 400 devices !!!

FYI, the former text is not i18n.

### DIFF
--- a/www/app/hardware/Hardware.html
+++ b/www/app/hardware/Hardware.html
@@ -1007,11 +1007,11 @@
 			</tr>
 			<tr>
 				<td align="right" style="width:110px"><label for="adddayforecast"><span data-i18n="Daily forecast">Daily forecast</span>:</label></td>
-				<td><input type="checkbox" id="adddayforecast">&nbsp;<label for="adddayforecast">&nbsp;Forecast for today and coming 7 days (if available).</label></td>
+				<td><input type="checkbox" id="adddayforecast">&nbsp;<label for="adddayforecast">&nbsp;Forecast for today and coming 7 days (if available). !! 80 devices will be created !!</label></td>
 			</tr>
 			<tr>
 				<td align="right" style="width:110px"><label for="addhourforecast"><span data-i18n="Hourly forecast">Hourly forecast</span>:</label></td>
-				<td><input type="checkbox" id="addhourforecast">&nbsp;<label for="addhourforecast">&nbsp;Forecast for each hour for 48 hours (if available).</label></td>
+				<td><input type="checkbox" id="addhourforecast">&nbsp;<label for="addhourforecast">&nbsp;Forecast for each hour for 48 hours (if available). !! 384 devices will be created !!</label></td>
 			</tr>
 		</table>
 	</div>


gizmocuz:
@jmberhault , thank you for this PR. Actually this is totally ridiculous indeed. It should be solved differently, maybe with a link that you can click

kiddigital:
@jmleglise Maybe a warning in the UI might be useful for those that just 'click-and-try' without reading any documentation.

The documentation on the Wiki is clear about the number of devices created.
https://www.domoticz.com/wiki/Open_Weather_Map

@gizmocuz By default, just 9 sensor about the current weather are created.

But if people decide they want the hourly forecast on 48! hours, offcourse a larger number of devices gets created. That is the whole purpose of wanting to add the 48 hours of detailed weather information in their setup. That is why OWM provides these types and details.

If users only care about a visual weather forecast, they can just use the weather forecast in the UI without the need to created these devices or even use the OWM module.

But if you for example want to create a more sophisticated automated garden watering system, you might be very well interested in hourly details about amount of downfall, precipation percentages, temperatures, etc. and are very happy that the OWM module makes the data that OWM provides available in their Domoticz setup so they can leverage it in an easy way.

(Without the need to custom code your own DYI solution. If so, one could even do that fine without using Domoticz, but I think that is one of the mean reason for an Home Automation system so you don't have to DIY everything :) )

jmleglise:
@kiddigital 

(It's inappropriate for me to comment your work because I don't know myself how to develop this)

But 80 devices + 380 devices does not seem reasonable to me. Moreover, if you remove the openweathermap hardware, are these devices removed? And if you recreate OWM afterwards, it adds another 400devices? (or are the same ones reused?)

A solution could be for the 7days, and the 48-hour forecast, to use only 9 devices each. Each concatenating the data of the forecast series. For example for the temperature, in a custom device text containing 48 figures : "-3,-2,-1,0,1,2 ...".  The users would then have to split it in order to use it in their script. Not a big deal.

kiddigital:
> @kiddigital 
> 
> 
> 
> (It's inappropriate for me to comment your work because I don't know myself how to develop this)
> 

No problem... always open to improve...

> 
> 
> But 80 devices + 380 devices does not seem reasonable to me. 

Why? Is it the amount of devices itself? The computer/Domoticz doesn't care 😁


> Moreover, if you remove the openweathermap hardware, are these devices removed? 

Yes, sort of.. they are not visible anymore but kept in the DB (as far as I am aware)

> And if you recreate OWM afterwards, it adds another 400devices? (or are the same ones reused?)
> 

I think Domoticz tries to re-use the already existing ones.

> 
> 
> A solution could be for the 7days, and the 48-hour forecast, to use only 9 devices each. Each concatenating the data of the forecast series. For example for the temperature, in a custom device text containing 48 figures : "-3,-2,-1,0,1,2 ...".  The users would then have to split it in order to use it in their script. Not a big deal.

Your suggestion 'just' lowers the amount of devices (why does that help the user?) but makes live for users much harder because they have to rely on scripting and process the data before the can use it. When using Blockly or other 'simple' automation it makes things unnecessary more complicated.

rwaaren:
Please note that this comment does not mean I personally object to the use of this many devices.
As long as the user is aware of it and of the CPU / Memory implications I guess it is up to her/him. 
But I would prefer another way with less devices needed if/ when such other way could be made available. 


> Why? Is it the amount of devices itself? The computer/Domoticz doesn't care 😁

True: "care" is a human emotion but an increase of devices will make domoticz use more memory and CPU cycles.
 
> > Moreover, if you remove the openweathermap hardware, are these devices removed?
>   -        Yes, sort of.. they are not visible anymore but kept in the DB (as far as I am aware)

No, devices are removed from the database when you remove the hardware where they are connected to. (when using the GUI or API to delete the hardware) 
 
 
> > And if you recreate OWM afterwards, it adds another 400devices? (or are the same ones reused?)
>    -       I think Domoticz tries to re-use the already existing ones.
No, see above.

gizmocuz:
@kiddigital , I am also not pointing to anyone, but there has to be a way to not create 48 devices.
Can't we just point to a URL with a forecast for 7 days ? (Based on the GPS coordinates) ?
We used to have (and maybe still have) a Forecast url option where the user could click on

kiddigital:
> @kiddigital , I am also not pointing to anyone, but there has to be a way to not create 48 devices.
> 
> Can't we just point to a URL with a forecast for 7 days ? (Based on the GPS coordinates) ?
> 
> We used to have (and maybe still have) a Forecast url option where the user could click on

Actually that is still there and even improved 😁 Still by default it gives the info from DarkSky (forecast.io) but you can configure it now to use OpenWeatherMap instead and make use of the ApiKey that is configured for the OWM module. The forecast screen will then show a gadget from OWM.

And btw.. also possible now to choose Buienradar and then a Buienradar gadget is shown.

I have modified/updated the 'weatherforecastinfo' command (not sure about the name, have to look it up) to return the forecastinfo (and url) for the given module (OWM, Buienradar or fallback to DarkSky). And updated the forecast screen to show the forecast info from the selected provider.

One thing I didn't do (yet) is properly document how to configure this and maybe even add it to the settings screen somewhere?

And to be clear, it is not needed to activate the forecast settings of the OWM module (which creates the (many) forecast devices) if you just want to use the forecast gadget of OWM on the forecast screen in the UI.

gizmocuz:
@kiddigital , what is the benefit to have all those devices created internally against using the forcast from OWM ?
Maybe this option could/should be removed ?

kiddigital:
> @kiddigital , what is the benefit to have all those devices created internally against using the forcast from OWM ?
> 
> Maybe this option could/should be removed ?

If/when a user decides that the OWM forecast screen does not fit their needs, than they can opt to have the OWM module to create 'forecast data' devices which can be used to automate things (which can not be done just using the forecast screen).

Example: 
- automate the garden watering using the hour data.
- automate opening/closing windows/door/rooflids before the actual rain (so a rain detection sensor does not fit)
- automate climate when away using daily forecast data
- probably much more we didn't think of yet 😁

Personally I do not think that the amount of devices created (only when the users opts to do so) is a problem for any Domoticz installation. Or is it?

gizmocuz:
Every new device confuses users and takes up database space.
Sure hourly forecast if it rains is nice, this is also in the buienradar script (as a switch), maybe it is possible to look at this implementation.
Automatic climate... this can be done in real-time ..

kiddigital:
Maybe we can add an option for the user to enter the amount of forecast days (now 7) or forecast hours (now 48) they want devices for? So if you are only interested in tomorrow or next hour, only a few devices are created.

Maybe something like this?
owm-settings-new

kiddigital:
@gizmocuz , @rwaaren , @jmleglise 
Is the above suggested direction an acceptable compromise? Is clear (hopefully, text suggestions welcome) on what the different options are and what to expect from them. User is in control how many hours, days of forecast information they want devices for to make it fit their use-case.

jmleglise:
My opinion :
- If you keep this way to create devices : it is enough to put a clear message for the user to decide for himself :
MSG = "This will create 8 new devices (UV, Temp, Wind, Cloud, Precipitation, description, icon, name) for each Forcast. So 64 for daily forecasting and  380 for hourly forecasting."
 
Personally, I wouldn't use that for myself, even, if  I find it very useful.

- I would have used it, only if it creates only 5 devices for all : UV, Temp, Wind, Cloud, Precipition, with 47 data serialized inside each. As I said, it is easy, to split it in Lua or python.   (of course, not in blockly. But Would we really make an irrigation system in blockly ?)

kiddigital:
> 
> 
> My opinion :
> 
>     * If you keep this way to create devices : it is enough to put a clear message for the user to decide for himself :
>       MSG = "This will create 8 new devices (UV, Temp, Wind, Cloud, Precipitation, description, icon, name) for each Forcast. So 64 for daily forecasting and  380 for hourly forecasting."
> 

Good clear message... I like that.

> 
> Personally, I wouldn't use that for myself, even, if I find it very useful.
> 
>     * I would have used it, only if it creates only 5 devices for all : UV, Temp, Wind, Cloud, Precipition, with 47 data serialized inside each. As I said, it is easy, to split it in Lua or python.   (of course, not in blockly. But Would we really make an irrigation system in blockly ?)

Using a sensor to basically be a container for an array of values is not something I would be in favor of. The sensor does not behave as a normal sensor and clutters the view of what sensors are for. Optionally one could think to store these 'arrays of values' in a User variable for example and use that Uservariable in your scripting.

(We could make all these things configuration options but it might also make a 'complicated' settings screen. Maybe add an 'advanced settings' button?)

gizmocuz:
I would just use a link to the forecast page on the openweather website, don't make it complicate !
You could use a text sensor, but the forecast link is the simplest.
And to be honest, the easiest way to find out the forecast is still using the weather widget/application on your mobile/pad

kiddigital:
See PR #4698 which is an extended alternative showing a clear message and offers additional switches in the settings.

kiddigital:
@jmleglise Hopefully you are happy with the changes (see other PR). Thanks for raising your concern.

(Feel free to close this PR and/or add suggestions)

jmleglise:
Thanks @kiddigital . I close now this PR. Resolved by #4698 .

PR #4639, opened 2021-02-14
Provide Configuration and LogLevel as Parameter dict

pipiche38:
2 new fields have been recently added to the Plugin Hardware, they are not yet available in the python plugin framework

This PR allow the python plugin framework to get the value of the Configuration and LogLevel parameter

### DIFF
--- a/hardware/plugins/Plugins.cpp
+++ b/hardware/plugins/Plugins.cpp
@@ -1265,7 +1265,7 @@ namespace Plugins
 			m_sql.GetPreferencesVar("Language", sLanguage);
 
 			std::vector<std::vector<std::string>> result;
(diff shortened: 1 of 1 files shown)


gizmocuz:
Would it also be possible to 'set' the configuration parameter ? This way python plugins can store all settings in this field (as JSON for example)
The Mode fields are numeric fields and are currently misused, meaning that once we switch to a stricter database, all python plugins stop working

pipiche38:
From my knowledge the 'Parameters' as Mode1, Mode2 .... are only available in Read Only inside the python Plugin.

For instance if you take those parameters ... they are set on Domoticz side when editing the Hardware .

While I like the idea I think I saw few steps to be addressed:

1/ We need to add 1 more API to the python Framework to all Editing the Parameters
2/ We need to provide a compatibility mode ( in such where the infos are store doesn't matter for the plugins )

CC: @dnpwwo

pipiche38:
@gizmocuz do you want me to correct anything in this PR, before merging it ?

gizmocuz:
@pipiche38 , well it would be great if we could  'set' the configuration parameter as well
So then it's complete

Issue #4643, opened 2021-02-16
Timplan bug on new install 


Eucliwood090:
Hello, I would like to use the Timplan, I tried on my domoticz which I use every day but I have errors.

I then tried on a new stable installation on an ubuntu 20.0 distribution, I got an error.

`
2021-02-16 11:10:31.298 Error: SQL Query("SELECT Active, DeviceRowID, Time, Type, Cmd, Level, Days, UseRandomness, Hue, [Date], MDay, Month, Occurence, Color FROM Timers WHERE (TimerPlan==2) ORDER BY ID") : no such column: Hue
2021-02-16 11:10:31.298 Error: SQL Query("SELECT Active, SceneRowID, Time, Type, Cmd, Level, Days, UseRandomness, Hue, [Date], Month, MDay, Occurence FROM SceneTimers WHERE (TimerPlan==2) ORDER BY ID") : no such column: Hue
`

these errors are on a new installation, they appear when I duplicate a timeplan. but domoticz crash completely at 00h00

So I can't use timplan without a complete domoticz crash

Thanks

rwaaren:
What is your domoticz version number on your every day domoticz?
What command do you give when you see this error?

There is indeed a bug in the timeplan code (referring to none existing table Hue) and I will create a PR to correct it but the bug should only cause issues when duplicating a timeplan.

Eucliwood090:
Hello, I am sending you a little video, it's easier. 
https://youtu.be/zQ1y2-cWID0

rwaaren:
The PR I just prepared (and which wil be available in a next build) will solve the SQL errors with duplicating a timerplan but you will have to refresh your browser cache / and probably browser history to fix the GUI issue.

Eucliwood090:
Perfect thank you . You are doing a great job

PR #4645, opened 2021-02-16
Fix for #4643 Timeplan duplication copies none existing Hue field

rwaaren:
Hue field was transferred to Color in database version 124 but left
behind in Scheduler.cpp code.
Because field stil exists in databases where it was created once, the
issue only surface on installations with new or recently created
databases.


gizmocuz:
Thank you !!

Issue #4644, opened 2021-02-16
Commit e5cb7d3f breaks building on Debian 9


Locusi:
Hello!

I get following errors after trying to build domoticz on Debian GNU/Linux 9.13 (stretch):

```
[ 23%] Building CXX object CMakeFiles/domoticz.dir/main/mainworker.cpp.o
/.../Domoticz/dev-domoticz/main/mainworker.cpp: In instantiation of ‘MainWorker::AddDomoticzHardware(CDomoticzHardwareBase*)::<lambda(auto:1, auto:2, auto:3, auto:4, auto:5)> [with auto:1 = CDomoticzHardwareBase*; auto:2 = const unsigned char*; auto:3 = const char*; auto:4 = int; auto:5 = const char*]’:
/.../Domoticz/boost_1_75_0/boost/function/function_template.hpp:158:11:   required from ‘static void boost::detail::function::void_function_obj_invoker5<FunctionObj, R, T0, T1, T2, T3, T4>::invoke(boost::detail::function::function_buffer&, T0, T1, T2, T3, T4) [with FunctionObj = MainWorker::AddDomoticzHardware(CDomoticzHardwareBase*)::<lambda(auto:1, auto:2, auto:3, auto:4, auto:5)>; R = void; T0 = CDomoticzHardwareBase*; T1 = const unsigned char*; T2 = const char*; T3 = int; T4 = const char*]’
/.../Domoticz/boost_1_75_0/boost/function/function_template.hpp:940:38:   required from ‘void boost::function5<R, T1, T2, T3, T4, T5>::assign_to(Functor) [with Functor = MainWorker::AddDomoticzHardware(CDomoticzHardwareBase*)::<lambda(auto:1, auto:2, auto:3, auto:4, auto:5)>; R = void; T0 = CDomoticzHardwareBase*; T1 = const unsigned char*; T2 = const char*; T3 = int; T4 = const char*]’
[...]

PR #4646, opened 2021-02-17
std::bind, lambda, and GCC6 fixes

neheb:
GCC6 has a bug with lambdas and captured pointers. Work around it by being explicit.

### DIFF
--- a/hardware/MQTT.cpp
+++ b/hardware/MQTT.cpp
@@ -153,8 +153,8 @@ void MQTT::on_connect(int rc)
 			_log.Log(LOG_STATUS, "MQTT: connected to: %s:%d", m_szIPAddress.c_str(), m_usIPPort);
 			m_IsConnected = true;
 			sOnConnected(this);
-			m_sDeviceReceivedConnection = m_mainworker.sOnDeviceReceived.connect([this](auto id, auto idx, const auto &name, auto cmd) { SendDeviceInfo(id, idx, name, cmd); });
-			m_sSwitchSceneConnection = m_mainworker.sOnSwitchScene.connect([this](auto scene, const auto &name) { SendSceneInfo(scene, name); });
+			m_sDeviceReceivedConnection = m_mainworker.sOnDeviceReceived.connect([this](auto id, auto idx, auto &&name, auto cmd) { SendDeviceInfo(id, idx, name, cmd); });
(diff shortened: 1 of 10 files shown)


neheb:
Fixes: https://github.com/domoticz/domoticz/issues/4644

gizmocuz:
@neheb, thank you for this PR. It looked better previously without the 'this' pointer of course...

Question, there are a few 'this' pointers with classes in them:

void ssl_server::init_connection() {
	new_connection_.reset(new connection(io_service_, connection_manager_, request_handler_, timeout_, context_));
	// the following line gets the passphrase for protected private server keys
	context_.set_password_callback([this](auto &&...) { return this->ssl_server::get_passphrase(); });

Could/Should this be 

	context_.set_password_callback([this](auto &&...) { return this->get_passphrase(); }); 

without the 'ssl_server' ?

neheb:
Done.

Although that was the only example I found.

gizmocuz:
Thanks! One thing I find strange (and should not be needed) is using 'this' in a class/function.
Sure this is valid, but should not be needed, it is already 'this'
I think you added this because of another issue reported, but it seems that is solved by installing clang 6, so I don't think this is needed

neheb:
It is as long as CMake does not error on GCC6,

neheb:
Eh removed the GCC6 stuff. It's old at this point.

gizmocuz:
Thank you for this ! One request (when you want to, otherwise I will do this after the merge),
Specially in the WebServer.cpp class

Use the real variable names:

RegisterCommandCode("zwaveisnodeincluded", [this](auto &&PH1, auto &&PH2, auto &&PH3) { Cmd_ZWaveIsNodeIncluded(PH1, PH2, PH3); });

to:

RegisterCommandCode("zwaveisnodeincluded", [this](auto &&session, auto &&req, auto &&root) { Cmd_ZWaveIsNodeIncluded(session, req, root); });

It will make it a bit more readable as well

neheb:
Main reason I didn't do that is that it's time consuming. The other is that it would force clang-tidy to break it up to multiple lines. It's better to merge now I think.

gizmocuz:
Okey I will merged this but will do the time consuming job myself
Clang tidy could be modified to not break it

gizmocuz:
Thank you for all the work again!!

PR #4649, opened 2021-02-17
Add API/JSON to delete shortlog and or calendarlog on a per idx base

rwaaren:
detailed history for some devices might be interresting while for other devices it will never be looked at
this API call provides the option to selectively delete the shortlog and/or calendarlog of a device without the need
to use external tools.
Like it is already available for switchtype devices and text sensors.

/json.htm?idx=<idx>&type=command&param=deletehistory&historytype=calenderlog
/json.htm?idx=<idx>&type=command&param=deletehistory&historytype=shortlog

### DIFF
--- a/main/WebServer.cpp
+++ b/main/WebServer.cpp
@@ -496,6 +496,8 @@ namespace http {
 			RegisterCommandCode("downloadready", std::bind(&CWebServer::Cmd_DownloadReady, this, _1, _2, _3));
 			RegisterCommandCode("update_application", std::bind(&CWebServer::Cmd_UpdateApplication, this, _1, _2, _3));
 			RegisterCommandCode("deletedatapoint", std::bind(&CWebServer::Cmd_DeleteDatePoint, this, _1, _2, _3));
+			RegisterCommandCode("deletehistory", std::bind(&CWebServer::Cmd_DeleteHistory, this, _1, _2, _3));
+
 			RegisterCommandCode("customevent", std::bind(&CWebServer::Cmd_CustomEvent, this, _1, _2, _3));
 
 			RegisterCommandCode("setactivetimerplan", std::bind(&CWebServer::Cmd_SetActiveTimerPlan, this, _1, _2, _3));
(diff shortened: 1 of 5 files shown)


main/WebServer.cpp
```diff
@@ -496,6 +496,8 @@ namespace http {
 			RegisterCommandCode("downloadready", std::bind(&CWebServer::Cmd_DownloadReady, this, _1, _2, _3));
 			RegisterCommandCode("update_application", std::bind(&CWebServer::Cmd_UpdateApplication, this, _1, _2, _3));
 			RegisterCommandCode("deletedatapoint", std::bind(&CWebServer::Cmd_DeleteDatePoint, this, _1, _2, _3));
+			RegisterCommandCode("deletehistory", std::bind(&CWebServer::Cmd_DeleteHistory, this, _1, _2, _3));
```
neheb, line 499: I would say avoid std::bind here.

rwaaren: @neheb 
Thx for the review. Please understand I am not fluent in c++ 
I copied an existing function with similar functionality and only really changed the SQL's inside. 
So much obliged if you could elaborate a bit on why this std::bind is different then the ones already used and/or on the reason why this is bad coding practice. 

Always eager to learn!

neheb: meh nvm. it's fine for this PR.

gizmocuz:
There is already a function doing this. This is duplicate code that now needs to be modified on two locations when we add a new log table
Best to use one function to perform this task (in the original function that does this job call the new function)

Still it is very dangerous...

rwaaren:
> There is already a function doing this. This is duplicate code that now needs to be modified on two locations when we add a new log table
> Best to use one function to perform this task (in the original function that does this job call the new function)
> 
> Still it is very dangerous...

@gizmocuz 
Sorry but I am not aware of an existing function for this task.
I only found DeleteDataPoint but considered it as too different. 
Are you referring to that one or is there another?

Thx!

rwaaren:
Looking at DeleteDatapoint again, I might be able to create a PR to implement your proposed solution. Will have a look today/this weekend

gizmocuz:
@rwaaren , just take the 'delete' calls for the logs from 'SQLHelper.cpp::DeleteDevices' and put them in a new function

void CSQLHelper::DeleteHistory(const std::string& idx, const bool bDeleteShortlog, const bool bDeleteDayLog)

Try to keep the internal common used variable names (idx) and please don't use const char* but const std::string& instead

gizmocuz:
And still it's dangerous and I'm not sure plugins/scripts should be able to do this

rwaaren:
Understand. Closing this PR and will prepare a new one.

gizmocuz:
Before you do, why is this needed?

PR #4656, opened 2021-02-20
Dockerized builder tools

markruys:
This builder still a bit work on progress. It uses Docker to compile Open ZWave and Domoticz. To speed up the build process on macos and Windows it uses a persistent volume for the actual compilation. Currently it supports only builds to amd64 targets. Using `docker buildx` it should be possible to add architectures like Raspberry Pi.


gizmocuz:
@markruys , thank you for your PR.
I am not sure if this is needed as the build is already happening externally (by github, appveyor, arm build system (32/64))

If your project follows this guide: https://www.domoticz.com/wiki/Build_Domoticz_from_source
It should/could work also for 32 bit arm and 64bit Linux

I do have some suggestions:

Please change " markr/domoticz-build" to " domoticz/domoticz-build"

Instead of using docker commands directly, could you use a docker-compose file, it's also easier to copy / duplicate and remove

Please change "Open ZWave" to "OpenZWave"

Some people would like to use special options for the CMake command building Domoticz (using external libraries for example, see CMakeList.txt).
Could you add a environment variable for additional build arguments. I think this is also needed for the apt-get install command (install jsoncpp, minizip, mqtt, sqlite, python, lua ...)

Would be great if you could use a command to replace the decimals to underscore for

ENV BOOST_VERSION=1.75.0
ENV BOOST_VERSION_=1_75_0

This way only 1 environment variable is needed

I would trim down all those run commands to a few run commands that belong together, see

https://github.com/domoticz/domoticz-docker/blob/main/Dockerfile

Because it creates additional 'layers' inside the docker, see

https://docs.docker.com/develop/develop-images/dockerfile_best-practices

For example one for the packages (with the cleanup see previous example file), one for boost, cmake (or combine them as it's still a additional package)

Not sure if we need to build OpenZWave and Domoticz separately, I think a single build command is okey ?

Add a option to build the beta or release

Maybe we could use use some commands from the makebeta/makerelease scripts to output a build file/archive with os and version number included ?

I know it is a lot of requests, but then again it could be a good idea to make these changes/additions

markruys:
Actually I don't know either if this PR should make it into the development tree. But I'm using it and thought it could help others. Anyway, I'll go through your appreciated feedback.

markruys:
I think I've implemented all of your suggestions, except:

>  Add a option to build the beta or release 

This is not clear to me as it seems that a beta release is just the same as a production release, except that a different Git repository is being used. 

To clarify my goal of writing these scripts is to cross compile Domoticz/OpenZWave on my workstation for my Debian Domoticz server. Second goal was to have a fast development environment for writing & testing Domoticz/OpenZWave enhancements. If others find this build tools useful it would be a good idea to support other architectures before merging this PR.  Not sure if I can find time to do this.

gizmocuz:
@markruys , there is a big difference between the development branch and the stable branch (like 100.000 changes)
I could image someone only wants to build the stable release (this means checking out the stable branch instead of the default branch (which is the development branch)

gizmocuz:
Thank you for the changes, much appreciated !
I do have one additional request, instead of using the /root folder, maybe the /opt folder ? Like:

/opt/open-zwave-read-only
/opt/dev-domoticz

(This will then also use the same folder structure as on our build wiki)

markruys:
I've moved the build script logic to the docker-compose.yaml and the image, more the Docker way so to say.

Also I renamed `/root/open-zwave` to `/opt/open-zwave-read-only` and `/root/domoticz` to `/opt/dev-domoticz`, although I don't really like those names ;)

kiddigital:
> 
> 
> @markruys , thank you for your PR.
> I am not sure if this is needed as the build is already happening externally (by github, appveyor, arm build system (32/64))
> 
> If your project follows this guide: https://www.domoticz.com/wiki/Build_Domoticz_from_source
> It should/could work also for 32 bit arm and 64bit Linux
> 

@gizmocuz , @markruys , I think this PR makes sense although it might not be strictly 100% needed as there are already multiple ways to reach the same goal -> setting up a development enviroment.

But as most developers, and the goal is to make it as easy and attractieve as possible for new contributers, start with checking out the code, this PR makes it real easy to start developing. As with this PR straight after the checkout, you can start building using Docker. And I think (assume) that most developers nowadays already have a Dockerized environment.

Ofcourse for those without Docker (like me, I use the [Vagrant way](https://www.domoticz.com/wiki/Build_Domoticz_from_source_with_Vagrant)) there are many alternatives and good documentation.

But with this PR and maybe some small documentation updates (@waltervl maybe?) to the Wiki (refer to Github instead of the other way around), [INSTALL.md](https://github.com/domoticz/domoticz/blob/development/INSTALL.md) and maybe a few others, it makes life a little easier (and also some pieces of Documentation closer to the source where they belong :) )

gizmocuz:
Okey I am a bit more confident with docker-compose now, and it could indeed make sense for some users to build via docker.
One change request @waltervl , the openzwave code needs to be checked out from this repo

https://github.com/domoticz/open-zwave.git

could you change this ?

It might also be a good idea to place this underneath build/docker ?

But maybe we should create a own repository for this ? domoticz-build-docker ?
And I can buildx this to dockerhub as well

waltervl:
Change for me? I think for @markruys

gizmocuz:
Yeah sorry ;) it was morning ;)

gizmocuz:
@markruys , thank you for the change, I only think something did not went well

Before:
$ git clone https://github.com/OpenZWave/open-zwave
$ git clone https://github.com/domoticz/domoticz.git

After:
$ ls domoticz/build/docker-compose.yml
domoticz/build/docker-compose.yml
$ git clone https://github.com/domoticz/open-zwave.git

I think it should be
$ git clone https://github.com/domoticz/open-zwave.git
$ git clone https://github.com/domoticz/domoticz.git

markruys:
I was thinking, when you read this README.md, you already have checked out the https://github.com/domoticz/domoticz.git repository. But if you would like me to change it into what you propose, that's absolutely fine with me.

gizmocuz:
Yes you are right ;) Unless we put this in a separate repository.. but not for now maybe

But this line could or should be removed  ?

$ ls domoticz/build/docker-compose.yml

gizmocuz:
The only issue I have with this PR is that it is using the 'build' folder... that suggests all build scripts are inside this folder.
Currently the CMake project file is in the root, the Visual Studio project file is inside msbuild

Maybe it is worth mentioning inside the build/README.MD that Domoticz can be build in three ways:

- Command line (CMake), and point to the wiki build page
- Windows (Visual Studio), point to the project file in the msbuild folder
- Docker (continue reading here)

Would that make sense ?

gizmocuz:
Thanks!

gizmocuz:
Sorry had to revert this PR, it created invalid dockers with executables that are not able to run

PR #4663, opened 2021-02-21
Query the OT Bridge parameters

brucemiranda:
The latest version of the Evohome firmware has stopped the controller querying the OT parameters from the OT Bridge. So now none of the OT devices get updated. This PR returns that functionality by making the Gateway request the OT parameters instead. Anyone with a previous firmware should stay unaffected, but the devices will get updated twice as often as the Controller and the Gateway will both request the same information.

### DIFF
--- a/hardware/EvohomeBase.cpp
+++ b/hardware/EvohomeBase.cpp
@@ -197,6 +197,20 @@ void CEvohomeBase::SetGatewayID(unsigned int nID)
 }
 
 
+unsigned int CEvohomeBase::GetOpenThermBridgeID()
+{
+	std::lock_guard<std::mutex> l(m_mtxOpenThermBridgeID);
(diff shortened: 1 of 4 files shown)


gizmocuz:
@brucemiranda , is it possible to check the firmware version and not update the devices twice ?

brucemiranda:
> @brucemiranda , is it possible to check the firmware version and not update the devices twice ?

It can be done but Resideo force a software update and have already started the rollout. So sooner or later all users will only use this new functionality. It's only in the interim period where any sensors set up will get updated twice as often. But its definitely better than not at all.

gizmocuz:
Thank you !!

brucemiranda:
Now fixed in PR https://github.com/domoticz/domoticz/pull/4753

PR #4665, opened 2021-02-22
Increase buffer size for HEOS to fix failure in case of more than 6 players

MarcBeige:
Fix the problem that if you have more than 6 players the HEOS system gives a parse error and fails to set or update the player devices. The message that contains the list of players in the system contains about 150 characters per player and the buffer size is 1028, so if you have more than about 6 players the problem occurs. Increasing it to 16384 defers the problem to a system size of more than about 100 players per network.

### DIFF
--- a/hardware/HEOS.h
+++ b/hardware/HEOS.h
@@ -62,7 +62,7 @@ class CHEOS : public CDomoticzHardwareBase, ASyncTCP
 
 	std::string m_szIPAddress;
 	unsigned short m_usIPPort;
-	unsigned char m_buffer[1028];
+	unsigned char m_buffer[16384];
 	int m_bufferpos;
 	std::shared_ptr<std::thread> m_thread;
 };


gizmocuz:
This is 16 times bigger... is this needed ?

MarcBeige:
There is only one HEOS device per network, so the cost of increasing the buffer size to allow up to 100 devices is very low. Increasing the buffer to 16k increases the footprint of domoticz by 0.04% and allows up to 100 players in a HEOS system. Of course it can be smaller, but there is little reason to do so. I think the minimum is 5 kB which would allow for about 30 devices. There will be very few people with more than 30 devices, but there might be. Reducing to 30 devices saves 11k or 0.024% of the footprint as compared to a setting for 100 devices. I advise to allow for 100 devices. It would be even better to use a dynamic buffer size but I think it is not worth the trouble as there is only one device per network.
This reminds me of the old days when I had 1 kB of RAM on a ZX81 for an assembler in machine code ;)

gizmocuz:
Thank you for your PR!

PR #4666, opened 2021-02-22
Fix HEOS parsing of list of players

MarcBeige:
Fix broken loop to parse the list of players.


neheb:
Oh I see. Broken by https://github.com/domoticz/domoticz/commit/53f73dbcf0f0686c3555d4dda45c25c519427a82

Bad transformation.

edit: That's strange that it even compiles...

gizmocuz:
Thank you for the fix and PR!

PR #4667, opened 2021-02-23
remove std::fill

neheb:
It's too verbose.

### DIFF
--- a/hardware/KMTronicBase.cpp
+++ b/hardware/KMTronicBase.cpp
@@ -18,7 +18,7 @@ KMTronicBase::KMTronicBase()
 {
 	m_bufferpos = 0;
 	m_TotRelais = 0;
-	std::fill(std::begin(m_bRelaisStatus), std::end(m_bRelaisStatus), false);
+	m_bRelaisStatus = {};
 }
 
 void KMTronicBase::ParseData(const unsigned char *pData, int Len)
--- a/main/WebServer.cpp
+++ b/main/WebServer.cpp
@@ -15135,7 +15135,7 @@ namespace http {
 					if (!result.empty())
 					{
 						std::map<int, int> _directions;
-						int wdirtabletemp[17][8];
+						std::array<std::array<int, 8>, 17> wdirtabletemp = {};
 						std::string szLegendLabels[7];
 						int ii = 0;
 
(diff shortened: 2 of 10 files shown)


hardware/KMTronicBase.cpp
```diff
@@ -18,7 +18,7 @@ KMTronicBase::KMTronicBase()
 {
 	m_bufferpos = 0;
 	m_TotRelais = 0;
-	std::fill(std::begin(m_bRelaisStatus), std::end(m_bRelaisStatus), false);
+	m_bRelaisStatus = {};
```
gizmocuz, line 21: I would prefer m_bRelaisStatus.fill(false) here, this for readability ...

main/WebServer.cpp
```diff
@@ -15147,7 +15147,7 @@ namespace http {
 					if (!result.empty())
 					{
 						std::map<int, int> _directions;
-						int wdirtabletemp[17][8];
+						int wdirtabletemp[17][8] = {};
```
gizmocuz: I would have expected a conversion to std::array from you here *g* ;)

neheb: :).

The reason for the other std::array conversions is because bool[d] = {} does not work. std::array<bool, d> = {} does.

edit: it works here though. But didn't on the other 2d array ¯\_(ツ)_/¯

gizmocuz: I tested it by just declaring it as

`						std::array<std::array<int, 8>, 17> wdirtabletemp = {};`

and removing the for loop to zero the array

Or you could use the fill method like:

`wdirtabletemp .fill({});`

neheb: that's quite verbose. = {} does the same thing. If using fill might as well set to 0...

gizmocuz: Yes, that's why the first declaration is perfect too. I was experimenting here trying to fill it whole array with a number other then zero... failed without creating a for loop... But not needed in this case ;)

neheb:
On second thought, I'd rather leave it as it.

Specifically in the case of multidimensional arrays, Using .fill fills one dimension. {} I believe is guaranteed to fill everything. {} is equivalent to doing memset of 0.

gizmocuz:
But it's not an array now, and also not consistent of all the changes you are making recently.

neheb:
You mean wdirtabletemp? I can change to array.

gizmocuz:
Yes sorry, that was what i meant

neheb:
Done.

gizmocuz:
Thank you again !!

PR #4668, opened 2021-02-23
constexpr conversions of point

neheb:


neheb:
C++17 would allow the whole namspace to be constexpr. Oh well.

gizmocuz:
Thank you !!

Issue #4670, opened 2021-02-23
Replace device keeps the new device


jmichault:
Replacing device keeps the new device, whereas it should keep the old one.
As a result, the device loses its name, and especially the ID is not kept.

waltervl:
I am not sure as the naming in the UI is strange but according manual you should click on edit on the new device, click on replace and select the old device (but replace dialog is saying new device, see screenshot !!). After clicking on OK, the old device is deleted and the logging is taken over by the new device.
![Selection_031](https://user-images.githubusercontent.com/16327072/108894663-da5dc780-7612-11eb-947d-601bf48c53dc.png) So to keep this behavior the text in the Replace dialog needs to be changed from new to old.

But if this PR works as expected and when clicking edit on the old device, replace with new device and the old device is deleted and the logging is taken over by the new device it is more logical.

kiddigital:
Hm... so code and UI are not consistent here...
2 scenario's
1) From the 'old' device -> Replace this old device by the selected new device
2) From the 'new' device -> Let this device replace the selected old device
Depending where the User starts (old or new device) the UI should be different, but the code should/could be the same -> replaceIDX(old, new)

```
		//Will transfer Newest sensor log to OLD sensor,
		//then set the HardwareID/DeviceID/Unit/Name/Type/Subtype/Unit for the OLD sensor to the NEW sensor ID/Type/Subtype/Unit
		//then delete the NEW sensor
		void CWebServer::RType_TransferDevice(WebEmSession & session, const request& req, Json::Value &root)
		{
			std::string sidx = request::findValue(&req, "idx");
			if (sidx.empty())
				return;

			std::string newidx = request::findValue(&req, "newidx");
			if (newidx.empty())
				return;

```
The code comments are clear on the goal. And it receives an old 'idx' an a 'newidx', but then
```
			//Check which device is newer

			time_t now = mytime(nullptr);
			struct tm tm1;
			localtime_r(&now, &tm1);
			struct tm LastUpdateTime_A;
			struct tm LastUpdateTime_B;

			result = m_sql.safe_query(
				"SELECT A.LastUpdate, B.LastUpdate FROM DeviceStatus as A, DeviceStatus as B WHERE (A.ID == '%q') AND (B.ID == '%q')",
				sidx.c_str(), newidx.c_str());
			if (result.empty())
				return;

			std::string sLastUpdate_A = result[0][0];
			std::string sLastUpdate_B = result[0][1];

			time_t timeA, timeB;
[...]

PR #4671, opened 2021-02-23
incorrect time comparaison in RType_TransferDevice

jmichault:
resolve issue #4670


jmichault:
Needs more corrections to be OK : technical datas from new device (type , ID, ...) must be transferred to old device, if we want to keep old ID.

gizmocuz:
We don't want to keep the old ID as mentioned in the Issue, we replace an old device with a new (existing) device

PR #4673, opened 2021-02-23
reply: constexpr conversions

neheb:
Size improvement locally.

From 8317280 to 8313184 locally.

### DIFF
--- a/webserver/reply.cpp
+++ b/webserver/reply.cpp
@@ -13,109 +13,89 @@
 #include "utf.hpp"
 #include <string>
 #include <fstream>
-#include <boost/algorithm/string.hpp>
 
 namespace http {
 namespace server {
 
 namespace status_strings {
 
-const std::string switching_protocols =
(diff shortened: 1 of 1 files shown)


webserver/reply.cpp
```diff
@@ -280,9 +244,11 @@ void reply::add_header(reply *rep, const std::string &name, const std::string &v
 {
 	size_t num = rep->headers.size();
 	if (replace) {
-		for (size_t h = 0; h < num; h++) {
-			if (boost::iequals(rep->headers[h].name, name)) {
-				rep->headers[h].value = value;
+		for (auto &h : rep->headers)
+		{
+			if (h.name == name)
```
gizmocuz: boost::iequals is a case insensitive compare function. You are making it case sensitive which should not be the case.,
I already made a new function in the helper class and will make a quick patch before merging this PR

gizmocuz:
Thank you again for this PR!

PR #4676, opened 2021-02-23
BleBox: constexpr conversions

neheb:

### DIFF
--- a/hardware/BleBox.cpp
+++ b/hardware/BleBox.cpp
@@ -13,27 +13,26 @@
 
 struct STR_DEVICE {
 	int			unit;
-	std::string api_name;
-	std::string name;
+	const char *api_name;
+	const char *name;
 	int			deviceID;
 	uint8_t		subType;
 	int			switchType;
-	std::string api_state;
+	const char *api_state;
 };
 
-#define TOT_DEVICE_TYPES 9
-
-const STR_DEVICE DevicesType[TOT_DEVICE_TYPES] =
-{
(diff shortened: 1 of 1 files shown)


gizmocuz:
code reading wise... I liked the std::string more then char*
Is it not possible to use std::string or maybe 'auto' here ?

neheb:
std::string is not possible in constexpr context as it has a con/destructor. C++17 has std::string_view which does work in constexpr context. I tried using boost::string_view but it doesn't work in constexpr context.

edit: I think it's just defined to the std variant in C++17 mode.

I don't know where you would want to use auto.

```
constexpr std::array<STR_DEVICE, 9> DevicesType{
```

can be changed to

```
constexpr auto DevicesType = std::array<STR_DEVICE, 9>{
```

And in C++17:

```
constexpr auto DevicesType = std::array{
```

because of CTAD.

edit2: actually if all the members are changed to STR_DEVICE {x } , the extra parentheses can be removed.

gizmocuz:
I was aiming at 'auto api_name;' , but no that does not work..... *g*

gizmocuz:
Thank you !

neheb:
Interesting...

LLVM's libcxx supports std::string_view under an experimental header. That's fairly useless unless building on FreeBSD.

fmtlib provides an implementation of string_view. I wasn't successful in getting the CMake stuff right though.

edit: NVM. https://en.cppreference.com/w/cpp/string/char_traits/length is the problem.

PR #4677, opened 2021-02-23
Honeywell: constexpr conversions

neheb:
Size from 8317280 to 8313184


gizmocuz:
Thank you again for this PR, much appreciated as always !

PR #4679, opened 2021-02-24
manual conversion to use std::make_unique

neheb:
Simpler.


gizmocuz:
Thank you again for the work !!

gizmocuz:
@neheb , could you try to fix this, or revert this PR ?

neheb:
Just revert.

gizmocuz:
@neheb , sorry I could not. Because I fixed some compile warning issues yesterday, I was unable to revert.
The only solution was a hard reset/push

The serial issue was inside "AsyncSerial::writeEnd", but there could be others as well.

PR #4680, opened 2021-02-25
RFLinkBase: constexpr conversions

neheb:
Small size improvement


gizmocuz:
Aldo it might have a few bytes left in the output, it does not improve readability, which is way more important than size
Closing this PR

PR #4681, opened 2021-02-25
EventSystem: constexpr conversions

neheb:
Small size reduction.

### DIFF
--- a/main/EventSystem.h
+++ b/main/EventSystem.h
@@ -201,8 +201,9 @@ private:
 	std::string m_lua_Dir;
 	std::string m_szStartTime;
 
-	static const std::string m_szReason[], m_szSecStatus[];
-	static const _tJsonMap JsonMap[];
+	static const std::array<const char *, 7> m_szReason;
+	static const std::array<const char *, 3> m_szSecStatus;
+	static const std::array<_tJsonMap, 52> JsonMap;
 
 	//our thread
 	void Do_Work();
--- a/main/EventSystem.cpp
+++ b/main/EventSystem.cpp
@@ -46,81 +46,79 @@ extern PyObject * PDevice_new(PyTypeObject *type, PyObject *args, PyObject *kwds
 #endif
 
 // Helper table for Blockly and SQL name mapping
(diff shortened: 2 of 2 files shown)


main/EventSystem.h
```diff
@@ -201,8 +201,9 @@ class CEventSystem : public CLuaCommon, StoppableTask, CNotificationObserver
 	std::string m_lua_Dir;
 	std::string m_szStartTime;
 
-	static const std::string m_szReason[], m_szSecStatus[];
-	static const _tJsonMap JsonMap[];
+	static const std::array<const char *, 7> m_szReason;
```
gizmocuz, line 204: If we rename m_szSecStatus to m_szEventsystemSecStatus , we could remove them from the header file.
(Same for the other array's)
The problem now/currently is that you need to specify the size on two locations (header and source file)
Declaring them only in the source file has also the benefit that no header needs to be changed and less source files need to be recompiled after a change
Is it possible to do this in the PR as well ?

neheb: Doesn't look like it. It's being used in main/dzVents.cpp

gizmocuz:
Yes, but you can declare it as

extern m_szEventsystemSecStatus[]

in dzVents ?

neheb:
I'll try that.

gizmocuz:
Do we need to declare the size ? If yes, then we should leave it like it is, otherwise it would be a plus ;)

neheb:
For std::array , yes. C++17 has CTAD which avoids the need for it.

gizmocuz:
Then I am afraid you need to declare the size in main/dzVents.cpp as well.
Then we better leave it like it is now, thanks ! ;)

neheb:
Yeah.

```
main/dzVents.cpp:758:60: error: no member named 'm_szSecStatus' in 'CEventSystem'; did you mean 'm_SecStatus'?
        luaTable.AddString("Security", m_mainworker.m_eventsystem.m_szSecStatus[secStatus]);
```

Has to be a member of CEventSystem

gizmocuz:
Thank you once again !!

PR #4684, opened 2021-02-25
Add option to delete daterange of one ID in history tables

rwaaren:


main/SQLHelper.cpp
```diff
@@ -7823,28 +7838,10 @@ void CSQLHelper::DeleteDataPoint(const char* ID, const std::string& Date)
 		tLastUpdate.tm_min += 2;
```
gizmocuz, line 7838: I think I added this for a specific reason... probably because the date/time also includes seconds

main/WebServer.cpp
```diff
@@ -3481,6 +3481,12 @@ namespace http {
 
 		void CWebServer::Cmd_DeleteDateRange(WebEmSession &session, const request &req, Json::Value &root)
 		{
+			if (session.rights != 2)
+			{
+				_log.Log(LOG_ERROR, "User: %s tried to delete a daterange!", session.username.c_str());
```
gizmocuz: We don't use these log messages, but only issue session.reply_status = reply::forbidden;
(Just for a very few sporadic calls there is a log message, and for the uservariable it is not needed as well)

main/SQLHelper.cpp
rwaaren: I understand.  That's why I did not take it out deletedatapoint. 
For deletedaterange it is not needed.

I added the requirement to have admin rights.

rwaaren: OK took them out

gizmocuz: Is deletedaterange  not the replacement as well for deletedatapoint?

rwaaren: Not really a replacement. 
deletedatapoint is used as pre-processor for deletedaterange when a single datapoint must be removed (by keeping the code that adds the extra seconds to the datapoint time)  and to maintain backward compatibility. 
The SQL functionality is moved to deletedaterange

rwaaren:
@gizmocuz, thx for reviewing. 
Does this PR require more changes before it can be  merged?

gizmocuz:
No, I just wanted to look at the complete source file and just needed some time for this
Thank you for the PR!!

PR #4687, opened 2021-02-26
Properly terminate plugin subinterpreters on exit

moroen:
This should fix "Fatal Python error: PyInterpreterState_Delete: remaining subinterpreters" on exit when using python versions 3.8.x and lower.


gizmocuz:
Thank you for this PR, hope it works!

PR #4688, opened 2021-02-26
add keep button to replace dialog, and code to manage it.

### LINKED ISSUE #4670: Replace device keeps the new device

jmichault:
resolves #4670


www/index.html
```diff
@@ -418,8 +418,13 @@
 			bootbox.alert($.t('No New Device Selected!'));
 			return ;
 		}
+		var KeepOld="n";
```
gizmocuz: better to use true/false here

gizmocuz:
You always go from a old device to a new device and not the other way around.
Why would you transfer from a new device to a old device ? And why would you want to keep the old device ? (as it has been replaced)

jmichault:
> You always go from a old device to a new device and not the other way around.
> Why would you transfer from a new device to a old device ? And why would you want to keep the old device ? (as it has been replaced)

Whenever I replace the battery in one of my devices, I have to review all of my blockly events to fix the device, since apparently blockly use the idx.

Maybe I can add two buttons:
   * one to copy name, description, used a favorite to the new device.
   * one to also keep the idx.

gizmocuz:
Maybe we should move all data from the new device into the old device and change the idx of the new device to the old device and remove the new device ?
I don't think it is bad to keep the old IDX as default

jmichault:
I have changed y/n to true/false.
I have changed default for keep button to checked.
While doing some tests, I realized that the CSQLHelper :: TransferDevice function was only designed to transfer logs from the old idx to the new one, so I modified it so that it can be used in both directions. 

So with my last commit:
* by default (keep button checked): idx, name, description and the two flags favorite and used are kept. All other fields are copied from the new device. Logs from new device are transferred to the old device. New device is deleted.
* if keep button is unchecked, everything happens as in the current version: logs from old device art transferred to the new device. Old device is deleted.

gizmocuz:
@jmichault , I have checked the changes again but....

You are using 'Id', the field is ID with capital letter

But the change is simpler if we want to keep the old logs and everything

The only thing that needs to be copied into the old device are the following fields (from the new device)

DeviceID, Unit, Type, SubType

After this, move the logs from the new device to the old device (change the IDX in the logs from 'new_idx' to 'old_idx')

Then delete the new_idx (DeviceStatus) record

For example, old device is A and new device = B

Pseudo code:

UPDATE DeviceStatus SET A.DeviceID=B.DeviceID, A.Unit=B.Unit, A.Type=B.Type, A.SubType=B.SubType FROM DeviceStatus as A, DeviceStatus as B WHERE (A.ID = OldIDX) AND (B.ID = NewIDX)

Next change all log records that have DeviceRowID=B.ID to A.ID

for example

UPDATE Temperature SET DeviceRowID=OldIDX WHERE (DeviceRowID = NewIDX)

and last but not least

DELETE FROM DeviceStatus WHERE (ID = NewIDX)

main/SQLHelper.cpp
```diff
@@ -7529,6 +7529,30 @@ void CSQLHelper::DeleteScenes(const std::string& idx)
...
 
+void CSQLHelper::CopyNewDevice(const std::string& newidx, const std::string& idx)
+{
+	// This function copy fields describing hardware from the new device to the old.
+	safe_query("UPDATE DeviceStatus set (DeviceID, Unit, Type, SubType) = (select DeviceID, Unit, Type, SubType from DeviceStatus d2 where d2.ID = '%q') where ID='%q'" , newidx.c_str(), idx.c_str());
+}
+
+// tranfer logs from idx to newidx for table "table"
+void CSQLHelper::Transfer1Log(const char* table,const std::string& idx, const std::string& newidx)
+{
```
gizmocuz, line 7706: I'm not quite sure what happens here...
Not sure if deleting log records from the new device is wanted.
(My new device could be running for some days before I want to exchange it)
But sure if we transfer from A to B, log records for B should not overlap A, so that is correct to delete them (B)

After this, the only thing needed to be done is set the DeviceRowID from B to A and update the DeviceStatus (CopyNewDevice)

May I suggest to rename 'CopyNewDevice' to 'TransferDevice' and as parameters 'old_idx' and 'new_idx" to make it clearer
And rename 'Transfer1Log' to 'TransferLog' (without the 1) and also 'old_idx' and 'new_idx"

Issue #4691, opened 2021-02-27
An integer overflow when parsing Samsung TV IR codes.


derrij:
https://github.com/domoticz/domoticz/blob/aad65002eea99b8336cefd773e7422f0a7f1873f/hardware/MySensorsBase.cpp#L2081

derrij:
Changed it to `pChild->SetValue(vType, (int)strtoul(payload.c_str(), NULL, 10));` , works fine for me

gizmocuz:
@derrij , do you know what the content/value of the 'payload' string is ?

derrij:
payload was '3772833823'

gizmocuz:
Thanks!

It seems you seem to convert it to a 64bit value, and after that truncating it to a 32bit value **(int)**
Could you also test with

pChild->SetValue(vType, (int)std::stoul(payload.c_str()));

If that works, could you make a PR against the active branch (development)

derrij:
Yes, with `pChild->SetValue(vType, (int)std::stoul(payload.c_str()));` works fine.

`2021-02-28 11:32:32.710 MySensors: NodeID: 3, ChildID: 7, MessageType: 1, Ack: 0, SubType: 33, Payload: 3772833823`
`2021-02-28 11:32:32.738 (usb) Light/Switch (IR Command Volume UP TV)`

gizmocuz:
Thank you for the PR, but you have to make the changes in the active branch , which is the development branch

rwaaren:
@derrij Can you please close the issue when tested and OK ?

PR #4696, opened 2021-02-28
Fix for integer overflow when parsing Samsung TV IR codes.

derrij:
line https://github.com/domoticz/domoticz/blob/development/hardware/MySensorsBase.cpp#L2058
change to `pChild->SetValue(vType, (int)std::stoul(payload.c_str()));`

#4691


rwaaren:
> #4691

this PR should not be made against the (protected) master but against the active branch (**development**)

gizmocuz:
Wrong branch

PR #4698, opened 2021-02-28
OpenWeatherMap: updated hardware settings UI incl. warning text

kiddigital:
This PR is an extended alternative to PR #4638

It explains/warns the user on the creation of many devices if the user decides they want forecast devices.

Also now it has a clear setting to use the OWM module for the Domoticz forecast screen (instead of the default DarkSky info).

See the updated settings screen below:
owm-settings

### DIFF
--- a/hardware/OpenWeatherMap.cpp
+++ b/hardware/OpenWeatherMap.cpp
@@ -51,12 +51,14 @@ std::string ReadFile(std::string filename)
 #define OWM_icon_URL "https://openweathermap.org/img/wn/"	// for example 10d@4x.png
 #define OWM_forecast_URL "https://openweathermap.org/city/"
 
-COpenWeatherMap::COpenWeatherMap(const int ID, const std::string &APIKey, const std::string &Location, const int adddayforecast, const int addhourforecast) :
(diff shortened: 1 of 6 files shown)


gizmocuz:
Thank you !!

RicardP:
@kiddigital I wonder if your forecast data could be presented as "forecast graphs", like historical curve? Maybe a nice alternative would be showing actual time as a line in center of the curve graph, i.e. temp, history building up left of actual time line (as well predicted temp kept with historical measured temp to see predicted vs. real temp outcome curves).  And on the right side of "actual time line" showing forecast temp curve? :)

kiddigital:
@RicardP , the data is there in several devices so you probably could build upon that.

But if I understand it correctly, you are looking for a 'dataset' which combines historical measurements (collected by Domoticz sensors) combined with forecast data? (Which are also historically collected when putting the values into sensors). At the moment, there is not such a 'collected/combined' dataset function.

Maybe 'easiest' to do something similar is to push the data to Influxdb and build a Grafana dashboard there?

I do like the idea though 👍

ghost:
@kiddigital I am just a user so I am not able to provide coding or methods, sorry. However I thought that not so many users would need one device of each received forecast data value, however as you say the data is available now so using same provided OWM Forecast data shown like the historical curves would be very nice in addition :D

Issue #4655, opened 2021-02-19
modify updatebeta.sh so it creates a backup of beta or domoticz folder for easy restore


waltervl:
A lot of people just run updatebeta without precautions and run into trouble when the beta has a major failure (shit can happen ;-) ).
As proposed in Forum topic https://domoticz.com/forum/viewtopic.php?f=6&t=34721&p=266798
Modify the updatebeta script to copy the downloaded domoticz_beta.tgz to a save location in case you need to revert to a previous version.

For example:
```
# START OF MOD
# rm domoticz_beta.tgz DEACTIVATED
mv domoticz_beta.tgz ~/domoticz_beta.tgz_$(date +%Y%m%d_%H%M%S)
# keep only the last 2 copies
ls -t ~/domoticz_beta.tgz* | tail -n +3 | xargs -r rm
# END OF MOD
```

Even better solution would be to copy and tar the current domoticz folder to a save location and then update to beta.
eg tar -zcvf ~/domoticz.backup.tar.gz  ./domoticz

waltervl:
If the principle is acceptable I could try to make a pull request for it after testing the scripts on my environment.

gizmocuz:
@waltervl , Not only users run the updatebeta or updaterelease scripts, but also the application itself when you update via the web interface.
You should however 'copy' this file, in the installation folder (so ./ and NOT ~/)
Even better would be to store it in the userdata folder, but that will make it very complex as this is a command line parameter.

But, sometimes there are database patches and you can not revert to a previous version, making things much worse

So best practice is to backup the complete Domoticz folder (or at least the Domoticz executable and the database), or having a second Domoticz installation where you can test the beta, and when OK update your main Domoticz.

the principle/idea is fine of course ! ;)

waltervl:
status update: Script is ready including copy of complete domoticz folder. Manual beta update works. Now I have to check a couple of days to see if it works for update through Domoticz web UI.

gizmocuz:
@waltervl , I would not copy the complete Domoticz folder as this takes 'a lot' of time and we want the update process as quick and smooth as possible.
It is also very easy to do this by hand ?

waltervl:
Live update succeeded for the first time.
The backup process I have configured,  see also the code snippet from the update script:

1. remove the old backup (if existing) to prevent backing up the backup. Name based on installation folder. I do not know if someone decided to install domoticz in another folder than domoticz.
2. process is started from domoticz folder so go 1 step higher
[...]

PR #4702, opened 2021-03-01
Update updatebeta

waltervl:
#4655 As discussed.
Tested with command line and with update through web interface on Linux 20.04 LTS.
Should be compatrible with RPi but please confirm.

### DIFF
--- a/updatebeta
+++ b/updatebeta
@@ -16,19 +16,21 @@ then
 fi
 
 echo "Stopping Domoticz..."
-base=$(basename $PWD)
-timestamp=$(date +%Y%m%d_%H%M%S)
-# remove old backup
-rm -f $base.backup*.tar.gz
-cd ..
-# create backup in parent folder (mostly home)
 sudo service domoticz.sh stop
-echo "Making backup of current installation..."
-sudo tar --exclude $base/backups -czf $base.backup_$timestamp.tar.gz $base
-# move backup to domoticz folder
-mv $base.backup_$timestamp.tar.gz $base/$base.backup_$timestamp.tar.gz
-echo "Backup finished..."
-cd $base
+if [ $1 != "-nobackup" ]; then
+    # create backup in parent folder (mostly home)
+    base=$(basename $PWD)
+    timestamp=$(date +%Y%m%d_%H%M%S)
+    # remove old backup
+    rm -f $base.backup*.tar.gz
+    cd ..
+    echo "Making backup of current installation..."
+    sudo tar --exclude $base/backups -czf $base.backup_$timestamp.tar.gz $base
+    # move backup to domoticz folder
+    mv $base.backup_$timestamp.tar.gz $base/$base.backup_$timestamp.tar.gz
+    cd $base
+    echo "Backup finished..."
+fi
 echo "Downloading latest beta version..."
 wget -4 -O domoticz_beta.tgz --no-check-certificate "https://www.domoticz.com/download.php?channel=beta&type=release&system=${OS}&machine=${MACH}"
 if [ $? -ne 0 ]
--- a/updatebeta
+++ b/updatebeta
@@ -2,6 +2,8 @@
 
 # This script can be used to upgrade to the latest beta version
 
+set -e
+
 lowercase(){
     echo "$1" | sed "y/ABCDEFGHIJKLMNOPQRSTUVWXYZ/abcdefghijklmnopqrstuvwxyz/"
 }
@@ -22,7 +24,7 @@ else
 	exit 1
 fi
 echo "Updating to latest beta version..."
-echo "Stop Domoticz and create backup of current install...
+echo "Stop Domoticz and create backup of current install..."
 base=$(basename $PWD)
 timestamp=$(date +%Y%m%d_%H%M%S)
 # remove old backup
--- a/updatebeta
+++ b/updatebeta
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/env bash
 
 # This script can be used to upgrade to the latest beta version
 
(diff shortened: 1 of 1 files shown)


rwaaren:
When executed it tar's the active database including the Write Ahead Log (wal) file and the SHaredMemory (shm)  file.
There is a (small) chance that these files, even when restored together, will result in a corrupt database.
See [how to corrupt a sqlite database](https://www.sqlite.org/howtocorrupt.html) -->  1.2. Backup or restore while a transaction is active

Safer methods would be to 

do the tar when no process access the database (= after domoticz stopped). 
or
before the tar do a sudo sqlite3 domoticz.db ".backup domoticz.backup" 
or
before the tar do a sudo sqlite3 domoticz.db "vacuum into 'domoticz.backup'"

updatebeta
```diff
@@ -13,17 +13,39 @@ then
...
+cd $base
+echo "Downloading and installing new beta..."
 echo "Please Standby..."
 wget -4 -O domoticz_beta.tgz --no-check-certificate "https://www.domoticz.com/download.php?channel=beta&type=release&system=${OS}&machine=${MACH}"
 echo "Checking file Integrity..."
 tar -tzf domoticz_beta.tgz >/dev/null
 if [ $? -ne 0 ]
 then
  echo "Problem in downloaded Domoticz archive. Stopping update!"
+ sudo service domoticz.sh start
```
gizmocuz, line 43: This will leave you with a correct and partial Domoticz install.
Best to check if the wget command succeeded, if not, stop (and restart Domoticz)
You could check if the tar is correct, if not, stop (and restart Domoticz)
Extract , and restart Domoticz

But as mentioned, this script is also used internally and it will slow down the update proces quite a bit

Could you add a date/time timestamp in the filename as well ?

rwaaren (commented):
Tested on RPI Buster and NUC Debian Bullseye but 
script fails on both systems.


```
You are root.
Updating to latest beta version...
./updatebeta: line 55: syntax error near unexpected token `('
./updatebeta: line 55: `echo "Starting Domoticz... (please standby...)"'

```
Why changing to bash? It is not the same as sh and behaves different.
I assume ficho is a typo ?

waltervl:
I changed it to bash because it is now the same way as the main installation script. Also I had to change it for the root access check (the same as the installation root access check),

rwaaren:
> I changed it to bash because it is now the same way as the main installation script. Also I had to change it for the root access check (the same as the installation root access check),

The main installation script is always executed from the command line . 

The updatebeta and updaterelease scripts can be executed from the command line and from the domoticz GUI. 

When executed from the command line I still get

```
You are root.
Updating to latest beta version...
./updatebeta: line 56: syntax error near unexpected token `('
./updatebeta: line 56: `echo "Starting Domoticz... (please standby...)"'

```

When updating from the GUI, I don't see any error message but I also see nothing happening in the domoticz log so I assume it is the same error.

waltervl:
Sorry for this. First time working with github. 
I copied now the script as is that is on my system. This should work.

rwaaren:
> I copied now the script as is that is on my system. This should work.

Script now tested OK on NUC Debian bullseye  and RPI-4 buster. On slowest system it took 37 seconds in total ( 23 seconds extra compared to current updatebeta) when triggered from the GUI. 

Would it be considered useful to include version build number in backup name?
`build=$(./domoticz --version | grep Giz | awk '{print $5"_"$7}')
`

waltervl:
I do not know if version adds something, it will not hurt to add it.
What perhaps could be an issue is that there is only one backup. What if the user is updating, has issues, updates again the next day, still has issues. 
He cannot return to the original working environment. Is this a common scenario? And if yes, what is the max number of backups to keep?

rwaaren:
> I do not know if version adds something, it will not hurt to add it.
> What perhaps could be an issue is that there is only one backup. What if the user is updating, has issues, updates again the next day, still has issues.
> He cannot return to the original working environment. Is this a common scenario? And if yes, what is the max number of backups to keep?

My 2 cents; with this safeguard domoticz already takes a big part of the responsibility for failed updates up to a level that not many other software distributions do. I would leave it at 1 backup (at most)

waltervl:
So, let's merge it! Or do I have to do something urgent?

gizmocuz:
@waltervl , the modifications are good, but it introduces a huge delay in updating the application
At the present we have some other urgent matters to address, but I am thinking about adding a option in the settings that the user can enable to create backups for each update
So this PR will be used somehow

waltervl:
What is the problem with a huge delay for beta update as it it safe? I would not mind waiting 20 seconds extra.
Who is going to raise an issue? Better safe than sorry......
Why not just merge it (with an announcement on the forum) and see what the community thinks....

gizmocuz:
- I also need to do more testing... (different enviroments)
- test if we need this change "#!/usr/bin/env bash"
- and some prompts I would like to change

As mentioned this PR is not going to get wasted !

gizmocuz:
Question, at the beginning of the script you check if the user is root (or executed via sudo)
In the script you call sudo

I think we should either use sudo in the script, or check if it is executed with sudo rights and remove the sudo's in the script

What if Domoticz is not run as root ? sudo is used for stopping/starting the service
I know you did this for the backup archive creation, maybe use sudo there ?

gizmocuz:
Is it possible to create the backups in a/the 'backup' folder ?

waltervl:
> Question, at the beginning of the script you check if the user is root (or executed via sudo)
> In the script you call sudo
> 
> I think we should either use sudo in the script, or check if it is executed with sudo rights and remove the sudo's in the script
> 
> What if Domoticz is not run as root ? sudo is used for stopping/starting the service
> I know you did this for the backup archive creation, maybe use sudo there ?

Yes you are right, sudo should not be needed in the script itself as we check for that. 

As the standard install script installs Domoticz as root I kept that as default (the root check is taken from install).

waltervl:
> Is it possible to create the backups in a/the 'backup' folder ?

Yes, that should be no problem. But I do not know what happens if there is no backup folder and we create one for the this update and later the user decides to switch on automated database backup. New folders are created and perhaps this creation will stumble on the update backup folder. So I kept away from this.

gizmocuz:
Thank you for this PR! I will continue to tweak it a little bit more

PR #4703, opened 2021-03-01
Harmonize and fix log messages in HEOS.cpp

MarcBeige:
Let all message start with 'HEOS'. Some log messages start with 'DENON by HEOS', others with 'HEOS by DENON'. The first is incorrect because HEOS is a system from Denon, not the other way around, in both cases Denon should be capitalized, not all caps, and it is not good that there are two styles of messages. Proper labels would be 'HEOS by Denon', 'Denon HEOS' or 'HEOS'. The latter is used as it is simpler and the manufacturer also refers to the system as HEOS.


kiddigital:
@MarcBeige , this module still uses an old style of logging using the old logger functions.

If you would remove the '_log.' before each logging/debug statement, it will use the logger in the HardwareBase class that will append the name of the hardware module for you.

So no need anymore to put 'HEOS: ' before each line 😁

MarcBeige:
I have updated HEOS.cpp to the new style of logging.

gizmocuz:
Great, thank you for this!

PR #4706, opened 2021-03-02
Process negative OT data from Evohome

brucemiranda:


gizmocuz:
Thank you for the PR and work !

PR #4709, opened 2021-03-06
replace boost::array with std version

neheb:
The documentation says:

Update: std::array is (as of C++11) part of the C++ standard. The
differences between boost::array and std::array are minimal. If you are
using C++11, you should consider using std::array instead of boost::array.


gizmocuz:
Thank you!!

PR #4710, opened 2021-03-06
replace boost::function with std version

neheb:


gizmocuz:
Thank you for this PR!!

PR #4711, opened 2021-03-06
replace boost::shared_lock with std version

neheb:

### DIFF
--- a/main/EventSystem.cpp
+++ b/main/EventSystem.cpp
@@ -24,6 +24,7 @@
 #include "../main/LuaTable.h"
 #define __STDC_FORMAT_MACROS
 #include <inttypes.h>
+#include <shared_mutex>
 
 extern "C" {
 #include <lua.h>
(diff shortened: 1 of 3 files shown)


main/EventSystem.cpp
```diff
@@ -612,7 +613,7 @@ void CEventSystem::GetCurrentMeasurementStates()
 	m_windgustValuesByID.clear();
 	m_zwaveAlarmValuesByID.clear();
 
-	boost::shared_lock<boost::shared_mutex> devicestatesMutexLock(m_devicestatesMutex);
+	std::shared_lock<boost::shared_mutex> devicestatesMutexLock(m_devicestatesMutex);
```
gizmocuz, line 616: Is it maybe an idea to also use std::shared_mutex ... if it needs to be shared, maybe std::mutex will work as well ?

neheb: std::shared_mutex is C++17

gizmocuz: True, but does it need to be a shared mutex or is a normal mutex sufficient ?

neheb: I don't know. I prefer not to edit the code.

gizmocuz:
I think this is currently been done in

#4715

Needs more testing, but so far no issues detected

main/EventSystem.cpp
keryell: Is there a serious reason for not switching to C++17 anyway?

PR #4715, opened 2021-03-08
Changed some boost shared mutex to std mutexes

gizmocuz:
Needs more testing and work...


gizmocuz:
Thanks for the PR!

I think this is currently been done in

https://github.com/domoticz/domoticz/pull/4715

Needs more testing, but so far no issues detected

rwaaren:
Tested with dzVents integration tests and with some Python plugins (during normal processing, disabling and enabling plugins, stop and start eventsystem, stop and start dzVents.
No errors encountered

PR #4717, opened 2021-03-09
replace boost::unique_lock with std version

neheb:


neheb:
Hmm conflicts with https://github.com/domoticz/domoticz/pull/4715 . nvm.-

gizmocuz:
Yeah, there is a PR from me which should eliminate the boost shared mutexes (it's not finished yet).
It has been tested this far...
The eventsystem class has way to many mutexes in my opinion... One was/is enough (devicemutex), no need for uservariable/scene/... mutex, the speed difference would not be that much

PR #4718, opened 2021-03-09
clang-tidy: run through performance checks

neheb:

### DIFF
--- a/hardware/Limitless.cpp
+++ b/hardware/Limitless.cpp
@@ -167,10 +167,6 @@ CLimitLess::CLimitLess(const int ID, const int LedType, const int BridgeType, co
 	Init();
 }
 
-CLimitLess::~CLimitLess(void)
-{
-}
-
 void CLimitLess::Init()
 {
 	m_CommandCntr = 1;
--- a/main/SQLHelper.cpp
+++ b/main/SQLHelper.cpp
@@ -7809,13 +7809,12 @@ void CSQLHelper::DeleteDateRange(const char *ID, const std::string &fromDate, co
 	if (result.empty())
 		return;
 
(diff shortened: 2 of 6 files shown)


hardware/Limitless.cpp
```diff
@@ -167,10 +167,6 @@ CLimitLess::CLimitLess(const int ID, const int LedType, const int BridgeType, co
 	Init();
 }
 
-CLimitLess::~CLimitLess(void)
-{
```
gizmocuz, line 171: I added this back, each class should have a deconstructor in .cpp file as it's easier for possible additions
I have no problem with a empty one

main/SQLHelper.cpp
```diff
@@ -3332,7 +3332,7 @@ void CSQLHelper::ManageExecuteScriptTimeout(int pid, int timeout, bool *stillRun
 }
 #endif
 
-void CSQLHelper::PerformThreadedAction(const _tTaskItem tItem)
+void CSQLHelper::PerformThreadedAction(const _tTaskItem &tItem)
```
gizmocuz: This is where it crashed last time, please do not undo my changes
Also we do not need to modify the taskItem after the function returns

hardware/Limitless.cpp
neheb: The goal is size reduction.

main/SQLHelper.cpp
neheb: This was automated. I'll look into it.

neheb: Removed. Easier than figuring out why clang-tidy thinks this should be a reference.

gizmocuz:
Thank you!!

PR #4721, opened 2021-03-10
treewide: start using fmtlib

neheb:
fmtlib is faster than C format family of functions. It also allows to
clean up the code here.

Some of these are probably wrong. Just posting here for comments.


gizmocuz:
I think fmt can be used as a header only library correct?
Not sure if we need a external library or include it in the project
There are a few places where we can use this, but these places can also be converted to use stringstream
If it's just a few places we have to think if its worth the change

tesfayeawala:
- [ ] ,lff

gizmocuz:
Seems this library is causing some issues/crashes. Just made a commit for a simple printf replacement (std_format)

PR #4726, opened 2021-03-12
Added double quotes around double quotes in -log file path parameter

laplug:
The current script used one (1) double quote. Innosetup requires double quotes around them to retain them in the path name parameter for path names with spaces in them.


gizmocuz:
Thank you for the PR! Working good !

PR #4729, opened 2021-03-14
JSON Web Token support (RFC 7519)

kiddigital:
This PR adds support for JSON Web Tokens (JWT) as an additional mean to authorize access to Domoticz.

**General**
For more information, see the official RFC at: [https://tools.ietf.org/html/rfc7519](https://tools.ietf.org/html/rfc7519)

It is a first step in future support for OpenID Connect ([https://openid.net/connect/](https://openid.net/connect/)) which uses JWT and OAuth2 as underpinning technologies.

Also for future support of OpenAPI ([https://www.openapis.org/](https://www.openapis.org/)), having token support is mandatory when implementing OpenAPI support in Domoticz.

**How does it work (now)**

When you enable 'Website protection' in the settings, Domoticz will ask you to perform a login before you get access to the UI. Also any other requests to the Domoticz webserver, for example by scripts and/or external dashboards, apps, etc., you either need to provide authorization using Basic Authentication and/or providing a valid session cookie.

Now it is also possible to provide a JWT access token as a 'Bearer' token through the HTTP 'Authorization' header. No need anymore to keep providing a valid session cookie or passing a username/password combination with each request. Just add the JWT Token to each request via the 'Authorization' header.

_Bonus_: When using JWT Bearer tokens, no cookies are generated and have to be exchanged. Also, Domoticz does not need to generate a _UserSession_ to keep track of the cookies which saves a lot of SQL reads AND writes to the UserSession table. Nice bonus to lower the wear on for example SD-cards when using a Pi.

**What does it look like?**
A JWT token is an base64 encoded string consisting of 3 parts. Each of these parts have a function and contain information that can be leveraged. To _see_ what information is in a JWT Token, you can use [https://jwt.io/#debugger-io](https://jwt.io/#debugger-io) to take a look.

Although the information usually is readable, which is fine because it contains your information, it cannot be tampered with as it is signed. When receiving a JWT token, the signature is checked to ensure data is not corrupted.

The _payload_ of each JWT contains relevant information that is used by this PR and looks like the following: 
```
{
  "aud": "DomIssuer1",
  "exp": 1615728854,
  "iat": 1615725254,
  "iss": "Domoticz.com",
  "nbf": 1615725254,
  "sub": "tester"
}
```
Each of the _claims_ in the payload is mandatory and is also validated. 

The 'exp' (expires at), 'nbf' (not before) and 'iat' (issued at) are all Unix Timestamp values about when the token was issued and in what period it is valid.

The _Audience_ (aud) claim contains the _ClientID_ of the **Client** (~application, etc.) for which the token is given out.
The _Subject_ (sub) contains the **Resource Owner**, which is the actual user (you :) )
And the _Issuer_ is the **Authorization Server** that created the token, here it is the Domoticz authorization Realm

**How to retrieve a JWT token?**

At the moment, not a full OpenID Connect (or OAuth2) flow is implemented. A basic flow resembling the _Client Credentials Grant_ type ([https://tools.ietf.org/html/rfc6749#section-4.4](https://tools.ietf.org/html/rfc6749#section-4.4)) is available using the Domoticz normal _command_ structure. A *getjwttoken* command is added which, if called in the right way, returns a JSON response resembling an OAuth2 responds containing a new _access\_token_.

The command looks as followed:
```
https://<YOURDOMOTICZIP>/json.htm?type=command&param=getjwttoken&username=tester
```
The right way for this command is to call it using a POST request and also provide an _Authentication_ header containing a 'username/password' combination using the Basic authentication method of encoding. For example using Curl:
```
curl --request POST 'https://localhost:8443/json.htm?type=command&param=getjwttoken&username=tester' \
--header 'Authorization: Basic RG9tSXNzdWVyMTpEb21Jc3N1ZXI='
```
Here the 'username/password' are NOT the username and password of the user named 'tester', but the ClientID and ClientSecret of the **Client** (DomIssuer1) that wants the token _on behalf off_ the User (tester).

The _username_ parameter provided in the URI needs to be a valid Domoticz user 

and the result:
```
{
    "access_token": "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJhdWQiOiJEb21Jc3N1ZXIxIiwiZXhwIjoxNjE1NzM0NzA2LCJpYXQiOjE2MTU3MzExMDYsImlzcyI6IkRvbW90aWN6LmNvbSIsIm5iZiI6MTYxNTczMTEwNiwic3ViIjoidGVzdGVyIn0.m14_7PgD0yCocbZqLOwSjZDNLTS1eLv8QyGDnCY01YA",
    "expires_in": 3600,
    "status": "OK",
    "title": "getjwttoken",
    "token_type": "Bearer"
}
```

**Setting up a Client**
Before a _client_ can use JWT Tokens on behalf off a user, such a client needs to be setup.

This is done through the 'Users' configuration of Domoticz (See 'Setup/More Options/Edit Users').
To _register_ as client, create a User and give that user the *ClientID* rights, see screenshot

user-overview-clientid

The Username = ClientID and the Password = ClientSecret

**NOTE**:Actually, the generated MD5-based hash Domoticz generates before storing the password, is the ClientSecret that is used to do the actual signing and verification of the JWT's.


kiddigital:
@gizmocuz , this PR includes an external library (https://github.com/Thalhammer/jwt-cpp). Referencing Master at the moment, but that is the same as the v0.5.0 tag, so we could fix it v0.5.0 if we want.

I do not know (yet) how to add this library to the Windows build system (as I am building on Linux), so the Appveyor build fails as it cannot link to this new library.

Let me know what I can do to make it work.

gizmocuz:
@kiddigital , for windows we use VCPKG, and there seems to be a port for JWT ! (Very good!!!)
So we need to add this library... I can do this
About the user rights (client_id), how about we generate a client ID for each user by default ?

kiddigital:
> 
> 
> @kiddigital , for windows we use VCPKG, and there seems to be a port for JWT ! (Very good!!!)
> So we need to add this library... I can do this

That would be great... thx...

Not sure if it becomes an issue, but I noticed that VCPKG has an older version of the library.. opened a 'Question' to see if it can be updated.

> About the user rights (client_id), how about we generate a client ID for each user by default ?

Not sure what you mean here? A 'client' (with its clienid/clientsecret) is _not_ a user. It represents an 'application' (client) that wants to access the _Resource Server_ (Domoticz backend in this case) on behalf of a User.

The (Angular) front-end of Domoticz for example could be considered a 'client' consuming the backend (Resource Server).

Also a plugin, a dashboard like Dashticz or a mobile app like Domotix or Pilot Home Control are 'clients'.

kiddigital:
> 
> 
> @kiddigital , for windows we use VCPKG, and there seems to be a port for JWT ! (Very good!!!)
> So we need to add this library... I can do this

@gizmocuz, can you add the library to the Windows build? 

And there is a PR on VCPKG for the update to the latest release of jwt-cpp :)
see: [https://github.com/microsoft/vcpkg/pull/16799](https://github.com/microsoft/vcpkg/pull/16799)

prince-chrismc:
> The Username = ClientID and the Password = ClientSecret

Typically you do _not_ want to pass the password, the added indirection helps keep the system secure.


https://www.oauth.com/oauth2-servers/client-registration/client-id-secret/
Check this article

kiddigital:
> > The Username = ClientID and the Password = ClientSecret
> 
> 
> 
> Typically you do _not_ want to pass the password, the added indirection helps keep the system secure.
> 
> 
> 
> 
> 
> https://www.oauth.com/oauth2-servers/client-registration/client-id-secret/
> 
> Check this article 
> 
> 

I know 😁 This is a first iteration on route to adopt OpenID connect.. 

thx for the support on adding the latest release of JWT-CPP to VCPKG. Makes our (Windows) life much easier.

prince-chrismc:
> to VCPKG

Shameless plug... Check out Conan!

I have a lot of Home automation and IoT, been using HA for a while but it's nice to see alternatives. Very interesting project!

kiddigital:
@gizmocuz , the latest version of jwt-cpp has been added to the Master branch of VCPKG (thx to @prince-chrismc ) and I added it to the `msbuild/vcpkg-packages.txt` file.

But I think it needs to become part of the `win32-libraries` prebuild package before the automated Windows (appveyor) build could successfully build this PR. Correct? Is that something you can 'quickly' do or could I be of some assistence somehow? I don't have a Windows build environment at the moment :(

kiddigital:
@gizmocuz , when you have a moment, can you add the JWT-CPP package to the windows libraries so the Windows build of this PR can be build succesfully? Thx...

> 
> But I think it needs to become part of the `win32-libraries` prebuild package before the automated Windows (appveyor) build could successfully build this PR. Correct? Is that something you can 'quickly' do or could I be of some assistence somehow? I don't have a Windows build environment at the moment :(

gizmocuz:
Don't worry about the windows build... Going to test this real soon and if all is OK i'll add the library.
It's a header only library ? Why not include it ?

kiddigital:
> 
> 
> Don't worry about the windows build... Going to test this real soon and if all is OK i'll add the library.
> It's a header only library ? Why not include it ?

As far as I know indeed header only lib, but not sure (haven't look into it) how to make that work on Windows :(
Linux builds fine :)

gizmocuz:
I think just by including this header in the domoticz project and using this instead of external projects

gizmocuz:
I think when you create a folder in the domoticz code folder named "jwt-cpp" and place in here the two files (jwt.h and base.h)
then include jwt in the source file that needs it with

#define JWT_DISABLE_PICOJSON
#include "jwt-cpp\\jwt.h"

it should/could work

This way you also do not need changes for the CMake file and no submodule

kiddigital:
@gizmocuz , I think that will work (will give it a try) but that effectively means that we copy the code (headers) into the Domoticz code tree and loose the link with the external repo/library.

Isn't it preferred to keep the link so we can benefit from upstream patches and improvements?

Same goes for the other external modules (like jsoncpp, sqlite, openzwave, etc)?

Linux builds fine with the external lib :)

And as the latest version is now also in VCPKG, building (fast) on Windows should also work, is it not?

gizmocuz:
I'm also fine with using a submodule... it did seem easier for a header only library to include the header

Could you make the following change:

USE_EXTERNAL_JWTCPP --> USE_BUILDIN_JWTCPP

option(USE_EXTERNAL_JWTCPP "Use external JSON Web Token library from Thalhammer" YES)

As you can see in the CMake project all submodules use BUILDIN instead of EXTERNAL
EXTERNAL in this case means a library already installed on the target system, and not static link it
In this case, I don't think JWP_CPP is already found at a clean system (as it's header only)

gizmocuz:
Thanks!

I think when using it external you have to search in CMakeList.txt for

find_package(jwt-cpp CONFIG REQUIRED)

like the other packages

Not sure if you can install libjwt-cpp via apt ?

kiddigital:
Don't think this library is available as a (debian or other) package. Could not find it and also as a header only lib, not as useful as a package I would think.

For compile time, now both VCPKG and Conan support this library, so easy to include it that way.

And now via CMake having it as an external library being included, I think we are good to go :)

gizmocuz:
Okey, then there is no need for a optional CMake option to have it build in or external as it's not external available

kiddigital:
@gizmocuz , rebased this PR to fix merge conflict but now somehow now both builds have failed.

For Windows we know why (just need to include the JWT-CPP VCPKG package into the Win32 binaries)

But the Linux build failed on extracting/downloading Boost... and I cannot restart the pipeline :)
(other than adding some useless commit)

gizmocuz:
Just include the header only library in the project

gizmocuz:
Seems the boost download location has changed. I updated this in the workflow, and now using boost 1.76 as well

kiddigital:
Closed in favor of PR #4824

PR #4735, opened 2021-03-18
Corrected Speaker and Blinds in dashboard view

ridderr:
This change should correct that always the Speaker On image is displayed in the dashboard. Secondly the controller of Blinds in the dashboard should now work correct.

### DIFF
--- a/www/js/domoticzdevices.js
+++ b/www/js/domoticzdevices.js
@@ -1665,10 +1665,11 @@ Smoke.inheritsFrom(BinarySensor);
 function Sound(item) {
 	if (arguments.length != 0) {
 		this.parent.constructor(item);
+		var onoff = ((item.Status == "On") ? "On" : "Off")
 		if (item.CustomImage != 0) {
-			this.image = "images/" + item.Image + "48_On.png";
+			this.image = "images/" + item.Image + "48_" + onoff + ".png";
 		} else {
-			this.image = "images/Speaker48_On.png";
+			this.image = "images/Speaker48_" + onoff + ".png";
 		}
 		this.LogLink = this.onClick = "window.location.href = '#/Devices/" + this.index + "/Log'";
 	}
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thank you for your PR.
As far as I just tested there is nothing wrong with the speaker icon on the dashboard.
The speaker icon is just a icon for a switch. To test, create a dummy switch, edit the switch, change the icon to a speaker, and add it to the dashboard.
Now when you toggle the switch, the icon should toggle as well.

There is a 'sound' device, which is a sensor that measures the sound level in dB, there is no On/Off state, as it is a sensor and not a switch.

I'm also not sure about the blinds issue you are having. There should be two icons a open and a close. There is no status
So your switch type has to be set to type = blinds or blinds inverted

ridderr:
**+++Speaker**
The change in the speaker is because my speaker is always showing the On pictogram while it’s off. It’s a Harmony device. (See pictures). With the proposed change the right pictogram is shown.
DevicesSpeaker
DashboardSpeaker

**+++Blinds**
The change for the blind is because it doesn’t work in my situation where the Blind is a virtual device which controls via RFXCom a Window dimming foil.
When I click on the Close icon the action is “Schakel Aan” (Switch On). 
Also the Percentage is not working because the code checks on “Blinds Percentage” while the SubType is “Blinds Percentage Inverted” and therefor there will be no match.
DevicesBlinds
DashboadBlinds-2

gizmocuz:
How is the dashboard view of the blind ? You show me the floorplan view
Or is your dashboard the floorplan ?
If you use the normal dashboard, is that correct?

ridderr:
Sorry I mixed dashboard with the floorplan. Problems are on the floorplan, the dashboard is working. 
Also the devices are working correct on the Switches page.

gizmocuz:
Ahh than this indeed fixes things, thanks you for this PR!

PR #4737, opened 2021-03-20
remove excess {} in std::array

neheb:
Adding the initializer explicitly avoids having extra {}. Also adds
compatibility with C++17 CTAD.


gizmocuz:
Thank you for the PR. I don't see any difference, except that there is more code and it's harder to read
I don't see a benefit for this patch....?

PR #4743, opened 2021-03-21
Idx plugins backward compat

jkl1337:
Here's a rough proposal for using Device Idx based addressing in plugins with backward and easy forward compatibility (updated plugins can continue to work on old stable). 
Note the use of Type+Subtype for uniqueness rather than just DeviceID + Unit is a trivial modification - I'll fix it.

To use this a plugin calls Domoticz.UseDeviceIndex() at module load time.
After this Devices is now indexed by Idx and creating devices doesn't require a Unit - it changes the Devices constructor. Otherwise the API is similar, so this has minimal impact on existing plugins.
There are alternatives, but the advantage of this way is the plugin can detect the feature at load time and support both indexing methods easily - obviously just not doing the allocation of a Unit number if the Idx feature is available.

### DIFF
--- a/hardware/plugins/DelayedLink.h
+++ b/hardware/plugins/DelayedLink.h
@@ -29,6 +29,8 @@ namespace Plugins {
 #define DECLARE_PYTHON_SYMBOL(type, symbol, params)	typedef type (PYTHON_CALL symbol##_t)(params); symbol##_t  symbol
 #define RESOLVE_PYTHON_SYMBOL(symbol)  symbol = (symbol##_t)RESOLVE_SYMBOL(shared_lib_, #symbol)
 
+#define DECLARE_PYTHON_DATA_SYMBOL(type, symbol) typedef const type *symbol##_t; symbol##_t symbol;
+
 	struct SharedLibraryProxy
 	{
 #ifdef WIN32
@@ -62,6 +64,7 @@ namespace Plugins {
 		DECLARE_PYTHON_SYMBOL(const char*, PyUnicode_AsUTF8, PyObject*);
 		DECLARE_PYTHON_SYMBOL(char*, PyByteArray_AsString, PyObject*);
 		DECLARE_PYTHON_SYMBOL(PyObject*, PyUnicode_FromKindAndData, int COMMA const void* COMMA Py_ssize_t);
+		DECLARE_PYTHON_SYMBOL(PyObject*, PyLong_FromUnsignedLongLong, unsigned long long);
 		DECLARE_PYTHON_SYMBOL(PyObject*, PyLong_FromLong, long);
 		DECLARE_PYTHON_SYMBOL(PY_LONG_LONG, PyLong_AsLongLong, PyObject*);
 		DECLARE_PYTHON_SYMBOL(PyObject*, PyModule_GetDict, PyObject*);
@@ -135,6 +138,8 @@ namespace Plugins {
 		DECLARE_PYTHON_SYMBOL(PyObject*, PyObject_GetIter, PyObject*);
 		DECLARE_PYTHON_SYMBOL(PyObject*, PyIter_Next, PyObject*);
 
+		DECLARE_PYTHON_DATA_SYMBOL(PyTypeObject, PyUnicode_Type);
+
 #ifdef _DEBUG
 		// In a debug build dealloc is a function but for release builds its a macro
 		DECLARE_PYTHON_SYMBOL(void, _Py_Dealloc, PyObject*);
@@ -203,6 +208,7 @@ namespace Plugins {
 					RESOLVE_PYTHON_SYMBOL(PyByteArray_AsString);
 					RESOLVE_PYTHON_SYMBOL(PyUnicode_FromKindAndData);
 					RESOLVE_PYTHON_SYMBOL(PyLong_FromLong);
+					RESOLVE_PYTHON_SYMBOL(PyLong_FromUnsignedLongLong);
 					RESOLVE_PYTHON_SYMBOL(PyLong_AsLongLong);
 					RESOLVE_PYTHON_SYMBOL(PyModule_GetDict);
 					RESOLVE_PYTHON_SYMBOL(PyDict_New);
@@ -277,6 +283,8 @@ namespace Plugins {
 					RESOLVE_PYTHON_SYMBOL(PyFloat_AsDouble);
 					RESOLVE_PYTHON_SYMBOL(PyObject_GetIter);
 					RESOLVE_PYTHON_SYMBOL(PyIter_Next);
+
+					RESOLVE_PYTHON_SYMBOL(PyUnicode_Type);
 				}
 			}
 			_Py_NoneStruct.ob_refcnt = 1;
@@ -420,6 +428,7 @@ extern	SharedLibraryProxy* pythonLib;
 #define PyByteArray_AsString	pythonLib->PyByteArray_AsString
 #define PyUnicode_FromKindAndData  pythonLib->PyUnicode_FromKindAndData
 #define PyLong_FromLong			pythonLib->PyLong_FromLong
+#define PyLong_FromUnsignedLongLong  pythonLib->PyLong_FromUnsignedLongLong
 #define PyLong_AsLongLong		pythonLib->PyLong_AsLongLong
 #define PyModule_GetDict		pythonLib->PyModule_GetDict
 #define PyDict_New				pythonLib->PyDict_New
@@ -497,6 +506,8 @@ extern	SharedLibraryProxy* pythonLib;
 #define PyFloat_AsDouble		pythonLib->PyFloat_AsDouble
 #define	PyObject_GetIter		pythonLib->PyObject_GetIter
 #define	PyIter_Next				pythonLib->PyIter_Next
+
+#define PyUnicode_Type (*pythonLib->PyUnicode_Type)
 
 #ifndef _Py_DEC_REFTOTAL
 /* _Py_DEC_REFTOTAL macro has been removed from Python 3.9 by: https://github.com/python/cpython/commit/49932fec62c616ec88da52642339d83ae719e924 */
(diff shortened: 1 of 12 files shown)


gizmocuz:
I noticed some changes also in mainworker.cpp, not sure if those are needed.

How does this work, you provide a idx value (like 10) and the device is created with idx=10 ? Or is it the DeviceID that is using the number 10 ?

I still think the best method for now and the feature is to create the following functions

idx = createDevice(Type, SubType, Unit, optional switchtype, Name, , nValue/svalue)
updateDevice(idx, nvalue)
updateDevice(idx, svalue)
updateDevice(idx, nvalue, svalue)

jkl1337:
> I noticed some changes also in mainworker.cpp, not sure if those are needed.

The changes are pretty minimal and they are only to get the ID column propagated to the plugin. Right now the plugin extension points in there ride on the fact that Unit is carried around in one of the database result vectors.
 
> How does this work, you provide a idx value (like 10) and the device is created with idx=10 ? Or is it the DeviceID that is using the number 10 ?

Nope just do dev = Device('Name', DeviceID='arbitrary', Unit=any_number, Type=...), and then dev.Create(). Just like before, except Unit does not have to be unique. Now dev.ID (which always existed) will give the generated ID. The Devices dictionary will be keyed by Idx (ID column) and callbacks get Idx as the first argument.

As has been stated elsewhere the tuple (HardwareID, DeviceID, Unit, Type, Subtype) is a candidate key for devices, and HardwareID is fixed per plugin, which means the key within a plugin would be (DeviceID, Unit, Type, Subtype). One of the objections in the zigbee2mqtt issue thread for not migrating away from Unit was that it would be cumbersome to have to use a composite key within plugin code. Well the answer to that is just don't do that; there's a perfectly good primary key - Device Idx (aka ID column).
 
> I still think the best method for now and the feature is to create the following functions
> 
> idx = createDevice(Type, SubType, Unit, optional switchtype, Name, , nValue/svalue)
> updateDevice(idx, nvalue)
> updateDevice(idx, svalue)
> updateDevice(idx, nvalue, svalue)

This is effectively what this change does. After calling Domoticz.UseDeviceIndex() at initialization, the Device constructor API changes subtly to no longer require a Unit value (however DeviceID and Unit must be unique - as mentioned I can loosen this constraint to the entire 4-tuple). Then the key for the Devices dictionary is the Device Idx and all callbacks get the Idx instead of Unit.

The niceness of this is that implementing the feature this way is "upgrading" existing plugin code is easy - generate a unique DeviceID (which plugins like zigbee2mqtt already do) and don't do that painful Unit generation that a lot of plugins have to do. It is easy to feature test UseDeviceIndex at runtime and switch accordingly with one code base.

Unfortunately some of the internal callback functions are a little ugly now to be backward compatible, but there will be some price to pay.

gizmocuz:
Thanks for the clarifications! Makes sense !... the only thing is that this 'new' method is the preferred (correct) method, and currently we have to 'enable' the preferred method

I have a few things I would like to see changed:

No changes in mainworker (SwitchLightInt/SetSetPointInt), SQLHelper etc... should not be needed, you can build the vector
Also makes this PR cleaner , only python related code is changed

using nullptr is good (and should be used), but sometimes i noticed you are using

`if (pointer == nullptr)`

and sometimes you use

`if (!pointer)`

(i would prefer == nullptr)

or

`return nullptr;`

but sometimes (also in the same function) you use

`return 0;`

(I would prefer return nullptr)


If you also implemented the below new python calls, we could in the end remove all obsolete calls, and display warnings in the log if old functions are still used... 

idx = createDevice(Type, SubType, Unit, optional switchtype, Name, , nValue/svalue)
updateDevice(idx, nvalue)
updateDevice(idx, svalue)
updateDevice(idx, nvalue, svalue)
removeDevice(idx)

jkl1337:
> No changes in mainworker (SwitchLightInt/SetSetPointInt), SQLHelper etc... should not be needed, you can build the vector
> Also makes this PR cleaner , only python related code is changed

This is unavoidable. The changes in mainworker, SQLHelper, and EventSystem are touching python plugin specific code that was already present. The plugin code is tightly coupled in a bunch of places in the core. I didn't add any more coupling. main/EventSystem.cpp change is to the SecurityEventCallback already present. The 1st SQLHelper change is to pass the ID to the DeviceAdded callback already there, and the second is for the DeviceRemoved callback. (what I *will* do with that one is clean it up a bit.. I passed through it very quickly and now on review it is clear that the original code was quite ugly and deserves a minor cleanup.)

The first WebServer.cpp change is very minor related to the admin API switch debug code (I think this is the only thing not completely plugin related that got affected and I think it's pretty benign). The second change is for existing plugin related Node Failure notification.

I think changing SwitchLightInt to take idx is very reasonable, as its caller already takes it, and the same is true for SetSetPointInt. That these changes are only affecting internal API functions without much collateral damage around the code base I think is telling. 
 
> using nullptr is good (and should be used), but sometimes i noticed you are using
> ...
Agreed.

> `return 0;`
> (I would prefer return nullptr)

I do not disagree. I think the code you're referring to is some of the error handling code. This was cut-and-paste from elsewhere, there's a lot of existing code in those modules that still uses 0. I'll go through it in a separate commit to keep the diff noise down. (The insert code I put in can be cleaned up a bit).

> 
> If you also implemented the below new python calls, we could in the end remove all obsolete calls, and display warnings in the log if old functions are still used...
> 
> idx = createDevice(Type, SubType, Unit, optional switchtype, Name, , nValue/svalue)
> updateDevice(idx, nvalue)

I don't disagree with cleaning the API in general, however I think it's a separate problem.

tl;dr: i don't disagree with renaming things - but I don't see how to avoid an explicit "version" or "use" declaration.

Edit: on rereading and thinking about it a different way - the main practical problem I have with this is:
With the current proposal the new style plugin has to make a declaration call (that they can feature test) and then the entire semantics of Devices dictionary and Device changes cleanly and completely - it's a dichotomous API - the names are the same but that's it - the interface has actually changed. If we get rid of the declaration call and add a new parallel API then what do we do with the existing interface including Devices table? It isn't feasible to keep it around - something has to tell the framework to make the old 255 device limited table go away - we can't cleanly deprecate it and support more than 255 devices. Like if this is a problem with naming, I get it, the names can be changed, but that doesn't deal with the old API. At least the mechanism proposed deals with it.

SIdebar:
I think the other issue here is - as wonky and arbitrarily hobbled as the existing plugin API is, there are probably a lot of consumers of it that never run into its limitations - and what gain is there in forcing this long tail to migrate at all? Plugins like zigbee2mqtt and possibly a zwave equivalent, other general purpose plugins like for mqtt tasmota and esphome clearly have a benefit to overcoming this limitation - and its not the prettiest that they have to add some declaration - but they're going to have to change *anyway*. Maybe the name UseDeviceIndex is stupid (it is and it isn't) - call it UseAPIVersion(2) - but I think the backward/forward compatibility has some merit.  

I won't belabor this point further but it is something to consider.

jkl1337:
We could have a distinct import for the new API. The current name is `Domoticz` though, so what would an appropriate name be? `domoticz` is an option or `domoticz_api`?

pipiche38:
Yes please keep the backward compatibility, breaking all python plugin will really be a shame

BTW have you see that: https://www.domoticz.com/forum/viewtopic.php?f=65&t=35452 which looks pretty promising

gizmocuz:
The plugin framework is using functions from the mainworker/sqlhelper so they need to change.
I like your PR, but I'm not going to merge it with these changes.
Please only change the plugin related code

@pipiche38 , sooner or later this will need to happen
But that's why I opted for new function names

idx = createDevice(Type, SubType, Unit, optional switchtype, Name, , nValue/svalue)
updateDevice(idx, nvalue)
updateDevice(idx, svalue)
updateDevice(idx, nvalue, svalue)
removeDevice(idx)

This way after the next stable we could give a warning in the log that old plugin methods are used and that they need to be adopted to the new methods

pipiche38:
@jkl1337  could you kindly use the 2 Py_BEGIN/END_ALLOW_THREADS macros accordingly to PR #4518 especially when doing m_sql calls

jkl1337:
> The plugin framework is using functions from the mainworker/sqlhelper so they need to change.
> I like your PR, but I'm not going to merge it with these changes.
> Please only change the plugin related code

I don't understand the ask here. The code changed in SQLHelper and mainworker is entirely plugin related. Can you provide at least of couple examples or specific commentary on the offending lines and what a feasible alternative would be?

For instance mainworker.cpp:11430. That is simply the call into the plugin hardware, everything following is a large switch statement with other unrelated hardware specific implementation details. How should that be handled differently?

gizmocuz:
Sorry if I was unclear, I am talking about these two changes that are not plugin related and used on multiple locations
(And do not need to change)

```
bool MainWorker::SwitchLightInt(const std::vector<std::string>& sd, std::string switchcmd, int level, const _tColor color, const bool IsTesting, const std::string& User)
bool MainWorker::SwitchLightInt(uint64_t idx, const std::vector<std::string> &sd, std::string switchcmd, int level, const _tColor color, const bool IsTesting, const std::string &User)

bool MainWorker::SetSetPointInt(const std::vector<std::string>& sd, const float TempValue)
bool MainWorker::SetSetPointInt(uint64_t idx, const std::vector<std::string> &sd, const float TempValue)


```

jkl1337:
I'll gladly change it. It did seem like those particular functions were entirely internal to the class, used nowhere outside of it and probably could even be made private, though.

gizmocuz:
@jkl1337 , a quick search in all files for 'SwitchLightInt' would reveal it's used on multiple locations

jkl1337:
Ah yes for that one, I suppose I knew that as necessarily the calls were changed. SetSetPointInt was the one not called elsewhere. In any event I'll change it.

gizmocuz:
@jkl1337 , Did you find some time to make the changes ?

kovainfo:
any news?

Bartekn86:
hi, any news ?

gizmocuz:
I am not sure if @jkl1337 is working on this change anymore as the last reply was from 25 March.
If someone wants to continue this work, please do so

Bartekn86:
nothing ? :(

gizmocuz:
@Bartekn86 , If you want to continue with the PR, please go ahead

waltervl:
There is some progress on this topic by Dnpwwo, not sure if it is additional work on this PR: 
https://domoticz.com/forum/viewtopic.php?p=275976#p275976

PR #4745, opened 2021-03-22
Execution progress and result text not visible

ridderr:
If clicked in Hardware on the "Has Node Failed" button no text is displayed in the popup during execution and when status of request is display. This is caused by this color #FFF (which is the same as the background). 
When set to #000 the text in the popup is visible during execution of the command and when result is displayed.

### DIFF
--- a/www/css/style.css
+++ b/www/css/style.css
@@ -35,7 +35,7 @@ body {
 	background: #202020 url(images/img01.jpg) repeat;
 	font-family: Arial, Helvetica, sans-serif;
 	font-size: 10pt;
-	color: #FFF;
+	color: #000;
 }
 
 .bootbox-body {


ridderr:
hasNodeFailed-FFF
hasNodeFailed-000

gizmocuz:
Is this with the default theme ?

ridderr:
No, the default is working because that has a dark background popup box. All other themes do have a white background and therefore the text is not visible.
I'm not sure if the proposed change has to be applied to the body element because this could impact more. But I'm not skilled enough on css to make it more specific for the modal popup only.

gizmocuz:
Yes, you need to make a patch for the theme you are using.
If it is a Domoticz supplied theme, you can find the theme in the www/styles folder, each has a custom.css where you can override the colors

ridderr:
Agree, i will give it a try :-) and cancel this pull request.

PR #4749, opened 2021-03-24
Rollback moisture sensor Opus-XT300 from custom to previous moisture…

Chrominator:
Hi,

Got 3 Opus-XT300 moisture sensors that were fully functional last spring (2020) in Domoticz.
This spring I put new batteries in them, the already defined sensors in Domoticz did not fully work ; the temperature was OK, however the moisture was not updated.

I delete the old sensors in Domoticz and choose to re-create the sensors.
The type of the sensor moved  from moisture type to custom type.
The issue is that there are 3 temperature sensors created with each their own ID, but only one custom sensor with ID=00000000.
This moisture sensor is updated, however we should have 3 of them.

Domoticz Bug Opus

The program has been changed in last December : 

Screenshot_2021-03-24 domoticz domoticz

Since I do not find why the moisture sensor which used to be fine has been changed to custom sensor, I've rolled back the source to the previous status.

After compilation, having deleted the defective sensors and having started Domoticz, the new sensors have been created and they now work fine, as they did before.

Domoticz - Opus rollback

I suggest to keep those moisture sensor type, which works well and which uses the correct unit "cb" to measure soil moisture.

Thanks.

### DIFF
--- a/hardware/Rtl433.cpp
+++ b/hardware/Rtl433.cpp
@@ -395,8 +395,15 @@ bool CRtl433::ParseData(std::map<std::string, std::string>& data)
 	if (haveMoisture)
 	{
 		//moisture is in percentage
-		//SendCustomSensor((uint8_t)sensoridx, (uint8_t)unit, batterylevel, static_cast<float>(moisture), model, "%", snr);
-		SendMoistureSensor(sensoridx, batterylevel, moisture, model, snr);
+		if (haveChannel)	// Channel is used to identify each sensor
+		{
+			SendCustomSensor((uint8_t)sensoridx, (uint8_t)channel, batterylevel, static_cast<float>(moisture), model, "%", snr);
+		}
+		else
+		{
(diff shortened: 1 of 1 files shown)


gizmocuz:
I suspect this was requested because another one had a percentage moisture sensor
A custom sensor you can edit and change the percentage to cb without issues

Maybe you can swap the unit and sensoridx in the call to SendCustomSensor ?

Chrominator:
rtl_433 provides a percentage moisture even with my Opus-XT300, maybe it's right to use a custom sensor.
image
The issue is that the change is not working with several sensors.
I'll try to fix it with custom sensor.
Thanks for your time.

Chrominator:
I've preserved the code and added a testing on the channel of the sensor.
If there's a channel defined, it is used as the identifier.
This allows to get one device per Opus-XT300 in the device list rather only one with the previous code.

image

gizmocuz:
Great work!! Thank you for the PR!!

PR #4752, opened 2021-03-27
Prevent Async Secure Read Exception: 335544539

### LINKED ISSUE #124: [plugin Linky] Pb lors de l'ajout dans Domoticz
Je voulais utliser votre plugin Linky sous Domoticz mais lorsque je choisis LInky dans la liste des HW, il n'y a aucun champ relatif aux identifiants Enedis. Y-a-t-il des conditions pour afficher ces champs? C'est sur une autre page ?

J'utilise un NAS Synology DS218 et Domoticz 2020.1.
image

guillaumezin:
Prevent Async Secure Read Exception: 335544539, short read message
SSL/TLS short read should be ignored, it happens on some Python versions on some hardware implementations (Synology mainly)
It Should fix https://github.com/guillaumezin/DomoticzLinky/issues/124

### DIFF
--- a/hardware/plugins/PluginTransports.cpp
+++ b/hardware/plugins/PluginTransports.cpp
@@ -537,6 +537,7 @@ namespace Plugins {
 					(e.value() != 1) &&	// Stream truncated, this exception occurs when third party closes the connection
 					(e.value() != 121) &&	// Semaphore timeout expiry or end of file aka 'lost contact'
 					(e.value() != 125) &&	// Operation canceled
+					(e.value() != 335544539) &&	// SSL/TLS short read, should be ignored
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thank you for this PR!

gizmocuz:
@guillaumezin , this check if on multiple locations in PluginTransports.cpp, does this not have to be added on all of these locations ?

guillaumezin:
I've seen the problem only with "Async Secure Read Exception", and it seems quite specific to SSL/TLS connection, so I corrected it only in one place. See https://github.com/boostorg/beast/issues/38#issuecomment-230899097 and next comments for instance : it means that the connection is not gracefully closed from third party, but Domoticz can't do anything about it.

PR #4753, opened 2021-03-28
Update EvohomeRadio.cpp

brucemiranda:
Only print the OT Bridge detected message when logging is enabled.

### DIFF
--- a/hardware/EvohomeRadio.cpp
+++ b/hardware/EvohomeRadio.cpp
@@ -727,8 +727,10 @@ void CEvohomeRadio::ProcessMsg(const char* rawmsg)
 		if (msg.id[0].GetIDType() == CEvohomeID::devBridge) //if we get a message from the OT Bridge capture it's ID
 		{
                         if (!m_nOtbID)
+                        {
                                 m_nOtbID = msg.GetID(0);
-			Log(LOG_STATUS, "evohome: opentherm bridge detected, ID:0x%x", m_nOtbID);
+                                _log.Log(LOG_STATUS, "evohome: opentherm bridge detected, ID:0x%x", m_nOtbID);
+                         }		
 		}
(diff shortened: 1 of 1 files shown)


hardware/EvohomeRadio.cpp
```diff
@@ -728,7 +728,7 @@ void CEvohomeRadio::ProcessMsg(const char* rawmsg)
 		{
                         if (!m_nOtbID)
                                 m_nOtbID = msg.GetID(0);
-			_log.Log(LOG_STATUS, "evohome: opentherm bridge detected, ID:0x%x", m_nOtbID);
+			Log(true, LOG_STATUS, "evohome: opentherm bridge detected, ID:0x%x", m_nOtbID);
```
gizmocuz: I think this has to be:

if (!m_nOtbID)
{
   m_nOtbID = msg.GetID(0);
   Log(LOG_STATUS, "evohome: opentherm bridge detected, ID:0x%x", m_nOtbID);
}

This way it is only displayed once

brucemiranda: It's displayed once, each time the OT Bridge is detected on the network. So it could be displayed often if you have Status logging enabled.

gizmocuz: Don't need to as the ID won't change ?

brucemiranda: ID will not change. So maybe we can go with your suggestion. But it was set similar to this before the change. So I have no idea how doing what you suggested will make it display that message only once.

gizmocuz: Initially the value of **m_nOtbID** = 0, if it is zero, the ID will be set **m_nOtbID = msg.GetID(0);**
So this happens one time.
I the same time you can log the ID

The code for this I posted a few messages back

brucemiranda: Sent a new patch, with your suggestion

gizmocuz:
Thank you for the PR and the quick fix!!

PR #4754, opened 2021-03-29
Meteorologisk: Refactoring, saver handling of wrong parameters, etc

kiddigital:
switch away from deprecated API call, dedicated hardware settings screen.

### DIFF
--- a/hardware/Meteorologisk.cpp
+++ b/hardware/Meteorologisk.cpp
@@ -55,7 +55,6 @@ CMeteorologisk::CMeteorologisk(const int ID, const std::string &Location) :
 	m_Location(Location)
 {
 	m_HwdID=ID;
-	m_URL = "";
 }
 
 void CMeteorologisk::Init()
@@ -115,7 +114,8 @@ void CMeteorologisk::Init()
 	}
 
 	std::string szLoc = CURLEncode::URLEncode(m_Location);
-	sURL.flush();
+	sURL.str(std::string());
+	sURL.clear();
 	sURL << Meteorologisk_forecast_URL << szLoc;
 	m_ForecastURL = sURL.str();
 	Debug(DEBUG_HARDWARE, "Meteorologisk: Set Forecast URL to: %s", m_ForecastURL.c_str());
@@ -157,7 +157,7 @@ bool CMeteorologisk::StopHardware()
 void CMeteorologisk::Do_Work()
 {
 	Log(LOG_STATUS, "Started...");
-	Debug(DEBUG_HARDWARE, "Metereologisk module started with Location parameters Latitude %f, Longitude %f!", m_Lat, m_Lon);
+	Debug(DEBUG_NORM, "Metereologisk module started with Location parameters Latitude %f, Longitude %f!", m_Lat, m_Lon);
 
 	int sec_counter = 290;
 	while (!IsStopRequested(1000))
--- a/hardware/Meteorologisk.cpp
+++ b/hardware/Meteorologisk.cpp
@@ -125,10 +125,6 @@ bool CMeteorologisk::StartHardware()
 {
 	Init();
 
-	if(m_URL.empty()) {
-		Log(LOG_ERROR, "Unable to properly start due to missing or incorrect Location parameters (Latitude, Longitude)!");
-	}
-
 	RequestStart();
 
 	//Start worker thread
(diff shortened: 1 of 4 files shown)


hardware/Meteorologisk.cpp
```diff
@@ -48,22 +48,90 @@ std::string ReadFile(std::string filename)
...
 
+#define Meteorologisk_URL "https://api.met.no/weatherapi/locationforecast/2.0/complete.json"
+#define Meteorologisk_forecast_URL "https://darksky.net/#/f/"
+
 CMeteorologisk::CMeteorologisk(const int ID, const std::string &Location) :
 	m_Location(Location)
 {
 	m_HwdID=ID;
-	Init();
+	m_URL = "";
```
gizmocuz: Thank you for the PR.
You don't have to initialize a std::string in a constructor of a class. This is already done

hardware/Meteorologisk.cpp
```diff
@@ -48,22 +48,90 @@ std::string ReadFile(std::string filename)
...
+	}
+
+	if(!(m_Lat == 1 || m_Lon == 1))
+	{
+		sURL << Meteorologisk_URL << "?lat=" << m_Lat << "&lon=" << m_Lon;
+		m_URL = sURL.str();
+	}
+
+	std::string szLoc = CURLEncode::URLEncode(m_Location);
+	sURL.flush();
```
gizmocuz: best to use this to clear/reset a stringstream

std::stringstream().swap(sUrl);

hardware/Meteorologisk.cpp
kiddigital: done

hardware/Meteorologisk.cpp
kiddigital: used str(std::string()) and clear()

Is that ok too?

hardware/Meteorologisk.cpp
```diff
@@ -48,22 +48,90 @@ std::string ReadFile(std::string filename)
...
 }
 
 bool CMeteorologisk::StartHardware()
 {
+	Init();
+
+	if(m_URL.empty()) {
+		Log(LOG_STATUS, "Unable to start due to missing or incorrect Location parameters (Latitude, Longitude)!");
+		return false;
+	}
```
kiddigital: @gizmocuz , this check prevents the hardware module to actually start its thread because of missing parameters.

But the Domoticz watchdog will report 'unresponsive' as it still expects that a thread should be running to at least update the heartbeat I guess?

hardware/Meteorologisk.cpp
gizmocuz: Yes, but the swap is c++11 prefered way

hardware/Meteorologisk.cpp
gizmocuz: That's correct. You could add this line at the top of GetMeterDetails. It will log this every 5 minutes but that's fine

kiddigital: Done

kiddigital:
@gizmocuz , one more question.. I noticed `ERROR: Unknown Sub type for Packet type= 55 : 71` debug messages when calling `SendRainRateSensor` (function of `DomoticzHardware.cpp`). I did not change anything there, so it was probably there before.

Looking at the definitions, indeed the subtype 71 (sTypeRAINByRate) does not exist as a subtype of Type 55 (pTypeRAIN). Everything does work as expected as only the messages is generated and the data is stored, so no real problem (yet?).

hardware/Meteorologisk.cpp
```diff
@@ -168,13 +167,20 @@ void CMeteorologisk::Do_Work()
...
+				catch (...)
+				{
+					Log(LOG_ERROR, "Unhandled failure getting/parsing data!");
+				}
 			}
-			catch (...)
+			else
 			{
-				Log(LOG_ERROR, "Unhandled failure getting/parsing http data!");
+				Log(LOG_STATUS, "Unable to perform work due to missing/incorrect startup parameters!");
```
gizmocuz: This message is a bit unclear... imaging you are looking at this 1 day later... you totally missed the startup ...
I think it's clearer to log here 

"Unable to properly start due to missing or incorrect Location parameters (Latitude, Longitude)!"

(And remove it from StartHardware)

As **sec_counter** is initialized at 290 (I would initialize it to to **Meteorologisk_Poll_Interval -5**)
This log message is quickly displayed

gizmocuz:
I fixed the debug message "ERROR: Unknown Sub type for Packet type= 55 : 71 "

hardware/Meteorologisk.cpp
kiddigital: Also done :)

gizmocuz:
Thank you again for this PR and time!!

PR #4755, opened 2021-03-29
Adding support for Teleinfo connected over TCP.

lunakato:
Support for the french smart meter Teleinfo was limited to directly connected serial. 
This adds support for Teleinfo connected to a serial-to-TCP adapter.
The TCP client code was mostly taken from P1MeterTCP.

### DIFF
--- a/main/mainworker.cpp
+++ b/main/mainworker.cpp
@@ -53,6 +53,7 @@
 #include "../hardware/OTGWTCP.h"
 #include "../hardware/TeleinfoBase.h"
 #include "../hardware/TeleinfoSerial.h"
+#include "../hardware/TeleinfoTCP.h"
 #include "../hardware/Limitless.h"
 #include "../hardware/MochadTCP.h"
 #include "../hardware/EnOceanESP2.h"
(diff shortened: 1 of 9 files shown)


main/mainworker.cpp
```diff
@@ -13537,6 +13541,8 @@ void MainWorker::HeartbeatCheck()
 
 bool MainWorker::UpdateDevice(const int DevIdx, const int nValue, const std::string &sValue, const std::string &userName, const int signallevel, const int batterylevel, const bool parseTrigger)
 {
+	_log.Debug(DEBUG_NORM, "MainWorker::UpdateDevice: ID: %d", DevIdx);
```
gizmocuz: Could you remove this debug line (and extra return)

gizmocuz:
Thank you for this PR !!!
Question, I also see a rate limit change being set to 0, why is this ? Does this not cause more CPU cycles ?
One small remark, please undo the changes in mainworker::updatedevice

main/mainworker.cpp
lunakato: Not a necessary part of this change, so removed.

lunakato:
The rate limit in this code (and in the P1 Smart Meter code) is an optional limiter between receiving the smart meter messages and passing them on to the main Domoticz code. The main loop is not affected.

Perhaps it's better to keep the 60 seconds default.  I'll adjust the code.

gizmocuz:
Thank you for this PR and your time !!!

PR #4756, opened 2021-04-02
MQTT: Add option to set Retain flag for messages when using Index or Name publishing scheme

kiddigital:
With this PR it is now possible to let Domoticz publish messages with the Retain bit set to True.

This option is available only for the Index and Name publishing scheme's as they are publishing on a topic per device bases.

Also, now Domoticz uses QoS = 1 for publishing messages (at least once) instead of best effort, ensuring that messages do get published.

The outgoing messages now also contain the LastUpdate data from each device, so in case of retained messages (which the client can check when looking at the Retain bit), a client can also see how old these messages are.

NOTE: When testing this, remember to clear the MQTT brokers persistancy database to actually flush Retained messages. Domoticz does not 'clean' these Retained messages when selecting a different publishing scheme!

### DIFF
--- a/hardware/MQTT.cpp
+++ b/hardware/MQTT.cpp
@@ -17,7 +17,6 @@
 #define TOPIC_OUT	"domoticz/out"
 #define TOPIC_IN	"domoticz/in"
 #define QOS         1
-#define RETAIN		true
 #define RETAIN_BIT	0x80
 
 namespace
--- a/hardware/MQTT.cpp
+++ b/hardware/MQTT.cpp
@@ -17,6 +17,7 @@
 #define TOPIC_OUT	"domoticz/out"
 #define TOPIC_IN	"domoticz/in"
 #define QOS         1
+#define RETAIN		true
 
 namespace
 {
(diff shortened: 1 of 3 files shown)


gizmocuz:
Thank you for this PR! Question, why not just make a option for this (checked/unchecked) in the Gui, like it is in the C++ code.
That's much clearer/cleaner.

kiddigital:
> Thank you for this PR! Question, why not just make a option for this (checked/unchecked) in the Gui, like it is in the C++ code.
> 
> That's much clearer/cleaner.
> 
> 

After some discussion and thought, the Retain option only makes sense when you select a publishing scheme that uses a topic per device.

For the other schemes like Flat and Hierarchical, the Retain bit just preserves the last message of the device that happens to be updated last. From a client perspective you can not determine anything from it.

Therefor I opted to only make it an option for schemes that make sense with this option. This way (dropdown) also the UI is clear instead of a toggle that actually works for some schemes (more confusing I think).

gizmocuz:
Thanks!!!

PR #4757, opened 2021-04-02
Tar extract less verbose

SkyLined:
Outputting the entire list of files in the tar cluttered the output. Tar output is now less verbose and shows a progress bar instead of the list of files.

### DIFF
--- a/updaterelease
+++ b/updaterelease
@@ -23,7 +23,7 @@ then
  echo "Problem in downloaded Domoticz archive. Stopping update!"
  exit 1
 fi
-tar xvfz domoticz_release.tgz
+tar xfz domoticz_release.tgz
 rm domoticz_release.tgz
 echo "Restarting Domoticz... (please standby...)"
 sudo service domoticz.sh restart


gizmocuz:
Thank you for this PR.
Could you change it to

tar tzf

We have a PR ready for updatebeta ... that will be merged as soon as we make a new release version.
Changes are we will adopt these changes in updaterelease as well

SkyLined:
I do not follow. Changing the code from "tar xfz" to "tar tzf" would make it *check* the file (again, see line 20) rather than extract it.

gizmocuz:
Yes sorry about this, i meant

tar xfz domoticz_xxx.tgz --checkpoint=.100

SkyLined:
Again, I do not follow: what use is a checkpoint without an action?

Also, it sounds like whatever this additional change is, it would fix something different than the verbosity and should probably be in a separate commit for that reason.

gizmocuz:
Okey, never mind then... As mentioned, we are working on the 'updatebeta' already (see the PR), once this get's merged the changes will be applied to 'updaterelease' as well

PR #2033, opened 2018-01-01
Change Devices view to add Zwave Nodeid to the Hardware column for ZWAVE devices.

jvanderzande:
This change will modify the Domoticz devices view and adds the ZWAVE device Nodeid to the Hardware field for the ZWAVE devices. This allows to sort the ZWAVE Domoticz devices on their ZWAVE Nodeid to determine which Domoticz devices belong to which physical ZWAVE device.
domoticz z-wave id

PR #4762, opened 2021-04-06
Add each zwave unit as a separate hardware in the device list

paalkr:
To make zwave device filtering easier in the device list. For large installations with hundreds of zwave units and thousands of Domoticz devices, filtering and searching for devices that belongs to the same zwave unit is hard. This PR adds each zwave unit as a separate "hardware" under the hardware tab in the device list filter, in addition to label each device with the same hardware name in the list accordingly.
z-wave-devices

### DIFF
--- a/www/app/devices/Devices.js
+++ b/www/app/devices/Devices.js
@@ -625,8 +625,17 @@ define(['app', 'livesocket'], function(app) {
                     if (response.result !== undefined) {
                         $ctrl.devices = response.result
                             .map(function(item) {
-                                var isScene = ['Group', 'Scene'].includes(item.Type);
+                                if (item.HardwareTypeVal == 21) {
+                                    var ZWID = item.ID.substr(-4, 2);
+                                    if (ZWID == '00') {
+                                        ZWID = item.ID.substr(-2, 2);
+                                    }
+                                    ZWID = '0x' + ZWID;
+                                    var ZWIDdec =  ("00" + parseInt(ZWID)).slice(-3);
+                                    item.HardwareName = item.HardwareName + " " + ZWIDdec + ' (' + ZWID + ')';   
+                                }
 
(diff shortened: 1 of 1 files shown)


paalkr:
The code is based on the work done by @jvanderzande over at https://github.com/domoticz/domoticz/pull/2033, but rebased according to the new device list code.

gizmocuz:
@paalkr , thank you for this PR. Aldo this is an option, another option is to just select(check) your zwave hardware (I would just label this ZWave bdw..) and in the filter at the top right enter 0F ... would that not also do the same ?
Or don't select the zwave device, and in the filter enter 'zwave 0f'

In the end, the zwave devices should actually be visible under the zwave setup button/page

paalkr:
If for example search for z-stick 03 in the search filed, in my use case i get 112 results in the device list. Yes, the zwave ID is highlighted on each of the 112 devices. But finding the exact 8 devices in that list corresponding to zwave node 03, is almost impossible. Selecting the zwave node 03 in the hardware list gives me exactly the 8 devices I'm looking for.

"Regular" search for Z-stick 03
image
...
image

Filtered using the new code
image

gizmocuz:
Thats true! Still in the end, I think we should add this 'node/device list' in the ZWave setup page as well.
Thank you for the PR and time!

paalkr:
Thank you very much. I absolutely agree that there might be better ways of selecting / displaying devices that belongs to a zwave node, for example adding a tree view in the zwave hardware page. But other solutions will probably require more work.

rwaaren:
@paalkr It seems that devices directly connected to the controller (in my case a Zwave stick) cannot be selected using this mechanism.

Also see [this forum post](https://www.domoticz.com/forum/viewtopic.php?p=275344#p275344)

PR #4766, opened 2021-04-10
USBtin & Multibloc improvement / Bloc 9 now supported

ghost:
This pull request contains many feature for USBtin and USBtin_MultiblocV8 improvement.
- Increase CAN buffer size receiver to avoid loss of frames.
- Increase Max number of blocs supported
- Add Bloc 9 management (like SFSP Bloc)
- Improve the auto release key code generation method when a user clicked on a virtual device who send his code on CAN bus. The old method was hazardous if more than one client clicked on 2 different virtual switch, the release keycode could not be send on CAN.
About that, I would like to know if there is a way with some switch, to detect a "press" state (who could send the "press" state) and detect the release of the switch (who could idealy send the release keycode) which would be idealy because a real EnOcean switch send a push key code and a release key code, this means that we could dim a light when we push the switch without release them.


gizmocuz:
Could you check your PR again ? It contains 2300 changes, I think something is wrong...

ghost:
I checked, for me it is normal because there is really a lot of change. The changes are only in USBtin files, no ?

ghost:
I can't acces to the button "resolve conflicts" is it normal ?

ghost:
Oh no.. I've understand. this is an encoding file problem ! thanks to windows... UTF8 has change to ansi or something else. I re check...

gizmocuz:
Maybe it is best to make a new fork of Domoticz, checkout the source code (using normal git commands, or under windows with ToirtuiseGit)
Make a new branch, make the changes and commit and make a pull request

PR #4767, opened 2021-04-12
Add ability to set an EvoHomeWeb setpoint until next in schedule

sidepipe:
If no "until time" is specified in the set setpoint command in TemporaryOverride
mode, use the next switchpoint time if possible. Also fixes a bug that used
uninitialised date if no until time was given.


rwaaren (approved):
Tested with the API 
`/json.htm?type=setused&idx=<id>&setpoint=<temperature>&mode=TemporaryOverride&used=true`

and updateSetPoint method of dzVents

Both with and without an until date

gizmocuz:
Thank you for the PR !!

PR #4769, opened 2021-04-13
USBtin & Multibloc improvement / Bloc 9 now supported 

ghost:
This pull request contains many feature for USBtin and USBtin_MultiblocV8 improvement.

    Increase CAN buffer size receiver to avoid loss of frames.
    Increase Max number of blocs supported
    Add Bloc 9 management (like SFSP Bloc)
    Improve the auto release key code generation method when a user clicked on a virtual device who send his code on CAN bus. The old method was hazardous if more than one client clicked on 2 different virtual switch, the release keycode could not be send on CAN.
    About that, I would like to know if there is a way with some switch, to detect a "press" state (who could send the "press" state) and detect the release of the switch (who could idealy send the release keycode) which would be idealy because a real EnOcean switch send a push key code and a release key code, this means that we could dim a light when we push the switch without release them.

[recreated following advice from gizmocuz , thanks to him]

### DIFF
--- a/hardware/USBtin.cpp
+++ b/hardware/USBtin.cpp
@@ -126,7 +126,7 @@ void USBtin::Do_Work()
 			{
 				switch(m_EtapeInitCan){
 					case 0 :
-						_log.Log(LOG_STATUS, "USBtin: Serial port is now open !");
+						Log(LOG_STATUS, "USBtin: Serial port is now open !");
 						CloseCanPort(); //more cleaner to close in first, sometimes the gateway maybe already open...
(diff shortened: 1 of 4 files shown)


gizmocuz:
Thank you for this PR !!
One question, could you replace the log calls from

_log.Log(....

to

Log(...

ghost:
Yes !

gizmocuz:
If you make the changes and commit they will automatically appear in this PR

gizmocuz:
Thank you for your PR and time!!

ghost:
with pleasure :-)

PR #4770, opened 2021-04-13
Neheb fmt

gizmocuz:


neheb:
I'd look at my format conversions. I haven't tested to see if they're correct.

hardware/Daikin.cpp
```diff
@@ -691,11 +692,7 @@ bool CDaikin::SetLedOnOFF(const bool OnOFF)
 void CDaikin::InsertUpdateSwitchSelector(const unsigned char Idx, const bool bIsOn, const int level, const std::string& defaultname)
 {
 	unsigned int sID = Idx;
-
-	char szTmp[300];
-	std::string ID;
-	sprintf(szTmp, "%c", Idx);
-	ID = szTmp;
+	std::string ID = fmt::format("{}", Idx);
```
neheb: I'm curious about this. What's the difference between this and fmt::to_string() ?

gizmocuz: I only changed the 'auto' to 'std::string' ..
But do you mean 'std::to_string' ? I guess in this case it's the same

gizmocuz: Looking at this again, the original code tries to print it to a character, but having looked again and again, this code is not correct. Looking at it again, that whole conversion part (and the variable ID) is not used and can be removed

Casting a unsigned char at the top to a unsigned int and putting it in a uint32_t value seems wrong as well.

Will correct it in this PR

gizmocuz:
Found some format mistakes... for example this will crash

`fmt::format("{x}", 1234)`

we need to use

`fmt::format("{:x}", 1234)`

I think all conversions are fine now

PR #4772, opened 2021-04-14
Added CMake Stuff to allow external libfmt

xbeaudouin:
On FreeBSD there is already a port of libfmt. This patch allow linking to external libfmt linking if we want to.

Regards

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,6 +24,7 @@ option(USE_BUILTIN_JSONCPP "Use builtin JSonCPP" YES)
 option(USE_BUILTIN_MINIZIP "Use builtin Minizip" YES)
 option(USE_BUILTIN_MQTT "Use builtin Mosquitto library" YES)
 option(USE_BUILTIN_SQLITE "Use builtin sqlite library" YES)
+option(USE_BUILTIN_LIBFMT "Use builtin FMT library" YES)
 
 # Optional dependencies
 option(USE_PYTHON "Use Python for Plugins and Event-Scripts" YES)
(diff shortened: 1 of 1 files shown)


neheb:
```
CMake Error at CMakeLists.txt:572:
  Parse error.  Expected "(", got newline with text "

  ".


-- Configuring incomplete, errors occurred!
```

CMakeLists.txt
```diff
@@ -563,9 +564,23 @@ ELSE(USE_BUILTIN_SQLITE)
...
 
 # FmtLib
-add_subdirectory (extern/fmtlib EXCLUDE_FROM_ALL)
-target_link_libraries(domoticz fmt::fmt)
-INCLUDE_DIRECTORIES(${CMAKE_SOURCE_DIR}/extern/fmtlib)
+IF(USE_BUILTIN_LIBFMT)
+  add_subdirectory (extern/fmtlib EXCLUDE_FROM_ALL)
+  target_link_libraries(domoticz fmt::fmt)
+  INCLUDE_DIRECTORIES(${CMAKE_SOURCE_DIR}/extern/fmtlib)
+ELSE
```
neheb: this should be ELSE() I think.

neheb: erm ELSE(USE_BUILTIN_LIBFMT)

gizmocuz:
What is the size difference in the executable if we leave this as it is ? This because it is a extreme small library and can also be used header-only. So I expect it to be very very minimal and in that case we just leave it as it is

xbeaudouin:
As usual, the size is not the problem but if the library is already on the system package why adding again the code statically linked on a package.
Also FreeBSD does not allow git pull when building project.
By the way minizip / libz / cereal / etc... can be used as external libraries and the option is turned of per default.

CMakeLists.txt
xbeaudouin: You are right. Fixed

gizmocuz:
Thanks for the PR!

Issue #4775, opened 2021-04-14
Can we PLEASE get rid of the MQTT message on every incoming message?


rrozema:
Version: 2020.2 (build 13176)
Build Hash: 8416fa012
Compile Date: 2021-04-14 17:17:08
dzVents Version: 3.1.7
Python Version: 3.7.3 (default, Jul 25 2020, 13:03:44) [GCC 8.3.0]

![image](https://user-images.githubusercontent.com/1747982/114779894-f2caa280-9d76-11eb-89e2-24f72a0a0119.png)

2021-04-14 22:49:32.120 MQTT: Topic: domoticz/in, Message: {"command":"udevice","idx":657,"nvalue":0,"svalue":"0;0"}
2021-04-14 22:49:32.279 MQTT: Topic: domoticz/in, Message: {"command":"udevice","idx":658,"nvalue":0,"svalue":"0;0"}
2021-04-14 22:49:32.432 MQTT: Topic: domoticz/in, Message: {"command":"udevice","idx":659,"nvalue":0,"svalue":"0;0"}
2021-04-14 22:49:32.600 MQTT: Topic: domoticz/in, Message: {"command":"udevice","idx":660,"nvalue":0,"svalue":"0;0"}
2021-04-14 22:49:32.734 MQTT: Topic: domoticz/in, Message: {"command":"udevice","idx":661,"nvalue":0,"svalue":"0;0"}
2021-04-14 22:53:31.142 MQTT: Topic: domoticz/in, Message: {"command":"udevice","idx":649,"nvalue":0,"svalue":"0;0"}
2021-04-14 22:53:31.276 MQTT: Topic: domoticz/in, Message: {"command":"udevice","idx":650,"nvalue":0,"svalue":"0;0"}
2021-04-14 22:53:31.409 MQTT: Topic: domoticz/in, Message: {"command":"udevice","idx":651,"nvalue":0,"svalue":"0;0"}
2021-04-14 22:53:31.542 MQTT: Topic: domoticz/in, Message: {"command":"udevice","idx":652,"nvalue":0,"svalue":"0;0"}
2021-04-14 22:53:31.686 MQTT: Topic: domoticz/in, Message: {"command":"udevice","idx":653,"nvalue":0,"svalue":"0;0"}
2021-04-14 22:53:31.824 MQTT: Topic: domoticz/in, Message: {"command":"udevice","idx":654,"nvalue":0,"svalue":"0;0"}
[...]

PR #4776, opened 2021-04-14
MQTT use new Log method to allow user controlled logging

rwaaren:
Will satisfy request as per issue https://github.com/domoticz/domoticz/issues/4775

#4775


gizmocuz:
Thanks for this PR! I will also check this for other hardware classes

PR #4778, opened 2021-04-15
[DEV] - fixe crash with log message

ghost:
Added security on bloc list name passed in Log message to avoid crash if index is out of list size (possible if wrong Can frame is received).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,4 +50,3 @@ sqlite3_config.h
 /extern/
 /node_modules/
 /domoticz.iml
-/extern/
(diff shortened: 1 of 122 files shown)


gizmocuz:
Sorry I can not see the changes because you have reformatted all whitespaces, or have started not with the latest source code (previous)
I have made already a commit that should solve the issue and new beta's are available, could you check this ?

ghost:
I've check the 2 files an resolve problem, all is fine for me.

gizmocuz:
Thank you for the feedback!

PR #4789, opened 2021-04-19
Remove libfmt dependency since it is not used anymore

xbeaudouin:
Since libfmt is not used anymore. There is no reasons (?) to add this dependency on CMakeFile ?
Regards


gizmocuz:
Thanks!!

PR #4799, opened 2021-04-23
Logging: only log updateDevice from LUA message during debug

rwaaren:
See forum topic  [HTTP/HTTPS poller - Logs ](https://www.domoticz.com/forum/viewtopic.php?f=61&t=35959)


gizmocuz:
Thanks !!

Issue #4174, opened 2020-05-22
Checkauthtoken session id not found


heggink:
Many users have a bunch of these errors in the log again. Any idea what's causing this?

DAVIZINH0:
Hello,
I have this problem with a lot of:
CheckAuthToken(xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx) : session id not found

I use domoticz in docker (version: V.411786) (last versión available for linuxserver docker)
App for mobile is Android paid version. Last version released (22/05)

If you want any information tell me

thanks

FireWizard52:
Hi,

See also: https://www.domoticz.com/forum/viewtopic.php?f=6&t=32680&p=248248#p248248

To many users are experiencing this issue.

I use the Domoticz Lite app on 4 devices (3 tablets and 1 phone)
All the devices are Android.

3 of them has been updated (automatically) to version 0.2.224 (7535) last evening (guess araound 22:00h) and
the other one was still on version 0.2.223 (7499).

So it wasn't the beta branch.

Remark: Mostly I saw this issues during the night, if their is not any activity from PC's or something. The PC's are off.

Regards

heggink:
I just uninstalled the android domoticz app and the problem went away so no domoticz linux problem.

kitopopo:
Hi dear friends,

The error appear when i use the domoticz pay app in android. When i use a widget appears the error at the moment. I have cleared the app cache and the frequency of the error is lower but it continues to appear . each time it appears with a different token code . We have been with this error for a long time, I think they should fix it

DAVIZINH0:
same for me kitopopo but the creator say that he cant replicate the problem :-(

kitopopo:
> same for me kitopopo but the creator say that he cant replicate the problem :-(

the creator can not replicate the problem? This error appeared more than 6 months ago and it is a constant error that many people have, it is also quite annoying, I do not understand how they cannot reproduce the problem. I have verified that it is related to the android domoticz payment application. From my point of view I do not know how they have been able to close the issue when the problem is not solved. @gimocuz know about this problem?

DAVIZINH0:
> > same for me kitopopo but the creator say that he cant replicate the problem :-(
> 
[...]

PR #4802, opened 2021-04-24
webserver Change loglevel on empty session id from error to debug

rwaaren:
See Checkauthtoken session id not found #4174

### DIFF
--- a/webserver/cWebem.cpp
+++ b/webserver/cWebem.cpp
@@ -1889,7 +1889,7 @@ namespace http {
 			WebEmStoredSession storedSession = sstore->GetSession(session.id);
 			if (storedSession.id.empty())
 			{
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thanks!!

kiddigital:
@rwaaren (and @gizmocuz ), not sure if you always want to report an error because there are a number of situations in which case the session ID is/should be empty.

In PR #4729 I did some refactoring and one of the things was to add an 'AUTH' debug level showing all authentication and authorization related messages.

When using script calls (with user/pass) for example, cookies are not passed so no sessionID will be found.

And when using JWT, the sessionID is not even needed anymore.

Once we can merge the PR (so the refactored code is in the dev branch), I am glad to help track down remaining issues like the one (if it is an issue) we try to find here.

rwaaren:
>not sure if you always want to report an error because there are a number of situations in which case the session ID is/should be empty.

@kiddigital ; I agree and that is why in this case (where the condition does not really refer to an error) log.debug is better suited.

PR #4804, opened 2021-04-26
use stat to check if a folder exists

rrozema:
path names without a '.' in them are not by definition folder paths. Use stat instead to see if the path is a folder.


gizmocuz:
Thanks for the PR... Seems other patches are done as well (no problem!)

could you change

if (tmpV.size() > 0

to

if (!tmpV.empty()

rrozema:
Yeah, sorry for that. Still having difficulties with grasping the ideas behind github. Both changes -albeit unrelated- are relevant though.

gizmocuz:
Thank you for this PR !!

PR #4808, opened 2021-04-28
(Linux)compile time warnings fix. Ensure correct integer sizes.

kiddigital:
Current Plugin code gives compile warnings because Integer sizes do not correspond.


gizmocuz:
Thank you for the PR.
Actually it has to be **%zu** ... but if it is expected that the size will not be high, I would leave it at %d, but cast the size_t variable to int like
`printf("%d", (int)myarray.size())`

kiddigital:
Reverted back to %d and now cast longints to ints

gizmocuz:
Thanks!!

PR #4812, opened 2021-05-04
OpenWebNet: added probe mode and selector status

Spugna85:
Added probe mode and selector status.
Added decoding of OWN messages (WHO 4)


gizmocuz:
Thank you !!

PR #4818, opened 2021-05-11
fix tfa drop rain amount and battery percentage

hvbommel:
The TFA 30.3233.01 rain sensor, in combination with RFXtrx433E and RFXtrx433XL, reported a factor 2.54 too low TotalRain. Also the battery level is now reported in steps of 10 from 10% to 100%. Note: For those who are running a build that does not include this fix can also correct total rain by setting the edit/multiply value to 2.54


gizmocuz:
Thanks for this PR!!

nipsgur:
Cool thanks,
for a long time I thought my TFAs 30.3233.01 RF signal to domoticz was too bad or broken since the small hardware receiver/display close to the meter show the total rain correct. But I could not find anything about a calibration problem of the TFA rain meter. Last week I just put 2.5 as a factor into domoticz out of curiosity and this week it got solved with a factor 2.54. I'm amazed!

PR #4819, opened 2021-05-12
Database: changed default journal mode to WAL (write ahead) for Windows

rwaaren:
Harmonize database pragma settings (journal mode ) between Windows and Linux type platforms.

Tested on Windows 10, Raspberry PI and Intel NUC (both Debian)

### DIFF
--- a/History.txt
+++ b/History.txt
@@ -2,6 +2,7 @@ Version 2021.xxxxx (xxxx)
 - Implemented: InfluxDB Data push, Influx 2.0 support
 - Updated: dzVents; version 3.1.8 ( https://www.domoticz.com/wiki/DzVents:_next_generation_LUA_scripting )
 - Updated: OpenZWave; Added Electric pulse sensor
+- Changed: journal mode for windows from DELETE to WAL (write ahead)
 - For a full detailed overview visit: https://www.domoticz.com/wiki/Domoticz_versions_-_Commits
 
 Version 2021.1 (April 17 2021)
--- a/History.txt
+++ b/History.txt
@@ -2,7 +2,7 @@ Version 2021.xxxxx (xxxx)
 - Implemented: InfluxDB Data push, Influx 2.0 support
 - Updated: dzVents; version 3.1.8 ( https://www.domoticz.com/wiki/DzVents:_next_generation_LUA_scripting )
 - Updated: OpenZWave; Added Electric pulse sensor
-- Changed: journal mode for windows from DELETE to WAL (write ahead)
+- Changed: default journal mode for windows now WAL (write ahead) add command line option to override
 - For a full detailed overview visit: https://www.domoticz.com/wiki/Domoticz_versions_-_Commits
 
 Version 2021.1 (April 17 2021)
--- a/dzVents/documentation/README.md
+++ b/dzVents/documentation/README.md
@@ -1222,6 +1222,7 @@ There are many switch-like devices. Not all methods are applicable for all switc
  - **levelActions**: *String*. |-separated list of url actions for selector switches.
  - **levelName**: *String*. Current level name for selector switches.
  - **levelNames**: *Table*. Table holding the level names for selector switch devices.
+ - **levelVal**: *Number*. Specific for selector switches.
  - **maxDimLevel**: *Number*.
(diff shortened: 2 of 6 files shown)


gizmocuz:
Changes looks good, but I assume you have read the pros and cons and SQLITE_BUSY at this page:

https://sqlite.org/wal.html

rwaaren:
> Changes looks good, but I assume you have read the pros and cons and SQLITE_BUSY at this page:
> 
> https://sqlite.org/wal.html


Yes, i did and from what I understand the advantages outweigh the disadvantages Maybe for some  custom operating systems without support for shared-memory primitives it could be an issue if / when compiled "as is"

Happy to change this PR and try making journal-mode an additional command line option with a default setting of WAL. Would that be a preferred approach?

gizmocuz:
It's more like using a database on a shared network resource that could cause issues
Maybe a command line option to turn it off could be useful, and default enabled ...

rwaaren:
> Maybe a command line option to turn it off could be useful, and default enabled ...

OK will look into it and try to amend the PR.

rwaaren:
@gizmocuz While testing this PR I noticed that a call to cmdLine.GetArgumentCount("- <switch>" returns 1 if  there is 1 argument behind the switch but also when there is no argument behind it.
I expected that it would return 0 when there was no argument behind the switch. 

Is my assumption right or do I overlook something?

gizmocuz:
`While testing this PR I noticed that a call to cmdLine.GetArgumentCount("- " returns 1 if there is 1 argument behind the switch but also when there is no argument behind it.
`
I don't think so... have a look in it's code... 

`iArgumentCount = static_cast<int>((*theIterator).second.m_strings.size());`

So it should return -1 if the switch is not found, 0 if no arguments found, or higher number

rwaaren:
> So it should return -1 if the switch is not found, 0 if no arguments found, or higher number

@gizmocuz. It should but it does not seem to do it.

Here it continues with an empty database
```
./domoticz -dbase 
2021-05-14 10:21:14.479  Status: Domoticz V2021.1 (build 13256) (c)2012-2021 GizMoCuz
2021-05-14 10:21:14.479  Status: Build Hash: ab915f0a4-modified, Date: 2021-05-11 15:57:09
2021-05-14 10:21:14.479  Status: Startup Path: /opt/domoticz/
2021-05-14 10:21:14.527  Status: PluginSystem: Started, Python version '3.9.2'.
```

And here it tries to open an empty string?
```
./domoticz -f
2021-05-14 10:21:40.248  Error: Could not open Configuration file ''
```

I also checked with an extra debug line and cmdLine.GetArgumentCount does return 1 regardless if an argument is supplied to the switch or not.

gizmocuz:
Not sure what you try to archive... I will look at this PR soon...
-f ? we use longer names, and for this PR I would expect something like -dbase_disable_wal_mode as a flag maybe ?
When specified don't use wall mode ?

rwaaren:
> Not sure what you try to archive... I will look at this PR soon...
> -f ? we use longer names, and for this PR I would expect something like -dbase_disable_wal_mode as a flag maybe ?
> When specified don't use wall mode ?

I used -journal_mode as switch and WAL or DELETE as arguments (where WAL = default journal_mode) and that works without issues.
The issue is that the check on the number of arguments seems not to work as expected. 

So the code does allow something like ./domoticz -journal_mode -dbase where I would expect an error message because of the missing argument.

kiddigital:
@rwaaren , for me _journal_mode_ as an option isn't directly clear.
Wouldn't _db_journal_mode_ or _sqllite3_journal_mode_ be more clear?

rwaaren:
@kiddigital 

I guess so. I have no strong feelings on the name of the switch but as @gizmocuz suggested just a default of WAL and a switch to change it to DELETE would be even better.

gizmocuz:
Thank you for the change. There seems to be some 'extra' files in the last commit (dzVents/levelVal)

rwaaren:
> Thank you for the change. There seems to be some 'extra' files in the last commit (dzVents/levelVal)

These were already in first commit and is a minor dzVents documentation (wiki) update . If you want I will remove them from this PR and merge them separately

gizmocuz:
No problem, just checking if it's was meant to be in here ;)

PR #4820, opened 2021-05-12
RainRate fix for TFA 30.3233.01

hvbommel:
For TFA Raindrop 30.3233.01 RainRate was showing as a negative value where it should be zero, example "Rate: -0.04 mm/h". Also 2 digits after the comma where shown instead of just one, example "Rate: 1.12 mm/h". Also did some minor code changes to improve code readability without functional changes.


gizmocuz:

PR #4821, opened 2021-05-13
double to float for rain

hvbommel:
Sorry this takes 3 steps but I still noticed rate did not return to zero one hour after a rain session ended. I made changes to webserver to prevent conversion errors because of float and double mixing so I turned them all into floats. Now I tested it for a rain session of 30 minutes. An hour later RainRate indeed turned back to zero (0.0) instead of another small value (like 0.1). So I hope this is the last PR for the topic. I apologize for in inconvenience and thanks for the fast merges.

### DIFF
--- a/main/WebServer.cpp
+++ b/main/WebServer.cpp
@@ -10254,7 +10254,7 @@ namespace http {
 
 							if (!result2.empty())
 							{
-								double total_real = 0;
+								float total_real = 0;
 								float rate = 0;
 								std::vector<std::string> sd2 = result2[0];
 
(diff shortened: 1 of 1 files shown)


gizmocuz:
The current change is giving a lot of casting errors

float total_min = atof

should be

float total_min = (float)atof

But without this change, what happens when you use "%g" instead of "%.1f" ?

hvbommel:
"%g" gives a result like "Rate: 1.47 mm/h" instead of "Rate: 1.5 mm/h". "%.1f" fixed. But after that rainrate still did not return to "0.0 mm/h" after an hour. I am still not 100% sure why the value does not return to exactly 0.0. I may be beacuse of some rounding problem somewhere so that the MIN value becomes larger than the MAX rain value. It may be because the bucket size of the rain drop is .254 So maybe it is best to close the PR and I debug it further to get a better solution.  I will close it now and create a new PR when I know exactly what causes the problem. I did not notice the casting errors in my build. Hints on why the value goes negative are welcome.

hvbommel:
I think I found the real problem. It will test it and will do a new PR.

PR #4822, opened 2021-05-13
TFA 30.3233.01 RainRate fix

hvbommel:
Total rain is a multiple of 0.254 mm. This value is now rounded so that rain rate can be calculated correctly. I tested it. Rain total and rate are now reported correctly and rain rate drops back to 0.0 mmh/h after one hour as required. Note: I reverted previous change in WebServer.cpp because this is no longer need.

Due to rounding in previous PR rain-min, of an hour ago, could become a bit larger then rain-max (latest). This resulted in a sometimes negative rain rate value or non zero value after one hour. 

Note: For use of the TFA Raindrop rainsensor 30.3233.01 in combination with RFXtrx433E or RFXtrx433XL a future firmware release (likely 1044 or later) will be required.

### DIFF
--- a/main/WebServer.cpp
+++ b/main/WebServer.cpp
@@ -10281,7 +10281,7 @@ namespace http {
 
 								sprintf(szTmp, "%.1f", total_real);
 								root["result"][ii]["Rain"] = szTmp;
-								sprintf(szTmp, "%.1f", rate);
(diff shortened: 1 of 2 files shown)


hvbommel:
Will make some more tweeks to make it compatible with RTLSDR.

PR #4823, opened 2021-05-14
TFA drop RainRate

hvbommel:
Total rain is a multiple of 0.254 mm. This value is now rounded so that rain rate can be calculated correctly. I tested it. Rain total and rate are now reported correctly and rain rate drops back to 0.0 mmh/h after one hour as required. Note: I reverted previous change in WebServer.cpp because this is no longer needed.

Due to rounding in previous PR rain-min, of an hour ago, could become a bit larger then rain-max (latest). This resulted in a sometimes negative rain rate value or non zero value after one hour. Raincount has decimal 10 added to it so that the initial value of raincount starts at zero. This prevents rollover over the TotalRain counter on fresh install of the 30.3233.01.

Note: For use of the TFA Raindrop rainsensor 30.3233.01 in combination with RFXtrx433E or RFXtrx433XL a future firmware release (likely 1044 or later) will be required.

### DIFF
--- a/main/WebServer.cpp
+++ b/main/WebServer.cpp
@@ -10281,7 +10281,7 @@ namespace http {
 
 								sprintf(szTmp, "%.1f", total_real);
 								root["result"][ii]["Rain"] = szTmp;
-								sprintf(szTmp, "%.1f", rate);
+								sprintf(szTmp, "%g", rate);
 								root["result"][ii]["RainRate"] = szTmp;
 								root["result"][ii]["Data"] = sValue;
 								root["result"][ii]["HaveTimeout"] = bHaveTimeout;
--- a/main/mainworker.cpp
+++ b/main/mainworker.cpp
@@ -3043,7 +3043,8 @@ void MainWorker::decode_Rain(const CDomoticzHardwareBase* pHardware, const tRBUF
 	int Rainrate = 0;
 	float TotalRain = 0;
 	if (subType == sTypeRAIN9) {
-		TotalRain = float((pResponse->RAIN.raintotal2 * 256) + pResponse->RAIN.raintotal3) * 0.254F;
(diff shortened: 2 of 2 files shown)


gizmocuz:
Why is there a new firmware required ? Did this work before this PR (and the previous) or do you have been in contact with RFXCom about this ?

hvbommel:
I have indeed been in contact with RFXcom and have a beta version of the 1044 firmware (It is still being developed so not final yet). The new firmware makes the 30.3233.01 work correctly with RFXcom. The versions prior to 1044 say they support the 30.3233.01 but it does not work in practice. The new beta does but is not available for general public as far is if I know. But once that new firmware is available Domoticz needs the changes I made to get correct results (ie correct rain mm and mm/h). 

By the way, so far I have been using my TFA raindrop in combination with RTLSDR. RTLSDR was, and still is after this change, working correctly with 30.3233.01. These changes do (should) not affect it as they are a RAIN3 subtype RFXcom is a RAIN9. In the tests I did have both RFXtrx and RTLSDR listen to the raindrop and compared results. I also used RFXmngr to verify results and protocol data. It all looks good now.

I made the note about RFXtrx firmware so that people don't waste their time trying to make the TFS-raindrop work with older versions of the RFXtrx firmware. I hope this is not too confusing and I answered your question.

and the good news ... it rained today so got a free test data. I all looks good, both rain mm and rain mm/h.

gizmocuz:
Thanks for the heads-up ... no I just wanted to double check it.
Thanks for the PR

hvbommel:
Note: RFXtrx433E and RFXtrx433XL firmware 1044 is now available. It can be downloaded here:
http://www.rfxcom.com/WebRoot/StoreNL2/Shops/78165469/MediaGallery/Downloads
This version is recommended for use with TFA raindrop 30.3233.01

PR #4824, opened 2021-05-14
JSON Web Token support (Alternative to #4729)

kiddigital:
This PR is an alternative PR to #4729. Functionally completely the same and code wise almost.

For an extensive explanation on how this works, see the PR text of the original PR.

Difference that it does not add a submodule for the JWT-CPP library but includes it into the Domotics source-tree.

At this moment, it still uses the picojson header library also. 
As Domoticz has a different JSON library (JSONCPP) and the JWT-CPP uses traits for the JSON library, it could be possible to disable the picojson and uses the JSONCPP lib instead if the traits are defined correctly. Failed to complete that task in the first tries (not enough knowledge of the JSONCPP lib), so I (or someone else) might attempt that again later if/once this PR is merged. Maybe also someone can also try to remove the included base64 encoding/decoding lib and (re)use the Domoticz ones if that is possible. 

The original PR builds fine on Linux/Pi but not on Windows as the windows build system uses submodules differently. The JWT-CPP library is available in the VCPKG library, so adding it that way should be an option as well.

This PR should be able to build on Windows as well because all libraries are now in the main source tree.

Personally I prefer the original PR as it is more 'clean' when it comes to using a 'external code' in a way that it is easy to integrate upstream (bug/security)fixes and features. But maybe that is 'harder' on the Windows (build) platform.

_More detailed explanation (from the original PR)_
---
This PR adds support for JSON Web Tokens (JWT) as an additional mean to authorize access to Domoticz.

**General**
For more information, see the official RFC at: [https://tools.ietf.org/html/rfc7519](https://tools.ietf.org/html/rfc7519)

It is a first step in future support for OpenID Connect ([https://openid.net/connect/](https://openid.net/connect/)) which uses JWT and OAuth2 as underpinning technologies.

Also for future support of OpenAPI ([https://www.openapis.org/](https://www.openapis.org/)), having token support is mandatory when implementing OpenAPI support in Domoticz.

**How does it work (now)**

When you enable 'Website protection' in the settings, Domoticz will ask you to perform a login before you get access to the UI. Also any other requests to the Domoticz webserver, for example by scripts and/or external dashboards, apps, etc., you either need to provide authorization using Basic Authentication and/or providing a valid session cookie.

Now it is also possible to provide a JWT access token as a 'Bearer' token through the HTTP 'Authorization' header. No need anymore to keep providing a valid session cookie or passing a username/password combination with each request. Just add the JWT Token to each request via the 'Authorization' header.

_Bonus_: When using JWT Bearer tokens, no cookies are generated and have to be exchanged. Also, Domoticz does not need to generate a _UserSession_ to keep track of the cookies which saves a lot of SQL reads AND writes to the UserSession table. Nice bonus to lower the wear on for example SD-cards when using a Pi.

**What does it look like?**
A JWT token is an base64 encoded string consisting of 3 parts. Each of these parts have a function and contain information that can be leveraged. To _see_ what information is in a JWT Token, you can use [https://jwt.io/#debugger-io](https://jwt.io/#debugger-io) to take a look.

Although the information usually is readable, which is fine because it contains your information, it cannot be tampered with as it is signed. When receiving a JWT token, the signature is checked to ensure data is not corrupted.

The _payload_ of each JWT contains relevant information that is used by this PR and looks like the following: 
```
{
  "aud": "DomIssuer1",
  "exp": 1615728854,
  "iat": 1615725254,
  "iss": "Domoticz.com",
  "nbf": 1615725254,
  "sub": "tester"
}
```
Each of the _claims_ in the payload is mandatory and is also validated. 

The 'exp' (expires at), 'nbf' (not before) and 'iat' (issued at) are all Unix Timestamp values about when the token was issued and in what period it is valid.

The _Audience_ (aud) claim contains the _ClientID_ of the **Client** (~application, etc.) for which the token is given out.
The _Subject_ (sub) contains the **Resource Owner**, which is the actual user (you :) )
And the _Issuer_ is the **Authorization Server** that created the token, here it is the Domoticz authorization Realm

**How to retrieve a JWT token?**

At the moment, not a full OpenID Connect (or OAuth2) flow is implemented. A basic flow resembling the _Client Credentials Grant_ type ([https://tools.ietf.org/html/rfc6749#section-4.4](https://tools.ietf.org/html/rfc6749#section-4.4)) is available using the Domoticz normal _command_ structure. A *getjwttoken* command is added which, if called in the right way, returns a JSON response resembling an OAuth2 responds containing a new _access\_token_.

The command looks as followed:
```
https://<YOURDOMOTICZIP>/json.htm?type=command&param=getjwttoken&username=tester
```
The right way for this command is to call it using a POST request and also provide an _Authentication_ header containing a 'username/password' combination using the Basic authentication method of encoding. For example using Curl:
```
curl --request POST 'https://localhost:8443/json.htm?type=command&param=getjwttoken&username=tester' \
--header 'Authorization: Basic RG9tSXNzdWVyMTpEb21Jc3N1ZXI='
```
Here the 'username/password' are NOT the username and password of the user named 'tester', but the ClientID and ClientSecret of the **Client** (DomIssuer1) that wants the token _on behalf off_ the User (tester).

The _username_ parameter provided in the URI needs to be a valid Domoticz user 

and the result:
```
{
    "access_token": "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJhdWQiOiJEb21Jc3N1ZXIxIiwiZXhwIjoxNjE1NzM0NzA2LCJpYXQiOjE2MTU3MzExMDYsImlzcyI6IkRvbW90aWN6LmNvbSIsIm5iZiI6MTYxNTczMTEwNiwic3ViIjoidGVzdGVyIn0.m14_7PgD0yCocbZqLOwSjZDNLTS1eLv8QyGDnCY01YA",
    "expires_in": 3600,
    "status": "OK",
    "title": "getjwttoken",
    "token_type": "Bearer"
}
```

**Setting up a Client**
Before a _client_ can use JWT Tokens on behalf off a user, such a client needs to be setup.

This is done through the 'Users' configuration of Domoticz (See 'Setup/More Options/Edit Users').
To _register_ as client, create a User and give that user the *ClientID* rights, see screenshot

user-overview-clientid

The Username = ClientID and the Password = ClientSecret

**NOTE**:Actually, the generated MD5-based hash Domoticz generates before storing the password, is the ClientSecret that is used to do the actual signing and verification of the JWT's.


gizmocuz:
Thanks for this PR!
Question about 'URIGHTS_CLIENTID' ... what does the right 'ClientID' mean ? I do not understand this.
What right does it have ? Can it perform admin actions ? Or has it the same rights as a user ?
Is it 'Application' or 'Token' access maybe ?

> This PR should be able to build on Windows as well because all libraries are now in the main source tree.
It does compile, but I assume you can also test this I hope ?

What's the best way to test this ? I think I can also remove the picojson dependency but would like to test this
Do you work with postman ? or have some http url's ?

kiddigital:
> Thanks for this PR!

As always.. you are welcome.. happy to keep Domoticz moving..

> 
> Question about 'URIGHTS_CLIENTID' ... what does the right 'ClientID' mean ? I do not understand this.
> 
> What right does it have ? Can it perform admin actions ? Or has it the same rights as a user ?
> 
> Is it 'Application' or 'Token' access maybe ?
> 

Check to other PR for more details, but indeed the 'clientID' represents an Application/front-end. Not a real user.

> 
> 
> > This PR should be able to build on Windows as well because all libraries are now in the main source tree.
> 
> It does compile, but I assume you can also test this I hope ?
> 

Test it on Linux for sure, but don't have a Windows test/build machine available at the moment. Maybe I need it?

> 
> 
> What's the best way to test this ? 

See the URL's in the other PR description.

Make sure you have 'secure access' turned-on in Domoticz :)

> I think I can also remove the picojson dependency but would like to test this
> 

That would be nice.. less duplicate code and functions. I did take a peek, but I am not familiar enough yet with the whole source tree to oversee it.

> Do you work with postman ? or have some http url's ?

I use Postman (amongst others).

Easiest to test is requesting the settings (via json command). Without proper auth you should get a 401. Once you supply a valid access_token, it gives the output and a 200.

Btw.. use the added 'auth' debug level to see more details on what is going on. Especially to see why for example a token is not accepted (or found).

Once this gets merged, I will continue with the OpenID Connect (and OAuth) calls leveraging OpenAPI (see my OpenAPI branch.. not recently updated, but will do).

gizmocuz:
Is it possible you could make/share your postman collection to test this ? Or post some url's ?
(Creating the token, using the token, revoking the token)

About the 'ClientID' , looking at the code, it seems you are giving this Application 'admin' rights

`session.rights = 2;`

Is it necessary for applications to have 'admin' rights ?

Is 'Application' a better name for 'ClientID' in the web interface as it will confuse a lot of users

kiddigital:
> Is it possible you could make/share your postman collection to test this ? Or post some url's ?
> 
> (Creating the token, using the token, revoking the token)
> 
> 

Not near my dev/dev test rig at the moment (doing this via the GitHub mobile app :) ), but will do once I get the change.

Btw.. there is only 1 call at the moment to request the token. 

And no 'revoke' call. To revoke, just disable the 'Application' (ClientID).

> 
> About the 'ClientID' , looking at the code, it seems you are giving this Application 'admin' rights
> 
> 
> 
> `session.rights = 2;`
> 
> 
> 
> Is it necessary for applications to have 'admin' rights ?
> 

Hm.. the token is on behalf of a real user (the one you provide when requesting the token). The rights should be set according to the rights of that user. Have to look where in the code the rights are set accordingly.


> 
> 
> Is 'Application' a better name for 'ClientID' in the web interface as it will confuse a lot of users

ClientID is the 'name' that is used in the official RFC texts, so I thought better to stick to that.

gizmocuz:
Okey yes ClientID is indeed from the RFC.

So a 'User' can have multiple Applications connected to it. Would it therefor not be better to indicate the amount of connected applications in the user overview, and when you select the user, to see a list of all tokens, and have the option to revoke it from there as well ?

The rights are set in WebServer.cpp ~ line 1874

`if (_ah.method == "JWT")`

kiddigital:
> Okey yes ClientID is indeed from the RFC.
> 
> 
> 
> So a 'User' can have multiple Applications connected to it. Would it therefor not be better to indicate the amount of connected applications in the user overview, and when you select the user, to see a list of all tokens, and have the option to revoke it from there as well ?
> 
> 

A ClientID is on a 'per client/application' bases, meaning that a single Client just needs 1 ClientID for all users.

Example, say I develop an (iOS or website) app called 'MySmartHouse'. This app needs to communicate with the users Domoticz instances. So the user has to 'configure' this app to access Domoticz on behalf of the real user.

As developer of the app I would generate a ClientID and ClientSecret (which I use in my app) and tell the users to configure this in their Domoticz instance.

Now the app can request an access_token from the Domoticz instance on behalf of the user.

In theory, instead of letting users create/manage ClientIDs, we (=Domoticz) could manage ClientIDs centrally. Example, for Dashticz or MyDomoticz, etc. we could create ClientIDs (and secrets) and let Domoticz retrieve them from a secure location so no need for users to configure this themselves.

(This is how most applications integrate with others, both dev sides know each other and create this 'back-end' link. Requesting an access_token is normally also an action that takes place between both back-end systems and do not go through the end-users app/UI)

Users can just uses these apps and only have to provide their username as the access_token is on their behalf.


> 
> The rights are set in WebServer.cpp ~ line 1874
> 
> 
> 
> `if (_ah.method == "JWT")`

I will have a look. Quick look shows that it does not set the rights associated to the user at that point. It does for the name...

kiddigital:
@gizmocuz , thx for pointing out that the userrights were not correctly transferred from the user settings. User always got admin which is offcourse not as it should be.

Also thank you for removing the dependency on the include base64 class. Signing of new JWT's didn't work anymore but that was easy fix :)

Only point to resolve is that JWT base64 en/decoding expects base64url en/decoding, so not use '+' and '/', but '-' and '_'. Also no padding (adding trailing '=').

For now it works (as we both generate (encode) and interpret (decode) the JWT), but when using other tools, they will probably complain and Domoticz won't accept (valid) JWT's that use base64url encoding. I will see if I can work on that...

kiddigital:
Extended Base64 functions to support Base64url en- & decoding. (Probably room for both code- and speed optimizations, but fine for now I think.)

JWT uses this.

And added KID in header and perform checking to prevent algorithm/key abuse.

And rebased :)

gizmocuz:
Thanks for the changes... If it is possible, please revert the name changes in the base64 cpp file. (I think you are using a editor that is not UTF8/Unicode compatible)
I recommend Visual Studio! ;)

The only issue I have is with the label ClientID in the user interface. That will not say anything to normal users.
Can you change this to Application ?

It could be interesting to store the issuer as well and display this in the web gui ?

kiddigital:
@gizmocuz , done.. weird.. using Visual Studio Code which nicely says the file is UTF-8 encoded... not sure why it changed these characters, but now they are back :)

gizmocuz:
Thanks! ;) ... Best to use Visual Studio ;)

gizmocuz:
@kiddigital , there is a "issuer" field in JWT, that seams to be used to identify the application that is requesting a JWT token.
It could be useful to store this
Could you provide some URL's for testing this ? (Request a token, using the token, deleting the token)

kiddigital:
Hi @gizmocuz , the 'issuer' is the field that tells which IAM (Identity and Access Management) provider has generated the JWT. Currently that is Domoticz as well, but it could also be some other system (for example a Microsoft Active Directory with Federation Services (ADFS) which support OAuth2, or one of the other (Cloud) Identity services/servers)

I will send some test URL's asap.

kiddigital:
_From the original PR description_

(This PR does call _ClientID_ different in the UI, it is called _Application_)

Example to call something with a JWT:
```
curl --location --request GET 'https://localhost:8443/json.htm?type=settings' \
--header 'Authorization: Bearer <Your-JWT>'
```

**How to retrieve a JWT token?**

At the moment, not a full OpenID Connect (or OAuth2) flow is implemented. A basic flow resembling the _Client Credentials Grant_ type ([https://tools.ietf.org/html/rfc6749#section-4.4](https://tools.ietf.org/html/rfc6749#section-4.4)) is available using the Domoticz normal _command_ structure. A *getjwttoken* command is added which, if called in the right way, returns a JSON response resembling an OAuth2 responds containing a new _access\_token_.

The command looks as followed:
```
https://<YOURDOMOTICZIP>/json.htm?type=command&param=getjwttoken&username=tester
```
The right way for this command is to call it using a POST request and also provide an _Authentication_ header containing a 'username/password' combination using the Basic authentication method of encoding. For example using Curl:
```
curl --request POST 'https://localhost:8443/json.htm?type=command&param=getjwttoken&username=tester' \
--header 'Authorization: Basic RG9tSXNzdWVyMTpEb21Jc3N1ZXI='
```
Here the 'username/password' are NOT the username and password of the user named 'tester', but the ClientID and ClientSecret of the **Client** (DomIssuer1) that wants the token _on behalf off_ the User (tester).

The _username_ parameter provided in the URI needs to be a valid Domoticz user 

and the result:
```
{
    "access_token": "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJhdWQiOiJEb21Jc3N1ZXIxIiwiZXhwIjoxNjE1NzM0NzA2LCJpYXQiOjE2MTU3MzExMDYsImlzcyI6IkRvbW90aWN6LmNvbSIsIm5iZiI6MTYxNTczMTEwNiwic3ViIjoidGVzdGVyIn0.m14_7PgD0yCocbZqLOwSjZDNLTS1eLv8QyGDnCY01YA",
    "expires_in": 3600,
    "status": "OK",
    "title": "getjwttoken",
    "token_type": "Bearer"
}
```

**Setting up a Client**
Before a _client_ can use JWT Tokens on behalf off a user, such a client needs to be setup.

This is done through the 'Users' configuration of Domoticz (See 'Setup/More Options/Edit Users').
To _register_ as client, create a User and give that user the *ClientID* rights, see screenshot

user-overview-clientid

The Username = ClientID and the Password = ClientSecret

**NOTE**:Actually, the generated MD5-based hash Domoticz generates before storing the password, is the ClientSecret that is used to do the actual signing and verification of the JWT's.

kiddigital:
@gizmocuz , you can use the online [JWT Debugger](https://jwt.io/#debugger-io) to decode a JWT and even generate your own. Helps with testing and validating that the Token contains the right information.

gizmocuz:
I'm not sure how to debug with this.... I need some Domoticz URL's to test against, and also eliminate the new JSON dependency

kiddigital:
@gizmocuz , here is my 'testplan'
1) Run Domoticz with Auth debugging enabled (and/or any other level you prefer/need)
```
sudo ./bin/domoticz -sslwww 8443 -wwwroot www -loglevel normal,status,debug,error -debuglevel auth
```
2) Make sure Domoticz is configured with security active
- Website protection: Put in a User/Pass and select 'Login page' as Authentication
  (Remember: the User provided here does NOT show up in de User list under 'Setup/Edit Users')
- Make sure that 'Local networks' is empty (so no access without valid credentials even from local URLs)

3) Test 1; request an JSON command URL (see https://www.domoticz.com/wiki/Domoticz_API/JSON_URL%27s) that requires valid authentication (either through your browser or a tool like curl or Postman):
- https://localhost:8443/json.htm?type=command&param=getforecastconfig
- https://localhost:8443/json.htm?type=command&param=getversion
- https://localhost:8443/json.htm?type=command&param=settings

The first and last one should give a (401) Unauthorized message (and return code 401) as these commands are only allowed for authorized users.

NOTE: when for example using a Browser and you have been logged in to Domoticz (even when performing a logout before), the Browser might still have a Domoticz cookie (called DMZID) that is valid and it will use that when requesting the URL's. As long as the cookie is valid (duration), Domoticz will accept the request. So please remove the cookie for testing purposes.

4) Now using the Domoticz UI, configure a test user called 'mytestuser' and give this user for example the 'Viewer' rights. This is the real user on whom behalf we want to perform actions.
- Test if you are able to login/logout with this user. It should work.

5) We need to add a 'Client' (~Application) that performs the actual actions on behalf off 'mytestuser'.
   So we add another 'User' (through the UI) called 'mdtcapp' (MyDomoticzTestClientApp) but give it the 'Application' rights.
   (And we will for now use the password 'MyMDTCappPass', but could be any secure password)

- Also try to login with this 'User'. This should NOT work (because it is not an actual real user)

*The real testing starts here*

6) Let's request a JWT from Domoticz for the 'mdtcapp' on behalf of 'mytestuser'
```
curl --location --request POST 'https://localhost:8443/json.htm?type=command&param=getjwttoken&username=mytestuser' --header 'Authorization: Basic bWR0Y2FwcDpNeU1EVENhcHBQYXNzCg=='
```
The 'tricky' part here is the (Base64) encoded Basic Authorization header. This is the \<user\>:\<password\> combination and than Base64 encoded. (Try to decode the string above and find the used user/pass).
The user/pass combination is NOT the one from 'mytestuser' but actually the combi from 'mdtcapp'!

This should return JSON resultdata (and HTTP status 200 OK) that contains an *access_token*. Copy this access_token as we need it to test out earlier step 3 requests again.

7) Now lets test the earlier requests from step 3 again
Call the same URL's but now provide the access_token from the previous step as the 'Bearer token' through the Authorization header.
```
curl --location --request GET 'https://localhost:8443/json.htm?type=settings' --header 'Authorization: Bearer <ENTER-ACCESS_TOKEN-HERE>'
```
Now the request should nicely work and return a (200 Ok) JSON response :)

8) Go to the JWT Debugger at 'https://jwt.io/#debugger-io' and paste the access_token into the box and see what the Token actually holds. You can even modify the payload data for example and it will update the token.
But ofcourse this updated token will not work (try it :)! ) as it has not been properly signed.


NOTE: the 3 test URL's are just a random choice and you can test it with any Domoticz URL.

gizmocuz:
@kiddigital , thank you for the explanation.

The first part I do not understand is why you need to create a second user for the application.
If you look in my OATH2 branch this is not needed. Normally when you request a token for the first time, you are presented with a login screen where you need to authenticate.
In this case, you can authenticate for a user with 'viewer' rights, or user rights, or admin rights...
Then you can a token back.

Imagine a user downloads a Domoticz application for it's mobile. The application asks for the username/password, and is now handshaking via the Domoticz (oauth) methods to request the (refresh) token. This is the only token the application stores.
It does not store the username/password that is being used to authenticate.

It seems in your case, the user needs to create 'application' users first.... That part should not be needed

kiddigital:
> 
> 
> @kiddigital , thank you for the explanation.
> 
> The first part I do not understand is why you need to create a second user for the application.
> If you look in my OATH2 branch this is not needed. Normally when you request a token for the first time, you are presented with a login screen where you need to authenticate.
> In this case, you can authenticate for a user with 'viewer' rights, or user rights, or admin rights...
> Then you can a token back.
> 
> Imagine a user downloads a Domoticz application for it's mobile. The application asks for the username/password, and is now handshaking via the Domoticz (oauth) methods to request the (refresh) token. This is the only token the application stores.
> It does not store the username/password that is being used to authenticate.

You are right that to be as secure as we are now (username/password) we could do without the clientid/cliensecret part. From an OAuth2 perspective this is called the 'Resource Owner Password Credentials' grant (https://datatracker.ietf.org/doc/html/rfc6749#section-1.3.3). Such a scenario is acceptable for a lot of Domoticz use-cases where Domoticz is running on an internal network (which it is most of the time) AND the clients accessing it, are also connected to that same local network.

> 
> It seems in your case, the user needs to create 'application' users first.... That part should not be needed

True in the above example. No need to create a single ClientID (Application as we have called it) for all users that use that same 'Client' (Application).

But when you want to use a different Client (Application) and that Client is NOT in the local network, you need to open up access to your Domoticz instance (in the local network) to become accessible from the outside. Usually by some port forwarding to the internal IP of Domoticz. Now it is possible to login to Domoticz with 'just' a Username/Password from all over the world :) !

But also by any 'client' just needing the user/pass combination. So domoticz is now 'open' to brute force user/password attacks (and with all these User/pass databases floating around :( )

Adding an extra layer of security is to have the 'Client' (Application) be known to Domoticz so the amount of possible Clients to access Domoticz can be limited (and so the attack risk). This can be done using the 'Authorization Code' grant (https://datatracker.ietf.org/doc/html/rfc6749#section-1.3.1).

_The implementation in this PR is taking a number of shortcuts and re-using the 'Users' table (to also hold the ClientID's which are not actual users but 'applications'). The main goal of the PR is to, next to some refactoring of the code, to add JWT tokens as an additional way next to either DMZID session cookie or passing username/password with each privileged request._

It is not a full blown OAuth2/OpenID Connect implementation. But OIDC (and OAuth2 could) requires JWT support, so that is the reason to start there.

The current 'getjwttoken' command is just a simple (not following proper specs) way to just do a login once and have a token ,similar to the DMZID cookie, that can be used within scripts, clients, etc. without the need to keep passing the cookie or user/pass all the time AND bonus benefit that there is no need to use the UserSession table all the time (saves a lot of DB activity).

1) If you want, I can make adding a 'ClientID' (Application) optional? By just passing 'none' as the _ClientSecret_ when requesting a token, we can skip the checking of the existence of the ClientID (and validate the ClientSecret). And still keep the benefit of knowing which application (ClientID) wanted to perform the request :)

2) And would you prefer not to keep track of 'ClientID/ClientSecrets' (Applications) in the User table? We could create a seperate table for that or maybe even better a local 'clientID' file, environment variables or through command line which get read during start-up (fits well with containerized setups to have these 'secrets' in a save place instead if in the container itself)?

gizmocuz:
I think it is wise to use the UserSessions table for this, maybe expand it a little.
The reason I created the OAuth2 branch is that existing third party OAuth libraries just work with it without any special tweaking.
I'm sure you know how OAuth works reading from your above replies, and it's a way cleaner approach.
If you could also embed this in this PR then app developers can choose what they want to use. But I hope they stick to OAuth2

That you need to open firewalls/ports , thats still something you have to do, otherwise none can reach the API

The whole reason for a 'refresh' token is to not use a username/password, and this is only required when you register the application for the very first time, or when you force it to logout (revoke the token)

Me myself still prefer to use a VPN (wireguard) connection and then use any app or web interface to my local network from a unknown location

kiddigital:
Closed in favor of PR #4872

PR #4829, opened 2021-05-19
Correction of VLD telegrams management

efdeb:
New tests I made show thas the previous pull request I proposed yesterday for VLD messages management is not sufficient.
Sorry!

This one is a better proposal considering that :
- VLD message m_buffer[1] does not contain EEP func, nor type, so lines 2078/79/80 are erroneous
- Known VLD Nodes are loaded in m_VLDNodes table, so no need to request the database
- Sender ID location in VLD message is always located in m_buffer[m_DataSize - 5] to  m_buffer[m_DataSize - 2]
...


gizmocuz:
No worries, thanks for correcting this!!

Issue #4647, opened 2021-02-17
Presentation of numbers in the web ui


SargonofAssyria:
There is small bug in the presentation of numbers in the web ui.
![2021-02-17_09-37-39](https://user-images.githubusercontent.com/24824113/108177874-0ef9ec80-7104-11eb-957b-09ad47a61adb.png)
It should be presented as:    Lux:  4.582 lx   or   Lux: 4,582 lx depending on what decimal system you use.
This is present in more devices if not all.

gizmocuz:
I think this has to be fixed in the initializer of HighCharts. I'm not aware we say which decimal system to use, so it should work automagically I suppose ?

PR #4830, opened 2021-05-19
Charts: added compare charts

michahagg:
# Pull Request changes

I have been working on compare charts and would like to receive comments on how to improve on them.

## Compare chart
A new chart is added for devices of type Counter (Gas, Water, Counter, P1, Instant-And-Counter). It presents usage data in a way
that it makes it easy to compare different years, quarters and months.

## Charts behavior
* Added ability to delete datapoints on touch device. (Double-tap, hold a second and release.)

## Charts appearance
* Don't show minimum-length columns for a value of 0, except when the actual value is shown above the column.
* Show vertical grid lines.
* Show a zoomed-trendline in addition to the overall trendline.
* Cycle through min and max settings on y-axes. (Click on the y-axis label.)

## Charts fixes
* (#4647) Attempt to use browser setting to format numbers using thousand-separators and decimal-separator.

## General appearance
* Show back-button aligned with next-button in theme 'elemental'
* On Hardware page, properly show device's bullet items in theme 'elemental'

## Configuration
* Configure bind-address for http and https ports in configuration file. Use 'http_address' and 'ssl_address' in addition to the existing 'http_port' and 'ssl_port'.

## Logging
* Show status message when binding on configured ip-address fails and :: or 0.0.0.0 is used instead.

## Updates
* jQuery: 3.6.0
* Highcharts: 9.1.0


michahagg:
Hi @gizmocuz, I have finally been able to find the time and energy to create the 'compare charts'. The P1 and Gas data you sent before have proven to be very useful in testing! :-D

Please let me know what you think of it.

CMakeLists.txt
```diff
@@ -720,7 +720,7 @@ ELSE()
   MESSAGE(STATUS "==== OpenZWave not found, support disabled!")
 ENDIF(OpenZWave)
 
-IF(EXISTS /sys/class/gpio)
+IF(EXISTS /sys/class/gpio OR WITH_GPIO)
```
gizmocuz: Question, if someone uses -DWITH_GPIO=YES , it is already added to the definitions ? It seems with this change it is added twice ?

gizmocuz:
Great work! Can't wait to test this soon!!!
Do a see a patched HighCharts ? Is this really needed? It's easier to just update highcharts with a new version when it's available.
The map file dependency can be removed by removing the last line in the .js files that are referring to it... Always disliked them.. ;)

CMakeLists.txt
michahagg: You're right. This is quite confusing. I've changed it to 'OR FORCE_WITH_GPIO'.

michahagg:
The patch is indeed quite annoying and a little hassle to apply. Even though it's only two changes. There is an issue that addresses the weird behavior: highcharts/highcharts#9612 and it has gotten some attention from someone from Highcharts. Although a final solution is yet to be provided. For now, I just apply the patch.

michahagg:
The .map files are very handy when it comes to debugging or just viewing the javascript source. The minified files are used when running the application. When you open the Developer Tools in the browser and start viewing the source, that is when the .map files are downloaded and are used to provide the full and comprehensible source code.

CMakeLists.txt
gizmocuz: Thanks, but not sure why this is needed. Even if I compile on a Linux system /sys/class/gpio exists. Why do you want to force this ? Are you compiling a Arm/Linux build on a Windows/Mac system ?

gizmocuz:
Thanks for the heads-up with the map files.. Yes that indeed provides the original source files... did not know this ;)
Nice for when you're developing

gizmocuz:
Windows build system has a issue building this. (boost/format), please don't use this, if possible use the domoticz helper function "std_format"

main/WebServer.cpp
```diff
@@ -16629,1342 +16656,1517 @@ namespace http {
...
+        void CWebServer::GroupBy(
+                Json::Value &root,
+                std::string dbasetable,
+                uint64_t idx,
+                std::string sgroupby,
+                std::function<std::string (std::string)> counterExpr,
+                std::function<std::string (std::string)> valueExpr,
+                std::function<std::string (double)> sumToResult
+        ) {
+            std::string query = boost::str(boost::format(std::string("")
```
gizmocuz: This could be a normal std::string, but the safe_query function also accepts parameters and is much clearer

michahagg: I'm cross-compiling on WSL2, in a Debian distro. It works very well, but I guess some things are not easy to emulate. I'm cross-compiling for a RPi3, so I do need to have the GPIO code in the executable.

michahagg: Thanks for pointing it out, the function std_format! I hadn't seen it before. I made a similar function and called it StdFormat ;-) Yours is much better so I'm using it, of course. For the boost::format, I made an alternative that replaces positional placeholders; I find it much better to have the arguments just passed once and be able to refer to them by index. It's something the Java-equivalent of printf can do, but unfortunately in C++ this was never implemented.

gizmocuz:
@michahagg , thank you for the changes.
I don't think the ReplacePlaceholders function is needed and can be done with StdFormat, and the last usage as mentioned before is not needed as you can can call m_sql.safe_query with arguments

m_sql.safe_query(
" some string "
" some string on another line ",
" ...",
arg1, arg2 etc

main/WebServer.cpp
gizmocuz: I am glad it's not in c++ ;) ... It's clearer to use the safe_query function
Is it possible to use 'std_format' instead of StdFormat, No need for new functions, will also make the application bigger and could confuse feature developers... which function to use...

michahagg:
@gizmocuz , I do see what you mean. Of course, safe_query uses a printf-method to format the query using placeholders. It's very convenient that way. There is only one issue that I don't like about it, it does not allow one argument to be used multiple times in the format. The tablename is referred to a couple of times in my query, and with safe_query, for each reference, it must be added as an argument. This is quite tedious and error-prone.
BTW I am using safe_query to substitute the device index in the query 3 times. The 'idx'-argument must be present 3 times.

Do you think there is a way in the middle? A way where the query is built using safe_query, but where arguments can be referred to by index?

gizmocuz:
I don't really see the difference by using a reference or directly using it the normal way as arguments...
It is also error sensitive if you reference a wrong index i presume ?
And a new function StdFormat instead of the existing std_format is a bit double work...

gizmocuz:
Thank you for this!!

michahagg:
> Thank you for this!!

In the end this was the better solution! Thank you!

gizmocuz:
@michahagg , chart is working great!
If I may have one critical opinion on it... it is the colors... they look 'dull' ... soft ... not sprinkling/shining
Plz let me know when you want this to be merged!

michahagg:
image

I hope you like these colors better!

If you do, it's **ready to be merged**! Thanks!

gizmocuz:
Love the colors! But do you need to set them explicitly ? what happens when you let highcharts decide the colors ?
(Happy to merge bdw.. but just wondering!)

michahagg:
I understand why you ask. Normally it would be ok to set a bunch of colours and let Highcharts decide which ones to use. What I wanted to achieve is that each year can be recognized by the colour in every situation. That requires that, no matter what set of years is represented in the chart, each year is assigned the same colour everytime.
Another reason is that in the current decade (2020-2029), every year has a different hue. But in the previous decade (2010-2019), the corresponding year has the same hue but a different saturation and in the following decade (2030-2039) the corresponding year has the same hue but different value (luminance).

Hope that makes it clear.

BTW If you (or anyone) have another idea for the colour palette, please let me know!

gizmocuz:
Perfect then ;)
One question, Do you think it is possible in the popup to have a quick indication on the right side with a arrow if it went up or down ? (or stable) , like in the report overview ?
That would make it perfect.... could also be another PR later of course

michahagg:
Yeah sure. That should be possible. Let me find out how that's gonna work.

michahagg:
image

I think you mean something like this. It compares each month/year value with the value before it, chronologically. So in the screenshot it shows a down-arrow for value Aug/2017, because value Jul/2017 was higher. For January/\<year\> it would compare the value with the value of Dec/\<year-1\>.
Right now it compares the value with the nearest value before it. If for some reason, for a month, the value of the previous month is missing, it will compare with the value of a _more previous_ month.

dannybloe:
Well, regarding color choices. I am color blind and always have problems with charts where a line or area is only defined by a color. So, the color palette could be chosen such that they work with most types of color blindness. Or better, make a setting for the current user where you can change the palette. Or, that when you hover over a legend item you can see the corresponding line/bar stand out in some way (border or such).

michahagg:
Hi @dannybloe, you are very right! Your suggestions could be easily made.
For example, when using a mouse, when you hover over a legend-item, Highcharts highlights that item in the chart (by dimming all the columns in the other series.) Does that already work for you? (I'm not sure if that currently works when using a touch device, but that could be made to work as well.)

gizmocuz:
@michahagg , WoW this looks great !!! Let me know when you think we should merge this !

michahagg:
I'm glad you like it, @gizmocuz ! It's ready to merge. If I want to add more, I'll just make a new PR.

gizmocuz:
Thank you both for this great PR !!!! Looking great!

rwaaren:
@michahagg, Looks great. 

Can I please ask you for the favor of updating the relevant chapters of 
[Domoticz API / JSON wiki page](https://www.domoticz.com/wiki/Domoticz_API/JSON_URL%27s)?

albert607:
After adding an energy pulse counter yesterday I detected that loading the log page take a long time. Probably due to the newly added comparing graph which looks at follows in my setup (Domoticz build 13360):  
image

Any suggestions? 

ps. I personally prefer line graphs over bar graphs which are showing on my setup in the year views nowadays.

dobber81:
I got the same issue as above when adding a new energy pulse counter. Takes forever to load log page and looks just like image above.

gizmocuz:
So it might be best to revert this PR ?

dobber81:
After 3 or 4 days new graphs are ok. I like the graph when it's working, but it's a bit frustrating to wait for 4 days before it works properly.

albert607:
Personally I discontinued the monitoring of the S0 counter so I cannot verify at my side. But if after a few days the graph is showing properly without long waiting times then I would recommend to not revert this PR. 
Perhaps a solution for the first few days might be worthfull to investigate (perhaps force a line or bar graph over time  is possible?)

gizmocuz:
@michahagg , Is it possible to have a look at the above issue ? It is very easy to replicate.
Add a dummy electric (instant+counter) device, wait till 5 minute have passed then go to the log
You will see the screen from Albert and it takes about 60 seconds to load the page
Also it seems the top chart is not functioning anymore
Hope you can find the issue

PR #4838, opened 2021-05-22
Favorites; Fix for loosing user's favorites when shared device is added

rwaaren:
Reported on the forum in post [User loosing favorites when adding device](https://www.domoticz.com/forum/viewtopic.php?p=275370#p275370)


gizmocuz:
Thank you for this PR!!

PR #4846, opened 2021-05-30
RTL-SDR TFA Raindrop ID

hvbommel:
For RTL-SDR only the lower 8 bits of the TFA Raindrop SensorID 30.3233.01 where used. This may lead to conflicts when using multiple Raindrop devices. 
Note: After this change the device needs to be added again. After that replace the old with the new device to keep rain history.


hvbommel:
need to do a empty string test for the model field. Will do the 1st and do a new PR.

PR #4847, opened 2021-05-31
RTL-SDR TFA Raindrop DeviceID

hvbommel:
1. For RTL-SDR only the lower 8 bits of the TFA Raindrop SensorID 30.3233.01 where used but the actual DeviceID contains 20 bits (2 and a half bytes). This may lead to conflicts when using multiple Raindrop devices. This has been fixed by this change. Device ID now also matches logging of well known program rtl_433. Example: previous device id 0x7d (8bits) now shows up as 0xe3a7d (20bits).

Note: After this change, for TFA Raindrop 30.3233.01 devices connected to RTL-SDR, the device needs to be added again to Domoticz because its device id changes. After adding the new device replace the old device with the new one to keep rain history.

2. I studied the TFA message a bit more and noticed the TFA Raindrop battery states are "ok" and "low", not percentage. Changed code for RFXtrx to report these 2 states instead of percentage so it more accurate reflects actual battery state as reported by the Raindrop sensor firmware.

### DIFF
--- a/hardware/Rtl433.cpp
+++ b/hardware/Rtl433.cpp
@@ -343,6 +343,9 @@ bool CRtl433::ParseData(std::map<std::string, std::string>& data)
 
 
 	unsigned int sensoridx = (id & 0xff) | ((channel & 0xff) << 8);
+	if (!model.empty() && (!strcmp(model.c_str(), "TFA-Drop"))) {
+		sensoridx = id;
+	}
 
 	if (haveTemp && haveHumidity)
 	{
--- a/hardware/Rtl433.cpp
+++ b/hardware/Rtl433.cpp
@@ -343,7 +343,7 @@ bool CRtl433::ParseData(std::map<std::string, std::string>& data)
 
 
 	unsigned int sensoridx = (id & 0xff) | ((channel & 0xff) << 8);
-	if (!model.empty() && (!strcmp(model.c_str(), "TFA-Drop"))) {
+	if (model == "TFA-Drop") {
 		sensoridx = id;
 	}
 
--- a/hardware/Rtl433.cpp
+++ b/hardware/Rtl433.cpp
@@ -343,6 +343,9 @@ bool CRtl433::ParseData(std::map<std::string, std::string>& data)
 
 
 	unsigned int sensoridx = (id & 0xff) | ((channel & 0xff) << 8);
+	if (!strcmp(model.c_str(), "TFA-Drop")) {
+		sensoridx = id;
+	}
 
 	if (haveTemp && haveHumidity)
 	{
--- a/hardware/Rtl433.cpp
+++ b/hardware/Rtl433.cpp
@@ -343,7 +343,7 @@ bool CRtl433::ParseData(std::map<std::string, std::string>& data)
 
 
 	unsigned int sensoridx = (id & 0xff) | ((channel & 0xff) << 8);
-	if (!strcmp(model.c_str(), "TFA-Drop")) {
+	if (!model.empty() && (!strcmp(model.c_str(), "TFA-Drop"))) {
 		sensoridx = id;
 	}
 
--- a/main/RFXtrx.h
+++ b/main/RFXtrx.h
@@ -186,7 +186,7 @@ SDK version 7.01
 	SelectPlus200689103 Black Chime added
 
 SDK version 7.00
-	TEMP7 - TSS330 added and TH9 - TSS320 added
+	TEMP7 - TSS330 added and TH9 – TSS320 added
 	BlindsT8 = Chamberlain CS4330CN added
 	SelectPlus200689101 White Chime added
 	Interface command - start receiver added
--- a/main/RFXtrx.h
+++ b/main/RFXtrx.h
@@ -186,7 +186,7 @@ SDK version 7.01
 	SelectPlus200689103 Black Chime added
 
 SDK version 7.00
-	TEMP7 - TSS330 added and TH9 � TSS320 added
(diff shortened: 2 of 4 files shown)


gizmocuz:
And you have multiple rain devices with almost the same id ? The odds are very very small (1%)

hvbommel:
I recently bought my 3th 30.3233.01 TFA Raindrop and ran into the problem of duplicate device id's. So I think TFA must have extended the device id to even more than 2 bytes (ie 20 bits) for a reason. I my case, only using the least significant byte of the 20 bit device id, caused caused the duplication.

Also note that RFXtrx (not rtl-sdr) is only using 16 bits of the 20 because and they will leave it like that because of the API standard. That makes sense of course, I would do the same. Besides that there is a currently little problem in the latest 1044 firmware. It uses effectively 12 bits instead of the 16 bit id field because the most significant nibble is currently always "3". That will be fixed in the next firmware release and they will use the 2 lower bytes of the 20 bits as the device id. But that will be in a fw release later then 1044.

Note: in the RAIN structure byte id0 is added to the very end so that the structure is still compatible with RFXtrx433. RFXtrx433 does not use the new id0 byte. RTL-SDR does use the id0 byte for bit16 to bit19 device id storage.

I have tested both RFXtrx433/TFA-Raindrop and RTL-SDR/TFA-Raindrop. All rain sensors work fine for 2 days now.

PS: If you prefer a single commit branch just let me know and I will create it for you and do a new PR. (I did a rebase but that ended up a bit messy, no file content has changed)

hardware/Rtl433.cpp
```diff
@@ -343,6 +343,9 @@ bool CRtl433::ParseData(std::map<std::string, std::string>& data)
 
 
 	unsigned int sensoridx = (id & 0xff) | ((channel & 0xff) << 8);
+	if (!model.empty() && (!strcmp(model.c_str(), "TFA-Drop"))) {
```
gizmocuz: model is a std::string, please do not use the old 'c' method strcmp to search/compare the string.
use == or one of the find functions please

main/RFXtrx.h
```diff
@@ -186,7 +186,7 @@ SDK version 7.01
 	SelectPlus200689103 Black Chime added
 
 SDK version 7.00
-	TEMP7 - TSS330 added and TH9 � TSS320 added
+	TEMP7 - TSS330 added and TH9 - TSS320 added
```
gizmocuz, line 189: Please do not modify this file, it is a external file and copyrighted

main/RFXtrx.h
```diff
@@ -2242,6 +2242,7 @@ typedef union tRBUF {
 		BYTE	battery_level : 4;
 		BYTE	rssi : 4;
 #endif
+		BYTE	id0;
```
gizmocuz: Please do not modify this file. It is read-only and copyrighted

hvbommel: Ok, that is a bit more work. I will work on it. The first 2 reviews are now covered in my commit 5 minutes ago.

gizmocuz: Thank you. Still the odds that you have a device that have the same 16 bits is very slim.

hvbommel: I agree, but lets make it right for once and forever.

gizmocuz: I don't share the same idea... 
- It is working currently for everyone, and no one will have multiple rain devices, and if they do, the chance that it has the same 16 bits is the same as winning the big lottery ticket
- You are trying to squeeze in bits that may be there in RTL-433, but not on other RF hardware that is mainly used inside the application
- For every received device (RTL433) a extra string compare is performed (slowing down the application, not by much, but still)

If the battery is being correctly with the below, then that would be a great PR

`decode_BateryLevel(subType == sTypeRAIN1, pResponse->RAIN.battery_level & 0x0F);`

hvbommel: So I propose to leave the structure alone and use the 16 bits of the 20 instead of just 8. That will avoid the conflict I noticed with minimal change. By the way I also prefer RFXtrx but for the rain sensor it is current less stable. Devices are lost and I dont know the reason but rtl-sdr works all the time without loosing devices. That is why I use rtl-sdr and run rfxtrx in parrallel to help RFXcom debug/test. By the way I have 5 rain sensors. Two for master rain to determine if I need to rain the garden and 3 more to monitor the groups of the rain system.

hvbommel:
I will close this one and start with a fresh clone and then do a new PR. I dont understand the build failures. Also the "TEMP7 - TSS330 added and TH9 – TSS320 added" Is strange as I dont see any change. Copy it for the master branch did not help.

main/RFXtrx.h
gizmocuz: We are using RFXCom RFXtrx433 (XL) for many many years without any issues... Maybe you enabled to many protocols. Some protocols have influence on others.
It is also a transceiver instead of only a receiver.

You could have rebased this PR , but I can also squash it into a single commit.

Okey so the only change will be the battery ?

hvbommel: I agree the RFXtrx433E and XL are great devices. They work fine in most cases. It is just for the TFA rain sensor where the SDR works better. I run the RFXtrx and the rtlsdr in parrallel for debugging. For the first 2 days or so both the RFXtrx and RTL-SDR reports all 3 TFS Raindrops perfect. At some pint the RFXtrx stops reporting the Raindrop while the RTL-SDR continues as expected. I have no clue what it is. For other devices the RFXtrx433 continues to function normal. That is why I currently use the RTL-SDR with Raindrop on my active system. Thats all

PR #4850, opened 2021-06-03
RTL-SDR TFA Raindrop 16bit device id, ignore 20:16

hvbommel:
For more comments see closed PR #4847. Simplified a lot.
- 16bit device id for TFA Raindrop connected to RTL-SDR only to prevent duplicate id's
- Battery state for TFA Raindrop 30.3233.01 connected to both RFXtrx433 and RTL-SDR


gizmocuz:
Thank you for this PR!!

Issue #4844, opened 2021-05-30
Travis-ci.org moves to Travis-ci.com


michahagg:
Hi @gizmocuz I just clicked on the build status icon at the bottom of the main domoticz GitHub page. It went to Travis-ci.org that displayed a warning it will be made unavailable end of May. That is tomorrow! Maybe it's nothing, maybe the build checker won't work after tomorrow 🤨

PR #4851, opened 2021-06-03
Fix for empty P1 year and month charts + Travis-CI.ORG moves to Travis-CI.COM

michahagg:
# Contents

## Fix for issue with P1 charts
This PR includes a fix for the problem described in issue #4852.
For the P1-compare-chart, it's necessary that a proper sensorarea is provided from the app to WebServer. When it's missing, an error message is written to the log. Incorrectly, it expected a sensorarea for the other chart-types as well (month and year).

## Fix for Travis-CI urls
This PR resolves issue #4844.
The domain name travis-ci.org is not being continued. The replacement travis-ci.com should be used from now on. The urls in the README are corrected accordingly.


gizmocuz:
Thank you for this PR!!

PR #4853, opened 2021-06-04
Add CEnOceanEEP : base class of CEnOceanESP2 & CEnOceanESP3 to provide shared enums and helper functions

efdeb:
Add CEnOceanEEP : base class of CEnOceanESP2 & CEnOceanESP3 to provide shared enums and helper functions.

- create enums for :
    ENOCEAN_MANUFACTURER
    ESP3_RORG_TYPE

- add helper functions to get text label (short desc) and description (long desc) from enum values
    GetManufacturerLabel & EnOceanGetManufacturerName
    GetRORGLabel & GetRORGDescription
    GetProfile & EnOceanGetProfileTypeLabel & EnOceanGetProfileTypeDescription

- add helper functions to deal with node IDs
    GetINodeID (uint32_t value) & GetNodeID (string value)

- rewrite helper function GetDeviceValue

Update CEnOceanESP3

- add enums for :
    ESP3_PACKET_TYPE
    ESP3_RETURN_CODE
	ESP3_EVENT_CODE
	ESP3_COMMON_COMMAND
	ESP3_SMART_ACK_CODE
	ESP3_FUNC_RETURN_CODE

- add helper functions to CEnOceanESP3 to get text label (short desc) and description (long desc) from enum values
    GetPacketTypeLabel & GetPacketTypeDescription
    GetReturnCodeLabel & GetReturnCodeDescription
    GetEventCodeLabel & GetEventCodeDescription
    GetCommonCommandLabel & GetCommonCommandDescription
    GetSmarkAckCodeLabel & GetSmartAckCodeDescription
    GetFunctionReturnCodeLabel & GetFunctionReturnCodeDescription

Adapt EnOceanESP2.h/cpp and EnOceanESP3.h/cpp files to take these changes into account

Change m_id_base type for both protocols from unsigned long to uint32_t conforming to ESP3 specification v1.51 §2.5.10

These changes should not alter the way EnOcean ESP2/ESP3 protocols are managed by domoticz.
Otherwise, please let me know as I shall fix any issues.
I shall propose other improvements as well as bug fixes later on.


efdeb:
Need to compile and link EnOceanHelper.cpp to build domotisz.exe.
;^)

gizmocuz:
Question, all enums from the new helper header are used in the .cpp file correct?
If that is the case, could you move them all to the top of the .cpp file ?

efdeb:
To take your demand into account, I moved enums/helper functions specific to ESP3 in EnOceanESP3.cpp.

Also I changed the proposed implementation, to scope common helper functions in a CEnOceanEEP base class that CEnOceanESP2 / CEnOceanESP3 inherit.

CEnOceanEEP enums provide public identifiers/values than may be used by both ESP2 and ESP3 protocols. That's why they are in the header and not in the implementation file. Example. ELTAKO manufacturer, RORG_4BS. Please note that some values may be hardcoded in source, as 0xnn hex values instead of enum identifiers.

Corresponding string values returned by herper functions are located in implementation file, besause they are not public.

A next step could be to have all strings located in an EnOcean configuration file and loaded just in time when needed. But I don't know where to place that files for EnOcean, because it seems that Config directory only contains ZWave related files.

gizmocuz:
Thank you for the changes. Please let us know when you would want this to be merged

efdeb:
Hi,
These changes might be merged in the current state.

I am currently working on other  improvements but I think it is preferable to distinguish the pull requests in case of regressions. I will use EnOcean ESP3 and EEP 2.6.8 specifications, but I do not have each and every device at home, so testers who own these devices are welcome.

Is there any documentation available about the code used to support hardware devices (hardwaretypes.h, RFXtrx.h, ... meaning of  pTypeXXX and sTypeXXX define, adùmissible values to pass to sDecodeRXMessage, ...) ?

gizmocuz:
Thank you for this PR and your time !!!

PR #4855, opened 2021-06-05
Correction of example in DzVents Timer docummentaion

jp-keros:
Missing quote in DzVents Timer docummentaion


rwaaren (changes requested):
Please use a descriptive name for your PR to show what it fixes. Just **_Update_** does not help when searching through the stack of changes. Moreover so because you use the same name for many similar Pull Requests.

The dzVents documentation is build from source (README.md.) to target (README.wiki) using the pandoc conversion tool. 

See howto on dzVents/documentation/pandoc.md

jp-keros:
There is a quote missing at the end of the line `'xx minutes after <astroMoment>, ` so the example does not show properly.

rwaaren:
Thx for reporting.
Fixed in build 13324, commit 9f4603f2b4cb6f5b146d26b19893146c4bc5e523

PR #4857, opened 2021-06-06
Add missing .c_str() to please clang

xbeaudouin:
Small PR to fix a missing .c_str() that make a fatal compilation error with Clang on FreeBSD 13.0


gizmocuz:
Thanks!!

PR #4858, opened 2021-06-07
Fix Logger binary Debug flags

kiddigital:
This PR ensure the correct binary settings of the Debug Flags. Before unexpected debug logging could occur.

For example, when selecting debug_threadids (old value 0x30), one would also get debug_python (0x20) and debug_eventsystem (0x10). Not expected :)

Now all levels have their own bit again.


gizmocuz:
Thank you !!

Issue #4860, opened 2021-06-14
problem with Enocean device


titi234:
Hi,

From the release 2021.1 I have a problem with my Enocean device, with each check it creates a new device and not update the existing with the same ID but new IDX and unit member. Do you have a solution ?

Thanks

gizmocuz:
Did you try the beta version ?

labelette91:
hi,
there is a regression when adding the rssi level in commit SHA-1: d21a535841d5ef812132185e527f9505b5e18049
Enocean: First implementation of RSSI reporting in device list

For temperature sensor / rfx sensor / occupancy sensor , the RSI field value is used for the enOcean byte0 device addresses.

cdt

labelette91:
find the correction for testing
[EnOceanESP3.zip](https://github.com/domoticz/domoticz/files/6657710/EnOceanESP3.zip)

cdt

gizmocuz:
@labelette91 , I'm sorry we can not easily compare against a zip file.
In Domoticz the RSSI value has a range of 0-11 , and 12 = NA (Not used)

I can't see what you changed by looking into a zip file

labelette91:
[EnOceanESP3.cpp.txt](https://github.com/domoticz/domoticz/files/6660708/EnOceanESP3.cpp.txt)

titi234:
> Did you try the beta version ?

Yes and it's the same

titi234:
> find the correction for testing
> [EnOceanESP3.zip](https://github.com/domoticz/domoticz/files/6657710/EnOceanESP3.zip)
> 
> cdt

where do I put the cpp file ?

gizmocuz:
Posting a complete source file does not show us what has been changed.
Maybe post the relevant code change and the difference in the result

labelette91:
![image](https://user-images.githubusercontent.com/1288018/122582438-76935a80-d058-11eb-84cd-ab38c5b0874e.png)
Find text compare.

I have tested with simulation mean. Not with real hardware.
Copy the file to hardware in order to test.

cdt

gizmocuz:
Thank you for the changes, PR is merged

efdeb:
Hi gizmocuz,

To avoid future regressions, it would be interresting to add comments to indicate that it is a trick to use RFXSENSOR.filler and RFXSENSOR.rssi to transmit EnOcean DeviceID/UnitID to decode_RFXSensor in mainworker.cpp.

[...]

PR #4863, opened 2021-06-18
correction problem with Enocean device #4860 :

labelette91580:
there is a regression when adding the rssi level value in commit SHA-1: d21a535 : Enocean: First implementation of RSSI reporting in device list

For temperature sensor / rfx sensor / occupancy sensor , the RSI field value is used for the enOcean byte0 device addresses

revert to previous commit .


gizmocuz:
Thank you very much!!

PR #4872, opened 2021-07-06
Security updates and implementation of OAuth2 (RFC6749), PKCE extension (RFC7636) and JWT Tokens (RFC7519)

### LINKED ISSUE #5259: Potentially Vulnerable to Clickjacking
Plugin Details
Severity:Medium
ID:85582
Version:$Revision: 1.7 $
Type:remote
Family:Web Servers
Published:August 22, 2015
Modified:May 16, 2017
Risk Information
Risk Factor: Medium
CVSS v2.0 Base Score: 4.3
CVSS v2.0 Vector: CVSS2#AV:N/AC:M/Au:N/C:N/I:P/A:N
Reference Information
CWE:  [693](http://cwe.mitre.org/data/definitions/693)
MEDIUMWeb Application Potentially Vulnerable to Clickjacking
Description
The remote web server does not set an X-Frame-Options response header or a Content-Security-Policy 'frame-ancestors' response header in all content responses. This could potentially expose the site to a clickjacking or UI redress attack, in which an attacker can trick a user into clicking an area of the vulnerable page that is different than what the user perceives the page to be. This can result in a user performing fraudulent or malicious transactions.

X-Frame-Options has been proposed by Microsoft as a way to mitigate clickjacking attacks and is currently supported by all major browser vendors.

Content-Security-Policy (CSP) has been proposed by the W3C Web Application Security Working Group, with increasing support among all major browser vendors, as a way to mitigate clickjacking and other attacks. The 'frame-ancestors' policy directive restricts which sources can embed the protected resource.

Note that while the X-Frame-Options and Content-Security-Policy response headers are not the only mitigations for clickjacking, they are currently the most reliable methods that can be detected through automation. Therefore, this plugin may produce false positives if other mitigation strategies (e.g., frame-busting JavaScript) are deployed or if the page does not perform any security-sensitive transactions.
Solution
Return the X-Frame-Options or Content-Security-Policy (with the 'frame-ancestors' directive) HTTP header with the page's response.
This prevents the page's content from being rendered by another site when using the frame or iframe HTML tags.
See Also
http://www.nessus.org/u?399b1f56
https://www.owasp.org/index.php/Clickjacking_Defense_Cheat_Sheet
https://en.wikipedia.org/wiki/Clickjacking
Output
The following pages do not use a clickjacking mitigation response header and contain a clickable event :

  - http://domoticzl/
  - http://domoticz./
  - http://domoticz/secpanel/index.html
  - http://domoticzl/secpanel/index.html

kiddigital:
This PR adds and improves several/most aspects of Domoticz Security and User Management functionalities.

All security related settings have been moved to a _Security_ tab with the Settings.

Also added a SECURITY_SETUP.md file (in the root) to describe the Security and User Management options and how they work (needs to be reviewed!).

Now some OAuth2- and basic OpenID Connect (OIDC) flows are also supported:
* authorization_code
* authorization_code with PKCE
* password
* refresh_token

Only the authorization_code grant type will return a refresh_token. The password grant will not.

Registering clients can be done by registering a new application via the new Applications screen.

It does not 100% comply to all requirements, for example no need (and way) to register 1- or more redirect URI's for registered clients. So there is no checking of the redirect URI's (only that they have to be absolute and HTTPS).

A _refresh_token_ can only be used once! Even if the exchange fails, the token is invalidated (removed)

The authorization code is only valid for 60 seconds (1 minute)! So it needs to be used quickly :) (as it should)

Support for RS256 and PS256 (Asymmetric algo using public/private key pair) as Signing algorithm as well as HMAC 256, 384, 512 (Symmetric algo).

Using the _access_token_ (JWT) is done through passing it as a _Bearer_ token in the _Authorization_ header of each request. This way there is NO need to keep track of a _session_ by using a Cookie (DMZID cookie).
Scripts and other _client_ applications (like webapps as Dashticz or mobile apps) can now use these JWT tokens to interact with the Domoticz backend functions in a secure and authorized way.

Also some other improvements to the security and user management functions have been done,
- Quite some cleanup and refactoring of the old code, needed to make things more clear, readable and understandable. AND also more consistent and less code duplication.
- no need anymore to set a `<username>/<password>` under 'Settings/Website protection' first and later add other users. Just add any User to the User Management system and the 'website protection' becomes active.
- New _Applications_ management screen to register (public) client Applications.
- Previously the local network of the IP-adress Domoticz was bound to, was automagically added to the (empty?) list of _Local Networks_ (that bypass protection). This is removed as it can have unexpected behavior especially for machines with multiple interfaces. The expected local network should now be configured explicitly.
- But _Local Network_ exceptions should not really be needed anymore as by using the 'password' grant, a Token can be requested to gain access to Domoticz for the duration of the token expire time (1 hour).
- Added additional AUTH debug flag, to see all Authentication and Authorization actions. 

And a few things more...take a look at the code (and commits as they hopefully explain/clarify things as well)

**NOTE** I think this PR is as complete as it should be, but more testing, documenting and maybe code bug fixing, cleanup/refactoring might still be needed.
_BUT_ I can really need some assistance and support, comments, etc. on the current state of this PR.
Most things can not directly be tested through the UI, so you need to craft some Curl requests or use Postman for example.

Endpoints:
- /oauth2/v1/authorize (GET) : to get an authorization code
example `https://localhost:8443/oauth2/v1/authorize?state=bla&response_type=code&client_id=DomIssuer1&redirect_uri=https://localhost:8443/#Dashboard`
- /oauth2/v1/token (POST) : to exchange the authorization code for access/refresh token, or request them directly (password grant) or exchange the refresh token for new tokens
example `https://localhost:8443/oauth2/v1/token`
- /.well-known/openid-configuration (GET): for basic _OpenID Connect Discovery_ of the different options supported
example `https://localhost:8443/.well-known/openid-configuration`

afbeelding

I run domoticz with some debug flags and ONLY with SSL active (disabled HTTP server)
`sudo ./bin/domoticz -www 0 -sslwww 8443 -wwwroot www -loglevel normal,status,debug,error -debuglevel webserver,auth`

PR also superseeds PR #4824 (you can find more details there)

**TO-DO**
* Update Wiki also
* Add testcases (at least descriptions, maybe _Gherkin_ style)
* i18n updates
* ...

I will update this PR (and this PR description) when changes are made.


gizmocuz:
Thank you for all this work and the PR!!!
I don't think a user should have 'application' rights, because it also does not need this rights when logging in via the web gui/application
Every user should be able to use a application and is limited by it's own rights (user/viewer/admin)

kiddigital:
> Thank you for all this work and the PR!!!
> 
Your welcome 😁

> I don't think a user should have 'application' rights, because it also does not need this rights when logging in via the web gui/application
> 
> Every user should be able to use a application and is limited by it's own rights (user/viewer/admin)
> 
> 

So would you prefer to create an 'Application' table in the DB instead of (ab)using the User table to register client applications?

gizmocuz:
I think we should extend the session table a little bit (with a token/refresh-token, and maybe a few other fields)
It's like a normal browser session where also the user is part of the cookie

galadril:
It would be really cool to have OAuth as an option for the mobile apps to authenticate!
So how can we push this forward??

kiddigital:
I will see if I can update the documentation some more soon and maybe even some (postman collection?) testcalls.

Would really be useful if some else (besides me) also gives this branch a (quick) test to check that I am not overlooking anything or develop some blind spots ;)

kiddigital:
> 
> 
> I think we should extend the session table a little bit (with a token/refresh-token, and maybe a few other fields)
> It's like a normal browser session where also the user is part of the cookie

I added a separate __Applications__ screen and menu option where one can manage _Applications_ (Clients)

This makes the _creative use_ of _Users with ClientID rights_ obsolete.

galadril:
Would it be helpful to add information about this settings page and usage in the Domoticz wiki??
https://www.domoticz.com/wiki

kiddigital:
> It would be really cool to have OAuth as an option for the mobile apps to authenticate!
> 
> So how can we push this forward??

@galadril , this PR is now part of the Beta. Could you give it a go to see if it works as expected when connecting an app to it?

galadril:
I would love to, but im looking at the applications page, can create a new app.. but i dont really get what client Id / secret to use to authenticate. (that was why i suggested a wiki maybe)

kiddigital:
> I would love to, but im looking at the applications page, can create a new app.. but i don't really get what client Id / secret to use to authenticate. (that was why i suggested a wiki maybe)

Hi @galadril , I started with a [Wiki page](https://www.domoticz.com/wiki/Security) as suggested.

kiddigital:
@galadril , let me know if I can be of assistance... PM me if needed (or Slack)

PR #4877, opened 2021-07-20
Rework teachin procedures, Improve ESP3 database management and debugging, Fix a few things

efdeb:
Implement 1BS (D5-00-01) teachin, see EnOcean Equipment Profiles 2.6.8, appendix 3.2, page 320.
D5-00-01 need to be teached in before they are managed by Domoticz.

Implement UTE (D4-xx-xx) teachin, see EnOcean Equipment Profiles 2.6.8, appendix 3.6, page 327.
nb. UTE teachout is not supported to avoid conflict with teachin. Use Domoticz interface to delete devices.

Implement RPS (F6-xx-xx) teachin, see EnOcean Equipment Profiles 2.6.8, appendix 3.1, page 320.
nb 1. as stated in EnOcean Equipment Profiles 2.6.8, §1.7, p.13, RPS teachin is implemented to avoid inadvertent learning : RPS telegrams have to be triggered 3 times within 2 seconds to allow teach-in
nb 2. as this teachin method does not provide the EEP of the device, F6-02-01 is assumed by default. It can later be changed by editing the database (domoticz.db). I shall later on develop an interface to do it directly from Domoticz.
nb 3. add support for F6-01-01 and F6-02-02 EEP (even if  F6-02-01 is assumed by default...)

Create (or update) functions to encapsulate database access to EnoceanSensor table and generalize use of m_nodes cache to limit database operations.

	void LoadNodesFromDatabase();
	Loads all EnOcean Nodes attached to ESP3 hardware into m_nodes map

	NodeInfo *GetNodeInfo(const uint32_t iNodeID);
	NodeInfo *GetNodeInfo(const std::string nodeID);
	Search node by its nodeID into m_nodes map, return NodeIndo if found and nullptr if node doesn't exist

	void TeachInNode(const std::string nodeID, const uint16_t manID, const uint8_t RORG, const uint8_t func, const uint8_t type, const bool generic);
	Teach a new EnOcean node in, creating it both in the database and in m_nodes cache

	uint32_t GetINodeID(const std::string nodeID);
	Get uint32_t nodeID value from string nodeID
	
	struct NodeInfo
	Add RORG field to allow recording of full EEP profile.
	Add boolean generic field : true if node has been created with default EEP, false if EEP has been provided during teachin

Improve ESP3 protocol and ESP3 devices debugging
- Add DumpESP3Packet functions to improve readability of ESP3 protocol debug for packets read from, or sent, to ESP3 hardware
	std::string DumpESP3Packet(uint8_t packettype, uint8_t *data, uint16_t datalen, uint8_t *optdata, uint8_t optdatalen);
	std::string DumpESP3Packet(std::string esp3packet);
- Add Readcallback test cases
- Add functional tests case for several supported EnOcean ESP3 devices  (to be continued...)

Add GelClockTicks in Helper.h / Helper.cpp, time in msecs since origin
	time_t GetClockTicks()

Fixes
- Fix GetDeviceValue in case rangeMin > rangeMax
- Fix A5-07-xx and A5-09-xx management

-------------------------
That's all for this contribution.
Next steps mid august.
Testers are welcome.


gizmocuz:
Thank you for all the patches!
Could you change one thing, declare const string arguments in functions as

(const std::string &nodeID);

instead of

(const std::string nodeID);

efdeb:
@gizmocuz
Yes, of course.
Done in the commit above.

gizmocuz:
Great!, Thank you for this PR and time again!!

PR #4880, opened 2021-07-22
Some improvements for RFPlayer

ultrasuperpingu:
* Add Manual Switchs management for supported protocols
 * Add more protocols handling in reception
 * Jamming detection switch is created and updated on jamming detection
 * A Jamming Simulation button is created and allow to simulate Jamming


gizmocuz:
Is it possible to place the function ConvertToGeneralSwitchTypeRFP as static function in the ZiBlueBase class ? And then call ZiBlueBase::ConvertToGeneralSwitchTypeRFP in the webserver code?
This keeps the code a bit in place (... if this is possible)

I think al those "Edisio Command Field" defines can be in the .cpp file

Thanks!

ultrasuperpingu:
@gizmocuz 
Yes, of course, I will do it.
For Edisio fields, it is data provided by RFPlayer specs, so it was for me the right place to put that. I didn't implemented edisio management because I don't own the required devices (so it's not used)  but I will be useful for ones who want to implement it.

ultrasuperpingu:
Few thoughts about this:
 * I only tested with with the devices I own. That means few Oregon Temp+Hum and Temp sensors and RTS blinds and remotes. If somebody is able to test it with other devices (Oregon rain/wind..., X10/AC/ARC/Visonic/X2D remotes/switches/shutters...), it would be nice but I may be able to have more feedback once the PR in beta release.
 * Oregon sensors send an idChannel field containing the configured channel of the sensor and a random number that change when the sensor is reset (or battery is changed). This is good because (at least on mines), the sensors only propose 3 differents channels, so if you own more of them, you're still able to separate the different device. But I own less of them and when I need to change the battery, another device is created in Domoticz and I have to make a SQL request to merge the old one with the new one. Is-it possible to store and edit an additional parameter in a Temperature device to handle if it need to use the random part for the ID or do I need to use a hardware wide parameter to implement such a feature ? (probably not in this PR anyway)
 * NodeIDs of sensors used to be 16bits length so we needed to use hashcode to build a valid NodeId like (pSen->idPHY ^ pSen->idChannel). Is-it still the case ?
 * The RFPlayer is able to detect Jamming. To get the jamming alert in Domoticz, I create a simple switch (On/Off) but I'm not sure it is the best way to do it because the user is able to switch this device (in this case, I simply return false in the WriteHardware method, so an error is raise in the web app). I'm pretty sure there is a better way to do it. It's the same thing for RTS remotes (the device in domoticz only provides the last command sent, it shouldn't be editable). Except using Text devices, I didn't found an other way to do it. How would you do (I'm sorry, I'm pretty sure the answer is obvious but I don't get it) ?
 * When you add a Manual switch in domoticz interface, you have dozens of protocols in the list with a lot of them that are not supported by the selected device. I thought it could be interesting to add the ability for devices to declare the protocols and parameters for the protocols they are able to handle. It should be possible to add a method like GetAddManualSwitchProtocols in CDomoticzHardwareBase class (returning nothing in this class but overridable) and send that to json if it doesn't return nothing to present only the supported protocols in interface (otherwise, the old code would be executed). It should be possible to make this feature available to Python plugin Hardwares too by doing that. Do you think it is a good idea ?

ultrasuperpingu:
@gizmocuz 
Do you need more work from me on this ?

gizmocuz:
@ultrasuperpingu , no looks perfect, thank you.
About the 'add manual switch', I think that's a good idea, and also keep using the button from the switched page and not create 100 of options in the hardware page

gizmocuz:
Thank you for this PR!

PR #4886, opened 2021-08-05
replace push_back with emplace_back

neheb:
emplace_back forwards the arguments to the constructor directly.


neheb:
Rebased.

gizmocuz:
Thank you for this PR. Aldo those might work, it makes the coffee less readable and could introduce errors in run time.
I am all for readable code in favour of optimized code (in this case where we are not using this 100.000 times a second there is also not a real performance improvement)

https://abseil.io/tips/112

neheb:
That tip is misleading. All changes here call the constructor explicitly. emplace does the same thing.

PR #4887, opened 2021-08-11
replace boost::tuple with std variant

neheb:
boost variant offers nothing


gizmocuz:
@neheb , thank you for this PR! Always great to use default stuff....
In **webserver/request_parser.hpp** there is still boost/tuple use with your PR, was this intended ?

neheb:
Yeah. That uses boost::tuple for some boost APIs. That one can't be swapped.

gizmocuz:
Perhaps then the current patch in this file is converting the std variant to a boost::tuple again, maybe it's best to keep the existing code and directly use the boost::tuple ?

neheb:
Not sure what you mean. The only change in request_parser.hpp is replacing make_tuple with {}. The only point of make_tuple is to avoid template parameters.

gizmocuz:
Thank you for this PR and your time !!

PR #4889, opened 2021-08-11
get rid of boost::thread

neheb:
No real advantage over std


gizmocuz:
Thank you again!

PR #4890, opened 2021-08-11
remove const type in std::map

neheb:
const here prevents copy elision. Remove it.


gizmocuz:
Thanks!!

PR #4891, opened 2021-08-11
use auto with iterators

neheb:
const vs non-const iterators are a complex situation that the compiler
can handle.


neheb:
Rebased.

keryell:
At least the code looks quite cleaner.

gizmocuz:
Yes it is ;) ... but coming from a non 'auto' area the current code looks clear to me as well
Verified these changes, merging PR

Thank you for the PR!

main/Helper.cpp
```diff
@@ -968,9 +968,7 @@ void rgb2hsb(const int r, const int g, const int b, float hsbvals[3])
 
 bool is_number(const std::string& s)
 {
-	std::string::const_iterator it = s.begin();
-	while (it != s.end() && (isdigit(*it) || (*it == '.') || (*it == '-') || (*it == ' ') || (*it == 0x00))) ++it;
-	return !s.empty() && it == s.end();
+	return std::any_of(s.begin(), s.end(), [](char c) { return isdigit(c) || (c == '.') || (c == '-') || (c == ' ') || (c == 0x00); });
```
gizmocuz, line 971: Not working correctly, reverted in later commit

keryell: Looking at the *Notes* section of https://en.cppreference.com/w/cpp/algorithm/all_any_none_of I guess this is because `std::all_of` should be used instead of `std::any_of`.
Otherwise I would prefer to use `std::isdigit` instead `isdigit`, but this is just a coding standard personal taste to be more coherent about C++.

gizmocuz: Yep that would work

PR #4892, opened 2021-08-11
Updating from dbversion < 134 fails, hardware table already has LogLevel column

dotnetniet:
When updating from dbversion < 134, the hardware table is recreated. In dbversion < 148 the column LogLevel was then duplicate. Added check if column already exists.

### DIFF
--- a/main/SQLHelper.cpp
+++ b/main/SQLHelper.cpp
@@ -2773,7 +2773,10 @@ bool CSQLHelper::OpenDatabase()
 		}
 		if (dbversion < 148)
 		{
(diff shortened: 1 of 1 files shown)


gizmocuz:
Assuming the LogLevel column was added @ database version 148, could it be that you downgraded your domoticz version and this has caused the duplicate column ? (Sure it's always good to check if a column already exisited)

dotnetniet:
Hi, upgraded from dbversion 116. In dbversion < 134 check, the hardware table is fully recreated, so that one added the LogLevel column already.

gizmocuz:
Thank you for this PR ! Yes you're correct, also in db update 110

PR #4894, opened 2021-08-12
unique_ptr conversions

neheb:
This is C++14. new can be replaced with make_unique.


gizmocuz:
Thank you!!

PR #4896, opened 2021-08-12
replace single character replace with std

neheb:
More efficient.


gizmocuz:
Thank you!

Issue #4895, opened 2021-08-12
(Windows install) Updating from 2017 version to 2021.1 stable missing vcruntime140.dll


dotnetniet:
Version: 2017 -> 2021.1 stable
Platform: Windows 10 Pro
Description: Missing Visual Studio C++ Redistributable Dll's after upgrading to latest stable version 2021.1

Domoticz is running on a clean Windows 10 Pro as service. It's the only active application on that system and has run for more then 5 years without problems with version from 2017.

Did a full backup and run the installer. After that the service started, but status was paused. Stopped the service and started domoticz manually.
Multiple error pop-ups where shown with missing dll's, like vcruntime140.dll and the other VC redistributable ones.

System has the x64 bit versions that come with windows, but it's missing the x86 versions in the syswow64 folder that are used by 32bit applications, like Domoticz.
After installing the x86 version (https://aka.ms/vs/16/release/vc_redist.x86.exe), service started normal.

Suggesting to add this vc_redist_x86.exe into the installer if these dll's are not present on the system.

Is this the correct solution or should all the used dll's in the installer be using a static build?
Seems that the Lua.dll is not statically linked (verified with dependency walker tool).

Looking to add this check into the inno setup, so it's a permanent solution, but do not know if this is the correct action.

dotnetniet:
Added pull request https://github.com/domoticz/domoticz/pull/4897

gizmocuz:
Sorry missed the link to this issue.
It should be statically linked. Or the two missing dll's needs to be shipped with domoticz.
This PR is adding 13MB to the installer ....
It's the first time I hear about this, do you know which file needs this dependency ?

dotnetniet:
There are multiple runtime dll's missing in this situation and it's not recommended to copy these runtime dll's 'manually'. Microsoft recommends the redistributable installer.

I also wanted to use Windows server 2019 and that has even less standard dll's installed., but think
 no normal person  will use  windows server😀

Regarding what dll's are not statically linked, I found that the lua.dll is not static, but did not look at other dll's. Will need to have a look at those.

dotnetniet:
I checked on the clean Windows server 2019 install, my Windows 10 is not clean anymore.

The following dll’s from the application folder need the vcruntime140.dll and MSVCP140.dll files:

lua.dll
libcurl.dll
jsoncpp.dll
libcrypto-1_1.dll
libssl-1_1.dll
zlib1.dll
sqlite3.dll
mosquitto.dll
[...]

PR #4897, opened 2021-08-13
Added check if vcruntime140.dll (32bit/x86) is installed.

dotnetniet:
Added check if vcruntime140.dll (32bit/x86) is installed. If not installed or older version then 14.29, VC_redist.x86.exe is used to install VC++ runtime libraries.


gizmocuz:
This will add 13MB to the installer. Is this needed ? It's the first time I hear about this .... Assuming you are running windows 10 with all updates installed ?

dotnetniet:
Correct, Windows 10 with all updates installed. Also checked on Windows server 2019 and that one even has less runtime dll's. 
Microsoft recommends to use the redistributable installer, because of these different possible environments.

dotnetniet:
Regarding the extra 13MB, Inno setup can also use the DownloadTemporaryFile function, but you need to have an internet connection to use that. That will also introduce an external dependency and will fail when Microsoft moves the file on there end.
I like to have everything in one package, that's why is chose for this solution.

dotnetniet:
See https://github.com/domoticz/domoticz/issues/4895 for further discussion.

dotnetniet:
We choose to include missing dll's in setup: https://github.com/domoticz/win32-libraries/pull/2

This change is not needed.

Issue #4898, opened 2021-08-14
Updating from 2017 to 2021.1 stable gives selector actions url encoded (bug in dbversion 128)


dotnetniet:
Version: stable 2017 -> 2021.1 stable (dbversion 116 -> 148)
Platform: Windows 10/Server 2019

Hi,

After updating from version 2017 (dbversion 116) to the current stable 2021.1 version, all url's in selector actions are url encoded and stopped working.

I found that in this update the urldecode was removed:
_Revision: 16731807489d518b1134346242bb7c04db3d9dd0_
_Author: Rob Peters <info@domoticz.com>_
_Date: 2018-05-22 16:27:34_
_Message: Selector Switch, fix action entry (utf/url) and hide off level action_

In main/SQLHelper.cpp the CURLEncode::URLDecode is removed from line 4282 and 4283. After that all older selector action values that still have an encoded url not work anymore.
This change is with dbversion 128.

I can add a fix with dbversion < 128 check, but should it be a hotfix from master or feature fix from develpment branch?

![image](https://user-images.githubusercontent.com/21026753/129457655-8bf32a6c-c45b-429b-8bce-9737ec48ae9c.png)

gizmocuz:
Always make PR's to the active branch (Development in this case)
You where upgrading from a very old version... I don't think there are much more users that where using a version like you did
It's base64 encoded now for various (good) reasons.
If you have a few selector switches you could use a website to decode them and enter them again to solve this

gizmocuz:
Please create only one issue instead of two.

https://github.com/domoticz/domoticz/pull/4899

dotnetniet:
Other PR https://github.com/domoticz/domoticz/pull/4892 was also needed, so I had to merge master back into development. Otherwise db update crashed on development.

I'm used to work with hotfixes on master, but in this environment it's probably better to use development for all fixes.

Can you update development with this PR https://github.com/domoticz/domoticz/pull/4892 ?

gizmocuz:
@MarkVeldt , sorry did not see it was made against the master branch. I reverted the PR. Is it possible to create a PR for the development branch?

dotnetniet:
Created new PR https://github.com/domoticz/domoticz/pull/4901 for development about hardware LogLevel column.

dotnetniet:
PR #4899 is single issue now.

gizmocuz:
Thanks for reporting. Aldo it's a very old installation ;) PR merged

PR #4899, opened 2021-08-14
Selector actions are url encoded (bug in dbversion 128)

dotnetniet:
Issue https://github.com/domoticz/domoticz/issues/4898


gizmocuz:
Thank you!!

PR #4905, opened 2021-08-18
Fix unreported bug: Send command to RFPlayer devices through dzEvents…

ultrasuperpingu:
… not working


gizmocuz:
Thank you again!

PR #4908, opened 2021-08-22
[USBtin/Multibloc V8] - IBS integration - timeleft correction + some over modifications

ghost:
there was an error calculating the time remaining before charging / discharging with IBS informartion.

### DIFF
--- a/hardware/USBtin_MultiblocV8.cpp
+++ b/hardware/USBtin_MultiblocV8.cpp
@@ -1393,10 +1393,10 @@ void USBtin_MultiblocV8::Traitement_IBS(const unsigned int FrameType, const unsi
 		FrameType == type_IBS_5_1 ||
 		FrameType == type_IBS_6_1 ){
 			
-		//if( m_BOOL_DebugInMultiblocV8 == true ){
+		if( m_BOOL_DebugInMultiblocV8 == true ){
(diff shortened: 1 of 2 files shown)


hardware/USBtin_MultiblocV8.cpp
```diff
@@ -303,11 +305,10 @@ void USBtin_MultiblocV8::StopThread()
 	ClearingBlocList();
 }
 
-void USBtin_MultiblocV8::ManageThreadV8(bool States, bool debugmode)
+void USBtin_MultiblocV8::ManageThreadV8(bool States, unsigned int debugmode)
```
gizmocuz: Why did you change this back to unsigned int and using/comparing it as a bool ? It is a boolean

gizmocuz:
Thank you for the PR. Could you undo all changes and ONLY make the change for the time correction ?

ghost:
Ok, yes I'll do that tomorrow. I don't understand why sometime when I get the code from origin reposit there is spaces and some over changes like that (tabulation, etc...), also the encoding change, from UNIX(LF) to WIN(CR+LF) so all the file is modified when I do a winmerge between domoticz reposit and mine :-D , Is there a mechanism that does automatic layout on the CI / CD side ?

gizmocuz:
@Reivax851 , This CR/LF should be handled automatically. Maybe it is your editor ? I use Visual Studio

ghost:
In addition, I conditioned some logs + change cast float to int on the 2 values Soc and Soh.  I compiled and tested directly on my server, everything works fine :-)

gizmocuz:
Thank you for this PR and time again!

PR #4911, opened 2021-08-24
Improve messages when opening DB at startup

efdeb:
**Add more user frienfly messages when opening Domoticz database.**

Helpful to debug, particularly when a version upgrade occurs.
Otherwise, need to sqlite3 open the database file to get it's version number.

Helpful to locate precisely the database file.

The following messages are proposed :

Error message in case of problem when opening the database:

    Error: Error opening <database name> database: <error cause>

When opening an existing database matching the current version:

    Status, Open <database name> database (version: <current version number>)

When creating a missing database :

    Status, Creating <database name> database (version: <current version number>)

When upgrading an existing database to the current version:

    Status, Upgrade <database name> database from version <version number> to <current version number>

Error message in case an existing database version is higher than the current version:

    Error, <database name> database incompatible with this Domoticz version.
    Error, Cannot downgrade version from <current version number> to <version number>!

### DIFF
--- a/main/SQLHelper.cpp
+++ b/main/SQLHelper.cpp
@@ -38,7 +38,8 @@
 #define __STDC_FORMAT_MACROS
 #include <inttypes.h>
 
-#define DB_VERSION 148
+#define DB_NAME "domoticz.db"
+#define DB_VERSION 149
 
 extern http::server::CWebServerHelper m_webservers;
 extern std::string szWWWFolder;
@@ -611,7 +612,7 @@ CSQLHelper::CSQLHelper()
 	m_bPreviousAcceptNewHardware = false;
 	m_bLogEventScriptTrigger = false;
 
-	SetDatabaseName("domoticz.db");
+	SetDatabaseName(DB_NAME);
 }
 
 CSQLHelper::~CSQLHelper()
@@ -626,7 +627,7 @@ bool CSQLHelper::OpenDatabase()
 	int rc = sqlite3_open(m_dbase_name.c_str(), &m_dbase);
 	if (rc)
 	{
-		_log.Log(LOG_ERROR, "Error opening SQLite3 database: %s", sqlite3_errmsg(m_dbase));
+		_log.Log(LOG_ERROR, "Error opening %s database: %s", m_dbase_name.c_str(), sqlite3_errmsg(m_dbase));
 		sqlite3_close(m_dbase);
 		return false;
 	}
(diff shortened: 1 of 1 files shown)


gizmocuz:
I looked at this PR over and over again and I don't think we need this as it will confuse a lot of people.
Opening database xyz.db (version 168) ? No normal user should even know this
upgrading database xyz.db from version 148 to 190 ... No normal user should know this, this is all internal
If a user is stupid enough to downgrade his/her domoticz system and expect it to work..... then a message is already displayed that this is not possible.
And that's it....

These days when people upgrade to a new (beta) version there is a full backup made and it is easy to switch back (including the database)

kiddigital:
@gizmocuz , @efdeb ,

I think maybe most of the status messages in this PR should be Debug messages (NORMAL). For troubleshooting purposes it is very useful information.

For the Error messages, I think these are good improvements giving more insight directly. Especially also when asking for help on the forum and other places.. this information is almost always needed.

I do agree that for normal operations, these Domoticz internals should be kept from the user in normal Info/Status messages.

efdeb:
Hi @gizmocuz, @kiddigital,

I think you both are right : the messages I added are intended for debug and not for the end user, but they still might be useful for developers.

Should I propose a new version taking that into account ?

option 1 : using CLogger::Debug to be able to dynamically activate them
option 2 : using #ifdef DEBUG_FLAG bla bla bla #endif conditional compilation directives

Thanks in advance for your opinions.

kiddigital:
@efdeb Dynamic otherwise it is to hard for regular users. Using an optional flag is something users can do.

galuschka:
I have troubles to get 2023.2-release working. My db is in version 159 and DB_VERSION in SQLHelper.cpp is 163. But the condition dbversion > DB_VERSION gets true!!!
So a more detailed debug output would help a lot. That output is not seen by "normal" users anyway.

However: in case you know your database version (sqlite3 - search for "DB_Version") you may want to see the history of DB_VERSION in evolution of SQLHelper.cpp of future versions:

```
for i in $(git log --oneline --follow HEAD..origin/master main/SQLHelper.cpp | cut -d " " -f 1)
do
    echo -n       $(git describe --first-parent --tags --abbrev=5 $i)
    echo -n " /"  $(git describe --first-parent --abbrev=5 $i)
    echo -n " in" $(git describe --contains $i)
    echo -n ": "
    git show $i:main/SQLHelper.cpp | grep DB_VERSION | head -1
done
```

gizmocuz:
2023.2 is not released yet. There is a PR with the draft

galuschka:
Oh! I did not recognize... I just got an email about to update and the web interface shows this "update available" message.

gizmocuz:
There are quite a few updates but no stable yet ... Should be very soon...

galuschka:
And due to some more debug info in SQLHelper, I meanwhile found the reason of that dbversion > DB_VERSION:
The database.db file used is the one in /opt/domoticz/... So I just checked the wrong file.
But thanks for quick response! Almost unbelievable! ;-)

gizmocuz:
Glad it's working!

efdeb:
Hello,
It was to help solving such problems that I proposed this PR some time ago.
Have a nice day.

PR #4913, opened 2021-08-25
replace most usages of std::list with deque

neheb:
std::list has a massive runtime cost while not being
much better than deque.

The other places use sort() and remove() which are not
available with deque.


gizmocuz:
Thank you!!

PR #4915, opened 2021-08-26
Improve DEBUG messages when opening DB at startup

efdeb:
Hi @gizmocuz, @kiddigital,

This is my last attempt for this PR.

Replaced Log messages by Debug message, so :

- normal users see only the former error messages in the log,
- launching Domoticz with '-loglevel all' adds the troubleshooting complements with database path and versions numbers.

PS. another PR will follow to fix '-debuglevel all' option which is not working in the current implementation

### DIFF
--- a/main/SQLHelper.cpp
+++ b/main/SQLHelper.cpp
@@ -39,7 +39,7 @@
 #include <inttypes.h>
 
 #define DB_NAME "domoticz.db"
-#define DB_VERSION 148
+#define DB_VERSION 149
 
 extern http::server::CWebServerHelper m_webservers;
 extern std::string szWWWFolder;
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thanks for this PR, but could you regret the log type messages (to not use debug).
If I read this PR correctly, and a user is using a wedding database(version) it is logged as a debug message but it should appear as normal because debug message could be disabled

efdeb:
Hi @gizmocuz,

You remember we had a discussion about the first version of this PR I proposed, about the messages needing to be visible for a normal user and additional information which could be of interrest for a power user or developper for troobleshooting.

In this version, I thus decided not to change the existing behaviour for the normal user, and only adding additional debug information for power users or developpers.

Unchanged : a normal user has only errors messages in his log when :
1. database file cannot be opened (old line 629)
2. database version is newer than the one expected (old line 649).

I added :
1. database file cannot be opened (line 631) => **database path**
2. database is empty (line 645) => **path and version of the created database**
3. database version is  the one expected (line 650) => **path and version of the database**
4. database version is older than the one expected (line 652) => **path and current/upgraded version of the database**
5. database version is older than the one expected (line 657) => **current/expected version of the database**

Of course I can change that, just tell which line to move to the log, because I do not catch what you call "wedding database(version)"...

efdeb:
Alse, to resolve conflict, line 42, 
#define DB_VERSION 148
has to be changed to 
#define DB_VERSION 149

PR #4916, opened 2021-08-26
Fix SetDebugFlag for 'all' or number values + minor adjustments

efdeb:
In CLogger::SetDebugFlags(const std::string &sFlags) :
- fix missing missing 'else', **causing 'all' flag being currently ignored**
- replace atoi by strtoul to get unsigned number value
- replace erroneous SetLogFlags call by SetDebugFlags to take it into account

In CLogger::SetLogFlags(const std::string &sFlags) :
- replace atoi by strtoul to get unsigned number value

In CLogger::CLogger() constructor :
- call SetLogFlags dedicated function to initialize m_log_flags
- call SetDebugFlags dedicated function to initialize m_debug_flags


gizmocuz:
Thank you!!

PR #4917, opened 2021-08-26
Fix --help for -loglevel & -debuglevel

efdeb:
Update message displayed when launching Domotocz with --help option for :

-loglevel (combination of: **all**,normal,status,error,debug)
-debuglevel (combination of: **all**,normal,hardware,received,webserver,eventsystem,python,thread_id,**sql**)

nb. don't speak about number value (bit mask), even if if supported by those options.


gizmocuz:
Thank you!!

PR #4919, opened 2021-08-26
Code refactoring in WebServer.cpp

efdeb:
**No functional changes : just code refactoring to improve readability.**

void CWebServer::HandleCommand(const std::string &cparam, WebEmSession &session, const request &req, Json::Value &root)
when (cparam == "testswitch")

1/ Reordering of "_if (lighttype == XX)_" blocks  by increasing XX values
2/ Simplification of _return_ statement for HTYPE_EnOceanESP2/HTYPE_EnOceanESP3/HTYPE_USBtinGateway when (lighttype == 70)

_Please note "else if (lighttype < 70)" becomes "else if (lighttype < 68)" because "else if (lighttype == 68)" and "else if (lighttype == 69)" are present._

void CWebServer::RType_Hardware(WebEmSession &session, const request &req, Json::Value &root)

1/ _#ifdef / #endif_ for Special case for openzwave (status for nodes queried)


kiddigital:
@efdeb , maybe a good idea to `git rebase -i upstream/development` (or whatever you have called _upstream_) and double check if everything still is and works as expected. There have been a few changes in the Webserver code recently, so checking is useful. Also to prevent any merge conflicts if any.

efdeb:
@kiddigital : I tried... with the poor result of pulling many, many, many unrelated changes. I definitely do not understand something with git rebase...

@gizmocuz : I close this PR, ans reopen a new one rebased on current development branch. The changes made appear very complex whel looking at a diff with the original file, but they are indeed relatively simple. I shall add additional comment about it in my new PR.

PR #4923, opened 2021-08-31
Allow hardware to declare manual switches they are able to handle.

ultrasuperpingu:
RFPlayer implements this feature.


waltervl:
Hello @ultrasuperpingu 
I see you have developed/maintaining the Ziblue RFPlayer integration.
Can you make a Wiki page for this as the only reference for this device is linking to an outdated Python Integration. Or is there a french page where we can link or translate with all the instructions and implemented functionality?
If you need a Wiki account please ask @gizmocuz through slack.

Old wiki Page: https://www.domoticz.com/wiki/Plugins/Ziblue-RFPlayer.html

waltervl:
@ultrasuperpingu I have added a simple Wiki page for you https://www.domoticz.com/wiki/ZiBlue_RFPlayer
Please update as it is very basic.

ultrasuperpingu:
@waltervl Thanks. I'll try to update this but the documentation would be very different if this PR is accepted. I think it's a bit early to update documentation for now. I will need a wiki access to do it indeed.

waltervl:
@ultrasuperpingu If you have a forum account please send a forum PM with your emailadres to gizmocuz.

ultrasuperpingu:
@waltervl Just create one but I can't send PM yet...

ultrasuperpingu:
@waltervl Can we delete those comment because they are not related to this PR ?

ultrasuperpingu:
@gizmocuz 
We talked about this feature on PR #4880
This implementation is just a proposition.
Do you think something like this can be integrated?

ultrasuperpingu:
@gizmocuz 
Hi,
I'd like to continue the development of RFPlayer for Domoticz and write the documentation for it but the tasks to do really depend if this PR will be integrated or not. Do you think this feature can be integrated? If yes I am, of course, ready to work on modifications you would need on this. If not, no problem, I will do otherwise.
By the way, how can I access to edit the wiki?

ultrasuperpingu:
@gizmocuz
Did you have time to look at this ?

gizmocuz:
@ultrasuperpingu , Still need to check out this PR if everything keeps working.
At the moment my time goes into MQTT which seems almost finished now

ultrasuperpingu:
@gizmocuz
I'm sorry to ask again but did you have time to look at this ?
I just would like to know if it is a no go, or if it will be integrated (of course with the modificationq that will be needed). I'd like to continue the development of RFPlayer for Domoticz but the tasks to do really depend if this PR will be integrated or not.

ultrasuperpingu:
There was conflicts so I merged the code with development branch. I don't know what causes the dependencies problem in the checks tho

gizmocuz:
Seems it was a problem with a lua library, seems to be resolved when I restarted the check...

gizmocuz:
Thank you for the changes! I will have to do some performance testing on them but it looks good !

Some questions, could you declare the json array in a little better readable way like:

```
std::string jsonConfig =R"([
	{"idx":"1","name":"Visonic 433MHz","parameters":[{"name":"housecode","display":"House Code","type":"housecode16"},{"name":"unitcode","display":"Unit Code","type":"unitcode16"}]},
	{"idx":"2","name":"Visonic 868MHz","parameters":[{"name":"housecode","display":"House Code","type":"housecode16"},{"name":"unitcode","display":"Unit Code","type":"unitcode16"}]},
	....
])";
```

That way it should also be easier to add more items

in GetManualSwitchParameters, could you use the following case system for the parameters

Parameters
SwitchTypeInOut
LightTypeInOut
dTypeOut
dSubTypeOut
devIDOut
sUnitOut

or comply with the rest of the code base

ID
unit
devType
subType
switchType

You don't have to specify eplicitly inout, it is confusing as well, take 'lighttypeInOut' ... it's used as input and can be 'const', you are not writing (out) to it.

I am also not sure why you need to push a multimap as parameter, better stick to just plain parameters, making things const where possible

Why are you using a string for the unit, this seems to be a int32_t internally

ultrasuperpingu:
It's done for the raw string literal to declare json.

For the multimap, it's filled dynamically with the parameters specified in the json. You can add in the json any parameters with the name and type you need and the values will be sent by javascript with the corresponding names. So, in WebServer, you don't know the name of those parameters, that's why GetManualSwitchParameters need the multimap containing all those parameters.

For the unit being a string, I just stick with types those variables was already. It's used later to fill a string vector used by m_mainworker.SwitchLightInt (in testswitch).

I'll try to change parameters names soon.

ultrasuperpingu:
The reason why SwitchType and LightType are inout is because RFLink use the method ConvertToGeneralSwitchType which modify those variables. I thought it was a good idea to let the hardware modify those values. So, if the RFLink implementation switch to this method, it may use it to integrate the ConvertToGeneralSwitchType method into GetManualSwitchParameters but it's maybe a bad idea...

gizmocuz:
Thank you for this PR and time !!

PR #4928, opened 2021-09-06
Set dynamically & per hardware Debug Levels

efdeb:
**As currently proposed for Log Levels, allow to set dynamically & per hardware Debug Levels  in domoticz hardware setup screen**

Log/Debug levels can be set either when creating a new hardware, or by updating an existing one.

This can be very useful for troobleshooting, making it possible to get debug information without recompiling or launching Domoticz manually with debug option enabled from the command line. In EnOcean ESP3 protocol, for instance, Normal debug will show how EnOcean telegrams are interpreted and Hardware debug will show exchanged ERP1 protocol data .

For that purpose, add 2 new checkboxes on Hardware setup screen and implement necessary logic.

**Hardware.html:**
Add 2 checkboxes => Debug Level : Nomal/Hardware

**Hardware.js / Mainworker.cpp / DomoticzHardware.h:**
_Unmodified:_
Use LogLevelEnabled 4 LSB bytes to exchange Log Level:
    0x00000001=LOG_NORM, 0x00000002=LOG_STATUS, 0x00000004=LOG_ERROR

Store current log level for hardware in m_LogLevelEnabled.

_New:_
Use LogLevelEnabled 4 MSB bytes to exchange Debug Level:
    0x00010000=DEBUG_NORM, 0x00020000=DEBUG_HARDWARE

Store current debug level for hardware in m_DebugLevelEnabled.

**DomoticzHardware.cpp:**
Implement necessary logic to follow what the user has defined in hardware setup screen.

**Logger.h/Logger.cpp**
Add 2 utility functions to get current Log/Debug flags:
	uint32_t GetLogFlags();
	uint32_t GetDebugFlags();

nb. The last commit "Set dynamically & per hardware Debug Levels" carries all the files changed by this PR. The others first 10  result in a noop...

### DIFF
--- a/hardware/DomoticzHardware.cpp
+++ b/hardware/DomoticzHardware.cpp
@@ -128,15 +128,29 @@ int CDomoticzHardwareBase::SetThreadNameInt(const std::thread::native_handle_typ
 void CDomoticzHardwareBase::Log(const _eLogLevel level, const std::string& sLogline)
 {
 	if (!(m_LogLevelEnabled & (uint32_t)level))
-		return; //this type of log is disabled
+		return; // Log level is disabled by hardware parameters
 
+	uint32_t currentLogFlags = _log.GetLogFlags();
+	_log.SetLogFlags(m_LogLevelEnabled);
+
+	// Save and force log level
+	
 	_log.Log(level, "%s: %s", m_Name.c_str(), sLogline.c_str());
+
+	// Restore previous log level
+
+	_log.SetLogFlags(currentLogFlags);
 }
 
 void CDomoticzHardwareBase::Log(const _eLogLevel level, const char* logline, ...)
 {
 	if (!(m_LogLevelEnabled & (uint32_t)level))
-		return; // this type of log is disabled
+		return; // Log level is disabled by hardware parameters
+
+	// Save and force log level
+	
+	uint32_t currentLogFlags = _log.GetLogFlags();
+	_log.SetLogFlags(m_LogLevelEnabled);
 
 	va_list argList;
 	char cbuffer[MAX_LOG_LINE_LENGTH];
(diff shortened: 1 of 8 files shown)


kiddigital:
Hi @efdeb , interesting PR.

On one hand I see how you make it _easier_ for end-users to get more insight/data in what is going on. But on the other hand when it comes to _debugging_ I think some more technical skills are required other than 'only' activating Debug level(s). Stopping, (re)starting or capturing _segmentation fault output_ etc. is often/usually also needed.

The idea that the user wanting to do the debugging is at least capable of stopping/starting with additional flags (like Debug levels, empty/different DB, other config, disabled modules, etc.) makes it much more likely that the debugging can be done succesfully.

But I do see a need for a little more _verbose_ level of logging than we have at the moment with the LOG_NORMAL. (But less verbose/detailed than DEBUG). Maybe adding a LOG_VERBOSE, and a switch to turn VERBOSE on/off for a module, is a good in-between direction. An end-user can easily turn-on VERBOSE when a new module is setup to see/validate that things work as expected (or not) and should provide details that are useful for end-users. For example for modules that connect over the internet, to see what gets communicated back-and-fort or when opening a hardware connection to a connected device, etc. And turn VERBOSE off when things run as normal.

efdeb:
Hi @kiddigital , thanks for your time & comments.

Concerning **LOG_VERBOSE**, il may of course be added and used instead, but I would be necessary to **review most Log/Debug calls** in hardware support code in order to introduce it. Is it really worth it ? Anyway, I can help anyway to implement it if need be.

For this PR, I was focussed using **existing log/debug levels** in order to limit changes.

Also, following former discussions with @gizmocuz and you concerning #4915, I choosed to use Debug(DEBUG_NORM or DEBUG_HARDWARE) to display troobleshooting data (not really debug data indeed) in EnOcean ESP3 hardware support :
- a normal user **do not see any of these messages** in his logs
- an advanced user can get more data without compiling code with #define DEBUG flags or even finding how to launch Domoticz with convenient command line options on his system.

For example, a user tries to use a new module, not really or completely supported by Domoticz. He reports in the forum and gets help from more experienced users / developpers who probably do not own the same module. He can easily generate troobleshooting data which may greatly help understanding / implementing the new module. Moreover, as the debug level is selected per hardware, he only gets the interresting debug messages.

waltervl:
From a user point of view we have now 3 standard logging levels (Info, status and error) that can be switched on/off.  Additional a user can start domoticz with additional debug loglevel  (-loglevel  normal,status,error,debug).
For me, more detailed info messages used for user level initial debugging may be shown as they can be easily switched off by the user if everythings is working as normal.

In a Python plugin the user can also switch on additional debug levels in the user interface https://www.domoticz.com/wiki/Developing_a_Python_plugin#C.2B.2B_Callable_API Ofcourse only valid when implemented by the hardware plugin developer.

In the wiki page https://www.domoticz.com/wiki/Developing_a_hardware_plugin there is no refference what each level means. That could be added to the wiki to give each C++ developer a better guidance.

gizmocuz:
I do agree that we have enough logging methods available.
When debugging a native hardware implementation you should compile in debug mode (which activates the #ifdef DEBUG option), and you could optionally use the log/debug feature.
I assume developers are using visual studio, and they set breakpoints, and do a step-by-step debugging session, once all is working, only a minimal set of debug log lines should exists.
Enabling the debug option at startup is therefor perfect
If a end-user needs to perform debugging tasks, it is easy to disable all other hardware (... because you only want to debug 'that' specific hardware) and use the debug startup options

PR #4934, opened 2021-09-16
Fixed Linux compile warning in parameters in Debug messages of EnOceanESP2/3

kiddigital:
Fix for incorrect (typo?) parameters in Debug messages resulting in compile warning under Linux

See https://github.com/domoticz/domoticz/pull/4872/checks?check_run_id=3625572567 as an example...

### DIFF
--- a/hardware/EnOceanESP2.cpp
+++ b/hardware/EnOceanESP2.cpp
@@ -1504,7 +1504,7 @@ bool CEnOceanESP2::ParseData()
 					tsen.TEMP_HUM.battery_level = 9; // OK
 					tsen.TEMP_HUM.rssi = 12; // Not available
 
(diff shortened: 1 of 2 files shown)


gizmocuz:
Thank you for this PR. I think the original author would like to print a percentage sign.
Maybe you can change it to %% instead of removing the percentage ?

kiddigital:
Hm.. good catch.. didn't read it that way. Will change the PR to use %%.

efdeb:
Hi both, you are right, the intention is to print a % sign.
Thanks for pointing it !
It is fixed in #4937 pull request I opened today.
So if mine is merged, this one can be closed.

kiddigital:
Fixed :)

gizmocuz:
Thank you!

PR #4935, opened 2021-09-17
Remove unnecessary print format in Logger.cpp

efdeb:
void CLogger::Log(const _eLogLevel level, const std::string &sLogline)
calls void CLogger::Log(const _eLogLevel level, const char *logline, ...)

There is no need to use "%s" format, just log directly the string value.

Nb. only the last change in this PR is significant
First 11 changes result in a noop...


efdeb:
Seems valid to me, but generate compilation warning.
I close the PR.

PR #4942, opened 2021-09-24
Refactored way of serving theme files

kiddigital:
This PR refactors the _theme_ processing within the Domoticz webserver(s).

- All theme processing logic  is now more in 1 place and not at different places and sourcefiles anymore.
- Theme processing is only done when not using the Default theme (so less code to run when using Default theme)
- The complete Domoticz UI (www folder) can now be served statically by any webserver (and not only the Domoticz webserver)
-- But only for the _default_ theme ofcourse

- Also refactored handling of serving _directories_ (rewrite request to serve index.html of directory if exists) as it was done at 2 places and one (request_handler) seemed broken (not working).

Main reason for this refactoring is to get a clearer separation between the UI and the Webserver (which now more and more becomes a WebService). It will make testing of the UI and testing of the WebService easier and can be done independant (for example by mocking the WebService).

Also it makes it 'easier' to develop new UI's/front-end's and/or serve the UI from a different place (webserver) than where the WebService might run. For example, one could serve the UI directly from an AWS S3 bucket (as Static files) and proxy only the JSON commands to a different (internal) location.

Have not tried it yet, but I think using AWS API gateway and AWS S3 static content serving it should be possible to create a setup now where the Domoticz UI (the 'www' folder) is served from a public location and the calls to _json.htm_ can be proxied to an internal Domoticz webserver/webservice. And therefor also secured properly! Anyone :) ?

For those on Linux: I added a [_lighttpd_](https://www.lighttpd.net/) config in the 'test' directory. If you install it, you can start Domoticz (at port 8080) and also start _lighttpd_ (`lighttpd -D -f test/lighttpd.conf`) which runs at port 8888.

Now you can request the Domoticz UI at port 8888 which gets served from lighttpd and only the API calls are handled by Domoticz :)

### DIFF
--- a/main/WebServer.cpp
+++ b/main/WebServer.cpp
@@ -812,7 +812,6 @@ namespace http
 							for (const auto &file : _ThemeFiles)
 							{
 								std::string tfname = file.first.substr(szWWWFolder.size() + 1);
-								stdreplace(tfname, "styles/" + sWebTheme, "acttheme");
 								response += tfname + '\n';
 							}
 							continue;
--- a/main/WebServer.cpp
+++ b/main/WebServer.cpp
@@ -337,15 +337,20 @@ namespace http
 			m_pWebEm->RegisterIncludeCode("combolanguage", [this](auto &&content_part) { DisplayLanguageCombo(content_part); });
 
 			m_pWebEm->RegisterPageCode("/json.htm", [this](auto &&session, auto &&req, auto &&rep) { GetJSonPage(session, req, rep); });
+			// These 'Pages' should probably be 'moved' to become Command codes handled by the 'json.htm API', so we get all API calls through one entry point
+			// And why .php or .cgi while all these commands are NOT handled by a PHP or CGI processor but by Domoticz ?? Legacy? Rename these?
+			m_pWebEm->RegisterPageCode("/logincheck", [this](auto &&session, auto &&req, auto &&rep) { PostLoginCheck(session, req, rep); }, true);
 			m_pWebEm->RegisterPageCode("/uploadcustomicon", [this](auto &&session, auto &&req, auto &&rep) { Post_UploadCustomIcon(session, req, rep); });
-			m_pWebEm->RegisterPageCode("/html5.appcache", [this](auto &&session, auto &&req, auto &&rep) { GetAppCache(session, req, rep); });
-			m_pWebEm->RegisterPageCode("/camsnapshot.jpg", [this](auto &&session, auto &&req, auto &&rep) { GetCameraSnapshot(session, req, rep); });
+			m_pWebEm->RegisterPageCode("/storesettings", [this](auto &&session, auto &&req, auto &&rep) { PostSettings(session, req, rep); });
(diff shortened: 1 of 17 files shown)


gizmocuz:
Thank you again for this PR!!

kiddigital:
@galadril , how do you propose this _variable_ should be implemented? You can do a _getthemes_ command giving you a list of available theme's. Actually it gives you the list of _directorynames_ within the 'www/styles' folder, which makes it hard/impossible to detect as a theme which one is yours. Or do a _getsettings_ command and look at the 'WebTheme' variable also containing the directoryname of the current active theme.

kiddigital:
@waltervl , I will try to answer your questions (refrasing and extending on the earlier mentioned points):

The _problem_ with the old 'acttheme' was not a (obvious) problem from end-user perspective.
1) In my opinion it was a software architectural problem where there was a strong dependency between the UI application (AngularJS front-end) and the back-end (C++ (web)server) on functionality that is only relevant for the front-end. Other front-end applications like the mobileApps, myDomoticz, Dashticz, etc. can not (and should not) use this and have to solve 'theming' within the front-end itself.
2) From a security standpoint installing a theme directly into the exposed 'www' directory poses risks when unintended files become _accessible_ through the webserver and could be exploited.
3) Developing a theme was not straight forward (not hard either, I agree) because as developer you need to know how the _acttheme alias_ works and can only test it using the Domoticz webserver.
4) From a user/browser perspective, a URI should always point to 1 single resource and not to something that changes when a different theme is chosen. Especially when the resources are considered static resources (like javascript, css, image files, etc.).
5)  Nowadays it is considered best practise to be able to split an application into (functional) parts that should be as independant as possible (micro-services architectures) for numerous reasons.

Theme's like Aurora did indeed work fine whatever the name of the installation folder.
1) As long as a simple theme did not need any (relative) URI linking to an other object within the theme, their was no problem.
2) But once it needs to point to some object withing the theme (for example loading another javascript file or including another css file, etc.), the theme file(s) need to know the relative path (URI) to that object. And here is where things got messy because the relative path should be something known to the theme (for example 'aurora') and _not_ something that depends on the name of the directory the user decided to use when installing the theme.

The extra files exposed when installing a theme should indeed not be used and probably won't by normal users. But that is the security problem, these thing not intended for use are often the ones that are _abused_ by non-normal users.

_End goals_: 
- Having a clear separation and no unneeded interdependency between the webapp/UI and the back-end service/server other than exchanging data through the exposed API's
- Ability to _serve_ the UI from anywhere using any (cloud)webserver without depending on the Domoticz own webserver.
- Remove the functionality of ServerSideInclude completely from the Domoticz webserver (although it works fine at the moment) as only used very little (2-3 cases), but needs to be maintained, tested, etc. while it also poses potential (future) security risks. Reducing the codebase and possible attack vectors is always a good idea in the long(er) term :)
- When someone (or group) decides the develop a new modern front-end for Domoticz, it is very very unlikely that they will make use of ServerSide Include functionality (no modern front-end framework uses it) and by removing it, it can not be used anyway.


Hope the above helps.

BTW: any idea how many themes are out there at the moment? And if/which are still having issue's? Are the maintainers aware of any issue and maybe working on it?

Issue #4944, opened 2021-09-26
dzvents generic_devices - error in error message for devices without lastUpdate


rrozema:
As reported in the forum: [https://domoticz.com/forum/viewtopic.php?f=59&t=36994](https://domoticz.com/forum/viewtopic.php?f=59&t=36994)

If a device gets triggered that doesn't have a valid lastUpdate value a message should be shown that this device will be skipped. The error message however crashes as it tries to call a function domoticz.utils._.str() that doesn't exist.

Error message listed instead of the intended informational message is:
`2021-09-13 12:56:31.895 Error: EventSystem: in /home/pi/domoticz/dzVents/runtime/dzVents.lua: ...oticz/dzVents/runtime/device-adapters/generic_device.lua:67: attempt to index a nil value (field '_')`

rrozema:
I created PR #4945 to fix the debug message not crashing the dzvents code. The lastUpdate value not being set is an issue in various hardwares and as such is out of the scope of this incident. According to the messages in the forum hardwares that fail to properly set lastUpdate are Dummy and zigbee2mqtt. But possibly others don't properly set it too.

gizmocuz:
This PR has been merged. Thank you for reporting and the PR!

PR #4945, opened 2021-09-26
dzvents - fix for #4944, skipping device without lastUpdate

rrozema:
This is a fix for issue 4944. It doesn't fix the lastUpdate not being valid, but it does fix dzVents correctly reporting the device skipped.

https://domoticz.com/forum/viewtopic.php?f=59&t=36994


gizmocuz:
Thank you!!

PR #4947, opened 2021-09-28
Fix for currentValue containing a JSON with the value field....

jvanderzande:
Fix for currentValue containing a JSON with the value field not updating the device. These messages are send by the ZWAVEJS2MQTT gateway.
It works for these type of payloads:
**mosquitto_pub -h 192.168.0.30 -t 'zwave/Screen1/38/0/currentValue' -m '50'**
This one is failing and is fixed by this PR: 
**mosquitto_pub -h 192.168.0.30 -t 'zwave/Screen1/38/0/currentValue' -m '{"time":1632837866709,"value":40}'**


gizmocuz:
thank you !!

Issue #4930, opened 2021-09-09
When creating a new sensor history shows up from year 0 to 2020


rtm-soft:
When creating a new dummy P1 smart meter I get history from year 0 to 2020, see picture below. Latest software on multiple different systems. Didn't have this problem before. Thanks Rene

![Screenshot 2021-09-09 at 17-30-55 Domoticz](https://user-images.githubusercontent.com/14265890/132765795-0fc72b9d-9780-46d6-80f1-c964f97498f0.png)

rtm-soft:
After 3 days this issue seems to fix itself. I can confirm this after reading the same issue in one of the issue reports

dobber81:
https://github.com/domoticz/domoticz/issues/4856 and https://github.com/domoticz/domoticz/pull/4830

PR #4948, opened 2021-09-28
Compare-chart should be able to handle new devices

### LINKED ISSUE #4907: Error log / graphs in a new energy (instant+counter meter)
When i add a new electric (instant + counter ) i can not view the log. 

image

**The log is working for old existing devices.**

My app cache was clear and i have flagged chrome as Disable cache in dev options

image

Comparing usage is showing from 0 to 2021

image

Let my know if you need more info.

========================

Version: 2021.1 (build 13422)
Build Hash: 56d1c95a0
Compile Date: 2021-08-16 13:41:21
dzVents Version: 3.1.8
Python Version: 3.7.3 (default, Jan 22 2021, 20:04:44) [GCC 8.3.0]

OS Raspbian buster

michahagg:
When a device is newly created, it does not have any data collected for it. Hence, the compare-chart should handle that situation and show nothing. This should fix #4907 and #4930.

### DIFF
--- a/www/app/log/CounterLog.js
+++ b/www/app/log/CounterLog.js
@@ -207,6 +207,9 @@ define(['app', 'lodash', 'RefreshingChart', 'DataLoader', 'ChartLoader', 'log/Ch
 
                                     function categoriesFromGroupingBy(groupingBy) {
                                         if (groupingBy === 'year') {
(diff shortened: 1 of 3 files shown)


gizmocuz:
Thank you !!

PR #4950, opened 2021-09-30
Recovered charts for Counter Incremental

michahagg:
This PR contains the following fixes/changes:
- The missing charts for Counter Incremental devices are now showing.
- Update of AngularJS to version 1.8.2


gizmocuz:
@michahagg , thank you for this fast fix! One more question, it seems when there is no data for the last xx days, the day/week graph is totally empty, is also does not display the table header 'Last days' or 'last week', is it possible to add this (back) ?
(Maybe if possible in the graph in the center a message 'no data' ?)

michahagg:
@gizmocuz I'll look into it!

michahagg:
Hi @gizmocuz it could be that PR #4954 fixes this problem. The chart area is now displayed properly with the title and all, but it is of course still empty. I'll look for a way to display some hint that there is no data to display.

PR #4951, opened 2021-10-01
Fix custom.css to find custom resources

michahagg:
With the new way of serving theme resources, the base-directory for the custom.css has changed.


gizmocuz:
Thank you!

gizmocuz:
I don't think this is the sollution .... there are other themes that are not distributed by default
The custom.css resides in the 'theme' folder, so anything that is refferenced should be directly accessable and not use ../

michahagg:
I was afraid so. Let's see if there is a better solution.

PR #4952, opened 2021-10-01
Fix for updated themeprocessing to serve other relative custom theme files correctly

kiddigital:
With PR #4942 a bug was introduced with serving certain theme specific assets like fonts, images, etc. due to incorrect processing.

PR #4951 was a temporary fix and this PR reverts those changes with a better solution.

Thx to @dobber81, @michahagg for finding/(temp)fixing/supporting

Should be all good now...


gizmocuz:
Thank you (all) for the quick fix and PR!

PR #4954, opened 2021-10-02
Charts: several fixes

### LINKED ISSUE #4856: Build 13314 empty graph issue with type General, Counter Incremental
@michahagg the problem with P1 smartmeter is solved (thanks for that!), but there now is an issue with an empty graph with type General, Counter Incremental.
Don't know if it is related to the the issue with the P1 graph, since it is different type sensor.
No error messages to provide a clue as to what is wrong.

TypeGeneralCounterIncremental

michahagg:
### Fixes
- Some attempts to invoke a function on an object/value, causing a chart not to be drawn (this could fix https://github.com/domoticz/domoticz/pull/4950#issuecomment-932106319)
- The value unit in compare-chart of Counter (Incremental) devices (was Wh, should be kWh)
- Show previous-year line in front of this-year columns
- Don't show zero-valued columns
- Don't adjust the values in compare-chart with the custom-adjustment in Counter devices (fixes https://github.com/domoticz/domoticz/issues/4856#issuecomment-932190890)

### DIFF
--- a/main/WebServer.cpp
+++ b/main/WebServer.cpp
@@ -13461,17 +13461,30 @@ namespace http
 					dbasetable = "Fan_Calendar";
 				else if (sensor == "counter")
 				{
-					if ((dType == pTypeP1Power) || (dType == pTypeCURRENT) || (dType == pTypeCURRENTENERGY) || (dType == pTypeAirQuality) ||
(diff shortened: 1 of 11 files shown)


gizmocuz:
Thank you!!

PR #4958, opened 2021-10-04
Charts: minor improvements

michahagg:
This PR contains the following:
- zero-valued columns in compare-chart are not shown
- when no data is available to display, the chart area displays an appropriate message
- further extend disabling text selection to improve panning/dragging chart on mobile device


michahagg:
Hi @gizmocuz this PR contains the feature you requested, to display a message when no data is available. If you have any suggestions/comments for improvement, please let me know.

michahagg:
@gizmocuz  could you please hold this PR for now? I found that the charts show all available series, even the ones that are not used with a particular device. I would like to fix that first. Thanks!

I'll put a comment when it's fixed and ready to be merged.

gizmocuz:
Sure, and thanks for looking into this !

michahagg:
@gizmocuz, it should now work in an acceptable manner. When no data is available, the temperature+ charts show only the temperature series plus an appropriate message. Then when data is available, it shows the series for which data is actually available.

Currently it is not possible to show all the series that are applicable with a specific device, because of how series are selected. It detects the presence of certain series from the data available. When there is no data, there is no way to determine which series would be present when data is present. I chose to only show temperature when no data is available, as temperature is always shown. Maybe in the future I could change the selection of the series in a chart by looking at the device type. But before that, I think I better migrate the other chart types to the auto-refreshing mechanism.

Anyway, it's ready to be merged.

Thanks!

gizmocuz:
Great PR, thanks for your time once again ! ;)

Issue #4960, opened 2021-10-07
Need to add "index.html" to path


kofec:
I'm using domoticz on Openwrt 21_02. Probably after the commit: https://github.com/domoticz/domoticz/commit/07e0c125ac314bec3f62f4c4c290ba7f1c90902c
I have to add "index.html" so the full path is eg:
http://192.168.1.35:8080/index.html#/About
Before it was:
http://192.168.1.35:8080/#/About

If I don't add "index.html" the website is not "interpreted" and I see code like:

```
<!DOCTYPE html>
<html manifest="html5.appcache">
<head>
		<meta charset="utf-8">
		<title>Domoticz</title>
		<meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0, user-scalable=no">
		<meta name="apple-mobile-web-app-capable" content="yes">
		<meta name="mobile-web-app-capable" content="yes">
		<meta name="apple-mobile-web-app-status-bar-style" content="black">

		<meta http-equiv="content-type" content="text/html; charset=utf-8">
		<meta name="description" content="">
		<meta name="keywords" content="">
		<meta http-equiv="X-UA-Compatible" content="IE=edge">
		<link rel="shortcut icon" href="favicon.ico">
		<link rel="apple-touch-icon" href="images/iphone-icon.png">
		<link rel="icon" sizes="196x196" href="images/logo.png">
		<link rel="icon" sizes="192x192" href="images/logo192.png">
...............................
i18n="Logout">Logout</span></a></li>
				</ul>
			</div>
		</div>
	</div>
	<div id="holder">
		<div class="container-fluid">
			<div alertarea></div>
			<div id="notification" style="display: none;"></div>
			<div class="bannercontent">
				<div id="main-view" ng-view></div>
			</div>
			<div id="copyright" style="display: none;" has-User>
				<p>&copy; 2012-{{currentyear}} Domoticz | WWW: <a href="http://domoticz.com/" target="_blank">Domoticz.com</a></p>
			</div>
		</div> <!-- /container -->
	</div>
</body>
</html>
```

The issue is seen on latest version:
About Domoticz
Version: 2021.1 (build 13690)
Build Hash: 96ca6028
Compile Date: 2021-10-06 08:48:51
dzVents Version: 3.1.8
Python Version: 3.9.7 (default, Feb 16 2021, 08:53:24) [GCC 8.4.0]

Uptime: 5 Minutes, 50 Seconds

[...]

PR #4962, opened 2021-10-10
Fix for serving already compressed sourcefiles and some refactoring

kiddigital:
This PR is to fix issue #4960 (caused by PR #4942).

That issue was found on the _Openwrt_ build where by default most webserver content files (the _www_ directory) are being zipped to save diskspace.

This PR also does some more refactoring of the webserver code base.
- Now it looks for the 'Cache-Control' header and when 'no-cache' is requested, it will serve the local file even if it doesn't seem modified.
- Adding the (optional) charset encoding (UTF-8) is now done at 1 place instead on multiple pieces of code


gizmocuz:
Could you please undo the utf-8 charges as this is needed for a lot more types and this has been done in the code.
Also worked previously without issue and should still work correctly.

kiddigital:
Hi @gizmocuz , I can do that but not sure if it is necessary.

I removed the code that adds the UTF-8 in several places as it was duplicate to reach the same goal. Also, when for whatever reason the content-type could not be determined based on the file extension (unknown or no extension) than the content-header was just wrong with only containing ';charset=UTF-8' without a proper actual type.

Also the way the extension and content-type where determined was different at different places and should now be at 1 place and consistent.

Do you foresee any issues somewhere? Or have certain (edge)cases in mind that I can test and verify if everything still works as expected?

Btw: for which additional content-types do you think we should add the charset type?

kiddigital:
@gizmocuz , removed the UTF-8 charset from the mime-types (so back to the original code) and have the 'add_header_content_type' function (reply.cpp) take care of adding it when needed. And made sure that other code pieces don't do it manual anymore but use this function.

kiddigital:
This PR is ready as far as I can tell. Once merged, people can create a new (beta)build for platforms like OpenWRT that work again with serving compressed content.

See issue #4960

gizmocuz:
Thanks! I just want to verify the workings before merging, hope to do shortly!

gizmocuz:
Something is not correct.... For example go to the settings, you will see there are no languages to select anymore

kiddigital:
Weird.. you think that it is related to this change? I will look into it asap.

Let me know if you find anything else.

Thx for testing.

kiddigital:
@gizmocuz , found the issue. Include processing didn't work properly anymore because content was compressed before include.

Now works fine again.

gizmocuz:
Great, so we do not the other PR ? There are multiple places where these comboboxes are used

kiddigital:
@gizmocuz , for this PR the other PR is not needed indeed.

The other PR is to remove the only ServerSideInclude from the Settings screen. That would make it possible to serve the settings screen from another webserver instead of the Domoticz webserver itself..  Could not find any other place where the language combobox gets embedded other than on the settings screen. The _switchtypes_ and _metertypes_ combo's are used still at a few (found 4) places.

So if we do or don't merge the other PR is irrelevant for this PR. This one fixes the open issue.

gizmocuz:
Thank you for the change! Maybe you could tweak the other PR to not touch the settings page, but leave the GetLanguages call, and if you want you can do the same for the other ServerSideIncludes...
Going to test/merge this soon

gizmocuz:
Thank you !!

PR #4963, opened 2021-10-11
OpenWeatherMap: Fix to resolve OWM City for default location. Used by OWM based forecast screen

kiddigital:
Small PR to solve an issue when specifying 0 (zero) as the OWM location, which means that Domoticz own home location should be used. The OWM module did already use that for the data, but it did not try to resolve the lat/lon to an OWM CityID. Therefor when selecting _'Use OpenWeatherMap for UI forecast screen.'_ , it did not work. The _old_ Darksky iframe kept being used instead of the OWM embedded widget.


gizmocuz:
Thank you!!

PR #4964, opened 2021-10-11
Added cover_has_stop_command to MQTT discovery

moroen:
Adds a setting "cover_has_stop_command" to the MQTT config topic for blinds. 

If set to "true" a blind with STYPE_BlindsPercentageWithStop will be created, when set to false create a blind  STYPE_BlindsPercentage.

Defaults to "true" in accordance with currenct functionality.

Config json-object example:
{

  "command_topic": "tradfri/65560/blind/set",
  "position_topic": "tradfri/65560/blind",
  "set_position_topic": "tradfri/65560/blind/set",
  "cover_has_stop_command": false,
  "unique_id": "tradfri_65560_blind",
  "device": {
    "identifiers": [
      "tradfri_65560_blind"
    ],
    "manufacturer": "IKEA of Sweden",
    "model": "KADRILJ roller blind",
    "name": "Stuevindu",
    "sw_version": ""
  },
  "name": "Stuevindu"
}

### DIFF
--- a/hardware/MQTT.cpp
+++ b/hardware/MQTT.cpp
@@ -2974,13 +2974,14 @@ bool MQTT::SendSwitchCommand(const std::string &DeviceID, const std::string &Dev
 
 		if (command == "On")
 		{
-			level = pSensor->position_open;
-			szValue = pSensor->payload_open;
+			level = pSensor->position_closed;
+			szValue = pSensor->payload_close;
+			
 		}
 		else if (command == "Off")
 		{
-			level = pSensor->position_closed;
-			szValue = pSensor->payload_close;
+			level = pSensor->position_open;
+			szValue = pSensor->payload_open;
 		}
 		else if (command == "Stop")
 		{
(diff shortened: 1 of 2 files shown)


gizmocuz:
Thank you for this PR!
This "cover_has_stop_command" is that documented on the HA website ?

moroen:
No, I haven't found a official way to specify if a cover supports a stop command or not. It might be better to use the stop_payload in some capacity. I can take a further look on how (if?) HA supports covers both with and without support for a stop command.

moroen:
So, it appears that if payload_stop is set to null or is empty, the stop button in HA is disabled, though it's still visible in the interface. I would argue that having a visible, but disabled "stop" button is bad interface design and I suggest that sending an empty or null payload_stop should create a STYPE_BlindsPercentage device. Specifying a payload_stop or not specififying the payload_stop at all, should create a STYPE_BlindsPercentageWithStop.

If this is acceptable, I'll gladly modify the PR!

gizmocuz:
@moroen , Sounds like a good idea.
Could you also undo the changes in SendSwitchCommand ?
Maybe revert the current change and only use the check if stop_payload quals null?

moroen:
I'm closing this pull request, and I'll issue a new pull request with these changes!

PR #4966, opened 2021-10-12
MercedesMe: Changed variable type to unit32_t to prevent issues on 32-bit platforms

kiddigital:
Fix so RefreshToken is properly stored and retrieved in the UserVariables table on 32-bit platforms (like RaspberryPi)

### DIFF
--- a/hardware/eVehicles/MercApi.h
+++ b/hardware/eVehicles/MercApi.h
@@ -57,7 +57,7 @@ private:
 	std::string m_accesstoken;
 	std::string m_refreshtoken;
 	std::string m_uservar_refreshtoken;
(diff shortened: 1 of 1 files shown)


gizmocuz:
Maybe it needs to be uint32_t ?

kiddigital:
Could be as well, but isn't _int_ not more suitable as it respects the platform (so either 32 or 64 depending on the platform)? Value is indeed an unsigned value (db field).

kiddigital:
@gizmocuz , changed it to an uint32_t (instead of replacing the uint64_t with int)

gizmocuz:
Thank you!! Yes it's better to use the correct value type

PR #4971, opened 2021-10-14
Fix missing thermostat-climate thermostat_command_template

heggink:
The existing code was erroneously using temperature_state_template for the setsetpoint response formatting whereas it should use temperature_command_template. For my zwave and zigbee devices, the template is not defined and the response payload should just consist of a single value instead of a json message. Zwavejs seems to accept a json message but zigbee2mqtt does not and errors out. This PR fixes the issue where both now work.


gizmocuz:
Thank you!!

PR #4972, opened 2021-10-14
Use uint32_t instead of std::string to uniquely reference EnOcean nodes

efdeb:
Use uint32_t instead of std::string to uniquely reference EnOcean nodes to :
 - better conform to EnOcean standard (see EnOcean ERP1 v1.2 and EnOcean EEP v3.1),
 - simplify the code,
 - avoid numerous unnecessary conversions.

(Next step should be to update EnoceanSensors table in domoticz database accordingly)

Fix a few things :

- replace atoi by std::stoul, when converting strings to unsigned integers
- replace %d by %u for unsigned integers in safe_query format strings
- improve management of virtual switches
- improve management of RPS telegrams send by VLD nodes
- fix some log/debug messages

### DIFF
--- a/hardware/EnOceanEEP.cpp
+++ b/hardware/EnOceanEEP.cpp
@@ -644,35 +644,18 @@ const char *CEnOceanEEP::GetEEPDescription(const int RORG, const int func, const
 	return ">>Unkown EEP... Please report!<<";
 }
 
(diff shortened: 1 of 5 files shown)


gizmocuz:
Thank you again!

PR #4975, opened 2021-10-15
Improve DEBUG messages when opening DB at startup (rebased replacement of PR #4915)

efdeb:
Hi @gizmocuz,

This PR is a rebased version of the previous : Improve DEBUG messages when opening DB at startup #4915 

You remember we had a discussion about the first version of this PR I proposed, about the messages needing to be visible for a normal user and additional information which could be of interrest for a power user or developper for troobleshooting.

I thus decided not to change the existing behaviour for the normal user, and only adding additional debug information for power users or developpers.

Unchanged : a normal user has only errors messages in his log when :

    database file cannot be opened (old line 687)
    database version is newer than the one expected (old line 707)

Added debug information :

    database file cannot be opened (line 689) => **database path**
    database is empty (line 703) => **path and version of the created database**
    database version is the one expected (line 708) => **path and version of the database**
    database version is older than the one expected (line 710) => **path and current/upgraded version of the database**
    database version is older than the one expected (line 714) => **path and current/expected version of the database**

Of course I can change that, just tell which line to move to the log, because I do not catch what you call "wedding database(version)"...

### DIFF
--- a/main/SQLHelper.cpp
+++ b/main/SQLHelper.cpp
@@ -38,6 +38,7 @@
 #define __STDC_FORMAT_MACROS
 #include <inttypes.h>
 
+#define DB_NAME "domoticz.db"
 #define DB_VERSION 149
 
 extern http::server::CWebServerHelper m_webservers;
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thank you for this PR. These log messages are not needed and could only confuse/scare users.
The reason why I also used SQLite3 in the sentence is that this does display the format, so in the slight case that this error occurs the users has a good point where to look, what tools to use to debug/repair

PR #4976, opened 2021-10-15
unique_ptr conversions

neheb:
This is C++14. new can be replaced with make_unique.

### DIFF
--- a/webserver/connection.cpp
+++ b/webserver/connection.cpp
@@ -263,8 +263,8 @@ namespace http {
 			{
 #define FILE_SEND_BUFFER_SIZE 16*1024
 				if (!send_buffer_)
-					send_buffer_ = new uint8_t[FILE_SEND_BUFFER_SIZE];
-				size_t bread = static_cast<size_t>(sendfile_.read((char*)send_buffer_, FILE_SEND_BUFFER_SIZE).gcount());
+					send_buffer_ = std::make_unique<std::array<uint8_t, FILE_SEND_BUFFER_SIZE>>();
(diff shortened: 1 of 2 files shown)


gizmocuz:
@neheb , I do thank you for your efforts to contribute to this project, but your recent merged PR's are causing Domoticz too crash. This has happened multiple times before as well.
Either the changes are incorrect or not tested good enough. This could be because you have limited hardware to test with (No RF 433/868/zwave/ZigBee/MQTT) or not many realtime devices.
Please have a look at the latest commits where 4 of your PR's have been reverted to resolve a crash that happened in minutes after running.
Maybe you can review them again.
For now I find it to risky to accept any of your patches.

neheb:
Most likely the unique_ptr commit was faulty. The rest should be fine.

gizmocuz:
Thank you for this PR. I needed to remove the duplicate FILE_SEND_BUFFER_SIZE define. Did you not get a compile warning during testing ?

neheb:
I believe I did but ignored it IIRC.

gizmocuz:
Not a good thing to ignore compile warnings ;)

neheb:
Sure. OTOH the codebase is full of them ;).

kiddigital:
> Sure. OTOH the codebase is full of them ;).

Interesting, as when I compile under Linux (Ubuntu 21 gcc 10 at the moment) I don't see any on my clean environment.

If I encounter one or more, I fix them asap, so it stays clean (for me at least).

What warnings do you get in your environment? What (additional?) settings do you give the compiler that could generate these warnings?

I agree with @gizmocuz that we should not ignore warnings as they are there often for a reason.

gizmocuz:
I don't see them as well... only when I compile under Windows with a 64bit platform...

neheb:
Probably need to enable Wextra with gcc.

PR #4984, opened 2021-10-20
fix for cover devices status updates received.

jvanderzande:
Fix for cover devices taking  JSON "position" into account for set level and config fields "pos_clsd" & "pos_open"


gizmocuz:
Thank you!!

PR #4987, opened 2021-10-23
Webserver: add getLanguages command

kiddigital:
This PR replace the way the combo box for the languages on the Setup screen is populated based on the available languages.

Instead of the current/old method of using a ServerSide Include (#embed tag in the View) where the view is _enriched_ with content from the ServiceSide Include method _DisplayLanguageCombo_ , now the UI uses a new _GetLanguages_ command in the same way as it does for Themes and TimerPlans.

Note for later: The _RegisterIncludeCode_ and the function code for both the Languages and Timerplans can be removed (after this PR is merged) because both are not used anymore (actually the Timerplans include isn't used anymore already).

Reason to do this is to make it possible to fully serve the _static_ content from the current UI from a different webserver without depending on the Domoticz internal webserver. And also possible for other (future?) UI's to do all the settings.

### DIFF
--- a/main/WebServer.cpp
+++ b/main/WebServer.cpp
@@ -375,6 +375,8 @@ namespace http
 			RegisterCommandCode(
 				"getlanguage", [this](auto &&session, auto &&req, auto &&root) { Cmd_GetLanguage(session, req, root); }, true);
 			RegisterCommandCode(
+				"getlanguages", [this](auto &&session, auto &&req, auto &&root) { Cmd_GetLanguages(session, req, root); }, true);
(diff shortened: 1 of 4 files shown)


kiddigital:
This PR is independant to PR #4962 but in a way related as that PR fixes the Server Side Include processing where this PR actually removes the need for Server Side Includes when it comes to getting and selecting the language(s).

kiddigital:
@gizmocuz , removed the changes to the settings page. This now just adds the new getLanguages command

gizmocuz:
Thanks!!

PR #4988, opened 2021-10-25
Update autodiscover climate  creating  devices when config is received.

jvanderzande:
This PR will enable device creation for climate devices at auto discovery time. 
Also the check for already discovered devices is commented to allow for device updates while running to avoid the need for a restart when a device changes and send a new "config" topic.


gizmocuz:
Thank you!!

PR #4989, opened 2021-10-25
Fix cover perc recalc

jvanderzande:
The cover percentage recalculation also needs to be applied when there is just a value without template usage.


gizmocuz:
Thank you!!

PR #4991, opened 2021-10-26
More fixes for the cover devices:

jvanderzande:
- proper recalc fromm 0-100 to pos_c value for both input and output message


jvanderzande:
Will update to latest commit and resubmit

PR #4992, opened 2021-10-26
More fixes for cover:

jvanderzande:
- proper recalc fromm 0-100 to pos_c value for both input and output message


jvanderzande:
Will apply the required patches to the current development branch state and make a new PR

PR #4995, opened 2021-10-26
More changes for cover:

jvanderzande:
- proper recalc from 0-100 to pos_c value for both input and output message


gizmocuz:
Thank you again!

PR #4997, opened 2021-10-27
Empty payload_stop creates STYPE_BlindsPercentage

moroen:
In accordance with HA, an empty payload_close now creates a STYPE_BlindsPercentage device. If non-empty, or not defined, a STYPE_BlindsPercentageWithStop is created instead.


gizmocuz:
Thank you!

PR #4998, opened 2021-10-28
Add running automated tests to GitHub action for PR workflow (and fix buildnr and more)

kiddigital:
This PR extends the GitHub action for _pull_requests_ with:

- Setting up some environment variables
- Make sure to fetch full GIT history to be able to create the proper BUILD number (was wrong)
- Runs the available automated Functional tests (added with PR 4962)
  - Currently does __NOT _fail___ when not all tests succeed (but they should)
  - Other automated tests (UI and dzVents) are not yet added (on to-do list)
- Creates an _Artifact_ of the build that can be downloaded
  - Could be used for example with Docker so without the need to compile, people can create a Docker container running a PR specific build.
  - This would allow people to start testing a PR build _before_ it is merged into development/Beta
  - To-do: create an example Docker file on how to use these PR builds
  - PR build artifacts are only stored for a period of _7 Days_
 
See [here](https://github.com/kiddigital/domoticz/actions/runs/1402535308) for the output of a run.

_I suggest to make it mandatory that all tests succeed in the near future forcing developers to act on failed tests._

### DIFF
--- a/.github/workflows/pull_request_check.yml
+++ b/.github/workflows/pull_request_check.yml
@@ -64,7 +64,7 @@ jobs:
         run: |
           cd $GITHUB_WORKSPACE
           cd ..
-          git clone https://github.com/OpenZWave/open-zwave.git open-zwave-read-only
+          git clone https://github.com/domoticz/open-zwave.git open-zwave-read-only
           cd open-zwave-read-only
           make
           sudo make install >> /dev/null 2>&1
@@ -75,7 +75,6 @@ jobs:
           cd $GITHUB_WORKSPACE
           cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_LIBRARY_PATH=open-zwave-read-only CMakeLists.txt
           make
-          cat appversion.h.txt
 
       # Run automated tests
       - name: test domoticz
@@ -88,14 +87,13 @@ jobs:
         continue-on-error: true
 
       # Upload artifacts
-      - uses: actions/upload-artifact@v2
-        with:
-          name: DomArtifacts-${{ env.BUILDNR }}-${{ env.BRANCH_NAME }}
-          path: |
-            domoticz
-            appversion.h.txt
-            test/
-            www/
-            lib/
-            !**/*.tar.gz
-          retention-days: 7
+      #- uses: actions/upload-artifact@v2
+      #  with:
+      #    name: Domoticz-${{ env.BUILDNR }}-${{ env.BRANCH_NAME }}
+      #    path: |
+      #      domoticz
+      #      appversion.h.txt
+      #      www/
+      #      lib/
+      #      !**/*.tar.gz
+      #    retention-days: 7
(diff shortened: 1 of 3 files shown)


gizmocuz:
There is no need for FULL history on checkout, the PR needs to build a.s.a.p., it's to check if the code compiles, not to check if the code is working correct or generate build numbers ;)

gizmocuz:
I think the tests should be run by the developers before committing their work ?
It depends how long the test takes (and this extra step sudo apt-get install python3-pytest python3-pytest-bdd)
Uploading build artifacts for PR should also not be done

gizmocuz:
Basically, quite happy as it is now... sorry

kiddigital:
hm.. had hoped that it would be received more enthusiastically :)

Let my explain why I am enthusiastic about it:

- Running automated tests automatically politely forces PR developers to be more aware of what they submit as these tests might reveal that something, for them unexpectedly, is broken (at least does not function as intended).
- Even when you as a developer are not (yet) that _experienced_ in automated testing, it helps and assists a lot in creating and maintaining code functionality.
- Not every developers that makes changes has a full overview of all functionality Domoticz offers (at least I know that I don't) so it is hard to determine if changes did not affect something that you are not familiar with. These tests help you.
  - Good example, the automated tests of _dzVents_ (thanks to the great work of the late rwaaren ). Hard to know as much about dzVents as the creator/developer and so when making changes, these tests help a lot to determine of not something is broken unintentionally.
- When a developer (or developers) adds new functionality (good example could be the MQTT Autodiscovery feature), writing automated tests for it helps in testing (prevent regressions during development process and later on for example) AND also helps other (new) developers in getting to understand the functionality as intended. And again helps them to determine if they did not break anything when wanting to add something.

About your concerns:

- _Just_ checking the code compiles (validating it is syntax correct C++) is relevant but checking of the code is working correctly seems even more important to me (remember a few PR's that had to be reverted because they compiled correctly but did not function properly).
- The reason for a FULL history checkout is to be able to determine the build number (otherwise rev-list always equals 1)
  - This only adds about 15 seconds to the whole workflow
  - And yes can be skipped, but the build number is used in the Artifact name
- Agree that ideally developers have done proper testing, but each developer is different, tests differently, has/reserves limited time to test and test only the things they _think_ are affected by the changes and have to be tested. And than they _only_ test what they can think of. Therefor these tests, how well intended as they are, often don't cover enough (and take precious time).
  - But running automated tests to assist the developers to point out possible things they have overlooked, helps them a lot. Yes, they can run these tests locally and probably should.
  - Adding the dependencies only adds about 4-5 seconds to that step.
- Running the (just a small amount just now) tests now only takes about 3 seconds, but running more tests will in the future hopefully take more time, maybe even a few minutes.
  - But as Compiling Domoticz (30+ minutes) and OpenZWave (+/- 3 minutes) take more than 30+ minutes, adding a few minutes (even it becomes 3 minutes) is less than 10% of the total workflow time, so I think not really a big deal. And I also don't think that developers that submit a PR are _waiting_ anxiously for the PR workflow run result before doing something else or are waiting to fix any found Compile errors. They will get notified anyway.
-  We can skip uploading the Artifact for sure. But it doesn't really cost anything and brings a possible gain, the easier ability to test a PR _before_ it has been merged by others other than the PR submitter, _without_ the need to build/compile the PR yourself.

Happy to discuss further if needed (on Slack?)

gizmocuz:
Okey that all makes sense... only not so happy with the artifacts

kiddigital:
Once merged, I will pickup the to-do to include running the already available tests for dzVents and the front-end tests from @michahagg

gizmocuz:
Thank you for this PR!!

PR #5003, opened 2021-11-01
MQTT Cover Blinds inverted patch

jvanderzande:
MQTT Cover Blinds inverted patch so both the Open/Close and percentage will be inverted when selected. 
This should fix the issues we have when a shutter operates in a reversed way.


gizmocuz:
Thank you!!

PR #5004, opened 2021-11-01
Fix for brightness_topic payload.

jvanderzande:
This change will just set the payload for configs with a brightness_command_topic & brightness_value_template defined:
On will just set the payload to brightness_scale when the template is "value" else the found "keyword": brightness_scale .
Off will just set the payload to 0 when the template is "value" else the found "keyword": 0.
Set Level will just set the payload to sLevel when the template is "value" else the found "keyword": sLevel.


gizmocuz:
Thank you!!

PR #5005, opened 2021-11-01
The Things Network (TTN): Switch to new V3 stack

kiddigital:
This PR makes the TTNMQTT hardware module compatible with the new Community Edition (v3) of The Things Network.

It is only handling v3 as v2 will be turned off per 1st December 2021 (very soon).

The amount of LoRa devices I have been able to test it with is very limited, _so please test it with your devices_.

It should work (as it did before) for devices that use the [Chayenne LPP](https://www.thethingsindustries.com/docs/integrations/payload-formatters/cayenne/) encoded payloads.

You can use custom encoders and the module will try to determine if it recognizes measurements. As the module looks for certain names of the measurements, it is now also possible to specify aliasses for these measurement names using an aliasses file in JSON format (see `ttnmqtt_aliasses.json`). Most LoRa products that are registered with TTN (see [GitHub repo](https://github.com/TheThingsNetwork/lorawan-devices) have default custom en/decoders. But they all might use different names for the sensor reading, so now you can specify if these names should be interpreted as aliasses for another name that this module recognizes.

### DIFF
--- a/hardware/TTNMQTT.cpp
+++ b/hardware/TTNMQTT.cpp
@@ -121,6 +121,20 @@ bool CTTNMQTT::StartHardware()
 		}
 	}
 
+	// Check if there is a Certificate file given to use for TLS encryption
+	std::vector<std::string> strarray;
+	StringSplit(m_CAFilename, ";", strarray);
+
+	if (strarray.size() == 3 || strarray.size() == 4)
+	{
+		m_CAFilename = strarray[0];
+	}
+	else
+	{
+		Log(LOG_STATUS, "Unexpected parameters given! Unable to determine CA Certificate filename (%s)", m_CAFilename.c_str());
+		m_CAFilename = "";
+	}
+
 	//force connect the next first time
 	m_IsConnected = false;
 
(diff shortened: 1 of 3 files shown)


kiddigital:
@gizmocuz , I was thinking about using YAML instead of JSON for the Aliasses config but that would need a good YAML library which is not in Domoticz yet.

Would it be an option to include [this](https://github.com/jbeder/yaml-cpp/blob/master/README.md) library as an external lib? Also available in VCPKG 😁

Reason to use YAML is the better readability for Users and with JSON-Schema the YAML can be validated nicely.

Let me know if that is an option and then I will give it a try.

kiddigital:
@gizmocuz , I prefer to merge this PR first before thinking to switch to use YAML as the deadline for the current V2 TTN environment is 1 Dec 2021! In a few weeks 🤨 So if this is merged, it gives people a little time to test it before the mandatory switch.

The YAML thing can de done later if decided to go that route.

gizmocuz:
Thank you!!

PR #5007, opened 2021-11-02
Reverted the ON Level to 255 in stead of max level

jvanderzande:
Reverted the ON Level to 255 to ensure it goes back to the last level in stead on max.


gizmocuz:
Thank you!!

PR #5008, opened 2021-11-02
Fix for Brightness update status not showing xx%

jvanderzande:
Fix for Brightness update status not showing xx%


gizmocuz:
Thank you!

Issue #5013, opened 2021-11-05
Python plugin framework crashes on badly-formed http response


bloob00k:
Version 2021.1 (build 13830)
Raspberry pi

My python plugin calls the Drayton Wiser API.  They recently updated their firmware / API such that it now returns an invalid response; specifically, the response uses chunked encoding, and has *two* terminator chunks.  I.e. it ends like this
0\r\n
\r\n
0\r\n <--- This and the next line should not be here.
\r\n

This causes the framework to crash.  Whilst the bug is clearly in the API rather than domoticz, the framework should nevertheless be more resilient to badly-formed data since it can't control it.  I have a patch to fix this which I will submit as a PR.

bloob00k:
PR #5014

waltervl:
Can be closed as 5014 is implemented with commit https://github.com/domoticz/domoticz/commit/3822d7a21625b4923616f8271a92ef47b7a04215

PR #5014, opened 2021-11-05
Be (slightly) more robust in the face of malformed http response

bloob00k:
See issue #5013

Python plugin using the HTTP protocol can cause the framework to crash if the HTTP response is badly formed.  Eg a recent update to third party Drayton Wiser API now causes it to return a badly-formed HTTP response; specifically, the response uses chunked encoding, and has two terminator chunks. I.e. it ends like this
0\r\n
\r\n
0\r\n <--- This and the next line should not be here.
\r\n

This causes the framework to crash. It processes the "real" message correctly, consuming all the data up to and including the proper terminator chunk.  However, the second chunk is left in the buffer, and gets processed as if it was the start of the next message.  The code assumes that the transaction either starts with a valid HTTP response line (HTTP/1.1 200 OK) or valid HTTP request (GET / HTTP/1.1).  Since it doesn't start with the string "HTTP" it processes it as if it was a request.  When it tries to extract the URL it crashes because it looks beyond the end of the string for a URL which isn't there.

Whilst the bug is clearly in the API rather than domoticz, the framework should nevertheless be more resilient to badly-formed data since it can't control it. This patch goes some way towards addressing that.  Ideally it would just throw away the extraneous data altogether, but I can't see an easy way to do that, so instead it "safely" parses the invalid data and passes it on to the plugin to deal with.  I.e. in this case the plugin will receive a message with
Verb: 0
URL: <empty string>
No headers
No payload

The plugin is unlikely to be expecting this either, but is better placed to decide what to do with it than the framework (given that the invalid response could be absolutely anything, it just so happens in this case it is a terminator chunk).

### DIFF
--- a/hardware/plugins/PluginProtocols.cpp
+++ b/hardware/plugins/PluginProtocols.cpp
@@ -689,7 +689,12 @@ namespace Plugins {
 					if (PyDict_SetItemString(DataDict, "Verb", pObj) == -1)
 						_log.Log(LOG_ERROR, "(%s) failed to add key '%s', value '%s' to dictionary.", "HTTP", "Verb", sVerb.c_str());
 
-					std::string		sURL = sFirstLine.substr(sVerb.length() + 1, sFirstLine.find_first_of(' ', sVerb.length() + 1));
+					// Beware - the request may be malformed; so make sure there is more data to process before trying to parse it out
+					std::string sURL;
+					if (sFirstLine.length() > sVerb.length())
+					{
+					        sURL = sFirstLine.substr(sVerb.length() + 1, sFirstLine.find_first_of(' ', sVerb.length() + 1));
+					}
 					PyNewRef pURL = Py_BuildValue("s", sURL.c_str());
 					if (PyDict_SetItemString(DataDict, "URL", pURL) == -1)
 						_log.Log(LOG_ERROR, "(%s) failed to add key '%s', value '%s' to dictionary.", "HTTP", "URL", sURL.c_str());


gizmocuz:
@bloob00k , thank you for this PR. Maybe we should log a error when it's malformed and don't continue the function ?

bloob00k:
I think in principle that is a good idea, but there are several complications.  The function doesn't have enough state or context to be able to tell where one request / response finishes and another starts.  By which I mean that in this particular example, all of the data being processed is part of the same single response from the device; the response contains some trailing garbage.  Everything up to the trailing garbage is well-formed, valid, and useful.  The function successfully parses it out and forwards it to the plugin as a message, and the plugin can process that.  Then when the function sees the garbage it is unaware that it is part of the previous server response; if it was, then logging an error and throwing it away would absolutely be the best course of action.  But as it is, it doesn't know that the response to the last request has been completely received and processed, and that there has been no intervening request since.  It just looks at the first four bytes of the buffer, and if they are "HTTP" then it processes the buffer as if it was a response, and otherwise processes as it as if it is a request.  So here, to the function, it looks like there is a new malformed request, when in fact it it is trailing garbage on the previous malformed response.  Since it is not actually a request it would be safe to throw it away and log an error.

What I am worried about is behaviour in other permutations.  There are so many different ways in which the data could be malformed.  Eg the start of the original response is wrong in one of many different possible ways; it is actually a request from the plugin which is malformed; there is a chunked response but one of the chunk sizes is wrong, so everything gets out of sync in the parsing, or there is no terminating chunk.

It might be sufficient for the code path that handles chunked encoding to "drain" all the data from the buffer.  At the moment, it takes no notice of the connection state.  It will continue trying to read data, and not processing what it has received so far, until it has received the terminating chunk, even if the connection has been closed and the response is truncated.  And similarly it continues to process data from the response even after a complete response has been decoded and processed (i.e. the trailing data in this case).  If it specifically checked (a) is there still more data available - if not, then return whatever of the message we have received so far; and (b) have I already received a complete response - if so, throw away any additional data, that would certainly address a number of potential error situations.  But there are still a lot of assumptions that could trip it up; to avoid those it would need a more reliable way to know whether this is a request or response than simply looking at the start of the buffer.

Fundamentally, I think the plugin is the thing which knows best how to handle bad responses.  The framework needs to protect itself against crashing if something is malformed, but I would leave it to the plugin to determine how to deal with that.

Checking the state of the connection / more data available would allow us to do what you suggest in this case; I can look into that when I have a bit of time if you like.  It would certainly be an improvement.

dnpwwo:
The plugin protocol layer is designed to do what the name suggests, it encodes and decodes protocols for the plugins.  It does not attempt to handle data that is invalid from a protocol perspective because there are a huge number of cases that would need to be handled.  If a plugin is receiving invalid data either because the transport is error prone or because the connected device has a protocol error then I would recommend that the plugin author falls back to the 'None' protocol and then takes care of the messages themselves within the plugin.

I would recommend NOT blurring the line between the Transport and Protocol layers in the way you are suggesting because it will introduce a lot of complexity.  Not all transports have active connections when a message is being processed, inbound UDP/IP for example, by its nature, is fundamentally connectionless (ICMP as well from memory).  Today any Protocol can be used with any Transport through the magic of polymorphism and that needs to be maintained.

bloob00k:
I actually think that is what I was originally proposing in my PR.  The PR fundamentally just avoids crashing if the data is malformed.

Regarding separation of protocol and transport,  I agree.  But you need to consider that for HTTP/1.0 they are unavoidably linked, and that link is already in the code for the non-chunked path, albeit in a way that does not directly bind the protocol decoding to a specific transport.  That is, HTTP/1.0 does not require a content-length header, and in the absence of one the end of the message is denoted by the end of the connection.  The existing code handles this thus:
if ((m_ContentLength == sData.length()) || (Message->m_Buffer.empty()))
I haven't looked closely, but I assume that m_Buffer.empty() is only true when the connection is closed.  I don't know how that would come to be true on a UDP connection.

So, for chunked encoding we _could_ process data until m_Buffer.empty() is true, rather than stopping as soon as the terminator chunk is received, without creating any new dependency on the underlying transport - unless persistent connections are used.  But again I think it is good enough just not to crash, as per the PR.

Using the None protocol is fine if as you say the device is known to be error prone, etc.  But in this case the plugin has been working fine for a couple of years, and then one day the vendor did a remote firmware update that introduced a protocol error, and all of a sudden the framework (not the plugin) started crashing.  It seems to me that having to suddenly rewrite the plugin with the None protocol handler just to get around that until the vendor hopefully issues another firmware update is not very friendly.

Are there better ideas on how to address this?

gizmocuz:
This PR prevents a crash, so thats good. But I suggest to add a error message that something is incorrect. In the mean time you could ask your vendor to make a new firmware update fixen this issue maybe

gizmocuz:
Thank you for the PR. I made the change and added a log line

bloob00k:
Thanks.  Sorry I didn't get around to adding the error message, I was away a lot recently.

Issue #4936, opened 2021-09-17
Domoticz process segfaults when saving settings


siemhermans:
**System Information**
Version: 2021.1 (build 13499)
Build Hash: eee4de721
Compile Date: 2021-09-16 07:36:53
dzVents Version: 3.1.8
Python Version: 3.8.10 (default, Jun 2 2021, 10:49:15) [GCC 9.4.0]

```
user@nuc:~/domoticz$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 20.04.3 LTS
Release:	20.04
Codename:	focal
```

**Issue**
When saving settings in the web GUI menu, the Domoticz process segfaults. Specifically, I'm trying to enable allowing new hardware.

```
Sep 17 20:27:56 nuc domoticz[722663]: Fatal Python error: Segmentation fault
Sep 17 20:27:56 nuc domoticz[722663]: Thread 0x00007fe2215ad9c0 (most recent call first):
Sep 17 20:27:56 nuc domoticz[722663]: <no Python frame>
Sep 17 20:27:56 nuc domoticz[722663]: 2021-09-17 20:27:56.976  #033[1;31mError: Domoticz(pid:722663, tid:722695('WebServer_8080')) received fatal signal 11 (Segmentation fault)#033[0;0m
Sep 17 20:27:56 nuc domoticz[722663]: 2021-09-17 20:27:56.976  #033[1;31mError: siginfo address=0x3e8000b06e7, address=0x7fe2224a324b#033[0;0m
Sep 17 20:27:57 nuc domoticz[722663]: 2021-09-17 20:27:57.595  #033[1;31mError: Thread 7 (Thread 0x7fe21e0e4700 (LWP 722695)):#033[0;0m
Sep 17 20:27:57 nuc domoticz[722663]: 2021-09-17 20:27:57.595  #033[1;31mError: #0  0x00007fe222205dff in __GI___wait4 (pid=722754, stat_loc=0x7fe21e0e0b64, options=0, usage=0x0) at ../sysdeps/unix/sysv/linux/wait4.c:27#033[0;0m
Sep 17 20:27:57 nuc domoticz[722663]: 2021-09-17 20:27:57.595  #033[1;31mError: #1  0x000055bf870d36b7 in dumpstack_gdb(bool) ()#033[0;0m
[...]

PR #5016, opened 2021-11-07
Improved robustness of processing Settings update

kiddigital:
This PR made changes to the cmd_PostSettings to make the processing of updated settings more robust.

Some users (see forum topics and issue #4936 for example) report that they have crashes when trying to save Settings.

Although it is not really clear why these crashes occur, it makes sense that the Domoticz back-end handles wrong data more gracefully.

The code has been refactored, cleaned-up and re-arranged to be more readable and better error handling is added including some extra information in case of caught exceptions.

And a (Linux) compiler warning has been tackled.

### DIFF
--- a/hardware/plugins/PluginManager.cpp
+++ b/hardware/plugins/PluginManager.cpp
@@ -143,7 +143,7 @@ namespace Plugins {
 			m_InitialPythonThread = PyEval_SaveThread();
 
 			m_bEnabled = true;
-			_log.Log(LOG_STATUS, "PluginSystem: Started, Python version '%s', %d plugin definitions loaded.", sVersion.c_str(), m_PluginXml.size());
+			_log.Log(LOG_STATUS, "PluginSystem: Started, Python version '%s', %d plugin definitions loaded.", sVersion.c_str(), (int)m_PluginXml.size());
 		}
 		catch (...) {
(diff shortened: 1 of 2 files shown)


gizmocuz:
The settings are applied but it seems to crash after it. When restarting the settings are applied.
So maybe it's not inside this function ?

kiddigital:
If you are just referring to the mentioned example issue, it is partly true.

During the process of storing the settings, for a number of settings not only the DB update is done but also one or more specific setting related function calls are executed from within this command. In this example issue it is likely that one of these calls (reload Python plugins I guess) is crashing.

But as the initiative to perform this call is from the PostSettings command and the change of the settings requires another function to be executed, it makes sense to consider the PostSettings call as failed (although most or even all DB updates have been done).

More in general this call should not crash and handle the data better. Also telling the User that if something failed, the state of the system is in limbo (what is stored and what not, what parts/modules have adapted to the new settings and what not, etc).

And from security perspective handling the data better makes abuse a little harder.

gizmocuz:
I don't think when the application is crashing it will return true/false to the user....
But restarting python plugins could indeed be the issue... If it is still crashing as some user, maybe we could remark this code and do some testing.
I will merge the PR, thank you!

kiddigital:
Thx..

With this change, the end-user will get a message 'Problem saving settings!' (See SetupController.js line 743) because the back-end will now return status 'ERR' instead of crashing 😁.

(And there is some additional information in an MSG field that could contain a clue on where/why it crashed. At least a starting point 😉)

PR #5019, opened 2021-11-08
Decoded the latest version of the OT Bridge messages

brucemiranda:
The newer OTBs from Resideo (Honeywell) are called R8820A and seem to have newer firmware than the older R8810A. Unhelpfully both devices are labeled and sold as R8810A, however the firmware in them is different. This PR, makes the code work with both versions.


gizmocuz:
Thank you !!

PR #5022, opened 2021-11-12
Recovered .clang-format file

michahagg:
It went missing...

@gizmocuz When you want clang-format to leave your nicely formatted if's alone, you could use comments to turn it off and on.

```
if ( // clang-format off
    expr1
    || expr2
    || expr3
) // clang-format on
{
    statements
}
```

### DIFF
--- a/hardware/MQTT.cpp
+++ b/hardware/MQTT.cpp
@@ -2104,7 +2104,7 @@ void MQTT::GuessSensorTypeValue(const _tMQTTASensor* pSensor, uint8_t& devType, 
 		subType = sTypeKwh;
 
 		float fUsage = 0;
-		float fkWh = static_cast<float>(atof(pSensor->last_value.c_str()));
+		float fkWh = static_cast<float>(atof(pSensor->last_value.c_str())) * 1000.0F;
 		_tMQTTASensor* pWattSensor = get_auto_discovery_sensor_unit(pSensor, "W");
 		if (pWattSensor)
 		{


kiddigital:
@michahagg , I have issues to get the 'mocha' unit-tests running.

Could you see if and how they still work and maybe to update the README?

The unit-tests of dzVents and some first functional test for the Domoticz back-end are ran as part of the GitHub action pipelines. Would be nice if the UI unit-tests could be added as well 😁

gizmocuz:
@michahagg , I don't like to add comments for every if/then/else. Is there a setting to let clang don't touch them ?

michahagg:
I'm afraid there isn't any other way. Except trying to have that feature added to clang format itself. 

Not all if's need to be skipped for formatting, only the specific ones with long condition lists. And the comments are hardly noticeable with those long lists anyway.

Plus I'm not entirely sure it works in Visual Studio. I know it works in CLion, which I use. Could you give it a try? And see if it works for you. It beats having no formatter at all, right?

gizmocuz:
At the moment I write all my if/then/else in the above format. It is very easy to read, and more importantly, very easy to add a few lines
Having to write comments if of course annoying. The only thing clang may perform for if/then/else the the position of the opening bracket (after or under the if statement)

I would vote to remove the clang formatting at all if the above is not possible. I am afraid I don't have much experience with clang to know which option to turn off/on , but after getting so frustrated with it I removed the file and accidently committed it 
I have not missed it for 1 day.

The other annoying thing with the current clang is class definitions like:

```
class CMyClass
{
public:
  function bla bla
  function bla bla
  function bla bla
private:
  function bla bla
  member 1
  member 2
  member 3
}
```

As you can see above, the public/private statements are all the way to the left, perfect

With the current clang we have

```
class CMyClass
{
    public:
  function bla bla
  function bla bla
  function bla bla
    private:
  function bla bla
  member 1
  member 2
  member 3
}
```

which I really find frustrating as well....

As you can guess... I'm getting frustrated with clang and because of it, it does not get my vote to use it again

kiddigital:
@gizmocuz , @michahagg , I agree that having a format is better than having none.

But when I look at the formatted webserver source file I also don't like that as for me the readability got worse (in my opinion at least).

So maybe we need to give it some more thought and look at some options to find a workable format (and yes, there will always be personal preferences that are not met).

Maybe start with finding a [style ](https://clang.llvm.org/docs/ClangFormatStyleOptions.html) to start with.

A quick scan (need more time to look at the details) showed me that the 'Google C++' style suites my preferences.

Once we have a starting style, we can make little Domoticz style tweaks if needed so it fits our needs even better.

Suggestions welcome 😃

michahagg:
@gizmocuz @kiddigital Thanks for looking at the PR. I can only say I agree with both of you. It's very annoying if the formatter can't be tailored to 100% of what we want. Even with a style, there will always be an aspect that someone wants to divert from. Like with the if's having each condition on a separate line starting with the conditional operator, even if the line length doesn't exceed the configured maximum.

I have spent some time trying to find a better formatter, but it's quite difficult to find a 100% matching solution.

Maybe we should leave it at this for now and just format by hand.

PR #5034, opened 2021-11-28
TTNMQTT: Fix for empty frm_payload

kiddigital:
It cannot be assumed that frm_payload is always there.

When using non LLP structured data from sensor and custom encoders to create a JSON payload, the frm_payload will not be present, only the 'decoded_payload'.

This PR removes the assumptions and properly checks the payload existence.


gizmocuz:
Thank you!

jeroenkl:
thxs Kiddigital!

PR #5035, opened 2021-11-29
Improved http get communication with denkovi hardware

denkovi1:
Implemented a retry mechanism for http get requests for denkovi hardware, which should improve the communication.


gizmocuz:
Thank you!!

PR #5036, opened 2021-11-29
TTNMQTT: Some decoded payload values might be objects. Handle it better.

kiddigital:
Small fix to handle some (pre-)decoded payloads better.

Depending on the decoder, some decoded payload values are objects and not real values. These are now handled (ignored) and prevents the processing to be aborted.

Thanks to @jeroenkl for providing some real data examples.


jeroenkl:
Hoping this update is soon available

jeroenkl:
@gizmocuz Polite request, I really like to test the Kiddigital PR. Could you please put this into the beta?
much appreciated!

BR, Jeroen

gizmocuz:
Thank you for this PR!

PR #5037, opened 2021-11-30
Updates to Daikin hardware and some typos

sidepipe:
The main change is adding a poll interval to the Daikin code, which needed an additional field in settings. I took the decision to use a validation function to avoid duplicating code ( which I hate. ) Ideally this function needs to be modified further to allow negative values and nil max/min when they're not set, but this does for my purposes. IMHO other parts should use this validation function and indeed others created, maybe in separate source. Further development would then involve an overall validator which is passed an array of things to validate ( required type, type options, etc. ) This could be then further extended to create input fields themselves, and hardware can simply register requirements rather than having bits of specific hardware code in Hardwate.js.

This is all pie in the sky of course because I personally don't have the time, and probably nobody else does either :-(

### DIFF
--- a/hardware/Daikin.cpp
+++ b/hardware/Daikin.cpp
@@ -83,7 +83,7 @@ CDaikin::CDaikin(const int ID, const std::string &IPAddress, const unsigned shor
 	m_HwdID = ID;
 	m_usIPPort = usIPPort;
 	m_bOutputLog = false;
-	m_poll = poll?poll:300;
+	m_poll = poll ? poll : 300;
 	Init();
 }
 
--- a/hardware/Daikin.cpp
+++ b/hardware/Daikin.cpp
@@ -148,6 +148,7 @@ void CDaikin::Do_Work()
 			m_force_sci = false;
 			m_last_setcontrolinfo = 0;
 			last_sci_update = current_time;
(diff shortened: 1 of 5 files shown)


sidepipe:
I've now updated validateInteger() to check for minVal/maxVal being a number type, so negative numbers can now be used. Also, I realised that despite the fact that the Daikin code now does a poll after a set, it was written in such a way if the set wasn't successful, Domoticz controls still wouldn't be updated to reflect the real state - this has now ( hopefully ) also been fixed.

The motivation behind this is that my units often respond with an "Invalid response" ( which is "PARAM NG" ) that I assume means the request was malformed in some way. I still need to look into this when and if I have time - it could mean that the request is actually invalid in some way, or else I've read that these units aren't the most reliable when it comes to IP control so it could just be that ( and the fact that subsequent requests for the same thing succeed may mean that this is the case. ) Regardless, at least now, if you turn off a unit and it ignores the command for example, the switch in Domoticz will revert to on after a couple of seconds, so can be checked by human or script.

gizmocuz:
Thank you for this PR. I made two small changes, if you find them correct I will merge this PR

sidepipe:
Yep, fine by me.

gizmocuz:
Thank you!

PR #5042, opened 2021-12-05
Simplify hardware.js

sidepipe:
Hardware.js is, IMHO, completely unmanageable at this point. It's not surprising because of how much things have grown over time, but there comes a point where code really needs to be restructured. I'm not sure I'm the best one to be doing it, because I really don't know what I'm doing when it comes to Javascript/JQuery, so this may not be the best way to do it ( though it seems OK to me 😀  )

I've used the Daikin hardware to create my version of how I think it should be structured, and I've tried to do it in such a way that the other hardware can be migrated as and when someone gets time. It removes pretty much all of the code duplication that currently exists, as well as the indexOf shenanigans. As I mentioned though, I'm not a Javascript coder so someone who is might want to have a look! 

( BTW, there will be plenty of hardware I suspect that won't need to load extra HTML or Javascript... for now all these need to do is set the calledFetch flag, but if everything was migrated to this new structure even that wouldn't be necessary.  )


gizmocuz:
Please check your PR, for example undo ask the extern changes

kiddigital:
@sidepipe , I like your idea to restructure/clean-up the hardware settings UI part by optionally loading the sub-html for the specific hardware. 

But I think that also the Javascript part needs additional splitting up as well as it is likely (mandatory actually to make it work) that for each hardware specific HTML, also specific Javascript is needed to handle the hardware specific fields/settings to read/set them.

Second suggestion, split this PR into 2 parts (rebase first btw) .. one for the update to Daikin module (with mode1 = poll frequence set to a default when none given)..
and a second one with the proposed UI changes carving out the Daikin specific HTML and javascript into separate code pieces from the one big hardware.js/.html

makes it easier to test, updates, etc.

Nice work! :)

BTW.. how does this PR relate to #5037 ?

sidepipe:
@gizmocuz - Apologies, I think I've finally sorted out the mess... I'm useless at working with git ( I only use it for my own local projects which are pretty self-contained. ) I think what happened was I compiled in my development tree at some point and didn't notice that the external modules were included in a commit - I generally use commit -a and didn't notice. Shouldn't .gitignore prevent this from happening?

@kiddigital Firstly PR #5037 is actually the first 6 commits of this one, because locally I created a new branch and continued from there... I really didn't know whether to essentially start again from the development HEAD, or continue, and since I still wanted the Daikin changes I did the latter. As I mentioned I don't really work on any github projects really, just the odd thing here and there, so don't really know how this should be done. Not sure if this ( partly ) addresses your second point, in that the first PR is the changes to the Daikin module without the restructure, and the second is the restructure using the Daikin code as an example. Thing is, if the second PR didn't include the first, and the first is accepted, I'm not sure if the second would then need a rebase on the new development branch with the first PR included, if I'd have started again with this one, if you see what I mean?

As to your first point, I'm not really sure I follow what you're saying? The Daikin code works as presented in this PR, and does indeed use Javascript to validate the additional fields in the HTML as you suggest? This would be the normal mode, BUT there's also the possibility that a device just wants to include additional HTML and no Javascript, for maybe adding a note with no parameters for example. I'm not sure how you're suggesting it should be split further?

gizmocuz:
Thank you for rebasing! Is it possible to undo all translation changes (as this means every translator has to translate this again)

`ShowNotify($.t`

What KitDigital means is that in this PR consists of two changes.

1) Daiking with poll interval
2) Restururing some parts of the hardware code

A split could be usefull

1) a PR for only the daiking poll interval
2) a PR for restructuring the hardware code

sidepipe:
Ah, sorry I hadn't noticed that the bad English was generated by translation code! I've simply reverted the commit in question in my PRs - hope that's ok?

As to splitting the two PRs - my point there was that the second one depends on the first, since it uses the Daikin code as an example. The problem is that if the first PR is accepted, the second would need re-writing ( if it was based on development HEAD. ) Essentially the first PR ( which is only the Dailkin changes ) is a subset of the second - so if the Daikin changes are merged into your development branch, the first commits of my second PR disappear because they're already included.

If I HADN'T done it this way then if the first PR is merged, I'd have then had to re-write the second one to avoid merge conflicts, and since the first one isn't particularly controversial ( I hope ) I thought it better to assume it would be merged and save myself some work.

To put it another way:
First PR = only Daikin fixes
Second PR = Daikin fixes plus simplification of Hardware,js

Hope this makes some sense!

gizmocuz:
I merged your other PR and this has cause this PR to have some merge conflicts, is it possible to have a look at this ?

sidepipe:
Yep - I anticipated that ( it's because of your formatting changes. ) Already on it - will be sorted in a few minutes.

gizmocuz:
Thank you for the changes.

The variable "extraHWTable" seems to contain Daikin hardware settings. Should this variable not be named something like "extraDaikingHWTable" or similar ?

sidepipe:
No, the idea is that ALL hardware that's coded into Hardware.js should use this system - once that's done it would simplify Hardware.js massively and make it much easier to manage ( that's the theory anyway. ) extraHWTable would grow to match hardware names to the additional html/js that's loaded ( or empty if it just uses the standard fields ) so would contain the names of all hardware using this new structure.

gizmocuz:
Okey that's clear. But you are now matching it against a string. Could you add also the real hardware type/id in this table and use this ? The ID will always stay the same, but strings can be changed in the feature...
The ID's are also used in add/update/delete

They can be found in RFXNames.h (_eHardwareTypes)

sidepipe:
Yes, I didn't like this either - though that's how the whole of Hardware.js
works ( using the indexOf stuff ) so I just stuck with it because that's
how it currently was! Is the value property of the selector set to the
correct hardware ID ( i.e. can I match on that? )

gizmocuz:
No, the hardware ID changes, when you add a new hardware it gets a new ID.
Best to stick with the Hardware Type value/field. I know current implementation could be optimized a lot...

var addId = oTable.fnAddData({

here you can add the type if you which with

"TypeID": item.Type

Idealy we should use a enum like in RFXNames.h instead of using the id directly

But maybe this is for another PR ?

sidepipe:
Unfortunately I don't really have time to dig into things too much more right now - what I've done took far too long but it just irritated me how difficult it was to work out how to add a simple field to the hardware parameters, and how much duplication there is in Hardware.js. My feeling is this needs to be thought about and yes, probably for another PR - so IMHO maybe stick with what's there for now and try to come up with something better later. Again IMHO, even though as you say if the name changes the entry in this table has to change, that's still far better than the current situation with all the other hardware I think - if you happen to add more hardware which matches one of the indexOf tests, all hell will break loose!!

My thoughts on how it SHOULD be done are basically that you shouldn't duplicate data tables if you can possibly avoid it ( which using the name as now actually violates, but as I say it's short term. I think the names, IDs and enums need to be specified once in the code - that same code can generate a table somewhere under www that Hardware.js can then pull in to use to match an enum when determining the hardware type?

gizmocuz:
Yes that is correct. Domoticz started with about 3 different hardware types... ;)
A start has been made to split different hardware in different classes... Still 100 to go ;)

sidepipe:
Absolutely - and I would never criticise for such a thing because it happens, and I've been VERY guilty of exactly the same thing with my own code... you just keep adding and adding because it's easy, and sorting out the mess becomes harder and harder! My hope is that the other hardware can be moved to this new system a piece at a time, so not putting too much pressure on anyone's time, and then at least Hardware.js will be in a much better state I think.

gizmocuz:
Thank you for the changes/updates, working great!
Could you maybe add another example besides Daikin , for example MQTT and Dummy ? This way we have 3 good examples how to convert the rest...

sidepipe:
Ok, I've had a go at MQTT, but unfortunately I've run out of time for now, so it's not been tested well and currently the labels for the topic fields don't appear for some reason. If you could have a look and maybe test a bit in the meantime that would be appreciated ( at least it should give an idea of having more than on device on the new scheme! )

sidepipe:
Sorry, has a small amount of time again this morning and realised that there were two other hardware types that matched "MQTT" ( proving my point about it all being difficult to manage! ) They were so similar that I've used the same code, and stored a function parameter in the window context so that specific hardware code in extra/ can access it. I've used a function parameter in this way rather than perpetuate the indexOf() stuff.

The labels for the MQTT topics still don't appear ( haven't looked any more why. ) The issue there is that I really don't know HTML or Javascript/JQuery very well at all, so whereas it would probably take me hours to find out why the labels aren't there, it might take you seconds!

gizmocuz:
Thank you for the changes and looking into this. For me it seems that you know more what you are doing than me!

Could it me that you are missing a `</div>` behind `<div id="divmysensorsmqtt">` in Hardware.html ?

Edit: No that is not it, I was just checking it on 'changed files' in this PR and it looked like it

sidepipe:
It's likely to be something like that though... some of what I've done is just cut and paste from the existing code, so something could easily have been missed. I did have a bit of a look yesterday but didn't get very far - I didn't even know the "label" tag existed until I spotted they weren't appearing, so it could be something REALLY stupid :D

gizmocuz:
There is a error (closing tag of TH) on line 1065, I think someone made a copy/paste error, that last part should just be a `</th>`

sidepipe:
Ah - well spotted! Want me to fix it?

gizmocuz:
Yes please, I will do the same here, so we do not get a merge conflict I hope...
The thing is, I just split a hardware class into two different once (MQTT and MQTT Auto Discovery) and need to add the configuration in the web interface.
I am a bit worried that when I make changes and push them, that this could result in merge conflicts with your PR
We now have about 4 types of MQTT based hardware types, and all could have other parameters.
For instance, now, when we have more than 4 extra fields, it is assumed that it is the auto discovery topic (like you also did), but maybe later there is a new MQTT based hardware, and this parameter is for something else.
Maybe it is better to have separate files/classes for all hardware ?
Maybe if you do them one by one (or start with one), the labels will return ?
A label field is just a tag that makes the text have a different style (like more bold)

gizmocuz:
Otherwise if you are running out of time, please revert the MQTT changes, and I can merge your PR and do my changes and then later you might have a look at this if you have time, or I am going to try it ;)

sidepipe:
I think that the MQTT variants that are there are so similar it doesn't make sense to split them - you end up duplicating code again which really should be avoided. If different types are added in future it should be easy just to use the additional parameter - only if there are large differences should there be separate files ( though that is easy too, there's no reason why ALL variants of MQTT need to be in the same file. )
The labels thing is strange and wouldn't be helped by this since it happened when I only had the original MQTT variant in there... there's no reason why this shouldn't work.
As for the changes you need to do - if you are happy with the concept then maybe you could merge this to a new branch for now, and make your changes on top of this? In theory it should be easier for you to make your changes on top of this variant of the MQTT code because all the code is in one place and there's not multiple copies etc etc. I really don't have much time, and my wife shouts when I do this stuff at the weekend - but I will try my best to get what you need sorted!!!

gizmocuz:
Yes you are right about that as well, the basics are always the same, address/port/username/password/certificate etc.. Only my latest one has no input/output topics and a additional discovery topic, internally the extra parameters is the same (but input/output topics are empty for example).
Okey let me see what I can do, I'm not that great with git either, I will try to merge your changes here locally and going to debug the label issue

Thanks already for your time!!

sidepipe:
No problem... Domoticz is a great project and I'm just happy I can contribute something!

It should be fairly simple to do your new variant, just maybe add the new field in the MQTT HTML, and use maybe "3" as the parameter in the table at the top of Hardware.js and hide/show the fields as needed. If you need me to do anything then give me a shout - I can't guarantee I'll be able to respond straight away but will try my best!

gizmocuz:
Found the issue, it seems the translation does not work (different language).
If you look at the missing labels, these are `<span>` elements with a data-i18n, there is no default text set
Somehow your extra classes/html need to use the translation as the others do

gizmocuz:
Fixed in **function fetchExtraHTML** , `$div.append(phtml).i18n();`

sidepipe:
Ah, makes sense, thanks. I really have no idea about how the translation stuff works so that would've taken me ages to find/fix! 
If you're now working on your MQTT changes then shall I leave my version as is? That way we'll avoid conflicts.

As I say though, please let me know if I need to do anything.

gizmocuz:
Thanks! I merged your and mine changes, Now let's see if I can hack the new hardware config in ;)

gizmocuz:
Hope you have time for a question, I don't understand the validation ....
I committed some new code in the current development branch, and it should now be possible to configure the new MQTT Auto Discovery hardware (in hardware MQTT param = 3)
I would like to validate if a 'string' is not empty (the Discovery prefix)
The validation function in hardware.js is only called for the port, but I don't see where that is done
I expected the validation to be happening in MQTTParams.js at **extraHWValidateParams** ?

gizmocuz:
Got it, I need to extend the validation function, I added a 'String' type

`validators["String"](topdisc, "Auto Discovery Prefix")`

sidepipe:
The idea is:

- Anything which is "standard" and used by many devices ( username, password, ip, port ) is validated in Hardware.js.
- Anything which is pretty hardware specific BUT is a standard type, such as an integer between certain values that may or may not be empty is validated within the extra/*.js code BUT using a validator function supplied by Hardware.js in the validators object, so the actual validation function only has to be written once.
- Anything else is validated in extra/*.js using custom code in the validate function ( can't remember the name - out and about so not easy to check but it should be fairly obvious from the MQTT/Daikin examples. )

Simply checking for an empty string is so simple it could be done in the extra/*.js file validate function ( which just returns true or false. ) The alternative ( and preferable, but overkill for your example ) would be to add a validators["string"] function which takes, maybe, min/max length, valid character set, etc etc - it can then be used by any hardware that needs it.

Edit: Ha! I see you beat me to it!

gizmocuz:
Thanks for this! Like this implementation more and more ;) should have had this years ago ;)

Issue #5025, opened 2021-11-18
Year/Month compare graph does not display correct icon


gizmocuz:
I think the icons are in the wrong direction

![Screenshot_2021-11-18-06-49-05-83_df198e732186825c8df26e3c5a10d7cd](https://user-images.githubusercontent.com/2265299/142360824-356827ce-a4c1-4119-9c82-5cbe60126626.jpg)

See for example the bottom one... Should not be green as usage is more then previous month

waltervl:
I think it is meant to be the difference with the month before on the same year. So in your example that October 2021 is lower than september  2021 and that October 2020 was higher than september 2020.

Olifant1990:
@waltervl is correct. I made that same mistake myself too. Had to find out it is compared to previous month in the same year.
But indeed, that is not expected, I also want to see in the same month over the years if it increased or decreased

gizmocuz:
I think we should compare over the years here 👍

michahagg:
You're probably right. Do you want me to change it? Should be no problem.

GravityRZ:
installed 13.872 but it seems this change ha not gone through.

gizmocuz:
Would be great if you can change it

michahagg:
Hi @gizmocuz the PR #5047 contains the fix for this issue

GravityRZ:
any idea how i can get rid of this faulty value.
the graphs show it happened in 2020 but shift leftclick states it is an entry from 1970.
shift leftclick can not remove this item.

![image](https://user-images.githubusercontent.com/16853535/145600495-77baf0d2-2319-4186-b42d-46f01f6b5921.png)

michahagg:
Hi @GravityRZ it makes no sense deleting points from the compare chart as it shows aggregated data only. But nevertheless the data shown seems incorrect. Probably it's a counter loop that is not being recognized.

If you let me analyse the data, I might just find a way to fix it. Could you execute the following query on the database?

`select * from Meter_Calendar where DeviceRowID = [device-id] and Date >= '2020-01-01'`

Fill in the index of the device at [device-id].

GravityRZ:
here it is.
the list is rather long but here i see things which look strange
0 numbers or big negative number

87	283	332976	2020-10-25
87	228	333204	2020-10-26
87	216	333420	2020-10-27
87	0	333420	2020-10-28
87	0	333420	2020-10-29
87	282	364723	2020-10-31
87	269	364995	2020-11-01
[...]

Issue #5024, opened 2021-11-16
History button in log view


ghost:
I was faced with a question regarding the history buttons selection in log views.
Example:
![image](https://user-images.githubusercontent.com/12487713/141987200-a618c6da-6e61-4e44-a811-6c265386c656.png)

After discussed in slack, it appears that it's not a good idea to always show "3 days" log when the user set in config less that 3 days of history because in this case the button seems broken.
![image](https://user-images.githubusercontent.com/12487713/141987267-9d6296de-b033-4c6f-a8e2-255474a455f7.png)

But it would be interesting to put on the button the same number of days as the one set in config (or remove this button if there is just one day of history).

michahagg:
Hi @Reivax851 it should be fixed in PR #5047

PR #5047, opened 2021-12-10
Chart changes

michahagg:
This PR contains the following:
- fix for Computed kWh-devices where the power value was never updated and the energy was computed from a constant and incorrect power value
- fix for the Compare chart to calculate the trend using the value in the same period of the previous year instead of the value in the previous period of the same year (#5025 )
- fix for the Compare chart to ignore negative counter values
- fix for Short Log Chart to show zoom buttons for 1d and 3d only when relevant in according to the range of the chart (#5024)
- fix for Counter Incremental to calculate the usage of the day with the first value of the day instead of the lowest value
- removed the ability to delete points on the Compare Chart
- a fix to prevent text selection when panning a chart on a mobile device (still not completely fixed)
- update for Highcharts library
- removed each Javascript .map-file and the reference to it from the .js-file, plus gzipped all .js-files in /js


michahagg:
Hi @gizmocuz, I'm sorry but I introduced a nasty bug on 3 Oct this year. For computed kWh-devices the power value was updated with the current value instead of the new value, hence the power value was never updated at all. Each time a new update would occur, the energy value was calculated with the stuck power value. This PR contains a fix for this issue. My apologies for causing this issue.

The commit causing the error was part of PR #4958 which was merged on 7 Oct.

gizmocuz:
No worries, great you solved it !! Thank you again for your support!

gizmocuz:
Happy to merge this, but is it correct that the current month has no up/down icon ?

image

michahagg:
In your situation, this is correct indeed. The trend for a period is determined by comparing its value with the value of the same period in the previous year. In your situation, the previous year for 2021 is 2020. For this year no value is given, hence no comparison can be made. It is quite unlikely for this to occur. It means, no counter value was recorded in Dec 2020. Is that correct? Could you check that?

kiddigital:
> Hi @gizmocuz, I'm sorry but I introduced a nasty bug on 3 Oct this year. For computed kWh-devices the power value was updated with the current value instead of the new value, hence the power value was never updated at all. Each time a new update would occur, the energy value was calculated with the stuck power value. This PR contains a fix for this issue. My apologies for causing this issue.
> 
> The commit causing the error was part of PR #4958 which was merged on 7 Oct.

That is why we have Beta versions :) Thanks to Beta testing, the problem has not made it into the Stable.


> No worries, great you solved it !! Thank you again for your support!

🥇

gizmocuz:
Your correct! In this domoticz instance there was no data in december 2020, there was for the other months.
Tested the PR and works perfectly, thanks!

PR #5048, opened 2021-12-10
Scenes/Groups: Fix sending color_temp & RGB info for MQTT devices

jvanderzande:
Groups/Scenes are not sending the set color_temp and/or RGB color settings for the Group/Scene since the command is "Set Level". Added logic to mqtt to change "Set Level" to "Set Color" for ColorMode lights.


gizmocuz:
Thank you!!

Dylantje:
@jvanderzande 
Is there a way for me to test?

PR #5050, opened 2021-12-12
Support for homebrew python on Apple Silicon

moroen:
Adds support for python3 installed by homebrew for MacOs running on Apple Silicon. 

Also change the deprecated find_package(PythonInterp) and find_package(PythonLibs) with the current find_package(Python3).


gizmocuz:
It seems you are using a development version of Python. We can not expect others to do the same.
If you installed the normal python release , it should be detected.

moroen:
No, I'm not...

I'm using python 3.9.9 installed with homebrew on a M1 based MacBook. 

The file location for the frameworks is different for M1 based Macs and Intel based Macs, and thus the FindLibrary in hardware/plugins/DelayedLink.h has to be amended to find the .dylib on M1 based systems!

In case you are thinking of the use of "Development" in CMakeLitsts.txt, it's the syntax for the new FindPython3 module of cmake. It tells cmake that we're looking  for the development libraries (which always has been needed to compile and use python in domoticz), not that we're looking for a development version of python...

The CMakeLists in this PR will find any development libraries for python versions 3.4 and onwards... The current version will not be able to find any python3 versions installed on M1 Macs without some rather clever path manipulation or the use of manual options when configuring the build with  cmake...

gizmocuz:
@moroen , is it possible for you to test a compilation on a system like Debian ? Does this still work ?

moroen:
@gizmocuz 

It's tested on M1 Macs, Intel Macs and on EndeavourOS (Arch-derivative). I'll be happy to spin up a Debian VM to make sure it compiles and works on .deb based systems as well, and "report back"!

gizmocuz:
Thank you for the explanation. I am to unfamiliar with this. @dnpwwo , is it possible you can have a look if this is OK to merge ?

dnpwwo (approved):
The DelayedLink change looks fine to me.
Don't know much about make so can't comment

moroen:
So, I've tested this on a fresh Debian 11 install. Domoticz compiles and works as expected, using the instructions for compiling from source found in the wiki!

gizmocuz:
Thank you for this !

gizmocuz:
Thank you for the PR!!

PR #5058, opened 2021-12-15
Fix compile warning Debug message escape percent sign

kiddigital:
Small PR to fix (Linux)compile warning on Debug message. Percent sign was not escaped properly.


gizmocuz:
Thank you!!

PR #5061, opened 2021-12-15
Added support for Heatit ZM Single Relay 16A

hansrune:
Domoticz seems to have it's own fork of open-wave these days . In case you now maintain open-zwave separately, please add this PR.

This PR is already [integrated in open-zwave](https://github.com/OpenZWave/open-zwave/pull/2620):

_My first attempt on device support. Based on config file from https://products.z-wavealliance.org/products/4062. Tested to work in Domoticz, and passes the make xmltest and dist-update_

### DIFF
--- a/Config/manufacturer_specific.xml
+++ b/Config/manufacturer_specific.xml
@@ -1771,6 +1771,7 @@
     <Product config="there/800z.xml" id="0001" name="ThereGate" type="0001"/>
   </Manufacturer>
   <Manufacturer id="019b" name="ThermoFloor AS">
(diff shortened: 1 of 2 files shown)


gizmocuz:
Thank you for the configuration changes and this PR.

We always handled the device configuration files seperatly.
OZW is not being development anymore. Only device configuration changes are changed.
For this reason we are moving to MQTT with auto discovery and zwavejs2mqtt.
You can already try this in the latest beta versions.

hansrune:
Thanks. That was quick . Yes - I'm aware and interested in moving as well.

PR #5063, opened 2021-12-15
Fix jpeg inline attachment handling for K9 mail

mvdwater:
Incorrect Mime type causes K9 mail to show image as an unknown attachment, instead of showing the picture inline.


gizmocuz:
Thank you for this PR!!

PR #5065, opened 2021-12-15
Fix inconsistent format of svalues in MQTT (domoticz#5043)

Igitigit2:


gizmocuz:
Thank you for this PR.
I was not aware of a sValue1 sValue2 and so on, where/why is this used and not completly send to MQTT ? It makes the data transfer also higher.

Maybe rewrite the code to

```
int sIndex = 1;
Query = "svalue" + std::to_string(sIndex);
while (!root[Query].empty())
{
  newSvalue += root[Query].asString();
  if (sIndex > 1) newSvalue += ";";
  sIndex++;
  Query = "svalue" + std::to_string(sIndex);
}
```

Igitigit2:
@gizmocus, yes, that change should work the same. Saves one increment at the cost of symmetry, decide as you like it, I have no strong feelings about that.

gizmocuz:
If you can change it please. But back to my initial question,  was not aware of a sValue1 sValue2 and so on, where/why is this used and not completly send to MQTT ?

Igitigit2:
Don't ask me, where those values originate from. To me it looks as if they are contained in the mqtt topics sent by zwavejs2mqtt. Also, I remember that I read something with this notation before in Domoticz, but don't recall where, so I was not really surprised that it existed (could it be used in dzvents somwhere?). What i do not understand, is what you mean by "not completely send to MQTT".

Igitigit2:
Now, that I implemented the fixes for the Fibaro dimmer, I understand a lot more about MQTT/AD and I learned, that the problem occured  because I made mistake in the configuration. I overlooked the sentence "If enabled leave Prefex Topic in and Out empty." Nevertheless, in https://www.domoticz.com/wiki/MQTT in the _Domoticz to MQTT_ section you find two examples of messages with svalue[n] notation, so this may actually happen (at least from Domoticz to MQTT). I therefore propose to include this fix, although personally I probably won't need it anymore.

gizmocuz:
In the very near feature there will be a new hardware type and your current MQTT hardware will be automatically be converted if you used the auto discovery topic.
I am not aware of svalue1 as mentioned before...

PR #5068, opened 2021-12-16
MQTT Cover fix for payload 100 & off showing wrong slider position.

jvanderzande:
MQTT Cover fix for received packages payload 100 & off showing wrong slider position.


gizmocuz:
Thank you again !!

PR #5070, opened 2021-12-16
MQTT cover slider: Fix issue with wrong location of  previous fix

jvanderzande:
MQTT cover slider: Fix issue with wrong location of  previous fix which could cause issues with JSON payload.


gizmocuz:
Thank you!!

Issue #5072, opened 2021-12-17
MQTT AD issue when state message contain many different info


megamarco833:
these are the logs from TASMOTA device

```
19:48:01.425 MQT: stat/tasmota_esp32/RESULT = {"POWER2":"ON"}
19:48:01.428 MQT: stat/tasmota_esp32/POWER2 = ON
19:48:01.431 MQT: stat/tasmota_esp32/RESULT = {"ShutterPosition1":12}
19:48:01.473 MQT: stat/tasmota_esp32/RESULT = {"Shutter1":{"Position":13,"Direction":-1,"Target":12,"Tilt":-20}}
19:48:02.124 MQT: stat/tasmota_esp32/RESULT = {"Shutter1":{"Position":12,"Direction":-1,"Target":12,"Tilt":-45},"Shutter2":{"Position":51,"Direction":0,"Target":52,"Tilt":-24},"Shutter3":{"Position":56,"Direction":0,"Target":57,"Tilt":0},"Shutter4":{"Position":41,"Direction":0,"Target":41,"Tilt":0}}
19:48:02.375 MQT: tele/tasmota_esp32/STATE = {"Time":"2021-12-17T19:48:02","Uptime":"7T07:51:36","UptimeSec":633096,"Heap":122,"SleepMode":"Dynamic","Sleep":50,"LoadAvg":20,"MqttCount":29,"Berry":{"HeapUsed":4,"Objects":61},"POWER1":"OFF","POWER2":"OFF","POWER3":"OFF","POWER4":"OFF","POWER5":"OFF","POWER6":"OFF","POWER7":"OFF","POWER8":"OFF","Wifi":{"AP":1,"SSId":"NETGEAR","BSSId":"E0:3F:49:A8:C5:70","Channel":11,"Mode":"11n","RSSI":94,"Signal":-53,"LinkCount":1,"Downtime":"0T00:00:06"}}
19:48:02.388 MQT: stat/tasmota_esp32/RESULT = {"POWER2":"OFF"}
19:48:02.391 MQT: stat/tasmota_esp32/POWER2 = OFF
19:48:02.892 SHT: Tilt not match -45 -> -20
[...]

PR #5074, opened 2021-12-18
Fix for #5072 : Only show MQTT Unhandled state received error message when debugging

jvanderzande:
Fix for #5072  
Only show MQTT Unhandled state received error message when debug is true.
This to avoid many error messages in the log when unit like TASMOTE send its general STATE updates which currently triggers this error:
`Error: MQTT_HA: Unhandled state received '{"Time":"2021-10-20T11:26:09","Uptime":"12T00:48:32",.......`


gizmocuz:
Thank you!!

PR #5075, opened 2021-12-19
Chart fixes

### LINKED ISSUE #5023: Power usage graph scale not showing properly
Hi,

I am monitoring my electricity consumption with a CM180i and since one year roughly, after a big update, the graphic scale does not show properly.
The Y scale with the Wh is set from 0 to 400k and is fixed now. So with my little consumption at 6k Wh maximum, its not showing well at all. I have no view full scale of the "power usage" like the temperature (the precise curve)

I made a gif of my screen to show you the bug :
Animation

I am running the last beta. I was in stable but the last update made this bug appear I think and the beta channel doesn't change the bug.
I am running on firefox but the same bug happend under chrome or edge
Can you take a look ?
Thanks !!

### LINKED ISSUE #5025: Year/Month compare graph does not display correct icon

michahagg:
This PR contains:
- Fix for the day-chart of devices of type ENERGY and POWER. The sValue for these devices has a scale of 2, i.e. two decimals after the decimal point. In the Meter table this value is stored times 100 in column Value. When sending this value to the browser for the day-chart, it needs to be divided by a 100 in addition to the regular energy divider from the Settings. (Fixes #5023 )
- Fix for the tooltip in the compare-chart, where sometimes the tooltip container is not wide enough to contain the up and down trend-arrows. (Fixes #5025 )
- Fix the color of the trend-arrow in the tooltip in the compare-chart. Normally the up arrow is red and the down arrow is green. For devices where an increase is good and a decrease is bad (Energy Generated), the arrow is now reversely colored. 
- I added the font-awesome.css to the index.html to enable it for all pages. (It was already used in Setup.html.) This css declares the css-class 'sr-only' and hides any element with that class, as it means 'only used for screen readers'. In the ZWave.html page the class 'sr-only' was used on button 'Node management', which suddenly makes the button invisible. The class is now removed from the button, as there is no other visible element to show the Node-management dropdown menu.


gizmocuz:
Thank you again for your work !!

PR #5077, opened 2021-12-20
Update manufacturer_specific.xml

ab10002:
Add Abus Door Lock cfa3010

PR #5078, opened 2021-12-20
Create cfa3010.xml

ab10002:
Add Abus Door Lock cfa3010..


kiddigital:
@ab10002 , see the comment of @Olifant1990 in your other PR #5077 . It applies to adding this file as well.

Get it added to the original OpenZwave branch, then it will get sync'ed to Domoticz...

gizmocuz:
Yes please add this in the original openzwave repository, we will sync from time to time their config folder

PR #5081, opened 2021-12-24
Fix mqtt ad select

jvanderzande:
Fix MQTT Selector switch not working when STATE payload contains the actual STATE in stead of a JSON payload.
Added MQTT AD check for valid STATE received and leave Current STATE in case an invalid STATE is received.
Fix for fixing 1 decimal for Electricity "W" Sensor.


gizmocuz:
Thank you for the PR! Still I would have expected a proper incoming MQTT value

Issue #5069, opened 2021-12-16
Zwave/MQTT Fibaro RGBW dimmer not autodiscovered as RGBW


Igitigit2:
Domoticz V2021.1 (build 13872)

Autodiscovery does not work for the Fibaro RGBW dimmers (FGRGBW). MQTT::on_auto_discovery_message(...) is missing the code to detect this. Only keywords "color_mode", "supported_color_mode", "rgb" and "color_temp" are checked, but this dimmer sends different ones. Here is the autodiscovery message as received by MQTT.cpp:

homeassistant/light/EZ-RBGW1/rgb_dimmer/config, Message: 

{
	"state_topic": "zwave/EZ/12/38/0/currentValue",
	"command_topic": "zwave/EZ/12/38/0/targetValue/set",
	"on_command_type": "brightness",
	"state_value_template": "{{ \"OFF\" if value_json.value == 0 else \"ON\" }}",
	"brightness_value_template": "{{ value_json.value }}",
	"brightness_scale": 99,
	"rgb_command_template": "{{ {'red': red, 'green': green, 'blue': blue}|to_json }}",
	"rgb_value_template": "{{ value_json.value.red }},{{ value_json.value.green }},{{ value_json.value.blue }}",
	"rgb_state_topic": "zwave/EZ/12/51/0/currentColor",
	"rgb_command_topic": "zwave/EZ/12/51/0/targetColor/set",
	"brightness_state_topic": "zwave/EZ/12/38/0/currentValue",
	"brightness_command_topic": "zwave/EZ/12/38/0/targetValue/set",
	"device": {
		"identifiers": [
			"zwavejs2mqtt_0xf3981245_node12"
		],
		"manufacturer": "Fibargroup",
		"model": "RGBW Controller (FGRGBW)",
		"name": "EZ-RBGW1",
		"sw_version": "25.25"
	},
	"name": "EZ-RBGW1_rgb_dimmer",
	"unique_id": "zwavejs2mqtt_0xf3981245_12-51-0-currentColor"
}

gizmocuz:
Thank you for reporting! Could you also post the value of zwave/EZ/12/51/0/currentColor ?

Igitigit2:
Here we go:

zwave/EZ/12/51/0/currentColor
{
  "time": 1639580828153,
  "value": {
    "warmWhite": 80,
    "red": 235,
    "green": 131,
    "blue": 0
  },
  "nodeName": "RBGW1",
[...]

PR #5082, opened 2021-12-26
Fix AD/MQTT for Fibaro FGRGBW dimmer

Igitigit2:
This fixes issue #5069 (as far as I need it at least). I am not sure, if that everything required is fixed, but as far as I can test it, it looks ok. Some of this could be fixed on the zwavejs2mqtt side, but they seem unwilling to work on this (issues were closed without solutions). Summary of my findings:

- The color dimmer subdevice "rgb_dimmer" does not advertise itself as an RGBW device in the AD message.
   -> Add "color_mode": "true", "supported_color_modes" : ["rgbw"] to the AD message
- The value template for the rgb_command_topic is incomplete (warmWhite missing)
   -> hardcoded now
- The color values are named red/green/blue, not r/g/b. This is correctly indicated by the value template, but  since warmWhite is missing there, all values are now hardcoded. Decoding such a template is not implemented in Domoticz anyway.
- Color values are contained in a "value" json object , not inside "color", as Domoticz expects.
- To switch off/on, "value":"0"/"99" must be sent to the dimmer's state topic. The AD template suggest "ON"/"OFF" and Domoticz would usually set the "state" key (not "value").
   -> hardcoded 99/0 as payload_on/off and added "value" to the message sent. 
- Some of the topics to subscribe to were not yet implemented in Domoticz.

### DIFF
--- a/hardware/MQTT.cpp
+++ b/hardware/MQTT.cpp
@@ -197,18 +197,31 @@ void MQTT::on_message(const struct mosquitto_message *message)
 		if (qMessage.empty())
 			return;
 
+		if (topic.find("currentColor")!=topic.npos)
+			int found = 1;
+		if (topic == "zwave/EZ/12/51/0/currentColor/2")
+			int found = 1;
+
+
 		if (topic != m_TopicIn)
 		{
 			if (topic.find(m_TopicDiscoveryPrefix) == 0)
 			{
+				// Handle all messages sent to the AD topic 
 				on_auto_discovery_message(message);
 				return;
 			}
-			if (m_subscribed_topics.find(topic) != m_subscribed_topics.end())
+
+			if (m_subscribed_topics.find(topic) != m_subscribed_topics.end())		
 			{
+				// Handle all messages from sensors which were aklready autodiscovered.
+				if (topic.find("currentColor") != topic.npos)
+					int found = 1;
 				handle_auto_discovery_sensor_message(message);
 				return;
 			}
+
+			// Not a topic we are interested in!
 			return;
 		}
 	}
@@ -217,6 +230,11 @@ void MQTT::on_message(const struct mosquitto_message *message)
 		Log(LOG_ERROR, "Exception: %s!", e.what());
 		return;
 	}
+
+	// From here on topics only, which are subscribed by using the domoticz prefix, 
+	// nothing which was autodiscovered already, but there maybe new sensors which 
+	// will be discovered here, when they are detected for the first time!
+	// Check truth!
 
 	Json::Value root;
 	std::string szCommand = "udevice";
@@ -230,10 +248,15 @@ void MQTT::on_message(const struct mosquitto_message *message)
 		goto mqttinvaliddata;
 	try
 	{
-		if (!root["command"].empty())
+		if (root["command"].empty())
 		{
-			szCommand = root["command"].asString();
+			// For currentColor topic, set szCommand = "setcolbrightnessvalue" (required for e.g. Fibaro FGRGBW dimmer)
+			std::vector<std::string> topicArray;
+			StringSplit(topic, ";", topicArray);
+			if (topicArray.back() == "currentColor")
+				szCommand = "setcolbrightnessvalue";
 		}
+		else szCommand = root["command"].asString();
 
 		// Checks
 		if ((szCommand == "udevice") || (szCommand == "switchlight") || (szCommand == "getdeviceinfo"))
@@ -433,6 +456,7 @@ void MQTT::on_message(const struct mosquitto_message *message)
 						color = _tColor(r, g, b, cw, ww, ColorModeCustom);
 						break;
 				}
+
 				if (iswhite == "true")
 					color.mode = ColorModeWhite;
 				// Debug(DEBUG_NORM, "setcolbrightnessvalue: trgbww: %02x%02x%02x%02x%02x, color: '%s'", r, g, b, cw, ww, color.toString().c_str());
@@ -452,6 +476,33 @@ void MQTT::on_message(const struct mosquitto_message *message)
 				if (iswhite == "true")
 					color.mode = ColorModeWhite;
 				// Debug(DEBUG_NORM, "setcolbrightnessvalue2: hue: %f, rgb: %02x%02x%02x, color: '%s'", iHue, r, g, b, color.toString().c_str());
+			}
+
+			else if (!root["value"].empty())
+			{
+				// Some devices (e.g. Fibaro FGRGBW) use red, green, blue instead of r, g, b
+				// This could also be a "preprocessor" to if (!root["color"].empty()) above.
+				// The FGRGBW does provide a template which could be used to construct the proper format, 
+				// but it is rgb only and white is missing.
+				//if (!root["value"]["red"].empty())	root["value"]["r"] = root["value"]["red"];
+				//if (!root["value"]["green"].empty())	root["value"]["g"] = root["value"]["green"];
+				//if (!root["value"]["blue"].empty())	root["value"]["b"] = root["value"]["blue"];
+				//if (!root["value"]["warmWhite"].empty()) root["value"]["ww"] = root["value"]["warmWhite"];
+				//if (!root["value"]["coldWhite"].empty()) root["value"]["cw"] = root["value"]["coldWhite"];
+				color = _tColor(root["value"]);
+				if (color.mode == ColorModeRGB)
+				{
+					// Normalize RGB to full brightness
+					float hsb[3];
+					int r, g, b;
+					rgb2hsb(color.r, color.g, color.b, hsb);
+					hsb2rgb(hsb[0] * 360.0F, hsb[1], 1.0F, r, g, b, 255);
+					color.r = (uint8_t)r;
+					color.g = (uint8_t)g;
+					color.b = (uint8_t)b;
+					brightnessAdj = hsb[2];
+				}
+				// Debug(DEBUG_NORM, "setcolbrightnessvalue: color: '%s', bri: '%s'", color.toString().c_str(), brightness.c_str());
 			}
 
 			if (color.mode == ColorModeNone)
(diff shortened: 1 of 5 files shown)


hardware/MQTT.cpp
```diff
@@ -3076,19 +3134,11 @@ void MQTT::InsertUpdateSwitch(_tMQTTASensor* pSensor)
 						szOnOffValue = "Set Level";
 						// recalculate level to make relative to min/maxpositions
 						if (pSensor->component_type == "cover") {
-							if (
```
gizmocuz, line 3079: Could you undo the formatting here. It makes the code harder to read

hardware/MQTT.cpp
```diff
@@ -3098,12 +3148,12 @@ void MQTT::InsertUpdateSwitch(_tMQTTASensor* pSensor)
 		{
 			if (root["brightness"].empty() && root["position"].empty())
 			{
-#ifdef _DEBUG
-				_log.Debug(DEBUG_NORM, "ERROR: Last Payload is missing state field (%s/%s)", pSensor->unique_id.c_str(), szDeviceName.c_str());
```
gizmocuz: I remember that we did this on purpose as else some users would have continues error lines because some hardware sends wrong MQTT messages
You will see them when you compile/test in debug anyway
So maybe best to revert this change ?

hardware/MQTT.cpp
```diff
@@ -3126,32 +3176,20 @@ void MQTT::InsertUpdateSwitch(_tMQTTASensor* pSensor)
 			{
 				szOnOffValue = "Set Level";
 				// Make level relative to 100.
-				if (
```
gizmocuz, line 3129: Could you undo the formatting here. It makes the code harder to read

hardware/MQTT.cpp
```diff
@@ -3126,32 +3176,20 @@ void MQTT::InsertUpdateSwitch(_tMQTTASensor* pSensor)
...
 					level = (int)(100 - ((100.0 / (pSensor->position_open - pSensor->position_closed)) * level));
-				}
 			}
 		}
 		if (!root["color"].empty())
 		{
 			Json::Value root_color;
 			bool ret = ParseJSon(sColor, root_color);
 			if (ret)
-			{
```
gizmocuz, line 3149: Could you undo the formatting here. It makes the code harder to read

hardware/MQTT.cpp
```diff
@@ -3160,14 +3198,29 @@ void MQTT::InsertUpdateSwitch(_tMQTTASensor* pSensor)
 			if (!root["color"]["b"].empty())
 				color_new.b = root["color"]["b"].asInt();
 			if (!root["color"]["c"].empty())
-				color_new.cw = root["color"]["c"].asInt();
+				color_new.cw = root["color"]["c"].asInt();
```
gizmocuz: Could you undo this change, or remove the spaces at the end?

hardware/MQTT.cpp
```diff
@@ -3593,14 +3666,8 @@ bool MQTT::SendSwitchCommand(const std::string &DeviceID, const std::string &Dev
 				if (!result.empty())
 				{
 					_eSwitchType sSwitchType = (_eSwitchType)atoi(result[0][0].c_str());
-					if (
```
gizmocuz, line 3596: Could you undo the formatting here. It makes the code harder to read

gizmocuz (changes requested):
Thank you very much for the changes!
Just some small changes in request

hardware/MQTT.cpp
Igitigit2: I don't know, why this was changed, I did not do it knowingly at least. Anyway, I will change it and also insert a Cr before the opening brace on the line before to bring it in line with he rest of the coding style.

hardware/MQTT.cpp
Igitigit2: Same here, I definitly did not touch this line. Is it possible that it was merged somehow when I checked it in? I am never really sure of what I do when using Github... Anyway, I will revert this.

hardware/MQTT.cpp
Igitigit2: Done

hardware/MQTT.cpp
Igitigit2: Same again...

hardware/MQTT.cpp
Igitigit2: Same again, reverted.

hardware/MQTT.cpp
Igitigit2: I do not quite agree to the readability here, but will do so.

Igitigit2:
Ok, I commited all requested changes, but still like to know where some ogf the edits came from. I don't think, it was the IDE auto formating (all off now anyway), since most of the edits that needed to be reverted were related to blinds and covers, code sections which I did not touch, although blame says differently. Any ideas? I would like to avoid this in the future.

gizmocuz:
First of all, thank you for your PR and time!

This PR consists of more changes than specified in the topic. For example extra empty lines, comments, changes in updatedevice (not mqtt related)
This makes it a bit hard to check

And (my fault?!) I just have jsut split the MQTT class into two hardware classes (default MQTT and MQTT Auto Discovery), this is causing a merge conflict.

I am not sure why this specific node(type) is being used differently by zwavejs2mqtt then the other RGB/W/CW nodes...
You have a explicit check on a string in this PR, maybe this should not be done as maybe in the feature this could work for other ? (If you think that without this it could break existing implementations by all means keep the string check)

You are introducing two new topics and templates, could you not use the already existing command/state topics/templates already in the system ?

Could you also post the following information from the node (MQTT)
- config object
- state object

Do you have a reference to the zigbee2mqtt issue that was closed ?

Igitigit2:
If somebody has to say thanX for the time, then it is probably us Domoticz users to you!

> This PR consists of more changes than specified in the topic. For example extra empty lines, comments, changes in updatedevice (not mqtt related)

Hm, I agree, there may be few added empty lines and perhaps three comments, but I don't think, I changed something outside of the MQTT class except for the color components in _tColor::fromJSON(), so it should all be mqtt related... The color components could of course also be patched inside the MQTT class, but that would be pretty bad style in my opinion. What makes the changes look a lot more than they are are probably the changes to the sections which I am not aware of having touched them at all (things related to covers and blinds, plus the one log message). I have really no clue where they came from, especially with the log message, which is not justdifferent formating, I am totaly lost. It must be related to git/github somehow and my improper use of those.

> And (my fault?!) I just have jsut split the MQTT class into two hardware classes (default MQTT and MQTT Auto Discovery), this is causing a merge conflict.

That of course makes it much harder to merge. If there is anyway for me to help, please let me know.

> I am not sure why this specific node(type) is being used differently by zwavejs2mqtt then the other RGB/W/CW nodes... You have a explicit check on a string in this PR, maybe this should not be done as maybe in the feature this could work for other ? (If you think that without this it could break existing implementations by all means keep the string check)

I do not at all like the check for the specific node type either, but I do not see any other option. The reason it is required is because zwavejs2mqtt is sending an incomplete config message which needs to be patched before it can be correctly processed by the mqtt class. This is a problem specific to the node type and not a generic thing, so I believe it is acceptible to introduce the specific node type check and patch here. BTW, homeassistant seems to have a similar problem with that node type. I would leave the check as I implemented it for now.

> You are introducing two new topics and templates, could you not use the already existing command/state topics/templates already in the system ?

I added the topics and templates which are sent in the config message. At least the rgb_value_template could probably be omitted, since it does not work anyway (it is incomplete in the config message), but the others should be required. I cannot simply use the state topic instead of the rgb state topic., since both are used. Also, I don't like it if variables are named purpose_x and used for purpose_y. Same holds for command_topic and rgb_command_topic. The dimmer exhibits all four topics and they are all in use (as far as I remember...).

> Could you also post the following information from the node (MQTT)
> 
> * config object
> * state object

Here we go (copied from older notes, I cannot simply switch to the dev system without annoying my wife too much ;-) ). Note that rgb_value_template is incomplete.

homeassistant/light/EZ-RBGW1/rgb_dimmer/config
{
	"state_topic": "zwave/EZ/12/38/0/currentValue",
	"command_topic": "zwave/EZ/12/38/0/targetValue/set",
	"brightness_scale": 99,
	
	"rgb_state_topic": "zwave/EZ/12/51/0/currentColor",
	"rgb_command_topic": "zwave/EZ/12/51/0/targetColor/set",
	"brightness_state_topic": "zwave/EZ/12/38/0/currentValue",
	"brightness_command_topic": "zwave/EZ/12/38/0/targetValue/set",
	"device": {
		"identifiers": [
			"zwavejs2mqtt_0xf3981245_node12"
		],
		"manufacturer": "Fibargroup",
		"model": "RGBW Controller (FGRGBW)",
		"name": "EZ-RBGW1",
		"sw_version": "25.25"
	},
	"name": "EZ-RBGW1_rgb_dimmer",
	"unique_id": "zwavejs2mqtt_0xf3981245_12-51-0-currentColor"
}


zwave/EZ/12/51/0/currentColor
{
	"time": 1640189400483,
	"value": {
		"warmWhite": "0",
		"red": "0",
		"green": "0",
		"blue": "99"
	},
	"nodeName": "RBGW1",
	"nodeLocation": "EZ"
}

> Do you have a reference to the zigbee2mqtt issue that was closed ?

??? I do not use zigbee2mqtt or are you referring to my other PR which was closed today and also related to zwavejs2mqtt? If yes, there is no dependency of these from each other and the closed one is probably really not required at all.

gizmocuz:
Sorry, I meant zwavejs2mqtt ;) you mentioned others had issues with this devices, requested help, but the issue was closed.

About re-using possible state/value topics, If the command_topic is not in use, but a rgb_command_topic is in use, I see no harm using the command_topic for this and not introduce 100+ new topics

Igitigit2:
I checked again in mqtt explorer and both topics are really used (command_topic and rgb_command_topic).

The closed issues where in the zwavejs2mqtt forum, but I do not have the links at hand anymore. I did a lot of wild searching initially when I came across this. I also read that the authors of zwavejs2mqtt want to stop working on autodiscovery, but I would not overweight this.

If it helps, then I can try to pull my changes into the current top revision and start a new PR based on that. Shouldn't be too much work, just testing and  potentially debugging will take some time again. Would be a chance for getting rid of my undesired changes as well... I even discovered a whole section related to covers and blinds which suddenly popped up in my code and which I definitely never wrote. I must have merged something without even noticing.

Igitigit2:
I merged my changes into the current head of the development branch and will test them tomorow. If everything is ok, how shall we proceed?

gizmocuz:
If you could make a new PR with the only changes required for this to work then please.

Igitigit2:
Changes here were based on the code before refactoring the MQTT class. New code changes after the refactoring are proposed in PR #5100. Discussion to be continued there.

PR #5084, opened 2021-12-28
Webserver Improvement: Serve files without extension (Certbot temp files for example)

kiddigital:
This PR adds (or brings back) the functionality to the Domoticz webserver to serve regular files that _do not_ have a file extension that could be used to determine the correct mime-type which the webserver should return when serving this file.

Before this PR, a file in the webroot of the fileserver that did not contain a '.' (dot) was not served but a '400 Bad request' was returned instead. Now it does.

The updated behavior is as followed:
- Non-existing files with a dot (having an extension) will give a 404
- Non-existing files without a dot will still give a 400 (and not a 404 File not found)
- Both existing- and non-existing files ending with a dot will give a 400
- Existing files with a dot will be served (200)
- Existing files _without_ a dot will now also be served (200)

Change is made to accommodate the use of Letsencrypt SSL certificates to encrypt Domoticz web-traffic for which it is needed to serve temporary verification files generated by `certbot` when creating and updating the SSL certificate.

See also forum topic https://www.domoticz.com/forum/viewtopic.php?f=47&t=37600

### DIFF
--- a/test/gherkin/resources/testwebcontent/test2
+++ b/test/gherkin/resources/testwebcontent/test2
@@ -1,0 +1,1 @@
+A regular file without extension!
--- a/test/gherkin/test_webserver.py
+++ b/test/gherkin/test_webserver.py
@@ -20,6 +20,14 @@ def test_uncompressedout():
 def test_decompressedout():
     pass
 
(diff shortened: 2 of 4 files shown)


gizmocuz:
Thank you for this PR!!

PR #5085, opened 2021-12-29
Prevent loosing wind devices

hvbommel:
TFA wind device keeps changing its device ID (composed using channel/rolling-code) even without battery change or power loss. As a result the device is lost in Domoticz. Forcing the device ID to 0 prevents loosing the device for both SDR433 and RFXcom wind devices.

### DIFF
--- a/main/mainworker.cpp
+++ b/main/mainworker.cpp
@@ -3217,7 +3217,6 @@ void MainWorker::decode_Wind(const CDomoticzHardwareBase* pHardware, const tRBUF
 	char szTmp[300];
 	uint8_t devType = pTypeWIND;
 	uint8_t subType = pResponse->WIND.subtype;
-	// uint16_t windID = (pResponse->WIND.id1 * 256) + pResponse->WIND.id2;
 	uint16_t windID = 0;
 	sprintf(szTmp, "%d", windID);
 	std::string ID = szTmp;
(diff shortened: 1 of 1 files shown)


Olifant1990:
I think I do not want this PR to be merged.
There can be (theoretically) more then one wind meter in a domoticz system and then you will merge both of them to one device.
Maybe the fix would be to look for a way to find out which device it is/was?

hvbommel:
Multiple wind devices are still supported and working ok. I have currently 3 active (1 TFA on SDR433, 1 TFA on RFXcom and 1 buienradar) and working ok all having a device ID of 0000 but all having a different index of course. For completeness: The failing windsensor is the "TFA Dostman 30.3149 Wind Transmitter".  See https://www.tfa-dostmann.de/en/product/wind-transmitter-30-3149/ for technical data and a picture.

Olifant1990:
Well, you have 3 different wind devices on 3 different types of "hardware".
But I forsee problems when you have multiple wind devices on the same hardware.
Because, which device with ID 0000 is which....
Maybe make an exception for problematic devices, but please, do not set device ID to 0 for every single wind device.

hvbommel:
Can you propose another fix? I don't think we have a way to the detect if the device is the TFA Dostmann 30.3149 and set device-id to zero, just for this device. Also I do not think many have multiple wind devices on a single domoticz system. TFA Dostmann is a very common manufacturer. Not supporting it would not be my preference.

gizmocuz:
We do have multiple wind devices on one systeem, same for temperatures. It's a known issue of these devices, today's why they're is a replace option when you press edit on the device

gizmocuz:
This PR cannot be merged, but thank you for your effort

PR #5086, opened 2021-12-30
Fix allowed TCP port ranges

keryell:
TCP ports are between 1 and 65535.
Modernize the code with C++11.

### DIFF
--- a/main/domoticz.cpp
+++ b/main/domoticz.cpp
@@ -862,6 +862,11 @@ int main(int argc, char**argv)
                  " or 0 to disable");
 				return 1;
 			}
+			else if (iPort >= 49152) {
+				_log.Log(LOG_NORM,
+                 "You are using a dynamic TCP port number (>= 49152)."
+                 " This might cause network conflicts.");
+			}
 			secure_webserver_settings.listening_port = wwwport;
 		}
 		if (!webserver_settings.listening_address.empty()) {
--- a/main/domoticz.cpp
+++ b/main/domoticz.cpp
@@ -855,10 +855,11 @@ int main(int argc, char**argv)
 			}
 			std::string wwwport = cmdLine.GetSafeArgument("-sslwww", 0, "");
 			auto iPort = std::stoi(wwwport);
-			if ((iPort < 1) || (iPort > 65535))
+			if ((iPort < 0) || (iPort > 65535))
 			{
 				_log.Log(LOG_ERROR,
-                 "Please specify a valid sslwww port between 1 and 65535");
+                 "Please specify a valid sslwww port between 1 and 65535"
+                 " or 0 to disable");
 				return 1;
 			}
 			secure_webserver_settings.listening_port = wwwport;
--- a/main/domoticz.cpp
+++ b/main/domoticz.cpp
@@ -42,7 +42,7 @@
 	#include <syslog.h>
 	#include <errno.h>
 	#include <fcntl.h>
-	#include <string.h> 
+	#include <string.h>
 #endif
 
 namespace
@@ -854,10 +854,11 @@ int main(int argc, char**argv)
 				return 1;
 			}
 			std::string wwwport = cmdLine.GetSafeArgument("-sslwww", 0, "");
-			int iPort = (int)atoi(wwwport.c_str());
-			if ((iPort < 0) || (iPort > 32767))
+			auto iPort = std::stoi(wwwport);
+			if ((iPort < 1) || (iPort > 65535))
 			{
-				_log.Log(LOG_ERROR, "Please specify a valid sslwww port");
+				_log.Log(LOG_ERROR,
+                 "Please specify a valid sslwww port between 1 and 65535");
 				return 1;
 			}
 			secure_webserver_settings.listening_port = wwwport;
--- a/main/domoticz.cpp
+++ b/main/domoticz.cpp
@@ -1,7 +1,7 @@
 /*
  Domoticz, Open Source Home Automation System
 
- Copyright (C) 2012,2020 Rob Peters (GizMoCuz)
+ Copyright (C) 2012,2022 Rob Peters (GizMoCuz)
 
  Domoticz is free software: you can redistribute it and/or modify it
  under the terms of the GNU General Public License as published
@@ -23,6 +23,7 @@
 #include <sys/types.h>
 #include <signal.h>
 #include <iostream>
+#include <string>
 #include "CmdLine.h"
 #include "Logger.h"
 #include "Helper.h"
@@ -587,7 +588,7 @@ bool ParseConfigFile(const std::string &szConfigFile)
 		}
 
 		else if (szFlag == "startup_delay") {
-			int DelaySeconds = atoi(sLine.c_str());
+			int DelaySeconds = std::stoi(sLine);
 			_log.Log(LOG_STATUS, "Startup delay... waiting %d seconds...", DelaySeconds);
 			sleep_seconds(DelaySeconds);
 		}
@@ -792,7 +793,7 @@ int main(int argc, char**argv)
 				_log.Log(LOG_ERROR, "Please specify a startupdelay");
 				return 1;
 			}
-			int DelaySeconds = atoi(cmdLine.GetSafeArgument("-startupdelay", 0, "").c_str());
+			int DelaySeconds = std::stoi(cmdLine.GetSafeArgument("-startupdelay", 0, ""));
 			_log.Log(LOG_STATUS, "Startup delay... waiting %d seconds...", DelaySeconds);
 			sleep_seconds(DelaySeconds);
 		}
@@ -813,7 +814,7 @@ int main(int argc, char**argv)
 				return 1;
 			}
 			std::string wwwport = cmdLine.GetSafeArgument("-www", 0, "");
-			int iPort = (int)atoi(wwwport.c_str());
+			int iPort = std::stoi(wwwport);
 			if ((iPort < 0) || (iPort > 32767))
 			{
 				_log.Log(LOG_ERROR, "Please specify a valid www port");
@@ -1161,7 +1162,7 @@ int main(int argc, char**argv)
 #ifndef _DEBUG
 	RedirectIOToConsole();	//hide console
 #endif
-	InitWindowsHelper(hInstance, hPrevInstance, nShowCmd, m_mainworker.GetWebserverAddress(), atoi(m_mainworker.GetWebserverPort().c_str()), bStartWebBrowser);
+	InitWindowsHelper(hInstance, hPrevInstance, nShowCmd, m_mainworker.GetWebserverAddress(), std::stoi(m_mainworker.GetWebserverPort()), bStartWebBrowser);
 	MSG Msg;
 	while (!g_bStopApplication)
 	{
--- a/main/domoticz.cpp
+++ b/main/domoticz.cpp
@@ -854,7 +854,7 @@ int main(int argc, char**argv)
 				return 1;
 			}
 			std::string wwwport = cmdLine.GetSafeArgument("-sslwww", 0, "");
-			auto iPort = std::stoi(wwwport);
+			int iPort = std::stoi(wwwport);
 			if ((iPort < 0) || (iPort > 65535))
 			{
 				_log.Log(LOG_ERROR,
--- a/main/domoticz.cpp
+++ b/main/domoticz.cpp
@@ -815,9 +815,11 @@ int main(int argc, char**argv)
 			}
 			std::string wwwport = cmdLine.GetSafeArgument("-www", 0, "");
 			int iPort = std::stoi(wwwport);
-			if ((iPort < 0) || (iPort > 32767))
+			if ((iPort < 0) || (iPort > 49151))
 			{
-				_log.Log(LOG_ERROR, "Please specify a valid www port");
+				_log.Log(LOG_ERROR,
+                 "Please specify a valid www port between 1 and 49151"
+                 " or 0 to disable");
 				return 1;
 			}
 			webserver_settings.listening_port = wwwport;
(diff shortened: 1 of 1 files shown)


kiddigital:
@keryell , nice thx.

It would even be better if the selected port for the SSL/TLS encrypted HTTP traffic would conform to internet netiquette as described in RFC's. 

[RFC 4340 section 19.9](https://datatracker.ietf.org/doc/html/RFC4340#section-19.9) , mentioned 3 groups of port ranges:

> Port numbers are divided into three ranges.  The Well Known Ports
    are those from 0 through 1023, the Registered Ports are those from
    1024 through 49151, and the Dynamic and/or Private Ports are those
    from 49152 through 65535.  Well Known and Registered Ports are
    intended for use by server applications that desire a default
    contact point on a system.  On most systems, Well Known Ports can
    only be used by system (or root) processes or by programs executed
    by privileged users, while Registered Ports can be used by ordinary
    user processes or programs executed by ordinary users.  Dynamic
    and/or Private Ports are intended for temporary use, including
    client-side ports, out-of-band negotiated ports, and application
    testing prior to registration of a dedicated port; they MUST NOT be
    registered.

So
0-1023 : Well Known ports -> here port 443 is the one port dedicated to the type of traffic we are talking about here
1024-49151: Registered ports
49152-65535: Dynamic ports

As the intended port specified here is the port number that the Domoticz webserver should use to handle SSL/TLS encryped HTTP traffic, it is safe to assume that the user wants to secure (by encrypting) traffic between the Domoticz webserver and clients because the webserver is exposed in an environment that is not considered fully safe. Most likely not a small local private network :). Therefor if it is not a small local private- and/or fully safe network , it is better to behave nicely and stick to proper netiquette.

Maybe an improved check could be something like:

```
if (!((iPort == 0) || (iPort == 443) || ((iPort >= 1024) && (iPort <= 49151)))
```

Makes sense?

A list of Named and Reserved ports can be found in [RFC 1700](https://datatracker.ietf.org/doc/html/rfc1700).

But RFC1700 also refers to the actual list of reserved/registered ports which can be found at the IANA website:
https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml

keryell:
There are various aspects in the discussion here and there is a difference between what is allowed and the best practices.
Since it is about a manual configuration, why not allowing whatever allowed value? In my case I hit the problem because I wanted to do some port forwarding at the entrance of my Internet router and it required a port >= 49152.
For example I am using an URL *à la* `https://...:55555`, keeping the same port number outside and inside to simplify the numbering.

keryell:
By the way, I have the feeling the old STD-1700 has been superseded by https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml

kiddigital:
> By the way, I have the feeling the old STD-1700 has been superseded by https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml

Yes, see the bottom of my previous comment 😁

keryell:
> Yes, see the bottom of my previous comment

I clearly need a new brain for 2022... :-(

kiddigital:
@keryell , I see what you want to do to keep things simple (and easy to remember). Probably you choose 55555 as it is easy as well and without the notion that you were picking a _dynamic_ port.

Also the chance is small that it will cause issues, but for example an online game from your game-console (if you would have one) could randomly pick the same port as well for a game session and that might lead to unexpected behavior.

So how about
```
if (!((iPort == 0) || (iPort == 443) || ((iPort >= 1024) && (iPort <= 65535)))
```
and add an additional warning
```
else if (iPort >= 49152)
    Log(LOG_WARN,"You are using a Dynamic port number (>49151)! Might cause network conflicts.");
```
Does that work for you? If so, could you update the PR?

keryell:
Thanks for the feedback.
But it is not clear to me why you are specially worried about the case 49152 and above but not the other ones.
For example you want to disable a warning for the port 443 but obviously if you have a plain web server in the machine with plain TLS support using plain TCP 443 port but Domoticz on the same machine also uses the same port but starts before, it will break the plain web server.
If you pick some ports starting at 6000 and you have X11 with remote access for display :0 you will break X11 on the same machine and so on.
Here we are talking about a network config option for Domoticz on a machine assuming and *allowing* some level of network planning on the same machine.
I do not understand what you mean by:
> but for example an online game from your game-console (if you would have one) could randomly pick the same port as well for a game session and that might lead to unexpected behavior.

What is the relation between a game console and my Raspberry Pi running Domoticz?
Do you mean using the Raspberry Pi running Domoticz also  as a game console with some emulator?
Do you mean installing a game server on the Raspberry Pi interacting with a game console and interacting badly on the TCP ports of the Raspberry Pi?
Could you please clarify your concern?

kiddigital:
> Thanks for the feedback. .... 
> I do not understand what you mean by:
> 
> > but for example an online game from your game-console (if you would have one) could randomly pick the same port as well for a game session and that might lead to unexpected behavior.
> 
> What is the relation between a game console and my Raspberry Pi running Domoticz? Do you mean using the Raspberry Pi running Domoticz also as a game console with some emulator? Do you mean installing a game server on the Raspberry Pi interacting with a game console and interacting badly on the TCP ports of the Raspberry Pi? Could you please clarify your concern?

It is _not_ the machine running Domoticz I am worried about. The concern I have is that the main reason to use _non-standard_ ports, here 443 being the standard for (domoticz) HTTP traffic, is to _open-up_ the Domoticz webserver to the outside world. Most of the time this means that the _internal_ Domoticz (SSL protected) webserver has to be reachable _through_ the network _gateway_. This gateway exposes a port on the outside and binds it to an internal port.

The reason I mention games/game-consoles, they are a good example of _applications/devices_ that make often use of _dynamic ports_ by using technologies like [STUN](https://en.wikipedia.org/wiki/STUN) or [IGD](https://en.wikipedia.org/wiki/Internet_Gateway_Device_Protocol) to ask the earlier mentioned network gateway (your router for example) to dynamically open a _random_ port (at least for a while, like a gaming sessions, etc.).

These _random_ ports should be ports in the _dynamic port_ range (> 49151). If you run a more permanent service on one of these ports (i.e. Domoticz on 55555), it could happen that your router (re)uses that port number when requested for a dynamic port. It really depends on how well the network gateway/router works and if it is smart enough to remember that 55555 in this case, is already used.

Therefor, at least giving a warning message make people aware that they might do something that potentially causes an issue in some cases. Without such a message, most people are not even aware. And often, just being aware is sufficient to choose a different port number.

Does that clarify? (and btw. it is never fool-proof as some routers/gateways also open up _non dynamic_ ports in a response to IGD/STUN/etc. requests..)

gizmocuz:
Thank you for this PR! If it possible to revert the 'auto' to 'int' again ?
Any router should be able to translate the outside port (49152) to a inside port on a machine these days I hope (49152 -> 192.168.0.10:8080)
I do agree that using high port numbers could possible break connections as these are normally used in communication
That is why I limited to 32K

keryell:
Thank you for the answer.
Yes I can update the PR but I want to be sure we do not over-engineer the problem here.
There is no `LOG_WARN` in `main/Logger.h` but that seems like a good idea. Using `

keryell:
Wrong button... Sorry.

keryell:
Thank you for the feedback.
Yes I can update the PR but I want to be sure we do not over-engineer the problem here, because we can probably contribute to Domoticz in more fundamental areas with our precious time. :-)
@kiddigital There is no `LOG_WARN` in `main/Logger.h` but that seems like a good idea to have one. Using `LOG_NORM` for now. I have not applied your restrictions on the ports yet because it would disallow some ports which have been allowed up to now.
It would make sense to generalize this to all the other port selection options like `-www` and so could be factorized out in a single function. Do you think we should go along this path?
@gizmocuz Yes, it is clearer to keep `int` here, and even shorter. :-) While I am here I have changed the other `atoi` in this file.

main/domoticz.cpp
```diff
@@ -854,7 +854,7 @@ int main(int argc, char**argv)
 				return 1;
 			}
 			std::string wwwport = cmdLine.GetSafeArgument("-sslwww", 0, "");
-			auto iPort = std::stoi(wwwport);
+			int iPort = std::stoi(wwwport);
```
kiddigital, line 860: @keryell , @gizmocuz , I am ok with using `int` here, but (as a not C++ programmer) it got me thinking about the use/usefulness of `auto`.

I can see why in a case like this using `auto` makes sense. A new variable gets initialized and is directly populated by a _conversion_ function. This conversion function is _casting_ in a sense from one type to another. The _conversion_ function (`std::stoi`) has an intended destination type so it makes sense to initialize the new variable with the same type. That is what `auto` is intended for and the compiler is very well aware if the output type of this _conversion_ function. We humans are maybe not, at least not directly. That is why coding it explicit as `int` makes it clear to us.

**But** it can also be read as another _implicit_ conversion if the _conversion_ function would be of a different type than `int`.

It can be argued that, although in the (_C_) past everything should be very explicit, in modern C++ it is fine to be _less explicit_ when it comes to certain declaration, etc. as the compiler knows how to cope with this and coding can be done at a higher abstraction similar to languages like Python or Go.

Just my 2 cents..

kiddigital (approved):

main/domoticz.cpp
```diff
@@ -813,7 +814,7 @@ int main(int argc, char**argv)
 				return 1;
 			}
 			std::string wwwport = cmdLine.GetSafeArgument("-www", 0, "");
-			int iPort = (int)atoi(wwwport.c_str());
+			int iPort = std::stoi(wwwport);
 			if ((iPort < 0) || (iPort > 32767))
```
gizmocuz: Could you also adjust the port range here as well (max port 49151!)

main/domoticz.cpp
```diff
@@ -854,12 +855,19 @@ int main(int argc, char**argv)
 				return 1;
 			}
 			std::string wwwport = cmdLine.GetSafeArgument("-sslwww", 0, "");
-			int iPort = (int)atoi(wwwport.c_str());
-			if ((iPort < 0) || (iPort > 32767))
+			int iPort = std::stoi(wwwport);
+			if ((iPort < 0) || (iPort > 65535))
```
gizmocuz: Could you change the maximum port range to 49151, these are 'user' ports and should stay this way.
Just forward another port like 44444 instead of 55555

keryell: Yes, I can pick any port.
So now this PR is pretty minimal.

gizmocuz: It's unfraternally not minimal enough. There are more changes than just the two port ranges.
Using std::stoi will result in a (unhandled) crash when garbage was supplied (a non valid number, like a text)
Better keep it basic atoi

kiddigital: std::atoi is C++
atoi is C
And indeed the exception throwing of `std::stoi` is different from `std::atoi`.

keryell: This is a question on point-of-view. :-)
You need to compare with the current situation.
I have looked at the `atoi` use cases in this file and there is currently no safety net in place, if I understand the code correctly.
The problem with `atoi` https://en.cppreference.com/w/cpp/string/byte/atoi is:
> Return value
> Integer value corresponding to the contents of str on success.
> If no conversion can be performed, ​0​ is returned.
> If the value of the result cannot be represented, i.e. the converted value falls out of range of the corresponding return type, behavior is undefined. 

Undefined behavior means that the computer can blow up or a cat can die, for example.

More concretely, currently with `atoi` is:
```
./domoticz -loglevel  all -startupdelay toto
2022-01-12 20:46:48.702  Status: Domoticz V2021.1 (build 13959) (c)2012-2022 GizMoCuz
2022-01-12 20:46:48.702  Status: Build Hash: 00f9ca475, Date: 2022-01-01 12:59:09
2022-01-12 20:46:48.702  Status: Startup Path: /home/keryell/perso/Informatique/RaspberryPi/domoticz/
2022-01-12 20:46:48.702  Status: Startup delay... waiting 0 seconds...
```
or
```
./domoticz -loglevel  all -startupdelay 99999999999999999999999999999999999999999999999999999999999999999999999
2022-01-12 20:52:38.716  Status: Domoticz V2021.1 (build 13959) (c)2012-2022 GizMoCuz
2022-01-12 20:52:38.716  Status: Build Hash: 00f9ca475, Date: 2022-01-01 12:59:09
2022-01-12 20:52:38.716  Status: Startup Path: /home/keryell/perso/Informatique/RaspberryPi/domoticz/
2022-01-12 20:52:38.716  Status: Startup delay... waiting -1 seconds...
```
compared with `stoi`:
```
./domoticz -www 8080 -loglevel  all -startupdelay toto
2022-01-12 20:39:16.456  Status: Domoticz V2021.1 (build 13960) (c)2012-2022 GizMoCuz
2022-01-12 20:39:16.457  Status: Build Hash: 96b4c6766-modified, Date: 2022-01-11 09:09:45
2022-01-12 20:39:16.457  Status: Startup Path: /home/keryell/perso/Informatique/RaspberryPi/domoticz/
terminate called after throwing an instance of 'std::invalid_argument'
  what():  stoi
Aborted
```
or
```
/domoticz -loglevel  all -startupdelay 99999999999999999999999999999999999999999999999999999999999999999999999
2022-01-12 22:07:43.045  Status: Domoticz V2021.1 (build 13960) (c)2012-2022 GizMoCuz
2022-01-12 22:07:43.045  Status: Build Hash: 78bccf18b, Date: 2022-01-11 09:18:48
2022-01-12 22:07:43.045  Status: Startup Path: /home/keryell/perso/Informatique/RaspberryPi/domoticz/
terminate called after throwing an instance of 'std::out_of_range'
  what():  stoi
Aborted
```
I agree that it is not an ultimate solution for great UX design :-) ,  but at the end the `stoi` is safer since it avoids running a program with such an error by stopping immediately.
Anyway this code using the `atoi` is mostly about option parsing and thus could be updated to use instead a library like https://www.boost.org/doc/libs/1_78_0/doc/html/program_options.html I use on some of my projects https://github.com/keryell/BitWizard_SPI_relay/blob/master/examples/bw_spi_relay_control.cpp#L170 and which provides type safety, argument checking, help display...
But this is outside the scope of this PR. :-)
So what is the plan? Just keep all the `atoi` as the *status quo*?

kiddigital: I favor going C++ (std::atoi) as that is the direction... so behavior stays the same (as plain atoi).

Domoticz already uses the Boost library. Rewriting the handling of program options to use this library so in the end things become even more stable, might be a good idea. But something for a separate PR I would say.

gizmocuz: This is getting way out of hand.
We keep the default atoi, and do we really want a higher port number ? If yes, I suggest to only add this change in this PR

gizmocuz:
Increased port range from www/ssl internally to 49151

Issue #5026, opened 2021-11-19
Domoticz crashes after updating to Raspbian Linux 11 bullseye with Python plugins


wirszews:
Hi,
I Have

Domoticz Version: 2021.1 (build 13872)
Build Hash: b8adb0e92
Compile Date: 2021-11-10 07:35:29
dzVents Version: 3.1.8
Python Version: 3.9.2 (default, Mar 12 2021, 04:06:34) [GCC 10.2.1 20210110

After updating Raspbian from 10 to 11 Domoticz started to reset if we have any active plugin that uses python.
The problem also occurs on a fresh Raspbian installation.
and it does not matter which plug I use. It applies to all plugins that use python.

My crash log
```
gdb: warning: Couldn't determine a path for the index cache directory.
[New LWP 6050]
[New LWP 6051]
[New LWP 6052]
[New LWP 6053]
[New LWP 6054]
[New LWP 6055]
[New LWP 6056]
[New LWP 6057]
[New LWP 6058]
[New LWP 6059]
[New LWP 6060]
[New LWP 6061]
[New LWP 6062]
[New LWP 6063]
[New LWP 6064]
[New LWP 6065]
[New LWP 6066]
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/arm-linux-gnueabihf/libthread_db.so.1".
0x768f0f8c in __GI___clock_nanosleep_time64 (clock_id=clock_id@entry=0, flags=flags@entry=0, req=0x7e8d89a0, req@entry=0x7e8d8998, rem=0x7e8d89b0, rem@entry=0x7e8d89a8) at ../sysdeps/unix/sysv/linux/clock_nanosleep.c:52
52 ../sysdeps/unix/sysv/linux/clock_nanosleep.c: No such file or directory.
Id Target Id Frame
* 1 Thread 0x75df6240 (LWP 6049) "domoticz" 0x768f0f8c in __GI___clock_nanosleep_time64 (clock_id=clock_id@entry=0, flags=flags@entry=0, req=0x7e8d89a0, req@entry=0x7e8d8998, rem=0x7e8d89b0, rem@entry=0x7e8d89a8) at ../sysdeps/unix/sysv/linux/clock_nanosleep.c:52
[...]

PR #5087, opened 2021-12-30
Add _FILE_OFFSET_BITS=64 to GCC builds. Fixes Python plugin issues on armv7 Debian 11 builds.

michlv:
When trying latest stable version of Domoticz (2021.1) on Armbian Bullseye (armv7, Debian 11), the python plugins do not seem to work and domoticz is core dumping. After some investigation, is seems to originate from fact that readdir call, returns NULL with errno set to 75. It then demonstrates as domoticz not being able to find python plugins.

The easiest way to replicate this without actual hardware is to run armv7 armbian bullseye under chroot on amd64 Debian 11, with using qemu-arm-static as emulator.

With setting -D_FILE_OFFSET_BITS=64 the resulting binary runs without any issue.

This is because on 32bit systems the file handles are 64bit for a while, but I suspect that there were some workarounds in place to handle backward compatibilty to apps with 32bit file handles. This may have changed around glibc 2.28. This setting is compatible with both 32bit and 64bit binaries (i.e. armv7, arm64 or amd64) and should also work on older versions of Debian (Buster), although I have not tried it.

Please suggest if there is better place to put this.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -64,6 +64,7 @@ IF(CMAKE_COMPILER_IS_GNUCXX)
   #SET(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wno-switch")
   #ADD_DEFINITIONS( -Wall -O0 -ggdb )
   #ADD_DEFINITIONS( -Wfatal-errors -Wformat=2 -Werror=format-security )
+  ADD_DEFINITIONS( -D_FILE_OFFSET_BITS=64 )
 ENDIF()
 
 ### VERSIONING SUPPORT


kiddigital:
Maybe relevant for #5026 ?

kiddigital:
@dnpwwo , Is this somehow relevant/helpful for your work on trying to address Python issues with 3.9+?

kiddigital:
@michlv , against what Python version(s) did you encounter this?

michlv:
@kiddigital Debian 11 has python 3.9

michlv:
> Maybe relevant for #5026 ?

I would think that this patch will also fix that. I compiled domoticz 2021.1 with this change on top and it runs on armv7 Debian 11  under arm qemu emulation, with all my python plugins, with no issue. I did not have chance to try this on armv7 Debian 11 on actual arm hardware yet, but would think there should be no difference.

michlv:
@kiddigital I also tried to compile 2021.1 with this change on top on armv7 Debian 10 (buster) with Python 3.7.3 and domoticz works without any issue.

gizmocuz:
Thank you for this PR!!

PR #5095, opened 2022-01-08
Fix warning in TeleinfoBase.cpp

mce35:


gizmocuz:
Thank you for the quick fix, would this be the same as writing it

`bool bContactState = (teleinfo.STGE & 0x01) ? true : false;`

mce35:
Yes, fixed

gizmocuz:
Thank you again!!

hardware/TeleinfoBase.cpp
```diff
@@ -398,7 +398,7 @@ void CTeleinfoBase::ProcessTeleinfo(const std::string& name, int rank, Teleinfo&
 	{ // Process status register if found (Linky standard mode only, refer to Enedis specs for details)
 		if((teleinfo.STGE & 0x1) != (teleinfo.prevSTGE & 0x1) || teleinfo.prevSTGE == UINT32_MAX)
 		{
-			bool bContactState = (teleinfo.STGE & 0x01 != 0)?true:false;
+			bool bContactState = (teleinfo.STGE & 0x01) ? true : false;
```
keryell, line 401: Then why not
```c++
bool bContactState = teleinfo.STGE & 0x01;
```
?

PR #5096, opened 2022-01-08
Updated the "match length of the UID" logic to make it a bit more eff…

jvanderzande:
Updated the "match length of the UID" logic to make it a bit more efficient in get_auto_discovery_sensor_unit().

### DIFF
--- a/hardware/MQTT.cpp
+++ b/hardware/MQTT.cpp
@@ -2229,11 +2229,12 @@ MQTT::_tMQTTASensor* MQTT::get_auto_discovery_sensor_unit(const _tMQTTASensor* p
 			if (pTmpDeviceSensor->unit_of_measurement == szMeasurementUnit)
 			{
 				// Check the "match length" of the UID of the DEVICE with the UID of the SENSOR to get the correct subdevice in case the are multiple
-				for (int i = 1; i < (int)pSensor->unique_id.size(); ++i)
-				{
-					if (strncmp(pSensor->unique_id.c_str(), pTmpDeviceSensor->unique_id.c_str(), i) != 0)
-					{
-						// In case of a longer matching string we assume this is a better sensor to use.
+				std::string::const_iterator unid1 = pSensor->unique_id.begin();
+				std::string::const_iterator unid2 = pTmpDeviceSensor->unique_id.begin();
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thank you for your PR.
I manually changed the code a little bit and also added a check where the UniqueID are equal as this currently returned a nullptr.

New routine now is:

```
MQTT::_tMQTTASensor* MQTT::get_auto_discovery_sensor_unit(const _tMQTTASensor* pSensor, const std::string& szMeasurementUnit)
{
	//Retrieved sensor from same device with specified measurement unit
	_tMQTTADevice* pDevice = &m_discovered_devices[pSensor->device_identifiers];
	if (pDevice == nullptr)
		return nullptr; //device not found!?
	
	int iUIDMatch = -1;
	_tMQTTASensor* pDeviceSensor = nullptr;

	// Check for the correct sensor with the largest match in the UID
	for (const auto ittSensorID : pDevice->sensor_ids)
	{
		if (m_discovered_sensors.find(ittSensorID.first) != m_discovered_sensors.end())
		{
			_tMQTTASensor* pTmpDeviceSensor = &m_discovered_sensors[ittSensorID.first];
			if (pTmpDeviceSensor->unit_of_measurement == szMeasurementUnit)
			{
				if (pSensor->unique_id == pTmpDeviceSensor->unique_id)
					return pTmpDeviceSensor;

				// Check the "match length" of the UID of the DEVICE with the UID of the SENSOR to get the correct subdevice in case the are multiple
				auto ittUnID1 = pSensor->unique_id.begin();
				auto ittUnID2 = pTmpDeviceSensor->unique_id.begin();
				int iTlen = (int)(pSensor->unique_id.size() < pTmpDeviceSensor->unique_id.size()) ? pSensor->unique_id.size() : pTmpDeviceSensor->unique_id.size();
				for (int i = 1; i < iTlen; ++i)
				{
					if (*++ittUnID1 != *++ittUnID2)
					{
						// In case of a longer matching string we assume this is a better sensor to use.
						if (i > iUIDMatch)
						{
							iUIDMatch = i;
							pDeviceSensor = pTmpDeviceSensor;
						}
						break;
					}
				}
			}
		}
	}
	return pDeviceSensor;
}

```

Issue #5097, opened 2022-01-09
(EventSystem) Symbolic links not working as expected? (Non-Windows)


kiddigital:
During some testing/debugging I found that on Linux the EventSystem does _not_ use files that are symbolic links to actual files.

Quick way to test:
* `ln -s script_device_demo.py script_device_test.py` (in the `script/python` directory)
* and with a Dummy Switch called 'Test'
* When pressing the Test switch, nothing happens but I expected the `_test.py` script would have been fired.

Debugging this 'issue' learned that to the EventSystem uses a number of helper function to look for available scripts/files.

From `helper.cpp` (lines 1057-1087)
```
bool dirent_is_directory(const std::string &dir, struct dirent *ent)
{
	if (ent->d_type == DT_DIR)
		return true;
#ifndef WIN32
	if (ent->d_type == DT_LNK)
		return true;
	if (ent->d_type == DT_UNKNOWN) {
		std::string fname = dir + "/" + ent->d_name;
		struct stat st;
		if (!lstat(fname.c_str(), &st))
			return S_ISDIR(st.st_mode);
	}
#endif
	return false;
}

bool dirent_is_file(const std::string &dir, struct dirent *ent)
{
	if (ent->d_type == DT_REG)
		return true;
#ifndef WIN32
	if (ent->d_type == DT_UNKNOWN) {
		std::string fname = dir + "/" + ent->d_name;
		struct stat st;
		if (!lstat(fname.c_str(), &st))
			return S_ISREG(st.st_mode);
	}
#endif
	return false;
}
```

The reason that the script does not get executed can be found in the `dirent_is_file` function and specific at the line:
`if (!lstat(fname.c_str(), &st))`.

Here the `lstat` function ([man page](https://linux.die.net/man/2/lstat)) is used which 'stats' the symbolic link it self and _not_ the file it refers to!

Suggestion would be to use the `stat` function instead so the actual file the link refers to is examined.

Unless this current behavior is/has been intended behavior? Hope someone can share some thoughts if this should be changed or not.

gizmocuz:
I see no harm changing this to **stat** in both functions. If possible please make a PR

PR #5099, opened 2022-01-09
(Non-Windows) Use stat instead of lstat to allow symlinks to be used

kiddigital:
PR to allow the usage of symlinked script files (Python, Lua, etc.) on (Non-Windows) OS'es that support symlinks.

Before, if for example a Python event script was supposed to be used by creating a symbolic link in the `script/python/` directory to the actual script file (which resides somewhere else on the filesystem, for example in a cloned GIT repo), that file did not get evaluated as the evaluation routine checks if the file is a regular file. When `lstat` is used, a symlink is _NOT_ consider a regular file (but a symlink :) ). By using `stat`, the actual file the link points to is being examined.

See issue #5097


gizmocuz:
thank you!!

PR #5100, opened 2022-01-10
MQTT/AD: Fixes for Fibaro FGRGBW dimmer (Issue #5069)

Igitigit2:
This is the continuation of  PR #5082, but based on the code after refactoring of the MQTT class. 
Changes are essentially the same:

- The color dimmer subdevice "rgb_dimmer" does not advertise itself as an RGBW device in the AD message.
-> Add "color_mode": "true", "supported_color_modes" : ["rgbw"] to the AD message

- The value template for the rgb_command_topic is incomplete (warmWhite missing)
-> hardcoded now

- The color values are named red/green/blue, not r/g/b. This is correctly indicated by the value template, but since warmWhite is missing there, all values are now hardcoded. Decoding such a template is not implemented in Domoticz anyway.

- Color values are contained in a "value" json object , not inside "color", as Domoticz expects.

- To switch off/on, "value":"0"/"99" must be sent to the dimmer's state topic. The AD template suggest 0 and >0 and Domoticz would usually set the "state" key (not "value").
-> hardcoded 99/0 as payload_on/off and added "value" to the message sent.

- Some of the topics to subscribe to were not yet implemented in Domoticz.

On the receiving end, the changes are slightly more generic now compared to #5082 and on the sending side, the inserted code block has been shifted to help github show the real difference.

@gizmocuz: I added a small helper function RenameKey(). Ideally this should go as a member into the Json::value object, but since that is a third party module it's probably not such a good idea to amend that. Any suggestion as to where this function should go? If not, it is fine for me to leave it where it is in MQTTAutoDiscover.cpp (line 2013 ff).


gizmocuz:
Thank you for the PR !!
Could you place the RenameKey in the json_helper.cpp/h and change it to:

```
void JSonRenameKey(Json::Value& value, const std::string& srcKey, const std::string& destKey)
{
	if(!value[srcKey].empty())
	{
		value[destKey]=value[srcKey];
		value.removeMember(srcKey);
	}
}
```

Igitigit2:
@gizmocus: Done.

gizmocuz:
Copying some previous mentioned MQTT config/payload here

```
homeassistant/light/EZ-RBGW1/rgb_dimmer/config
{
	"state_topic": "zwave/EZ/12/38/0/currentValue",
	"command_topic": "zwave/EZ/12/38/0/targetValue/set",
	"brightness_scale": 99,
	
	"rgb_state_topic": "zwave/EZ/12/51/0/currentColor",
	"rgb_command_topic": "zwave/EZ/12/51/0/targetColor/set",
	"brightness_state_topic": "zwave/EZ/12/38/0/currentValue",
	"brightness_command_topic": "zwave/EZ/12/38/0/targetValue/set",
	"device": {
		"identifiers": [
			"zwavejs2mqtt_0xf3981245_node12"
		],
		"manufacturer": "Fibargroup",
		"model": "RGBW Controller (FGRGBW)",
		"name": "EZ-RBGW1",
		"sw_version": "25.25"
	},
	"name": "EZ-RBGW1_rgb_dimmer",
	"unique_id": "zwavejs2mqtt_0xf3981245_12-51-0-currentColor"
}


zwave/EZ/12/51/0/currentColor
{
	"time": 1640189400483,
	"value": {
		"warmWhite": "0",
		"red": "0",
		"green": "0",
		"blue": "99"
	},
	"nodeName": "RBGW1",
	"nodeLocation": "EZ"
}


```

gizmocuz:
We had a quick look and maybe you could rebuild this more generic...

Remove this part:

```
if (object_id == "rgb_dimmer" && pDevice->model=="RGBW Controller (FGRGBW)")
		{
			root["color_mode"] = "True";
			root["supported_color_modes"][0] = "rgbw";
			root["brightness_value_template"] = "";
			root["payload_on"] = "99";
			root["payload_off"] = "0";
		}
```

And specify the correct pSensor->color_mode and pSensor->supported_color_modes["rgbw"]=1 in maybe this check

```
if (!root["rgb_value_template"].empty())
{
			pSensor->rgb_value_template = root["rgb_value_template"].asString();
			pSensor->supported_color_modes["rgbw"]=1
}
```

Why are you clearing the brightness_value_template and not use it ?

You can check for the '**on_command_type**', if it contains '**brightness**' and root["brightness_scale"] exists use:

```
pSensor->payload_on  = root["brightness_scale"].asInt();
pSensor->payload_off  = 0;
```

Igitigit2:
> And specify the correct pSensor->color_mode and pSensor->supported_color_modes["rgbw"]=1 in maybe this check
> 
> ```
> if (!root["rgb_value_template"].empty())
> {
> 			pSensor->rgb_value_template = root["rgb_value_template"].asString();
> 			pSensor->supported_color_modes["rgbw"]=1
> }
> ```

If I do this, then the fix will be applicable to all devices which expose that template and all will be treated as rgbw, even if they are rgb only. For me that's ok, but others might have a problem with that.... If this is desired, no problem.

A similar thing is in line 2243 (InsertUpdateSwitch()):

```
				if (root["value"].isObject() && (!root["value"]["red"].empty() && root["value"]["r"].empty()))
				{ ...
```
This is now only active for exactly the message sent for the FGRGBW. If it shall be more generic, I suggest to write:
```
				if (root["value"].isObject() && (!root["value"]["red"].empty() || !root["value"]["r"].empty())) 
				{
```
Shall I implement if that way?

> Why are you clearing the brightness_value_template and not use it ?

That may indeed be a relict from my early experiments when I tried to get things working at all. I will try what happens if I use it, but that may take until Thursday.

The entire template thing is something which is really weired in AD, since a lot seems defined by the internals of homeassistant and at least I haven't found a real spec yet on which one could build.

Here some complex templates I came across so far:

Igitigit2:
I could give the brightness_template a quick try and it fails. The reason is that the color definition as well as the brightness value use the key "value". Consequently, when the brightness is written to the Json object, it deletes the color definition. Seems like the template cannot be used for the RGBW set command, so I will leave that part as it is.

jvanderzande:
Shouldn't the Brightness and the COLOR json objects be 2 separate objects and messages for this type of bulb ?
```
"rgb_state_topic": "zwave/EZ/12/51/0/currentColor",
"rgb_command_topic": "zwave/EZ/12/51/0/targetColor/set",
"brightness_state_topic": "zwave/EZ/12/38/0/currentValue",
"brightness_command_topic": "zwave/EZ/12/38/0/targetValue/set",
```
..and in this case we should only used the "brightness_value_template" for the value received in topic "zwave/EZ/12/38/0/currentValue" .... right?

So where exactly is it going wrong in the code?

Igitigit2:
@jvanderzande I agree and that is how I implemented it. Domoticz in general does however allow brightness and color to be sent in one message, but this will only work if the keys for brightness and color differ, which is not the case with the Fibaro dimmer. The general code layout is (not my code...):

```
if (command == "Set Color")
{
    if (pSensor->supported_color_modes.find("rgbw"))
        Set color values in message  
    if ((pSensor->bBrightness) || (!pSensor->brightness_value_template.empty())
        Set Brightness value in message
}
Send message
```
For the Fibaro dimmer brightness and color both use the key "value", so only one of the two will work in one message. To allow the color to pass I delete the brightness template which disables setting the brightness and thus the color command will pass.
An alternative option could be to simply check if the key used for brightness does already exist and skip setting the brightness if it does. This would also be simple and I wil try how it behaves.

As a sidenote, we are just talking about the rgb_state_topic, the brightness_command_topic is also used, but as a seperate message (as it should be).

jvanderzande:
Could we have this discussion in SLACK to make it easier to have the conversation and go through your issues? 
I am available there as @Jos
To me we should fix the logic in a way that it will handle the Value's correctly as they have unique TOPICS, so we should be able to determine which Value is coming in.

Igitigit2:
We could but I never used Slack, so first need an account and probably an invitation. Also, this morning I‘ll have some real work to do ;-) and won‘t be available before 3pm or so.

gizmocuz:
Please use the following invite link

Igitigit2:
Thanks for the invite, I am there now with the user name starting with Chris S...

gizmocuz:
Thank you very much for this PR and all your time and patience!! (Specially the debug sessions and talk on Slack!)

PR #5103, opened 2022-01-11
Notifications: Add optional http/script action

gizmocuz:


waltervl:
For documentation purposes: What is the intended use for this optional script/action on notification?

gizmocuz:
With this it is possible to execute any HTTP/HTTPS or Script action for any notification.
The previous operation for notifications could only send a notification to one of the configured systems (like pushover, http, ...), but you could not execute custom actions.
You can use this for instance to execute a script that uses text to speech (or to an Alexa) and because you are fully flexible what to execute you can make it say whatever you want regardless of the notification or custom message.
Or trigger a external HTTP servers, or a script that does specific things for that specific notification etc etc etc

PR #5105, opened 2022-01-12
Webserver fix to add charset to stock reply content-type

kiddigital:
This PR fixes the omission in the stock reply Content-Type. Previously it did not return the charset. Now it also returns `charset=UTF-8` to set the correct encoding of the returned `text/html`.

Also updated the (automated) functional tests to keep checking this and created a 'conftest.py' to contain generic Step Definitions that can be re-used by other Feature file implementations (instead of duplicating them).


gizmocuz:
Thanks!

PR #5109, opened 2022-01-14
Webserver: Fix for dirs with a dot in name that messed up extension handling

kiddigital:
PR is a improvement of PR #5084 . Although files without an extension are being served by that PR, it does fail when there was a '.' (dot) in one of the directory names.

The logic to determine the extension of the requested file did not handle that correctly. Now it does.

Thx fanabullunet for persisting with testing.


gizmocuz:
Thank you!!

PR #5110, opened 2022-01-14
Change temperature digits from 1 to 2

jgaalen:
Use 2 digits for temperature as many sensors send 2 digits of temperature. This significance is now lost which.
Also use 2 digits for pressure as that significance is also lost.

### DIFF
--- a/hardware/AnnaThermostat.cpp
+++ b/hardware/AnnaThermostat.cpp
@@ -35,8 +35,6 @@
 #define sAnnaFailedBurnerStarts             17
 #define sAnnaBurnerOperationTime            18
 #define sAnnaHotWaterBurnerOperationTime    19
-#define sAnnaHotWaterFlow                   20
-#define sAnnaHotWaterTemperature            21
 
 const std::string ANNA_VERSION       = "1.0.1";
 const std::string ANNA_GET_STATUS    = "/core/appliances";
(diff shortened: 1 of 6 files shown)


gizmocuz:
Thank you for this PR.

These structures/routines are also used for other hardware that have no support for 2 digits temperature.

Two digits is also not needed, 1 digit is more then enough.

Sure you might connect a temp sensor to your ESP/Arduino project, but that sensor is not calibrated in a lab, and surely is not accurate enough.

jgaalen:
Regarding the 2 digits, there is a difference between accuracy, correctness and precision. Although sensors might not be accurate or correct on their measurement, the precision (rounding/significance) still makes sense. It should be up to the sensor or sensor plugin to determine the value and precision, not the core engine (domoticz). The core/engine/domoticz should take whatever value is provided and not discard any data.
I have a case (anna thermostat), which provides temp with 2 digits and also bar with 2 digits. Domoticz rounds it to 1 digit and that deletes some detail which does makes sense and is valuable.
For some reason, python plugins which sends temperature values can add values with 2 digits and I make very good use of the 2nd digit as it gives more precision (not accuracy). With multiple values you can round it, so it gets more precise with 2 digits. I you see it in a graph, same sensor, with 2 digits or rounded to 1 digit, you will immediately see that 2 digits does make sense and adds more sensible detail.
Also the pressure/bar value is very accurate, but domoticz discards the 2nd digit.

That majority of sensors sends data with 1 digit is still fine right? In the gui the values are rounded to 1 digit anyway, but I'd like to propagate the real value to influxdb and using it in dzvents without having it rounded by domoticz before.

gizmocuz:
The correct display of a temperature value in Domoticz is 1 digit and is perfectly fine

jgaalen:
Sure, for display purpose (which is then up to the gui designer), but the core value behind, which can be used in dzvents, influxdb whatever, also gets impacted by this decision. It should be up to the sensor and sensor plugin developer to determine which value should be stored right? Not the core engine which cannot interpret the correctness, it should just take and store the full value it is provided. Please consider it as it does makes sense (not only for temperature, but also for pressure and probably a lot of other types).

The python plugins (deconz with zigbee temp sensors) are perfectly able to store temperature values with 2 digits and domoticz seem to just operate fine. 

Someone please back me up on this discussion... :)

gizmocuz:
That last digit is totally bogus !

Anyway, this is the accuracy of your thermostat

 Temperature
» Measurement accuracy: ±0,5°C
» Measurement resolution: 0,1°C

Source: https://www.plugwise.com/product/anna

Now let's focus on things that matters more in life then 2 digits ;)

PR #5112, opened 2022-01-14
Merging RFLINK gateway MQTT device

pagocs:
Finally I would like to merge my RFLINK gateway MQTT device implementation to the official version because may somebody else also can use this not just me... :) For using this device it should make a little hardware changes on the original RFLINK hardware what I documented here: https://github.com/pagocs/esp32-rflinkmqttgateway

In the pull request there is a little thing what you should be consider. Because my continuous manual merge this implementation to the development branch until now  I added some reserve to the _eHardwareTypes in the RFXNames.h to avoid the device "driver" ID conflicts. Of course it is not necessary for the final merging so it could be removed.


gizmocuz:
Thank you very much for this PR! Does your project also work with a Arduino Ethernet Shield ?
This PR failed because you forgot to add the needed files to the visual studio project. Could you fix this ? (If possible test with visual studio)

There is also a small compile warning

RFLinkBase.cpp(358,13): warning C4244: 'initializing': conversion from 'double' to 'int', possible loss of data 

Would be great if you could support the MQTT Auto Discovery protocol, that might benefit other home automation projects as well.

Why did you not use the RFLink with a Ethernet Shield and use it with TCP, this method is already implemented in Domoticz

pagocs:
> Thank you very much for this PR! Does your project also work with a Arduino Ethernet Shield ?

No, my project works just with my ESP32 extension.

> This PR failed because you forgot to add the needed files to the visual studio project. Could you fix 
> this ? (If possible test with visual studio)

Sorry for this, I basically do not use win so I did not realised this issue. . I guess I fixed the  the compiling error, but I cannot tested yet because I do not have win compiling environment, but I am on it...

> RFLinkBase.cpp(358,13): warning C4244: 'initializing': conversion from 'double' to 'int', possible loss 
> of data

The warning is fixed

>Would be great if you could support the MQTT Auto Discovery protocol, that might benefit other 
> home automation projects as well.

I will check this, but I hope lack of this is not an obstacle  for the merge... :) 

> Why did you not use the RFLink with a Ethernet Shield and use it with TCP, this method is already 
> implemented in Domoticz

My original goal was create a solution for using  433Mhz devices (sensors and actors) independently on multiple host back in 2017.  The main reason behind this to build a home automation system which is could be fault tolerant at least partly. So based on my research I could not find any simple solution  which is support the connection from multiple sources so that is why I built my solution on a different approach. Of course the 433 mhz devices not the best solutions but there are some reliable temp sensors, weather stations and cheap sockets and they will remain with us for years... That is why I think this solution is could be helpful to others.

main/RFXNames.h
```diff
@@ -222,6 +222,11 @@ enum _eHardwareTypes {
 	HTYPE_AirconWithMe,         //123
 	HTYPE_TeleinfoMeterTCP,		//124
 	HTYPE_MQTTAutoDiscovery,	//125
+	HTYPE_RESERVED_FOR_OTHERS3,	//126
```
gizmocuz: Could you please not use 'HTYPE_RESERVED_FOR_OTHERS' here, but just your new hardware type ?
You can change the hardware type in your database

www/app/hardware/Hardware.js
```diff
@@ -3916,7 +3917,9 @@ define(['app'], function (app) {
 							else if (HwTypeStr.indexOf("HTTP/HTTPS") >= 0) {
 								HwTypeStr += ' <span class="label label-info lcursor" onclick="CreateDummySensors(' + item.idx + ',\'' + item.Name + '\');">' + $.t("Create Virtual Sensors") + '</span>';
 							}
-							else if (HwTypeStr.indexOf("RFLink") >= 0) {
+							else if( (HwTypeStr.indexOf("RFLink") >= 0 &&
+									HwTypeStr.indexOf("MQTT") == -1 ) ||
```
gizmocuz: Do you need both MQTT checks here ?

HwTypeStr.indexOf("MQTT") == -1 ) ||
									HwTypeStr.indexOf("RFLink Gateway MQTT")

Should this not be HwTypeStr.indexOf("MQTT") != -1 )

gizmocuz:
Could you make use of the 'StoppableTask' method, the bIsStopRequested boolean code is obsolete and slow
I also do not think you need to overwrite Start/Stop hardware, have a look in the MQTT Auto Discovery class what should be overwritten
You can close connections in on_disconnect

I don't understand that part with multiple IP connections... I think that should be removed as well
Your RFLink will publish to 1 destination IP/Broker i suppose, it will get a mad house if it transmits to random MQTT brokers
If your broker stops frequently you better check why this happens and/or install a process monitoring system like 'monit'

The on_disconnect error you see is not related to Domoticz, I don't see this, and neither had reports of this one.
I do know when you are debugging (with a breakpoint) and this takes to long, my MQTT broker kicks me out and we will reconnect
So please remove that , and check how the MQTT Auto Discovery class handles the on_disconnect please

www/app/hardware/Hardware.js
```diff
@@ -3916,7 +3917,7 @@ define(['app'], function (app) {
 							else if (HwTypeStr.indexOf("HTTP/HTTPS") >= 0) {
 								HwTypeStr += ' <span class="label label-info lcursor" onclick="CreateDummySensors(' + item.idx + ',\'' + item.Name + '\');">' + $.t("Create Virtual Sensors") + '</span>';
 							}
-							else if (HwTypeStr.indexOf("RFLink") >= 0) {
+							else if (HwTypeStr.indexOf("HTTP/HTTPS") >= 0) {
```
gizmocuz: I am not sure if the change in hardware.js is correct at the moment

pagocs:
> Could you make use of the 'StoppableTask' method, the bIsStopRequested boolean code is obsolete and slowI

I implemented the new method.

> also do not think you need to overwrite Start/Stop hardware,

It should to implement the start / stop because this class is derived from the CRFLinkBase not from the MQTT class. I tried to add the MQTT class to the hierarchy but unfortunately that is need some extra research and changes because the WriteInt member functions is different in:
rflinkbase.h:: virtual **bool** WriteInt(const std::string &sendString) = 0;
mqtt.h:: **void** WriteInt(const std::string &sendStr) override;

and the RFLINK base is used in more devices as parent not just in mine...

> have a look in the MQTT Auto Discovery class what should be overwrittenYou can close connections in on_disconnect

Sorry, but I am not override because my class is not derived from the same class. 

> I don't understand that part with multiple IP connections... I think that should be removed as wellYour RFLink will publish to 1 destination IP/Broker i suppose, it will get a mad house if it transmits to random MQTT brokersIf your broker stops frequently you better check why this happens and/or install a process monitoring system like 'monit'

The method is transparent so if you do not specify more target it looks like tthe multiple target is not implemented.  But anyway the concept behind this is the same as I mention earlier: the fault tolerance. The mosquitto broker is support the bridge mode (http://www.steves-internet-guide.com/mosquitto-bridge-configuration/) when the broker is share the configured topics with a remote broker. So with this you can use two broker which can distribute all the messages  with their connected clients.  My ESP32 firmware is also support this target shift so based on this if one of the brokers is died both the ESP32 and the and domoticz could be change the target and everything will be fine. Please keep in mind this method is planned for using iin multiple domoticz server environment, so if one node is broken the other one is can work further.
Monit is a good tool and I am using it. But I am afraid monit cannot do anithing if the SD card or SSD will die in a RPI node and you are at mountains for skiing, but somebody at home and (for example) your house heating is running on a domoticz system, like mine?

Please do not forget if you do not specify more IP this feature is fully transparent.

> The on_disconnect error you see is not related to Domoticz, I don't see this, and neither had reports of this one.I do know when you are debugging (with a breakpoint) and this takes to long, my MQTT  broker kicks me out and we will reconnectSo please remove that , and check how the MQTT Auto  Discovery class handles the on_disconnect please

I cannot use same method like MQTT Auto discovery as I mention earlier. Anyway the implementation basically  same as in the MQTT.CPP which is called in the MQTT Auto discovery on_disconnect member function except one error handling. Based on my experience the domoticz version 4.11554  has an MQTT related issue and this was a workaround  for it. After this fix I never checked further the issue existence so I do not know it is still can happen or not. The fix based on to skipping he whole reconnect sequence with leave untouched the m_bDoReconnect boolean variable. I commented out the workaround but I still leave there an error log for further detection this special case.

- [x] I will commit the cahnges tomorrow morning...

Sorry for the double commit....

gizmocuz:

pagocs:
Thank you for the merging!

PR #5113, opened 2022-01-17
Fix a bug in variables name in alertlevel function

guillaumeclair:
lines 85/93 : 'level' should be replaced by 'flevel'
line 95: 'level' should be equal to 4 when flevel > 98
lines 77/85/93 should be reordered.


gizmocuz:
Thank you for this PR!

PR #5120, opened 2022-01-21
MQTT AD Scripts add Userid and Password support

jvanderzande:
Added the options -u & -P to be able to specify the Userid and Password for the MQTT daemon.


gizmocuz:
Thank you!

PR #5126, opened 2022-01-23
Added get_build_versions to easily get ta list of Domoticz version

jvanderzande:
Added get_build_versions to easily get ta list of Domoticz version and the commit it is based upon.
Also shows the previous 20 commits (by default) to easily list what was changed before that build.

### DIFF
--- a/scripts/get_build_versions.sh
+++ b/scripts/get_build_versions.sh
@@ -1,0 +1,37 @@
+#!/bin/bash
+: '
+# Scriptname getcommitlog.sh     Ver: 20220123-02
+# Created by JvdZ
+# base info from https://www.domoticz.com/forum/viewtopic.php?f=47&t=30405#p230094
+
+Script will list 20 commitlog entries preceding provided version, to easily list changes made around that version
+
+usage:
+sh getcommitlog.sh [-v version] [-m max_logEntries]
+   -v domoticz version. default last
+   -m number of records to show. default 20
+'
+
+MAXREC=20;
+IVER=""
+# process parameters
+while getopts m:v: flag
+do
+	case "${flag}" in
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thank you!!

waltervl:
Thanks! If someone could find a way to get this uploaded in the wiki then this would be very helpful.

kiddigital:
> Thanks! If someone could find a way to get this uploaded in the wiki then this would be very helpful.

What do you mean 'uploaded to the wiki'? Paste the script there also? Why? Easiest way it to add a link like [this](https://github.com/domoticz/domoticz/blob/development/scripts/get_build_versions.sh).

But that shouldn't be needed, as this scripts expect that the GitHub repo of Domoticz is cloned first as it uses the Git information (so not a normal install using the installers) and then this script is of course also there as it is in the repo.

Are you looking for documentation on these scripts? Maybe and idea to create a MarkDown file and have the Wiki link (or even better embed) to this file.

@jvanderzande , not sure if the 'scripts' directory is the best place. It is a script indeed, but this directory is meant/used by Domoticz itself to execute scripts from. Which is not the purpose for this script. Maybe there should be a other/new directory with 'supportscripts' or something (including a READme.md :) )

jvanderzande:
> @jvanderzande , not sure if the 'scripts' directory is the best place. It is a script indeed, but this directory is meant/used by Domoticz itself to execute scripts from. Which is not the purpose for this script. Maybe there should be a other/new directory with 'supportscripts' or something (including a READme.md :) )

That's what @gizmocuz  requested. Either way works for me. ;) 
We could also move the MQTT record & send scripts in that new subdirectory, so people get them distributed when upgrading and keep these type of scripts together.

PR #5127, opened 2022-01-24
Fix logfilename when search string contains invalid filecharacter

jvanderzande:
Fix logfilename when search string contains invalid file characters by removing them from the logfilename


gizmocuz:
Thank you!

PR #5129, opened 2022-01-25
Teleinfo: Combine used/delivered energy in same meter

mce35:
Teleinfo: Instead of having separate meters for used and delivered energy, combine both in the same (existing) P1 Smart Meter.


gizmocuz:
Thank you!!

PR #5130, opened 2022-01-25
Allow to select serial ports using /dev/serial/by-path

mce35:
Allow to select serial ports using /dev/serial/by-path, this allows to distinguish between multiple USB serial ports using the same chips (in my case multiple PL2303, only 1 is visible in /dev/serial/by-id)

### DIFF
--- a/main/Helper.cpp
+++ b/main/Helper.cpp
@@ -348,7 +348,7 @@ std::vector<std::string> GetSerialPorts(bool &bUseDirectPath)
 
 #else
 	//scan /dev for /dev/ttyUSB* or /dev/ttyS* or /dev/tty.usbserial* or /dev/ttyAMA* or /dev/ttySAC* or /dev/ttymxc*
-	//also scan /dev/serial/by-id/* on Linux
+	//also scan /dev/serial/by-id/* and /dev/serial/by-path/* on Linux
 
 	bool bHaveTtyAMAfree=false;
 	std::string sLine;
@@ -464,23 +464,27 @@ std::vector<std::string> GetSerialPorts(bool &bUseDirectPath)
 	}
 
 #if defined(__linux__) || defined(__linux) || defined(linux)
-	d=opendir("/dev/serial/by-id");
-	if (d != nullptr)
+	auto scanDir = [&ret](std::string dir)
 	{
-		struct dirent *de = nullptr;
-		// Loop while not nullptr
-		while ((de = readdir(d)))
+		DIR *d = opendir(dir.c_str());
+		if (d != nullptr)
 		{
-			// Only consider symbolic links
-                        if (de->d_type == DT_LNK)
-                        {
-				std::string fname = de->d_name;
-				ret.push_back("/dev/serial/by-id/" + fname);
+			struct dirent *de = nullptr;
+			// Loop while not nullptr
+			while ((de = readdir(d)))
+			{
+				// Only consider symbolic links
+				if (de->d_type == DT_LNK)
+				{
+					std::string fname = de->d_name;
+					ret.push_back(dir + "/" + fname);
+				}
 			}
+			closedir(d);
 		}
-		closedir(d);
-	}
-
+	};
+	scanDir("/dev/serial/by-id");
+	scanDir("/dev/serial/by-path");
 #endif
 #endif
 	return ret;


gizmocuz:
It's unique as long as the manufacturer follows the rules about giving each device a unique serial number.
Cheap hardware might not do this.
The problem with using 'by-path' is that all gets mixed up when you stick them into different USB ports ?
Then you could just use /dev/ttyUSB2

mce35:
I agree that theses devices are missing a serial number (I have 3 different models (PL2303, CH340, Zigate v2), none of them has a SN). In such cases it is not possible to use symlinks in by-id when multiple devices of the same model are connected (it links to the last device connected).
If I use ttyUSBX, then I am not sure that the order of devices will stay the same at each reboot. At least when using by-path I am sure that the device will not change unless it is physically connected to another USB port.

gizmocuz:
The idea of using serial-by-id is that it does not matter in what port you stick the USB device
by-path is the same as using ttyUSBxx , so there is no reason for to add this as it will change when you switch usb ports

mce35:
by-path is different than ttyUSBX, links in by-path always point to the same USB (physical) port, whereas ttyUSBX depends on the order on which the devices are discovered/connected. For example, I just disconnected/reconnected 2 devices in the opposite order:
```
pi@rasp3:~/domoticz $ ls -lh /dev/serial/by-path/
total 0
lrwxrwxrwx 1 root root 13 23 janv. 14:26 platform-3f980000.usb-usb-0:1.2:1.0-port0 -> ../../ttyUSB2
lrwxrwxrwx 1 root root 13 23 janv. 13:17 platform-3f980000.usb-usb-0:1.3.1:1.0-port0 -> ../../ttyUSB3
lrwxrwxrwx 1 root root 13 23 janv. 14:25 platform-3f980000.usb-usb-0:1.3.3:1.0-port0 -> ../../ttyUSB0
**lrwxrwxrwx 1 root root 13 23 janv. 13:17 platform-3f980000.usb-usb-0:1.4:1.0-port0 -> ../../ttyUSB1
pi@rasp3:~/domoticz $ ls -lh /dev/serial/by-path/
total 0
lrwxrwxrwx 1 root root 13 26 janv. 12:11 platform-3f980000.usb-usb-0:1.2:1.0-port0 -> ../../ttyUSB1
lrwxrwxrwx 1 root root 13 23 janv. 13:17 platform-3f980000.usb-usb-0:1.3.1:1.0-port0 -> ../../ttyUSB3
lrwxrwxrwx 1 root root 13 23 janv. 14:25 platform-3f980000.usb-usb-0:1.3.3:1.0-port0 -> ../../ttyUSB0
lrwxrwxrwx 1 root root 13 26 janv. 12:11 platform-3f980000.usb-usb-0:1.4:1.0-port0 -> ../../ttyUSB2
```
As you can see ttyUSB1/2 have been reversed, so using by-path ensure that domoticz is still using the same device for the same meter (unless they are connected to different USB ports of course).

xbeaudouin:
Well you may also do what I use on FreeBSD : 

```
[root@domoticz /etc/devd]# cat rfx.conf
attach 10 {
     device-name "uftdi[0-9]+";
     match "vendor"            "0x0403";
     match "product"        "0x6001";
     match "sernum" "A11HEUGX";
     action "logger RFX device attached on $cdev/$device-name/$sernum/tty$ttyname, creating symlink and setting owner.;  /bin/ln -fs /dev/tty$ttyname /dev/ttyUrfx; chown root /dev/ttyUrfx";
};

```

This allow the device to be attached to a special /dev/ttyUrfx whatever it is plugged.

mce35:
Of course I could add a new udev rule to select each devices by path (as they do not have serial numbers) and link them to ttyUSB\<something>, but in this case why not just use already existing links in the by-path directory?

gizmocuz:
Because that will display a ton of new devices again in the serial port drop-down

mce35:
Except if you have many serial devices, it will only add a few items in the list (4 in my case).

gizmocuz:
So you have 4 ttyUSB , 4 serial-by-path (and users with serial devices that have a serial number get also 4 serial-by-path
I'm not to happy with this PR, I thank you for it of course, but your the first on in a very long time requesting this.
In your case I would make a udev rule

mce35:
If somebody has the same issue here is an example udev rules to create a symlink based on the path:
```
SUBSYSTEM=="tty",ENV{ID_PATH}=="platform-3f980000.usb-usb-0:1.4:1.0",SYMLINK+="ttyUSBPZ1"
```
This rule needs to be added in a new file in `/etc/udev/rules.d`. This will create a new link `/dev/ttyUSBPZ1`

To get available variables related to a device:
```
udevadm info /dev/serial/by-path/platform-3f980000.usb-usb-0\:1.4\:1.0-port0
```


To reload the rules without rebooting:
```
udevadm control --reload-rules && udevadm trigger
```

gizmocuz:
Why don't you use the serial-by-id?

mce35:
Because I have 2 devices for pointing to the same serial-by-id (no SN provided by the device, same chip)
```
/dev/serial/by-id:
total 0
lrwxrwxrwx 1 root root 13 20 janv. 08:50 usb-1a86_USB2.0-Serial-if00-port0 -> ../../ttyUSB2
lrwxrwxrwx 1 root root 13 20 janv. 08:50 usb-FTDI_FT232EX-if00-port0 -> ../../ttyUSB3
lrwxrwxrwx 1 root root 13 20 janv. 08:50 usb-Prolific_Technology_Inc._USB-Serial_Controller-if00-port0 -> ../../ttyUSB1

/dev/serial/by-path:
total 0
lrwxrwxrwx 1 root root 13 20 janv. 08:50 platform-3f980000.usb-usb-0:1.2:1.0-port0 -> ../../ttyUSB0
lrwxrwxrwx 1 root root 13 20 janv. 08:50 platform-3f980000.usb-usb-0:1.3.1:1.0-port0 -> ../../ttyUSB2
lrwxrwxrwx 1 root root 13 20 janv. 08:50 platform-3f980000.usb-usb-0:1.3.3:1.0-port0 -> ../../ttyUSB3
lrwxrwxrwx 1 root root 13 20 janv. 08:50 platform-3f980000.usb-usb-0:1.4:1.0-port0 -> ../../ttyUSB1
```

`ttyUSB0` and `ttyUSB1` would be the same in serial-by-id so only 1 of them is available.

gizmocuz:
Seems like a design fault for that device.. and it's not something you can solve by ft_prog ?
But this is a very rare situation

PR #5135, opened 2022-01-28
Added duplicated RFLINK MQTT packet filtering

pagocs:
I added a small function to filtering the duplicated RFLINK packets. It is very useful if you use multiple 433mhz decoder devices which can receive the same device signal parallel.


gizmocuz:
Thank you for this PR.
I am not sure what you mean by "if you use multiple 433mhz decoder devices which can receive the same device signal parallel"

Why would you do so ? Maybe a bigger antenna or other setup location ?

If this is from 1 RFLink device, I think this should be done inside RFLink

What would be the problem if two or three of the same temperature values are received ?

pagocs:
> I am not sure what you mean by "if you use multiple 433mhz decoder devices which can receive the same device signal parallel"
> Why would you do so ? Maybe a bigger antenna or other setup location ?

I develop custom sensors and actors based on ESP32 soc and I add RFLINK receiver to it, because:
- can add some currently not supported (for example Acurite rain sensor) 433 mhz devices (I contacted the RFLINK gateway developer but he still keep the source closed) 
- can cover some areas in my house where / from the signals cannot be reached elsewhere (for example my garage partly underground so I cannot receives signals from there)

> If this is from 1 RFLink device, I think this should be done inside RFLink

No, it is originated from different devices.

> What would be the problem if two or three of the same temperature values are received ?

Temperature is not a problem, but the buttons or remotes it could be if you receive the push signals (which is just could sen the pushed event) more times. It could be filter in the handler scripts but as I guess more reasonable if it filtered centrally...

pagocs:
@gizmocuz pls

hardware/RFLinkMQTT.cpp
```diff
@@ -259,11 +260,49 @@ void CRFLinkMQTT::on_message(const struct mosquitto_message *message)
...
+		if( seqnumend != std::string::npos && seqnumend > 3 )
+		{
+			struct timeval tnow;
+			std::string cpofmsg = qMessage;
+
+			// 20;77;EV1527;ID=0f7499;SWITCH=04;CMD=ON;
+			//    ^^^
+			// Remove the sequence number from payload
+			cpofmsg.erase( 3 , seqnumend-3 );
+			// Calculate a CRC value for paylod comparision -- Crc32 defined in Hhelper.cpp
```
gizmocuz: Please double check this sentence

hardware/RFLinkMQTT.cpp
```diff
@@ -259,11 +260,49 @@ void CRFLinkMQTT::on_message(const struct mosquitto_message *message)
...
+
+			// Get time in millisec for sub second comparision
+			gettimeofday(&tnow, nullptr);
+			time_t msecs = (tnow.tv_sec * 1000) + (tnow.tv_usec / 1000);
+
+			// _log.Log(LOG_NORM, ">>> RFLINK MQTT: Message to chksum: %s CRC: %x time diff: %ld" , cpofmsg.c_str() , crc , msecs-m_lastmsgTime );
+
+			if( m_lastmsgCRC == crc && (msecs-m_lastmsgTime) < 420 )
+			{
+				// if the CRC is equal in the allowed period it should be filtered
```
gizmocuz: if the CRC is equal in the specified period it should be ignored

gizmocuz:
Thank you for this PR!!

pagocs:
Thank you

PR #5138, opened 2022-01-31
add header and ns for strstr

jpalus:
Fixes build on my ARMv6/Linux/GCC 11.2.0

### DIFF
--- a/hardware/EnOceanRawValue.cpp
+++ b/hardware/EnOceanRawValue.cpp
@@ -1,6 +1,7 @@
 #include "stdafx.h"
 #include "EnOceanRawValue.h"
 #include <string>
+#include <string.h>
 
 
 


gizmocuz:
Thank you for your PR! Question, why not use the 'c' function strstr instead of std::strstr, you do not need the include for that ...

jpalus:
Actually original version failed compilation since it apparently missed either `<string.h>` or `<cstring>`. Decided to include `<cstring>` so I also went for `std::`, though admittedly I'm not C++ developer and don't know what are preferences in this regard. But I thought namespaced symbols are _the right way_  in C++ world.

gizmocuz:
I think it is best to include `<string.h>` and use `strstr ` as this is used throughout the program, there is not a single std::strstr that I am aware of ;)

gizmocuz:
Thank you!

PR #5143, opened 2022-02-04
Enocean changes

ludos78:
2 small adaptations :
- smoke detector value rarely change, so update value unconditionally to get last update date (typically my smoke detector send a frame every 20 minutes and i want to know if it is still alive)
- Automatic Meter : use Channel and Data Type values to separate data sources

note: ESP2 change not tested, only mimic changes done on ESP3 Automatic Meter.


gizmocuz:
Thank you for this PR!

I have a few comments/suggestions

Instead of using 'sDecodeRXMessage', please use SendKwhMeter, SendMeterSensor
Please do not store channel values in fillers etc as this structure can change any time
There are two id's (id1 and id2), see SendMeterSensor
Please do check if by using these functions you do not create a new device (use the correct node/child id, or make a sql update patch)

About the smoke sensor, with your current patch this will always tricker a On/Off event, and what you want is to update the LastUpdate value only if the switch state is the same
My suggestion (if you really really need that information ?) is to use part of the logic of the **CDomoticzHardwareBase::SendSwitch** function and update the LastUpdate value

pseudo:
```
	std::string szLastUpdate = TimeToString(nullptr, TF_DateTime);
	m_sql.safe_query("UPDATE DeviceStatus SET LastUpdate='%q' WHERE (ID == %" PRIu64 ")",
		szLastUpdate.c_str(),
		idx);
```

Or have a look at

CNestOAuthAPI::UpdateSmokeSensor

PR #5144, opened 2022-02-05
Remove Appcache functionality

gizmocuz:


kiddigital:
Any plans to start using Service Workers for certain/all pages?

gizmocuz:
Yes but not for all pages.. Do you think that's needed? Most browsers will cache the payload anyway.

kiddigital:
I think that the Domoticz UI can work very well without any Service Workers being the SPA that it is now. With the (current) use of web-sockets, bringing up-to-date dashboard and notifications, I don't see any additional features yet that Service Workers would bring. But that could also be because I am not a front-end developer :)

If it is just about caching, I think there is no need for Service Workers as caching is taken care of by the browser. Maybe some tweaking with some of the caching headers from the webserver perspective might be needed to optimize things, but not that I know off on forehand.

gizmocuz:
Closed PR because of upcoming commit

PR #5145, opened 2022-02-05
Feature/webserverlogs: Option for the webserver to produce Apache Combined LogFile format weblogs

kiddigital:
This PR adds a new commandline option `-weblog <weblog file>` which tells Domoticz webserver(s) to start logging all web request to the specified logfile.

The logfile uses the [Apache Combined Logfile Format](https://httpd.apache.org/docs/2.4/logs.html#accesslog) as this a very common single-line logformat that is understood by most logfile analysers. This means these logs can easily be (automatically) analyzed and presented in nice overviews.

**NOTE**: the logtime is in milliseconds and not just seconds! So instead of using the standard '%t' as time parser, here '%f' is used. This is also already common for a long time as a lot of requests are handled by webservers within a second, so having milliseconds to order these requests is very welcome. Depending on the loganalyser, you sometimes need to tell the analyser that it needs '%f' instead of '%t' to parse the time correctly.

The logfile is automatically closed and re-opened after writing 100.000 loglines (resulting in logfiles between 15-50 MB). This enables log rotating (and compression after it is closed) _without the need to re-start_ Domoticz.

A common use-case for this option is to log all access to Domoticz (webserver), done through the default UI, API's or other apps/clients, when the access to Domoticz is _open_ to the outside world, larger networks like buildings/venues or just to see the (ab)use in your own network.

Using modern tools like [ELK](https://www.elastic.co/what-is/elk-stack), [GrayLog](https://www.graylog.org/) or even [Splunk](https://www.splunk.com/), but also older/simpler command line tools like [goaccess](https://goaccess.io/man) or [AWstats](https://github.com/eldy/AWStats) can create insights from these logfiles. (Most of these tools are available in standard Debian or RPM based package managers).

Using `awstats`:
awstats

Realtime using `goaccess`:
goaccess

_This logging is NOT enabled by default. So if you want this logging to be generated, add this option to the Domoticz startup script._

**NOTE for Raspberry PI users**: Generating these logfiles means constant writing to the filesystem. When running Domoticz on a Raspberry Pi off a SD-card, this logging feature can really cause a lot of wear on such a card and shorten the lifespan of these cards. If you want to get these logs even when running on a Pi, either attach a separate (USB) harddrive to write to. Or have the logging forwarded to another machine using `syslog`. This can be done by specifying _'syslog:'_ as the _'weblog file'_. But this only works when you enabled `syslog` for the other logging as well! See the `-syslog` commandline option.

### DIFF
--- a/History.txt
+++ b/History.txt
@@ -1,5 +1,6 @@
 Version 2022.xxxxx (xxxx)
 - Implemented: Blockly, added 'Toggle' set command
+- Implemented: Commandline option '-weblog <weblogfile>' which creates Apache Combined Logformat logs containg all webrequests
 - Fixed: Crash when secure webserver could not start and applying settings
 - Fixed: Notification system, prevent double start
 - For a full detailed overview visit: https://www.domoticz.com/wiki/Domoticz_versions_-_Commits
--- a/main/Helper.cpp
+++ b/main/Helper.cpp
@@ -810,30 +810,34 @@ time_t GetClockTicks()
 	return(tv.tv_sec * 1000 + tv.tv_usec / 1000);
 }
 
+void CurrentDateTimeMillisecond(tm &timeinfo, timeval &tv)
+{
+#ifdef CLOCK_REALTIME
+	struct timespec ts;
+	if (!clock_gettime(CLOCK_REALTIME, &ts))
+	{
+		tv.tv_sec = ts.tv_sec;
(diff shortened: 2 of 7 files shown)


kiddigital:
Hm.. Windows build fails.. will take a look asap.

gizmocuz:
Thank you again for this PR!!

PR #5148, opened 2022-02-07
TTNMQTT: Fix to use DeviceSerial (again) as mac

kiddigital:
Small PR to use the correct Device Serial identifier.

This was changed in previous changes to switch to the new V3 stack, but the previous serial used was not unique. Now it should be (again as it was in V2).

See [this](https://www.domoticz.com/forum/viewtopic.php?p=286100) forum thread.


gizmocuz:
Thanks!

PR #5150, opened 2022-02-08
Update Python Events template with a working example

### LINKED ISSUE #3959: Python domoticz.log() broken
Hi

I have problem with running base Python script:

```
import domoticz
import random

domoticz.log("Hi Domoticz!")
```

The output log says:
```
2020-03-01 00:45:08.128 Error: EventSystem: Failed to execute python event script "TestPython"
2020-03-01 00:45:08.128 Error: EventSystem: TypeError: function takes exactly 1 argument (2 given)
2020-03-01 00:45:08.128 Error: EventSystem:
2020-03-01 00:45:08.128 Error: EventSystem: The above exception was the direct cause of the following exception:
2020-03-01 00:45:08.128 Error: EventSystem:
2020-03-01 00:45:08.128 Error: EventSystem: Traceback (most recent call last):
2020-03-01 00:45:08.128 Error: EventSystem: File "<string>", line 4, in <module>
2020-03-01 00:45:08.128 Error: EventSystem: File "/config/scripts/python/domoticz.py", line 42, in log
2020-03-01 00:45:08.128 Error: EventSystem: domoticz_.Log(0, " ".join([str(k) for k in args]))
2020-03-01 00:45:08.128 Error: EventSystem: SystemError: <built-in function Log> returned a result with an error set
2020-03-01 00:45:08.538 Error: Pyhton Event System: Failed to parse parameters: string expected.
```
I seen the same problem was reported before, but there was problem with function name where `log()` had to be `Log()`, now clearly from console you can see it is uppercase `Log()` but there seems extra argument appear

waltervl:
The current python example is completely outdated and throwing errors.
The python part is now the same as scripts/python/script_device_demo.py

### DIFF
--- a/scripts/templates/All.Python
+++ b/scripts/templates/All.Python
@@ -60,19 +60,19 @@ TODO: setting variables
 
 Calling Python's print function will not print to the domoticz console, see below
 """
-import domoticz
-import random
+import DomoticzEvents as DE
 
-print ("This will only show up in the shell where you start domoticz");
-domoticz.log("Hi Domoticz!")
+DE.Log("Python: Changed: " + DE.changed_device.Describe())
 
-domoticz.log("The device that got changed is: ", changed_device_name)
-# changed_device.name is the same string
+if DE.changed_device_name == "Test":
+    if DE.Devices["Test_Target"].n_value_string == "On":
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thank you!!

PR #5151, opened 2022-02-08
Master

Jorkek:
the P1 meter has an issue for brussels based (SIBELGA) smart meters. The string of the identifier is too long and this creates an error. I found the solution here https://en.domoticz.cn/forum/viewtopic.php?t=30562. My apologies if I am doing this wrong. I fixed the code on my machine and now would like to show you how I've done it


Jorkek:
sorry, wrong

PR #5152, opened 2022-02-08
Update P1MeterBase.cpp

Jorkek:
fixed to help with Sibelga smart meters that use a too long string as an identifier.
explained on : https://en.domoticz.cn/forum/viewtopic.php?t=30562

### DIFF
--- a/hardware/P1MeterBase.cpp
+++ b/hardware/P1MeterBase.cpp
@@ -224,7 +224,7 @@ bool P1MeterBase::MatchLine()
 	uint8_t i;
 	bool found = false;
 	const P1Match *t;
-	char value[20] = "";
+	char value[100] = "";
 	std::string vString;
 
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thank you for this PR!

I think this code could be optimized quite a bit.... I will have a go at it because I think value should be of type **std::string** as well

gizmocuz:
@Jorkek , I made some changes in the latest development branch related to the P1 code.
Is it maybe possible that you could verify as well that all is still working ?

Jorkek:
> @Jorkek , I made some changes in the latest development branch related to the P1 code.
> Is it maybe possible that you could verify as well that all is still working ?

Hi Gizmozuz, 

Thanks ! In tested it and it still works. 
thanks for cleaning up the code !

gizmocuz:
Thanks for the feedback and testing!

PR #5156, opened 2022-02-13
Round brightness/dimmer level to avoid 1% becoming 0% for ZWAVE

jvanderzande:
Fix for ZWAVE devices sending 0% when set to 1% in DOMOTICZ.


gizmocuz:
Thank you again!!

PR #5161, opened 2022-02-16
Minor fixes to EnOcean ESP3

efdeb:
Fix CheckAndUpdateNodeRORG behaviour :
Change node type from Generic to Teached-in only when RORG is updated
Add EEP name/description when EEP changes and not specified manually by the user

Apply minor fixes : typos, rescope functions as class members, etc...


gizmocuz:
Thank you again!!

PR #5167, opened 2022-02-19
TTNMQTT: Skip empty messages (and wrong external decoded messages)

kiddigital:
Small PR that skips message processing when there was not data payload in the base message (for example when only MAC commands or _ping/keepalive_ messages are send.)

Some external payload decoders generate a wrong decoded_payload message when they shouldn't as there is/was no payload data to begin with. These wrongly decoded payloads are now ignored as there is a check to see if there was any real original data payload to start with.

See [this](https://www.domoticz.com/forum/viewtopic.php?p=287052&sid=9a1cf94a066b63d39dc537d1a1d0b64e#p287052) forum topic for details.

### DIFF
--- a/hardware/TTNMQTT.cpp
+++ b/hardware/TTNMQTT.cpp
@@ -691,6 +691,11 @@ void CTTNMQTT::on_message(const struct mosquitto_message *message)
 		Json::Value endDeviceIds = root["end_device_ids"];
 		Json::Value applicationIds = endDeviceIds["application_ids"];
 
+		if (uplinkMessage["frm_payload"].empty())
+		{
+			return;		// When there is no frm_payload, there is no data. Not even from a payload decoder.
+		}
+
 		//Get data from message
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thank you!!

xbeaudouin:
Seems... on Clang  we have some warning:

```
/usr/home/kiwi/git/freebsd-ports/www/domoticz-devel/work/domoticz-8f01ed77d/main/../hardware/RFLinkMQTT.h:49:15: warning: 'WriteInt' overrides a member function but is not marked 'override' [-Winconsistent-missing-override]
        virtual bool WriteInt(const std::string &sendString); // override;
                     ^
```

kiddigital:
Different module? RflinkMQTT and not TTNMQTT?

gizmocuz:
Yes correct, it is for RFLinkMQTT.h

PR #5168, opened 2022-02-19
Changed P1 Max W from 17250 (25A) to 55200 (80A)

jvanderzande:
Changed P1 Max W from 17250 (25A) to 55200 (80A).
See issue reported in forum: https://www.domoticz.com/forum/viewtopic.php?p=286926#

### DIFF
--- a/hardware/P1MeterBase.cpp
+++ b/hardware/P1MeterBase.cpp
@@ -491,12 +491,12 @@ bool P1MeterBase::MatchLine()
 				break;
 			case P1TYPE_USAGECURRENT:
 				temp_usage = (unsigned long)(std::stof(sValue) * 1000.0F); // Watt
-				if (temp_usage < 17250)
+				if (temp_usage < 55200) // max 80A fuse 3fase 230V
 					m_power.usagecurrent = temp_usage;
 				break;
 			case P1TYPE_DELIVCURRENT:
 				temp_usage = (unsigned long)(std::stof(sValue) * 1000.0F); // Watt;
-				if (temp_usage < 17250)
+				if (temp_usage < 55200) // max 80A fuse 3fase 230V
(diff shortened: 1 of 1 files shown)


kiddigital:
@jvanderzande , maybe would should make this somehow configurable via the Settings screen and have a reasonable default (25A?)?

The check is there to prevent odd readings (spikes, glitches, transfer errors, etc.) so where to draw the line for _normal_ readings. I guess that most households nowadays are still at 25A? But with solar-panels, electric cars, electric heating systems, etc. there will definitely be more and more households that go above this.

What would be good _levels_ to offer users on such a settings screen? Only 16A, 25A, 80A ? or even more when people have a 400V connection? Or should it be in Watts as not only 230V is used, but some are at 110V or 400V.

jvanderzande:
> What would be good _levels_ to offer users on such a settings screen? Only 16A, 25A, 80A ? or even more when people have a 400V connection? Or should it be in Watts as not only 230V is used, but some are at 110V or 400V.

It looks like there is quite a variety of options of Main Fuses Amps used and internationally indeed the 110 V  (not really 400V as that is the same as 230V where 230V is the voltage between the Fase and NUL and 400 is the Voltage between 2 Fases.).
So I guess it would make sense to allow for a Max Total Wattage in stead of a Max AMPs when there is a need to make a Max for validity of the W value. 
Just wondering: Is there a past experience where this value was showing a wrong value ?  ... and was that between 17250 - 55200  or much larger?

gizmocuz:
Yes there was a problem with NeoCoolcam wall plugs. This is probably caused by a wrong bit inside OpenZWave and should not happen with ZWaveJS2MQTT hopefully

gizmocuz:
Thank you !!

PR #5169, opened 2022-02-19
Also increased the MAX Power per Fase from 10000 to 18400 (230V*80A)

jvanderzande:
Also increased the MAX Power per Phase from 10000 to 18400 (230V*80A)


gizmocuz:
Thanks again!!

PR #5171, opened 2022-02-19
Fix rounding issue in Counter Report Fixes 5162

### LINKED ISSUE #5162: Counter divider on RFXMETER Counter display issues
When setting a counter divider on RFXMETER type counter gives display issues in reports.
Seems to be js code solvable so will try to fix.

Screenshot_20220216-235159-840

waltervl:
Fix rounding issue in Counter Report Fixes https://github.com/domoticz/domoticz/issues/5162
The + sign will fix to 3 digits but remove trailing zero's So for integers it will stay integer.

PS previous Merge into Domoticz/Development was unintentional. Luckily a very small change.... Sorry for that.


gizmocuz:
Thanks for your PR but this is not the fix!
This is not the correct way to solve this...
I do not need any decimals because my data is delivered correctly and my divider is therefor 1

I made some changes , please have a look at: (beta 14220)

https://github.com/domoticz/domoticz/commit/fecded4ee21ca226eaadc86a0a8cf5ee69f87456

now these changes also need to be applied in the 'log' (day/week/month/year)

waltervl:
Yes, much better. Thanks!

Issue #5172, opened 2022-02-22
Error 'dzVents: local netWork not open for dzVents openURL call' for other service on same host - missing port check


sander1988:
There is a small bug in this check: https://github.com/domoticz/domoticz/blob/master/main/dzVents.cpp#L398

It doesn't check the port of the URL being called with the dzVents openURL function. So it fails with 'dzVents: local netWork not open for dzVents openURL call' when I call another service (a different port) on localhost.

Note that the value of setting 'Local Networks (no username/password)' is empty in my setup. Adding localhost here it would be a workaround but is not preferred for security reasons (in this case).

waltervl:
It is specified in the DzVents documentation that you have to set localhost 127.0.0.* in settings Network so perhaps it is intentional.

But if you think it can be fixed please make the change, and create a PR.

sander1988:
> It is specified in the DzVents documentation that you have to set localhost 127.0.0.* in settings Network so perhaps it is intentional.
> 
> But if you think it can be fixed please make the change, and create a PR.

Sure, I will make a PR. It might take some time, I have to setup a build env first.

For now I came up with a workaround... 

I created the record 'myhost' in /etc/hosts which also points to 127.0.0.1 . So when I call openURL with 'myhost' as domain, then it bypasses the check and still works.

jvanderzande:
Not sure whether the Domoticz port is a know value in dzventz, so wondering how that could be checked/fixed in the linked logic.

sander1988:
> Not sure whether the Domoticz port is a know value in dzventz, so wondering how that could be checked/fixed in the linked logic.

The check itself is in C (https://github.com/domoticz/domoticz/blob/master/main/dzVents.cpp#L398) so that would not be an issue I expect.

jvanderzande:
True... :)

jvanderzande:
@sander1988, Just tested adding a test for the http & https ports to the linked code,  so it will only do the check when the url contains either port and that seems indeed to fix the issue. 
Will make a PR unless you are ready to do so ....  just let me know. ;)

waltervl:
DzVents uses OpenURL also for some internal functions. Is that also still working?

Edit: example function device.setDescription(description)

https://github.com/domoticz/domoticz/blob/53758e5b4825d620f47c997a52847bf367848e09/dzVents/runtime/device-adapters/generic_device.lua#L135

sander1988:
[...]

PR #5173, opened 2022-02-24
Patch will fix issue #5172 to ensure only the http and https ports us…

jvanderzande:
Patch will fix issue #5172 to ensure only the http and https ports used by Domoticz will be checked in OpenUrl check for "Local Networks".
This will allow OpenUrl calls to 127.0.0.1 to other ports than used by Domoticz, when the Local Networks field doesn't contain 127.0.0.*
I've tested the Patch and it is working fine, but I still do not understand what won't work in dzVents when 127.0.0* isn't in the Local Networks.  So does anybody with more integral knowledge of dzVents know why that check was put in in the first place?

### DIFF
--- a/main/dzVents.cpp
+++ b/main/dzVents.cpp
@@ -395,16 +395,29 @@ bool CdzVents::OpenURL(lua_State *lua_state, const std::vector<_tLuaTableValues>
 	}
 
 	// Handle situation where WebLocalNetworks is not open without password for dzVents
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thank you for the PR! Going to to a little refactoring after the merging to make it a bit more c++

waltervl:
The test should be that if a user/password is used in Domoticz  the Network settings should contain 127.0.0.* else DzVents cannot work as it sometimes uses internal Json API URL's without User/password entries. If there is no User/password implemented in the Domoticz environment there is no need for this check within OpenURL()

PR #5174, opened 2022-02-25
Netatmo +  Hue Changes

### LINKED ISSUE #4882: Netatmo.cpp : fail to check for valves if user have both Weather Station and Thermostat
In hardware/Netatmo.cpp, the only way the code check for radiator valves is if m_NetatmoType is set to NETYPE_ENERGY. And this happens on line 1178 in GetMeterDetails() only if code fails to parse json API response for NETYPE_WEATHER_STATION and NETYPE_HOMECOACH (the same logic applies to setting the type to NETYPE_HOMECOACH)

As a consequence when the code check for thermostat data in the Do_Work() function, the GetThermostatDetails() , check for the m_NetatmoType, and if not set the NETYP_ENERGY, it the use the URI "https://api.netatmo.net/api/getthermostatsdata" which only provides data for the thermostat without the valves (and may be deprecated as no documentation is availabe for this in the Netatmo API), instead of calling https://api.netatmo.net/api/homesdata that does have all the data for the thermostat and the valves.

In my opinion, netatmo products are not exclusive : you can have the weather station, the aircare station and the thermostat range (boiler / valves), there is also security products (camera / smoke alarm / sensors)

BUT the API for the thermotsat (https://api.netatmo.net/api/homesdata and https://api.netatmo.net/api/homstatus) does also carry the data for the weather station (and I presume it does carry the data for the aircare station but I'm not sure as far as the API documentation does not state this is the case, even for the weather station)

So the correct way to handle this should be to rewrite the logical decision in the code , there is four product range in Netamo API : Security / Energy / Weather / Aircare. The "Energy" range can carry data for Weather (and maybe AirCare and Security). Curently only Energy / Weather / Aircare is supported and they should be checked in this order 👍 

1) Energy
2) Weather
3) Aircare

The quick fix would be in the function GetMeterDetails() to put the region "//Check if the user has an Enery device (thermostat)" begining at line 1147 through 1181 out of the condition block if (!ParseNetatmoGetResponse(sResult, m_NetatmoType, false)) begining at line 1107

Also a nice feature to add to this hardware is the ability to set the schedule of the thermostat : in the application you can create different temperatures set by room, so as to adapt the heating to some situations. For instance I use this to turn off heating in summer with a "summer" and a "winter" schedule.

The api https://api.netatmo.com/api/homesdata does allow to retreive the schedules (in the "schedules" array, data of interest is schedule id and schedule name) and  the api https://api.netatmo.com/api/switchhomeschedule allow to switch the schedules (home id and schedule id are required)

Note : as a Netatmo user with both weather station and thermostat with radiator valves, I can confirm that radiator valves are not discovered.

StephaneM60:
Netatmo changes to support weather station and radiator valves. 

Philips Hue changes to support color change on some vendors bulb (like LIDL) : the idea here is to send the color change as XY values instead of HLS (LIDL bulb doesn't handle HLS correctly)


kiddigital:
@StephaneM60 , there are a few files in this PR that are from your build environment and should not be part of the Domoticz source code (like .vs/ProjectSettings or out/build/x64-Debug/VSInheritEnvironments.txt). Please remove these from your PR.

Also, there are a few formatting changes, like indenting, spaces, tabs, etc.. These are not actual changes, just your editor with different settings. And it also makes it very unclear to see the actually intended changes in the source code. Would be much better if the PR just contains your actual changes (and leave out the formatting differences). You probably have to update some settings in your IDE to not automatically modify the formatting.

Would be great if you can update the PR. Thx.

gizmocuz:
Thank you for this PR. If possible in the feature split them into different PR's when they involve different topics.

StephaneM60:
Hi,

For the format changes, I checked and are your referring to changes like : char * transformed into char* (thoses are the only kind of changes I have not made purposefully) otherwise tab / space wise I made sure I used the same as the original.

I don't know where Visual Studio hide the settings reponsible for this for sure, I can see a "default format style" that is currently on "Visual Studio" (of course) and it offers me "LLVM" / "Google" / "Chromium" / "Mozilla" / "WebKit".

Also for the Netatmo.cpp, I have mad really heavy changes on this one as I refactored the code.

Best Regards.

gizmocuz:
The format changes are ok as these are correct and VS does that so no worries
Could you undo the changes in gitignore, or at least remove the last debug line , also make sure there is a return at the end of the last line

image

StephaneM60:
Sure I can do this, though If I remove the last line, how can I prevent VS to push the file back ?

gizmocuz:
Thank you for the changes! One more request, please undo the changes in

msbuild/domoticz.vcxproj

Or upgrade to VS 2022 ... It's a worthy upgrade and community edition is free t use (or commercial if your revenue is less then 1M $)

StephaneM60:
Ok I have manually changed the platform tools (I had made sure that in the IDE I reverted the changes in order to compile to verify all was Ok : as I made the changes back in August, to publish the modifications without complications, I deleted the project I was working / compiling in August and started from a fresh Git pull, reapplying the only modifications I made in summer then pushing them, without compiling anything, that's why there was build errors). Hopefully this time the IDE didn't try to force anything else without me noticing ;)

gizmocuz:
Thank you very much for this PR and work !!

Issue #3119, opened 2019-03-18
Netatmo: Error connecting to Server...


alaunay:
Hello,

Using the last git domoticz, I'm unable to connect to the netatmo api, domoticz gives me a

2019-03-18 12:24:47.053  Error: Netatmo: Error connecting to Server...

I modified a bit the cpp file to print what it sends:

2019-03-18 12:24:46.989  Error: httpUrl: https://api.netatmo.net/oauth2/token
2019-03-18 12:24:46.989  Error: httpData: grant_type=password&client_id=5588029e485a88af28f4a3c4&client_secret=6vIpQVjNsL2A74Bd8tINscklLw2LKv7NhE9uW2&username=modified%40modified.com&password=nope&scope=read_station read_thermostat write_thermostat read_homecoach

And tried directly on CLI:

$ curl -s -d "grant_type=password&client_id=5588029e485a88af28f4a3c4&client_secret=6vIpQVjNsL2A74Bd8tINscklLw2LKv7NhE9uW2&username=modified%40modified.com&password=nope&scope=read_station read_thermostat write_thermostat read_homecoach" "https://api.netatmo.net/oauth2/token"

{"access_token":"hexastuff|hexastuff","refresh_token":"hexastuff|hexastuff","scope":["read_station","read_thermostat","write_thermostat","read_homecoach"],"expires_in":10800,"expire_in":10800}

It works on CLI, but it seems the HTTPClient part in the domoticz doesn't do what it needs...

I've no idea how to debug this, c++ isn't something I know (more than printing a few variables).

Hints ?

lorenzo-deluca:
I've the same problem.
I receive errors on email and sometimes I receive several mail with "Netatmo: Error connecting to Server..." Probably this is a Netatmo server problem, is it possible to remove from errors, maybe only an info log message ?
It's on Netatmo.cpp

`bool ret = HTTPClient::POST(httpUrl, httpData, ExtraHeaders, sResult);
	if (!ret)
	{
		_log.Log(LOG_ERROR, "Netatmo: Error connecting to Server...");
		return false;
	}`

Thanks to everyone.

pipiche38:
same and still existing problems, can it be related to a too often polling

gizmocuz:
Yes it could be related. It is currently set to 500 seconds (5 minutes)
Is it possible you can change/test this ?

pipiche38:
@gizmocuz you know that I'm more a python man than a C++, but yes, I'll try that .

[...]

PR #5180, opened 2022-03-03
Preventing Server Error message

pipiche38:
This PR is about to address #3119.
Despite the fact the Netamo documentation say that they update their information every 10minutes, doing the polling with the same period seems to be creating issue.
Moving from 600s (10 minutes) to 900s ( 15 minutes) seems to reduce from my test drastically the number of Server Errors, but still getting one or two per day.


gizmocuz:
Thank you !!

PR #5189, opened 2022-03-13
Change .noselect to allow selection of log text

laplug:
Proposal to change the "user-select: none" to: "user-select: text" to allow selection and copying of log text. 
See: https://www.domoticz.com/forum/viewtopic.php?f=6&t=38189&sid=67c41df497c0dbf532c72136083c33de )
.noselect {
  -webkit-touch-callout:         none; /* iOS Safari */
    -webkit-user-select:         none; /* Safari */
     -khtml-user-select:         none; /* Konqueror HTML */
       -moz-user-select:         none; /* Firefox */
        -ms-user-select:         none; /* Internet Explorer/Edge */
            user-select:         text; /* Non-prefixed version, currently
                                          supported by Chrome and Opera */
	-webkit-tap-highlight-color: rgba(0,0,0,0);
}

### DIFF
--- a/www/css/style.css
+++ b/www/css/style.css
@@ -2248,7 +2248,7 @@ table.display thead th.sorting_disabled div.DataTables_sort_wrapper {
      -khtml-user-select:         none; /* Konqueror HTML */
        -moz-user-select:         none; /* Firefox */
         -ms-user-select:         none; /* Internet Explorer/Edge */
-            user-select:         none; /* Non-prefixed version, currently
+            user-select:         text; /* Non-prefixed version, currently
                                           supported by Chrome and Opera */
 	-webkit-tap-highlight-color: rgba(0,0,0,0);
 }


gizmocuz:
Thank you for this PR. The style 'no-select' should be no-select, so this patch is not valid and need to be done elsewhere.
But, When I go to Setup->Log I am able to copy/paste log text
Maybe you are talking about another log ?
Maybe you need to create a new style .doselect and assign this to some parts you want to select?

laplug:
Thank you for the feedback.

I do mean Setup->Log.
Without the 'solution' I cannot copy log text. I use Edge, Firefox, Chrome.
Others report the same: https://www.domoticz.com/forum/viewtopic.php?f=6&t=38189&sid=528222f397cc04e781df131be6c44a60 and also when copying became disabled (for us).
Could there be something else different between your and our setup or browsers that causes copy being disabled?

gizmocuz:
Please do not reply by email directly as it exposes sensitive information.

Are you using the default Domoticz theme ? I can really copy/paste/select there

laplug:
Correct, I use the default theme.
I really can't copy without the proposed 'solution' from the French forum.

gizmocuz:
**Please do not reply by email directly as it exposes sensitive information.**

I had no issues using the 'Edge' browser, but I added a new CSS style to enable selection and added this class to the log window

Could you check beta 14277, this should solve the issue

laplug:
Thank you, that works!
I will inform the users in the forum.

BTW: I didn't install the beta because I am way behind in updates. But after editing your solution in the log.html and style.css copying works fine.

laplug:
This PR is not valid and nolonger needed. 
Problem is solved in beta 14277

PR #5190, opened 2022-03-13
Fix (Linux)compile warnings by casting Py_ssize_t to int

kiddigital:
PR to fix compile warnings under Linux.

Example, see [this](https://github.com/domoticz/domoticz/runs/5526926451?check_suite_focus=true) output from one of the last build actions.

And the build output from the [PR-check action of this PR](https://github.com/domoticz/domoticz/runs/5527109232?check_suite_focus=true) should be free of these compile warning :)


gizmocuz:
Thank you! Yes there are some more, try building a 64bit windows version...

PR #5191, opened 2022-03-13
Implement force-select to allow copying About info

laplug:
A request was made (by the same users for the Log copying) to allow copying of About information, so I included the force-select.
I hope that it is a valid PR this time. It is handy to be able to copy the version info.
TIA


gizmocuz:
Thank you!!

PR #5203, opened 2022-03-27
Merge RemoteProxyIP's to Local Networks

### LINKED ISSUE #5199: Reverse proxy code broken
Today i found out that de settings for the reverse proxy are gone. 
I have my domoticz install behind a nginx reverse proxy. I whitelisted the ip-range of my home-network so no login is needed. My reverse proxy is in the same ip-range so it didn't need an password either.This was solved by the setting for the "RemoteProxyIP" so everyone outside of my homenetwork needed a password to access by the reverse proxy.

In the  code submit of 9th of march i see "Remove old proxy code".  Probably the setting and with that the security measure for asking the login credentials disappeared. Is this by design and do i need to find a new way to set up my reverse proxy or is this a bug?

gizmocuz:


kiddigital:
@gizmocuz , looks great! I will give it a try asap.

gizmocuz:
Thanks! As far as I could test it's working correctly. Old proxy addresses are added to the local network if it was not already there

www/views/setup.html
```diff
@@ -236,17 +236,6 @@ <h2><span data-i18n="AutoBackup"></span>:</h2>
 							</div>
 							<div class="row-fluid">
```
kiddigital, line 237: @gizmocuz , I think this whole diff has to be removed... now an empty line remains

kiddigital:
> Old proxy addresses are added to the local network if it was not already there

I could not find any code that migrates the `WebRemoteProxyIPs` configured IP's to the `WebLocalNetworks`? I think there should be a DB upgrade for that? Add the list of `WebRemoteProxyIPs` to the `WebLocalNetworks` and delete the Proxy variable.

kiddigital:
When looking and thinking about the code (which has been there for a long time already) and doing a little reading on Proxies (an interesting one at [Mozilla](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For)), some refactoring might be recommended.

But maybe we leave it for now, meaning back to a working Proxy situation as it was before.

When we do some security related updates, maybe this current behavior should be revisited and optimized (like first check for proxy header existence before going through the loop, remove X-real-IP as a valid header, add other headers, allow multiple occurrences of the header, etc.)

www/views/setup.html
gizmocuz: Done

gizmocuz:
@kiddigital , somehow the changes for the database migration where not pushed. Done now.
The detection of the headers could be done in the request_parser.cpp perhaps... 
Bdw got some issues with 'lodash.lua'... I know you already made a issue about this, but perhaps we could include this file or temp. disable these checks ?

kiddigital:
> @kiddigital , somehow the changes for the database migration where not pushed. Done now. 

Great

> The detection of the headers could be done in the request_parser.cpp perhaps... 

I will take a look...

> Bdw got some issues with 'lodash.lua'... I know you already made a issue about this, but perhaps we could include this file or temp. disable these checks ?

Indeed known issue... yesterday I (temporarily) disabled this check for now

PR #5212, opened 2022-04-06
Fix tracking of EnOcean blind position when using multiple blinds.

gpouger:
Fix for the issue discussed on the forum at https://www.domoticz.com/forum/viewtopic.php?p=288803 by adding tracking of the last used blind node id.
Also renamed the LastPosition member to match the naming convention of the rest of the class.

I compiled and tested my fix on my Raspberry Pi and checked that it now works as expected.

It's my first ever pull request, please be indulgent if I didn't do it right ;-)


gizmocuz:
Thank you very much for this PR!!

PR #5222, opened 2022-04-23
Fix chill calculation

IanDury:
When Domoticz calculates the wind chill temperature it wrongly shows a chill temperature of 0 when it does not meet:
if ((temp < 10.0) && (wspeedms >= 1.4))


gizmocuz:
Thank you for this PR!!

PR #5226, opened 2022-04-29
Add multiply field to UVI widget

IanDury:
UV sensors are typically very sensitive to aging.  Often this can be calibrated with a gain correction applied to the receiving weather station. This PR provides a solution to apply a gain correction to the Domoticz UVI widget.

image
image

### DIFF
--- a/main/mainworker.cpp
+++ b/main/mainworker.cpp
@@ -4330,7 +4330,7 @@ void MainWorker::decode_UV(const CDomoticzHardwareBase* pHardware, const tRBUF* 
 		BatteryLevel = 0;
 	else
 		BatteryLevel = 100;
-	float Level = pResponse->UV.uv;
+	float Level = float(pResponse->UV.uv) / 10.0F;
 	float AddjValue = 0.0F;
 	float AddjMulti = 1.0F;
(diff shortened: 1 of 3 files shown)


kiddigital:
Hi @IanDury , could you provide some description for this PR and a few screenshots so it is more clear why and what this PR does? Thx

IanDury:
Wrote description as requested

gizmocuz:
@IanDury , thank you for this PR and your time.
As you have seen, there are also UVI sensors that contains a temperature sensor.
Currently it is already possible to have a temperature correction, and this was done in the normal adjustment/multiply
In your PR I see you now assume that this is done with "GetAddjustment2", but that is not correct.
In your case, you do want to have the adjustment for the UVI to be in these fields and leave the 'GetAddjustment' for the temperature.
When you update the UVI sensor you need to use 'addjmulti2' instead of 'addjmulti' and use GetAddjustment2 in your code

IanDury:
I assumed GetAddjustment and GetAddjustment2 were identical and should be used in order of appearance.
Now using GetAddjustment2 and addjmulti2 for UVI multiply.

kiddigital:
@IanDury , thx for the update but shouldn't the UI update call also not be modified to store the multiplier into the addjmulti2 instead?

IanDury:
You're right. Overlooked this.

kiddigital:
Thanks!

PR #5230, opened 2022-05-08
Get QueuedCommands

louisthissen:
Would be nice if we could have a function that tels us if there is a queued command on a device.
For example:
`if domoticz.devices(53).QueuedCommands then...`

PR #5231, opened 2022-05-08
Get whoUpdate

louisthissen:
Would be nice to have a function that tels us who did the last update on a device.

For example:
If a curtain was closed by human it should not go open automatically after some time, but if it was closed by script it should.

`if domoticz.devices(53).whoUpdate == "script" then
domoticz.devices(53).open()
end`


kiddigital:
Same problem as #5230

PR #5253, opened 2022-06-03
Initialize hartbeats and small refactor (fixes #5252)

### LINKED ISSUE #5252: Domoticz does not start correctly at power on time
Hello,

Context of the issue :

- Domoticz V2022.1 (beta build from source 14338)
- Python 3.7.3
- running on RPI3B+

Sometimes Domoticz does not restart correctly after a power on.

It generates a signal SIGUSR1 (10) which stops domoticz. 

Consequently I have to launch manually the command "sudo reboot" to restart again Domoticz which is a trouble after a storm caused poweroff.

I discovered the potential root cause of the issue.

In main\SignalHandler.cpp a watchdog thread is created and tests in heartbeat_check() function if the Domoticz  main thread and the mainworker thread are still alive by  checking if the elapsed time from respectively m_LastHeartbeat and  m_mainworker.m_LastHeartbeat have not timed out.

If the watchdog thread is running before the main thread or the mainworker thread have initialized their own  m_LastHeartbeat  data with the actual time then the watchdog thread might raise a SIGUSR1 signal depending on random memory values of these data.

A potential fix for the issue is either for the watchdog thread to:

Option 1 : wait a bit before entering  its working loop : sleep_milliseconds(30000) to let the main thread and mainworker thread initialize the data
Option 2 : to initialize the m_LastHeartbeat and  m_mainworker.m_LastHeartbeat data with the actual time  before entering  its working loop.
	m_mainworker.m_LastHeartbeat = mytime(nullptr);
	m_LastHeartbeat = mytime(nullptr);

BR

kiddigital:
Small fix for issue #5252 preventing the watchdog to stop Domoticz before it actually started the mainthread and mainworker. This could happen occasionally due to not properly initialized heartbeat variables and 'slow' startup.  (thx @mealgi)

### DIFF
--- a/main/SignalHandler.cpp
+++ b/main/SignalHandler.cpp
@@ -493,6 +493,7 @@ void signal_handler(int sig_num
 
 static void heartbeat_check()
 {
+	bool bTimeOut = false;
 	time_t now;
 	mytime(&now);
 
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thank you both for this PR!!

PR #5256, opened 2022-06-08
get rid of std::ptr_fun

neheb:
Seems it's deprecated under GCC 12

### DIFF
--- a/hardware/Comm5SMTCP.cpp
+++ b/hardware/Comm5SMTCP.cpp
@@ -16,7 +16,7 @@
 */
 
 static inline std::string &rtrim(std::string &s) {
-	s.erase(std::find_if(s.rbegin(), s.rend(), std::not1(std::ptr_fun<int, int>(std::isspace))).base(), s.end());
+	s.erase(std::find_if_not(s.rbegin(), s.rend(), ::isspace).base(), s.end());
 	return s;
 }
 
--- a/hardware/Comm5TCP.cpp
+++ b/hardware/Comm5TCP.cpp
@@ -18,7 +18,7 @@
 */
 
 static inline std::string &c5_rtrim(std::string &s) {
-	s.erase(std::find_if(s.rbegin(), s.rend(), std::not1(std::ptr_fun<int, int>(std::isspace))).base(), s.end());
+	s.erase(std::find_if_not(s.rbegin(), s.rend(), ::isspace).base(), s.end());
 	return s;
 }
 
--- a/hardware/PiFace.cpp
+++ b/hardware/PiFace.cpp
@@ -92,14 +92,14 @@ CPiFace::CPiFace(const int ID)
 // trim from start
 std::string & CPiFace::ltrim(std::string &s)
 {
-    s.erase(s.begin(), std::find_if(s.begin(), s.end(), std::not1(std::ptr_fun<int, int>(std::isspace))));
+    s.erase(s.begin(), std::find_if_not(s.begin(), s.end(), ::isspace));
     return s;
 }
 
 // trim from end
 std::string & CPiFace::rtrim(std::string &s)
 {
-    s.erase(std::find_if(s.rbegin(), s.rend(), std::not1(std::ptr_fun<int, int>(std::isspace))).base(), s.end());
+    s.erase(std::find_if_not(s.rbegin(), s.rend(), ::isspace).base(), s.end());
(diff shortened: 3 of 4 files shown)


kiddigital:
@neheb , thx for the finding.

Looking at the code, it seems we have multiple implementation of trim functions in our code. Maybe we should do a small refactor and have the 3 hardware modules addressed in the PR, to use the Helper functions instead of doing their own implementations?

And maybe we can speed optimize the Helper functions by rewriting it differently?
And also assumption is now that it removes 'spaces', but should it be removing any White Space (\r \n \t etc) as well?

Here are some nice examples I found.
https://www.techiedelight.com/trim-string-cpp-remove-leading-trailing-spaces/

@gizmocuz , any preferences on direction to address this (upcoming) issue?

neheb:
tested that link with godbolt. 67 instructions vs 111. not bad.

kiddigital:
> tested that link with godbolt. 67 instructions vs 111. not bad.

Which one... the first option (using `find_first_not_of()`)? Compared it with? Your proposed code?

neheb:
Yeah.

gizmocuz:
the trim function should only remove spaces from the front or behind ... not the spaces in '\r \n \r \n'

gizmocuz:
There already exists some optimized code in the helper functions but it is marked out... needs to be tested if that works correctly.
It's the same as from your link (option 2)

kiddigital:
> the trim function should only remove spaces from the front or behind ... not the spaces in '\r \n \r \n'

So only spaces and not whitespace characters like '\t' or '\n'.

Is there a separate cleaning function for that?

neheb:
> > the trim function should only remove spaces from the front or behind ... not the spaces in '\r \n \r \n'
> 
> So only spaces and not whitespace characters like '\t' or '\n'.
> 
> Is there a separate cleaning function for that?

He’s saying beginning and end, not in the middle. This patch keeps things as they are. Just removes deprecated/removed C++98 stuff.

kiddigital:
> > > the trim function should only remove spaces from the front or behind ... not the spaces in '\r \n \r \n'
> > 
> > 
> > So only spaces and not whitespace characters like '\t' or '\n'.
> > Is there a separate cleaning function for that?
> 
> He’s saying beginning and end, not in the middle. This patch keeps things as they are. Just removes deprecated/removed C++98 stuff.

Agree that it is only about _space_ at the beginning or end, and not somewhere in the middle.

Question is about the definition of __space__. Do we mean the single space character (ASCII 0x20) or what commonly is referred to as _whitespace_ (see [here](https://www.techiedelight.com/remove-whitespaces-string-cpp/) for example). Many standard library _space removal_ functions remove whitespace characters and not only the '  ' (0x20). For example looking at what `std::isspace` refers to, it refers to a set of whitespace characters and not only the singe ' ' character.

So examples:
- '\n Hello John ' -> 'Hello John' or '\n Hello John' ?
- ' \tHello William' -> 'Hello William' or '\tHello William' ?
- '\r\nHi Jane\r\n' -> 'Hi Jane' or '\r\nHi Jane\r\n' ?
- '  \t  Goodbye Queen Mary \n  ' -> 'Goodbye Queen Mary' or '\t  Goodbye Queen Mary \n' ?

Personally, it makes sense to me that we mean all whitespace characters and not only the single 0x20 because most often these type of `trim()` functions are used to cleanup pieces of strings coming from certain input that has been chopped up in pieces (reading file lines, interpreting HTTP input, reading from i/o input, etc.).

Just want to come to a common understanding. I will be fine as well with a `trim()` functions (and related `ltrim()` and `rtrim()`) that only looks at the space character. Than it makes sense to have a `wtrim()` and `wltrim()`, etc. as well. Or maybe to be more inline with general usage looking at other languages that have build-in `trim()` function, have the default trim whitespace and have e separate `sltrim()`, etc. which only looks at the single space character.

gizmocuz:
trimming is meant for spaces infront or behind text. Not to cleanup whitespaces
Until this moment there has been no need yet for writing a wide-string implementation for this project...

kiddigital:
Ok.. clear.. 

And btw, I meant _whitespace_ when I wrote `wltrim()` :)

gizmocuz:
Thanks for the PR! I have removed the duplicate functions in the hardware classes and changed/tested the helper trim functions

Issue #5217, opened 2022-04-16
Crazy blinds behavior


TheUlpio:
Domoticz version 2022.1 build 14312 (the version doesn't matter)
Zwave js version 9.0.1
Zvavej22mqtt version 6.7.0
Actuators  FIBARO FGR222 Roller Shutter 2 fw 25.25

Sorry for my bad English. It's very hard for me to correctly explain things, so I've made a video [https://www.mediafire.com/file/zzmvwcechia9qds/2022-04-16_13-55-23.mkv](https://www.mediafire.com/file/zzmvwcechia9qds/2022-04-16_13-55-23.mkv). The quality is far from optimal, but isn't easy to compress a 3840*1920 resolution in a manageable file. Apologize me. The devices settings are shown in that video.

Maybe related to this: [https://github.com/domoticz/domoticz/issues/5124](https://github.com/domoticz/domoticz/issues/5124), but it's related to the "blinds inverted" type.
In the above indicated issue, was complained a strange behavior in blinds reaction over domoticz commands and related (wrong?) state shown in controls. Thanks to devs, that issue was resolved by restoring - for OZW users - the old behavior.  

Always a OZW user, my blinds worked well set as "blind inverted with positioning". Then I switched to mqtt-ad (built from sources and packed as an executable). Despite other devices (but this a forum argument), my FGR222 fas correctly recognized and related controls created. OK, but blinds now react in a crazy way.

Thank to the zwavejs2mqtt architecture, I can manage blinds in the control panel (and I've discovered that my blinds are **NOT** "inverted", them act exactly as set in control center. So, I've installed an home assistant instance on my second odroid, pointing to my zwavejs2mqtt server (the same domoticz does), so now I can see at the same time the zwavejs control panel, the domoticz instance and the home assistant one.

Well, I'm not an home assistant user (never seen before), so I use what is automatically set by his mqtt-ad, and can see that the zwavejs control panel, the home assistant control and the real blind status are synced. Domoticz, set as "blinds with positioning" (not inverted) no. It seems it applies some sort of logic that distorts the state of the controls. The slide is inverted, buttons act correctly and also the status text (open closed) and popups on buttons are correct (not the percentage, it's 100-x).

This strange behavior is relevant also in the group control, despite than before, it's simply inverted. If I push the green button (the hint popup tells "open"), blinds close. If I push the red button (the hint popup tells (close") blinds open.

[...]

PR #5260, opened 2022-06-14
Fixing crazy blinds on MQTT auto discover blinds types

CyrilB1531:
Fixing #5217 on MQTT auto discover blinds devices:
- Adding sTypeBlindsT21 as generic MQTT blinds
- Setting MQTT Autodiscover code to match generic blinds
- Changing WebServer to send good values to web site and good action to MQTT Auto discover
- Changing LightController and DashboardController to reflect changes

Tested with a Tasmota dual R3 as blind switch on Compact, Normal and Mobile Dashboard of web site

### DIFF
--- a/main/RFXtrx.h
+++ b/main/RFXtrx.h
@@ -8,11 +8,11 @@
 //
 
 /*
-                                                                   
+
 Copyright 2011-2022, RFXCOM
 
 ALL RIGHTS RESERVED. This code is owned by RFXCOM, and is protected under
-Netherlands Copyright Laws and Treaties and shall be subject to the 
+Netherlands Copyright Laws and Treaties and shall be subject to the
 exclusive jurisdiction of the Netherlands Courts. The information from this
 file may freely be used to create programs to exclusively interface with
 RFXCOM products only. Any other use or unauthorized reprint of this material
@@ -85,7 +85,7 @@ SDK version 9.22	Aug 18, 2018
 	Zemismart blinds added
 	Async port added
 	Firmware types added
-	Livolo 1-10 device changed 
+	Livolo 1-10 device changed
 
 SDK version 9.21	June 18, 2018
 	Fan LucciAir DC added
@@ -301,7 +301,7 @@ SDK version 6.07
 	CHIME structure and pTypeChime added for Byron SX Chime
 
 SDK version 6.06a
-    RFU4 changed to RSLenabled in IRESPONSE
+	RFU4 changed to RSLenabled in IRESPONSE
 
 SDK version 6.06
 	Lighting1 Energenie added
@@ -337,7 +337,7 @@ SDK version 5.00
 		cmdENABLEALL 0x04, cmdUNDEC 0x05
 		cmdDISX10 0x10   to    cmdDISFS20 0x1C
 	CM180i CURRENT_ENERGY - ELEC4 added
-	code for pTypeGAS and pTypeWATER changed (not yet used) 
+	code for pTypeGAS and pTypeWATER changed (not yet used)
 
 SDK version 4.36
 	security - #define sStatusIRbeam 0x8 added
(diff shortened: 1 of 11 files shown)


gizmocuz:
After a quick check, there is one problem ... **'#define sTypeBlindsT21 0x15	//Generic'** in RFXtrx.h, this file is from the manufactorer and will be overwritten by new updates.
Could you add a new blinds in 'RFXNames.h' ? (Maybe label it sTypeBlindsMQTT)

CyrilB1531:
Thanks for your feedback, 
If I add new blinds, I will add:
	STYPE_BlindsMQTT,
	STYPE_BlindsInvertedMQTT,
	STYPE_BlindsPercentageMQTT,
	STYPE_BlindsPercentageInvertedMQTT,
	STYPE_BlindsPercentageWithStopMQTT,
	STYPE_BlindsPercentageInvertedWithStopMQTT,

May be I can use Hardware HwdType instead. I check that

gizmocuz:
@CyrilB1531 , I do not think we need any additional blind types as you know the hardware is MQTT correct?
So obviously in the MQTT class you do not need them, and for the eventsystem you can check the hardware type
There should also be no difference in behaviour ?

main/RFXtrx.h
```diff
@@ -881,10 +881,9 @@ SDK version 4.9
 #define sTypeBlindsT18 0x12	//Cherubini
 #define sTypeBlindsT19 0x13	//Louvolite vertical
 #define sTypeBlindsT20 0x14	//Ozroll E-Trans
-#define sTypeBlindsT21 0x15	//Generic
 
-#define blinds_sOpen 0x0
-#define blinds_sClose 0x1
+#define blinds_sClose 0x0
```
gizmocuz: You can not modify anything in this file

CyrilB1531: Done, I've reverted my changes

gizmocuz:
Thank you for all the changes and time!!

One question, in the lightscontroller.js i noticed:

```
if(item.HardwareTypeVal != 125)
								xhtm += '\t      <td id="type">' + item.Type + ', ' + item.SubType + ', ' + item.SwitchType;
							else
								xhtm += '\t      <td id="type">' + item.Type + ', ' + item.SwitchType;

```

It is a bit scary to check against 125.... But I did not see this in the DashboardController.... Is there a reason to remove the display of the subtype ?

CyrilB1531:
I've just remove it, it was when MQTT AD Covers was set as blind Type.

gizmocuz:
@CyrilB1531 , thank you very much for this patch and work!! (And also to all testers!!)

CyrilB1531:
Thank you all

gizmocuz:
@CyrilB1531 , is it possible to have a look at this issue:

https://github.com/domoticz/domoticz/issues/5282

Maybe it's because of the changes in the web interface? Seems like adding a virtual blind (dummy hardware) is not working , and I suspect other (real) hardware as well.
Maybe we could keep the web interface the same (when there are no issues with it of course) and patch the code in MQTT AD ?

CyrilB1531:
I Gizmocuz,

Of course I can and I will check it


De : gizmocuz ***@***.***>
Envoyé : samedi 23 juillet 2022 08:07
Cc : CyrilB1531 ***@***.***>; Mention ***@***.***>
Objet : Re: [domoticz/domoticz] Fixing crazy blinds on MQTT auto discover blinds types (PR #5260)


@CyrilB1531<https://github.com/CyrilB1531> , is it possible to have a look at this issue:

#5282<https://github.com/domoticz/domoticz/issues/5282>

Maybe it's because of the changes in the web interface? Seems like adding a virtual blind (dummy hardware) is not working , and I suspect other (real) hardware as well.
Maybe we could keep the web interface the same (when there are no issues with it of course) and patch the code in MQTT AD ?

kiddigital:
@CyrilB1531 , would it be possible to maybe add some tests? See the test folder.

Even just writing the feature files, without actual implementation, would already be helpful so people can read the intended behavior (it acts as documentation at least).

If you need help or have questions, feel free to use our slack channels. Thx.

gizmocuz:
@CyrilB1531 , this PR is causing the light/dimmer slider on the dashboard to not update anymore when a new value is introduced
A quick reproduction is two add a dummy light/dimmer and open two browsers, one on the switches tab and one on the dashboard
Next set a level on the switches tab

If I check for both _slider and #slider then it works ... but is there a need to introduce a _slider ? all was #slider before and there was no need to check for existence of both

CyrilB1531:
@gizmocuz hello,
if, you've got more than one dimmer device the previous code creates devices with the same id="slider" and when you change the value of one of these devices, all slider change their values. That's why I've changed it. I'll check it to reproduce your issue and fix it then

gizmocuz:
@CyrilB1531 , thanks for checking. Yes you would expect id="123 slider" not "slider"

PR #5264, opened 2022-06-24
using OpenSSL 3.0.0

### LINKED ISSUE #5233: Domoticz doesn't start after ubuntu update to 22.04 (Due to OpenSSL3)
Since 22.04LTS ubuntu switches to newer libssl which causes domoticz startup failure.  
`error while loading shared libraries: libssl.so.1.1: cannot open shared object file: No such file or directory`

CyrilB1531:
Fixing #5233
But it needs some tests with Xiaomi gateway


kiddigital:
Hi @CyrilB1531 , thank you for the PR.

Our automated build for Windows does complete successfully but it depends on the OpenSLL 1.1 lib for Windows.

The Linux test build fails. It runs on a Ubuntu 20.04LTS environment which by default still has OpenSSL 1.1. But it seems to fail on the CMake OpenSSL library search complaining about the version parameter. Not sure (and unable to test at the moment) if that is a CMake problem or just simply saying it can not find version 3 of OpenSSL. But is stops the building process which it should not.

The building process should be building successfully for both OpenSSL 1.1 and 3. It will take a while before other Linux version (and for example the ARM based ones for the Raspberry Pi, etc.) have switch to using OpenSSL 3 as the new default.

Can you try to make it build under both Ubuntu 20.04 and 22.04 using their respective default OpenSSL libs?

gizmocuz:
I think we don't need to have changes in the CMake project file as it will automatically picks the current installed OpenSSL version installed on the system.
If a system has no OpenSSL 3.0 but 1.1, it will compile/work with 1.1, if 3.0 is installed it will pick that one
The other changes are needed as some functions are deprecated 

Maybe instead of using the _ex functions like EVP_CipherInit_ex we can use EVP_CipherInit so it will compile with the 1.1 library as well ?

gizmocuz:
Could you test the output of the original Xiaomi code against the new one ?
Maybe also rewrite the function to use

`	auto ctx = std::unique_ptr<EVP_CIPHER_CTX, decltype(&EVP_CIPHER_CTX_free)>(EVP_CIPHER_CTX_new(), EVP_CIPHER_CTX_free);
`
and 
```
EVP_CipherInit
EVP_CipherUpdate
EVP_CipherFinal

```
I am having a hard time testing this (output does not match), maybe also in the original code 'sizeof(plaintext)' was not correct and should have been 'strlen(plaintext)' ?

CyrilB1531:
I will check it thank you for your feed back

gizmocuz:
@CyrilB1531 , thanks for the changes.
Is it possible to undo all changes in 'msbuild/domoticz.vcxproj' and 'CMakeList.txt', I don't think we need any change in there

gizmocuz:
Thank you very much for this PR and your time!

acaonweb:
Ok, so what I have to do? Just wait😄?

gizmocuz:
@CyrilB1531 , I was a bit afraid of this.
Made a small patch but not sure if this is going to solve the issue.
Code seems valid.

CyrilB1531:
I think we should use libssl 3.0 while buildind domoticz to close this issue :(

gizmocuz:
@CyrilB1531 , the above issue mentioned by StKob was caused by a translation from 1.xx to 3.0 and was not related to compiling Domoticz with the 3.0 library

I do agree we need to switch very soon when all platforms supports it
For example the ARM build systems (running debian 10 (buster)) does not have 3.0, when I force the build system to use 3.0, everyone on buster will have a non working system

We might force to link it statically but this has caused issues in the past and also introduces more memory usage and filesize

This is what it is when libraries get's updated

You can always build Domoticz yourself or just use the docker image (i would recommend so)

gizmocuz:
https://packages.debian.org/source/bullseye/openssl

PR #5265, opened 2022-07-02
Fix rain forecast owm

lokonli:
Via the OWM hardware plugin devices can be created to indicate the rain forecast for the coming days.
However, this doesn't work correctly. There are two issues:

* Rain rate devices are being created. However, this results in unreliable data, since rain rate is transformed into daily rainfall  by Domoticz, giving incorrect results.
* In case the predicted rain fall is 0mm, the OWM data doesn't contain the rain forecast data at all. That means the device is not updated, which is incorrect, because it should become 0.

Both issues will be fixed with this PR.
* Instead of devices of type rain rate the type custom will be used
* Devices will be updated to 0 in case rain rate data is not provided by OWM

Tested on Debian 10.

Remark on update:
If the rain forecast devices already exist they will not be updated or recreated automatically by Domoticz and will have the wrong device type.
Users have to manually delete the devices first. After that, they will be created again by the hardware plugin.

### DIFF
--- a/hardware/OpenWeatherMap.cpp
+++ b/hardware/OpenWeatherMap.cpp
@@ -486,7 +486,7 @@ bool COpenWeatherMap::ProcessForecast(Json::Value &forecast, const std::string &
 	{
 		try
 		{
-			float rainmm = 0.00F;
+			float rainmm = 9999.00F;
 			float uvi = -999.9F;
 			float mintemp = -999.9F;
 			float maxtemp = -999.9F;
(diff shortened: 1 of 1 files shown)


kiddigital:
Hi @lokonli , thx for the PR..

Good point to update the rain fall forecast back to 0 when No rain fall forecast data is in the payload, which means that there is no (0) rain fall forecasted.

I think I agree with you that the current RainRateSensor is not the best as the _translation_ to daily rain fall makes it hard to use also when looking at the historical prediction data.

Not sure yet if using a custom sensor type would be the best. I do understand that it is totally fine for example when using it in Dashticz (which I am using as well :) ). And I am very much in favor for using the Domoticz OWM module in Dashticz which makes a seperate OWM module in Dashticz obsolete.

Down side for the Domoticz UI is that these custom sensor appear on the 'Other' tab and not on the 'Weather' tab where one should expect these forecasts.

Let me look if adding a different Rain sensor would be workable.

gizmocuz:
Maybe the code of the rainrate sensor has to be checked
Should not be that complicated

Rain Rate:
"A measurement of the intensity of rainfall. It gives the amount of rain that would fall over a time period if the rain rate was constant over that time period"

usually expressed by mm/h

Instead of working from the start of the day the amount of rain from the last hour has to be taken in account

kiddigital:
@lokonli , made some small adjustments. Let me know if they work for you... if so, we can merge this PR

lokonli:
I just tested, almost ok:
Sometimes pop>0, while no rain forecast is provided by OWM.
In that case the rain forecast should be set to 0, and not ignored.
(rainmm is initialized to 9999, and not changed. The effect is that the device will not be updated)

rain forecast info can indeed be obtained for instance with DzVents:
`
domoticz.utils.stringSplit(device.sValue,';',true)[2]   
`

kiddigital:
Ok great.

If PoP > 0 but no rain (snow) data, why set it to 0 when PoP says that there will be something?

Does setting rain to 0 give the impression that it will remain dry while PoP says there is a change?

Any clue why there is no rain and PoP > 0?

kiddigital:
What about the idea for the rainsensors to add an optional 'prediction' flag? This flag can be used to differ in the totals calculations.

When the prediction flag is set, then alternative total rain calculation can be done.

PR #5266, opened 2022-07-02
Fix XiaomiGateway cipher

mce35:
After recent changes in XiaomiGateway (commit [21f97e4](https://github.com/mce35/domoticz/commit/21f97e4adb11008f1e59fca710cfc9cd5e2a9b46) and [3da4f5e](https://github.com/mce35/domoticz/commit/3da4f5edcff05a1195844b744918bc888f3682b1)), it was not working anymore (plugin reports "invalid key" in the logs). From [here](https://www.openssl.org/docs/man1.0.2/man3/EVP_CipherFinal_ex.html), buffer passed to EVP_EncryptFinal() must be after data just encrypted to avoid overwriting it.

### DIFF
--- a/hardware/XiaomiGateway.cpp
+++ b/hardware/XiaomiGateway.cpp
@@ -1121,8 +1121,9 @@ std::string XiaomiGateway::GetGatewayKey()
 		Log(LOG_ERROR, "GetGatewayKey Cannot Perform EVP Cipher Update");
 		return std::string("");
 	}
-	int len = cipherSize;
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thank you for the patch! Yes you are correct, but I don't think this solves the issue.
I made some more changes and hope it works correctly now
Do you have a gateway to test against ?

gizmocuz:
Problem has been solved. Thank you for the PR

mce35:
> Problem has been solved. Thank you for the PR

Hello, thank you, it is working with your fix.

PR #5271, opened 2022-07-06
avoid generating 0 values update in short log

badzz:
P1 meter is updated every ShortLogInterval (5mins ) but if the device did not update its accumulated values whithin this period the diff is going to be 0.
Hence it generates graph like waves ,eg:
device updates 
100 110 120 after 6mins each time
it will end up in short log as
0 10 0 20 
this fix makes sure that the data is younger than 5mins to write a value in short log

### DIFF
--- a/main/SQLHelper.cpp
+++ b/main/SQLHelper.cpp
@@ -6688,6 +6688,10 @@ void CSQLHelper::UpdateMultiMeter()
 
 			if (difftime(now, checktime) >= SensorTimeOut * 60)
 				continue;
+
+			if (difftime(now, checktime) >= m_ShortLogInterval * 60)
+                               continue;
+
 			std::vector<std::string> splitresults;
 			StringSplit(sValue, ";", splitresults);
 


badzz:
it is to avoid to see this type of graph on P1 Meters when it is being updated in more than 5mins 
image

gizmocuz:
@badzz , thank you for this PR!

Some thoughts:
- As far as I am aware, the P1 meter outputs the energy every 10 seconds... ? So this might be a custom sensor ?
- How about you set the timeout for this sensor to 5 minutes ?
- the 'UpdateMultiMeter' function (or all shortlog functions) are not calculating anything it just puts the last received value in the tables. So it seems this is something that should be solved in the routine that generates the data for this shortlog graph maybe ?

I am currently making some internal changes to make this an option for all type of sensors.. not sure if that's a good idea
(Add only newly received values to the short log)
On the other hand, this might save quite some disk space (and database queries/performance) when decreasing the shortlog interval

gizmocuz:
Is it possible you could check this PR:

https://github.com/domoticz/domoticz/pull/5273

badzz:
> * As far as I am aware, the P1 meter outputs the energy every 10 seconds... ? So this might be a custom sensor ?

Yes it is a custom sensor updated via a python plugin

> * How about you set the timeout for this sensor to 5 minutes ?

ouch, I did not find a way to set a timeout per sensor . let me dig deepr

> * the 'UpdateMultiMeter' function (or all shortlog functions) are not calculating anything it just puts the last received value in the tables. So it seems this is something that should be solved in the routine that generates the data for this shortlog graph maybe ?

My understanding is that the device updates DeviceStatus table and UpdateMultiMeter is called every 5mins to fill the shortlog.
It does not "calculated" anything but it updates the table with a wrong value when the lastupdate is higher than 5mins

PS: Sorry I pushed this PR because I am doing a summer cleaning and found this fix that I did last year and forgot about it. I am a bit rusted on the understanding of the mechanism.

badzz:
> Is it possible you could check this PR:
> 
> #5273

that looks really good .

gizmocuz:
The device status table is updated when a new sensor reading is made. UpdateMultiMeter is called every 5 minutes indeed.
Cool, I will then merge PR 5273! ;)
Thank you for this PR, hope to see more PR's when needed 👍😀

badzz:
Thanks a lot . Have a great weekend

badzz:
#5273

PR #5273, opened 2022-07-08
Option to add only newly received values in the short log

gizmocuz:
Currently every 5 minutes the last received value is stored in the short-log history
This option allows the user to only add newly received values in the log


kiddigital:
@gizmocuz do the translation files updates be part of this PR? Maybe separate them?

gizmocuz:
Yes I'm afraid I already changed them, but this could be a PR that is going to be merged anyway, and the default behaviour is like it currently is

PR #5283, opened 2022-07-23
Testing UTF-8 proof trim functions

kiddigital:
This PR adds a few more tests to testing the trim helper functions.

The proposed changes of PR #5256 have been applied to the Helper functions. See [this](https://github.com/domoticz/domoticz/commit/7d4e8eddaa31118bfc531928b8c3654b4eb57cd2) commit.

But they also have been [reverted](https://github.com/domoticz/domoticz/commit/81ce5798b496b153a7c8ede8c8d81bdeffa60705) afterwards as it was suspected that they do not work correctly with UTF-8 string input.

Therefor some tests have been added to test with UTF-8 input as well.

And it works ok.

The `domoticztester` utility now has an extra option to show the output as binary (Hexadecimal representation).

For example running ` ./domoticztester -module helper -function stdstring_trim -input " Δ€ UTF-8°  " -binary`
will show
```
Domoticz_tester V2022.1 (build 14420) (c)2022-2022 KidDigital
Build Hash: 81ce5798b, Date: 2022-07-22 12:02:09

Domoticztester testing function stdstring_trim of Module helper with input: . Δ€ UTF-8°  .
Executing : stdstring_trim (helper) | Result : .Δ€ UTF-8°. (b'0xCE 0x94 0xE2 0x82 0xAC 0x20 0x55 0x54 0x46 0x2D 0x38 0xC2 0xB0')
```

### DIFF
--- a/main/Helper.cpp
+++ b/main/Helper.cpp
@@ -583,26 +583,12 @@ double distanceEarth(double lat1d, double lon1d, double lat2d, double lon2d)
 
 std::string &stdstring_ltrim(std::string &s)
 {
-	s.erase(s.begin(), std::find_if_not(s.begin(), s.end(), ::isspace));
-/*	while (!s.empty())
-	{
-		if (s[0] != ' ')
-			return s;
-		s = s.substr(1);
-	} */
-	return s;
+	return s.erase(0, s.find_first_not_of(' ' ));
 }
 
 std::string &stdstring_rtrim(std::string &s)
 {
-	s.erase(std::find_if_not(s.rbegin(), s.rend(), ::isspace).base(), s.end());
-/*	while (!s.empty())
-	{
-		if (s[s.size() - 1] != ' ')
-			return s;
-		s = s.substr(0, s.size() - 1);
-	} */
-	return s;
+	return s.erase(s.find_last_not_of(" ") + 1);
 }
 
 // trim from both ends
--- a/main/Helper.cpp
+++ b/main/Helper.cpp
@@ -588,7 +588,7 @@ std::string &stdstring_ltrim(std::string &s)
 
 std::string &stdstring_rtrim(std::string &s)
 {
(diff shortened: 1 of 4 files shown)


gizmocuz:
The problem is with the ltrim function when using a string containing "Â°C" (which we get from mqtt)
This is solved with the updated code, previous code in this PR still caused a crash

```
---------------------------
Microsoft Visual C++ Runtime Library
---------------------------
Debug Assertion Failed!

File: minkernel\crts\ucrt\src\appcrt\convert\isctype.cpp
Line: 36

Expression: c >= -1 && c <= 255
---------------------------

```

kiddigital:
So instead of specifying " " specifying ' ' takes care of the implicit conversion of type? (See here my lack of C/C++ knowledge) Shouldn't we make it explicit for clarify? Explicit casting maybe?

kiddigital:
Could it be OS dependent? (no clue why though)

When I try (with the 'old' code of the PR) to test the mentioned input on Linux, it works just fine:
```
./domoticztester -module helper -function stdstring_ltrim -input " Â°C " -binary
Domoticz_tester V2022.1 (build 14420) (c)2022-2022 KidDigital
Build Hash: 81ce5798b, Date: 2022-07-22 12:02:09

Domoticztester testing function stdstring_ltrim of Module helper with input: . Â°C .
Executing : stdstring_ltrim (helper) | Result : .Â°C . (b'0xC3 0x82 0xC2 0xB0 0x43 0x20')
```

And with the new code it also works fine.

So I think we have a winner... :)

gizmocuz:
No, with " " you can also specify multiple characters like " \t\r\n" all those four characters will then be removed from the beginning or end

It could be OS depended; I check on Windows via Visual Studio and also in debug mode which sometimes provides more information

With the current modifications I think we're safe on all platforms

PR #5291, opened 2022-08-01
Honor cancel button in DeleteDataPoint dialog

### LINKED ISSUE #5290: Value gets removed when cancelling the "remove value" dialogue.
Version: 2022.1 (build 14440)
Build Hash: b793752ee

The dialogue that confirms that a value should be removed will remove the value even when the answer is "cancel". 

How to reproduce: I open the "Log" page of a counter. I Shift+Click a value in the "Last Month" view and a dialogue opens asking "Are you sure to remove this value at ?: Date 2022-07-19 Value: 24.181". No matter whether I now press "Cancel" "X" or "Yes", the value always gets removed.

Expectation: When pressing "Cancel" or closing the dialogue using "X" the value should not be removed.

kiddigital:
This PR fixes #5290 .

The actual delete was still performed even when the user pressed Cancel on the confirmation dialog.

### DIFF
--- a/www/app/log/factories.js
+++ b/www/app/log/factories.js
@@ -84,18 +84,20 @@ define(['app'], function (app) {
                     if (result !== true) {
                         reject();
                     }
-
-                    domoticzApi
-                        .sendCommand('deletedatapoint', {
-                            idx: deviceIdx,
-                            date: dateString
-                        })
-                        .then(resolve)
-                        .catch(function () {
-                            HideNotify();
-                            ShowNotify($.t('Problem deleting data point!'), 2500, true);
-                            reject();
-                        });
+                    else
+                    {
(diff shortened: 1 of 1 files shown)


kiddigital:
Maybe someone with better JavaScript/Angular knowledge than me can also have a look and explain what the expected behavior is of the `reject()` call in this piece of code? Is it expected to stop execution at that point and return (which it currently does not, hence the _else_)

@michahagg maybe?

gizmocuz:
By issuing a 'reject()' the function continues so in RefreshingChart.js the 'self.$route.reload();' is performed.
By issuing a 'return' execution is stopped. There was another mistake in the function that is now also corrected.

Thank you for the PR!

PR #5293, opened 2022-08-02
Fix build domoticztester when NOT using included third party projets.

xbeaudouin:
This simple fix, allow user that doesn't include third party external libs to build correctly domoticz. Used for exemple on FreeBSD.


kiddigital:
Thank you!

Issue #4055, opened 2020-04-05
Thermostat setpoint cannot go to minus


Dylantje:
I hope some one can make the Thermostat setpoint is going to minus.
Now only possible to 0 and + value.
For my heatpump i need minus value.

Hope there is a option..

writeangle:
Is this for the purpose of setting the heating mode to a setpoint that's negative?

Dylantje:
Could be.
And also for the shift option to set to -1 to -5

Vincent835:
Hi, 
I need the same function in order to shift +/-5°C temperature target. 
Thank you

gizmocuz:
@Dylantje , I don't understand, your 'heat pump' is for heating, why on earth do you need to send a value from 0 degrees or lower ?

kiddigital:
> Hi, 
> 
> I need the same function in order to shift +/-5°C temperature target. 
> 
> Thank you 

Is this because the device does not want to receive an absolute value (positive and higher than say 0 or more) but expects a relative value to deviate from a given 'middle' value? What would be a logical range? Somewhere between -10 to +10?

Which Domoticz module do you use to control the device?

Vincent835:
> Is this because the device does not want to receive an absolute value (positive and higher than say 0 or more) but expects a relative value to deviate from a given 'middle' value? What would be a logical range? Somewhere between -10 to +10?

For my heat pump, the room temperature setpoint is absolute.
But the water temperature is calculated automatically by the heat pump and can be adjusted by + or - 5 ° C.

> Which Domoticz module do you use to control the device?

I am writing a python plugin to monitor and configure my heat pump.
It is a DAIKIN ALTHERMA HT with an RTD-W modbus module.

The solution I found is a 11 level selector from -5 to + 5 ° C.

gizmocuz:
Could you guys please stay on topic. Which is:

"Thermostat setpoint cannot go to minus"

pipiche38:
I would tend to agree that having a Setpoint to - makes some sense. We might not have the use case right now, but why not having a connected fridge or icer where we have to setpoint to -25° !

gizmocuz:
@pipiche38 , I am not aware of such device.
Of course you can convince me by supplying a link to a API that uses set points < 0 degrees

vandermark1977:
I agree with the heatpump case: We shift the water outlet temperature from -5 to +5. It would be very useful if you can simply click down from zero te negative values in the thermostat.setpoint device.

[...]

PR #5297, opened 2022-08-03
JS UI Thermostat set point onclick fix.

ajarzyn:
Only bigtext was updated. Updating onclick function parameter.

On update_device only bigtext is updated.
`ShowSetpointPopup(event, 40, false, 1.0)` data value stayed the same thus after opening set point again value from page load was loaded.

```
<td id="bigtext">1.0° C</td>
<td id="img"><img src="images/override.png" class="lcursor" onclick="ShowSetpointPopup(event, 40, false, 1.0);" width="48" height="48"></td>
```

Change only for thermostat device.

PS.
I was trying to use following line:
UtilityController.js:814: `'</td>\n';` in place of regex to update img but unfortunately in `RefreshItem = function (item) {` `item` does not contain `Image` member thus regex.

It would be good to have initial definition of img field (line814) in one place not to double the code, but using regex here isn't (in my opinion) most efficient solution.

Could someone look into why there is not `Image` member?

### DIFF
--- a/www/app/UtilityController.js
+++ b/www/app/UtilityController.js
@@ -666,7 +666,7 @@ define(['app', 'livesocket'], function (app) {
 							}
 							else if (item.Type == "Thermostat") {
 								xhtm += item.Data + '\u00B0 ' + $scope.config.TempSign;
-                                $(id + " #img").attr('onclick', 'ShowSetpointPopup(event, ' + item.idx + ', ' + item.Protected + ', ' + item.Data + ')');
+								$(id + " #img").attr('onclick', 'ShowSetpointPopup(event, ' + item.idx + ', ' + item.Protected + ', ' + item.Data + ')');
 							}
 							else if (item.SubType == "Waterflow") {
 								xhtm += item.Data;
--- a/www/app/UtilityController.js
+++ b/www/app/UtilityController.js
@@ -433,6 +433,8 @@ define(['app', 'livesocket'], function (app) {
 				else if ((item.Type == "Thermostat") && (item.SubType == "SetPoint")) {
 					status = "";
 					bigtext = item.Data + '\u00B0 ' + $scope.config.TempSign;
+					const regex = /ShowSetpointPopup\(event, \d*, \w*, (\d+\.?)*\d+\);/gm;
+					img = $(id + " #img").html().replace(regex, 'ShowSetpointPopup(event, ' + item.idx + ', ' + item.Protected + ', ' + item.Data + ');')
 				}
 				else if (item.Type == "Radiator 1") {
 					status = item.Data + '\u00B0 ' + $scope.config.TempSign;
--- a/www/app/UtilityController.js
+++ b/www/app/UtilityController.js
@@ -433,8 +433,7 @@ define(['app', 'livesocket'], function (app) {
 				else if ((item.Type == "Thermostat") && (item.SubType == "SetPoint")) {
 					status = "";
 					bigtext = item.Data + '\u00B0 ' + $scope.config.TempSign;
-					const regex = /ShowSetpointPopup\(event, \d*, \w*, (\d+\.?)*\d+\);/gm;
-					img = $(id + " #img").html().replace(regex, 'ShowSetpointPopup(event, ' + item.idx + ', ' + item.Protected + ', ' + item.Data + ');')
+					$(id + " #img").attr('onclick', 'ShowSetpointPopup(event, ' + item.idx + ', ' + item.Protected + ', ' + item.Data + ')');
 				}
 				else if (item.Type == "Radiator 1") {
 					status = item.Data + '\u00B0 ' + $scope.config.TempSign;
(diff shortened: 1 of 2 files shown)


gizmocuz:
The "Image" field is only set when the user chooses a custom icon over the default one.
If none is provided, the default icon is set via the JS script.
In this case, there is also no override48_On.png as there is for the custom icons (s they have a on/off and small icon)

Happy to merge your PR if you don't see issues doing so

ajarzyn:
> Happy to merge your PR if you don't see issues doing so

I'm sorry but I do not understand this sentence. Is it politeness, good manners or real question?
I mean I would never made a PR if I wouldn't like it to be merged :)

About Images I do understand that it makes less data to be throw in the requests and that's fine.
With `data['Image']` it would be also 'no go' without refactoring styling the controls.
I've asked about it only because I wasn't sure that we should use regex/html parsing/string operations, but if you feel this is proper way to go please feel free and merge.

Best to you @gizmocuz and thank you for you good work.

www/app/UtilityController.js
```diff
@@ -433,6 +433,8 @@ define(['app', 'livesocket'], function (app) {
 				else if ((item.Type == "Thermostat") && (item.SubType == "SetPoint")) {
 					status = "";
 					bigtext = item.Data + '\u00B0 ' + $scope.config.TempSign;
+					const regex = /ShowSetpointPopup\(event, \d*, \w*, (\d+\.?)*\d+\);/gm;
```
ajarzyn: One last thought.
What if signature for ShowSetpointPopup will change?
Shouldn't we make it depended on the JS code we have?

www/app/UtilityController.js
```diff
@@ -433,6 +433,8 @@ define(['app', 'livesocket'], function (app) {
...
+					img = $(id + " #img").html().replace(regex, 'ShowSetpointPopup(event, ' + item.idx + ', ' + item.Protected + ', ' + item.Data + ');')
```
ajarzyn: Additional 'last thought'
wouldn't it be better to update only attribute for onClick?
Something like this (not tested yet):
`$(id + " #img").attr('ShowSetpointPopup(event, ' + item.idx + ', ' + item.Protected + ', ' + item.Data + ')')`

This way we do not need to use regex and we update this parameter directly under this if, not waiting for the later if with `img != ''`

Problem with doubling function signature will still exist.

ajarzyn (commented):
Please check additional comments.

gizmocuz:
The reason for asking is because you seem to have some doubts about using a regex. It did work nicely and my JS regex experience is limited.
I think the double function will still exist and updating only the attribute might seems better
Both work, I leave this up to you 😉

ajarzyn:
Please merge it. If I will find time (as I have some more debugging to do) related to Thermostat python plugin NameType. 
I will create another PR with only onclick attr set - as I think this is 'better' approach.

gizmocuz:
Thanks I will, just two more quick questions
Does the regex now have to be in quotes '<regex>', i always put strings in quotes
Do you also perform this action on the dashboard?

www/app/UtilityController.js
ajarzyn: Changed to attribute handling.

www/app/UtilityController.js
ajarzyn: > Do you also perform this action on the dashboard?

This proves my point, now I know about 3 (4?) different places where this is set.
- initialization of the web (dashboard)
- `refreshItem` (dashboard)
- During initialization of the web (Utils)
- `refreshItem` (Utils)

I feel like it should be wrapped into a class or some structure.
I would like to propose another PR, to parametrize this set point (I've already rose this in different [thread](https://github.com/domoticz/domoticz/issues/4055#issuecomment-993254561))

Step/limits would be useful.

ajarzyn:
ad 1. No it's just the language notation [stack answer](https://stackoverflow.com/questions/31491908/why-are-regular-expression-strings-not-encapsulated-in-quotes-in-javascript/31491967#31491967)
ad 2. No I haven't but I created another commit and hopefully it will be all good now: [5d706f4](https://github.com/domoticz/domoticz/pull/5297/commits/5d706f451f87c6afed31362e9877725bf2f65541) and corrected whitespaces in the next one: [04283e6](https://github.com/domoticz/domoticz/pull/5297/commits/04283e63a79d177c7b498a237891bf7457486138)

Please check my comment in the conversation.

ajarzyn:
I wanted to write that I tested it (it's late... almost 2AM), so I did one more check and it appeared that I've edited wrong line of thermostat code.

This time [23a79a9](https://github.com/domoticz/domoticz/pull/5297/commits/23a79a91e8514966ee1d1d8327523d35a1de0626) solution is fully tested, for Utilities and Dashboad.

Sorry for the mess in commits.

gizmocuz:
Thank you very much for the PR and your time! Looks great !

PR #5299, opened 2022-08-05
Build a Linux Beta artifact

kiddigital:
PR to extent the development github action pipeline to not only check if building Domoticz still works, but also perform the automated tests. And when everything is successful, a github artifact is created that can be downloaded afterwards (retention of 7 days) which contains the Linux x86_64 build for the specific version.

Also, both the Boost Library and CMake are updated to more recent versions.

### DIFF
--- a/.github/workflows/development.yml
+++ b/.github/workflows/development.yml
@@ -29,14 +29,26 @@ jobs:
           fetch-depth: 0
           submodules: true
 
+      # Prepare environment
+      - name: prepare environment
+        run: |
+          sudo timedatectl set-timezone Europe/Amsterdam
+          buildnr=$(git rev-list HEAD --count)
+          echo "BUILDNR=$(($buildnr + 2107))" >> $GITHUB_ENV
+
       #- name: Make an error (Enable debug)
       #  run: ./not-exist-file.sh it bloke build
 
       # install dependencies
       - name: dependencies
         run: |
-          sudo apt-get update && sudo apt-get install
+          sudo apt-get update && sudo apt-get --autoremove upgrade
           sudo apt-get install make gcc g++ libssl-dev git libcurl4-gnutls-dev libusb-dev python3-dev zlib1g-dev libcereal-dev liblua5.3-dev uthash-dev
+          sudo apt-get install python3-pytest python3-pytest-bdd
+          sudo apt-get install lua5.3 luarocks
+          sudo luarocks install busted
+          sudo luarocks install luacov
+      # Disabled March 2022 - sudo luarocks install lodash
 
       # get CMake
       - name: cmake-compile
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thanks for the PR looking good.

Does the SSH section need to be commented out on every line?
This because there is already a check to see if it needs to be enabled

kiddigital:
Commenting out all lines makes the step invisible, so it does not even appear as a skipped step.

With the 'false' the step does appear as one of the many steps, but is just not executed further.

I left it in there, but I think it can be removed totally. Not sure if/why we would use it now or in the future?

Is it an option to use the build artifact and upload it (ftp step) to the Domoticz server for download? That way the Linux build is also directly available like the Windows Beta's.

gizmocuz:
The SSH interface is sometimes needed when there are issues to solve and you want to test them on the build server first.

kiddigital:
> The SSH interface is sometimes needed when there are issues to solve and you want to test them on the build server first.
> 
> 

Understandable when you have a private build server. But on Github the builders are temporary (container) machines that are destroyed once the build has ended, either ok or in error state.

gizmocuz:
No works here as well

kiddigital:
> No works here as well

It's back in case we ever need it

gizmocuz:
I think changing

`sudo apt-get update && sudo apt-get --autoremove upgrade`

to

`sudo apt-get update`

Will have some performance increase as we only want to 'update' the package list because we are installing some packages, not because we need to upgrade the complete build enviroment

gizmocuz:
I am not sure if sftp is going to work with version 4.0

https://github.com/op-de-kop/op-de-kop/commit/2ee760913122c4d562070cf0925868fc8eca01f4

maybe use another action for SFTP like

https://github.com/marketplace/actions/sftp-deploy

Maybe have a peek at the windows appveyor script to see if we can use the same enviroments

  host: $(FTP_HOST)
  username: $(FTP_USER)
  password: $(FTP_PASSWORD)

I still don't think this has to be automated, but sure it's possible, but then we also need to look at the arm 32/64 builds

the current build server does the job ok

kiddigital:
> I think changing
> 
> `sudo apt-get update && sudo apt-get --autoremove upgrade`
> 
> to
> 
> `sudo apt-get update`
> 
> Will have some performance increase as we only want to 'update' the package list because we are installing some packages, not because we need to upgrade the complete build environment

Agree that it can save a little time, but in practice it will only save a few seconds as the whole 'dependencies' step takes less than a minute. Not sure who would benefit of this small reduction (1 percent of total?) in building time and I see a benefit in keeping it.

As with each run a new virtual machines/docker instance is created from a basis Ubuntu install, the image is not up-to-date. In case any up-to-date tools and libraries would break the build process, I think that it is a good thing to be aware of that early and it _only_ breaks building a Beta release.

Also I think that these Beta releases are often used on up-to-date images (and updated Docker images), so finding using during build is preferable I think over finding these issues during usage.

Agree?

kiddigital:
> I am not sure if sftp is going to work with version 4.0
> 
> [op-de-kop/op-de-kop@2ee7609](https://github.com/op-de-kop/op-de-kop/commit/2ee760913122c4d562070cf0925868fc8eca01f4)
> 
> maybe use another action for SFTP like
> 
> https://github.com/marketplace/actions/sftp-deploy
> 
> Maybe have a peek at the windows appveyor script to see if we can use the same enviroments
> 
> `host: $(FTP_HOST) username: $(FTP_USER) password: $(FTP_PASSWORD)`
> 
Hm.. ok... haven't been able to actually try it out as I don't have access to certain items.. but I will have a look.

> I still don't think this has to be automated, but sure it's possible, but then we also need to look at the arm 32/64 builds
> 
For the Windows builds it is automated, why not apply that for the Linux builds? And I think the Linux builds are the most used Beta's (both locally as in a container) so automated builds makes the latest changes available asap without depending on manual intervention (which could take time for many very good reasons). Automated builds can increase the speed of testing/verifying new changes and solve issues quicker.

Maybe we can make use of Docker `buildx` to even do cross-platform compilation making it possible to automate these platform builds as well in the future (but small steps, first things first).

> the current build server does the job ok
> 
True. A benefit of switching to using GitHub actions, is that it removes the need to maintain a build server for building the Linux builds (and maybe even ARM builds). We can just rely on _Infrastructure-as-Code_.

kiddigital:
Removed the FTP upload step as it is better/safer to use the [GitHub Action WebHooks](https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#workflow_run) and `workflow_run` event to retrieve the artifact and make it available on other channels for download as well.

PR #5300, opened 2022-08-06
Always setting previous dimmer level

gizmocuz:


CreasolTech:
Hi gizmocuz, 
I've just tried to 
gh repo clone domoticz/domoticz
gh pr checkout 5300
make
cp domoticz /home/pi/domoticz 
to replace the exec file in my domoticz working dir.
The problem was not solved.
Please note that dimmer level is managed perfectly by Domoticz.
The only problem is in the web UI, because 
* I set dimmer level to 80% , for example
* go to another panel, then return to switch panel, and see dimmer level to 100%, but a lua script that get the dimmer level, get the right level 80.
So, from my point of view, the only problem is in the UI: 
please check the video at https://youtu.be/I1RasA4DJ4Q (sorry for my bad english).
Thanks a lot for your great effort.

gizmocuz:
Thank you for testing! Is it possible to also test with the latest beta version?

CreasolTech:
Just load the latest beta version, then:
1. I change the dimmers value : slide stays to the right value
2. change domoticz room or panel
3. return to the original room/panel, but sliders are all full scale. The stupid LUA script I written to display in the log the current slider value always print the good value.
Thanks a lot for your effort.

gizmocuz:
Just to be sure, did you clear your browser cache as well? If you add a dummy switch (dimmer) does the slider work correctly?

CreasolTech:
Sorry for the late reply, but I've make several tests.
If I delete old virtual dimmer switches and create it again, now it works good. If I update its value from a LUA script, it works perfectly.

If I removed a dimmer that was created by a python plugin, restart domoticz (or restart only the plugin), the dimmer slider go to 100% even if the right value is lower.
Enabling a LUA script and updates the dimmer does not change: dimmer sliders go to 100% but the real value is always correct.

It's strange that virtual devices, after being removed and created, works good, while a real dimmer device managed by a python plugin displays bad.

gizmocuz:
@CreasolTech , thank you for testing. I think this PR is deprecated now and it should work correctly in the latest beta version

CreasolTech:
immagine
Build 14495 solved the issue, for me!
Very good work!!!
Thanks a lot!

gizmocuz:
Thanks for the feedback, closing this one!

Issue #5320, opened 2022-08-28
Feature: Parsing python plugin descriptions.


ajarzyn:
Hello all,

I would like to propose change for parsing python plugin description.

Right now all parsing is done on tha c++ side and it's static parse of xml document.
I wanted to have possibility to get dynamically generated description for plugins thus following POC:

Please check the following commit:
https://github.com/ajarzyn/domoticz/commit/99af100cb0f10368e869007f43bf8bbab2047868

I did not fully grasp on how to correctly incorporate importing python modules in the Domoticz, so I've only copied the way of how the python plugins are imported.

Followed by calling function object from the module:

```
PyObject* myFunction = PyObject_GetAttrString(m_PyModule, (char*)"get_plugin_description");
PyObject* myResult = PyObject_CallObject(myFunction, nullptr);
sXML = std::string(PyUnicode_AsUTF8(myResult)) + '\n';
```

I would appreciate some guidelines regarding multi thread import of python module, and if possible info of how python function should be called.

The commit is just a raw POC, with working solution.

To make it work plugin need a function with proper signature ex:

```
author = "ajarzyn"
version = "0.0.6"
def create_simple_list():
    return [a for a in range(0, 10)]

def get_plugin_description():
    return f"""
<plugin key="BasePlug" name="Basic Python Plugin Example" author="{author}" version="{version}" wikilink="http://www.domoticz.com/wiki/plugins/plugin.html" externallink="https://www.google.com/">
    <description>
        <h2>Plugin Title</h2><br/>
        {create_simple_list()}
        Overview...
        <h3>Features</h3>
        <ul style="list-style-type:square">
            <li>Feature one...</li>
            <li>Feature two...</li>
        </ul>
        <h3>Devices</h3>
        <ul style="list-style-type:square">
            <li>Device Type - What it does...</li>
        </ul>
        <h3>Configuration</h3>
        Configuration options...
    </description>
    <params>
        <param field="Mode6" label="Debug" width="150px">
            <options>
                <option label="None" value="0"  default="true" />
                <option label="Python Only" value="2"/>
                <option label="Basic Debugging" value="62"/>
                <option label="Basic+Messages" value="126"/>
                <option label="Queue" value="128"/>
                <option label="Connections Only" value="16"/>
                <option label="Connections+Queue" value="144"/>
                <option label="All" value="-1"/>
            </options>
        </param>
    </params>
</plugin>
"""
```

Thinks to consider:

[...]

PR #5323, opened 2022-08-29
Python plugin description parsing.

ajarzyn:
POC for parsing description for python plugins.
get_plugin_description recognized as parsing indicator.

Requested in #5320 by @kiddigital

### DIFF
--- a/hardware/plugins/PluginManager.cpp
+++ b/hardware/plugins/PluginManager.cpp
@@ -53,6 +53,7 @@
 
 extern std::string szUserDataFolder;
 extern std::string szPyVersion;
+extern MainWorker m_mainworker;
 
 #define GETSTATE(m) ((struct module_state*)PyModule_GetState(m))
 
@@ -90,9 +91,6 @@ namespace Plugins {
 			_log.Log(LOG_STATUS, "PluginSystem: Failed dynamic library load, install the latest libpython3.x library that is available for your platform.");
 			return false;
 		}
-
-		// Pull UI elements from plugins and create manifest map in memory
-		BuildManifest();
 
 		m_thread = std::make_shared<std::thread>([this] { Do_Work(); });
 		SetThreadName(m_thread->native_handle(), "PluginMgr");
@@ -152,6 +150,9 @@ namespace Plugins {
 			}
 
 			m_InitialPythonThread = PyEval_SaveThread();
+
+			// Pull UI elements from plugins and create manifest map in memory
+			BuildManifest();
 
 			m_bEnabled = true;
 			_log.Log(LOG_STATUS, "PluginSystem: Started, Python version '%s', %d plugin definitions loaded.", sVersion.c_str(), (int)m_PluginXml.size());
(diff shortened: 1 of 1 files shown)


ajarzyn:
Hello all,

I would like to propose change for parsing python plugin description.

Right now all parsing is done on tha c++ side and it's static parse of xml document.
I wanted to have possibility to get dynamically generated description for plugins thus following POC:

Please check the following commit:
https://github.com/ajarzyn/domoticz/commit/99af100cb0f10368e869007f43bf8bbab2047868

I did not fully grasp on how to correctly incorporate importing python modules in the Domoticz, so I've only copied the way of how the python plugins are imported.

Followed by calling function object from the module:

```c++
PyObject* myFunction = PyObject_GetAttrString(m_PyModule, (char*)"get_plugin_description");
PyObject* myResult = PyObject_CallObject(myFunction, nullptr);
sXML = std::string(PyUnicode_AsUTF8(myResult)) + '\n';
```

I would appreciate some guidelines regarding multi thread import of python module, and if possible info of how python function should be called.

The commit is just a raw POC, with working solution.

To make it work plugin need a function with proper signature ex:

```python
author = "ajarzyn"
version = "0.0.6"
def create_simple_list():
    return [a for a in range(0, 10)]

def get_plugin_description():
    return f"""
<plugin key="BasePlug" name="Basic Python Plugin Example" author="{author}" version="{version}" wikilink="http://www.domoticz.com/wiki/plugins/plugin.html" externallink="https://www.google.com/">
    <description>
        <h2>Plugin Title</h2><br/>
        {create_simple_list()}
        Overview...
        <h3>Features</h3>
        <ul style="list-style-type:square">
            <li>Feature one...</li>
            <li>Feature two...</li>
        </ul>
        <h3>Devices</h3>
        <ul style="list-style-type:square">
            <li>Device Type - What it does...</li>
        </ul>
        <h3>Configuration</h3>
        Configuration options...
    </description>
    <params>
        <param field="Mode6" label="Debug" width="150px">
            <options>
                <option label="None" value="0"  default="true" />
                <option label="Python Only" value="2"/>
                <option label="Basic Debugging" value="62"/>
                <option label="Basic+Messages" value="126"/>
                <option label="Queue" value="128"/>
                <option label="Connections Only" value="16"/>
                <option label="Connections+Queue" value="144"/>
                <option label="All" value="-1"/>
            </options>
        </param>
    </params>
</plugin>
"""
```


Thinks to consider:

- backward compability (if any) - meaning: dynamically generated description won't be able to be parsed by the old method thus, plugins using new method wouldn't be able to be run at older Domoticz versions.
- validation of generated xml data?

Why do I propose this change?
In https://github.com/ycahome/pp-manager

to add new plugin to the list you need to:

> In order to add your plugins to PP-Manager, make a "pull request" with your lines added:
> 
>     Increase version at "...author="ycahome" version="1.5.xx"..." in Line 12
> 
>     Increase version at "...Python Plugin Manager v.1.5.xx..." in Line 14
> 
>     Under 'label="Domoticz Plugin"' in alphabetical order by 'label' and
> 
>     Under 'self.plugindata' in alphabetical order by 'pluginText'

and instead of repeting (error prone) data insertions, it would be enought to change version, and add plugin data to the list, and plugin would generate sorted list by itself. (Commit made on different machine, I will post in next day or two with link to the example implementation).

Comments, help, opinions, most appreciated.

ajarzyn:
I would just add if that's the only way to import the module then next step would be to create a method or some kind of helper that will tackle all this functionality in one place.

ajarzyn:
Example dynamic description in pp-manager: https://github.com/ycahome/pp-manager/pull/42

dnpwwo:
I would like to see a clear statement of the problem that this is trying to solve.  Without that it is hard to comment on the solution.

At a high level though, if a change is made to move away from the plugin definition being a chunk of XML in a file to something more accessible in Python then this does not go far enough.  Why wouldn't the sections of the definition be exposed as attributes of the Plugin?

The other thing I would say is that the impact (and potential side effects) will need to be thought through.  When a Python module is imported it is also executed (this is a Python thing not a Domoticz one) which means any 'import's at the plugin level would also be imported (and executed) which can take a while.

ajarzyn:
> Example dynamic description in pp-manager: [ycahome/pp-manager#42](https://github.com/ycahome/pp-manager/pull/42)

@dnpwwo have you looked into this PR for pp-manager? There I've shown why I would like to introduce dynamic parsing of the description.

> move away from the plugin definition being a chunk of XML

I'm not sure I wanted to introduce such radical change. As it would break backward compatibility. I was thinking to make it an option to be used with plugin system. That's the reason I've left old implementation for xml parsing.
If `def generate_python_plugin_description():` won't be found before `<plugin ...` then plugin won't be imported, and current method of xml parsing would be used.

> Why wouldn't the sections of the definition be exposed as attributes of the Plugin?

I'm not sure if I understood you correctly. You suggest that description is a part of `class BasePlugin`?
I wasn't planning to make such a big change, as I thought it would be good to keep backward compatibility.
Additionally you pointed out that we will run the script (I forgot that in the plugin the Plugin Object is initialized)
```python
global _plugin
_plugin = BasePlugin()
```

For this I think of 2 solutions:
1. Extract description to different file ie. `description.py` then import it from `plugin.py` if needed, then import only description module.
2. move `_plugin - BasePlugin()` under `if __name__ == '__main__':` and run python plugins as separate processes initializing separate interpreters, but I expect that there would be need of interprocess comunication to be implemented for that, which could be to much.

pipiche38:
I'm not sure I followed, but the description.py will be executed at the start of plugin.py while the xml parsing is doing at the Domoticz level in order to publish the Hardware form.

So I'm not sure what is the purpose and how a backward compatibility can be save (old plugin format continue to work correctly on new Domoticz version)

I wonder if it won't be eassier to de-couple the xml from the plugin.py and so you can generate the xml as you want .
Side effect, I understand that Domoticz needs to be restarted as soon as we change the XML, that needs to be taken in consideration

ajarzyn:
```c++
if (!bFound && (line.find("<plugin") != std::string::npos))
		bFound = true;
	if (bFound)
		sXML += line + '\n';
	if (line.find("</plugin>") != std::string::npos)
		break;
}
readFile.close();
m_PluginXml.insert(std::pair<std::string, std::string>(plugin_Dir, sXML));
```
This part of the file should be left unchanged. So if the plugin would not implement dynamic description parsing, old method would be used. (And this is why I didn't wanted to extract description anywhere, because it will break backward compatibility).


My idea with `description.py` is rather simple, instead of looking for the special function in the code we could check for `import description` and if so, import description file **only**. But it would be impossible to use plugins using this system on the older versions of Domoticz.

I think description.py would be like de-coupling it from the `plugin.py` file. The real question is if it would be really needed to import it in `plugin.py` (I suppose it wouldn't be necessary) so then before parsing file looking for proper string, we could just check if there's proper `description.py` file in the plugin, and decide proper parsing method.


I'm sorry if all of that seems a bit chaotic but that's why I really wanted to consult it with others.

gizmocuz:
I'm closing this PR as it's not going anywhere and current system seems to work

PR #5333, opened 2022-09-07
Added support for EcoDevices RT2 with Firmware 3.00.00

xbeaudouin:
EcoDevices RT2 has a new firmware.
This trivial patch allow Domoticz to support it.


kiddigital:
Thank you!

PR #5335, opened 2022-09-09
Update Sterbox.cpp

houk001:
New Movement sensors require quicker read of sterbox devices

### DIFF
--- a/hardware/Sterbox.cpp
+++ b/hardware/Sterbox.cpp
@@ -11,7 +11,7 @@
 
 //http://sterbox.eu/index.php/sterbox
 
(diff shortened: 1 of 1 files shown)


kiddigital:
Hi @houk001 , how about the old ones? If the interval is shortened, does it affect the older hardware?

Otherwise this change might break functionality for existing users of older versions of the hardware.

houk001:
Old one will work without problem.  Tested by myself. Even work better. I have 8 devices and 4 seconds also work without problem

kiddigital:
Ok, great. Thx

PR #5337, opened 2022-09-12
Fix completly support for EcoDevices RT2 with firmware 3.00.xx or more

xbeaudouin:
EcoDevices RT2 with firmware 3.00.xx has new XML values for teleinformation and push the kWh data with '.' for some kWh that is mis interpreted by domoticz core.
I dunno if the RT2 with 2.00.xx has same issue or not, so the reason of testing about the major version of the RT2 version.

### DIFF
--- a/hardware/EcoDevices.cpp
+++ b/hardware/EcoDevices.cpp
@@ -8,11 +8,12 @@ http://gce-electronics.com/en/nos-produits/409-module-teleinfo-eco-devices.html
 Detailed information on the API can be found at
 http://www.touteladomotique.com/index.php?option=com_content&id=985:premiers-pas-avec-leco-devices-sur-la-route-de-la-maitrise-de-lenergie&Itemid=89#.WKcK0zi3ik5
 
-Version 3.3
+Version 3.4
 Author Blaise Thauvin
 
 Version history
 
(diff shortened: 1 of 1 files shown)


kiddigital:
Thx @xbeaudouin 

Maybe you can change the ifdebug compile time debugging by using `Debug(DEBUG_HARDWARE,"%s", message.c_str());`
as runtime debugging?

Makes it possible for others to gather debug details without the need to compile with debug first (just use debug parameters when starting Domoticz)

xbeaudouin:
@kiddigital Well I just patched the current code. 
Is there any README about Debug() stuff?

kiddigital:
> @kiddigital Well I just patched the current code. 
> 
> Is there any README about Debug() stuff?

Not specific that I know off, but look at the Logger class for details. It works in the same manner as the `Log()` function.

See also the domoticz commandline settings on how to activate the different debug levels.

xbeaudouin:
This should be done :D

kiddigital:
Nice.. one final _nitpicking_ :) you can remove the text 'DEBUG: ' at the beginning of each debug message as the Debug function will do that as well :)

xbeaudouin:
@kiddigital  Should be done :D

kiddigital:
Great and thx!

PR #5343, opened 2022-09-23
Fix protocol not set for ICMP and update Pinger plugin example

Ban-aan:
Hi,

It bugged me that the Pinger plugin did not work, so I did a little bit of investigating. 

It seems like the other Plugin Protocols set the 'ConnectionProtocol' on connect, but ICMP does not connect. 
I added the 'ConnectionProtocol' call just before creating 'CPluginTransportICMP', but I'm not sure if this is the right location. 

I have a few points before merging this PR:
- Is the 'ConnectionProtocol' call on the right position? 

- While debugging, after a while, I get the following Exception. In Release mode it runs without any crashes (tested multiple hours).
Domoticz-debug-crash

- I changed the 'Py_BuildValue' from `y#` to `s` because of the following Exception. I'm not sure if this is correct. 
Domoticz-debug-crash2

- Part of the Pinger.py update I took from [here](https://github.com/dnpwwo/Domoticz-Pinger-Plugin/blob/master/plugin.py). My addition is the change of `self.icmpConn.Name` to `self.icmpConn.Address`.

### DIFF
--- a/hardware/plugins/PluginProtocols.cpp
+++ b/hardware/plugins/PluginProtocols.cpp
@@ -1188,7 +1188,7 @@ namespace Plugins {
 
 		std::string		sData(Message->m_Buffer.begin(), Message->m_Buffer.end());
 		sData = sData.substr(iDataOffset, iTotalData - iDataOffset);
-		pObj = Py_BuildValue("y#", sData.c_str(), sData.length());
+		pObj = Py_BuildValue("s", sData.c_str(), sData.length());
 		if (PyDict_SetItemString(pDataDict, "Data", pObj) == -1)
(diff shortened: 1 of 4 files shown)


dnpwwo:
Couple of things:

1. "#y" has been problematic on recent Python versions and has been removed almost completely. This one was left because I can't really test on ICMP on Linux. Happy for it to be changed.  I've been converting the code to use the PyNewRef string constructor (see line 126 of PluginProtocols) but that's not crucial.
2. ConnectionListen is not the place to set a connection's protocol, none of the other transports have it set there and we need to be consistent.  If you are sure that it is not being set then you can set it using 				pConnection->pPlugin->MessagePlugin(new ProtocolDirective(pConnection)); in PluginTransports.cpp perhaps in CPluginTransportICMP::handleListen.

Ban-aan:
Applied your comments, thanks!

gizmocuz:

PR #5352, opened 2022-10-01
P1Meter: catch telegram line parsing errors

lokonli:
With my new energy meter I'm receiving a lot of P1 crc errors.
Unfortunately this also results in a Domoticz crash within 1-2 minutes.

This is caused by the stof (string to float) functions in P1MeterBase::MatchLine() throwing an exception when the parameter cannot be converted to a float.

A simple try-catch block prevents Domoticz to crash.

Verified on Ubuntu.

### DIFF
--- a/hardware/P1MeterBase.cpp
+++ b/hardware/P1MeterBase.cpp
@@ -258,402 +258,409 @@ void P1MeterBase::InitP1EncryptionState()
 
 bool P1MeterBase::MatchLine()
 {
-	if ((strlen(l_buffer) < 1) || (l_buffer[0] == 0x0a))
-		return true; //null value (startup)
+	try {
+		if ((strlen(l_buffer) < 1) || (l_buffer[0] == 0x0a))
+			return true; //null value (startup)
 
-	bool bFound = false;
+		bool bFound = false;
 
-	for (size_t i = 0; i < p1_matchlist.size(); ++i)
-	{
-		if (bFound)
-			break;
-
-		std::string sValue;
-
-		const P1Match* t = &p1_matchlist[i];
-		switch (t->matchtype)
+		for (size_t i = 0; i < p1_matchlist.size(); ++i)
 		{
-		case _eP1MatchType::ID:
-			// start of data
-			if (strncmp(t->key, l_buffer, strlen(t->key)) == 0)
-			{
-				m_linecount = 1;
-				bFound = true;
-			}
-			continue; // we do not process anything else on this line
-			break;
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thank you for this PR and the fix! 
Do you know the value of the string when this happens? Is the error caused by a parsing error before stof if called?

Is it possible to submit a PR with only the needed changes? You have a lot of line ident changes in this PR that should not be there

lokonli:
The P1 telegram contains errors. Examples below. The first line is the complete line. For some lines a key/value was found (second line). The value is the stof parameter that causes the exception:
1-0:72.7.0(:62.7.0(00.000*kW)
Key: voltagel3, Value: :62.7.0(00.000

1-0:52.7.0()

1-0:41.7.0(W)
Key: powerusel2, Value: W

Value is the part of the telegram line between opening bracket and (closing bracket or '*')

When Value doesn't start with a number, or when there is no Value (size 0), then stof will throw an exception.

The line indent changes are caused by the try/catch block. I can create a version without additional indents, but then the formatting doesn't look very nice. (github is not so smart in comparing changes, VSCode does a better job). Let me know what you prefer.

gizmocuz:
@lokonli , you could be very right about the format changes/shift caused by the try/catch block. Please ignore my comment ;)

Are those just randomly empty lines, or are these lines empty when no value update is to be reported?
If we ignore those empty lines, is the try/catch block still needed?

If yes, I'm happy to merge this

lokonli:
yes.

Values like 'W' and ':62.7.0(00.000' will throw errors as well.

lokonli:
Checking the code: it seems matchline will be called always, also when the telegram CRC is incorrect.
It would be better to first check the CRC, and then start interpreting the telegram.
But that is a somewhat bigger change.

gizmocuz:
okey happy to merge, but maybe above there should be a check if the value is empty (and then stop using it), or checking it is a valid number/float
But looks like a issue in the device itself... or it is allowed in the protocol specification

Yes I think continue parsing is okey as the lines should be matched for now

lokonli:
I can check for an empty value.
In case not empty use stof once to get a float, and catch the error (if any).
Will also make the indent impact smaller :)

lokonli:
Changed my mind. The try for the complete block is a better solution: Some keys may not have a value as parameter. That means the test for a float needs to be dependent on the key. This will add quite some code. So I propose to stick to the existing solution.

gizmocuz:
Thanks!!!

PR #5361, opened 2022-10-06
Changing default blind working to be consistent with other hard/software

gizmocuz:


gizmocuz:
Needs some more work:
- remove all inverted blind types from the code (making sure also in the JS/Python/dzVents code there is no reference to these old types)
- convert all current used inverted blind types to normal blind and set correct options

gizmocuz:
Should already work when using normal blind types and setting one of the new reverse options to 'true' when you edit the blind

gizmocuz:
image

jmcvieira1:
Hello i´m using  Version: 2022.1 (build 14535) i don't know if you are aware but the Venetian blinds EU / US button on the gui does not have the "expected" behavior, you press the close button it send´s the close command but highlights the open button
and the status word says open, my blind´s are dumb and they do not have any feedback. i see that you are working on this matter and did not see anyone complaining about it so i did not open any issue in case you already are working on this feel free to delete this message.
Thanks for hall your work.

gizmocuz:
@jmcvieira1 , thanks for reporting. I have tested the Venetian Blinds here with this PR and it works correctly (now) If you press Open, the left icon (open) is highlighted

gizmocuz:
The PR has been merged and there are no Inverted blinds anymore...
When you edit a switch/blind, you now have the option to reverse the state and/or position via checkboxes
Beta 14536 will be available within the hour for all platforms (lots of changes, lots to build)
Do not forget to clear your browser cache as the change involves some GUI changes/code as well!

PR #5375, opened 2022-10-19
Fix 'Boost asio error: no shared cipher'

Morolem:
Those changes fix a bug which describe in https://stackoverflow.com/questions/12217346/c-boost-asio-error-no-shared-cipher. I've seen this in connection.cpp connection::handle_handshake() when trying to connect via ssl socket.

This happens to me right after starting domoticz.


gizmocuz:

PR #5384, opened 2022-10-30
MercApi: Modified Token URL

kiddigital:
Mercedes is making a change to their Authentication and Authorization endpoint on the 2nd of November 2022.

See [here](https://developer.mercedes-benz.com/content-page/migration_2022) for more information.

This PR changes the endpoint to the new location.

Users also have to take some steps to re-create valid credentials and initial token, see the Domoticz forum thread [here](https://www.domoticz.com/forum/viewtopic.php?p=293880#p293880)


kiddigital:
@gizmocuz , I will merge this into the Beta on the 2nd of November (after doing some quick tests to see if it actually works).

Users that are already on the Beta channel, will be able to update once the Beta's are build.

Maybe we have to apply this very small change also to the Stable branch so users on the Stable channel can keep using the Mercedes Me Connect module.

PR #5387, opened 2022-10-31
fix SIBELGA P1meter garbage message

jacfont:
Belgiam P1Meter's from energy provider SIBELGA generate a garbage message which prevent the information to be processed.
The change, drop the garbage line.

### DIFF
--- a/hardware/P1MeterBase.cpp
+++ b/hardware/P1MeterBase.cpp
@@ -289,6 +289,9 @@ bool P1MeterBase::MatchLine()
 		if ((strlen(l_buffer) < 1) || (l_buffer[0] == 0x0a))
 			return true; //null value (startup)
 
+		if (strncmp("0-0:96.1.4(20202020", (const char *)&l_buffer, 18) == 0)
+			return true; //null value Sibelga
+
 		bool bFound = false;
 
 		for (size_t i = 0; i < p1_matchlist.size(); ++i)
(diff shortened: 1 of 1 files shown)


jacfont:
the wrong msg is the folllwing:
0-0:96.1.4(202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020)

gizmocuz:
@jacfont , is it possible you can post a few complete samples of the P1 telegrams?

jacfont:
@gizmocuz 
here is the description in attach.
[clear telegram.txt](https://github.com/domoticz/domoticz/files/9912007/clear.telegram.txt)

It was discussed on the foruum some time ago (seems not accessible anymore)
"[p1 energy meter not working - no devices shown](https://www.domoticz.com/forum/viewtopic.php?f=6&t=30562#p238994)"

I understand that this is very specific to this energy provider (and even potentially to a specific hardware/firmware of it).
I was wondering if there is a feature toggle possibility in a config file. 
( Does it improve code efficiency ? if (feature_xyz) { specific code} ) 

Without the proposed code, the telegram is dropped and no energy data is accepted.
The code detect the wrong line and just skip it.

gizmocuz:
Maybe it's the wrong topic you are pointing to.
But I do get a P1 device:

image

What version are you using? And if it's the stable, try the beta

jacfont:
We've the issue for about two years. At each upgrade we need to adapt the code and compile from source.
We've used the release version.
You see the error in the log at each telegram.
(about the forum topic, it can be it is related only partially (or not ?) with this isssue - as at the start of the setup there has been other issues)

kiddigital:
@jacfont , you are using the Stable.. could you quickly try the Beta version?

It seems that your problem is about flooding the (error)log. 

Of course we can just 'ignore' these errors. Either as you propose to filter out this very specific character sequence at the top, or maybe just drop the reporting (can you show some log output?).

But the `MatchLine()` function should be able to handle this without crashing (as it seems now). The matchlist seems to be aware of the specific Belgium version but it does not work (yet). Let's fix it there...

gizmocuz:
I don’t see any error in the log
As mentioned use the beta and if you see a error please post the exact error message here

jacfont:
Hello
Good news. We've tested with the release 2022.2. Everything is working fine.
The P1 telegram are processed correctly and there is no any error in the log anymore.
So I'll close this oull request.

PR #5388, opened 2022-10-31
teleinfo_hchp_pb

lcha78:
Hi,
this is a fix to french teleinfo with linky set in standard mode.
Depending on Energy provider, label NGTF may differ... so no KwhMeter is displayed  
I also remove badly implemented m_counter refresh limitation base on baud rate (the higher the baud rate, the less refress !). m_iRateLimit is already here to ajust refresh rate.

### DIFF
--- a/hardware/TeleinfoBase.cpp
+++ b/hardware/TeleinfoBase.cpp
@@ -46,7 +46,6 @@ void CTeleinfoBase::InitTeleinfo()
 {
 	m_bufferpos = 0;
 	m_teleinfo.CRCmode1 = 255;	 // Guess the CRC mode at first run
-	m_counter = 0;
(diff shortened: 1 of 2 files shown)


gizmocuz:
Thank you for this PR. If 'm_counter' is not used anymore, could you also remove it from the header file?

What is the default value of 'm_iRateLimit'? Can a user set this to 0? Is there a check for this? 
(Sorry did not look in the JS code)

lcha78:
Ok, remove 'm_counter' from header. 
Yes user can set value of m_iRateLimit right here :
image

m_iRateLimit can be set to 0 => No Rate limit.

gizmocuz:

PR #5420, opened 2022-11-18
Bugfix for MH9-CO2-WA / WD

MaartenJB:
Bugfix created by IanDury from the domoticz forum. This fixed the issue he had for me too. The original config template "mh9co2.xml" contains wrong values.

Source:
https://www.domoticz.com/forum/posting.php?mode=reply&t=35838

### DIFF
--- a/Config/mcohome/mh9co2.xml
+++ b/Config/mcohome/mh9co2.xml
@@ -3,27 +3,26 @@
     <!--MCOHome CO2 Monitor MH9-CO2-WD -->
     <!-- Configuration -->
     <CommandClass id="112">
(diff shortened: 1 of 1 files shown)


IanDury:
This fix will break devices using older firmware. See also: https://groups.google.com/g/openzwave/c/tPPx9v2JFLY

I've switched to Z-Wave JS UI last year which allows firmware version conditional configuration. OZW usage is deprecated and support for this will be dropped in a near beta version.

gizmocuz:
Thank you for this patch. Indeed best to switch to zwavejs

PR #5425, opened 2022-11-22
Better handling of remote and local proxy headers

### LINKED ISSUE #5403: Security improvements
> After updating Domoticz to the last version (2022.2) I do not have access to Domoticz (https://************.dataplicity.io port 80) via >Dataplicity (403 Forbidden). :(
>Log form Domoticz:
>2022-11-08 15:42:05.994 Error: Webserver: Multiple proxies are used (Or possible spoofing attempt), ignoring client request (remote address: 127.0.0.1)
>Version: 2022.2
>Platform: Raspberry Pi 4
>PS. Previous versions of Domoticz worked with Dataplicity without any problems

>@kiddigital
>You need to add the local network interface(s) to the (trusted) local networks.

>Previous version this was implicit, but also risky and unclear. Now it needs to be set explicitly.

>See [this](https://www.domoticz.com/forum/viewtopic.php?t=39312) forum topic for more info.

Even in earlier versions of domoticz, I had the local network set (127.0.0.1)
image
I have read SECURITY_SETUP.md and unfortunately I do not know how to set "Add clients" (I do not have / "do not see" the "Under 'Setup / More Options / Applications' it is possible to add / modify applications" (Domoticz 2022.2). Please guide me how to share Domoticz for Dataplicity (works under ZeroTier)

kiddigital:
This (draft) PR is a start to implement better processing of the Proxy headers coming from both local- and external proxies that might be used when trying to connect to the Domoticz built-in webserver.

The current (2022.2) processing of Proxy headers is limited and can cause issues with clients trying to connect, but getting rejected (HTTP 403 response) is cases where multiple proxies are involved. (see Issue #5403 for example).

Better support for multiple different proxy headers and different ways of how these headers behave. And also see if the _proposed_
'Forwarded HTTP Extension' [RFC7239](https://www.rfc-editor.org/rfc/rfc7239.html) could be supported as multiple webservers/proxies already implemented it.

### DIFF
--- a/test/lighttpd_proxy.conf
+++ b/test/lighttpd_proxy.conf
@@ -1,0 +1,68 @@
+# A lighttpd config file for Domoticz, serving the UI as Static content and proying only API calls to Domoticz
+#
+# Make sure you set the Document-root to your full-path of the Domoticz WWW directory
+#
+# All calls to 'json.htm' will be proxied to the Domoticz webserver/webservice that is expected to run on port 8080 (No SSL)
+# You can start Domoticz for example with the following config
+# sudo ./bin/domoticz -www 8080 -wwwroot www -loglevel debug,error -debuglevel webserver
+#
+# Domoticz will show the (web)requests it handles (should be the json.htm ones only)
+# And you can tail the lighttpd logfile (/tmp/lighttpd.log) to see all other static requests being handled by lighttpd
+#
+# Start this script from your Domoticz home directory with:
+# lighttpd -D -f test/lighttpd.conf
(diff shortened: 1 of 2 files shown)


kiddigital:
@gizmocuz , I would like to merge these small changes into Beta already. It allows users that having issues with multiple proxies to gather debugging information needed to better understand the real world scenario's that are out there already.

I will start a new PR for the actual implementation of better proxy header handling.

gizmocuz:
Yes sure go ahead!

PR #5430, opened 2022-11-27
Improve Trusted Network processing and no more anonymous user

### LINKED ISSUE #5396: Custom Page is always active (some browsers, not all)
When I open Domoticz via mobile, the Custom Page is always active.  
Raspberry pi 3bplus
 Domoticz version 2022.2 (build 14708)

### LINKED ISSUE #5423: Floorplan visable when button is disable
I notice it that the floorplan button was visible on my test system. (normal default install settings)
It is an clean setup that is 14 days old. cache, app cache and incognito modus make no difference. 

Domoticz build 14727

afbeelding

### LINKED ISSUE #5429: Mobile view on dz beta without floorplans
Hello, withlatest beta release in my  mobile smartphone if i use mobile browsers to see domoticz i see this menu without the floorolans menu and with all menu icons on top that are all placed on the right side. 
Here below a screenshot. 
With old releases i was able to see the menu of floatplanes and all the menu icons was on opposite side (left) 
Thanks 
Screenshot_20221124_235002

kiddigital:
This PR improves the processing and handling of **Trusted networks**, which was called _local networks_ before.

Now, when a User connects from a Trusted network, the identity of the first _admin user_ is taken. So although no explicit User (and password) are given, **everything happens under a known identity now!**

Impact:

- There now always is a 'Logout' button in the menu (but you will get logged in immediately again when you are in a trusted network) which shows the Username of the logged-in user.
- On the `Setup/settings/system`-tab, there is no longer a system-wide option to set certain menu tabs on or off as now always the User settings will be used. (as this was actually not system-wide but only used for _anonymous_ access)
- The `-nowwwpwd` settings previously did not load any Users (so only anonymous access was possible), but now still all users are loaded. Now when this option is specified during startup, the `0.0.0.0/0` network mask (IPv4) and `::` (IPv6) is added which effectively means that _every ip-address_ (also on the Internet!) is considered part of a _trusted network_. Therefor no Username and Password are requested, allowing Users to reset the lost password(s).

### DIFF
--- a/main/SQLHelper.cpp
+++ b/main/SQLHelper.cpp
@@ -3053,6 +3053,14 @@ bool CSQLHelper::OpenDatabase()
 					_log.Log(LOG_STATUS, "A default admin User called 'admin' has been added with a default password!");
 				}
 			}
+			// Remove these Pref vars as we do not use them anymore
+			DeletePreferencesVar("EnableTabLights");
+			DeletePreferencesVar("EnableTabTemp");
+			DeletePreferencesVar("EnableTabWeather");
+			DeletePreferencesVar("EnableTabUtility");
+			DeletePreferencesVar("EnableTabCustom");
+			DeletePreferencesVar("EnableTabScenes");
+			DeletePreferencesVar("EnableTabFloorplans");
 		}
 	}
 	else if (bNewInstall)
@@ -3241,34 +3249,6 @@ bool CSQLHelper::OpenDatabase()
 	if (!GetPreferencesVar("SmartMeterType", nValue))	//0=meter has decimals, 1=meter does not have decimals, need this for the day graph
 	{
 		UpdatePreferencesVar("SmartMeterType", 0);
-	}
-	if (!GetPreferencesVar("EnableTabLights", nValue))
-	{
-		UpdatePreferencesVar("EnableTabLights", 1);
-	}
-	if (!GetPreferencesVar("EnableTabTemp", nValue))
-	{
-		UpdatePreferencesVar("EnableTabTemp", 1);
-	}
-	if (!GetPreferencesVar("EnableTabWeather", nValue))
-	{
-		UpdatePreferencesVar("EnableTabWeather", 1);
-	}
-	if (!GetPreferencesVar("EnableTabUtility", nValue))
-	{
-		UpdatePreferencesVar("EnableTabUtility", 1);
-	}
-	if (!GetPreferencesVar("EnableTabCustom", nValue))
-	{
-		UpdatePreferencesVar("EnableTabCustom", 1);
-	}
-	if (!GetPreferencesVar("EnableTabScenes", nValue))
-	{
-		UpdatePreferencesVar("EnableTabScenes", 1);
-	}
-	if (!GetPreferencesVar("EnableTabFloorplans", nValue))
-	{
-		UpdatePreferencesVar("EnableTabFloorplans", 0);
 	}
 	if (!GetPreferencesVar("NotificationSensorInterval", nValue))
 	{
--- a/main/SQLHelper.cpp
+++ b/main/SQLHelper.cpp
@@ -40,7 +40,7 @@
 #define __STDC_FORMAT_MACROS
 #include <inttypes.h>
 
-#define DB_VERSION 157
+#define DB_VERSION 158
 
 #define DEFAULT_ADMINUSER "admin"
 #define DEFAULT_ADMINPWD "domoticz"
@@ -3048,11 +3048,14 @@ bool CSQLHelper::OpenDatabase()
 				nValue = atoi(result[0][0].c_str());
 				if (nValue == 0)
 				{
-					// Add a default User as no users exist
+					// Add a default Admin User as no admin users exist
 					safe_query("INSERT INTO Users (Active, Username, Password, Rights, TabsEnabled) VALUES (1, '%s', '%s', %d, 63)", base64_encode(DEFAULT_ADMINUSER).c_str(), GenerateMD5Hash(DEFAULT_ADMINPWD).c_str(), http::server::URIGHTS_ADMIN);
 					_log.Log(LOG_STATUS, "A default admin User called 'admin' has been added with a default password!");
 				}
 			}
+		}
+		if (dbversion < 158)
+		{
 			// Remove these Pref vars as we do not use them anymore
 			DeletePreferencesVar("EnableTabLights");
 			DeletePreferencesVar("EnableTabTemp");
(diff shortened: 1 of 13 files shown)


main/SQLHelper.cpp
```diff
@@ -3052,6 +3052,14 @@ bool CSQLHelper::OpenDatabase()
 					_log.Log(LOG_STATUS, "A default admin User called 'admin' has been added with a default password!");
 				}
 			}
+			// Remove these Pref vars as we do not use them anymore
+			DeletePreferencesVar("EnableTabLights");
```
gizmocuz, line 3060: Could you increase the DB_VERSION and make this a patch? (see examples like if (debversion < 155) etc

main/WebServer.cpp
```diff
@@ -318,19 +318,21 @@ namespace http
...
+			{
+				std::vector<std::string> strarray;
+				StringSplit(WebLocalNetworks, ";", strarray);
+				for (const auto& str : strarray)
+					m_pWebEm->AddTrustedNetworks(str);
+			}
+			if (bIgnoreUsernamePassword)
+			{
+				m_pWebEm->AddTrustedNetworks("0.0.0.0/0");
+				_log.Log(LOG_STATUS, "Allowing access without username/password as all incoming traffic is considered trusted! SECURITY RISK!");
```
gizmocuz: Maybe start this line with "SECURITY RISK: Allowing acc....."

gizmocuz (commented):
PR looks perfect! Just two minor change requests

gizmocuz:
Maybe also rename the preferences key 'WebLocalNetworks' to 'TrustedNetworks'
And delete 'WebUserName' (also in the code)

kiddigital:
> Maybe also rename the preferences key 'WebLocalNetworks' to 'TrustedNetworks' 

Being used in many places atm, so needs more time and testing...

> And delete 'WebUserName' (also in the code)

Only used in dzVents now, but needs more time and testing as well... (and not break the automated tests of dzVents)

kiddigital:
@gizmocuz , I think this PR is ready to be merged.

Once I got all dzVents tests running again, I will try to see if I can fully remove the `webUserName` and `localNetwork` references from the codebase. At least there are gone in the UI.

One other thought I have is if we should always use the first _admin_ user as the anonymous login user? Wouldn't it make more sense to have a certain user selected? Anonymous access makes sense in a known/trusted environment, but that does not perse mean it should be an admin... thoughts for future improvement 😁

gizmocuz:
It's not anonymous access. It is trusted access, and a admin can be trusted.
In the very old times, it was possible to use basic-auth where you could specify a user.
Better would be to use the open api with a JWT token.
I suspect it is used the most for scripts to update devices (maybe also create devices) and there is no need to do other admin things. (like delete or configuration etc)
I will start implementing the JWT token for user access soon. With this there should be no need to use the Trusted Network access to login without a username/password within your private network.
Just login once, and when you use the gui once in a week it should keep you logged in

main/SQLHelper.cpp
```diff
@@ -3048,20 +3048,31 @@ bool CSQLHelper::OpenDatabase()
...
+		{
+			// Remove these Pref vars as we do not use them anymore
+			DeletePreferencesVar("EnableTabLights");
+			DeletePreferencesVar("EnableTabTemp");
+			DeletePreferencesVar("EnableTabWeather");
+			DeletePreferencesVar("EnableTabUtility");
+			DeletePreferencesVar("EnableTabCustom");
+			DeletePreferencesVar("EnableTabScenes");
+			DeletePreferencesVar("EnableTabFloorplans");
+		}
```
gizmocuz, line 3067: Maybe also get the "WebLocalNetworks" variable, and put it in a TrustedNetworks variable, and delete the "WebLocalNetworks" variable
(And of course use TrustedNetworks troughouth the code/js

main/WebServer.cpp
```diff
@@ -318,19 +318,21 @@ namespace http
...
+				std::vector<std::string> strarray;
+				StringSplit(TrustedNetworks, ";", strarray);
+				for (const auto& str : strarray)
+					m_pWebEm->AddTrustedNetworks(str);
+			}
+			if (bIgnoreUsernamePassword)
+			{
+				m_pWebEm->AddTrustedNetworks("0.0.0.0/0");	// IPv4
+				m_pWebEm->AddTrustedNetworks("::");	// IPv6
+				_log.Log(LOG_ERROR, "SECURITY RISK! Allowing access without username/password as all incoming traffic is considered trusted!");
```
gizmocuz: Maybe also mention to change the admin password as soon as possible and restart Domoticz
Or when the password has changed the above trusted networks are removed?

main/WebServer.cpp
```diff
@@ -3182,30 +3184,21 @@ namespace http
 
 			std::vector<std::vector<std::string>> result;
 
-			if ((UserID != 0) && (UserID != 10000))
+			result = m_sql.safe_query("SELECT TabsEnabled FROM Users WHERE (ID==%lu)", UserID);
+			if (!result.empty())
```
gizmocuz: TabsEnabled always exitst so the if/then/else check can be deleted

www/app/app.js
```diff
@@ -562,6 +566,7 @@ define(['angularAMD', 'app.routes', 'app.constants', 'app.notifications', 'app.p
 			}
 		});
 		permissions.setPermissions(permissionList);
+		$rootScope.config.userName = permissionList.user;
```
gizmocuz: bit mixed feelings about this one.
You can set the userName within the function 'setPermissions', which I believe you already do now with the patch you made in app.permissions.js

gizmocuz:
PR looks great! Just a few small comments/suggestions

main/WebServer.cpp
kiddigital: removed (although from a defensive programming strategy it never hurts to double check if the working assumption is still valid ;) )

main/WebServer.cpp
kiddigital: extended the message

main/SQLHelper.cpp
kiddigital: will do in another PR, but need some more time... UI wise it is already gone..

www/app/app.js
kiddigital: I had the same... but the 'permissions' object is never added to the rootscope, so not accessible in the html template
Therefor this was my way to get access to the Username, but open for better alternatives (here my lack of front-end/angular knowledge plays a part as well).

main/WebServer.cpp
gizmocuz: Yes you are correct there, but there is no need for the 'else' statement, and if empty something serious was wrong (like database corruption) and you could bail out

www/app/app.js
gizmocuz: Okey maybe as alternative, where you call 'getconfig' (app.js/$rootScope.GetGlobalConfig), also return the username, and store it in the rootScope config ?

kiddigital: in the current order of processing (in the `app.run()`-callback method) that does not work as the order now is

- getConfig (json.htm?type=command&param=getconfig)
- getVersion (json.htm?type=command&param=getversion)
- getAuth (json.htm?type=command&param=getauth)
Only after the `getAuth` the Auth information is retrieved (incl. the Username) and stored into the Permission object.
And the getConfig does not retrieve this information so cannot set it into the rootScope config at that time.

Maybe we could consider to extend the 'getConfig' (which is _only_ used in the `app.js` controller) with information like the Version and Userinfo so everything can be retrieved in a single call (instead of 3 XHR calls at the moment). Would that be a better direction? (and we leave the other 2 as they are used for version checking and login/logout)

kiddigital: hm... just looking at the getConfig command, it already uses the session information to get the correct Tabs for the specific User :) So it already has User specific processing which makes adding the Username easy.

gizmocuz: Yep that was my idea as well (to extend the returned info)

kiddigital: done :) .. getConfig now includes the Username and all info from getVersion (simply by calling the command) and the UI no longer needs to do a separate request to getVersion when starting (saves a roundtrip)

gizmocuz:
Thanks again for all work!!!

PR #5439, opened 2022-12-02
Fix EcoDevices1 HP Jour Blanc (bug)

xbeaudouin:
On the EDF Tempo plan, there is 3 type of electricity bills. There were a mistake on HP JW (Heures Plein Jour Blanc) that still used the index of Jour Bleus (HPJB). This was maybe a typo, but fix the issue of not having HP when tarif is Jour Blanc.


gizmocuz:
Thanks for spotting and this PR!

PR #5443, opened 2022-12-05
Better handling of remote and local proxy headers - Continued

kiddigital:
This PR is a continuation of implementing better processing of the Proxy headers coming from both local- and external proxies that might be used when trying to connect to the Domoticz built-in webserver. (See PR #5425 )

The current (2022.2) processing of Proxy headers is limited and can cause issues with clients trying to connect, but getting rejected (HTTP 403 response) is cases where multiple proxies are involved. (see Issue #5403 for example).

The PR provides better support for multiple different proxy headers and different ways of how these headers behave. Also now it supports the proposed 'Forwarded HTTP Extension' [RFC7239](https://www.rfc-editor.org/rfc/rfc7239.html) as multiple webservers/proxies already implemented it.

It looks for the following headers, in order of importance (if it finds one, it will not look for others):

- `Forwarded` (RFC7239)
- `X-Forwarded-For`
- `X-Real-IP`

If it finds one of these headers, it will try to process it. 

_When the header can not be processed correctly because it does not comply to the specification, it is considered as an hacking attempt and the request will be denied_!

### DIFF
--- a/webserver/cWebem.cpp
+++ b/webserver/cWebem.cpp
@@ -1157,11 +1157,113 @@ namespace http {
 			m_session_clean_timer.async_wait([this](auto &&) { CleanSessions(); });
 		}
 
-		bool cWebem::HandleProxies(const request &req, WebEmSession &sessions)
+		bool cWebem::findRealHostBehindProxies(const request &req, std::string &realhost)
 		{
+			// Checking for 3 possible headers (in order of handling)
+			// "Forwarded"	RFC7239  (https://www.rfc-editor.org/rfc/rfc7239)
+			// "X-Forwarded-For" The defacto standard header used by many web/proxy servers (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For)
+			// "X-Real-IP"	The (old) default header used by NGINX  (http://nginx.org/en/docs/http/ngx_http_realip_module.html#real_ip_header)
+			//
+			// These headers can occur multiple times, so need to be 'squashed' together
+			// And a single line can contain multiple (comma separated) values in order
+
 			bool bStatus = true;
-			_log.Debug(DEBUG_WEBSERVER, "[web:%s] Checking proxy headers!", GetPort().c_str());
+			std::vector<std::string> headers;
+			std::vector<std::string> hosts;
+
+			if (sumProxyHeader("Forwarded", req, headers))
+			{	// We found one or more Forwarded headers that need to be processed into a list of Hosts
+				if (!parseForwardedProxyHeader(headers, hosts))
+				{
+					bStatus = false;
+				}
+			}
+			else if (sumProxyHeader("X-Forwarded-For", req, headers))
+			{	// We found one or more X-Forwarded-For headers that need to be processed into a list of Hosts
+				if (!parseProxyHeader(headers, hosts))
+				{
+					bStatus = false;
+				}
+			}
+			else if (sumProxyHeader("X-Real-IP", req, headers))
+			{	// We found one or more X-Real-IP headers that need to be processed into a list of Hosts
+				if (!parseProxyHeader(headers, hosts))
+				{
+					bStatus = false;
+				}
+			}
+
+			if (hosts.size() >= 1)
+			{
+				realhost = hosts[0];	// Even if we found a chain of hosts, we always use the first (= origin)
+			}
+
 			return bStatus;
+		}
+
+		bool cWebem::sumProxyHeader(const std::string sHeader, const request &req, std::vector<std::string> &vHeaderLines)
+		{
+			uint8_t iFound = 0;
+
+			for (const auto &header : req.headers)
+			{
+				if (sHeader.compare(header.name) == 0)
+				{
+					iFound++;
+					vHeaderLines.push_back(header.value);
+				}
+			}
+
+			return (iFound > 0);
+		}
+
+		bool cWebem::parseProxyHeader(std::vector<std::string> &vHeaderLines, std::vector<std::string> &vHosts)
+		{
+			uint8_t iFound = 0;
+
+			for (const auto sLine : vHeaderLines)
+			{
+				std::vector<std::string> vLineParts;
+				StringSplit(sLine, ",", vLineParts);
+				for (std::string sPart : vLineParts)
+				{
+					iFound++;
+					stdstring_trimws(sPart);
+					vHosts.push_back(sPart);
+				}
+			}
+
+			return (iFound > 0);
+		}
+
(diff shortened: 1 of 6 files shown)


webserver/cWebem.cpp
```diff
@@ -1157,6 +1157,115 @@ namespace http {
...
+
+			if (hosts.size() >= 1)
+			{
+				realhost = hosts[0];	// Even if we found a chain of hosts, we always use the first (= origin)
+			}
+
+			return bStatus;
+		}
+
+		bool cWebem::sumProxyHeader(const std::string sHeader, const request &req, std::vector<std::string> &vHeaderLines)
```
gizmocuz: could you make sHeader const (const std::string &sHeader)

webserver/cWebem.cpp
```diff
@@ -1157,6 +1157,115 @@ namespace http {
...
+		{
+			uint8_t iFound = 0;
+
+			for (const auto &header : req.headers)
+			{
+				if (sHeader.compare(header.name) == 0)
```
gizmocuz: quote:
The disadvantage of compare is that you need to check that str2.length() is no greater than str1.length() (or catch the resulting exception and treat it as a false result).

I think you want to check if header.name is equal or begins with sHeader?
Could you do

`if (header.name.find(sHeader)==0)`

webserver/cWebem.cpp
```diff
@@ -1157,6 +1157,115 @@ namespace http {
...
+				{
+					iFound++;
+					vHeaderLines.push_back(header.value);
+				}
+			}
+
+			return (iFound > 0);
+		}
+
+		bool cWebem::parseProxyHeader(std::vector<std::string> &vHeaderLines, std::vector<std::string> &vHosts)
```
gizmocuz: please declare vHeaderLines with const

`const std::vector<std::string> &vHeaderLines`

webserver/cWebem.cpp
```diff
@@ -1157,6 +1157,115 @@ namespace http {
...
+					iFound++;
+					stdstring_trimws(sPart);
+					vHosts.push_back(sPart);
+				}
+			}
+
+			return (iFound > 0);
+		}
+
+		bool cWebem::parseForwardedProxyHeader(std::vector<std::string> &vHeaderLines, std::vector<std::string> &vHosts)
```
gizmocuz: please declare vHeaderLines with const

`const std::vector<std::string> &vHeaderLines`

webserver/cWebem.cpp
```diff
@@ -1157,6 +1157,115 @@ namespace http {
...
+			{	// We found one or more X-Real-IP headers that need to be processed into a list of Hosts
+				if (!parseProxyHeader(headers, hosts))
+				{
+					bStatus = false;
+				}
+			}
+
+			if (hosts.size() >= 1)
+			{
+				realhost = hosts[0];	// Even if we found a chain of hosts, we always use the first (= origin)
```
gizmocuz: instead of a bStatus boolean we could return true here or else return false at the bottom of this function?

webserver/cWebem.cpp
```diff
@@ -1157,6 +1157,115 @@ namespace http {
...
+				StringSplit(sLine, ",", vLineParts);
+				for (std::string sPart : vLineParts)
+				{
+					iFound++;
+					stdstring_trimws(sPart);
+					vHosts.push_back(sPart);
+				}
+			}
+
+			return (iFound > 0);
```
gizmocuz: Instead of iFound we can use

`return !vHosts.empty();`

webserver/cWebem.cpp
```diff
@@ -1157,6 +1157,115 @@ namespace http {
...
+							iePos = sPart.find(";", isPos);
+						}
+						std::string sSub = sPart.substr(isPos, (iePos - isPos));
+						stdstring_trimws(sSub);
+						vHosts.push_back(sSub);
+					}
+				}
+			}
+
+			return (iFound > 0);
```
gizmocuz: Instead of iFound we can use

`return !vHosts.empty();`

webserver/connection.cpp
```diff
@@ -431,6 +431,8 @@ namespace http {
 							// LogFormat "%h %l %u %f \"%r\" %>s %b \"%{Referer}i\" \"%{User-agent}i\"" combined
 							// 127.0.0.1 - frank [10/Oct/2000:13:55:36.012 -0700] "GET /apache_pb.gif HTTP/1.0" 200 2326 "http://my.domoticz.local/index.html" "Mozilla/4.08 [en] (Win98; I ;Nav)"
 							std::string wlHost = request_.host_remote_address;
+							if (!reply_.originHost.empty())
+								wlHost = reply_.originHost;
```
gizmocuz: Might be faster to write

`std::string wlHost = (reply_.originHost.empty()) ? request_.host_remote_address : reply_.originHost;`

Otherwise you always assign 'host_remote_address ' and still always perform the check

gizmocuz (commented):
Some small comments/suggestions

webserver/cWebem.cpp
kiddigital: the True/False is about if we were able to process the headers correctly yes or no. That is independent of the fact if we found alternative hosts yes or no. Checking if an actual alternative host is found and what to do with it, is left up to the caller.

Personally I prefer to exit a function (`return`) at 1 (or as least as possible) location in the code so it is clear where and what the exit does. Improves readability and future maintenance (IMO).

webserver/cWebem.cpp
kiddigital: Yes/No... in this case it could be as the actual fact is that the input vector is most likely (and in this case) empty.

But when just looking at the function definition, it allows a vector as input and just ADD's values regardless if the input was empty or not. That is why the `iFound` indicates if values have been added or not.

Checking on `empty()` could be false even when nothing was added, but the input wasn't empty to begin with.

It is debatable I agree, but I prefer when looking just at the function (without considering its environment and other assumption and knowledge) it is self-explanatory. In this current way, it is in my opinion.

But if it needs to be changed...

webserver/cWebem.cpp
kiddigital: see above

webserver/connection.cpp
kiddigital: done

webserver/cWebem.cpp
gizmocuz: There is no need to have a counting value in this function, speed wise, code size etc...

webserver/cWebem.cpp
gizmocuz: There is no need to have a counting value in this function, speed wise, code size etc...
It is also cleaner to use 

`return !vHosts.empty();`

kiddigital: done

PR #5450, opened 2022-12-09
FIX: ensure an active admin is present

### LINKED ISSUE #5445: After last upgrade frontend stop works (when no active admin user in the system)
image

kiddigital:
In PR #5430 the 'anonymous' usage of Domoticz was replaced by _known_ usage by ensuring that everything was done under a known admin account, especially when coming from a _Trusted network_ meaning no login (asking for Username / Password) required.

In certain setups this could lead to a strange situation that when in a _trusted network_ no _active_ admin could be found, the system was in an unknown state.

This PR adds checking and prevents disabling the last remaining active User with administrative privileges.

In the rear situation that somehow no active admin is available, one can restart the system with the `-nowwwpwd` option and a _tmpadmin_ will be used allowing to do the needed User management and making sure that their is an active admin in the system.


gizmocuz (approved):
Looks perfect!

PR #5451, opened 2022-12-10
Fix #5448

### LINKED ISSUE #5448: Arilux - Mono light type are added as RGB
When adding a light in Arilux hardware, even if we select "Mono" as Light Type:

Domoticz_MagicHome_20221208_162759

the device is added as RGB:

Domoticz_MagicHome_20221208_162914

I guess it may be related to the mismatch between the label and the data-i18n element in [Hardware.html](https://github.com/domoticz/domoticz/blob/f64304efb7c74a7f37f237a06c16c8a5544a14f7/www/app/hardware/Hardware.html#L1622):

https://github.com/domoticz/domoticz/blob/f64304efb7c74a7f37f237a06c16c8a5544a14f7/www/app/hardware/Hardware.html#L1622

But still, it would be nice to have all the possibilities offered by ColorSwitch and to be able to add Single Color LED controller without the need to patch the subtype in the database.

rpeyron:
Hello, should fix #5448 

I replaced values of subtype in Hardware.html by those in ColorSwitch.h to avoid extra values mapping, and added a value check for sanity.
I defined Mono, RGB, RGBW, RGBWZ, but I can test only Mono.

Thanks

### DIFF
--- a/hardware/Arilux.cpp
+++ b/hardware/Arilux.cpp
@@ -221,8 +221,12 @@ namespace http {
 
 			int HwdID = atoi(idx.c_str());
 
+			// stype values aligned to ColorSwitch.h enum values in Hardware.html
+			int subtype = std::stoi(stype);
+			if ( (subtype <= 0) || (subtype > 8) ) subtype = 1;
+
 			Arilux Arilux(HwdID);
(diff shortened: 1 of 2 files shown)


gizmocuz:

PR #5471, opened 2022-12-18
Improve use of Basic Auth as authentication method for API calls within- and outside Trusted Networks

### LINKED ISSUE #5461: retrieving / updating via http api doesn't work anymore
Docker image was updated to build 14852, updating T/H sensors or retrieving data from domoticz using curl doesn't work anymore. Authentication is done as described here: https://www.domoticz.com/wiki/Domoticz_API/JSON_URL%27s#Format

http://<username:password@>domoticz-ip<:port>/json.htm?api-call

the request returns:
`<html><head><title>Unauthorized</title></head><body><h1>401 Unauthorized</h1></body></html>`

### LINKED ISSUE #5465: Unauthorized errors with username+password in URL
Since build 14855 or probably a couple before, I'm unable to login using the username + password in the url. Suddenly after upgrading dmomoticz,  my dashticz and locative aren't working anymore, as Domoticz responds with HTTP-401 Unauthorized. I've tried adding a fresh user, put in the base64 encoded username/passwords, doesn't work.
Also tried adding my local network to the allowed list (it wasn't added before), but that also doesn't work. Always getting an HTTP-401. It worked for years, I guess something got broken in recent builds

kiddigital:
This PR addresses how Basic Auth authentication can be used when performing API calls (`/json.htm`).

Domoticz **always** requires that actions are performed by an Authenticated user (see the 'Users' settings to manage User rights, etc.).

Authentication can be provided in several ways:

- Providing valid login credentials on the Login screen (_not usable for API calls_)
- Providing a valid Token (JWT) or certificate (X509)
- Providing a valid Basic Auth header
- When the request originates from within the Trusted Network and **no** other Authentication method is provided, Domoticz will look for the first User with admin privileges and assume the identity of that User. _This is very useful for the UI so no login screen is presented to users in the Trusted network_.

As _Basic Auth_ is not really considered safe anymore, it _was_ limited to only use it within the Trusted Network.

This PR changes that limitation.

**Now _Basic Auth_ is accepted when done over HTTPS, also from outside the Trusted Network.**

**Additional Setting**: A new setting is introduced on the 'Security' page of the Settings. In the 'Website Protection' block, it is now possible to _Allow Basic Authentication over plain HTTP_ regardless if it is coming from within- or outside the Trusted network.

image

With this setting, API calls using Basic Authentication will also be accepted when _NOT_ using secured HTTPS calls.

**NOTE**: For setups where a (reverse) Proxy is used, the IP address of that Proxy should be in the list of Trusted Networks. If that is the case, Domoticz will trust the Proxy Header information from the Proxy to determine the _origin IP address_ of the request. This IP address is than used as the real source of the request, meaning that Domoticz will look at that address to determine if it originates from the Trusted network or not. 

It should address issues #5465 , #5461

### DIFF
--- a/main/WebServer.cpp
+++ b/main/WebServer.cpp
@@ -762,6 +762,13 @@ namespace http
 			m_pWebEm->SetWebCompressionMode(gzmode);
 		}
 
+		void CWebServer::SetAllowPlainBasicAuth(const bool allow)
+		{
+			if (m_pWebEm == nullptr)
+				return;
+			m_pWebEm->SetAllowPlainBasicAuth(allow);
+		}
+
 		void CWebServer::SetAuthenticationMethod(const _eAuthenticationMethod amethod)
 		{
 			if (m_pWebEm == nullptr)
@@ -8667,10 +8674,15 @@ namespace http
 				m_pWebEm->SetAuthenticationMethod(amethod);
 				cntSettings++;
 
+				bool AllowPlainBasicAuth = (request::findValue(&req, "AllowPlainBasicAuth") == "on" ? 1 : 0);
+				m_sql.UpdatePreferencesVar("AllowPlainBasicAuth", AllowPlainBasicAuth);
+
+				m_pWebEm->SetAllowPlainBasicAuth(AllowPlainBasicAuth);
+				cntSettings++;
+
 				std::string WebLocalNetworks = CURLEncode::URLDecode(request::findValue(&req, "WebLocalNetworks"));
 				m_sql.UpdatePreferencesVar("WebLocalNetworks", WebLocalNetworks);
 				m_webservers.ReloadTrustedNetworks();
-				cntSettings++;
 				cntSettings++;
 
 				if (session.username.empty())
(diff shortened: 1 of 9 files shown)


kiddigital:
@gizmocuz , this PR is Draft as I will be adding an additional setting on the 'Security' page to also allow Basic Auth over Non HTTPS so the User has an option to allow this.

Although that is even more unsafe, it might be needed as not always Domoticz runs over HTTPS and the User explicitly needs to allow it. Off by default of course.

pipiche38:
I  do not understand why on my local network I should have https running (which is the purpose of the proxy), and so if I do consider that my local network is a trusted one, why should I have to do Basic auth, which them will required me to go with a Username/Password using Base64 .

Please let the user make their choice.

kiddigital:
> 
> Please let the user make their choice.

Yes, that is why the option to accept Basic Auth headers in non HTTPS setups is also available.

And in a Trusted Network you could do _without_ any Authorization (so no need for Basic Auth or others) if you want. Although it is a good practice to be explicit on the User that you expect to perform the operation.

kiddigital:
@gizmocuz , PR is ready and can be merged (after reviewing of course)

gizmocuz (approved):
Looking good!

PR #5472, opened 2022-12-18
hardware/EnOceanRawValue.h: include stdarg.h

ffontaine:
Include `stdarg.h` to avoid the following build failure on uclibc:

```
In file included from /home/fabrice/buildroot/output/build/domoticz-2022.2/hardware/EnOceanRawValue.cpp:2: /home/fabrice/buildroot/output/build/domoticz-2022.2/hardware/EnOceanRawValue.h:109:83: error: 'va_list' has not been declared
  109 | uint32_t SetRawValuesNb(uint8_t * data, T_DATAFIELD * OffsetDes, int NbParameter, va_list value);
      |                                                                                   ^~~~~~~
```

### DIFF
--- a/hardware/EnOceanRawValue.h
+++ b/hardware/EnOceanRawValue.h
@@ -1,6 +1,7 @@
 #ifndef _SetGetRawValue
 #define _SetGetRawValue
 
+#include <stdarg.h>
 #include <stdint.h>
 #include <vector>
 #include <stdio.h>


kiddigital:
@ffontaine , thx... can you tell a little more on the environment in which you get this Build error? (version of OS, compiler, CMake, etc.) as in our current Build pipelines this error does not occur (yet).

ffontaine:
Build failure is raised on buildroot with uclibc (as already stated in the commit message), gcc 11.3.0 and domoticz in version 2022.2.
`va_list` is defined in `stdarg.h` as stated in various sources including wikipedia: https://en.wikipedia.org/wiki/Stdarg.h.

kiddigital:
Thank you!

gizmocuz:
Moved the include to the top of the .cpp file
The include was already in the .cpp file arround line 130

ffontaine:
Why are you moving the include in the `.cpp` file? `va_list` is used in `EnOceanRawValue.h`. This doesn't seem future-proof as a build failure will be raised if  `EnOceanRawValue.h` is included in any source file which doesn't include `stdarg.h` first.

gizmocuz:
@ffontaine , in that case, the code should be moved to the .cpp file as this is a header file

gizmocuz:
Because i noticed it was included in the .cpp file somewhere in the middle
Anyway, did some cleaning up as well you should be able to compile now without this error

PR #5475, opened 2022-12-18
Removes 'Return' Key event's handling from 'dialog-edittempdevice'

soif:
In temperatures page, it was not possible to enter a multi-line description, 
because the Return key event is intercepted by the javacript, to rather click on the first button.

This PR fixes this, by removing keyCode 13 handling from JS.


gizmocuz:
Thank you for this PR!

PR #5480, opened 2022-12-19
Moved ShowUpdateNotification() to right scope

### LINKED ISSUE #5445: After last upgrade frontend stop works (when no active admin user in the system)

kiddigital:
PR with possible fix for UI issues like #5445


gizmocuz:
Thank you for the PR, this should be solved by now

PR #5484, opened 2022-12-20
Fully responsive SelectMenu

soif:
Light/Switch Page:
 When reducing the browser windows width, the items **SELECT menus** widths are not properly resized, and they exceed on the right from the item block.
 
This PR removes the inline width style (built by $ui.selectmenu() ), so the width can now be controlled from the CSS, thus making it fully dynamic / responsive...


gizmocuz:
Thanks!

Issue #5395, opened 2022-11-07
Blind + Stop inconsistancies EnOcean


didi31:
I have enOcean USB (ESP3) hardware which created the "Volet cuisine" device.
domoticz.devices("Volet cuisine").state  = "Open" and domoticz.devices("Volet cuisine").level  = 100
The GUI shows the picture of the device "open" active and the slider is on the left side.
![image](https://user-images.githubusercontent.com/46863073/200346387-8fc1f351-f977-491b-b962-b76282581881.png)

I enter the following code in dzvents
**Problem**
I enter the code below:
`domoticz.devices("Volet cuisine").close()`
The blind is closed but dzVents return the following:

- domoticz.devices("Volet cuisine").state  = "Open"
- domoticz.devices("Volet cuisine").level  = 0

The GUI shows 
- the picture of the device "open" active
- the slider is on the left side
- the state is shown at 0%

If I execute the same code again, all is fine
The GUI shows 
- the picture of the device "Closed" is active
- the slider is on the right side
- the state is "Closed"

Now I execute the following code:
`domoticz.devices("Volet cuisine").open()`
The blind is then opened but the slider is on the right side.
![image](https://user-images.githubusercontent.com/46863073/200346226-4ca73871-7f53-4709-966a-5d64adee5b47.png)

didi31:
From the previous state:
domoticz.devices("Volet cuisine").state = "Open"
domoticz.devices("Volet cuisine").level = 100
domoticz.devices("Volet cuisine").lastLevel = 0

if I click on the "Open" button, the slider move on the left side and dzVents return the following:

domoticz.devices("Volet cuisine").state = "Open"
domoticz.devices("Volet cuisine").level = 100
domoticz.devices("Volet cuisine").lastLevel = **154**

walawa75:
Hi,

Same issue here with Zwave devices. I confirm too that this new stable version is very buggy and unusable. I rolled back to beta (stable and usable)

kiddigital:
> Hi,
> 
> Same issue here with Zwave devices. I confirm too that this new stable version is very buggy and unusable. I rolled back to beta (stable and usable)

Zwave in general or are you referring to Blinds specifically?

Did you look at [this](https://domoticz.com/forum/viewtopic.php?p=293724) forum thread?

Did you read the [wiki](https://www.domoticz.com/wiki/Blinds)?

Does Domoticz crash? Or just behave different than you expected? How does 'buggy' manifest itself?

[...]

PR #5488, opened 2022-12-21
Enoblinds

efdeb:
Review implementation of EnOcean Blinds Control (EEP D2-05-XX) to conform to EnOcean EEP 2.6.8 protocol and adapt to recents changes on blinds made by @gizmocuz .

Should solve issue Blind + Stop inconsistancies EnOcean #5395 

Create virtual blinds control switches for all blinds channels, using general switch implementation to get a more accurate 0%..100%  control.

Pre-define the correct blinds controls behaviour by setting appropriate switches parameters : 
- Reverse Open/Close State is not set
- Reverse Position is set


gizmocuz:
Thank you for your PR.

- Is there a reason why you specifically use sSwitchTypeAC instead of sSwitchGeneralSwitch ?

- Maybe it is better to 'create' the blinds first (see MQTT Autodiscovery class) and not have two functions named 'SendBlindSwitch'

- You also seem to create inverted blinds by default, the idea is that it seems 'normal' in Domoticz and the event system

In-  your code you invert as well on various places like

SendBlindSwitch(senderID, CHN + 1, gswitch_sSetLevel, 100U - POS

If you do not do the inversion in the code, you also do not have to create a inverted blind as well?

This will also make the code more readable

Maybe also not use 100U but just 100

efdeb:
Thanks for your quick answer.

- I used pTypeGeneralSwitch/sSwitchTypeAC just because I didn't found documentation about general switches subtypes and this seems to be the default value for _tGeneralSwitch class type/subtype => I change to sSwitchGeneralSwitch.

- OK => I rename the first SendBlindSwitch to CreateBlindSwitch

- EnOcean EEP 2.6.8 specifies  0% for an open blind and 100% for a closed one, which is the reverse of domoticz default convention. This is the reason why I have to use 100U - POS in some places, to get the EnOcean blind react as expected by domoticz without needing any end user actions after teaching-in the blind control node.

- The preset of the 'Reverse Position' option is another and unrelated subject. This is not a 'double inversion' but just a 'look & feel' choice for the EnOcean Blind Control widget : it  aims to have the Open state at the left of the slider and the Closed state at the right, which I find more consistent considering the position of the Open/Close buttons. This can be removed from the code if you prefer (or changed later on by the end user if need be).

- Used unsigned value (ex. 100U) because the expression (100U - POS) results in an unsigned uint8_t => change to signed value (ex. 100) to improve readability and let the compiler do the appropriate cast.

Just tell me : I have some time this afternoon !!
;^))

gizmocuz:
Thank you for your answers. The blinds should look/feel like the default blinds in Domoticz.
Best to add a dummy hardware switch, set this to blind percentage, or see the description here:

https://www.domoticz.com/wiki/Blinds

My feeling was a long time, left=close, slider right=open, but this should not be, other HA software also did it the other way

gizmocuz:
position can only be from 0-100, so a uint8_t is fine, but there is no need to use 100U, it complicates the code and is not the style used anywhere (except for some uin64_t maybe 😁)

efdeb:
OK, you decide.
I will reset Reverse Position to default value.

But do not merge the PR Yet since I have a side effect of a concurrent commit done yesterday : Remarked unused code in EnOceanRawValue.h / EnOceanRawValue.cpp.

efdeb:
@gizmocuz
Took into account your remarks, PR is now ready to merge.

Had to revert some changes made yesterday causing EnOcean VLD switches not working anymore.
I shall have a look to this code in a near future.

Thanks

gizmocuz:
Thanks again!

Question, does this mean all switches do not work anymore? Maybe it is best to wait for your changes?

gizmocuz:
Because you now work with default blind behavior, non-inverted by default in Domoticz, I don't think we need the extra CreateBlind function but we can just use the SendBlindSwitch function?

didi31:
@gizmocuz 
> Question, does this mean all switches do not work anymore? Maybe it is best to wait for your changes?

Please consider that the enocean blind are not working any more on the last build and publish this PR as soon as possible even it is not perfect.

efdeb:
> Thanks again!
> 
> Question, does this mean all switches do not work anymore? Maybe it is best to wait for your changes?

With the current development version, EnOcean Pilotwire and Blinds dummy switches created at teach-in won't work anymore. Need roll back EnOceanRawValue.h / EnOceanRawValue.cpp to 3b96ccf or the ones in this PR.

No need to rollback changes to Visual Studio Projects


> Because you now work with default blind behavior, non-inverted by default in Domoticz, I don't think we need the extra CreateBlind function but we can just use the SendBlindSwitch function?

Not sure I can work on it before next monday / tuesday... Maybe possible to merge as is and change next week ? Or just roll back EnOceanRawValue.h / EnOceanRawValue.cpp in the meantime ?

gizmocuz:
Not sure what is wrong with EnOceanRawValue, you are pointing to a change that includes a header, does not seem a big change? 😁

efdeb:
I mean reversing changes concerning EnOceanRawValue.h + EnOceanRawValue.cpp in the 4 commits you made on  Dec 20, 2022 (016d0ab, f36b228, 7bcda7f & 07c4215) and returning for these two files (only) at their statuses on Jan 31, 2022 after 3b96ccf.

gizmocuz:
I did not remove anything from the file, i did commented out some unused functions and did a small cleanup

efdeb:
Don't know yet why... and not much time to investigate before next week, but I confirm **I am sure** **after compiling & testing both versions** that Jan 31, 2022  EnOceanRawValue.h + EnOceanRawValue.cpp code **is working** and Dec 20, 2022 **is no more**.

Maybe moving **stdarg.h** in cpp is wrong, because it defines **va_list** that appears in some functions declarations ?

gizmocuz:
I am not sure about this, in the header I declared the va_list as char*, va_list is a char*
in the .cpp file i cast the char* back to va_list

gizmocuz:
I think it goes wrong here (and yes caused by my patch)

CEnOceanESP3::sendVld (one of them that calls SetRawValues)

I will merge your pr, but when you have time please consider to have a look/change:

- there are many functions with the same name that calls each other
- you are declaring global functions in 'EnOceanRawValue.h' , I will put them all into a namespace (done)
- also take care of what include files you include as most of them are already in stdafx.h
- it looks like there are defines in EnOceanRawValue.h that are not used, if there will be used someday in a .cpp file, please move them there (done)

efdeb:
Thanks !
Not my code these files, but I'll have a look next week.
Merry Christmas !

gizmocuz:
Thanks! Merry Christmass!!

PR #5504, opened 2022-12-29
Improved IP address checking function (used in Proxy header processing)

kiddigital:
This PR adds the `bool IsValidIP(std::string &ip)` function which first validates _AND_ reformats the IP address if needed.

It works for both IPv4 and IPv6 addresses.

This new function is used to validate IP addresses seen in Proxy headers.

### DIFF
--- a/webserver/cWebem.cpp
+++ b/webserver/cWebem.cpp
@@ -1158,6 +1158,54 @@ namespace http {
 			m_session_clean_timer.async_wait([this](auto &&) { CleanSessions(); });
 		}
 
+		static bool isValidIP(std::string &ip)
+		{
+			if (ip.empty())
+				return false;
+
+			std::string cleanIP = stdstring_trimws(ip);
+			bool bIsIPv6 = (cleanIP.find(':') != std::string::npos);
+			if (bIsIPv6)
+			{
+				// IPv6 addresses can be written as quoted strings and between brackets (See RFC5952)
+				if (cleanIP.find('"') != std::string::npos)
+				{
+					cleanIP = cleanIP.substr(1,cleanIP.size()-2);	// Remove quotes from begin and end
+				}
+				if (cleanIP.find('[') != std::string::npos)
+				{
+					cleanIP = cleanIP.substr(1,cleanIP.size()-2);	// Remove brackets from begin and end
+				}
+			}
+		#ifndef WIN32
+			else
+			{
+				// Convert from 'IPv4 numbers-and-dots notation' to 'IPv4 dotted-decimal notation' (or sometimes called: IPv4 dotted-quad notation)
+				struct in_addr addr;
+				if (inet_aton(cleanIP.c_str(), &addr) != 1)
+					return false;
+				char str[INET_ADDRSTRLEN];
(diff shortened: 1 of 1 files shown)


gizmocuz:
We have some build issues under win32:

> C:\projects\domoticz\main\Helper.cpp(662,7): error C3861: 'inet_aton': identifier not found C:\projects\domoticz\main\Helper.cpp(664,12): error C2065: 'INET_ADDRSTRLEN': undeclared identifier C:\projects\domoticz\main\Helper.cpp(665,38): error C2065: 'INET_ADDRSTRLEN': undeclared identifier C:\projects\domoticz\main\Helper.cpp(665,7): error C3861: 'inet_ntop': identifier not found C:\projects\domoticz\main\Helper.cpp(670,6): error C3861: 'inet_pton': identifier not found C:\projects\domoticz\main\Helper.cpp(673,12): error C2065: 'INET6_ADDRSTRLEN': undeclared identifier C:\projects\domoticz\main\Helper.cpp(674,62): error C2065: 'INET6_ADDRSTRLEN': undeclared identifier C:\projects\domoticz\main\Helper.cpp(674,7): error C3861: 'inet_ntop': identifier not found

kiddigital:
hm... some WIN32 specific includes missing it seems. These functions are used at other parts like 'hardware/csocket.cpp' and 'webserver/cWebem.cpp'

Maybe add to 'helper.cpp'?
```
#ifdef WIN32
...
#include <winsock2.h>
#include <Ws2tcpip.h>
#else
...
```

Unfortunately I am not able to test this quickly myself.
Otherwise I might have to _abuse_ the Windows build system to fix it by trail/error :(

gizmocuz:
You are only using the **isValidIP** function in cWebem, please put the function in this class
If you need to test this... make it a static function so you can call it from your testing helper

gizmocuz:
You could also use inet_pton
But still please make this a (static) function of cWebEm as it's only used in this class

kiddigital:
Agree that it is currently only used in the cWebEm class.

Reason to put in into the Helper is because it is intended to be used wider as I see that other hardware modules also do IP checking (some with `inet_pton`, some with `inet_addr` and some even fully custom build) but all in limited form, most IPv4 only (and also only _dotted decimal_, no _numbers and dots_) and no/bad support for IPv6.

I did not undertake any refactoring of those other pieces as I don't have the time and/or ability to test them properly (knowing it compiles is not enough).

gizmocuz:
Yes, that is why i suggested to make the function static so you can use/call it everywhere

gizmocuz:
Could you move the 'bool isValidIP(std::string &ip)' out of the helper header/cpp and into cWebem and make it static so you can use it in your test code if you want to?

kiddigital:
@gizmocuz , moved to cWebem as static function and removed from tester

gizmocuz:
> @gizmocuz , moved to cWebem as static function and removed from tester

Thanks for this. This works (actually it does not have to be static anymore now), but what I actually meant was declare it as static in the header (cWebem.h) inside the class cWebem

then you can access it anywhere you like with

cWebem::IsValidIP(....)

As long as you include cWebem.h

Going to merge, thanks once again!!

PR #5505, opened 2022-12-29
Virtual Thermostat hardware

labelette91580:
Hello : Virtual Thermostat Hardware implementation

I have develop a new hardware device : VirtualThermostat. 
The purpose of this hardware is to act as a virtual thermostat for an electrical radiator using a switch light for 
powering it and a temperature sensor to read room temperature. 
The temperature sensor could be put every where in the room. 
The radiator is power by an electrical switch that is connected to the 220v power or to the pilot wire. 
The room temperature is taken for a standard temperature sensor. 
The VirtualThermostat uses the room temperature and user setpoint temperature to power On/Off the electrical 
radiator . 
The temperature regulation law is not simply On/Off type but a Proportional/integrator regulation law.

Nice if you could merge it !

cdt
see pdf doc
[virtualThermostathardware.pdf](https://github.com/domoticz/domoticz/files/10319174/virtualThermostathardware.pdf)


xbeaudouin:
I love this virtual hardware :D

labelette91580:
bonjour,
merci
vous avez travailler sur le planning timer ?
cdt
De : Xavier Beaudouin ***@***.***>
Envoyé : jeudi 29 décembre 2022 11:45
Objet : Re: [domoticz/domoticz] Virtual Thermostat hardware (PR #5505)


I love this virtual hardware :D

gizmocuz:
Please keep it in English and do not reply via email else all other details are visible here as well

labelette91580:
ok !sorry

kiddigital:
@labelette91580 , I like your idea of the _VirtualThermostat_ device and this 1st implementation of combining an 'on/off switch' and a 'temperature sensor' 👍 

When I have some time, I will also take a look at your proposed code and come back with my 2 cents.

labelette91580:
ok! perfect !

hardware/VirtualThermostat.cpp
```diff
@@ -0,0 +1,740 @@
...
+#include "../main/SQLHelper.h"
+#include "../main/RFXtrx.h"
+
+#include "VirtualThermostat.h"
+
+#include "../main/localtime_r.h"
+#include "../main/mainworker.h"
+#include "../notifications/NotificationHelper.h"
+
+#include <map>
```
gizmocuz: This include (map) should not be needed and can be removed

hardware/VirtualThermostat.cpp
```diff
@@ -0,0 +1,740 @@
...
+{
+	if (fabs(LastValues[index] - value)  > delta)
+	{
+		LastValues[index] = value;
+		return true;
+	}
+	return false;
+}
+
+std::string ToString(double value, const char * format = "%f")
```
gizmocuz: these functions should not be needed, you can use the c++ functions 

**std::to_string**

Or you can use the helper function

**std_format**

hardware/VirtualThermostat.cpp
```diff
@@ -0,0 +1,740 @@
...
+	char buf[64];
+	snprintf(buf, sizeof(buf) - 2, format, value);
+	return buf;
+}
+
+
+//option management
+void getOption(TOptionMap &Option, const char * optionName, int & value)
+{
+	value = atoi(Option[optionName].c_str());
```
gizmocuz: maybe better to use the c++ function std::stoi, std::stol,....

hardware/VirtualThermostat.cpp
```diff
@@ -0,0 +1,740 @@
...
+		m_thread->join();
+		m_thread.reset();
+	}
+	m_bIsStarted = false;
+	return true;
+}
+void VirtualThermostat::Do_Work()
+{
+	int sec_counter = 0;
+	_log.Log(LOG_STATUS, "VirtualThermostat: Worker started...");
```
gizmocuz: you should not call the _log functions, please use **Log** directly (this will also print the hardware name during logging and things can be disabled (info/status/error)

hardware/VirtualThermostat.cpp
```diff
@@ -0,0 +1,740 @@
...
+};                  
+
+//compute the thermostat output switch value 
+// the output is modulated in a time period of 10 minute
+// each minute , the output is activated dependeing time and percent.
+//input : Min : the minute counter 0..59
+//input : PowerPercent : the power modulation  0..100%
+//output 1 : activated
+
+short VirtualThermostat::ComputeThermostatOutput ( int Min , int PowerPercent )
```
gizmocuz: instead of short/long/word could you use the stdint types like uint16_t, uint8_t, uint32_t

hardware/VirtualThermostat.cpp
```diff
@@ -0,0 +1,740 @@
...
+  //get row value for ThermostatOutput
+  PowerPercent = (PowerPercent  ) / MODULATION_STEP ;
+  short switchValue = ThermostatOutput[PowerPercent][Min%MODULATION_DURATION]; 
+  return switchValue;
+}                    
+  
+//return the poqer modulation in function of Room , Exterior and Target Temperature,
+int VirtualThermostat::ComputeThermostatPower ( int index , double RoomTemp , double TargetTemp , double ExteriorTemp , double CoefProportional , double CoefIntegral , double CoefDerivated  )
+{
+    (void)CoefDerivated;
```
gizmocuz: I don't think this line should be here?

hardware/VirtualThermostat.cpp
```diff
@@ -0,0 +1,740 @@
...
+    (void)ExteriorTemp;
```
gizmocuz: I don't think this line should be here?

hardware/VirtualThermostat.cpp
```diff
@@ -0,0 +1,740 @@
...
+		}
+        else
+ 		    _log.Log(LOG_ERROR,"Invalid Switch command : %s ", Cmd.c_str() );
+	}
+}
+
+//get level from imput cmd : selectof value 
+int translateCmd(std::string &Cmd,TOptionMap & options)
+{
+int Level=-1;
```
gizmocuz: Please check the indent of your source code (in visual studio ctrl-k, ctrl-d)

hardware/VirtualThermostat.cpp
```diff
@@ -0,0 +1,740 @@
...
+	Level = GetSelectorSwitchLevel(options, Cmd);
+	if (Level >= 0) { 
+        Cmd = "Set Level " + std::to_string(Level);
+    }
+    return Level;
+}
+void VirtualThermostat::ScheduleThermostat(int Minute )
```
gizmocuz: Please add one empty line between each function

hardware/VirtualThermostat.cpp
```diff
@@ -0,0 +1,740 @@
...
+    int switchtype;
+
+try
+{	
+//select all the Virtuan device
+  TSqlQueryResult result=m_sql.safe_query("SELECT A.Name,A.ID,A.Type,A.SubType,A.nValue,A.sValue, A.Options, A.HardwareID , B.Type,A.DeviceID     FROM DeviceStatus as A LEFT OUTER JOIN Hardware as B ON (B.ID==A.HardwareID) where (B.type == %d )",HTYPE_VirtualThermostat );
+
+	//for all the thermostat switch
+    unsigned int nbDevices     =  result.size() ;
+    double scrollDevice  =  double(MODULATION_DURATION)/nbDevices;
```
gizmocuz: What happens when result is empty? (nbDevices = empty), you get a division by zero, you never know why edits the code in the feature... best to check

hardware/VirtualThermostat.cpp
```diff
@@ -0,0 +1,740 @@
...
+
+	for (unsigned int i=0;i<nbDevices;i++)
```
gizmocuz: better to have nbDevices of type size_t

you can also use

```
for (const auto &itt : result)
{
}

```

no need to check if result is empty

hardware/VirtualThermostat.cpp
```diff
@@ -0,0 +1,740 @@
...
+
+#include "../main/localtime_r.h"
+#include "../main/mainworker.h"
+#include "../notifications/NotificationHelper.h"
+
+#include <map>
+
+std::string GetDeviceValue(const char * FieldName, const char *Idx)
+{
+	TSqlQueryResult result = m_sql.safe_query("SELECT %s from DeviceStatus WHERE (ID == %s )", FieldName, Idx);
```
gizmocuz: we use the 'auto' for these results in our code these days

auto result = m_sql.sa...

hardware/VirtualThermostat.cpp
```diff
@@ -0,0 +1,740 @@
...
+    }
+return TargetTemp;
+}
+
+//return next  thermostat target temperature after Time
+int VirtualThermostat::getNextThermostatProg ( const char * devID , char * CurrentTime , std::string &Time )
+{
+  int TargetTemp=0;
+    TSqlQueryResult result = m_sql.safe_query("SELECT Time,Temperature FROM SetpointTimers where (DeviceRowID==%s) and ( Time > '%s' ) order by time asc limit 1" ,devID, CurrentTime  ) ;
+    if (result.size() ){
```
gizmocuz: instead of result.size(), please use

if (!result.empty())

hardware/VirtualThermostat.cpp
```diff
@@ -0,0 +1,740 @@
...
+        time_t lutime;
+		ParseSQLdatetime(lutime, LastUpdateTime, sLastUpdate);
+
+		result=true;
+	}
+	
+	return result;
+}
+
+#include <stdarg.h>     /* va_list, va_start, va_arg, va_end */
```
gizmocuz: Please do not include this in the middle of the code but at the top

hardware/VirtualThermostat.h
```diff
@@ -0,0 +1,122 @@
+#pragma once
+
+#include <map>
+#include "ThermostatHardware.h"
+
+typedef std::map<int,  float  > T_Map_LastRoomTemp ;
+
+//duratoiopn of integration window in min
+#define INTEGRAL_DURATION 10
```
gizmocuz, line 6: please do not define anything in headers that are not used in the header.
For instance, put them in the .cpp code file

hardware/VirtualThermostat.h
```diff
@@ -0,0 +1,122 @@
+#pragma once
+
+#include <map>
```
gizmocuz: map should not need to be included here, it is already included in stdafx.h

hardware/VirtualThermostat.h
```diff
@@ -0,0 +1,122 @@
...
+	Eco=0,
+	Confor,
+  FrostProtection,
+  Off,
+  EndMode
+};
+
+#define AVALAIBLE_MODE "Eco,Conf,Frost,Off"
+
+class CircularBuffer {
```
gizmocuz: I prefer if you define this class (and LastValue ) underneat the VirtualThermostat class

class VirtualThermostat()
{
private:
  class LastValue()
  {

Or use a namespace in the source/headerfile

Reason, Everywhere this class is included now, we can not use LastValue as it is defined here

main/RFXNames.cpp
```diff
@@ -275,6 +275,7 @@ static const STR_TABLE_SINGLE HardwareTypeTable[] = {
 	{ HTYPE_OctoPrint, "OctoPrint (MQTT/Gina Haussge) with LAN interface", "OctoPrint" },
 	{ HTYPE_Meteorologisk, "Meteorologisk institutt Norway (Weather Lookup)", "Meteorologisk" },
 	{ HTYPE_AirconWithMe, "AirconWithMe Wifi Airco module", "AirconWithMe" },
+	{ HTYPE_VirtualThermostat, "Virtual Thermostat"							,       "VirtualThermostat" },
```
gizmocuz: Please check the indent, it is a space, not 100 spaces ;)

main/SQLHelper.cpp
```diff
@@ -5972,9 +6021,29 @@ void CSQLHelper::UpdateTemperatureLog()
 			{
 			case pTypeRego6XXTemp:
 			case pTypeTEMP:
-			case pTypeThermostat:
 				temp = static_cast<float>(atof(splitresults[0].c_str()));
 				break;
+			case pTypeThermostat:
+				{
+
+					//if not a virtual thermostat
```
gizmocuz, line 6089: better check if hardware type is the virtual thermostat, you never know in the feature each thermostat has options

main/WebServer.cpp
```diff
@@ -68,6 +68,8 @@
 #define __STDC_FORMAT_MACROS
 #include <inttypes.h>
 
+#include "../hardware/VirtualThermostat.h"
```
gizmocuz: Do we need this here?

main/mainworker.cpp
```diff
@@ -13095,6 +13103,11 @@ bool MainWorker::SetThermostatState(const std::string& idx, const int newState)
 		pGateway->SetProgramState(newState);
 		return true;
 	}
+	else if (pHardware->HwdType == HTYPE_VirtualThermostat)
+	//virtual thermostat set state confor/eco/off/frozen
```
gizmocuz: Please put the comment inside the { } block

msbuild/domoticz.vcxproj
```diff
@@ -29,27 +29,27 @@
     <ConfigurationType>Application</ConfigurationType>
     <UseDebugLibraries>true</UseDebugLibraries>
     <CharacterSet>MultiByte</CharacterSet>
-    <PlatformToolset>v143</PlatformToolset>
+    <PlatformToolset>v142</PlatformToolset>
```
gizmocuz: Please undo these changes, better to undo all in this file
Also upgrade your visual studio ;)

gizmocuz (changes requested):
Thanks for this PR. Just some comments/suggestions/requests

gizmocuz:
@labelette91580 , thank you for this PR. I made a quick review.
Question, this virtualthemostat could also have been done with dzVents?

labelette91580:
@gizmocuz : i don't know  dzVents ?

gizmocuz:
@labelette91580 , It is the Domoticz internal scripting language, worth having a look at this.
You can do all logic in here (Proportional/integrator regulation law)

hardware/VirtualThermostat.cpp
labelette91580: ok done

hardware/VirtualThermostat.cpp
labelette91580: ok for std::string ToString(int value

for using std_format , i have vs2019 on windows 7 and std_format is not available

hardware/VirtualThermostat.cpp
labelette91580: ok done

hardware/VirtualThermostat.cpp
labelette91580: ok. done also for Debug.

hardware/VirtualThermostat.cpp
labelette91580: ok use int

main/RFXNames.cpp
labelette91580: ok del tab

main/SQLHelper.cpp
labelette91580: ok! check hardware type

main/WebServer.cpp
labelette91580: yes ok ! done

main/mainworker.cpp
labelette91580: ok done

msbuild/domoticz.vcxproj
labelette91580: ok ! done ! i was on Winsows 7 so VS2019

hardware/VirtualThermostat.cpp
labelette91580: it is to avoid unused argument warning

hardware/VirtualThermostat.cpp
labelette91580: it is to avoid unused argument warning

hardware/VirtualThermostat.cpp
labelette91580: ok  done

hardware/VirtualThermostat.cpp
labelette91580: ok done

hardware/VirtualThermostat.cpp
labelette91580: yes right ! but it is floating division and scrollDevice take inf value ! so it work ok 
but i move line

hardware/VirtualThermostat.cpp
labelette91580: ok ! done ! new for me .

hardware/VirtualThermostat.cpp
gizmocuz: std_format is a internal Domoticz function that is declared in the helper class... very usefull instead of using sprintf

hardware/VirtualThermostat.cpp
gizmocuz: Looking at the code, the result is a 10 bit value, so uint16_t will do?

hardware/VirtualThermostat.cpp
gizmocuz: Thats an option... the other option is to remove these arguments.. 😁

msbuild/domoticz.vcxproj
gizmocuz: Windows 7 is deprecated as well as 8.1, better move to Windows 11, pretty all right once you get used to it (and stay of the inner circle development releases... 😁)

hardware/VirtualThermostat.cpp
labelette91580: ok ! done

hardware/VirtualThermostat.cpp
labelette91580: ok done

hardware/VirtualThermostat.h
labelette91580: ok done ! this was use in previous version

hardware/VirtualThermostat.h
labelette91580: ok done

hardware/VirtualThermostat.h
labelette91580: ok done

hardware/VirtualThermostat.cpp
labelette91580: ok : done

gizmocuz (approved):
All code changes look good, thank you for this!

I do have some remarks about indenting.
You are using Visual Studio correct? If you press control-k and then control-d , it will format the code
Could you do this in both the header and source files you changed/created?

for example

image

waltervl:
Is this somehow related to the plugin Smart Virtual Thermostat? https://www.domoticz.com/wiki/Plugins/Smart_Virtual_Thermostat.html and https://www.domoticz.com/forum/viewtopic.php?t=19418

labelette91580:
i think it is due to the tab programmation in text editor VS : i use 4 space for Tab .
What i a have to use Tab au space ? for all c++/html/js files ?

labelette91580:
in fact i have begin this developemnt before this python plugin exist ! 
and i think that native hardware is perhaps more stable than external plugin.

labelette91580:
i am also woking on the nativ implementation of Imperihome ISS Api !

waltervl:
> in fact i have begin this developemnt before this python plugin exist ! and i think that native hardware is perhaps more stable than external plugin.

I do not mind, just wondering if it has the same functionality. 
If added in a beta also the Wiki should be updated..
Reading the attached PDF I am wondering: 
There should already be a Domoticz Temperature sensor and an On/Off switch device or Selector Switch device present? These could be python plugin, Zwave, RFLink etc controlled devices? 

If a user wants to change the temperature device connected to the Virtual Thermostat how to do this? Edit device?

And what is Power (%)? How is it related to the On/Off or selector switch?

labelette91580:
There should already be a Domoticz Temperature sensor and an On/Off switch device or Selector Switch device present? These could be python plugin, Zwave, RFLink etc controlled devices?
--> the sensor should exist before switch and temperature sensor : 
when you create the virtual thermostat device, you must choose one in combo box :
    Then you have to assign a temperature Sensor and Switch. 
    Select the temperature sensor in Temperature Combo box. 
    Select the Switch actuator Switch Combo. 

to change temperature : it is like for setpoint : click on icon ! 
in the pdf : To edit setPoint , click on image :

The power is the computation resultat : 
0% the radiator will never been switch on 
10% the radiator will  been switch  on during 1 minutes / 10 min
50% the radiator will  been switch  on during 5 minutes / 10 min
100% the radiator will been switch  on during 10 minutes / 10 min

CMakeLists.txt
```diff
@@ -186,6 +186,7 @@ hardware/AnnaThermostat.cpp
 hardware/Arilux.cpp
 hardware/ASyncSerial.cpp
 hardware/ASyncTCP.cpp
+hardware/VirtualThermostat.cpp
```
keryell: Sort.

hardware/ThermostatHardware.h
```diff
@@ -0,0 +1,53 @@
+#pragma once
+
+#include <iostream>
+#include "DomoticzHardware.h"
+
+class CThermostatHardware : public CDomoticzHardwareBase
+{
+public:
+
+	//thermostat available mode in string "OFF,ECO,CONFOR,AUTO," separated by comma,
```
keryell: Use English instead of French abbreviation?
```suggestion
	//thermostat available mode in string "OFF,ECO,COMFORT,AUTO," separated by comma, 
```

hardware/ThermostatHardware.h
```diff
@@ -0,0 +1,53 @@
...
+#include <iostream>
+#include "DomoticzHardware.h"
+
+class CThermostatHardware : public CDomoticzHardwareBase
+{
+public:
+
+	//thermostat available mode in string "OFF,ECO,CONFOR,AUTO," separated by comma, 
+	// and in the order from 0 to N and 0 = OFF 1 = ECO, 2 = CONFOR 3 = AUTO
+	std::string  AvailableMode;
```
keryell:
```suggestion
	std::string AvailableMode;
```
Use a code formatter for new code

hardware/ThermostatHardware.h
```diff
@@ -0,0 +1,53 @@
...
+
+	//thermostat function
+	//return the thermostat available mode in string "OFF,ECO,CONFOR,AUTO," separated by comma , and 0=OFF 1=ECO, 2=CONFOR 3=AUTO
+	virtual std::string GetAvailableMode() { return AvailableMode; };
+	virtual void        SetAvailableMode(std::string  pAvailableMode) { AvailableMode = pAvailableMode; };
```
keryell:
```suggestion
	virtual void SetAvailableMode(std::string pAvailableMode) { AvailableMode = pAvailableMode; };
```
Just curious about why they are not `const std::string&` everywhere?

hardware/ThermostatHardware.h
```diff
@@ -0,0 +1,53 @@
...
+	int  ThermostatModeStringToInt(std::string& mode)
+	{
+		std::vector<std::string> ModeStr;
+		StringSplit(AvailableMode, ",", ModeStr);
+
+		for (int i = 0; i < (int)ModeStr.size(); i++)
+			if (strcmp(mode.c_str(), ModeStr[i].c_str()) == 0)
+				return  i;
+		return 0;
+
```
keryell: I have the feeling that what you mean is something like
```suggestion
		return (std::find(ModeStr.begin(), ModeStr.end(), mode) - ModeStr.begin()) % ModeStr.size();
```
https://en.cppreference.com/w/cpp/algorithm/find

hardware/ThermostatHardware.h
```diff
@@ -0,0 +1,53 @@
...
+	}
+	//convert interger thermostat state to string state : 0--> OFF 1-->ECO
```
keryell: Life is easier when using a spell-checker :-)
```suggestion
	//convert integer thermostat state to string state : 0--> OFF 1-->ECO
```

hardware/ThermostatHardware.h
```diff
@@ -0,0 +1,53 @@
...
+	std::string   ThermostatModeIntToString(int Mode)
+	{
+		std::vector<std::string> ModeStr;
+		StringSplit(AvailableMode, ",", ModeStr);
+		if (Mode < (int)ModeStr.size())
+			return ModeStr[Mode];
+		else
+			return "UNKNOWN";
```
keryell: I guess for negative `Mode` you will get into trouble too.
Otherwise there is `ModeStr.at(Mode)` if you are OK with exceptions, but I do not know if it fits the design.

hardware/ThermostatHardware.h
```diff
@@ -0,0 +1,53 @@
...
+			if (strcmp(mode.c_str(), ModeStr[i].c_str()) == 0)
+				return  i;
+		return 0;
+
+	}
+	//convert interger thermostat state to string state : 0--> OFF 1-->ECO
+	std::string   ThermostatModeIntToString(int Mode)
+	{
+		std::vector<std::string> ModeStr;
+		StringSplit(AvailableMode, ",", ModeStr);
```
keryell: Is it OK to split every time? How often do `AvailableMode` change?

hardware/VirtualThermostat.cpp
```diff
@@ -0,0 +1,715 @@
...
+
+void VirtualThermostat::CircularBuffer::Clear()
+{
+	for (int i = 0; i < Size; i++)Value[i] = 0;
+	index = 0;
+	Sum = 0;
+}
+VirtualThermostat::CircularBuffer::CircularBuffer(int pSize)
+{
+	Value = new double[pSize];
```
keryell: Do not use raw memory https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rr-newdelete
Use containers like `std::vector` instead.
But at the end, why not using something like https://www.boost.org/doc/libs/1_81_0/doc/html/circular_buffer.html instead of reinventing it?

hardware/VirtualThermostat.cpp
```diff
@@ -0,0 +1,715 @@
...
+		if (m_stoprequested)
+			break;
+		sec_counter++;
+
+		if (sec_counter % 12 == 0) {
+			m_LastHeartbeat = mytime(NULL);
+		}
+		time_t atime = mytime(NULL);
+		struct tm ltime;
+		localtime_r(&atime, &ltime);
```
keryell, line 170: Use C++ `chrono` library instead.
https://en.cppreference.com/w/cpp/header/chrono

hardware/VirtualThermostat.cpp
```diff
@@ -0,0 +1,715 @@
...
+{
+	return true;
+}
+
+//time for the power time modulation in minute
+#define MODULATION_DURATION 10
+//  number of step in percent %
+#define MODULATION_STEP     10
+
+//guve the output power swicth value in the time from the modulation percent
```
keryell:
```suggestion
//give the output power switch value in the time from the modulation percent
```
Are there boolean values otherwise?

hardware/VirtualThermostat.cpp
```diff
@@ -0,0 +1,715 @@
...
+	}
+	Log(LOG_STATUS, "VirtualThermostat: Worker stopped...");
+}
+bool VirtualThermostat::WriteToHardware(const char* pdata, const unsigned char length)
+{
+	return true;
+}
+
+//time for the power time modulation in minute
+#define MODULATION_DURATION 10
```
keryell: Replace macros by `constexpr` variables when possible.
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-macros2

hardware/VirtualThermostat.cpp
```diff
@@ -0,0 +1,715 @@
...
+//input : Min : the minute counter 0..59
+//input : PowerPercent : the power modulation  0..100%
+//output 1 : activated
+
+int VirtualThermostat::ComputeThermostatOutput(int Min, int PowerPercent)
+{
+	if (PowerPercent > 100) PowerPercent = 100;
+	//get row value for ThermostatOutput
+	PowerPercent = (PowerPercent) / MODULATION_STEP;
+	int switchValue = ThermostatOutput[PowerPercent][Min % MODULATION_DURATION];
```
keryell: I have the feeling that you can compute the output directly instead of enumerating everything in `ThermostatOutput`.

hardware/VirtualThermostat.cpp
```diff
@@ -0,0 +1,715 @@
...
+	return switchValue;
+}
+
+//return the poqer modulation in function of Room , Exterior and Target Temperature,
```
keryell:
```suggestion
//return the power modulation in function of Room , Exterior and Target Temperature,
```

hardware/VirtualThermostat.cpp
```diff
@@ -0,0 +1,715 @@
...
+		//PID regulation
+		// coef integral is the sum of delta temperature since last 10 minutes.
+		PowerModulation = int(DeltaTemp * CoefProportional + Delta->GetSum() * CoefIntegral / INTEGRAL_DURATION);
+	}
+	else
+		Delta->Clear(); //clear integral part
+
+	if (PowerModulation > 100) PowerModulation = 100;
+	if (PowerModulation < 0) PowerModulation = 0;
+	return PowerModulation;
```
keryell, line 245: https://en.cppreference.com/w/cpp/algorithm/clamp

keryell (commented):
Thanks! I was just looking for such a thing. :-)

CMakeLists.txt
labelette91580: ok done

hardware/ThermostatHardware.h
labelette91580: ok ! which when using a spell-checker  you use ?

hardware/ThermostatHardware.h
labelette91580: ok ! done ! set Mode as unsigned

hardware/ThermostatHardware.h
labelette91580: no AvailableMode  is constant for a thermostat instance.
this function is used for display

hardware/VirtualThermostat.cpp
labelette91580: ok. but i dont know boost library !

hardware/VirtualThermostat.cpp
labelette91580: i reused code from void CEventSystem::Do_Work() !

hardware/VirtualThermostat.cpp
labelette91580: ok.
Yes

hardware/VirtualThermostat.cpp
labelette91580: yes ! you are right !

hardware/VirtualThermostat.cpp
labelette91580: ok done

hardware/VirtualThermostat.cpp
labelette91580: ok ! done . i dont know

hardware/VirtualThermostat.cpp
labelette91580: i will do

hardware/VirtualThermostat.cpp
labelette91580: the allocation is done one time during reset and is not freed after
so it is more effecient than the std::vector that the allocation  is done during runtime

keryell: I guess you would create a vector only once with all its elements initialized to 0 with
```c++
std::vector<double> Value(pSize);
```

hardware/ThermostatHardware.h
keryell: The spell-checker depends on your IDE, which I do not know.
Just in case, I use ```(add-hook 'prog-mode-hook 'flyspell-prog-mode)``` with Emacs.

gizmocuz:
Is it possible you can also include the header file in the 'VirtualThermostat' filter in visual studio?
At the moment I only see the cpp file

image


Edit: Also ThermostatHardware.h, or declare this code inside your VirtualThermostat header so this file is not needed

gizmocuz:
std::clamp is not available in C++14

hardware/ThermostatHardware.h
```diff
@@ -0,0 +1,48 @@
...
+#include <iostream>
+#include "DomoticzHardware.h"
+
+class CThermostatHardware : public CDomoticzHardwareBase
+{
+public:
+
+	//thermostat available mode in string "OFF,ECO,COMFORT,AUTO," separated by comma, 
+	// and in the order from 0 to N and 0 = OFF 1 = ECO, 2 = CONFOR 3 = AUTO
+	std::string AvailableMode;
```
gizmocuz: Could you make member variables start with 'm_'
I think this variable and other functions below should be plural (m_AvailableMode, GetAvailableModes, ...)

gizmocuz (changes requested):

keryell:
> std::clamp is not available in C++14

Ah yes, good point. Happy new year 2023, the year of C++23. :-)
Any idea about when Domoticz could jump to C++17 (now 6 year old)? It looks like for example the RaspberryPi supports most of C++20 now.

gizmocuz:
Well not so long ago we jumped to C++14 .... 
I don't think the default raspbian has support for this
And I think we should support the default buster users as well
Of course with funky tricks you can install the latest and greatest gcc
But it is easier to recode the clamp function

hardware/VirtualThermostat.cpp
labelette91580: std::clamp  not compile on raspbery pi : c++17 !

hardware/VirtualThermostat.cpp
labelette91580: ok done

hardware/ThermostatHardware.h
labelette91580: ok !done

hardware/VirtualThermostat.cpp
labelette91580: ok ! use circular buffer

hardware/VirtualThermostat.cpp
labelette91580: ok done

labelette91580:
@kyrell : i have use the boost::circular_buffer template but it does not compile on windows  
the reason in : the header is not present in the msbuild\Windows Libraries\include\boost.
but it is present in vcpkg\installed\x86-windows\include\boost 

in local i have delete Windows Libraries\include\boost directory.

so i think you have to delete boost include from windowslibrairy.zip because only open zwave is used ?

gizmocuz:
Please avoid any boost dependencies

labelette91580:
do you have a domoticz test database with all possible devices created ? for testing ?

labelette91580:
hello : do you expect anything else from me ?

kiddigital:
Hi @labelette91580 , not really at the moment.

We are looking into it and discussing if/how we can incorporate your idea of virtual hardware and how to implement it.

Just need some time to have a good look and work out some details.

Sorry it takes some time (work is always in the way of hobbies 😁)

labelette91580:
ok ! no problem ! thanks

keryell:
> Please avoid any boost dependencies

On the other hand, Boost is already used by Domoticz, so it does not really add a new dependency. Between using Boost or reimplementing some existing stuff with some possible bugs and technical debt, there is obviously some trade-off to be found. :-)

gizmocuz:
We are trying to limit its usage. It is mostly used for Asio stuff. Best to find a default std way

main/SQLHelper.h
```diff
@@ -321,21 +321,23 @@ typedef std::vector<std::string> TSqlRowQuery;
 // result for an sql query : Vector of TSqlRowQuery
 typedef std::vector<TSqlRowQuery> TSqlQueryResult;
 
+typedef   std::map<std::string, std::string> TOptionMap;
+
```
keryell, line 325: You should use
```suggestion
using TOptionMap = std::map<std::string, std::string>;
```
in new code https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rt-using

hardware/ThermostatHardware.h
```diff
@@ -0,0 +1,50 @@
...
+
+#include <iostream>
+#include "DomoticzHardware.h"
+
+class CThermostatHardware : public CDomoticzHardwareBase
+{
+private:
+
+	//thermostat available mode in string "OFF,ECO,COMFORT,AUTO," separated by comma, 
+	// and in the order from 0 to N and 0 = OFF 1 = ECO, 2 = CONFORT 3 = AUTO
```
keryell: Use English instead of French
```suggestion
	// and in the order from 0 to N and 0 = OFF 1 = ECO, 2 = COMFORT 3 = AUTO
```

main/SQLHelper.h
```diff
@@ -344,140 +346,142 @@ class CSQLStatement
...
+		std::string& devname, const bool bUseOnOffAction, const char* User = nullptr);
+	uint64_t UpdateValue(int HardwareID, const char* ID, unsigned char unit, unsigned char devType, unsigned char subType, unsigned char signallevel, unsigned char batterylevel,
+		const char* sValue, std::string& devname, const bool bUseOnOffAction, const char* User = nullptr);
+	uint64_t UpdateValue(int HardwareID, const char* ID, unsigned char unit, unsigned char devType, unsigned char subType, unsigned char signallevel, unsigned char batterylevel, int nValue,
+		const char* sValue, std::string& devname, const bool bUseOnOffAction, const char* User = nullptr);
+	uint64_t UpdateValueLighting2GroupCmd(int HardwareID, const char* ID, unsigned char unit, unsigned char devType, unsigned char subType, unsigned char signallevel, unsigned char batterylevel,
+		int nValue, const char* sValue, std::string& devname, const bool bUseOnOffAction, const char* User = nullptr);
+	uint64_t UpdateValueHomeConfortGroupCmd(int HardwareID, const char* ID, unsigned char unit, unsigned char devType, unsigned char subType, unsigned char signallevel, unsigned char batterylevel,
```
keryell: You could use a global substitution
```suggestion
	uint64_t UpdateValueHomeComfortGroupCmd(int HardwareID, const char* ID, unsigned char unit, unsigned char devType, unsigned char subType, unsigned char signallevel, unsigned char batterylevel,
```

main/mainworker.cpp
```diff
@@ -13090,6 +13098,11 @@ bool MainWorker::SetThermostatState(const std::string& idx, const int newState)
 		pGateway->SetProgramState(newState);
 		return true;
 	}
+	else if (pHardware->HwdType == HTYPE_VirtualThermostat)
+	{
+		//virtual thermostat set state confor/eco/off/frozen
```
keryell:
```suggestion
		//virtual thermostat set state comfort/eco/off/frozen
```
etc.

main/SQLHelper.h
labelette91580: ok done

main/SQLHelper.h
labelette91580: i have not developped this function ?
do you want that i rename this function?

labelette91580: ok rename done

gizmocuz:
I still think this all could be done via a nice dzVents script

labelette91580:
ok. perhaps.
but for the end user , it will be more easy to create rapidly  many hardware through  an ihm
than create many dzVents script with own configuration.
mainly if you  don't know dzEvent langage or no programming knowledge.



cdt

De : gizmocuz ***@***.***>
Envoyé : mardi 14 février 2023 10:57
Cc : labelette91580 ***@***.***>; Mention ***@***.***>
Objet : Re: [domoticz/domoticz] Virtual Thermostat hardware (PR #5505)


I still this all could be done via a nice dzVents script

gizmocuz:
Well, I mean, you would copy/paste the dzVents script and change some parameters at the top and the user is ready.
This way it is a much more straight forward way instead of making new devices and code in Domoticz

labelette91580:
yes !
in fact, I have begin this development 5 year ago when there is no python or dzevent plugins in domoticz ! only LUA .
So i decide to do this in code c (more easy to develop and test than LUA).
It work fine for me since.
Now , effectively, it could be done otherwise .
I Just to propose to the other user something "plug & play" working  without any programation.

cdt


De : gizmocuz ***@***.***>
Envoyé : mercredi 15 février 2023 10:48
Cc : labelette91580 ***@***.***>; Mention ***@***.***>
Objet : Re: [domoticz/domoticz] Virtual Thermostat hardware (PR #5505)


Well, I mean, you would copy/paste the dzVents script and change some parameters at the top and the user is ready.
This way it is a much more straight forward way instead of making new devices and code in Domoticz

hardware/ThermostatHardware.h
```diff
@@ -0,0 +1,50 @@
...
+		std::vector<std::string> ModeStr;
+		StringSplit(m_AvailableMode, ",", ModeStr);
+		return (std::find(ModeStr.begin(), ModeStr.end(), mode) - ModeStr.begin()) % ModeStr.size();
+	}
+	//convert integer thermostat state to string state : 0--> OFF 1-->ECO
+	std::string   ThermostatModeIntToString(unsigned int Mode)
+	{
+		std::vector<std::string> ModeStr;
+		StringSplit(m_AvailableMode, ",", ModeStr);
+		if (Mode < (int)ModeStr.size())
```
MilanFranc: Variable 'Mode' is type unsigned int, I suggest make type-cast to the same type -> unsigned int.
Also I would make it: `static_cast<unsigned int>(ModeStr.size())`

hardware/VirtualThermostat.cpp
```diff
@@ -0,0 +1,682 @@
...
+	{
+		Sum += m_Value[i];
+	}
+	return Sum;
+}
+
+//option management
+void getOption(TOptionMap& Option, const char* optionName, int& value)
+{
+	value = std::stoi(Option[optionName].c_str());
```
MilanFranc: function std::stoi or std::stol can rise an std::exception. Is that ok?  and std::stoi can work with std::string

hardware/VirtualThermostat.cpp
```diff
@@ -0,0 +1,682 @@
...
+	else
+		return  "";
+}
+
+void VirtualThermostat::CircularBuffer::Clear()
+{
+	for (int i = 0; i < m_Size; i++)m_Value[i] = 0;
+	m_index = 0;
+}
+VirtualThermostat::CircularBuffer::CircularBuffer(int pSize)
```
MilanFranc: `unsigned int` would be better here and also in member variables and for cycles.

hardware/ThermostatHardware.h
```diff
@@ -0,0 +1,50 @@
...
+	std::string m_AvailableMode;
+
+public:
+
+	//thermostat function
+	//return the thermostat available mode in string separated by comma 
+	virtual std::string& GetAvailableMode() { return m_AvailableMode; };
+	virtual void SetAvailableMode(std::string& pAvailableMode) { m_AvailableMode = pAvailableMode; };
+	//return the thermostat mode 
+	virtual std::string GetCurrentMode(std::string& devIdx) = 0;
```
MilanFranc: This string can be changed? If not, please make it `const std::string&`

hardware/ThermostatHardware.h
```diff
@@ -0,0 +1,50 @@
...
+	//return the thermostat room temperature 
+	virtual std::string GetRoomTemperature(std::string& devIdx) = 0;
+	//return the thermostat current setpoint 
+	virtual std::string GetSetPoint(std::string& devIdx) = 0;
+	//set the thermostat mode : newState = integer thermostat mode , in the order in AvailableMode
+	virtual bool SetThermostatState(const std::string& devIdx, const int newState) = 0;
+
+	//convert thermostat string state to int state : OFF-->0  ECO-->1
+	unsigned int  ThermostatModeStringToInt(std::string& mode)
```
MilanFranc: please `const std::string& `

main/SQLHelper.cpp
```diff
@@ -5632,6 +5665,22 @@ bool CSQLHelper::GetLastValue(const int HardwareID, const char* DeviceID, const
...
+	std::vector<std::vector<std::string> > sqlresult;
+
+    sqlresult = safe_query("SELECT %s FROM DeviceStatus WHERE ( ID= %s )" , FieldName, Idx.c_str() );
+
+	if (!sqlresult.empty())
+	{
+		return sqlresult[0][0] ;
+	}
+
+	return "" ;
```
MilanFranc: Much better (cleaner) were `return std::string();`

hardware/ThermostatHardware.h
```diff
@@ -0,0 +1,50 @@
+#pragma once
+
+#include <iostream>
```
MilanFranc: I don't see 'iostream' objects in this header files. But I'm missing string, vector, algorithm (std::find) includes

hardware/ThermostatHardware.h
```diff
@@ -0,0 +1,50 @@
...
+	//thermostat available mode in string "OFF,ECO,COMFORT,AUTO," separated by comma, 
+	// and in the order from 0 to N and 0 = OFF 1 = ECO, 2 = COMFORT 3 = AUTO
+	std::string m_AvailableMode;
+
+public:
+
+	//thermostat function
+	//return the thermostat available mode in string separated by comma 
+	virtual std::string& GetAvailableMode() { return m_AvailableMode; };
+	virtual void SetAvailableMode(std::string& pAvailableMode) { m_AvailableMode = pAvailableMode; };
```
MilanFranc: please `const std::string&`

hardware/VirtualThermostat.h
```diff
@@ -0,0 +1,92 @@
...
+
+	//set the thermostat mode 
+	virtual bool SetThermostatState(const std::string& deviceIdx, const int newState);
+	//convert interger state to string state : 0--> OFF 1-->ECO
+
+	bool StartHardware();
+	bool StopHardware();
+	void Do_Work();
+
+	std::shared_ptr<std::thread> m_thread;
```
MilanFranc, line 80: Is there a reason that this thread and other member variables are in public section ? If not please put it in private section of the class.

main/SQLHelper.cpp
gizmocuz: in case of sql queries, we use 'auto' in domoticz, makes it easier

`auto sqlresult = safe_query("SELECT %s FROM DeviceStatus WHERE ( ID= %s )" , FieldName, Idx.c_str() );`

hardware/VirtualThermostat.cpp
```diff
@@ -0,0 +1,682 @@
...
+
+#include "../main/localtime_r.h"
+#include "../main/mainworker.h"
+#include "../notifications/NotificationHelper.h"
+#include <stdarg.h>     
+
+std::string AVALAIBLE_MODE = "Eco,Conf,Frost,Off";
+
+std::string GetDeviceValue(const char* FieldName, const char* Idx)
+{
```
gizmocuz: Seems this function is not used as you made a new function in SQLHelper.cpp

main/SQLHelper.cpp
```diff
@@ -5632,6 +5665,22 @@ bool CSQLHelper::GetLastValue(const int HardwareID, const char* DeviceID, const
 	return result;
 }
 
+std::string  CSQLHelper::GetDeviceValue(const char* FieldName, const std::string& Idx)
```
gizmocuz: why using char* here and not std::string?

hardware/VirtualThermostat.cpp
```diff
@@ -0,0 +1,682 @@
...
+//return the thermostat room temperature 
+std::string VirtualThermostat::GetRoomTemperature(std::string& devIdx)
+{
+	std::string temp = GetOptionFromDeviceId(devIdx, "RoomTemp");
+	return (temp);
+};
+
+std::string VirtualThermostat::GetSetPoint(std::string& devIdx)
+{
+	return (m_sql.GetDeviceValue("svalue", devIdx.c_str()).c_str());
```
gizmocuz: you do not to use .c_str() here on both places

hardware/VirtualThermostat.cpp
```diff
@@ -0,0 +1,682 @@
...
+}
+
+float VirtualThermostat::GetSetPointTemperature(std::string& devIdx)
+{
+	return ((float)std::stof(GetSetPoint(devIdx).c_str()));
```
gizmocuz: you do not need to use .c_str() here

hardware/VirtualThermostat.cpp
```diff
@@ -0,0 +1,682 @@
...
+}
+//return temperature value from Svalue : is code temperature;humidity;???
+float VirtualThermostat::getTemperatureFromSValue(const char* sValue)
+{
+	std::vector<std::string> splitresults;
+	StringSplit(sValue, ";", splitresults);
+	if (splitresults.size() < 1)
+		return 0;
+	else
+		return (float)std::stof(splitresults[0].c_str());
```
gizmocuz: You do not need .c_str() here and also the old_school cast to float is not needed

```
    std::string myFloatStr = "2.34";
    float my_float = std::stof(myFloatStr);

```

hardware/VirtualThermostat.cpp
```diff
@@ -0,0 +1,682 @@
...
+	time_t now = mytime(NULL);
+	struct tm tm1;
+	localtime_r(&now, &tm1);
+
+	sprintf(szTmp, "SELECT nValue,sValue,LastUpdate FROM DeviceStatus WHERE ( ID=%s ) ", DeviceID);
+	auto sqlresult = m_sql.safe_query(szTmp);
+
+	if (!sqlresult.empty())
+	{
+		nValue = (int)std::stoi(sqlresult[0][0].c_str());
```
gizmocuz: you do not need .c_str() here, also the cast to (int) is not needed

main/SQLHelper.cpp
```diff
@@ -6017,9 +6066,32 @@ void CSQLHelper::UpdateTemperatureLog()
...
+			case pTypeThermostat:
+				{
+
+					const std::string& HwdId = sd[7];
+					//get virtual thermostat hardware
+					CDomoticzHardwareBase* Hardware = m_mainworker.GetHardwareByIDType( HwdId, HTYPE_VirtualThermostat);
+					//if not a virtual thermostat
+					if (Hardware==nullptr)
+					{
+						temp = static_cast<float>(std::stof(splitresults[0].c_str()));
```
gizmocuz: you do not need .c_str() here, also the cast to float is not needed (std::stof is returning a float)

main/SQLHelper.cpp
```diff
@@ -6017,9 +6066,32 @@ void CSQLHelper::UpdateTemperatureLog()
...
+					}
+					else
+					{
+						//for virtual thermostat record Room Temperature / set point & power level 0..100%
+						//set point temperature record as chill
+						std::string Options  = sd[6];
+						chill = static_cast<float>(std::stof(splitresults[0].c_str()));
```
gizmocuz: you do not need .c_str() here, also the cast to float is not needed (std::stof is returning a float)

main/SQLHelper.cpp
```diff
@@ -6017,9 +6066,32 @@ void CSQLHelper::UpdateTemperatureLog()
...
+						//record power % as humidity
+						humidity = std::stoi(VirtualThermostatGetOption("Power", Options).c_str());
```
gizmocuz: you do not need .c_str() here, also the cast to int is not needed (std::stoi is returning a integer)

main/SQLHelper.cpp
```diff
@@ -6017,9 +6066,32 @@ void CSQLHelper::UpdateTemperatureLog()
...
+						//record room temp 
+						temp = (float)std::stof(VirtualThermostatGetOption("RoomTemp", Options).c_str());
```
gizmocuz: you do not need .c_str() here, also the cast to float is not needed (std::stof is returning a float)

main/SQLHelper.cpp
```diff
@@ -5632,6 +5665,22 @@ bool CSQLHelper::GetLastValue(const int HardwareID, const char* DeviceID, const
 	return result;
 }
 
+std::string  CSQLHelper::GetDeviceValue(const char* FieldName, const std::string& Idx)
+{
+	bool result = false;
+	std::vector<std::vector<std::string> > sqlresult;
+
+    sqlresult = safe_query("SELECT %s FROM DeviceStatus WHERE ( ID= %s )" , FieldName, Idx.c_str() );
```
gizmocuz: please use the 'auto' keyword instead of declaring the vector, can also be a online liner

main/WebServer.cpp
```diff
@@ -10560,6 +10568,20 @@ namespace http
...
 						root["result"][ii]["HaveTimeout"] = bHaveTimeout;
 						root["result"][ii]["TypeImg"] = "override_mini";
+
+						if (_hardwareNames[hardwareID].HardwareTypeVal == HTYPE_VirtualThermostat)
+						{
+							TOptionMap optionsMap = m_sql.BuildDeviceOptions(sOptions);
+							//build option values
+							for (const auto& itt : optionsMap)
+							{
+								std::string optionName = itt.first.c_str();
```
gizmocuz: please remove the .c_str()

main/WebServer.cpp
```diff
@@ -10560,6 +10568,20 @@ namespace http
...
+								std::string optionValue = itt.second;
+								root["result"][ii][optionName.c_str()] = optionValue.c_str();
```
gizmocuz: please remove the .c_str() on both places

main/WebServer.cpp
```diff
@@ -12964,7 +12986,8 @@ namespace http
 
 			if (!devoptions.empty())
 			{
-				m_sql.safe_query("UPDATE DeviceStatus SET Options='%q' WHERE (ID == '%q')", devoptions.c_str(), idx.c_str());
+				uint64_t ullidx = std::stoull(idx);
+				m_sql.SetDeviceOptions(ullidx, m_sql.BuildDeviceOptions(devoptions.c_str(), false));
```
gizmocuz: please remove the .c_str() the function takes a normal std::string

main/SQLHelper.cpp
MilanFranc: return type of this function is 'std::string' also I sugested to use that type. Has nothing to do with auto..

keryell: Do it the modern and efficient way
```suggestion
	return {};
```

gizmocuz: or the readable way 😁

` return std::string("")`

hardware/ThermostatHardware.h
labelette91580: ok.done !

hardware/VirtualThermostat.cpp
labelette91580: ok.done !

hardware/ThermostatHardware.h
labelette91580: ok.done !

main/SQLHelper.cpp
labelette91580: ok.done !

hardware/ThermostatHardware.h
labelette91580: ok.done !

hardware/ThermostatHardware.h
labelette91580: Ok!done

hardware/VirtualThermostat.cpp
labelette91580: Ok!done delete

main/SQLHelper.cpp
labelette91580: Ok done.

hardware/VirtualThermostat.cpp
labelette91580: Ok done.

hardware/VirtualThermostat.cpp
labelette91580: Ok done.

main/SQLHelper.cpp
labelette91580: Ok done.

main/SQLHelper.cpp
labelette91580: Ok done.

main/SQLHelper.cpp
labelette91580: Ok done.

main/SQLHelper.cpp
labelette91580: Ok done.

main/SQLHelper.cpp
labelette91580: Ok done.

main/WebServer.cpp
labelette91580: Ok done.

main/WebServer.cpp
labelette91580: Ok done.

main/WebServer.cpp
labelette91580: Ok done.

main/SQLHelper.cpp
labelette91580: Ok done.

gizmocuz (approved):

kiddigital:
Hi @labelette91580 

> @labelette91580 , I like your idea of the _VirtualThermostat_ device and this 1st implementation of combining an 'on/off switch' and a 'temperature sensor' 👍
> 
> When I have some time, I will also take a look at your proposed code and come back with my 2 cents.

Maybe we need to consider naming it differently, at least this implementation. Your implementation is 1 implementation of a Virtual Thermostat. Following your idea, there could be others in the future.

So we need a name that reflects this particular implementation of the vthermostat, like 'Vtherm 1tempsensor 1on/off'? And also in the code base.

Another request, there are a lot of changes in this PR but some of them are not related to this PR but are formatting changes in non-related codelines. Would be very useful to revert such changes so the PR becomes a little more clear on what is changes to support this without being 'polluted' with non-related changes.

BTW. feel free to suggest code formatting/refactoring changes, etc. in a separate PR. There is always room for improvement. :)

main/SQLHelper.cpp
keryell: > or the readable
> 
> ` return std::string("")`

I guess what you mean by readable is the default constructor `return std::string()` or `return std::string{}`. Otherwise `std::string("")` will have to look at the C-string you pass before constructing the `std::string`, which is a waste of time and memory. But probably not the critical hot path in Domoticz for performance, I agree. :-)

mojso:
when it will be implemented in the beta version ?

kiddigital:
Hi @labelette91580 , any thoughts on using an alternative name for this specific Virtual Thermostat implementation?

labelette91580:
hi,
yes .
could you tell me the name you want !
i will modify it !
De : KidDigital ***@***.***>
Envoyé : jeudi 6 avril 2023 19:11
Cc : labelette91580 ***@***.***>; Mention ***@***.***>
Objet : Re: [domoticz/domoticz] Virtual Thermostat hardware (PR #5505)


Hi @labelette91580<https://github.com/labelette91580> , any thoughts on using an alternative name for this specific Virtual Thermostat implementation?

999LV:
As the original author of the Smart Virtual Thermostat python plugin that I have seen mentioned in this thread that I am just discovering, I just want to comment that I do not understand why a virtual thermostat should be hardcoded as hardware in Domoticz… it is all about logic rather than hardware control and having it hardcoded makes it inflexible, while a script or a plugin can be tweaked endlessly. The python plugin that I developed has been tweaked and forked many times over by users that customized it to their precise setup and needs.
But this is just a point of view obviously. Though I’d be interested to see a guidance on what should be hardcoded v.s. coded as script or plugin from @gizmocuz ?

gizmocuz:
@999LV , that is also exactly my idea mentioned in one of the first comments.
Either a python or dzVents script can do this and is easier to implement

megamarco833:
hello i give my two cents
i basically i agree with @gizmocuz better to have it on dzvents so everyone can customise it.
but could be an option introduce a new dummy devices as a thermostat? maybe this new thermostat dummy device can be a "merge" of:
- selector switch (to control status like on, off,cooling, heating, vacation, etc...)
- setpoint
- temperature
- humidity

when the device will be created, there should be the possibility to have if customized with the above function.
in this case there will be the advantage that with only one device we can display the setpoint, the status, the temperatue and/or humidity
of course will be fantastic have the possibility to update it with API and MQTT
and have the possibility to update this device only with parameters that you need

gizmocuz:
You could also make a custom template page for this device where all is displayed.
Currently all above are implemented, and indeed seperate (except temp+hum), not that humidity has anything to do with a setpoint device

gizmocuz:
Best to use dzVentz for this

PR #5517, opened 2023-01-05
Fixes Weather Tab

### LINKED ISSUE #5510: Log page of weather sensor show also the weather widgets
Hello @gizmocuz 

Various users, including myself, see if they open the log page for the weather sensors, also the widgets below the log graphics.
In my case it is all, except the widget for visibility and rain, but this may be different for other users.

For more details, including screenshots, see: https://www.domoticz.com/forum/viewtopic.php?p=296681#p296681

Forget, what I mentioned there about the log in "Scenes". This is solved, as I do not see it in version 14901.

Regards and a Happy new Year

soif:
fixes #5510

### What was the issue
When the WS triggers **RefreshItem**, **RefreshLiveSearch** is launched and it changes the "#weatherwidgets" _display_ property, thus removing the _display:none_ , normally set when the Log is displayed. 

### What does this patch:
It ensures that **LIveSearch** does not change the _display_ property more than ONCE, so when **RefreshItem** retriggers the **RefreshLiveSearch**, it does not revert the _display_ property back to a "shown" state.

### Better Fix
IMHO, **RefreshItem**, should ONLY happen, when devices are supposed to be shown, but NOT when they are not displayed.
- Better coding practice
- This would avoid this patch
- Refreshing devices when not needed, adds some browser's (+ domoticz server) stress, and might even lead to memory leaks in some cases

As you said that you would refactor the whole Angular code, (to normalize pages currently served by 2 different code designs), might I suggest that you _refactor the  **RefreshItem** code , to only happens WHENEVER needed._

HTH

### DIFF
--- a/www/js/domoticz.js
+++ b/www/js/domoticz.js
@@ -6292,17 +6292,22 @@ function WatchLiveSearch(){
 		var div=$('.divider');
 		var cont=$('.devicesList');
 		if(query.length == 0){
-			cont.removeClass('devicesListFiltered');
-			div.css('display','block');
-			div.addClass('row');
-			div.find('.clearfix').show();
-			$('.itemBlock').show().removeClass('liveSearchShown');
+			if(cont.hasClass('devicesListFiltered')){
+				cont.removeClass('devicesListFiltered');
+				div.css('display','block');
+				div.addClass('row');
+				div.find('.clearfix').show();
+				$('.itemBlock').show().removeClass('liveSearchShown');	
+			}
 		}
 		else{
-			cont.addClass('devicesListFiltered');
-			div.css('display','inline');
+			if(! cont.hasClass('devicesListFiltered')){
+				cont.addClass('devicesListFiltered');
+				div.css('display','inline');
+			}
 			div.removeClass('row');
 			div.find('.clearfix').hide();  /* only for Wheater and Temperatures */
+
 			$('.itemBlock').each(function(index){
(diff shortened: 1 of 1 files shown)


gizmocuz:
The refactoring could take some while... So if you feel up to it, please go ahead already

I am not 100% what you mean by 'to only happens WHENEVER needed'

I have not written that part of the code, But from what I understand is that you get updates, find the item, update the item, and done
I don't think you want to make a request with 80 parameters (80 items on that page)

gizmocuz:
Thank you for the patch! And the quick response!

soif:
> The refactoring could take some while... So if you feel up to it, please go ahead already

As I said, I'm not confortable with Angular, so I'm not the right guy to refactor it properly... sorry 😞 

> I am not 100% what you mean by 'to only happens WHENEVER needed'

I mean that when the devices are displayed (ie Weather Tab default view), it makes sense to have the WS periodically querying for any device changes, and updating it (BTW updating ONLY those who have changed would also be better/faster, instead of  updating EVERY devices)
BUT, when we are displaying a "Sub page", ie "Log", "Edit", "Timers", "Notifications", etc... the devices list is not longer displayed, and in theses cases, the WebSocket refresh should NOT be triggered, until the View is reverted back to the device list.

gizmocuz:
From the application the web page gets a push through the websocket with new updates for 'a' device

I think when you go to a sub page, the web socket could be closed, and when you go back, enable it again

PR #5518, opened 2023-01-05
LiveSearch also searches IDx

soif:
This quick PR improves the LiveSearch feature to add the ability to also search for the device's IDx.

### DIFF
--- a/www/app/DashboardController.js
+++ b/www/app/DashboardController.js
@@ -1649,7 +1649,7 @@ define(['app', 'livesocket'], function (app) {
 								if (($scope.config.DashboardType == 2) || (window.myglobals.ismobile == true)) {
 									xhtm +=
 										'\t    <tr id="scene_' + item.idx + '">\n' +
-										'\t      <td id="name" class="name item-name" data-desc="'+item.Description.replace('"',"'")+'">' + item.Name;
+										'\t      <td id="name" class="name item-name" data-idx="'+item.idx+'" data-desc="'+item.Description.replace('"',"'")+'">' + item.Name;
 									xhtm +=
 										'</td>\n';
 									var status = "";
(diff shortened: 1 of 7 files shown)


soif:
_BTW, I dont know how easy it would be to prevent the Github CI to always build the whole Domoticz code, when changes are made outside of the C code (ie: in the "www" folder), but it would certainly be way faster, and it would avoid wasting build ressources...._

kiddigital:
> _BTW, I dont know how easy it would be to prevent the Github CI to always build the whole Domoticz code, when changes are made outside of the C code (ie: in the "www" folder), but it would certainly be way faster, and it would avoid wasting build ressources...._

It is easy, but...
The PR CI/CD pipeline does not only Build but also _Tests_ and _creates Artifacts_.
Some tests could/should use the front-end (www folder) and these tests need to be performed _especially_ when changes have been made to the front-end.
And when new code appears in the WWW folder, a new Artifact for the PR should be created so potential interested (Test) Users can download the Artifact and test the new code.
But to run the tests and create the Artifact, Domoticz itself needs to be compiled first as every CI/CD run starts with an empty (container) environment. And yes, building the Domoticz binary does take some time.

soif:
>every CI/CD run starts with an empty (container) 

Obviously, without first building Domoticz, tests can not be performed...  😢

gizmocuz:
Thanks!! Maybe for the switches page data-status (on/off/closed/open, the bigtext contents)
With this you can see quickly all switches that are on or off
In the JS check if it is not undefined and perform the compare...
But maybe this is too much 😁

soif:
Thank for the merge! 👍 

> Maybe for the switches page data-status

Good idea! and it should be very fast to code ....

1) Can you please first list me the full list of possible status, or simply point me to the C file that define/list them.

2) would you like if the search would "group" status results? I mean: 
- Typing "on" would return all devices that are either "on" or "closed". 
- Typing "off" would return all those that are either "off" of "open".

gizmocuz:
Maybe if you put the big text inside data-status that you have everything you want.
No need to group as On/Closed if for some On/Open

PR #5522, opened 2023-01-06
Removal of Website Protection setting (Only Login screen, Basic-Auth only for API)

kiddigital:
This PR removes the old 'Website Protection' method where the user could choose between 'Login Page' and 'Basic-Auth'.

Using Basic-Auth for access to the UI was never really useful and was only intended for use with the API so non-users (scripts, etc.) could authenticate in another way instead of having to fill in credentials on the Login screen every time (which does not work for scripts).

As now the Basic-Auth (and X509 and JWT) method is improved for use with API's, only 'Login Page' is the option to Authenticate for the UI. Therefor this choice has become irrelevant.


waltervl:
Changes needed on wiki page https://www.domoticz.com/wiki/Security ??

kiddigital:
> Changes needed on wiki page https://www.domoticz.com/wiki/Security ??

Yes and also [here](https://www.domoticz.com/wiki/Application_Settings#Website_Protection)

(Do we have a way on the Wiki to differentiate between what is valid for the Stable and what is WIP for the Beta?)

waltervl:
I think we do all WIP on the security wiki and make some remarks on the other pages touching security that things have changed. If a new Stable is going to be released we update all pages to the new situation.

Currently there is no wiki for Beta. If on other functions/integrations something changes we normally mention that in the wiki page: From beta release xxxx something has been changed/ added.

gizmocuz:
Thanks again for this PR and time!!!

PR #5525, opened 2023-01-07
LiveSearch also searches in Status + displays results count

soif:
as asked in #5518

enjoy 😎

### DIFF
--- a/www/js/domoticz.js
+++ b/www/js/domoticz.js
@@ -6293,16 +6293,19 @@ function WatchLiveSearch(){
 
 	$('.jsLiveSearch').off().on('keyup change',function(e){
 		if(_debug_livesearch)  console.log('LiveSearch: processing on keyup - "'+$(this).val()+'"');
-		var query=$(this).val().toUpperCase();
-		var div=$('.divider');
-		var cont=$('.devicesList');
+		var query	=$(this).val();
+		var div		=$('.divider');
+		var cont	=$('.devicesList');
+		var items	=$('.itemBlock');
+		var cl_shown='liveSearchShown';
+
 		if(query.length == 0){
 			if(cont.hasClass('devicesListFiltered')){
 				cont.removeClass('devicesListFiltered');
 				div.css('display','block');
 				div.addClass('row');
 				div.find('.clearfix').show();
-				$('.itemBlock').show().removeClass('liveSearchShown');	
+				items.show().removeClass('liveSearchShown');	
 			}
 		}
 		else{
@@ -6313,24 +6316,24 @@ function WatchLiveSearch(){
 			div.removeClass('row');
 			div.find('.clearfix').hide();  /* only for Wheater and Temperatures */
 
-			$('.itemBlock').each(function(index){
+			items.each(function(index){
 				var name	=$(this).find('#name').html()				|| '';
 				var desc	=$(this).find('#name').attr('data-desc')	|| '';
 				var idx		=$(this).find('#name').attr('data-idx')		|| '';
 				var status	=$(this).find('#name').attr('data-status')	|| '';
 
-				name	=name.toUpperCase();
-				desc	=desc.toUpperCase();
-				status	=status.toUpperCase();
+				var safe_query=query.replace('\\','').replace('[','\\[').replace(']','\\]').replace('.','\\.')
+				var re		= new RegExp(safe_query, "mi");
+				var re_s	= new RegExp('^'+safe_query, "i");
 
 				var to_hide=$(this);
-				if ( name.indexOf(query) > -1 || desc.indexOf(query) > -1 || idx.indexOf(query) > -1 || status.indexOf(query) > -1) {
+				if ( re.test(name) || re.test(desc) || re.test(idx) || re_s.test(status) ) {
 					to_hide.show();
-					to_hide.addClass('liveSearchShown');
+					to_hide.addClass(cl_shown);
 				}
 				else{
 					to_hide.hide();
-					to_hide.removeClass('liveSearchShown');
+					to_hide.removeClass(cl_shown);
 				}
 			});
 		}
--- a/www/js/domoticz.js
+++ b/www/js/domoticz.js
@@ -6317,13 +6317,14 @@ function WatchLiveSearch(){
 				var name=$(this).find('#name').html().toUpperCase();
 				var desc=$(this).find('#name').attr('data-desc');
 				var idx=$(this).find('#name').attr('data-idx');
+				var status=$(this).find('#name').attr('data-status').toUpperCase();
 				if(desc === undefined){
 					desc='';
 				}
 				desc=desc.toUpperCase();
 
 				var to_hide=$(this);
-				if ( (name.indexOf(query) > -1) || desc.indexOf(query) > -1 || idx.indexOf(query) > -1) {
+				if ( name.indexOf(query) > -1 || desc.indexOf(query) > -1 || idx.indexOf(query) > -1 || status.indexOf(query) > -1) {
 					to_hide.show();
 					to_hide.addClass('liveSearchShown');
 				}
--- a/www/js/domoticz.js
+++ b/www/js/domoticz.js
@@ -6289,7 +6289,7 @@ function RefreshLiveSearch(){
 
 /* Watch the LiveSearch INPUT field  */
 function WatchLiveSearch(){
-	if(_debug_livesearch) console.log('LiveSearch : Start Watching ...');
+	if(_debug_livesearch) console.log('LiveSearch: Start Watching ...');
 
 	$('.jsLiveSearch').off().on('keyup change',function(e){
 		if(_debug_livesearch)  console.log('LiveSearch: processing on keyup - "'+$(this).val()+'"');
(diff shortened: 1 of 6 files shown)


www/js/domoticz.js
```diff
@@ -6317,13 +6317,14 @@ function WatchLiveSearch(){
 				var name=$(this).find('#name').html().toUpperCase();
 				var desc=$(this).find('#name').attr('data-desc');
 				var idx=$(this).find('#name').attr('data-idx');
+				var status=$(this).find('#name').attr('data-status').toUpperCase();
```
gizmocuz: Just checking, if (for example for a utility devices) there is no status, would this not give an assertion?
Maybe status should be checked like 'desc' below, if it is undefined, we make it empty else we make it uppercase?

gizmocuz:
Fast as lightning ;)

www/js/domoticz.js
soif: my bad! You're right!

_--> I always forget that there is currently no global template for devices 😉_ 

A fix will follow ASAP.

soif: fixed!

soif:
- I fixed the `undefined` case
- Search now uses RegEx: It should be better for internationalization + allow typing some basic Regex, ie `^name`  (starts with) or `pattern$` (ends with)
- Status are searched only from the **start** of the data-status txt (better filtering IMHO)

soif:
As a Bonus:
- Search results count is displayed while searching.
- A close icon allows to quickly reset the search

Hope you like it?

kiddigital (commented):
@soif , it is good practice to keep a PR branch up-to-date with its base branch to prevent any merge conflicts when the PR gets merged. (so thx for that).

As an alternative to `merge` the updated base branch back into the PR branch (which creates a merge-commit in the PR branch), you can _rebase_ the PR branch on its base branch (`git rebase -i upstream/development` for example). This effectively takes the updated base branch and tries to re-apply all commits in the PR branch.

Great thing also is that in case of any merge conflicts, they will appear during the rebase and you can fix them immediately. The necessary changes to fix the merge conflicts become part of the commit that caused the merge conflict, so no extra commit(s) needed to get the intended result of the initial commit but now applied on the updated base branch. Than you can `git push -f` the PR branch again and the PR branch is up-to-date and ready to be merged.

This way you keep the PR branch both up-to-date and ready for merging without having extra merge commits and possible merge conflict solving commits :)

soif:
@kiddigital 
Thank for the helpful explanation. 😉 
I'm aware of the git rebase mechanism, but I've never used it.
_I did not found the time to make a test project, and then experiment how the rebase works... But I promise I will do it one day or later..._ 😎

gizmocuz:
Thank you for this PR!! Looking good... two small issues

When I search for 'on' it also displays 'off", same for 'off' it displays also the 'on' switches

image

This is probably caused by the type indication? Maybe we have to change the on/off to normal, or ignore this on/off?

The font with the counter could maybe be a little bolder or bigger... it's a bit small

soif:
>When I search for 'on' it also displays 'off", same for 'off' it displays also the 'on' switches

I dont know for the **Off** case, but regarding the **On** case,  from your screenshot,  the device is shown because the name is  testbutt**ON** 😛 

>This is probably caused by the type indication? 

Nope, type is not searched, only fields in the data-XXX parameters...

>The font with the counter could maybe be a little bolder or bigger... it's a bit small

I've done this because while the count info is interesting, its not VERY important, so it can be smaller.

I could make it a bit larger, but as you may have noticed, ASAP as the count is displayed, the search field width is reduced (by the average amount of the count field width) so that, the following blocks (ie Room plans) are not shifted to the right (even more important on mobiles, which could lead to have plans popup to "jump "to the next line.)

So I can enlarge the font, but it will reduce the search field a little bit more.. Do U Want it?

gizmocuz:
@soif , you're right ;) it's because 'on' is part of the name , same on a device that has 'motion' in the name ;)
No please leave the count as it is ;)

Thinking further, I'm on my Utility view and wanted to quickly see all my kWh sensors, or Usage/Watt sensors, or Lux sensors, or percentage sensors, or all my custom sensors...
Maybe for this view it would be nice to have the 'type' field searchable?

soif:
Per request, I just added Search in the Status from the Utility page.

I rather search in "Status", than  in "Type", because:
- searching the unit used in the Status (ie "Watt", "° C", "Kb/s") mostly archieves the same  results
- It make the Search UI more consistent: The user expects to search in "Name,Desc,Ids,Status" everywhere, but not in the "type"

Note that to allow to search _somewhere_ in the Status, I've removed the RegEx that was searching only from start.

Sounds Good?

soif:
New Bonus: Also searches Status from the Scenes page.

_--> again this makes the search UI more consistent_

------

BTW: I've also intended to add Status Search in Temperatures/wheather pages, BUT it is very very very tedious, because of a design flaw (IMHO) in these pages.. 

_--> The final Status (=BigText) is built from the Angular template (xxx_widget.html), rather than from the controller._

 Might I suggest that when you'd redesign the whole Angular framework, you might add a global JS function ie _"LoadItemsFormatted(url)"_ (that would be called by each page controller), that loops thru each retrieved item, and formats each item (maybe in a sub _FormatItem(item) function), to create a new **item.html_value** (=BigText=Status) in the _items_ object. That way, each xxx_widget.html template just have to print the **item.html_value** without having to worry about complex types combinations logic...

_This item.html_value, could also be directly built from the C code, so that it would also appears in API answers..._

 See what I mean?
 Makes sense ?

gizmocuz:
Thank you once again! Your fast!!! 😁
Yes it makes sense to refactor the HTML/JS code, but to be honest I think it will take quite some time to make a new UI based on typescript/Angular instead of AngularJS.
But typescript makes it a lot more error free

About the generic function (js or in c++) that's a good idea. In principle the idea was to use the 'Data' field.
But some sensors have multiple values (like a kWh sensor (today counter and current usage, same for P1 meter, gas and a few others)
Maybe for some devices the 'Data' field should be extended with the unit
But for most of the devices the bigtext='Data' field

soif:
> Thank you once again! Your fast!!! 😁 

I don't like to leave a PR not **fully** finished... 😇 

>Yes it makes sense to refactor the HTML/JS code, but to be honest I think it will take quite some time to make a new UI 

Huge work, I know.... 😲 

> About the generic function (js or in c++) that's a good idea. In principle the idea was to use the 'Data' field.

IMO it would be best in the C++ code!!!
-  API users wouldn't have to fully replicate this logic in their "**Views**", anymore to correctly print data, whatever the type! 
- That's the way a smart API is usually built... 😺
- On the other hand, it would allow to drastically simplify the actual JS/Angular code ASAP, without having to wait for the complete Angular refactoring.... 

------

ready/happy to merge ?

gizmocuz:
I have tested your additions ;) ... I still have a problem finding my kWh devices in the utility tab
In the seach box I enter 'kWh' and I only see the devices that have this text inside the title/name

But this device is not found

image

gizmocuz:
The text kWh is also in the today, but also in the 'type'

soif:
>I rather search in "Status", than in "Type", because:

Again 😉 only STATUS (aka BigText) is searched, not TYPE...

Search for "Watt".... 🙄

gizmocuz:
Yes, that will work but, on some systems, this produces 50% wrong/not needed results 😁
If you have a additional data field to search for (which is never used, but in this case filled with the Type for Utility sensors) this would do the trick
At the moment you have multiple data types, maybe it is an idea to have one and add all data into this you want to search for?
It would maybe make the search function simpler and faster?
But happy to merge if you say it's ready to be!

soif:
> Yes, that will work but, on some systems, this produces 50% wrong/not needed results 😁 

That's obviouly the drawback of searching MULTIPLE fields... I can not do any magic here ATM 🙄 
_Any better idea is welcome !_

>If you have a additional data field to search for (which is never used, but in this case filled with the Type for Utility sensors) this would do the trick 

I can, but you initially asked to search in the "Status" field, and this is what this PR basically does...

>At the moment you have multiple data types, maybe it is an idea to have one and add all data into this you want to search for? 

Yep, I might add **type** search (and certainly also subtype search) to some (not all) pages... But can we add/discuss this in another PR?

>It would maybe make the search function simpler and faster? 

Nope (or I did not understood). It would simply add another field searched, in some case (ie :only in the Utility page).... This would make the Search UI less consistent, because only, in this case (Utility page), the search would ALSO search in type...

>But happy to merge if you say it's ready to be!

NO, you're the master of merge, and only YOU can say if it deserves to be merged ATM, or not 😄 

From my point of view, this PR fits your initial request, but  if you want me to first improve it b4 merging, I'm here to serve!

gizmocuz:
Thanks again!!

soif:
Thanks for the merge! 

You're also (with @kiddigital ) nominated as the the fastest merger of the universe... 🤣  🤣  🤣 

If needed, please feel free to start an issue (notifying me) to also search in Type, in some cases...

Best regards

gizmocuz:
Thanks, Okey I will spawn an idea in a new issue ;)

PR #5526, opened 2023-01-08
Remove console.log(item)

soif:
_( Was introduced in 365fdd1e5717e1fb4ccbb06d8ed40de8b2d652f2 )_ 

A basic no-brainer merge request 😉

### DIFF
--- a/www/app/LightsController.js
+++ b/www/app/LightsController.js
@@ -323,7 +323,7 @@ define(['app', 'livesocket'], function (app) {
 			var img3 = "";
 			var status = "";
 
-			console.log(item);
+			//console.log(item);
 
 			var bigtext = TranslateStatusShort(item.Status);
 			if (item.UsedByCamera == true) {


gizmocuz:
Thanks for the PR!

soif:
@gizmocuz : You're welcome... But this PR was really easy 😃 
 
@kiddigital : You're now nominated as the **faster merger of the universe**! 🤣🤣🤣🤣

soif:
BTW @kiddigital, I've seen my commit added to the develop branch but with a slightly different message/email, and my 'RemoveLog' branch was not seen as merged, on my side.

Why this?
Is this a sort of cherryPick ? 
Some GitHub mechanism I'm not aware of ?

_Not important, just for my knowledge..._

kiddigital:
@soif , your small PR has been merged via 'Squash and commit' mechanisme which creates a single commit (with PR number in the message) instead via the normal 'Merge' method which takes all commits in the PR and add a merge commit as well.

The only 'problem' with 'Squash and commit' is that your PR branch commit(s) cannot be found back in the upstream branch (develop) and therefor git cannot determine if commits have made it into the 'upstream' branch.

That is why you have to delete the PR branch locally after pulling the upstream branch.

BTW, me being 'fast' is thanks to the excellent Github mobile app (iOS) 😁

soif:
Many thanks for the explanation ! 🍻🍻🍻 

I'm forced to say that you're also nominated as the **faster to reply** ! 🤣 

_Thank you! 😇 : it's really a pleasure to commit here..._

PR #5052, opened 2021-12-13
Domoticz Weekly Planning Selector

xbeaudouin:
This Pull Request is the portage of the following patch : https://github.com/syrhus/domoticz-weekly-planning

Idea is to have choice on the weekly planning to allow for example heating to be easily set.
Author didn't made a PR, so I made it for him since I really think this should be a nice addition to Domoticz.
Some Screenshots : (but the original project has some more as well)

Capture d’écran 2021-12-13 à 18 04 21
Capture d’écran 2021-12-13 à 18 04 09

PR #5542, opened 2023-01-14
correct  Domoticz Weekly Planning Selector #5052 :

labelette91580:
This are some correction about PR from xbaudouin : Domoticz Weekly Planning Selector #5052 
 
display table by default
correct background select entry
correct scene timer regression
no display Grid in case Led/scene
correct display of blinds in grid

I hope this will suit to merge it!

### DIFF
--- a/www/app/timers/DeviceTimersController.js
+++ b/www/app/timers/DeviceTimersController.js
@@ -1,4 +1,4 @@
-define(['app', 'timers/factories', 'timers/components'], function (app) {
+define(['app', 'timers/factories', 'timers/components','timers/planning'], function (app) {
 
     app.controller('DeviceTimersController', function ($scope, $routeParams, deviceApi, deviceLightApi, deviceRegularTimersApi, deviceSetpointTimersApi, deviceTimerOptions, deviceTimerConfigUtils, utils) {
         var vm = this;
@@ -43,7 +43,7 @@ define(['app', 'timers/factories', 'timers/components'], function (app) {
                     ? deviceSetpointTimersApi
                     : deviceRegularTimersApi;
 
-                if (vm.isSelector) {
+                 if (vm.isSelector) {
                     vm.levelOptions = device.getSelectorLevelOptions();
                 }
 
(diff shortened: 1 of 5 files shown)


gizmocuz:
Yes, great!!!! thanks again!!

Some small requests if possible
- the red/green are 100% saturated, my eyes are in pain... could you make it a bit less green/red
- the selected mode tab (Table/Grid), could you make it the same as in Settings (also the selected style)

If possible the color rectangle a little bit down so it is vertically aligned with the text in front of it:
image

For the thermostats, please change the default frost protection temperature to 5 degrees.

The only thing that feels a bit weird is that I'm used to draw a rectangle and when I release the left mouse it should change the selection.
At least, this is my experience from camera monitoring systems
Here when you have the left mouse button down, it directly sets the value and it is very easily to move your mouse to a wrong value
But that's how the control is designed.... 

Looking great!

gizmocuz:
Looking great!
More than happy to merge!

Question, are you able to draw a rectangle between the left mouse button down and up (or touch down/up) instead of displaying the blocks?

labelette91580:
i will see if it is simple to do !
De : gizmocuz ***@***.***>
Envoyé : dimanche 22 janvier 2023 15:59
Objet : Re: [domoticz/domoticz] correct Domoticz Weekly Planning Selector #5052 : (PR #5542)


Looking great!
More than happy to merge!

Question, are you able to draw a rectangle between the left mouse button down and up (or touch down/up) instead of displaying the blocks?

gizmocuz:
Lightning fast! Thanks for another amazing job!!!

I noticed it is not (yet) possible to use the grid with RGB(w/ww/..) lights.
I guess this is because we need to specify a colour to turn on to.
Maybe this is possible to specify once the rectangle has been drawn (when ON was specified), than op popup to specify the RGB/W/WW value

Thanks again!

labelette91580:
no.
this is not possible because the color identify the state.
or for rgb there are to many differents state.

for example I have to reduce tthe state for the dimmer to 10 color

De : gizmocuz ***@***.***>
Envoyé : lundi 23 janvier 2023 12:14
Objet : Re: [domoticz/domoticz] correct Domoticz Weekly Planning Selector #5052 : (PR #5542)


Lightning fast! Thanks for another amazing job!!!

I noticed it is not (yet) possible to use the grid with RGB(w/ww/..) lights.
I guess this is because we need to specify a colour to turn on to.
Maybe this is possible to specify once the rectangle has been drawn (when ON was specified), than op popup to specify the RGB/W/WW value

Thanks again!

PR #5545, opened 2023-01-15
[Enphase] Add installer password to get inverter details

mce35:
On some firmware of the Enphase Envoy (mine is D5.0.55), a password is required to get the individual  inverter production. This password is derived from the serial number. This PR adds the password for the inverter API call.
Without the password the box returns 401:
```
2023-01-15 19:29:14.659  Debug: SQLH UpdateValueInt Enphase kWh Production HwID:2  DevID:00000201 Type:243  sType:29 nValue:0 sValue:0.000;3660303.000 
2023-01-15 19:29:14.659  Enphase: General/kWh (Enphase kWh Production)
2023-01-15 19:29:14.661  Debug: SQLH UpdateValueInt Enphase total-consumption HwID:2  DevID:00000202 Type:243  sType:29 nValue:0 sValue:2354.000;6051140.000 
2023-01-15 19:29:14.661  Enphase: General/kWh (Enphase total-consumption)
2023-01-15 19:29:14.662  Debug: SQLH UpdateValueInt Enphase net-consumption HwID:2  DevID:00000203 Type:243  sType:29 nValue:0 sValue:2348.000;3810153.000 
2023-01-15 19:29:14.662  Enphase: General/kWh (Enphase net-consumption)
2023-01-15 19:29:14.677  Debug: HTTP 401: Unauthorized
2023-01-15 19:29:14.677  Error: Enphase: Invalid data received! (inverter details)
```

With password the plugins is able to update the details of each inverter:
```
2023-01-15 23:39:58.202  Debug: SQLH UpdateValueInt 122138073676 HwID:2  DevID:00000280 Type:243  sType:29 nValue:0 sValue:0.000;337.000 
2023-01-15 23:39:58.204  Enphase: General/kWh (122138073676)
2023-01-15 23:39:58.205  Debug: SQLH UpdateValueInt 122134049595 HwID:2  DevID:00000281 Type:243  sType:29 nValue:0 sValue:0.000;342.000 
2023-01-15 23:39:58.207  Enphase: General/kWh (122134049595)
2023-01-15 23:39:58.208  Debug: SQLH UpdateValueInt 122134052922 HwID:2  DevID:00000282 Type:243  sType:29 nValue:0 sValue:0.000;340.000 
2023-01-15 23:39:58.210  Enphase: General/kWh (122134052922)
2023-01-15 23:39:58.212  Debug: SQLH UpdateValueInt 122134051631 HwID:2  DevID:00000283 Type:243  sType:29 nValue:0 sValue:0.000;334.000 
2023-01-15 23:39:58.213  Enphase: General/kWh (122134051631)
2023-01-15 23:39:58.215  Debug: SQLH UpdateValueInt 122134051700 HwID:2  DevID:00000284 Type:243  sType:29 nValue:0 sValue:0.000;336.000 
2023-01-15 23:39:58.216  Enphase: General/kWh (122134051700)
2023-01-15 23:39:58.218  Debug: SQLH UpdateValueInt 122134052139 HwID:2  DevID:00000285 Type:243  sType:29 nValue:0 sValue:0.000;338.000 
2023-01-15 23:39:58.219  Enphase: General/kWh (122134052139)
2023-01-15 23:39:58.221  Debug: SQLH UpdateValueInt 122134052445 HwID:2  DevID:00000286 Type:243  sType:29 nValue:0 sValue:0.000;340.000 
2023-01-15 23:39:58.222  Enphase: General/kWh (122134052445)
2023-01-15 23:39:58.224  Debug: SQLH UpdateValueInt 122134052356 HwID:2  DevID:00000287 Type:243  sType:29 nValue:0 sValue:0.000;338.000 
2023-01-15 23:39:58.225  Enphase: General/kWh (122134052356)
```

I hope this doesn't break newer firmware.


gizmocuz:
Thank you for this PR and your time.

What is the function 'emupwGetMobilePasswd' trying to do? What is the response? And what does this function tries to strip?
Decrementing counters does not seem a good and stable idea

Please do not modify the code in HTTPClient, if you want to supply a username, i suppose this can be done with

http(s)://<username>@<ipaddress

But your Enphase is on a old firmware. Why don't you update to version 7 and you can use this class without modification?

mce35:
> Thank you for this PR and your time.
> 
> What is the function 'emupwGetMobilePasswd' trying to do? What is the response? And what does this function tries to strip? Decrementing counters does not seem a good and stable idea

The algorithm was reversed engineered by someone based on the Enphase installer toolkit. I agree it looks a bit strange. You can find more details [here](https://thecomputerperson.wordpress.com/2016/08/28/reverse-engineering-the-enphase-installer-toolkit/)
The name of the function is from the original library in the installer toolkit.

> Please do not modify the code in HTTPClient, if you want to supply a username, i suppose this can be done with
> 
> http(s)://@<ipaddress

OK, it was actually already used to send another user/password, maybe for another version of the firmware. So I changed the code to try first with the existing user/password (envoy/last six digits of the SN). If it fails I try with the installer password.

> But your Enphase is on a old firmware. Why don't you update to version 7 and you can use this class without modification?

I will try to do this using the installer toolkit, I hope it still works without internet access with firmware v7.

Thank you.

gizmocuz:
Thanks for the explanation. Happy to merge, or you want to update first to version V7 and if that works we keep it this way?

mce35:
I think that even if I update to v7 it is worth merging. With v7 it will use the token authentication, and people with older versions may find this useful.

gizmocuz:
Maybe people with older firmwares need to update?

hardware/EnphaseAPI.cpp
```diff
@@ -227,6 +227,55 @@ int EnphaseAPI::getSunRiseSunSetMinutes(const bool bGetSunRise)
...
+static std::string emupwGetPasswdForSn(const std::string &serialNumber, const std::string &userName, const std::string &realm)
+{
+    std::string szMD5 = GenerateMD5Hash(std::string("[e]") + userName + "@" + realm + "#" + serialNumber + " EnPhAsE eNeRgY ");
+
+	return szMD5;
+}
+
+static std::string emupwGetMobilePasswd(const std::string &serialNumber, const std::string &userName, const std::string &realm)
+{
+    auto digest = emupwGetPasswdForSn(serialNumber, userName, realm);
```
gizmocuz: please change auto to std::string

hardware/EnphaseAPI.cpp
```diff
@@ -227,6 +227,55 @@ int EnphaseAPI::getSunRiseSunSetMinutes(const bool bGetSunRise)
...
+	if(digest.length() <= 8)
+		return "";
+
+	auto countZero = std::count(digest.begin(), digest.end(), '0');
```
gizmocuz: please change auto to int

hardware/EnphaseAPI.cpp
```diff
@@ -227,6 +227,55 @@ int EnphaseAPI::getSunRiseSunSetMinutes(const bool bGetSunRise)
...
+	auto countOne = std::count(digest.begin(), digest.end(), '1');
```
gizmocuz: please change auto to int

hardware/EnphaseAPI.cpp
```diff
@@ -227,6 +227,55 @@ int EnphaseAPI::getSunRiseSunSetMinutes(const bool bGetSunRise)
...
+    std::string szPassword = "";
```
gizmocuz: no need to initialize with = "";

hardware/EnphaseAPI.cpp
```diff
@@ -227,6 +227,55 @@ int EnphaseAPI::getSunRiseSunSetMinutes(const bool bGetSunRise)
 	return 0;
 }
 
+// emupwGetPasswdForSn / emupwGetMobilePasswd taken from https://github.com/sarnau/EnphaseEnergy
+static std::string emupwGetPasswdForSn(const std::string &serialNumber, const std::string &userName, const std::string &realm)
```
gizmocuz: this function only generated a MD5 hash, this can be done inside emupwGet... function, so please do not use a separate function for this

hardware/EnphaseAPI.cpp
```diff
@@ -227,6 +227,55 @@ int EnphaseAPI::getSunRiseSunSetMinutes(const bool bGetSunRise)
...
+{
+    std::string szMD5 = GenerateMD5Hash(std::string("[e]") + userName + "@" + realm + "#" + serialNumber + " EnPhAsE eNeRgY ");
+
+	return szMD5;
+}
+
+static std::string emupwGetMobilePasswd(const std::string &serialNumber, const std::string &userName, const std::string &realm)
```
gizmocuz: Please make this a member of the class and not static
Also maybe rename this to V5_GetMobilePassword

gizmocuz (changes requested):
Thanks for this PR
Some small changes request

Could you also please check your indenting!? I see tabs/spaces etc...
All our code is tab based

mce35:
> Maybe people with older firmwares need to update?

I agree but since the update is not automatic, I guess most users will not update their envoy box unless they have an issue with it.

hardware/EnphaseAPI.cpp
mce35: done

hardware/EnphaseAPI.cpp
mce35: done

hardware/EnphaseAPI.cpp
mce35: donedone

hardware/EnphaseAPI.cpp
mce35: done

hardware/EnphaseAPI.cpp
mce35: done

hardware/EnphaseAPI.cpp
mce35: done, I leave the emupw prefix to stick with the original function name from the library.

gizmocuz:
Could it be that 'std::string sUserPwd' is not used? (if not, please remove it 😁)

xbeaudouin:
> > Maybe people with older firmwares need to update?
> 
> I agree but since the update is not automatic, I guess most users will not update their envoy box unless they have an issue with it.

I have emphase here... And ... the envoy still on R5.0.55 (and the envoy app doesn't even show there is an update)...

rezzalex:
> > > Maybe people with older firmwares need to update?
> > 
> > 
> > I agree but since the update is not automatic, I guess most users will not update their envoy box unless they have an issue with it.
> 
> I have emphase here... And ... the envoy still on R5.0.55 (and the envoy app doesn't even show there is an update)...

On my side, a phone call to Enphase Support did the trick to get the latest firmware version installed

xbeaudouin:
> > > > Maybe people with older firmwares need to update?
> > > 
> > > 
> > > I agree but since the update is not automatic, I guess most users will not update their envoy box unless they have an issue with it.
> > 
> > 
> > I have emphase here... And ... the envoy still on R5.0.55 (and the envoy app doesn't even show there is an update)...
> 
> On my side, a phone call to Enphase Support did the trick to get the latest firmware version installed

Since in my case this is my 'own made' emphase installation, I really doubt enphase support will do it for me :) 

So yeah support "old firmware" should be a good idea :)

mce35:
> Could it be that 'std::string sUserPwd' is not used? (if not, please remove it grin)

Removed.

mce35:
I have one question regarding the existing code, maybe someone can help clarify this. For the widget to work properly, it is needed to configure it to compute the total production of each individual inverter by checking this in the GUI:
image

Is there a way to do this programmatically so that the user doesn't need to do this?

This is because each inverter only reports the current production and the max production (for its lifetime I guess) in Watts, it does not report the total energy produced.
```
  {
    "serialNumber": "122138073676",
    "lastReportDate": 1673887334,
    "devType": 1,
    "lastReportWatts": 0,
    "maxReportWatts": 337
  },
```

I don't understand passing the maxReportWatts as the 4th argument here but I may be missing something.

```
int musage = itt["lastReportWatts"].asInt();
int mtotal = itt["maxReportWatts"].asInt();

SendKwhMeter(m_HwdID, iIndex++, 255, musage, mtotal / 1000.0, szSerialNumber);
```

mce35:
> > > > Maybe people with older firmwares need to update?
> > > 
> > > 
> > > I agree but since the update is not automatic, I guess most users will not update their envoy box unless they have an issue with it.
> > 
> > 
> > I have emphase here... And ... the envoy still on R5.0.55 (and the envoy app doesn't even show there is an update)...
> 
> On my side, a phone call to Enphase Support did the trick to get the latest firmware version installed

I tried to update using the installer toolkit but it seems to be restricted to professional installers. Same on the website, the upgrade should be possible but one need access to their "Enlighten Manager" which I don't have.
I have 2 options:
 - Ask my installer to provide me access if this is possible (not sure)
 - Send a support request to Enphase

In the mean time I am stuck with the V5 firmware :(

So definitely supporting old firmware versions is a good idea.

gizmocuz:
About the password, did you try username:envoy password: last six digits of the serial number?
(Some report admin/admin, but above worked here)

Regarding the individual inverter details, I hoped that the Max value was a continues counter, but I guess it's not

I made a patch (already committed) that changes the type to computed.

gizmocuz (approved):

gizmocuz:
Maybe the username/password you are looking for is the one you calculate with 'V5_emupwGetMobilePasswd' ? ;)

gizmocuz:
One other change please....

At startup the software version is retrieved

Only when it is less then 7 request the MobilePassword 

Add the check below these lines please:

```
	if (
		(m_szSoftwareVersion.size() < 2)
		|| (m_szSoftwareVersion[0] != 'D')
		)
	{
		Log(LOG_STATUS, "Unsupported software version! Please contact us for support!");
	}

```

mce35:
> Regarding the individual inverter details, I hoped that the Max value was a continues counter, but I guess it's not
> 
> I made a patch (already committed) that changes the type to computed.

There was a small issue in setting the option, it was never called, I fixed it on this PR.

> One other change please....
> 
> At startup the software version is retrieved
> 
> Only when it is less then 7 request the MobilePassword

OK I added a check, in `getInverterDetails` it is still trying with the envoy:<last 6 digits SN> user/pwd first and if it fails and if lower than v7 then tries with the installer password.

I don't know exactly which password should be used for which firmware version. I was not able to find a definitive answer looking at forums. It looks like v3 is using the `envoy` username, but I am not sure of this. So maybe
- v3: envoy/last 6 digits SN
- v5: installer/generated password from SN
- v7: token only? not sure if the installer password would work on v7

We would need confirmation before implementing this.

gizmocuz:
Thanks for the fix. It was a left over debug line

It seems to me that when you run the installer application and you can login, you have the right password?

https://support.enphase.com/s/question/0D53m00007EXzi1CAD/how-do-i-login-to-my-envoy-communication-gateway-monitor

To login locally to your V7 firmware you need a token, a token you can get online after login.

gizmocuz:
So, when you just update your firmware this PR is not needed?

mce35:
> So, when you just update your firmware this PR is not needed?

Probably, it looks like with v7 only token authentication is possible.

**But** token authentication **do** require internet access (see [here](https://support.enphase.com/s/question/0D53m00006ySLuRCAW/unimpressed-with-loss-of-local-api-connectivity-to-envoys)). Tokens seem to be valid 6 months.

So I will **not** upgrade my Envoy, and it has no internet access.

gizmocuz:
Why does it not have internet access? I assume you are using this at home.
Your Domoticz system has internet access, as well as many other devices.

There is a good reason they switched to JWT tokens as this makes it a lot more secure.
Imagine you have a Wi-Fi camera that gets hacked (or you are using a not-updated/old operating system that gets hacked)
They can take over your envoy system because of its weak security. (Still wonder what they are going to do with this)

Aldo the token is indeed valid for 180 days (6 months), you are able to use any other device to generate a new token and copy/paste this in your Domoticz system (or envoy), it's inside a user variable

mce35:
> Why does it not have internet access? I assume you are using this at home. Your Domoticz system has internet access, as well as many other devices.

My domoticz has internet access and can be accessed through a reverse proxy. But the Envoy is on a separate VLAN without internet access as I don't trust it.

> There is a good reason they switched to JWT tokens as this makes it a lot more secure. Imagine you have a Wi-Fi camera that gets hacked (or you are using a not-updated/old operating system that gets hacked) They can take over your envoy system because of its weak security. (Still wonder what they are going to do with this)

I agree JWT is more secure, actually the Envoy security seems to be quite low, but this is the same for many other systems like this (Cameras, ...). I also have some cameras (without Wi-Fi of course), but as with the Envoy they are on a separate VLAN without Internet access.

> Aldo the token is indeed valid for 180 days (6 months), you are able to use any other device to generate a new token and copy/paste this in your Domoticz system (or envoy), it's inside a user variable

What would happen if for some reason Enphase decided not to generate tokens anymore? Users would have to rely on their cloud based API to access their data. I don't really like systems that depends on the vendors clouds to operate properly.

gizmocuz:
That's a good question to ask Envoy, maybe contact support or add a post.
Wonder what they say.

gizmocuz:
Thanks for this PR! Hope you can upgrade to V7 soon (and later 8,9,10,11)

PR #5546, opened 2023-01-16
Allow wildcards in MQTT auto discovery  topics

StephenR1:
Find all subscriptions which might have triggered the message delivery using a wildcard match and pass the original subscription topic downstream to allow matching to devices.

Hard code to avoid trying to match the global wildcard 'm_TopicDiscoveryPrefix'/# to continue to allow sensor data to be held in the same topic root as the sensor descriptor data.

### DIFF
--- a/hardware/MQTTAutoDiscover.cpp
+++ b/hardware/MQTTAutoDiscover.cpp
@@ -1,6 +1,7 @@
 #include "stdafx.h"
 #include "MQTT.h"
 #include "MQTTAutoDiscover.h"
+#include "../extern/mosquitto/include/mosquitto.h"
 #include "../main/json_helper.h"
 #include "../main/Helper.h"
 #include "../main/localtime_r.h"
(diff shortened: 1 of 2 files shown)


StephenR1:
Hi,

I tried to follow the 'contribution' request to write all code changes against the master branch but I notice that other pull requests are against the development one so I wonder if I misunderstood what was being asked?

On a separate note, I see that the CI task has failed and doesn't like the reference to the mosquitto.h in 'extern'. I am not sure on this one so help would be great. It builds for me and the .h file is already there (i.e. I didn't add it). I did a search for other references and I see there (only one) it uses the unqualified <mosquitto.h>, should I use that?
I only want the helper function to compare a wildcard topic with a FQ version (and I wrote my own version of that before seeing an existing one here!) so am happy to revert to mine?
I could add the ref to the base class wrapper but was trying to keep my changes to a minimum so as not to break anything else...

Sorry.

StephenR1:
I am going to assume I got wrong end of stick and move the changes to the development branch.

I also now see that I can remove the explicit ref to 'mosquitto.h' because it gets included (indirectly) already.

gizmocuz:
Yes that's correct , please use the development branch

gizmocuz:
Yes you need to use

<mosquitto.h>

that is, 'if' you need to use this

I am still not sure why we need a wildcard

Issue #5280, opened 2022-07-18
MQTT AD missing call to create sensors


redswan01:
Hi,
I have been trying to debug why my Sonoff SNZB-02 were not being created despite being discovered. I have come to the conclusion that there's a missing call in the function "on_auto_discovery_message" in MqTTAutodiscover.cpp at line 1046/7, 
`		if (pSensor->component_type == "sensor")
		{
			if (pSensor->state_topic.empty())
			{
				Log(LOG_ERROR, "A sensor should have a state_topic!");
				return;
			}
			handle_auto_discovery_sensor(pSensor, message);  // This line is missing
		}
`
After adding this line then devices are created, though in my case I get individual devices for for temperature, humidity and battery as opposed to one combined temp/humidity device.

Maybe someone with more knowledge of the intricacies could confirm this is the correct place to do this, and also make a pr for me as I'm not even a novice where github is concerned.

I first came across this problem with 2022.1  build 14352 on a Rpi 3b with Bullseye, to debug I cloned 2022.1 build 14394 on windows and built/debugged with VS2022.

Thanks.

Chris

gizmocuz:
Sensors are 'detected' and 'created' when actually received
Only setpoints/lights/switches are created when detected

redswan01:
OK, thanks for the explanation, that explains why I thought it was missing.
However, in may case, this does not happen, neither for the Sonoff SNZB-02 nor for temp/humidity sensor I try to create following the example in the HomeAssitant doc, I've added a unique_ID so that Domoticz doesn't create a new one at each discovery.
Both these sensors work correctly in HA, in fact for the moment I use HA to forward the data to Domoticz!!
Apparently Domoticz discovers these sensors, but sending data doesn't create any devices!

2022-07-21 11:23:47.762  Status: MQTT broker: discovered: Temp_Sensor_1/Temp_Sensor_1_temperature (unique_id: 0x00124b00251ce5b2_temperature_homeassistant)
2022-07-21 11:23:47.870  Status: MQTT broker: discovered: Temp_Sensor_1/Temp_Sensor_1_humidity (unique_id: 0x00124b00251ce5b2_humidity_homeassistant)
2022-07-21 11:23:47.984  Status: MQTT broker: discovered: Temp_Sensor_1/Temp_Sensor_1_voltage (unique_id: 0x00124b00251ce5b2_voltage_homeassistant)
[...]

PR #5548, opened 2023-01-16
Allow wildcards in MQTT auto discovery topics

StephenR1:
Find all subscriptions which might have triggered the message delivery using a wildcard match and pass the original subscription topic downstream to allow matching to devices.

Hard code to avoid trying to match the global wildcard 'm_TopicDiscoveryPrefix'/# to continue to allow sensor data to be held in the same topic root as the sensor descriptor data.

### DIFF
--- a/hardware/MQTTAutoDiscover.cpp
+++ b/hardware/MQTTAutoDiscover.cpp
@@ -50,19 +50,6 @@ MQTTAutoDiscover::MQTTAutoDiscover(const int ID, const std::string& Name, const 
 	}
 }
 
-bool MQTTAutoDiscover::IsWildcardMatch(const std::string st_topic,const std::string m_topic)
-{
-	bool result;
-	if(mosquitto_topic_matches_sub(st_topic.c_str(), m_topic.c_str(),&result) == MOSQ_ERR_SUCCESS && result == true)
-	{
-		Debug(DEBUG_HARDWARE, "Wildcard compare of %s and %s matched", st_topic.c_str(), m_topic.c_str());
-		return true;
-	}
-	
-	Debug(DEBUG_HARDWARE, "Wildcard compare of %s and %s did not match", st_topic.c_str(), m_topic.c_str());
-	return false;
-}
-
 void MQTTAutoDiscover::on_message(const struct mosquitto_message* message)
 {
 	std::string topic = message->topic;
@@ -78,7 +65,10 @@ void MQTTAutoDiscover::on_message(const struct mosquitto_message* message)
 		bool topicMatch = false;
 		for (auto& itt : m_subscribed_topics)
 		{
-			if (itt.first != m_TopicDiscoveryPrefix + "/#" && IsWildcardMatch(itt.first,topic))
+			bool result = false;
+			mosquitto_topic_matches_sub(itt.first.c_str(), topic.c_str(),&result);
+
+			if (itt.first != m_TopicDiscoveryPrefix + "/#" && result == true)
 			{
 				handle_auto_discovery_sensor_message(message,itt.first);
 				topicMatch = true;			
--- a/hardware/MQTTAutoDiscover.cpp
+++ b/hardware/MQTTAutoDiscover.cpp
@@ -50,6 +50,19 @@ MQTTAutoDiscover::MQTTAutoDiscover(const int ID, const std::string& Name, const 
 	}
 }
 
+bool MQTTAutoDiscover::IsWildcardMatch(std::string st_topic, std::string m_topic)
+{
+	bool result;
+	if(mosquitto_topic_matches_sub(st_topic.c_str(), m_topic.c_str(),&result) == MOSQ_ERR_SUCCESS && result == true)
+	{
+		Debug(DEBUG_HARDWARE, "Wildcard compare of %s and %s matched", st_topic.c_str(), m_topic.c_str());
+		return true;
+	}
+	
+	Debug(DEBUG_HARDWARE, "Wildcard compare of %s and %s did not match", st_topic.c_str(), m_topic.c_str());
+	return false;
+}
+
 void MQTTAutoDiscover::on_message(const struct mosquitto_message* message)
 {
 	std::string topic = message->topic;
@@ -62,11 +75,16 @@ void MQTTAutoDiscover::on_message(const struct mosquitto_message* message)
 		if (qMessage.empty())
 			return;
 
-		if (m_subscribed_topics.find(topic) != m_subscribed_topics.end())
+		bool topicMatch = false;
+		for (auto& itt : m_subscribed_topics)
 		{
-			handle_auto_discovery_sensor_message(message);
-			return;
+			if (itt.first != m_TopicDiscoveryPrefix + "/#" && IsWildcardMatch(itt.first,topic))
+			{
+				handle_auto_discovery_sensor_message(message,itt.first);
+				topicMatch = true;			
+			}
 		}
+		if(topicMatch == true) return;
 
 		if (topic.substr(0, topic.find('/')) == m_TopicDiscoveryPrefix)
 		{
@@ -1212,9 +1230,9 @@ void MQTTAutoDiscover::ApplySignalLevelDevice(const _tMQTTASensor* pSensor)
 	}
 }
 
-void MQTTAutoDiscover::handle_auto_discovery_sensor_message(const struct mosquitto_message* message)
+void MQTTAutoDiscover::handle_auto_discovery_sensor_message(const struct mosquitto_message* message,std::string subscribed_topic)
 {
-	std::string topic = message->topic;
+	std::string topic = subscribed_topic;
 	std::string qMessage = std::string((char*)message->payload, (char*)message->payload + message->payloadlen);
 
 	if (qMessage.empty())
--- a/hardware/MQTTAutoDiscover.cpp
+++ b/hardware/MQTTAutoDiscover.cpp
@@ -50,7 +50,7 @@ MQTTAutoDiscover::MQTTAutoDiscover(const int ID, const std::string& Name, const 
 	}
 }
 
-bool MQTTAutoDiscover::IsWildcardMatch(std::string st_topic, std::string m_topic)
+bool MQTTAutoDiscover::IsWildcardMatch(const std::string st_topic,const std::string m_topic)
 {
 	bool result;
 	if(mosquitto_topic_matches_sub(st_topic.c_str(), m_topic.c_str(),&result) == MOSQ_ERR_SUCCESS && result == true)
@@ -1230,7 +1230,7 @@ void MQTTAutoDiscover::ApplySignalLevelDevice(const _tMQTTASensor* pSensor)
 	}
 }
 
-void MQTTAutoDiscover::handle_auto_discovery_sensor_message(const struct mosquitto_message* message,std::string subscribed_topic)
+void MQTTAutoDiscover::handle_auto_discovery_sensor_message(const struct mosquitto_message* message,const std::string subscribed_topic)
 {
 	std::string topic = subscribed_topic;
 	std::string qMessage = std::string((char*)message->payload, (char*)message->payload + message->payloadlen);
--- a/hardware/MQTTAutoDiscover.cpp
+++ b/hardware/MQTTAutoDiscover.cpp
@@ -62,13 +62,19 @@ void MQTTAutoDiscover::on_message(const struct mosquitto_message* message)
 		if (qMessage.empty())
 			return;
 
+		if (topic.substr(0, topic.find('/')) == m_TopicDiscoveryPrefix)
+		{
+			on_auto_discovery_message(message);
+			return;
+		}
+
 		bool topicMatch = false;
 		for (auto& itt : m_subscribed_topics)
 		{
 			bool result = false;
 			mosquitto_topic_matches_sub(itt.first.c_str(), topic.c_str(),&result);
 
-			if (itt.first != m_TopicDiscoveryPrefix + "/#" && result == true)
+			if (result == true)
 			{
 				handle_auto_discovery_sensor_message(message,itt.first);
 				topicMatch = true;			
@@ -76,11 +82,6 @@ void MQTTAutoDiscover::on_message(const struct mosquitto_message* message)
 		}
 		if(topicMatch == true) return;
 
-		if (topic.substr(0, topic.find('/')) == m_TopicDiscoveryPrefix)
-		{
-			on_auto_discovery_message(message);
-			return;
-		}
 		return;
 	}
 	catch (const std::exception& e)
--- a/hardware/MQTTAutoDiscover.h
+++ b/hardware/MQTTAutoDiscover.h
@@ -139,7 +139,7 @@ public:
 	void on_disconnect(int rc) override;
 
 private:
-	bool IsWildcardMatch(std::string st_topic, std::string m_topic);
+	bool IsWildcardMatch(const std::string st_topic,const std::string m_topic);
 	void InsertUpdateSwitch(_tMQTTASensor* pSensor);
 
 	void UpdateBlindPosition(_tMQTTASensor* pSensor);
@@ -153,7 +153,7 @@ private:
 	void ApplySignalLevelDevice(const _tMQTTASensor* pSensor);
 
 	void on_auto_discovery_message(const struct mosquitto_message* message);
-	void handle_auto_discovery_sensor_message(const struct mosquitto_message* message,std::string subscribed_topic);
+	void handle_auto_discovery_sensor_message(const struct mosquitto_message* message,const std::string subscribed_topic);
 
 	void handle_auto_discovery_availability(_tMQTTASensor* pSensor, const std::string& payload, const struct mosquitto_message* message);
 	void handle_auto_discovery_sensor(_tMQTTASensor* pSensor, const struct mosquitto_message* message);
--- a/hardware/MQTTAutoDiscover.h
+++ b/hardware/MQTTAutoDiscover.h
@@ -139,6 +139,7 @@ public:
 	void on_disconnect(int rc) override;
 
 private:
+	bool IsWildcardMatch(std::string st_topic, std::string m_topic);
 	void InsertUpdateSwitch(_tMQTTASensor* pSensor);
 
 	void UpdateBlindPosition(_tMQTTASensor* pSensor);
@@ -152,7 +153,7 @@ private:
 	void ApplySignalLevelDevice(const _tMQTTASensor* pSensor);
 
 	void on_auto_discovery_message(const struct mosquitto_message* message);
-	void handle_auto_discovery_sensor_message(const struct mosquitto_message* message);
+	void handle_auto_discovery_sensor_message(const struct mosquitto_message* message,std::string subscribed_topic);
 
 	void handle_auto_discovery_availability(_tMQTTASensor* pSensor, const std::string& payload, const struct mosquitto_message* message);
 	void handle_auto_discovery_sensor(_tMQTTASensor* pSensor, const struct mosquitto_message* message);
--- a/hardware/MQTTAutoDiscover.h
+++ b/hardware/MQTTAutoDiscover.h
@@ -139,7 +139,6 @@ public:
 	void on_disconnect(int rc) override;
 
 private:
-	bool IsWildcardMatch(const std::string st_topic,const std::string m_topic);
 	void InsertUpdateSwitch(_tMQTTASensor* pSensor);
 
 	void UpdateBlindPosition(_tMQTTASensor* pSensor);


StephenR1:
Hi gizmocuz

To answer your question from the  wrong PR I closed out:

I am not sure if you are asking why TheengsGateway (and OpenMQTTGateway which it is based on) use wildcards or why Domoticz needs to support them so will try to answer both.

If I just go for the later question first (because easier). Simple way to answer is that the two gateways above (and maybe others) use wildcarded topics in their autodiscovery messages and that does not work in Domoticz at the moment because the code which distributes messages to the 'owner' sensor currently does a straight equality check between the message topic (which is a fully qualified path) and the subscription topic (which is wildcarded)

(There is a summary from the OpenMQTTGateway forum)
The 'why do they use them' part of the question relates to situations where you have more than one gateway. (Large house maybe).
If you have more than one then you may get BT messages picked up by both gateways. If both gateways use different root topics then the devices appear twice. If they use the same root then you get some 'corruption'. You could blacklist the 'middle' devices in one of the gateways but that means a manual job every time you add a new device.
The solution they went for is quite nice and uses a wild card for the subscription topic so it doesn't matter if the message comes from one gateway or the other, they still both update the UI but the messages can be located in different root topics in the queue so they don't interfere with each other. The only downside is that the UI may sometimes process the same status update twice.
I hope that makes sense!

Thanks

Stephen

hardware/MQTTAutoDiscover.cpp
```diff
@@ -50,6 +50,19 @@ MQTTAutoDiscover::MQTTAutoDiscover(const int ID, const std::string& Name, const
 	}
 }
 
+bool MQTTAutoDiscover::IsWildcardMatch(std::string st_topic, std::string m_topic)
```
gizmocuz: could you declare the string as const like

(const std::string &st_topic, const std::string &m_topic)

gizmocuz:
Could you give me a few examples of wildcard topics?

This because at the moment there should be one auto discovery topic (homeassistant/domoticz/location_1/....)
and everything behind there is treated as config

In the config payload it is specified what are (for example) the state_topics
Then we subscribe to these topics

Some configurations specify a home topic (specified with a ~ sign)
But wildcard topics I have not seen yet.

Can you post a config payload as well? Because maybe we can detect this and add a different topic/string in **m_subscribed_topics** so we do not have to modify the routine

hardware/MQTTAutoDiscover.cpp
StephenR1: Sure thing. I have updated that (and also the new argument to handle_auto_discovery_sensor_message)
Does the 'pull' automatically update? I think I see it does but just checking that I don't need to do something else to make the update visible.
I will get some sample message logged in a min. (Just have a work meeting...)

gizmocuz: Yes you can keep pushing to this branch/PR

gizmocuz:
Meanwhile I found the forum discussion and could test/experience with this myself.

Two questions, instead of a new function 'IsWildcardMatch', we could just call **mosquitto_topic_matches_sub** directly?

Are you deliberately going over all elements in the for loop for a second match?
Can't we return after the message has been handled? (after calling **handle_auto_discovery_sensor_message** return true)
This also removes the extra boolean

StephenR1:
In case useful to have/see the messages here as well as the bits from the forum:

The 'descriptor' messages go to (in my 'default' setup) - homeassistant/sensor/A4C1382C87D4-tempc/config

and look like:
{
  "stat_t": "+/+/BTtoMQTT/A4C1382C87D4",
  "dev_cla": "temperature",
  "unit_of_meas": "°C",
  "name": "LYWSD03MMC_PVVX-tempc",
  "uniq_id": "A4C1382C87D4-tempc",
  "val_tpl": "{{ value_json.tempc }}",
  "state_class": "measurement",
  "device": {
    "identifiers": [
      "A4C1382C87D4"
    ],
    "connections": [
      [
        "mac",
        "A4C1382C87D4"
      ]
    ],
    "manufacturer": "Xiaomi",
    "model": "LYWSD03MMC_PVVX",
    "name": "ATC_2C87D4",
    "via_device": "TheengsGateway"
  }
}

The value messages for this then come with the following as the message topic  - home/TheengsGateway/BTtoMQTT/A4C1382C87D4

And look like:
{
  "name": "ATC_2C87D4",
  "id": "A4:C1:38:2C:87:D4",
  "rssi": -85,
  "brand": "Xiaomi",
  "model": "LYWSD03MMC",
  "model_id": "LYWSD03MMC_PVVX",
  "tempc": 16.86,
  "tempf": 62.348,
  "hum": 55.71,
  "batt": 82,
  "volt": 2.945
}

StephenR1:
> Meanwhile I found the forum discussion and could test/experience with this myself.
> 
> Two questions, instead of a new function 'IsWildcardMatch', we could just call **mosquitto_topic_matches_sub** directly?
> 
> Are you deliberately going over all elements in the for loop for a second match? Can't we return after the message has been handled? (after calling **handle_auto_discovery_sensor_message** return true) This also removes the extra boolean

I put in the new function in case it needed to be called by other parts of the code later (and to keep the original function cleaner). It is slightly messier than it might be because mosquitto routine has the 'match result' is passed by by ref and the return code from the function containing failure info. I wasn't 100% sure what the result code for the match would say if the routine failed for a 'system' reason so just coded for belt and braces.
Also, wrapping in another function allowed adding some debug logging which could be really useful for  these matches if people have problems.

I tried to make as few changes as I could but this routine (if it stays) would be much better of in the mosquittodz base class probably.

I loop through because there may be more than one match and I was trying to cover all situations.
At the moment, there is a check for an existing subscription before a new one is added but this is not 'wildcard' aware. There is an edge case which this allows for where one sensor has a wildcarded topic which is a superset of another. In that case, the subscription topics will be different and both sensors need to be updated.

E.g.
If one sensor subscribes to all updates at 
+/test/sensor 
and another one subscribes to updates at 
+/+/sensor
Then a message from my/test/sensor should go to both of them but the sensors are 'keyed' on their subscription topic so you have to call handle_auto_discovery_sensor_message twice (once with each subscription topic) to allow both sensor devices to update.

I put the 'hard coded' exception to m_TopicDiscoveryPrefix + "/#" in there to cope with the (also probably wrong) case I read about where people had both their descriptor messages in homeassistant (or where-ever) and also their value messages going to the same root. (Like I say, not a nice setup). The '#' subscription here is a global wildcard so would match on any topic coming in. My logic just says that if no more specific topics have matched an actual sensor value subscription then drop through to allowing the system to see if it is a sensor descriptor instead.

Sorry for not very well worded responses...

gizmocuz:
Thanks for the explanation.
Meanwhile I also tried another approach:

- handle_auto_discovery_sensor_message needs the subscribed_topic, same as you did

- written a new SubscribeTopic function in the MQTT-AD class that subscribes to the topic, but removes any '+/ ' from the topic string and use this in **m_subscribed_topics**
So it becomes 'BTtoMQTT/A4C1382C87D4' and covers +/ and +/+/ etc

this will also simplify the code in **on_message** as there is no need to call a mosquitto compare function

What do you think about going this way?

Anyway, currently 700 subscriptions and continuing the loop is very expensive cpu wise
If a match is found in the first 10 compares, it will continue to compare 690 other strings in a wild(card) way

StephenR1:
I think that won't work (in a general sense) because wildcards don't have to be at the start.
(Totally thinking hypothetically) I can see that a composite device might want to subscribe to :
my/test/sensor/+ if the underlying sensor report temp and humidity on different value message topics?

Worse (though I can't think why it would be done) would be something like +/test/+/sensor/+ 

On the performance point, even with 700 subscriptions this is a local string compare (admittedly with the wildcard logic in the compare) and should be very fast. (C being very good at that kind of thing).
Also, given that a 'random' message in the above case would take 350 compares, we are only doubling that so it is probably not too big an issue?

On he other hand, I did also think this as I wrote the code and the two optimisations I thought were:
Store a cache of fully qualified message topics against wildcard subscriptions so one only has to do the wildcard match for new ones.
Store the wildcard subscriptions in a different map so you do a simple 'find' against the fully qualified topics first and then do a check of the wild card ones after. You still have to do both unfortunately because a message might have a fully qualified subscription and a wildcard one but at least you could stop the FQ search after a single match.

I discounted both in the end because I think the wildcard match is pretty fast and the two ideas were overly complex and more likely to lead to errors than the benefit they gave...

(I know almost nothing about MQTT so sorry if the next bit is silly)
Do you happen to know if there is some sort of 'correlation' Id you can pass to MQTT when subscribing which you can get the broker to pass back on every matching message? Or, can you identify the subscription from the callback with the message? Either of those would allow the subscription topic to be 'attached' back to the messages as they came in and allow a simple full string compare to find the topics? (If the number of subscriptions was high, we could even sort the subscriptions into alphabetical order and do a binary search to find the match rather than a interactive one)

As I write the above, I realise that I don't know how cpp maps work and maybe they do use a hashing store and search method... I haven't written C++ for over 20 years and have lost touch a bit!

StephenR1:
Sorry, other other thing I just thought.

That proposed approach would make :
+/+/topic/thing
look like
+/topic/thing

But are obviously different topics. 

I think you still need to search the full list as well because the end of the topic string you get left with might match more than one sensor for the same reason as the wildcard does?

In the bit a few up for example, I think you would have subscription topic keys of:
/sensor
and
/test/sensor

A message coming in from my/test/sensor matches both.

StephenR1:
Just had thoughts on two more suggestions:

I think I like the first one best:
Make 'Support MQTT wildcards' an option flag in Domoticz. If the flag is no then use the original 'find' method, otherwise do the new iteration through?
That way anyone with no wildcards (The majority of users) don't have to suffer because of the minority but the minority still get everything working with the small performance hit!

Second option is to do the 'Find' method first and call the sensor update and return if one is found.
Then do an iterative match with wildcards if no 'total' match is made.
The wildcarded messages will be even slower in this case because the 'find' will need to look though all the subscriptions and the second iteration then goes through the whole lot again... But if majority of subscriptions are fully qualified then this might be a better compromise? 
(This has a small hole in the logic too in the very edge case where there is a fully qualified subscription and a wildcarded one which is a superset of the FQ one. I bet that *never* happens for real!)

I still like first option the most. Partly because it leaves things alone for people who don't want it!

StephenR1:
I haven't tested this but I think this new feature looks like a potential solution which would be both robust and quick.

Downside - It would need pretty major changes potentially! Maybe a thought for future development?

Looking at the new v5 MQTT features, I think the 'subscription Identifier' would allow the original subscription topic to be linked back to from the message. The description they give certainly seems pretty much the problem we are trying to fix.

https://www.emqx.com/en/blog/subscription-identifier-and-subscription-options

gizmocuz:
Okey let's hope it is fast enough ;)

Still one (the same) request, please use **mosquitto_topic_matches_sub** directly, no need for this extra function and it is flooding my debug log 100% (1000000%)
It is just a function that returns true/false, no big deal... it's like logging a debug message for std::string::find

StephenR1:
If I separate those in two.
I can definitely remove the debug logging! I hear what you are saying about the number but, to be fair, it was invaluable for me to see it was working properly and that is kind of what debug logging is for...

Also, totally happy to make the other change (it is your code!). My only point here is that the mosquitto_topic_matches_sub  doesn't return a true/false. The true false is returned in the argument you pass by reference.
The other advantage in wrapping this in another function is that it does allow the actual function to return true/false and that allows its use to be more (very slightly) more declarative and less imperative which makes it very slightly easier to follow. (But only to my mind I absolutely admit!)

Give me the nod though and I will absolutely make both changes and commit.

StephenR1:
Changes made.

Thanks.

gizmocuz:
Thanks for the change! Happy to merge... Just one question regarding:

`if (itt.first != m_TopicDiscoveryPrefix + "/#"`

Is it really possible that someone sends a message to **m_TopicDiscoveryPrefix + "/#"`** ?
I would expect that people send message to **m_TopicDiscoveryPrefix + "/some/topic"`**
And that you can not include **#** as output topic

should the code not be

`if (itt.first.find(m_TopicDiscoveryPrefix)==std::string::npos`

So any topic that starts with the discovery topic is not allowed

StephenR1:
Hi,

This was so as not to break other things that were already there.

You (i.e. the Domoticz autodiscover plugin) are the ones who subscribe to m_TopicDiscoveryPrefix + "/#" :)

If I put in the code as you suggest (or indeed, just move the code block from below to above this one to match on m_TopicDiscoveryPrefix and process them then return as the code used to) then it will break the process for the change someone asked for a while ago who had their value messages put under m_TopicDiscoveryPrefix.

The issue which caused that was fixed by you I think (sorry for the bitmap. I am sure there is a nicer way to reference a check-in but I don't know how to...)

image

StephenR1:
P.S. I think it is a really bad idea to have the value topics there but only did this so as not to step on other people's toes...

gizmocuz:
Yes, values should 100% not be inside the auto discovery topic.
I think this patch was made to ignore the errors some users where having.

This is the current code:

```
		if (m_subscribed_topics.find(topic) != m_subscribed_topics.end())
		{
			handle_auto_discovery_sensor_message(message);
			return;
		}

		if (topic.substr(0, topic.find('/')) == m_TopicDiscoveryPrefix)
		{
			on_auto_discovery_message(message);
			return;
		}
```

Now we could move the last part to the top:

```
		if (topic.substr(0, topic.find('/')) == m_TopicDiscoveryPrefix)
		{
			on_auto_discovery_message(message);
			return;
		}
```

So, this leaves this piece of code you have currently modified:

```
		if (m_subscribed_topics.find(topic) != m_subscribed_topics.end())
		{
			handle_auto_discovery_sensor_message(message);
			return;
		}
```

In this code I see no reference to 

**if (itt.first != m_TopicDiscoveryPrefix + "/#"**

So i think we could remove this check?

StephenR1:
Sorry. I was showing where you moved 

if (topic.substr(0, topic.find('/')) == m_TopicDiscoveryPrefix)
		{
			on_auto_discovery_message(message);
			return;
		}

To be below the call to update the sensor status (back in July last year).

If we move it back above (or don't call updates for any topics starting with m_TopicDiscoveryPrefix) then it will break the thing you fixed then?

This is really hard for me in the sense that I don't mind but I am trying not to break things for others.
I can make that change to reverse the change you made in July but it will break the system for [redswan01] that you fixed in issue https://github.com/domoticz/domoticz/issues/5280

gizmocuz:
Well the only thing is , I don't think the check **if (itt.first != m_TopicDiscoveryPrefix + "/#"** should be made, as this was never made before?

StephenR1:
No. But the wildcard match now matches on that so sends all messages to do with autodiscovery to the sensor code when they should go to the autodiscovery bit.

You didn't get this before because you were not doing a wildcard match,

gizmocuz:
What you say I understand (when they should go to the autodiscovery bit)
But what I find strange (as explained above) is that you can not send a message to **m_TopicDiscoveryPrefix + "/#**
I am very sure you are not allowed to send messagr to a topic like mytopic/#
You are able to send something to 'mytopic' or mytopic/subtopic

Also I have never seen a config file that has a (state) topic with a # inside

StephenR1:
I will write an explanation but it will be a little long so might take me a mo...

StephenR1:
Taking a very simple example of a single (mythical) sensor. I will also assume homeassitant as the root for the autodiscovery messages to make things easier to read.
I will assume the 'bad' case that we are trying to allow for redswan01 to continue to work.

Domoticz starts up and connects to MQTT
It subscribes to homeassistant/# (so that it receives autodiscovery messages)
As it subscribes, it puts the subscription topic (homeassistant/#) into m_subscribed_topics

A sensor linked to a gateway causes a descriptor message  message to be written.
Lets say that goes to homeassistant/sensor/IDofSensor
The descriptor says that the state topic is homeassistant/Temp_sensor_1
Domoticz subscribes to homeassistant/Temp_sensor_1 and puts the detail in m_subscribed_topics

Now looking at things before and after.
Before the wildcard matching, when we got messages from the autodiscovery queue then there was no match with the topic that had been put in m_subscribed_topics because homeassistant/# != homeassistant/sensor/IDofSensor for example
However, after the wildcard code is put in place then homeassistant/# does equal homeassistant/sensor/IDofSensor with a wildcard match. This makes the autodiscovery message go to the sensor update... Bad...

So we can fix that easily by having 

```
if (topic.substr(0, topic.find('/')) == m_TopicDiscoveryPrefix)
{
on_auto_discovery_message(message);
return;
}
```
*before* the sensor message dispatcher.

But if we do...Oh no... Now the value topics from redswan01 which come in on topic homeassistant/Temp_sensor_1 now match that code first and how value updates go to the autodiscovery code!

There are definitely nicer ways round this (like not adding the autodiscovery topic to m_subscribed_topics) but in the end I thought that having a simple rule to explicitly exclude homeassistant/# from going to the sensor was the simplest?
Then they just drop out as they should to the actual autodiscovery bit.
You are right that this does assume that no *sensor* would register a status topic of homeassistant/# but if it did then it probably deserves not to work!

I hope that makes sense! I was hard to write in an easy to understand way...

StephenR1:
One other thing. I assume you know that it is the autodiscovery process which is subscribing to #?

In MQTTAutoDiscover::on_connect(int rc)

```
	if (!m_TopicDiscoveryPrefix.empty())
		{
			SubscribeTopic((m_TopicDiscoveryPrefix + std::string("/#")).c_str());
			//Send online status
			SendMessage(m_TopicDiscoveryPrefix + std::string("/status"), "online");
		}

```

gizmocuz:
Hmmmz

No, we are not going to allow the bad case for redswan01... he messed up his zigbee2mqtt configuration.

gizmocuz:
> One other thing. I assume you know that it is the autodiscovery process which is subscribing to #?
Sure, but you cannot send a message to # as mentioned above. you send a message to topic/subtopic/../../.. not topic/#

StephenR1:
Ah. I think I see where you might have got the wrong end of the stick!

It is not doing a test in the code for a message having been sent to homeassistant/#
It is doing a check to see if the *subscription* was made to homeassistant/#

Basically it is wanting not to send messages that it gets in and should go to the autodiscovery code accidently to the sensor update code. If it did, the autodiscovery would never see them because any message which get distributed to the sensor update code causes the message processing to exit.

All that said, I just saw your comment about not keeping the  bad case for redswan01.
Assuming that is the case (and I agree there) then I can just switch round the check to look for autodiscovery messages first and remove the extra check for the sensors!

Shall I go ahead with that?

gizmocuz:
Okey that indeed explains ;) Yes I believe there should be a difference between the discovery topic and the actual sensor topic. They should not be the same.
In Zigbee2MQTT (as well as ZWaveJS-UI) you can configure both. (base-topic)

Yes please that would be great and we can merge this PR 👌👌

StephenR1:
Brilliant. Will do.

StephenR1:
Sorry. As I do this I notice one other thing it is probably best to ask you about.

The code does that pair of if() clauses and returns from each one if they match but the final one doesn't strictly need to do that because it will hit the final return anyway
i.e. we have

```
if(x)
{
    blah
    return ;
}

if(y)
{
    blah
    return ;
}

return

```

Would yo like me to simplify the code simply to drop out of the second if and pick up the final return or leave as is which makes it a bit more flexible if there is ever a third 'if' added.
I think leaving the extra return in the second if is better (even though currently redundant) but wanted to make sure you were happy!

StephenR1:
(Code isn't quite as I wrote but am sure you see what I mean!)

gizmocuz:
Yes that's correct, but who knows what the feature brings, maybe another compare ;)

StephenR1:
Thanks very much.

That update made and it looks all to be running fine I think.

gizmocuz:
Thank you for this PR and time/patience! 😁

StephenR1:
No problem at all! Thank you very much for the fantastic app. My bit here is minuscule compared to all the work you have done.

Really really appreciated!

StephenR1:
Hi Gizmocuz

Just a quick note to say that you do know that you changed the code in 
your last refactor to make it work differently?

It is totally ok (and not odd) for a message to match more than one wild 
card subscription and it is impossible to know which one did it (without 
adding the function available in v5 to link a subscription to a message)

The code before would call all subscriptions which matched but your new 
code only calls the first I think?

Thanks

Stephen

StephenR1:
Probably not a big deal though!

gizmocuz:
Yes I know, but i 'm quite certain that your MQTT devices still work and that the change that this is needed is < 1%
As mentioned, this will also make the lookup very slow as you iterate over all possible topics over and over again.
This is no issue when you have only 10 devices reporting 5 topics, but if you have 100 devices it's a different story.
But happy to undo when there is a issue reported of course

StephenR1:
Brilliant. Thanks Gizmocuz!

redswan01:
Ha, at least my messed up configuration, due to my ignorance on how the autodiscovery worked, has had some use as an example of what to avoid. Maybe it could be useful to have an error message if the sensor topic is declared in the AD path.

gizmocuz:
Yes, there is already a message, but just made a commit that should display more errors

IanDury:
Just updated to beta 14979 this morning and I get the following error log using MQTT AD for zigbee2mqtt. Could it be related to this PR?

`2023-01-21 07:08:23.232 Error: MQTT ZigBee: MQTT_Discovery: Invalid/Unhandled data received! (Topic: zigbeeauto/update/0x00178801062b65e7/update/config, Message: {"availability":[{"topic":"zigbee2mqtt/bridge/state","value_template":"{{ value_json.state }}"},{"topic":"zigbee2mqtt/Hue Fuzo outdoor wall light/availability","value_template":"{{ value_json.state }}"}],"availability_mode":"all","command_topic":"zigbee2mqtt/bridge/request/device/ota_update/update","device":{"identifiers":["zigbee2mqtt_0x00178801062b65e7"],"manufacturer":"Philips","model":"Hue Fuzo outdoor wall light (8718696170625)","name":"Hue Fuzo outdoor wall light","sw_version":"1.88.1"},"device_class":"firmware","entity_picture":"https://github.com/Koenkk/zigbee2mqtt/raw/master/images/logo.png","json_attributes_topic":"zigbee2mqtt/Hue Fuzo outdoor wall light","latest_version_template":"{{ value_json['update']['latest_version'] }}","latest_version_topic":"zigbee2mqtt/Hue Fuzo outdoor wall light","name":"Hue Fuzo outdoor wall light","payload_install":"{\"id\": \"0x00178801062b65e7\"}","state_topic":"zigbee2mqtt/Hue Fuzo outdoor wall light","unique_id":"0x00178801062b65e7_update_zigbee2mqtt","value_template":"{{ value_json['update']['installed_version'] }}"})`

gizmocuz:
Nope, It was caused by another patch.
This type is unhandled and therefor a message was given.
Beta 14980 silenced the message again

IanDury:
Confirmed. With beta 14980 the above error log doesn't show anymore. Thanks.

PR #5557, opened 2023-01-22
[OAuth2] Present a nice Auth form instead of using Browser Basic-Auth dialog

kiddigital:
This PR adds a nicer form to the Domoticz IAM service.

Now when an external app that wants to interact with Domoticz is redirecting the User to the Domoticz IAM service to request the User to Allow access, a Domoticz Authentication & Authorization dialog is presented. Once the User has been Authenticated AND granted the access, the IAM service redirects the User back to the origin app.

image

The PR also includes a bug fix in detecting the correct signing mechanism.
And now does proper checking of the Client Secret to validate if it matches.

At this moment, the IAM settings can not be set through command-line switches or config file settings, but the code has been prepared to propagate other than default settings.

Some code cleanup and changing DEFINE constants in proper settings so the could be changed runtime in the future.
Also some parts of the code from the IAM service are now in a separate source-code directory to make things more clear.


gizmocuz:
Looking great!
Is it possible to change the Login title to '.... your Domoticz system. ' (added the word system)
With this PR we now have two login screens, maybe it's an idea to make a style sheet (css) for this?

gizmocuz (approved):

PR #5567, opened 2023-01-25
Weekly planning selector

labelette91580:
hello,
optimize timer display size to be displayed on full screen on mobile android tablette .

Capture d’écran 2023-01-24 220937


gizmocuz:
Thank you again!!

PR #5570, opened 2023-01-26
update automaticaly the device id of sensor that battery as been replaces

labelette91580:
hello,
this a function that will detect automatically if a battery of a device with low battery as been replaced :
it update the new device ID that have been changed due to the battery replace.
An option as been added in settings to allow this function ; other/"Allow Device Replace On Battery Changed Detected" shall be ON
a log is written .
algo

if a new device is detected with batery level 100% (new batery)
 we search if a same device type/subtype/Unit already exist with batery low
if one is found , we assume that it is the same device that we have replace batery and a new ID as been modified. the old DeviceID is replaced by the new DeviceID
to allow this , the settings option  other/"Allow Device Replace On Battery Changed Detected" shall be ON

### DIFF
--- a/main/SQLHelper.cpp
+++ b/main/SQLHelper.cpp
@@ -5059,6 +5059,52 @@ uint64_t CSQLHelper::GetDeviceIndex(const int HardwareID, const std::string& ID,
 	return std::stoull(result[0].at(0));
 }
 
+/// @brief check if batery has been changed in a device with a new ID
+//  if a new device is detected with batery level 100% (new batery) 
+//  we seach if a same device type/subtype/Unit already exist with batery low
+// if one is found , we assume that it is the same device that we have replace batery and a new ID as been modified.
+// the old DeviceID is replaced by the new DeviceID
+// to allow this , the settings option  other/"Allow Device Replace On Battery Changed Detected" shall be ON
+//return true if device ID as been replaced 
+bool replace_device_when_changed_baterie_detected(const int HardwareID, const char* deviceID, const unsigned char unit, const unsigned char devType, const unsigned char subType, const unsigned char batterylevel, const char* User)
+{
+    if (!m_sql.m_bAcceptNewHardware)
+			    return false;
+	//if battery not full (replaced)
+    if (batterylevel != 100)
+			    return false;
+
+	 int iAllowDeviceReplaceOnBatteryChanged =0;
+	m_sql.GetPreferencesVar("AllowDeviceReplaceOnBatteryChanged", iAllowDeviceReplaceOnBatteryChanged);
+	if (iAllowDeviceReplaceOnBatteryChanged==0)
+		return false;
+
+//test if device exist : done before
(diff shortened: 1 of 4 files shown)


gizmocuz:
Thanks for this PR. That won't work 100%
- Battery level could be 80% depending on the battery you use
- Have multiple 433Mhz temperature sensors with the same unit with different device ID

It is quite easy to edit the old device and click on the replace button

labelette91580:
100% is the level after the batery is replaced : it shoul be 100% if batery is new

Yes the device shall have different unit value from 1..8.

If this this case , nothing is done and you have to do that manually
De : gizmocuz ***@***.***>
Envoyé : jeudi 26 janvier 2023 12:15
Objet : Re: [domoticz/domoticz] update automaticaly the device id of sensor that battery as been replaces (PR #5570)


Thanks for this PR. That won't work 100%

  *   Battery level could be 80% depending on the battery you use
  *   Have multiple 433Mhz temperature sensors with the same unit with different device ID

It is quite easy to edit the old device and click on the replace button

gizmocuz:
Got about 10 433Mhz temp+hum devices and some of them have duplicate units of course (as you can set mostly 3 different addresses)
I replace my battery when it's nearly empty (otherwise your device can damage if you wait to long)
My batteries are on the shelf or are pre-loaded, but not all of them are 100%, some are ~95, and sometimes even worse.
Does not matter as it will take a long time until it reaches a low value

Then we have other devices that are for instance zigbee, that have a value of 0-100, but the device_id/unit will luckly stay the same

I'm sorry, this is a very dangerous patch, and as mentioned, it is very easy to replace a old sensor with a new one

labelette91580:
ok.
i don't know for zigbee device.

But if the algorytm found more than one device with same device type/subtype/unit/battery level low ,
it does not do any thing : it shall be do manually

for zigbee : perhaps we can set
      - a power level for battery low : 20 %
      - a power level for battery high : 80%

in any case if there is more than one device , no thing is done.

cdt

De : gizmocuz ***@***.***>
Envoyé : vendredi 27 janvier 2023 08:43
Objet : Re: [domoticz/domoticz] update automaticaly the device id of sensor that battery as been replaces (PR #5570)


Got about 10 433Mhz temp+hum devices and some of them have duplicate units of course (as you can set mostly 3 different addresses)
I replace my battery when it's nearly empty (otherwise your device can damage if you wait to long)
My batteries are on the shelf or are pre-loaded, but not all of them are 100%, some are ~95, and sometimes even worse.
Does not matter as it will take a long time until it reaches a low value

Then we have other devices that are for instance zigbee, that have a value of 0-100, but the device_id/unit will luckly stay the same

I'm sorry, this is a very dangerous patch, and as mentioned, it is very easy to replace a old sensor with a new one

gizmocuz:
So, I have a low battery temp sensor , maybe I know, maybe not
I install a new temp sensor in another room
Bang things are mixed and broken

As said this is a dangerous patch and if needed, it is easy to replace a good sensor

PR #5583, opened 2023-02-03
Log filter

labelette91580:
add the possibility to filter the log/debug more detailed :
- for example display only the lines with contaning specific words
- or filter  only the lines with contaning specific words ( for example device reception dump)

 add "debugfilter" switch  in command line :

-debugfilter word1 :  filter all the LOG containing word1
-debugfilter all       :  filter all the LOG : to be used with  all keyword

-debugfilter all,+word1   :  display only log contaning word2 ( filter all word except word1 )
-debugfilter all,+word1,+word2   :  display only log contaning word1 & word2 ( filter all word except word1 & word2 )

### DIFF
--- a/main/Logger.cpp
+++ b/main/Logger.cpp
@@ -509,10 +509,10 @@ bool CLogger::NotificationLogsEnabled()
 void CLogger::SetFilter(const std::string  &pFilter)
 {
 	std::vector<std::string> FilterList;
-    _log.Log(LOG_STATUS, "debugfilter:%s", pFilter.c_str());
+	_log.Debug(DEBUG_NORM, "debugfilter:%s", pFilter.c_str());
 	FilterStringList.clear();
 	KeepStringList.clear();
-	StringSplit(pFilter, ";", FilterList);
+	StringSplit(pFilter, ",", FilterList);
 	for (size_t i = 0; i < FilterList.size(); i++)
 	{
 		if (FilterList[i][0] == '+')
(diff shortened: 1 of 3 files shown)


gizmocuz:
Thank you for this PR!
When you log to a file, you can already use 'tail -f' on this file and use the default logging tools from your system (grep/tee/cut/../../)
In Domoticz in the log window you can also enter the text to filter

labelette91580:
yes
but you are display only the line containing a single word .

here you can filtered many word list and display many word list .
On windows , this is available directly on the windows console with out using the http log windows.

so for debugging , it is very pacticle !
cdt




De : gizmocuz ***@***.***>
Envoyé : vendredi 3 février 2023 12:46
Objet : Re: [domoticz/domoticz] Log filter (PR #5583)


Thank you for this PR!
When you log to a file, you can already use 'tail -f' on this file and use the default logging tools from your system (grep/tee/cut/../../)
In Domoticz in the log window you can also enter the text to filter

gizmocuz:
I don't consider a windows user a serious Domoticz debugger 😁
And in my opinion you can easily disable all hardware except the one you are debugging and then use the log window
(Or the tail -f or console or the developer debug console)
That is if you are writing a new hardware class

labelette91580:
ok😥
i have also develop a log debug function with the level argument as a string
in order not to  be force to declare

  1.  a new _eDebugLevel entry
  2.  update SetLogFlags with the test of new level.

it is more evolutionary
use case  Debug("imperihome","message ")
void CLogger::Debug(const char*  level, const std::string &sLogline)

gizmocuz:
The current log implementation is sufficient and when you call the right log functions it already logs the hardware name with it.
Making a new log with a filter will cause every log line to be handled more slower, making everything slower.
It is very easy to make your own filter
As you already know, it is also possible to output the log file to disk. I enabled this option in my debug environment (visual studio, yes Windows 👌)
I can look into this file in realtime. (I use total commander), but I can also use the unix filter tools
You can also use a windows tool like 'Tailviewer' for this if you want
Sorry closing this PR, thank you for it but it is not needed

PR #5591, opened 2023-02-09
Livesearch enhancements

gizmocuz:


www/js/domoticz.js
```diff
@@ -6282,6 +6282,89 @@ function fromInstanceOrFunction(functionTemplate = f => f()) {
...
+		return current_data;
+	return current_data + " " +  new_value;
+}
+
+GenerateLiveSearchTextDefault = function (item) {
+	var searchText = "";
+	searchText = AddToLiveSearch(searchText, item.idx);
+	searchText = AddToLiveSearch(searchText, item.Name);
+	searchText = AddToLiveSearch(searchText, item.Description.replace('"',"'"));
+	searchText = AddToLiveSearch(searchText, item.Type);
```
soif, line 6308: AddToLiveSearch(searchText, item.Type) breaks the ability to correctly search in Light/Switch Status, (because for example, it adds the "Switch On/Off" to the search text), thus preventing to find Switch by status.

gizmocuz: There is no "Switch On/Off" type, code here is correct

PR #5592, opened 2023-02-09
Update SBFSpot.cpp

Bertels999:
Frequency of PV invertor is displayed as percentage using the SBFSpot plugin. It should be Hz.


gizmocuz:
Thank you for this PR!!

PR #5598, opened 2023-02-13
Adding Time-based One-Time Passcode (RFC6238) support for 2FA to Login and IAM service

kiddigital:
This PR adds a way to support 2-Factor Authentication.

By default Domoticz activates TOTP support for each User that has enabled 2FA (via the MyProfile screen). The Authentication screen requires a 6-digit passcode as additional Authentication factor next to the Username and Password.

This can be disabled using a command-line option (`-disable2fa`).

image

The standard Login screen of the Domoticz UI will request an TOTP code (on a second window) as well for Users that have set a TOTP secret. A check will be performed to validate the TOTP code against the Users secret.
 
It works with several Authenticators like [FreeOTP](https://freeotp.github.io/), [Google Authenticator](https://googleauthenticator.net/) and [Microsofts Authenticator](https://www.microsoft.com/en-us/security/mobile-authenticator-app). See screenshot below:

Depending on the Authenticator used, you can either enter the secret manually (often the base32 encoded string) or you need a QR-code to setup the Authenticator. 

### New 'My Profile' menu-option
There is a new '**_My Profile_**' menu option under 'Setup' (for Admin) or 'Profile' (for non-admins) where you can enable Two-Factor Authentication for the logged-in User (but only when Domoticz is running with the `-enable2fa` flag).

(You can also change the users password here :) No need to do that via an Admin and User management anymore)

When you '_enable MFA_' , an additional field appears where you need to enter the (base32 encoded) TOTP secret. A QR-code is display alongside that can be used to configure the Authenticator used.

At the moment, the secret has to have a length of 20 characters (at least 16 as it will be filled-up to 20) and only `sha1` is supported. Also only the standard 30-seconds interval is supported.

#### More on TOTP MFA QR-code 
On Linux a QR-code can be generated as followed (using the Python QR-code library):
```
echo -n "MySuperSecret1234567"|base32|qr "otpauth://totp/domoticz?algorithm=SHA1&digits=6&secret=$(</dev/stdin)"
```
(or leave out the part after 'base32' to just get the base32 encoded version of the specified secret.)

You can also use the [2fa-qr](https://stefansundin.github.io/2fa-qr/) site to enter the info and create a QR code on the fly.


kiddigital:
@gizmocuz , I did some updating to this PR and now when Domoticz is started with `-enable2fa`, the standard Login screen will ask for an (optional) TOTP code. If a user has set a TOTP-secret, a 2FA check will be performed.

gizmocuz:
I think it is best that this is an additional step ... Username/Password screen ---> check if user has 2FA enabled ---> Show 2FA screen for TOTP code --> inside
Would that make sense security wise? Also we could start with normal username/password for users that don't have this enabled as we don't know in advance who's going to login

kiddigital:
Created a 'MyProfile' page looking like this:
image

Needs some labels/translation to be filled in but its there :)

@gizmocuz , for non-admin there now is a small menu as well instead of the 'logout'-button. Hope you like it.

kiddigital:
> I think it is best that this is an additional step ... Username/Password screen ---> check if user has 2FA enabled ---> Show 2FA screen for TOTP code --> inside Would that make sense security wise? Also we could start with normal username/password for users that don't have this enabled as we don't know in advance who's going to login

@gizmocuz  Done :)

gizmocuz:
Thanks for all work! Looking great! As this is now a branch, we can delete this PR

kiddigital:
> Thanks for all work! Looking great! As this is now a branch, we can delete this PR

Sort of agree :) I like the PR to stick around for the documentation and information for now. I will keep the PR and Branch in-sync.

(Advantage in my opinion in merging the PR, is that a provides a good history while branches should be removed after a while).

kiddigital:
This PR has been followed-up by #5712

PR #5600, opened 2023-02-14
Align data points

hreintke:
Hi,

In the 2022.2 version, the column counter graphs for the daily, weekly, monthly and yearly are not aligned to the datapoints.
(Compare graphs does correctly align).

See f.e. a monthly column graph: 

Monthly2022 2

And weekly column graph

Weekly2022 2

In the beta version there is a commit fbac957 dd 25/12/2022 which solves this for the weekly graph.

WeeklyBeta

This PR applies the same update for the daily, monthly and yearly graphs.

MonthlyPR

Disclaimer : I only have a small Domoticz implementation and cannot test/verify all configurations.


gizmocuz:
Thank you for this PR! Is it possible to undo the changes in the ZWaveBase.cpp file?
Reason for this is that this file is going to be removed this week

hreintke:
Hi,

I don't know where the ZWaveBase.cpp update comes from.
It is not in my local repo, neither in my cloned version on Github.

Can it be caused by my PR to the development branch instead of master ? 

If so do you want me to close this PR and submit another ?

gizmocuz:
No you are correct to make PR's against the developers branch, this is the default
You can keep pushing against your forked and this PR will automatically update.

gizmocuz:
But you can create a new PR if you wish

hreintke:
Sorry, I read somewhere that PR's should go to master.

I branched from master in my local to create the PR. Can that be the reason ?

gizmocuz:
No, PR's should be made against the active branch (which is the dev branch in our case)
Maybe best to delete this PR and make a new one.

hreintke:
Redone, See PR #5601

arjendekorte:
I'm not certain that the changes in PR #5600 (or PR #5601 for that matter) are aesthetically pleasing for the day graph. Unlike the week/month graphs where this does help, it makes more sense to display the hourly usage of say 15:00 to 16:00 on a graph between the markers 15:00 and 16:00, rather than centered on either the 15:00 or 16:00 marker.

hreintke:
I consider the timing of the measurement, in your example 16:00 as the point of data.
I did not test and don't know how it would display when there is only an update every 2 hours. 
Will test if you want.
I you want to revert part of the PR, do you want me to submit one with the reversion code ?

gizmocuz:
@hreintke , no I think it is find. The issue was something else and arjen should not have mentioned this PR
Thanks for responding!

arjendekorte:
Nope, different issue here. I fully agree that it make a lot of sense to align the graphs with the data points for the week and month. For the daily graph, I'm not so sure.

The bars in the daily graph indicate usage not at a certain point in time (or a day in the weekly or monthly graphs), but in a one hour period. Previously it was very clear which one. The graph for 14:00 (usage between 14:00 and 15:00) would be displayed between the 14:00 and 15:00 markers (even when the latter wasn't shown). With this change, it is displayed at the 14:00 marker, making it not immediately clear which one hour period it represents).

Not a big deal, but IMHO aesthetically not pleasing.

hreintke:
@gizmocuz OK, I take no action from my side now.
If you want something changed. just let me know.

gizmocuz:
Thanks, If you have/see other great ideas please continue making PR's 👌

PR #5605, opened 2023-02-16
Removing obsolete OpenZWave (Use ZWaveJS-UI with MQTT-AD)

gizmocuz:


.github/workflows/development.yml
```diff
@@ -19,7 +19,7 @@ jobs:
   # This workflow contains a single job called "build"
   build:
     # The type of runner that the job will run on
-    #runs-on: ubuntu-latest   # Per Dec 2022 latest = 22.04 with GCC 11, OpenSSL3.0 and this breaks building OpenZwave
+    #runs-on: ubuntu-latest   # Per Dec 2022 latest = 22.04 with GCC 11, OpenSSL3.0
     runs-on: ubuntu-20.04
```
kiddigital: Let's also switch to 22.04 now, so we have some time to see if everything plays nice with OpenSSL3 in the coming Beta's

gizmocuz: I will upgrade the workflow, but it's not used in the final builds

MikeyMan83: Why remove this? I have zero appetite to go to the MQTT zwave plugin after testing. And I'm not the only one. Just add a comment to it and leave it in.

MadPatrick: My vote for keeping OZW, but it look like it will be inevitable.
Just leave it out for Ubuntu if this is the problem

gizmocuz: This has mentioned before.
OZW is not being developed anymore.
ZWaveJS-UI is actively developed and working way better and faster.
It is not so hard to switch, done it myself a long time ago.
You do not need to learn all your devices again.
It is a matter of replacing the old with the new devices in Domoticz
Yes it will take some time, but than that is it.

"I am still using Windows 98 because I do not like the new Windows"

Again, for about a year this is mentioned, I suppose to test it asap

I put hundreds (maybe about a thousand) hours in OZW but it's time to say goodbye!

MikeyMan83: Lots of people don't see the benefit of the swap. The comparison to windowd98 is poor to say the least. Its not an OS and there is no harm in keeping it in just like any other plugin. If it works it works. And that it does. With many, many hours of configuration in scripts and automation.

I tried the move and found it to be a step back. And im not the only one. Just getting it set up requires a degree in advanced IT and imo a dedicated mqtt server and takes up much more resources. How would a simple domoticz user ever do it? Honestly, it's less of a hassle to move to HA. 

Replacing the old with the new doesn't work at all and the order all devices is totally messed up. If it ever shows up because there wasn't an mqtt message yet. The whole principe of using a message service got this kind of device integration is wrong.

Oh and btw, half of my devices are not supported. While they are working fine in that old 'windows 98'. So no, its not an improvement in any way. Apart from the theory of being 'developed'. Whatever that may mean.

gizmocuz: The comparison is spot on!

Firstly, ZWave is not for a simple Domoticz user.
Secondly, Installing MQTT via docker compose as well as ZWaveJS-UI is not complicated.

Third, honestly, if you move to HA, you will be using ZWaveJS-UI.

Replacing old with new is working, you will not be the first one doing this.
The IDX of the new device will be the same as the old device.
Even if you use the device names as ID, naming the device correct will make sure your scripts are continue to work

MikeyMan83: Not that hard? It took me several days to get this shit up and running. Compared to originally plugging in a USB stick and adding a plugin. Which once took me ten minutes to set up at most. And by the looks of the reactions on the forum I'm not the only one. The migration manual that user andrehj has written is multiple pages long. And prone to errors.

The comparison isn't spot on at all. The reason for not using win98 anymore is that it's successors are better and or safer. Not the subjective mantra of being supported. 

Like I said, half of my devices is not supported, and the whole mqtt mechanism for this kind of integration is flawed. Is that an improvement?

I have been able to use the MQTT devices in domoticz, but replacing a device that is not available is technically impossible. You can only do that from the dashboard tabs, and guess what happens to devices that are not available.

All well and nice that it is working for you. But there's a broader userbase than just programmers. in theory I understand the whole 'in development' argument, buy at this point I'm not buying for a second that the mqtt solution is an objective improvement at all.

But the whole point was; putting the new stuff in is one thing. Buy why actively break the old stuff?

gizmocuz: So did you install everything via docker compose? If not, here's your problem.

Regarding your comment "not all my devices are supported", I would suggest opening a ticket for this ASAP.
Of course, from a simple user's perspective it's nice to wait until something should work and then take the jump.

Why remove old not-supported stuff from a project? To clean it up. I guess you have to look at this PR to see all the extra shit we need to keep in here.

PR #5613, opened 2023-02-20
Update service-worker.js

arjendekorte:
I see error messages in the console when Domoticz lives in web_root=home (Serviceworker registration failed). Changing the OFFLINE_URL and self.importScrits from absolute to relative paths fixes this.


gizmocuz:
Thanks for spotting and the PR!

Issue #5612, opened 2023-02-20
Thread concurrency issue with asynchronous executeShellCommand


DGA45:
I use executeShellCommand to execute bash scripts every 5 minutes as well on motion sensor devices changes.
A loop is used to execute 4 times the same script with different argument (IP address):
```
local ping = {}
ping['192.168.1.100']= 'Device 1'
ping['192.168.1.101']= 'Device 2'
ping['192.168.1.102']= 'Device 3'
ping['192.168.1.103']= 'Device 4'

for ip, device in pairs(ping) do
  domoticz.executeShellCommand({
      command = '/home/pi/domoticz/scripts/MyPing.sh '..ip..' "'..device..'"',
      callback = 'CallbackPing',
      timeout = 120,
  })
end
```
 
The script can take between 1 and 100 seconds (depends if the device respond quickly or not) and outputs the ping attempts.

From time to time, I observe the output of one script which is interrupted and replaced by the output of another script.
Example:
The normal output is:
```
192.168.1.102
MyPing ----------
PING 192.168.1.102 (192.168.1.102) xx bytes of data
etc...
```
One corrupted output is:
```
192.168.1.102
MyPing ----------PING 192.168.1.101 (192.168.1.101) xx bytes of data
etc...
```
We can see 2 different IP address in the output, while the bash script only handles one.

I also have the following error in the log when it happens:
```
2023-02-20 08:15:52.780 Error: unable to remove file /home/pi/domoticz/scripts/dzVents/data/domscript3032.out
2023-02-20 08:15:52.781 Error: unable to remove file /home/pi/domoticz/scripts/dzVents/data/domscript3032.err
```

I had a look in the executeShellCommand implementation and I noticed that the variable "scriptoutputindex" is not protected against concurrency:
```
scriptoutputindex++;
if (scriptoutputindex > 10000) // should be a big number, to prevent parallel scripts having the same output files. 250 concurrent will probably never be reached
{
  scriptoutputindex = 1;
}
std::string scriptoutputindextext = std::to_string(scriptoutputindex);
```
Thus, if a new executeShellCommand thread is started after "scriptoutputindex++" and before "scriptoutputindextext = std::to_string(scriptoutputindex)", both threads will end with the same used scriptoutputindextext, which would explain my 2 observations:
- 2 threads writing in the same file (output corruption)
- 2nd thread cannot delete the temporary file (already deleted by the first one)

I see 3 possibility to fix this:
- A mutex could be used
[...]

PR #5623, opened 2023-02-28
Fix concurrency issue with scriptoutputindex (#5612)

DGA45:
Replace scriptoutputindex by the thread ID to ensure unique filename


kiddigital:
Thank you!

Issue #5624, opened 2023-03-02
Api call : Did not find suitable Authorization 


PsyOhm23:
Hi,

Since the last 2023.1 version, the authorization header is not detected.

Here is my request from logs : 

```
2023-03-02 22:12:17.337  [72b7b200] Debug: [web:8080] Host:127.0.0.1 Uri:/json.htm?type=devices&used=true&displayhidden=1
2023-03-02 22:12:17.337  [72b7b200] Debug: [web:8080] Request Headers:
Host: mydomain.com
X-Real-IP: xx.xx.xx.xx
X-Scheme: https
Connection: close
Authorization: Basic xxxxxxxxxxxxxxxxxxxxxx==
User-Agent: PostmanRuntime/7.31.1
Accept: */*
Cache-Control: no-cache
Accept-Encoding: gzip, deflate, br

2023-03-02 22:12:17.338  [72b7b200] Debug: [web:8080] Did not find suitable Authorization!
2023-03-02 22:12:17.338  [72b7b200] Debug: Web ACLF: 127.0.0.1 - - [02/Mar/2023:22:12:17.337 +0100] "GET /json.htm?type=devices&used=true&displayhidden=1 HTTP/1" 401 91 - "PostmanRuntime/7.31.1"
```

I've checked [the code from where the `Basic` header is read](https://github.com/domoticz/domoticz/blob/3e50bcbb3f63fe83197b663ad995dda899d59db6/webserver/cWebem.cpp#LL1436C31-L1436C31) 

I should have the log written by the following line no? :
`_log.Debug(DEBUG_AUTH, "[Basic] Found a Basic Auth Header (%s)", ah->user.c_str());`

Instead of that, I have the `Did not find suitable Authorization!` log 

Could you tell me what's wrong ?

Thanks in advance

kiddigital:
I see a Proxy is used.

Is the IP address of the proxy in the Trusted Network?

PsyOhm23:
The local IP address of my proxy is 192.168.1.1 and my trusted network list is 192.168.*

If we can see the Authentication header in the log, I don't understand why the header is not interpreted by Domoticz  ?

Thank you for your time

kiddigital:
Try to add the 'auth' debug flag as well and see what additional info is logged.

kiddigital:
> The local IP address of my proxy is 192.168.1.1 and my trusted network list is 192.168.*
> 
> 
[...]

PR #5625, opened 2023-03-05
Invalid Basic auth API calls should be rejected even in Trusted network

kiddigital:
PR that improves/fixes some unwanted behavior (Issue #5624) that even when calling the API with _invalid_ Basic Auth credentials, the call would be handled when coming from within the Trusted Network but then also with an Admin user and not the User specified in the Basic Auth header.

Now when invalid Basic auth credentials are provided for an API call, the call will be denied.

And when _not_ providing any credentials _but_ coming from within the Trusted network, the calls are allowed.

Also, some additional debugging info (auth-level) is added to easier see what is going on.

### DIFF
--- a/webserver/cWebem.cpp
+++ b/webserver/cWebem.cpp
@@ -2160,24 +2160,38 @@ namespace http {
 				}
 				else if (_ah.method == "BASIC")
 				{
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thank you!

Issue #5470, opened 2022-12-18
1-wire error with multiple DS18B20 temp sensors


McGyver64:
I tried to read 3 different DS18B20 temperature sensors with Domoticz on Raspberry Pi 2B
The temperature value is changing in a 2 sec cycle 13.8C....13,6C....14,7C.......13.8C....13,6C....14,7C........13.8C....13,6C....14,7C........13.8C....13,6C....14,7C etc. So the 3 different temperatures are shown, but not as different sensor, but as one sensor showing different temperature in a 2 sec cycle

Version: 2022.2
Build Hash: eea9db734
Compile Date: 2022-11-05 13:05:35
dzVents Version: [3.1.8]
Python Version: 3.7.3 (default, Oct 31 2022, 14:04:00) [GCC 8.3.0]
_I tried the Beta version because some threads suggest this as a solution._

From the LOG file is clear that all 3 sensors are read.

2022-12-16 08:12:40.827 Status: 1Wire: Added Device: 28-4289811e64ff
2022-12-16 08:12:40.827 Status: 1Wire: Added Device: 28-fbb3811e64ff
2022-12-16 08:12:40.827 Status: 1Wire: Added Device: 28-6065851e64ff

but 1-Wire device IDs of all three sensors have the same last 2 bytes (64ff). Apparently last 2 bytes of 1-Wire device ID are used by Domoticz for temperature senzor device ID. And because they are the same in this case, Domoticz creates only to 1 device instead of 3 devices...

<img width="1592" alt="Schermafbeelding 2022-12-16 om 08 23 29" src="https://user-images.githubusercontent.com/69874589/208291978-227fe337-30ca-48bc-a9f6-479a65d67327.png">

This means that all 3 temperature values from three 1-wire devices (temperate sensors) are reported to this single Domoticz temperature device.

[](url)https://www.domoticz.com/forum/viewtopic.php?p=296107&hilit=1+wire#p296107

McGyver64:
I just installed an older version of Domoticz and the problem is the same

Version: 4.11803
Build Hash: f0aac95ee
Compile Date: 2020-03-21 16:52:51
dzVents Version: 3.0.1
Python Version: 3.7.3 (default, Oct 31 2022, 14:04:00) [GCC 8.3.0]

kiddigital:
@McGyver64 , can you show your Hardware settings?

Try to run Domoticz with additional Debug flags (like Hardware) and see if that provides some more insights.

[...]

PR #5629, opened 2023-03-06
1Wire temperature sensor DeviceIDs modified to use all six bytes

Zuikkis:
This is discussed here:
https://github.com/domoticz/domoticz/issues/5470

Old implementation for 1wire temperature sensors only used 2 bytes of the actual 6 byte ID. This results in duplicates as some sensors have lower bytes set to constant value.

Bytes are now XORed to create 16bit ID.

When scanning the devices, "old style" IDs are corrected in database.

NOTE: This means that if you go back from this version to regular Domoticz, new devices will be created with the old style IDs..

BatteryLevel field in new-style temperature sensors is set to 100 instead of 255. This allows us to detect which devices are already converted, preventing double conversion.

### DIFF
--- a/hardware/1Wire.cpp
+++ b/hardware/1Wire.cpp
@@ -304,11 +304,14 @@ void C1Wire::BuildSensorList() {
 		return;
 
 	std::vector<_t1WireDevice> devices;
+	int n1wireVersion=0;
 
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thanks for this PR and your time! Hope to see more 😂

Issue #5631, opened 2023-03-09
Dimmers forgets it's last level


lwalbert:
Hi,
After updating domoticz from Version: 2022.1 (build 14455) to the latest Domoticz V2023.1 (build 15132), 
I noticed that my dimmers started to forget it's last level.
When switching off dimmer, the LastLevel in DB changed to 0.
when switching on again, the dimmer indicated On at 0% instead of reverting to the last level.

Fault traced, and found that behaviour of dimmer was modified in https://github.com/domoticz/domoticz/pull/5361

in SQLHelper.cpp :

`bool bIsBlindsPercentage = (
				(switchtype == STYPE_Dimmer)
				|| (switchtype == STYPE_BlindsPercentage)
				|| (switchtype == STYPE_BlindsPercentageWithStop)
				);`

the (switchtype == STYPE_Dimmer) is included in bool for blindsPercentage.
Not sure if this is the correct solution, as I don't have blinds.

otherwise, the if-statement at line 5387 should be modified to "&& (bIsBlindsPercentage)" instead of 
"|| (bIsBlindsPercentage)"

from:
`if (
				((bIsLightSwitchOn) && (llevel != 0) && (llevel != 255))
				|| (bIsBlindsPercentage)
				)
			{`

to:
`if (
				((bIsLightSwitchOn) && (llevel != 0) && (llevel != 255))
				&& (bIsBlindsPercentage)
				)
			{`
By removing/commenting out the line  (switchtype == STYPE_Dimmer) in pos: 5382.  
This solved the issue I experienced.

Noticed that several other persons has encountered similar behaviour when searching reported Issues.

Domoticz V2023.1 (build 15132) (c)2012-2023 GizMoCuz
Build Hash: 43a1ec035-modified, Date: 2023-03-08 17:37:01
PluginSystem: Started, Python version '3.7.3'

Br
Lennart

gizmocuz:
Don't experience this issue here.
Did you try with creating a dummy switch, set this to dimmer, set the percentage to 40%, turn off the dimmer (by the button, not the slider) and turn it back on

lwalbert:
Hi,
Apparently it seems to work for dummy switch devices configured as dimmer. LastLevel is keept in database.

But for other hardware dimmer devices (zwave, zigbee), LastLevel changes to 0 in database. 
Somehow they manage to retain to last state when switched on again. 
[...]

PR #5636, opened 2023-03-11
Update SQLHelper.cpp

lwalbert:
Prevent lastlevel update when dimmers are switched OFF

reference to issue https://github.com/domoticz/domoticz/issues/5631


gizmocuz:
Thanks for the PR! Good spot, this should not have been here!

Issue #5581, opened 2023-02-02
Huge value of incremental counter


ZonPV:
Reading the gasmeter with a optical tracker sensor , TCRT5000 to a NodeMCU ESP8266 with ESPeasy version ESP_Easy_mega_20220809.. I selected a pulse counter in that device.
It works for years without any problem. The only thing is that every morning the sensor on the dashboard of Domoticz shows a very high value: 184467440737095520.000 m³. In a graph there is no pile. The value at that moment is 0.
As soon as I use some gas the daily counter starts at 0,01 m³ and shows the right value for the rest of the day.
It is a dummy, virtual sensor, Type General, SubType Counter Incremental, Field Data shows the correct total value of 36501.137 m³. Exactly as the physical gasmeter itself.

The actual build of Domoticz is 15006. I can tell this cosmetic failure is there for years, so in V2022.2, V2022.1, and so on.

It is not related to a browser. It is there in Firefox, Chrome and on all my devices like PC, smartphone or tablet.

I do not have a smart meter for reading electricity. Still the oldfashioned ferraris meter. No P1-reading.

Running Domoticz in Docker on my Synology NAS.
In the log, after the value is send to InfluxDB is shown: 
`2023-02-02 11:54:54.378 InfluxLink: value Counter-Incremental,idx=734,name=Gasmeter value=3650113.7`

Found no way in this forum to add a screenshot; sorry for that.

It is not a dramatic issue. I just want to tell it as feedback and I love Domoticz. Just to make the application better. ;-)

waltervl:
Please use the Forum for this type of issues. It seems that your NodeMCU ESP8266 with ESPeasy is not connecting to Domoticz correctly. Could still lead to a bug in Domoticz but there is a far bigger audience for these kind of issues then on this github.

Zuikkis:
This is actually a domoticz bug. I have a lot of incremental counters. They all move slightly backwards at midnight, and because domoticz counters don't handle negative numbers it shows very large positive integer instead.

It's fixed automatically whenever the counter is incremented forwards again.

gizmocuz:
I use counter incrementals myself and don't have these issues

Maybe you could adopt my watermeter code (MQTT Autodiscovery) to work with your optical sensor?

https://github.com/gizmocuz/esp_proximity_sensor_mqtt

When you make a screenshot, you can just paste it in here (control-v)

[...]

Issue #5544, opened 2023-01-15
request for price/cost device (similar to kWh but with configurable currency code)


WillemD61:
This is an enhancement request for a dummy device that can be used to track prices and/or costs, positive and negative whereby the user can indicate the currency used.

With the increased/increasing focus on energy cost such a device is really missed in Domoticz. Currently workarounds have to be used such as storing such prices/costs on kWh devices.

It can be a simple device, i.e. no calculation based on tariffs as suggested in https://github.com/domoticz/domoticz/issues/2132. 
Any calculation should be left to the user using dzVents scripts.

The device only needs to receive and display increasing, decreasing, positive and negative values, be able to display those in a graph and calculate daily totals for display in weekly, monthly and comparison graphs.

Currently one cannot store negative values on counter devices and the counter graphs cannot handle negative or even decreasing values. The only device that can handle it is a kWh device (General, kWh). The functionality of a kWH comes very close to the desired functionality of a cost device, but the new device should allow setting of the currency unit.
Required for a currency device ( and partly already possible on a kWh device):
1) Possible to increase and decrease in value.
2) Possible to handle negative values.
3) Intraday graph of type bar or line.
4) Graphs with daily values display bars of total (so no min,max or avg). The device values/counter should therefore be stored in the meter and meter_calendar table.
5) Daily switchover and insert into meter_calendar based on difference between first and last value of the day (like kWh), not on min/max, like counters.
6) Update of the device similar to updateElectricity(0,value) of a kWh device, i.e. no calculation based on time-based-rate.
7) It would be nice to have the option for DisableLogAutoUpdate and AddDBLogEntry so one can also enter future values via JSON calls in meter tables without affecting meter_calendar entries.

This request is based on the analysis shown here https://www.domoticz.com/forum/viewtopic.php?p=296335#p296335 and the fact that I am tired of using kWh devices as workaround. It is confusing/annoying and pollutes my Domoticz setup. As said, with increasing focus on costs and increasing use of hourly electricity prices, solar power, battery storage (resulting in negative/decreasing values), I am sure many more people struggle with this.

gizmocuz:
How about using a custom sensor?

gizmocuz:
What you might want to do is use the InfluxDB pusher and push the kWh (counter) and Price. Then you can create a chart with cost per hour/day

WillemD61:
> How about using a custom sensor?

[...]

Issue #5436, opened 2022-11-30
incorrect handling of negative values in counter day graph


WillemD61:
I am using a counter device (RFX meter counter, divider 100) to store costs in Euro. The costs are updated every hour and can increase and decrease. (I don't see other suitable devices for cost tracking. I have tried custom devices but these don't show daily totals, only max,min and avg)

Bug report: The increase/decreases are stored corrrectly in the meter table but are not displayed correctly in the day graph. The day graph shows the countertotal plus counterdecrease, for example at 11:00 it shows 5.09-0.03=5.06 instead of a -0.03.

In the example below the values of 11:00 to 14:00 are incorrectly displayed.

Notes:
* simliar errors might occur in weekly and monthly graphs, but I don't have negative values there yet.
* in below overview only values at 00 minutes are shown, the values in the meter table from 05 to 55 minutes are present in the table but not shown in below overview
* the device has a divider of 100
* the daily total shown in the device icon is correct (i.e. matching the table values)
* the "usage last 24 hour" graph is incorrect, but the "usage last week" is correct (i.e. showing correct daily total for current day)

Below is a table showing the values displayed in the day graph with next to it a column showing the values from the meter table and finally the calculated difference from one meter value to the other.

```
Displayed in graphs (xls download)			value in meter table	(difference to next value)/100
DateTime	Costs			
2022-11-29 18:00:00	0,27					261				0.27
2022-11-29 19:00:00	0,36					288				0.36
2022-11-29 20:00:00	0,25					324				0.25
2022-11-29 21:00:00	0,2					349				0.2
2022-11-29 22:00:00	0,16					369				0.16
2022-11-29 23:00:00	0,05					385				0.05
2022-11-30 00:00:00	0,21					390				0.21
2022-11-30 01:00:00	0,21					411				0.21
2022-11-30 02:00:00	0,04					432				0.04
2022-11-30 03:00:00	0,04					436				0.04
2022-11-30 04:00:00	0,04					440				0.04
2022-11-30 05:00:00	0,04					444				0.04
2022-11-30 06:00:00	0,06					448				0.06
2022-11-30 07:00:00	0,07					454				0.07
2022-11-30 08:00:00	0,21					461				0.21
2022-11-30 09:00:00	0,22					482				0.22
2022-11-30 10:00:00	0,05					504				0.05
2022-11-30 11:00:00	5,06					509				-0.03
2022-11-30 12:00:00	4,95					506				-0.11
[...]

PR #5637, opened 2023-03-12
Counters modified to support negative increments and negative totals.

Zuikkis:
Previous implementation was a bit flawed as nothing prevented you pushing negative values to the database, but then all graphs were calculated with unsigned 64bit integers. Negatives were shown as about 2E19 positive integers.

Since no one probably really has such large counter values, it is all the same to use signed 64bit in calculations which still gives 1E19 positive range but does not choke on negatives.

Another issue was that at midnight daily totals were calculated using MIN/MAX of daily values. This works for regular positively incrementing counters, but if negative increments are allowed, First/Last value of the day should be used instead.

All of these changes were already done for kWh counters several years ago, this commit just expands same behaviour to other type of counters.

This is related to issues:
https://github.com/domoticz/domoticz/issues/5544
https://github.com/domoticz/domoticz/issues/5436
https://github.com/domoticz/domoticz/issues/5581

### DIFF
--- a/main/SQLHelper.cpp
+++ b/main/SQLHelper.cpp
@@ -7273,7 +7273,8 @@ void CSQLHelper::AddCalendarUpdateMeter()
 
 			// if kwh counter => total_min = first value of the day, and total_max = last value of the day
 			// because last value can be lower than first value when consumed energy is negative (e.g. photovoltaic produces more than building usage)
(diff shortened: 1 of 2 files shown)


gizmocuz:
Thanks for the PR and your time!!

PR #5641, opened 2023-03-14
Flowchart fixes for SECURITY_SETUP.md

ferdnyc:
Two commits:

1. Un-hide Mermaid graph to prevent breakage
   It's a [reported issue][1] that Mermaid graphs containing labeled links (lines with text on them) fail to render inside a `<details>` section. If not collapsed, they render fine.

2. Clean up, simplify flowchart

    - There's no need for the inner "UIs" subgraph, if the flowchart is configured `TD` instead of `TB`. (I don't know why that matters, and it probably shouldn't. Seems to be a bug. But, it matters.)
    - "User Interfaces" has no apostrophe in it

[1]: https://github.com/mermaid-js/mermaid/issues/3504


gizmocuz:
Thank you for this!

PR #5643, opened 2023-03-16
Recheck and fix int64_t and uint64_t print formats

efdeb:
Just rechecked int64_t and uint64_t print formats in all code, and fixed some of them.
Have no more compilation warning about them on Ubuntu 20.04 LTS 64 bits.

### DIFF
--- a/hardware/EvohomeWeb.cpp
+++ b/hardware/EvohomeWeb.cpp
@@ -709,8 +709,9 @@ void CEvohomeWeb::DecodeDHWState(temperatureControlSystem* tcs)
 		}
 		else if ((result[0][1] != szId) || (result[0][2] != ndevname))
 		{
+			uint64_t DevRowIdx = (uint64_t) std::stoull(result[0][0]);
 			// also wipe StrParam1 - we do not want a double action from the old (python) script when changing the setpoint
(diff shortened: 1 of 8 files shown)


hardware/EvohomeWeb.cpp
```diff
@@ -709,8 +709,9 @@ void CEvohomeWeb::DecodeDHWState(temperatureControlSystem* tcs)
 		}
 		else if ((result[0][1] != szId) || (result[0][2] != ndevname))
 		{
+			uint64_t DevRowIdx = (uint64_t) std::stoull(result[0][0]);
```
gizmocuz: I don't think you need to use (uint64_t) here as it is already a uint64_t

efdeb: In fact, std::chrono::milliseconds is a int64_t, not a uint64_t. That's the reason why i converted. Anyway, this is not strictly needed as the conversion will be implicit when assigning the value to the uint64_t variable.

Up to you : just tell me.

gizmocuz: True, but std::stoull is already returning a uint64_t, so casting to this seems double (compiler will fix this, but doesn't look that pretty)

efdeb: Sorry, read too fast : I thought your comment concerned another fix.
I remove the cast, although std::stoull returns **unsigned long long** and not exactly **uint64_t**.

gizmocuz:
Thank you again for this PR and time!!

PR #5644, opened 2023-03-16
Auto enable debug and prevent crash

kiddigital:
Small PR that

- Automatically enables 'loglevel' `debug` when the `-debuglevel <somelevel>` is specified
- Prevent crash of Domoticz when for some reason (like too high DB version) the mainworker could not start


gizmocuz:
Thank you!!

PR #5653, opened 2023-03-24
WriteToHardware now returns true to allow events to happen (Fixes #5650)

### LINKED ISSUE #5650: CHttpPoller::WriteToHardware is not implemented
Domoticz version: 2023.1 build 15163.

Please see https://www.domoticz.com/forum/viewtopic.php?t=40110 including the post in there with my analysis: https://www.domoticz.com/forum/viewtopic.php?p=301077#p301077

That forum topic describes a valid scenario for having a virtual switch device as a child of HTTP/HTTPS poller hardware.
However, that doesn't work because CHttpPoller::WriteToHardware is not implemented.
It looks like this can be fixed by reusing the implementation from CDummy::WriteToHardware.

kiddigital:
Small PR to fix issue #5650

### DIFF
--- a/hardware/HttpPoller.cpp
+++ b/hardware/HttpPoller.cpp
@@ -48,7 +48,7 @@ void CHttpPoller::Init()
 
 bool CHttpPoller::WriteToHardware(const char* /*pdata*/, const unsigned char /*length*/)
 {
-	return false;
+	return true;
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thank you for this PR!!

PR #5662, opened 2023-04-03
Check HTTP headers case-insensitive (RFC7230)

### LINKED ISSUE #5661: Domoticz seems case sensitive when checking X-Real-IP and X-Forwarded-For
I am having a problem with pfsense, haproxy and domoticz.

Domoticz would always show the pfsense internal ip, not the external ip of the connected client in the log. 

I have found that haproxy rewrites X-Real-IP and X-Forwarded-For to all lower case.
This also affects security since I had my internal pfsense ip whitelisted since merge of proxy and client whitelist.

I did some wiresharking and found this:
haproxy:
    x-real-ip: 83.185.83.xxx\r\n
    x-forwarded-for: 83.185.83.xxx\r\n

Synology/nginx:
    X-Real-IP: 83.185.83.xxx\r\n
    X-Forwarded-For: 83.185.83.xxx\r\n

When domoticz searches for X-Real-IP it finds nothing when client is connected through haproxy. In my case leaving my domoticz installation wide open to the internet without password protection.

I would like to request that the search is non-case-sensitive and that proxy ips has its own box again. It is really dangerous to combine it with client whitelist, requiring no password if proxy is not configured properly.

For others with the same issue there is a temporary fix in haproxy until servers get up to spec with RFC7230 and HTTP/2:
https://cbonte.github.io/haproxy-dconv/2.2/configuration.html#h1-case-adjust

kiddigital:
Small PR which treats Proxy HTTP header case-insensitive as per RFC7230.

Fixes #5661

### DIFF
--- a/webserver/cWebem.cpp
+++ b/webserver/cWebem.cpp
@@ -1219,7 +1219,7 @@ namespace http {
 			std::vector<std::string> headers;
 			std::vector<std::string> hosts;
 
-			if (sumProxyHeader("Forwarded", req, headers))
+			if (sumProxyHeader("forwarded", req, headers))
 			{
 				// We found one or more Forwarded headers that need to be processed into a list of Hosts
 				if (!parseForwardedProxyHeader(headers, hosts))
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thanks for this PR!!

PR #5669, opened 2023-04-17
Fix Visual Crossing hardware cannot update

### LINKED ISSUE #5670: Visual Crossing small issues
@Olifant1990 
Thanks for the PR to implement Visual Crossing as alternative for Darksky. As it is implemented now in latest beta we saw some small isues. See also Forum topic https://www.domoticz.com/forum/viewtopic.php?p=302113

- Wind device and Visibility device got the name Unknown"
- The mixture of barometer device with temp/humi device (and the naming THB)
- Forecast is still going to old Darksky (forecast.io) and so still giving an error

Is it also possible to use the location of setup - settings as default? eg by leaving the location field empty or fill in a 0 (zero)

Olifant1990:
Fix Visual Crossing hardware cannot update

### DIFF
--- a/hardware/VisualCrossing.cpp
+++ b/hardware/VisualCrossing.cpp
@@ -323,7 +323,7 @@ void CVisualCrossing::GetMeterDetails()
 		at10 -= (tsen.WIND.chillh*256);
 		tsen.WIND.chilll = (BYTE)(at10);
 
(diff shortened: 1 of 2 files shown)


Olifant1990:
Last commit fixes default name problems from https://github.com/domoticz/domoticz/issues/5670

kiddigital:
Thx!

PR #5672, opened 2023-04-19
OpenWebNet: add energy management actuators

Spugna85:
added registration request to receive data from Energy Management Actuators


gizmocuz:
Thank you!

PR #5674, opened 2023-04-23
Add timesync check script to domoticz.sh

### LINKED ISSUE #5673: [Enhancement] Cold boot crash solution for systems without RTC
In version 2023.1, the issue where Domoticz will trigger a watchdog timer on a cold boot when here is no RTC (real time clock) and the time is being synced while Domoticz is starting up/running is still present. Checked on a clean install of Raspberry Pi OS Lite Bullseye with Domoticz installed using the `sudo bash -c "$(curl -sSfL https://install.domoticz.com)"` method.
In the [troubleshooting wiki](https://www.domoticz.com/wiki/Troubleshooting#Domoticz_stops_after_cold_boot) there is a mention about it should be fixed in 2022.2 stable, but it doesn't seem like it is. 
Also, the workarounds in the wiki either act upon Domoticz crashing first and then restarting it, installing other software/hardware (ntpdate, RTC module) or using a startup delay which does not check anything.

Another [workaround](https://github.com/domoticz/domoticz/issues/3371#issuecomment-514216432) as created by @agambier in #3371 also relies on the Domoticz service crashing first.
While these workarounds might be viable for many people, I wanted to try and prevent a crash instead of acting upon a crash.

I have come up with an addition that can be placed at the top in the `do_start()` function in the domoticz.sh script.
It first checks if the system is configured to sync to a timeserver by checking if `systemd-timesyncd` is installed and running, then if the time is already synced by checking if the file `/run/systemd/timesync/synchronized` exists indicating a successful sync (for ex. at a Domoticz service restart or system reboot), and if not it waits at most 30 seconds for the time to be synced and then start Domoticz.

```
# Check if systemd-timesyncd is enabled and running, otherwise ignore script
if systemctl --quiet is-enabled systemd-timesyncd && systemctl --quiet is-active systemd-timesyncd; then
    # Check if time is already synced, if so start Domoticz immediately
    if [ -f "/run/systemd/timesync/synchronized" ]; then
        printf "Time is already synced, starting Domoticz...\n"
    else
        # Check if custom time server(s) are defined
        if grep -q "^#NTP=" /etc/systemd/timesyncd.conf; then
            printf "Note: No time server(s) defined in /etc/systemd/timesyncd.conf, using default fallback server(s)\n"
        fi
        # Wait at most 30 seconds until sync file is generated indicating successful time sync
        printf "Waiting for time synchronization before starting Domoticz"
        count=30
		while [ ! -f "/run/systemd/timesync/synchronized" ]
        do
            count=$((count-1))
            if [ $((count)) -lt 1 ]
            then
		#If failed, print error message, exit loop and start Domoticz anyway
                printf "\nTime sync failed, check network and /etc/systemd/timesyncd.conf\n"
                printf "Starting Domoticz without successful time sync...\n"
                break
            fi
            printf "."
            sleep 1
        done
	#If the file was found in time, sync was successful and Domoticz will start immediately
        if [ -f "/run/systemd/timesync/synchronized" ]
        then
            printf "\nTime synced, starting Domoticz...\n"
        fi
    fi
else
    printf "Note: systemd-timesyncd is not enabled or running, starting Domoticz without time sync check...\n"
fi
```

I have specifically omitted any exit conditions to avoid Domoticz not starting when the script can't find the file or with other issues. My thought here is that, when the time can't be synced within 30 seconds, chances are there is a connection issue or something alike and the time might even never be updated after that. In that case Domoticz will not crash anyway since the time will not change and should start regardless.

Please let me know if this is a viable solution, if so I can make a pull request.

bonzi9:
This adds a script in the do_start() function to wait 30 seconds, or less until the file /run/systemd/timesync/synchronized is found indicating a successful time sync before starting Domoticz. This is to allow systems without a real time clock to finish synchronizing to a time server, because Domoticz can crash when the time change is larger than a set watchdog timer. The script also checks if systemd-timesyncd is active and running before executing, to avoid running it on systems that do not use it. When the time is already synchronized, the script will start Domoticz immediately. The script will always allow Domoticz to start, even if the sync fails, and all events generate a message for the user.

Resolves #5673

### DIFF
--- a/domoticz.sh
+++ b/domoticz.sh
@@ -53,6 +53,46 @@ pidof_domoticz() {
 #
 do_start()
 {
+
+# Script that waits for time synchronisation before starting Domoticz to prevent a crash
+# on a cold boot when no real time clock is available.
+
+# Check if systemd-timesyncd is enabled and running, otherwise ignore script
+if systemctl --quiet is-enabled systemd-timesyncd && systemctl --quiet is-active systemd-timesyncd; then
+    # Check if time is already synced, if so start Domoticz immediately
+    if [ -f "/run/systemd/timesync/synchronized" ]; then
+        printf "Time is already synchronized, starting Domoticz...\n"
+    else
+        # Check if custom time server(s) are defined
+        if grep -q "^#NTP=" /etc/systemd/timesyncd.conf; then
+            printf "INFO: No time server(s) defined in /etc/systemd/timesyncd.conf, using default fallback server(s)\n"
+        fi
+        # Wait a maximum of 30 seconds until sync file is generated indicating successful time sync
+        printf "Waiting for time synchronization before starting Domoticz"
+        count=30
+		while [ ! -f "/run/systemd/timesync/synchronized" ]
+        do
+            count=$((count-1))
+            if [ $((count)) -lt 1 ]
+            then
+		        #If failed, print error message, exit loop and start Domoticz anyway
+                printf "\nWARNING: Time synchronization failed, check network and /etc/systemd/timesyncd.conf\n"
+                printf "Starting Domoticz without successful time synchronization...\n"
+                break
+            fi
+            printf "."
+            sleep 1
+        done
+	    #If the file was found in time, sync was successful and Domoticz will start immediately
+        if [ -f "/run/systemd/timesync/synchronized" ]
+        then
+            printf "\nTime synchronization successful, starting Domoticz...\n"
+        fi
+    fi
+else
+    printf "INFO: systemd-timesyncd is not enabled or running\nINFO: Starting Domoticz without waiting for time synchronization...\n"
+fi
+
         # Return
         #   0 if daemon has been started
         #   1 if daemon was already running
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thank you very much for this PR.
I made some small changes to the indenting and text.
I also removed the below line

`# printf "INFO: systemd-timesyncd is not enabled or running\nINFO: Starting Domoticz without waiting for time synchronization...\n"
`

Because I suspect that everyone using a RTC will otherwise see this message?

If all is OK, please let me know and I will merge this.

bonzi9:
Thanks for reviewing and your note on the message if `systemd-timesyncd` is not running. I agree the message is not needed in case it is not used.
Using your commit with the indent and text changes, I committed a change where I deleted the message and related else statement.
For me it is ok, you can merge it if you agree with my last commit.

_I tested the updated code from my last commit on my own Domoticz 2023.1 installation and it works as expected. On a cold boot, it takes about 9-10 seconds (9-10 dots after the "Waiting for..." line) for the time to be synced, and that is on a single core Raspberry Pi Zero W. I expect the max. 30 second wait is plenty but not too long for most users._

gizmocuz:
Merged! Thank you for this PR!!!

PR #5678, opened 2023-04-24
Don't record NULL reads of a DS2423 1-Wire counter as a zero, NULL should be disgarded and re-read on the next interval.

simat-git:
Don't record NULL reads of a DS2423 1-Wire counter as a zero, zero can be a valid read, but a NULL should be disgarded and re-read on the next interval.


gizmocuz:
Thank you for this PR!!

PR #5682, opened 2023-04-29
Check for 32-bit userland on aarch64

### LINKED ISSUE #5679: Install script selects incompatible arch on Raspbian aarch64 with 32-bit userland
Reported by several users at https://www.domoticz.com/forum/viewtopic.php?t=40198

The installer script uses `uname -m` to determine the OS architecture:
```
MACH=`uname -m`
```

However, this returns the architecture of the kernel and it's possible to have a different architecture for userland binaries. Because of this, the installer script downloads the wrong binaries. For me, this resulted in a broken state when upgrading to the 2023 builds.

You can detect the userland architecture with `dpkg --print-architecture`.

For example, on my RPi 4:
```
pi@raspberrypi:~ $ uname -m
aarch64
pi@raspberrypi:~ $ dpkg --print-architecture
armhf
```

The fix would need to be in the installer script.
I'm not enough of a Linux expert to know how to map all the possible outputs of `dpkg --print-architecture` to the relevant binaries.
For example, it's clear for my case that "armhf" should map to the "armv7l" download. But what other architectures are possible?

Perhaps there's a better way to figure out the pointer width (32bit versus 64bit) of user-land binaries from a shell script.

pharring:
Fixes #5679

### DIFF
--- a/History.txt
+++ b/History.txt
@@ -10,6 +10,7 @@ Version 2023.xx
 - Fixed: Counter day charts, better hourly bars and usage
 - Fixed: Device reordering when a room plan was selected
 - Fixed: Forecast button now displays a working forecast
+- Fixed: Install and update fixed for Raspberry Pi 4 with 32bit (armhf) userland
 - Changed: MQTT-AD, better Unit calculation for scene/action devices
 - Removed: native GoodWe hardware (use the new GoodWe python plugin instead)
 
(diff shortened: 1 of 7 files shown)


gizmocuz:
Thanks for this great patch! Looking good! Just going to perform some tests tomorrow as well.

gizmocuz:
Thanks!!

PR #5688, opened 2023-05-07
Add helm chart to deploy domoticz in kubernetes clusters

jpradoar:
* Add: Helm chart to deploy Domotics on kubernetes
* Add: Small readme to know how to operate

The objective is to deploy an instance of Domotics on kubernetes cluster and can access it via load-balancer url (in the future could be use a real dns).
This is usefull because I can use it like "SaaS"

If you consider this job as an interesting, I can impruve all deployment and add more usefull configurations.

Thanks a lot!
Jonathan

### DIFF
--- a/helm-chart-k8s/.helmignore
+++ b/helm-chart-k8s/.helmignore
@@ -1,0 +1,23 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
+.DS_Store
+# Common VCS dirs
+.git/
+.gitignore
+.bzr/
+.bzrignore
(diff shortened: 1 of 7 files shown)


gizmocuz:
Looking good! I got some questions about this as I am very new to this.
Is it possible to join our slack channel and send me a message referring to this topic?

gizmocuz:
I think you can usebthis chart?

https://artifacthub.io/packages/helm/truecharts/domoticz

jpradoar:
> I think you can usebthis chart?
> 
> https://artifacthub.io/packages/helm/truecharts/domoticz

Ok, perfect. I will use it and modify my deployment to add all updated values

PR #5690, opened 2023-05-08
Add servercert.pem save/restore if non default file found (user cert)…

ghost:
… to updatebeta only for now (TBD: Do the same in updaterelease if no side effect).

### DIFF
--- a/updatebeta
+++ b/updatebeta
@@ -3,6 +3,9 @@
 # This script can be used to upgrade to the latest beta version
 # To skip making a backup, use -nobackup as commandline argument
 
+# To update if default servercert.pem change someday...
+MD5_DEFAULT_SERVERCERT="7671292113c501d9b432c825c09b693a"
+
 lowercase(){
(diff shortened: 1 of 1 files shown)


gizmocuz:
Or you could use another filename for your certificate, and modify the Domoticz startup parameters to use this file?

waltervl:
So it is better to modify the instruction in wiki as it is stating there that the original file name should be used? https://www.domoticz.com/wiki/Native_secure_access_with_Lets_Encrypt#Add_the_certificate_to_Domoticz

So better to use a new filename and modify /etc/init.d/domoticz.sh with -sslcert path_to_cert_file

ghost:
Hi Gizmocuz, That's indeed better to use startup params... Il just did not noticed this was possible using -sslcert. Anyway the PR had to be modified as there was a mistake in pushed file so better close it if useless!

gizmocuz:
@waltervl , yes, otherwise with each update that file is overwritten

waltervl:
Wiki is updated.

PR #5694, opened 2023-05-17
Proxy handling: Also remove quotes from IPv4 addresses (fixes #5683)

### LINKED ISSUE #5683: Use of a proxyserver Hiawatha
Sorry that I write an issue here but my account (PeterB1) in the forum has been disabled. Trying to change my e-mail address something went wrong. For some reason I cannot contact the forum admin to reset my account. 
Case: I am using the Hiawatha webserver (https://www.hiawatha-webserver.org/) as proxyserver. This worked with the n-1 version of Domoticz.  I have upgraded Domoticz to the last stable version. After that, the proxy does not work any more (403 Forbidden).
I have add the IP adress (192.168.2.65) of the webserver in the trusted networks field an I have created an user with admin rights.  

Using the proxyserver I get the error message in Domoticz:
`Unable to determine origin due to improper proxy header(s) (values) being used (Possible spoofing attempt!?), dropping client request (remote address: 192.168.2.65)`

I have contacted the developer of the Hiawatha-server. He asked me to find out what exactly has been changed in managing the request. I am not a developer and what to continue using Hiawatha.

kiddigital:
When a Proxy uses the 'forwarded'-header, the 'for'-field contains the IP-address of the source. For IPv6 addresses, these can be quoted and between brackets ([).

But it seems also allowed to put IPv4 addresses between quotes. It was found that the Hiawatha-proxy server puts the IPv4 addresses between quotes. This was not handled correctly as Domoticz only checked for quotes for IPv6 addresses.

This is now changed so both IPv6 and IPv4 addresses can be quoted and will be handled correctly.

This fixes #5683

### DIFF
--- a/webserver/cWebem.cpp
+++ b/webserver/cWebem.cpp
@@ -1165,13 +1165,14 @@ namespace http {
 
 			std::string cleanIP = stdstring_trimws(ip);
 			bool bIsIPv6 = (cleanIP.find(':') != std::string::npos);
+			// IPv6 and IPv4 addresses can be written as quoted strings
+			if (cleanIP.find('"') != std::string::npos)
+			{
+				cleanIP = cleanIP.substr(1,cleanIP.size()-2);	// Remove quotes from begin and end
+			}
 			if (bIsIPv6)
 			{
 				// IPv6 addresses can be written as quoted strings and between brackets (See RFC5952)
-				if (cleanIP.find('"') != std::string::npos)
-				{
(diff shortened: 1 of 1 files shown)


gizmocuz:
front and back functions are not exception safe (for example when the string is empty and you compare this to a character you will get an exception)
In c++20 we get **starts_with** and **end_with** functions for strings

How sure are we that cleanIP/ip is not empty? (can't see this by looking at this PR alone)

kiddigital:
function starts with
```
			if (ip.empty())
				return false;
```
:)

gizmocuz:
Thanks for this PR!!

PR #5697, opened 2023-05-21
Refactor/webserversimplification

kiddigital:
This PR is an attempt to simplify- and reorganize the current webserver implementation.

Over the years, many different methods to implement all kinds of functionality have been used. This is a start to make things more straight forward and moving to an 'API oriented' implementation.

This PR (and probably more will follow over time) does the following:

- Show (in DEBUG mode) the variety of methods used
```
Debug: WebServer(SSL) started with: 219 Commands - 35 RTypes
Debug: cWebEm Registration: 25 pages, 12 actions, 2 includes, 4 whitelist urls, 11 whitelist commands
```

- The old `handlecommand` function (now renamed to 'handlecommandparam' and moved to its own file 'WebServerCommands.cpp') also has over 80! _commands_ in a 4500+ lines long single function
- Removed the ancient (and not used) JSONP support
- Removed the UTF-16 'wide-character' support as it is not used anywhere
- Removed all 4 ServerSide includes
  - languages now uses the GetLanguages command
  - TimerTypes was not used. A Cmd_GetTimerTypes has been added in case in future it is needed
  - MeterTypes and SwitchTypes now use Cmds as well
- Changed **all** 30+ RTypes to proper Commands
  - Settings -> Cmd_GetSettings
  - Users -> Cmd_GetUsers
  - Devices -> Cmd_GetDevices
  - Hardware -> Cmd_GetHardware
  - Cameras -> Cmd_GetCameras
  - Scenes -> Cmd_GetScenes
- Some old RTypes are still recognized and mapped to their corresponding commands in case other clients use them, but also generate a message in the log to use the command instead!

And quite some moving code around to group things for better overview.

**Found possible issues**
- Seems that 'www/app/timers/factories/js' has its own `getTimerTypes` and not uses the backend! _Should it be rewritten to use the new command?_ Now the UI and backend maintain their own list that could ran out-of-sync!
- Command `createvirtualsensor` (Cmd_CreateMappedSensor in 'hardware/Dummy.cpp') is not used anywhere in the UI. Seems to be replace by the newer `createdevice` (also in Dummy.cpp). _Only used in DzVents at the moment._


gizmocuz:
Thank you very much for all this work!!! Much appreciated!

waltervl:
This PR introduced some strange log messages on depricated RType  see forum https://www.domoticz.com/forum/viewtopic.php?f=6&t=40415

kiddigital:
> This PR introduced some strange log messages on depricated RType  see forum https://www.domoticz.com/forum/viewtopic.php?f=6&t=40415

Hi @waltervl , it is not a strange messages but intended 😁

We are cleaning up some old code and inconsistencies and the replacement of 'RType' api calls to regular 'command' type is one of them.

Maybe it is a good idea to already mentioned this on the Wiki? At least for the next Stable the Wiki has to be updated anyway.

The message should not appear when using the Domoticz UI. If they do, that is most likely a caching issue.

For custom scripts, interfaces, UI's, etc., they should start to transition to the new calls. Therefor some of the old RType calls are still working but generate this message.

waltervl:
@kiddigital Is it easy for you to list all deprecated RTypes? Then I will update the wiki, and perhaps also in the forum... I foresee users and other developers asking questions about the API changes (What is the successor of RType XX?)
Or add the correct API to the deprecation warning in the log....

kiddigital:
@waltervl , I have added the list to the [forum topic](https://www.domoticz.com/forum/viewtopic.php?p=303581#p303581) about this issue.

Issue #5622, opened 2023-02-27
Github - Add built releases as assets in github versions


Krakinou:
Hello,
It's not really a bug but I don't see where to make this request otherwise : 
Would it be possible to add the built tgz released files (downloadable from the [Domoticz Release page](https://www.domoticz.com/releases/?dir=./release) ) as asset in the assets available for each version in the [github versions](https://github.com/domoticz/domoticz/releases)?
Or alternatively, to keep those previous version for a while in the [Domoticz Releases page](https://www.domoticz.com/releases/?dir=./release)?

The reason why I'm asking this is that I have packaged domoticz for [Yunohost](https://github.com/YunoHost/yunohost) and not being able to control the version installed (as only the last stable one is available at any moment in time) causes me quite some problems.
Also, it would allow people to install a previous version should they need it for whatever reason.
Thanks!

kiddigital:
Hi @Krakinou , the build services (both Windows and Linux) create downloadable artifacts so you can download the different Beta's and even the PR builds.

waltervl:
Those artifacts expire after 14 days.

kiddigital:
Older releases can be found (and downloaded) at https://github.com/domoticz/domoticz/releases

waltervl:
Those are only the sources, not the build files.
Best guess would be the docker environment, perhaps you can extract the built files from there. Only for example 2022.2  stable became by accident the same as 2023.1 stable. So next best option would be to build the sources yourself. Or create your own archive of stable releases.

Krakinou:
Hi!
Thanks for your answers! As stated by @waltervl, only the source files are available on the github repo, and having to build the binaries locally on each install may be quite a challenge on small hardware (like old Rasp), not even talking about all the dependencies issues to get to build them on different OS/version/images/etc...
My idea was just to have a way to download the previous build files, not only the current ones, so either:
- by having it as assets in the github version (tbh, I don't have the faintest idea of the amount of work required to do that, but I can see that a lot of github project seems to be doing it, for example [UMS](https://github.com/UniversalMediaServer/UniversalMediaServer/releases)
- or just by keeping the previous release on the domoticz release page, just by changing the file name for each release (eg : domoticz_linux_aarch64_20231.tgz instead of just domoticz_linux_aarch64.tgz ).

 Again, I don't know the amount of work involved for any of these options...
As for creating my own archive... why not... but how the hell do I do that?? :)

gizmocuz:
[...]

PR #5700, opened 2023-05-25
Add built releases as assets in github versions

Krakinou:
Hello:
Related to #5622 
This add a github action that will build and upload asset for x86_64, armhf and aarch64 architecture each time a new release is created.
It is inspired from the [development.yml](https://github.com/domoticz/domoticz/blob/development/.github/workflows/development.yml) action and uses [uraimo/run-on-arch-action](https://github.com/uraimo/run-on-arch-action) to emulate alternative architecture.
It takes into account the phase-out of Openzwave .

**Prerequisite**: The Workflow permissions should be "Read and Write" in the action settings of the repo:
image

**How to** : 
The action needs to be run manually once via workflow_dispatch event, this will create packages in the repository that contains a docker container with everything required to compile the assets.
This may run a few hours as cmake need to be built for armv7 and as it run in qemu it is far slower as a native x86_64. This has to be done in a separate job as github limit the duration of a job to 6h and compiling both cmake and domoticz will take longer.
Once this is done, each new Github Release will trigger the action that will build 3 assets and attach them to the release. 

You may find an example for a release and a pre-release on my repo that I used to test the action: https://github.com/Krakinou/domoticz_build_on_arch/releases

### DIFF
--- a/.github/workflows/build_arch.yml
+++ b/.github/workflows/build_arch.yml
@@ -56,21 +56,11 @@ jobs:
         ./b2 stage threading=multi link=static --with-thread --with-system --with-chrono >> /dev/null 2>&1
         sudo ./b2 install threading=multi link=static --with-thread --with-system --with-chrono >> /dev/null 2>&1
 
-    - name : Build open-zwave (x86_64)
-      if: ${{ ( contains(matrix.qemu_arch, 'x86') && github.event_name != 'workflow_dispatch' ) }}
-      run: |
-        cd $GITHUB_WORKSPACE/src
-        cd ..
-        git clone https://github.com/domoticz/open-zwave.git open-zwave-read-only
-        cd open-zwave-read-only
-        make
-        sudo make install >> /dev/null 2>&1
-
     - name : Build domoticz (x86_64)
       if: ${{ ( contains(matrix.qemu_arch, 'x86') && github.event_name != 'workflow_dispatch' ) }}
       run: |
         cd $GITHUB_WORKSPACE/src
-        cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_LIBRARY_PATH=open-zwave-read-only CMakeLists.txt
+        cmake -DCMAKE_BUILD_TYPE=Release CMakeLists.txt
         make
 
     - name: functional-tests domoticz (x86_64)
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thanks for this PR!

Not sure why you need to compile CMake for armv7 as it could be installed as described on our wiki?
You do this also for the other architecture

https://www.domoticz.com/wiki/Build_Domoticz_from_source

The reason why it is currently not automated is that not every commit should trigger a new release.
For example, if one letter is correct because it was wrongly spelled, or some code refactoring that does not change anything for the end user, there is no reason for it

kiddigital:
Maybe we could do the automated builds a little smarter by not triggering it for every commit, but only trigger it when a 'git tag' is added.

So when there is a good reason to release a new Beta, only a Tag has to be committed and new builds will be created with the tag name associated to it.

If that would be an option, I will have a go at it to see if I can create such a workflow.

gizmocuz:
Currently the build tag has a direct relation to the commit. This way we also know when something went wrong.
Fine to trigger a external trigger somehow to get this started, or a manual action on github would be fine as well

Krakinou:
> Not sure why you need to compile CMake for armv7 as it could be installed as described on our wiki?

Well, if I'm not mistaken, your wiki do also compile cmake?
The x86_64 compile cmake as it essentially does the same as  the initial [development.yml](https://github.com/domoticz/domoticz/blob/development/.github/workflows/development.yml) that I used as the starting point of this action. This is true that I could have just downloaded the binaries... I didn't even think about it :)
The aarch64 just install the precompiled version of cmake that is downloadable from the github project
The armv7 version is not available in the cmake github so I also had to compile it .

> The reason why it is currently not automated is that not every commit should trigger a new release.

The action is not triggered at every commit but at every github release (2022.2, 2023.1 and so on). This includes releases that are set as "pre-release" (alpha or beta release for example).
image

If you prefer, I can change the action so it is triggered by a tag? But I'm not sure we can upload asset in relation to a tag, I think only github releases allow this.

gizmocuz:
Yes, you're correct, for armv7 CMake has to be build as there is no binary available.
Performing the action on a new release tag is a good idea!

Krakinou:
I give it a look and it is not possible to link an asset to a tag. Only releases can have assets attached to them.

Anyway, I also learned that, in Github, tags can only be created along a release so there is always a one-one relationship (except in the case one deletes a release without deleting the tag afterward) : So I would say that's not a big deal and in the end, running the action on a Github release or a Github tag is essentially the same.

I added some info on the [initial post of this PR](https://github.com/domoticz/domoticz/pull/5700#issue-1726473685). The pre-release example is currently building but assets should be available soon.

gizmocuz:
Thanks for this PR and your time!

kiddigital:
@Krakinou , @gizmocuz , I gave this a try by creating a pre-release / Beta and the job got started nicely.

Unfortunately it still hit the 6h max run-time limit for the ARM build as the qemu container took already over 4 hours to being build.

Wouldn't it be possible to pre-build these qemu containers and publish them on Dockerhub for example and just pull them during this process?

Any other suggestions?

I do like the easy way to create such a pre-release in Github including the notes it generates..

And 2 of the 3 pre-release builds have succeeded and have been attached as downloadable assets. Nice 👍

Krakinou:
Hi,
Well something is not working as intended : 
The idea behind running the job once manually is to pre-build the system for qemu (install dependencies and compile cmake) so that it is not needed anymore during the actual compilation of domoticz and does not hit the 6h threshold.

I just tried on my forked repo and I have the same issue : it's rebuilding the system even if package was prebuild successfully 2 month ago... I'll give it a look and keep you posted.

kiddigital:
Thx for diving into it.

When I get the change, I will study it as well. I do not understand yet how it will re-use previously build assets (qemu environment) for new builds. But that is just me not understanding (yet) 😁

Krakinou:
Did you give read and write permission to the action so that it could re-use the package?

Krakinou:
Or it could be that the base image ubuntu-latest has changed between the moment you prebuild the image and the moment you tried to build domoticz, so it has to rebuild it.
Maybe better to stick with a stable number like ubuntu22.04?

PR #5708, opened 2023-06-04
Prevent info leakage with getVersion Cmd

### LINKED ISSUE #5713: API/JSON call getversion spills information
The JSON call 'getversion' does not require username/password. 

In my case (and maybe others too) i have domoticz exposed to the public internet via a 'reverse proxy'. I have experienced that the 'getversion' returns some data to anyone that knows the URL. As far as i know there is no way i can configure that this information is not spilled.

I realise that 'getversion' is also used by monitoring applications to check if a domoticz server is 'online'. For what i know, only a reply on the 'getversion' query satisfies the monitoring detection as 'live'.

I would like to suggest two possible solutions:
- change the 'getversion'  (and any other api call) to require authentification. Incase a unauthorized user issues a 'getversion'  query, domoticz responds not with a full message (as present), but only with a JSON response "Unauthorized".  This way a valid JSON transaction is completed, but no information is spilled to unautorized users. In case a valid username/password was passed, the existing response can be returned.
- create a json command 'ping' that always returns a "OK", possibly with a option to disable/enable a anonymous ping. Monitoring applications should migrate from 'getversion' to 'ping' in the future.

kiddigital:
Small PR based on forum discussion to disclose less information via the `getversion` command when a request is not coming from an authenticated user.

From a security perspective it is best to provide only the minimum of information to fulfill the request. Any other, although interesting, information can be _abused_ by bad actors.

See forum info at : https://www.domoticz.com/forum/viewtopic.php?t=40418

With this PR when a request is done without any authorisation, only the following is returned:
```
{
	"status" : "OK",
	"title" : "GetVersion"
}
```

When requested by a (non-admin) User, it gives (_which is the current default response without this PR_):
```
{
	"HaveUpdate" : false,
	"UseUpdate" : false,
	"build_time" : "2023-06-04 16:08:09",
	"dzvents_version" : "3.1.8",
	"hash" : "98a7d1eef-modified",
	"python_version" : "3.8.10 (default, Nov 14 2022, 12:59:47) \n[GCC 9.4.0]",
	"status" : "OK",
	"title" : "GetVersion",
	"version" : "2023.1 (build 15331)"
}
```

And for admins
```
{
	"DomoticzUpdateURL" : "https://www.domoticz.com/download.php?channel=stable&type=release&system=linux&machine=x86_64",
	"HaveUpdate" : false,
	"Revision" : 15091,
	"SystemName" : "linux",
	"UseUpdate" : true,
	"build_time" : "2023-06-04 16:08:09",
	"dzvents_version" : "3.1.8",
	"hash" : "98a7d1eef-modified",
	"python_version" : "3.8.10 (default, Nov 14 2022, 12:59:47) \n[GCC 9.4.0]",
	"status" : "OK",
	"title" : "GetVersion",
	"version" : "2023.1 (build 15331)"
}
```

Also the `getconfig` command now gives reduces information when called without any authorization.

### DIFF
--- a/main/WebServer.cpp
+++ b/main/WebServer.cpp
@@ -2561,12 +2561,11 @@ namespace http
 				root["dzvents_version"] = dzvents->GetVersion();
 				root["python_version"] = szPyVersion;
 				root["UseUpdate"] = false;
-				root["HaveUpdate"] = false;
+				root["HaveUpdate"] = m_mainworker.IsUpdateAvailable(false);
 
 				if (session.rights == URIGHTS_ADMIN)
 				{
 					root["UseUpdate"] = g_bUseUpdater;
-					root["HaveUpdate"] = m_mainworker.IsUpdateAvailable(false);
 					root["DomoticzUpdateURL"] = m_mainworker.m_szDomoticzUpdateURL;
 					root["SystemName"] = m_mainworker.m_szSystemName;
 					root["Revision"] = m_mainworker.m_iRevision;
@@ -2796,14 +2795,6 @@ namespace http
 			root["status"] = "ERR";
 			root["title"] = "GetConfig";
 
-			int iUser = -1;
-			unsigned long UserID = -1;
-			if (!session.username.empty() && (iUser = FindUser(session.username.c_str())) != -1)
-			{
-				UserID = m_users[iUser].ID;
-				root["UserName"] = m_users[iUser].Username;
-			}
-
 			std::string sValue;
 			int nValue = 0;
 			int iDashboardType = 0;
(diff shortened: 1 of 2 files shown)


gizmocuz:
You might also need to modify $rootScope.GetGlobalConfig

But I don't see any harm in the information returned... but

And alerting a non-admin user that there is a update is maybe not to bad as well.. they can signal an admin to update

kiddigital:
> You might also need to modify $rootScope.GetGlobalConfig

Check.. done

> But I don't see any harm in the information returned... but

You never know what bad actors can do with 'harmless' information :)

> 
> And alerting a non-admin user that there is a update is maybe not to bad as well.. they can signal an admin to update

Done..

gizmocuz:
Thanks for the PR!

PR #5712, opened 2023-06-06
Adding Time-based One-Time Passcode (RFC6238) support for 2FA to Login and IAM service

gizmocuz:
This PR adds a way to support 2-Factor Authentication.

By default, Domoticz activates TOTP support for each User that has enabled 2FA (via their Profile screen).
When enabled, a second authentication screen is presented that requires a 6-digit passcode next to the default Username and Password screen.

It works with several Authenticators like (but not limited to): [Authy](https://authy.com/), [FreeOTP](https://freeotp.github.io/), [Google Authenticator](https://googleauthenticator.net/) or [Microsofts Authenticator](https://www.microsoft.com/en-us/security/mobile-authenticator-app).

**New 'My Profile' menu-option**
There is a new 'My Profile' menu option under 'Setup' (for Admin) or 'Profile' (for non-admins) where you can enable Two-Factor Authentication for the logged-in User.

Users can also change their password here :) No need to do that via an Admin and User management anymore


gizmocuz:
Many many thanks to **kiddigital** for all the hard work in this PR!!!

PR #5714, opened 2023-06-08
Fix the functionaltests

kiddigital:
Small fix, updated documentation and added `runtests.sh` support script (in the 'test'-directory).

PR #5708 changed the default anonymous output of the 'getversion' command which was used to check to see if domoticz itself is running and reachable before executing the functional tests. The check was looking for the versionnumber which is no longer _leaked_ under anonymous access.


gizmocuz:
Thank you!!

PR #5715, opened 2023-06-08
Improved API compatibility and logging

kiddigital:
Small PR that (temporarily) improves the backward compatibility with old deprecated RType API calls that have been changes in PR #5697.

Also the STATUS logging is better showing the origin IP-address and if or if not the call is handled via fallback.


gizmocuz:
Thanks!!!

PR #5735, opened 2023-07-12
Replace SetCurrentCostUSBType action with new proper API command

kiddigital:
Instead of using the old 'webem' POST request form handling, now use a proper API command to set the correct values for this hardware type.

- Cmd added to hardware class
- Registered the new command to webserver API commands
- Updated the UI to use the new API command instead of the form POST

And fixed UI issue (not showing the current value) in the process.


gizmocuz (commented):
Looking good! Thanks for the PR
Looks like you found a bug in the previous code as well (item.Address)😎

PR #5740, opened 2023-07-15
Added support for Libre Hardware Monitor

Eegee:
On Windows, added support for Libre Hardware Monitor, with a fallback to Open Hardware Monitor, as the latter hasn't been maintained recently.

### DIFF
--- a/hardware/HardwareMonitor.cpp
+++ b/hardware/HardwareMonitor.cpp
@@ -10,7 +10,8 @@
 #define __STDC_FORMAT_MACROS
 #include <inttypes.h>
 
-//Note, for Windows we use OpenHardware Monitor
+//Note, for Windows we use Libre Hardware Monitor or Open Hardware Monitor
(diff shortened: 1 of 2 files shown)


kiddigital:
Thx, seems useful.

As I don't run Windows, did you test with both Open and Libre?

Which versions (and which Windows version).

gizmocuz:
@Eegee , I had a look at your PR and looks perfect! Still support for both monitors so current users won't have issues, but main set to Libre.
Great job, thanks for the PR!

Eegee:
I am very sorry, I made a mistake. The fallback does not seem to work properly: open hardware monitor is still not detected even if it is running. I will investigate further and will try to come up with a proper solution. @gizmocuz

gizmocuz:
No problem. I will take it out the release version for now. But might be better to switch to Libre 100% if you can't get it working

PR #5741, opened 2023-07-16
2023.2 stable

gizmocuz:


Eegee:
I am sorry, I made a mistake. The fallback in my PR https://github.com/domoticz/domoticz/pull/5740 does not seem to work properly. I will investigate further and will try to come up with a proper solution. @gizmocuz

PR #5742, opened 2023-07-16
Revert "Added support for Libre Hardware Monitor"

Eegee:
Reverts domoticz/domoticz#5740
Fallback in IsHMRunning doesn't work properly: check for (FAILED(hr)) doesn't work as as expected.


gizmocuz:
Not accepted ;) I will have a look at it as well ... not today, maybe tomorrow or later
See who comes up with a fix first 😂😎

PR #5743, opened 2023-07-16
Fallback to Open Hardware Monitor corrected

Eegee:
Windows executables for hardware sensors are now properly checked in order, first Libre Hardware Monitor, then Open Hardware Monitor. If an executable is found running, its corresponding WMI namespace is used in subsequent sensor data lookups.

Summary of the errors in my previous PR: I assumed HRESULTs would result in FAILED when they didn't and WMI namespaces can still exist even when the related executable is no longer running.

Tested on Windows 11 Pro 22H2 build 22621.1992 with Libre Hardware Monitor 0.9.2 and Open Hardware Monitor 0.9.6

### DIFF
--- a/hardware/HardwareMonitor.cpp
+++ b/hardware/HardwareMonitor.cpp
@@ -16,6 +16,14 @@
 
 #ifdef WIN32
 	#include <comdef.h>
+
+constexpr std::array<std::pair<const char*, const char*>, 2> hardwareMonitors {
+	{
+		{ "LibreHardwareMonitor.exe", "root\\LibreHardwareMonitor" },
+		{ "OpenHardwareMonitor.exe", "root\\OpenHardwareMonitor" },
+	}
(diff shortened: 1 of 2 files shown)


Eegee:
As was mentioned, it's also fine by me if you'd rather have Libre Hardware Monitor permanently be used instead of Open Hardware Monitor without any fallback. This would make the change in the code as from yesterday simpler but also would mean notifying in release notes any existing OHM user to switch to LHM and updating docs etcetera.

gizmocuz:
Sorry, did not see your PR after I committed a patch for this 😎

Unfortunately, 'ConnectServer' always returns true if one of the two applications have been started once before.
The only true check is to also call 'IsHMRunning'
There was also a small memory leak in this function, so I took the opportunity to patch that as well.

Should be running fine in beta 15435

PR #5750, opened 2023-07-22
Add cron job for Github action build last version

Krakinou:
Following #5700 , I added a cron job that will build /rebuild the qemu environnement every day at 23.00 UTC.
So no worries about running the job manually or whatever, it should work on its own and there should always be an up-to-date package for when you're releasing a version.

I pushed directly to master as it does not change the "real" code and it need to be in master for the cron job to be triggered.


gizmocuz:
I have absolute no problem building the stable releases by hand as I want to monitor this very closely.
Next to this, a patch/commit in the master branch should not always create a new stable release (like the fix PR from today)
Also, when making a new stable release there currently also needs to be build a new Windows installer, and I always make a post on the forum about the new stable version.

gizmocuz:
And after I'm sure the arm builds (x86/x64) and the Linux (x64) build correctly and are published, only then it is possible to build also the new docker.
So it's not a matter of letting everything quickly build.

Krakinou:
Hello,
This does not change anything to the current github action except that it will try to prepare the qemu environnement everyday to prevent the 6hours treshold issue stated [here](https://github.com/domoticz/domoticz/pull/5700#issuecomment-1635629539).
The docker package will be prebuild on a daily basis instead of manually, **no actual build from domoticz will be made.**

> Next to this, a patch/commit in the master branch should not always create a new stable release (like the fix PR from today)


I'm not sure of what you're saying? Did my PR trigger something? In this case sorry I was not aware.


> When making a new stable release there currently also needs to be build a new Windows installer


Yes, I never find a way to do it on github event...

gizmocuz:
Yes, but also building it every day (... we have a stable release every xx months?) taking up hours of the github servers, is not so convenient.
I would wait until they support arm building Kinda strange as they have a Arm windows build for tablets....

Krakinou:
The job will just check if anything changed. 
If nothing changed it's over in 2/3 minutes.
If something changed, a new qemu environnement is prepared.

gizmocuz:
As you can read above, a lot has to be done when creating a new stable release.

This is not something that can be done via a GitHub action, this can be done via GitHub helpers and we are looking into this option at the moment.
As we host multiple build servers to also overcome the use of qemu they need to be setup to work with this.
There is no need for a GitHub action to build a stable/master build.

gizmocuz:
(The only reason why it is here on the development branch is mainly to check for build errors)

Krakinou:
Well, as the github action we're talking about was [merged into the master branch](https://github.com/domoticz/domoticz/blob/25624ca148fa220a2ad63a4ff2faf614253c6111/.github/workflows/build_arch.yml) before 2023.2 release, I had mistakenly the impression that you and @kiddigital wanted this to be fixed in [this message](https://github.com/domoticz/domoticz/pull/5700#issuecomment-1635629539) :

> Wouldn't it be possible to pre-build these qemu containers and publish them on Dockerhub for example and just pull them during this process?


Anyway, I just provided a fix for this as requested, that's your repo and you know better than me what to do with it, feel free to close the PR or whatever.

gizmocuz:
Thanks for the fix and your time but releases (beta/stable/docker) can not be build via GitHub actions
The Travis and Appveyor compile tests are very useful for a quick check if everything is working.

We do need all 3 builds ready before a docker can be build containing them all

PR #5754, opened 2023-07-22
Phase out openzwave

gizmocuz:


.github/workflows/development.yml
kiddigital, line 1: The 'Config' directory (from OpenZWave) needs to be removed from the list of files/directories that should be tarred-up into the 'Package' (Packaging step)

gizmocuz: Done....

PR #5762, opened 2023-07-26
(IsValidIP) Remove IPv6 zone-index for link-local addresses

### LINKED ISSUE #5760: (fe80::6:5605:b4d9:dfe9%eth0) is not a valid Ipv4 or IPv6 address!
> 2023-07-24 18:55:54.056 Status: [web:8080] Given host (fe80::6:5605:b4d9:dfe9%eth0) is not a valid Ipv4 or IPv6 address! Unable to check if in Trusted Network!

As described in the forum
https://www.domoticz.com/forum/viewtopic.php?t=40576

kiddigital:
Small PR addressing #5760 

Domoticz now is able to handle link-local IPv6 addresses that contain a zone-index identifying the used machine interface.

Actually we just remote the zone-index information as it is not used during checking of valid IP addresses.

### DIFF
--- a/webserver/cWebem.cpp
+++ b/webserver/cWebem.cpp
@@ -997,6 +997,12 @@ namespace http {
 				{
 					cleanIP = cleanIP.substr(1,cleanIP.size()-2);	// Remove brackets from begin and end
 				}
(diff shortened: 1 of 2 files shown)


gizmocuz:
Thank you!!

PR #5770, opened 2023-08-02
Update AlfenEve.cpp, remove URLEncode of password

### LINKED ISSUE #5768: URLEncode of password should not be done, causes login failure.
On line 65 on AlfenEve.cpp in the hardware folder it shows:
m_szPassword(CURLEncode::URLEncode(szPassword))

If the szPassword (as provided in the Domoticz hardware setup page) contains any special characters, then the login fails as result of this URLencode.

If the owner password of the charger is changed so it does not contain any special characters and therefore the URLEncode has no effect, then the login to the charger is succesfull, devices are created, data is retrieved.

So the URLEncode should be removed.

Workaround for now is to use a owner password for the charger without special characters.

WillemD61:
Remove URLEncode of the password on line 65 since URLEncode is not needed and causes login failure if the password contains special characters.
Reference issue 5768

### DIFF
--- a/hardware/AlfenEve.cpp
+++ b/hardware/AlfenEve.cpp
@@ -62,7 +62,7 @@ std::string ReadFile(std::string filename)
 AlfenEve::AlfenEve(const int ID, const std::string& IPAddress, const uint16_t usIPPort, int PollInterval, const std::string& szUsername, const std::string& szPassword) :
 	m_szIPAddress(IPAddress),
 	m_szUsername(szUsername),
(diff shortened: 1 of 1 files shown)


WillemD61:
Sorry, don't know what I did wrong, just wanted to remvoe the URLEncode. Someone with more knowlegde of C help please?

gizmocuz:
Thank you very much for this PR! I make the change locally. 
You also needed to remove the CURLEncode:: part

gizmocuz:
Solved in beta 15477

PR #5796, opened 2023-08-23
Update MQTTAutoDiscover.cpp

heggink:
Add ignoring condition where a lock receives a state that is mapped in the value template


heggink:
typo fixed

PR #5801, opened 2023-08-25
(Refactor) Initialize Thermosmart with only relevant (checked) parameters

kiddigital:
The Thermosmart hardware was initialized with all (6) possible _Mode_ parameters while only 1 is used (and checked).

**To-do?** Make sure during Hardware validation that the device Index passed in the _Mode1_ parameters `IsTemp(...)` ?

**To-do?** Same problem (copy/paste) exist in the _AtagOne_ and the _OpenThermGateway_ (OTGW) hardware classes. Only Mode1 is used but all 6 are passed.

### DIFF
--- a/hardware/Thermosmart.cpp
+++ b/hardware/Thermosmart.cpp
@@ -56,29 +56,14 @@ std::string ReadFile(std::string filename)
 }
 #endif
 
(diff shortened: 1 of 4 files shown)


gizmocuz:
Thank you!!

waltervl:
Looking at the tweakers reviews the device is not working anymore... So perhaps time to deprecate in some release?https://tweakers.net/pricewatch/390714/thermosmart/reviews/?orderField=time&orderSort=desc#productReviews

kiddigital:
Hm... still works fine here

PR #5817, opened 2023-09-12
Setpoint Parameters

gizmocuz:
Converted 'Thermostat Setpoint' to generic 'Setpoint' type

Added the following options:

- user defined Unit (Like RPM/Watt/C)
- step size
- minimum value
- maximum value


waltervl:
Thanks!

gkirstei:
I have been using the homebridge-edomoticz plugin to integrate my Domoticz setup with Apple's HomeKit. However, following the aforementioned changes, the setpoint thermostats in HomeKit now show up as "not supported." It appears that the modifications to the Thermostat Setpoint Device have affected the compatibility with HomeKit via the homebridge-edomoticz plugin. This was a valuable feature for me and its loss affects the usability of the system. Addressing this issue would help in maintaining the seamless integration with HomeKit.

waltervl:
Notifying the author of the plugin would be the correct way to go.

PR #5818, opened 2023-09-12
Fixes for issues 5802 and 5808

### LINKED ISSUE #5802: Managed counter - dzVents does not return actual value
in a dzVents script, "domoticz.devices( [deviceid] ).counter" always returns the value of the counter at startup of domoticz. 
[deviceid] is the id of a managed counter.

"http://localhost:8080/json.htm?type=command&param=getdevices&rid=[deviceid]"  does return the correct actual value

probably missing:
_m_mainworker.m_eventsystem.ProcessDevice(HardwareID, ulID, unit, devType, subType, signallevel, batterylevel, nValue, sValue);_
at the end of function "CSQLHelper::UpdateManagedValueInt"  (SQLHelper.cpp line 5053)

### LINKED ISSUE #5808: Managed Counters - Comparing Graph doesn't display data
[2023.2 (build 15509)
the "Comparing Graph" of a manage counter doesn't display data.
managed counters use "value" and not the "counter" column of the meter_calendar table.

the CWebServer::GroupBy function (Webserver.cpp line 3528), called from WebServerHandleGraph line 2966) returns no data if there is no value in the "counter" column of meter_calender table.

Stripping it down to it's basics, using bIsManagedCounter as a parameter returns the correct values ( i'm aware that if there is no "value", result will be incomplete)
Attached the top-part of a stripped-down version of GroupBy, using bIsManagedCounter as a parameter.

[GroupBy-first_part.txt](https://github.com/domoticz/domoticz/files/12500736/GroupBy-first_part.txt)

robk358:
5802: there was no update to the internal (cached) datatable, fixed with adding "m_mainworker.m_eventsystem.ProcessDevice" to function CSQLHelper::UpdateManagedValueInt. 
5808: modified the query of CWebServer::GroupBy, created a new parameter "bUseValuesOrCounter", if true the is a check if the Counter column for a device contains values, if there are no values, data of the Value column will be used.

### DIFF
--- a/main/SQLHelper.cpp
+++ b/main/SQLHelper.cpp
@@ -5162,6 +5162,9 @@ uint64_t CSQLHelper::UpdateManagedValueInt(
 	}
 
 	safe_query("UPDATE DeviceStatus SET LastUpdate='%q', sValue='%q' WHERE (ID = %" PRIu64 ")", sLastUpdate.c_str(), sValue, ulID);
+	
+	m_mainworker.m_eventsystem.ProcessDevice(HardwareID, ulID, unit, devType, subType, signallevel, batterylevel, nValue, sValue);
+
 	return ulID;
 }
 
--- a/main/WebServer.cpp
+++ b/main/WebServer.cpp
@@ -3525,7 +3525,7 @@ namespace http
 		 * Takes root["result"] and groups all items according to sgroupby, summing all values for each category, then creating new items in root["result"]
 		 * for each combination year/category.
 		 */
-		void CWebServer::GroupBy(Json::Value& root, std::string dbasetable, uint64_t idx, std::string sgroupby, bool bUseValues, std::function<std::string(std::string)> counter,
+		void CWebServer::GroupBy(Json::Value& root, std::string dbasetable, uint64_t idx, std::string sgroupby, bool bUseValuesOrCounter, std::function<std::string(std::string)> counter,
 			std::function<std::string(std::string)> value, std::function<std::string(double)> sumToResult)
 		{
 			/*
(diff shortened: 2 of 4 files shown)


gizmocuz:
Thanks for the PR!

regarding
```
			m_mainworker.m_eventsystem.ProcessDevice(ParentHardwareID, ParentID, ParentUnit, ParentType, ParentSubType, signallevel, batterylevel, nValue, sValue);
			m_mainworker.sOnDeviceUpdate(std::stoi(sd[2]), std::stoll(sd[0]));

```
I think this should be done at the top of CSQLHelper::UpdateValue when it's not a light/switch ?

Not sure why you need a patch for 5808?

robk358:
5808: i am using a number of managed counters, but without the Counter values filled in, only the Value column. (calculated values, not based on a counter) I really like the "Comparing Graph". 

the graph before: 
without_5808
and now with patch:
with_5808

5802: until now i've only found the managed counter to have the issue. 
at CSQLHelper::UpdateValue, ProcessDevice is also used for light/switch devices
at CSQLHelper::UpdateValueInt ProcessDevice is only used for non-managed counters
at CSQLHelper::UpdateValueInt, UpdateManagedValueInt is called and returns without calling ProcessDevice

i _think_ UpdateManagedValueInt is the best place, unless if there is a check for managedcounters (+all other checks) at the the top of UpdateValue (isn't that what UpdateManagedValueInt does?)
(of course i don't know the code in depth as you do..)

gizmocuz:
Sorry for the delay and thank you for the PR and time!

robk358:
no problem and you're welcome.

PR #5825, opened 2023-09-27
Update KMTronicSerial.cpp

reavell:
Update to allow for more than six 8-relay boards on a KMTronic RS485 connection. The KMT protocol allows for up to 15 boards, so this patch also allows for up to 15 boards, to a total of 120 individual relays on one RS485 connection.


gizmocuz:
That's a lot of relays! Thank you for the PR!!

PR #5835, opened 2023-10-08
[USBtin/Multibloc V8] - Add support of Bloc 7 + some fixes

reivax85:
Fixed a bad sID used to write to supply voltage information on bloc 9/ bloc sfsp
Add support of Bloc 7 which allows users to be able to read analog and digital inputs coming from this block (and IBS information to, because bloc 7 has a LIN input like block 9 so we can connect 6x IBS on it !)
Stack upgraded to V04.00


reivax85:
@kiddigital  I think I'm ok now :)

gizmocuz:
Thank you for this PR and all work!!!

PR #5839, opened 2023-10-12
Bug fix: MQTT RGB dimmers do not work with scenes, and timers.

ajarzyn:
- change `allowed_components` to set to optimize verification
- add `SwitchCommands` `enum` not to compare `commands` strings
- return after send when changing only level
- on `color_set` set level first

### DIFF
--- a/hardware/MQTTAutoDiscover.cpp
+++ b/hardware/MQTTAutoDiscover.cpp
@@ -10,8 +10,9 @@
 #include "../main/WebServer.h"
 #include "../notifications/NotificationHelper.h"
 #include <iostream>
+#include <set>
 
-std::vector<std::string> allowed_components = {
+std::set<std::string> allowed_components = {
 		"binary_sensor",
 		"button",
 		"climate",
@@ -24,6 +25,16 @@ std::vector<std::string> allowed_components = {
 		"sensor",
 		"switch",
 		"fan"
+};
+
+enum SwitchCommands {
+	COMMAND_UNKNOWN = -1,
+	COMMAND_ON,
+	COMMAND_OFF,
+	COMMAND_OPEN,
+	COMMAND_STOP,
+	COMMAND_SET_LVL,
+	COMMAND_SET_COL,
 };
 
 #define CLIMATE_MODE_UNIT 1
@@ -551,7 +562,7 @@ void MQTTAutoDiscover::on_auto_discovery_message(const struct mosquitto_message*
 
 	component = strarray[0];
 
-	if (std::find(allowed_components.begin(), allowed_components.end(), component) == allowed_components.end())
+	if (allowed_components.find(component) == allowed_components.end())
 	{
 		//not for us
 		return;
@@ -3675,6 +3686,7 @@ bool MQTTAutoDiscover::SendSwitchCommand(const std::string& DeviceID, const std:
 
 	std::string szSendValue;
 	std::string command_topic = pSensor->command_topic;
+	SwitchCommands eCommand = SwitchCommands::COMMAND_UNKNOWN;
 
 	if (
 		(pSensor->component_type != "climate")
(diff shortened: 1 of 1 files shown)


ajarzyn:
Tested on Fibargroup 	RGBW CONTROLLER 2

gizmocuz:
Thank you for the PR!

Did you also tested with the latest Domoticz beta versions? As there have been some changes the last few days.

It is a bit hard to see what did not work, and what you changed to solve the error

I do like the idea for the enum instead of a string compare.

could you write the fill name, like LEVEL and COLOR
COMMAND_SET_LEVEL,
COMMAND_SET_COLOR,

But first please check if it is working as it is now with the latest beta version

ajarzyn:
Names changed, thank you for suggestion.

Regarding testing:
Initially I've noticed the problem on the stable release: `Version: 2023.2 (build 15457)`.
Then I reproduced the problem on `development` branch, compiling domoticz from source.

Then I developed the fix and tested in under the provided commit.

What is the problem:
In the file `domoticz\main\mainworker.cpp:11736` there's an if:
```c++
	if (pHardware->HwdType == HTYPE_MQTTAutoDiscovery)
	{
		// Special case when color is passed from timer or scene
		if ((switchcmd == "Set Level") && (color.mode != ColorModeNone))
		{
			switchcmd = "Set Color";
		}
		return ((MQTTAutoDiscover*)m_hardwaredevices[hindex])->SendSwitchCommand(sd[1], sd[9], Unit, switchcmd, level, color, User);
	}
```

I suppose it was introduced as a work around with this commit: https://github.com/domoticz/domoticz/commit/513b5774bc74226437c430e13f2cb3349856d446
But I created the scenario that from `scenes` and `timers` the `Set Level` command was never send.

Provided implementation change behaviour of `Set Color` in following:
```
if command == "Set Color":
    sendMQTTCommand(Set Level)
    sendMQTTCommand(Set Color)
if command == "Set Level":
    sendMQTTCommand(Set Level)
```

Thus instead of 1 message (for colour set) there are 2 messages sent.

Version of the `development` build is: `Version: 2023.2 (build 15599)`

gizmocuz:
Can you post your MQTT auto discovery config topic + json payload?
You don't want to send two commands, most (99,9%) of the cases it's a json structure thats being send, and this structure could contain the color and brightness
But we made some changes for this recently, that's why I asked if you experience this behavior also when using the latest beta
(15599 is good to test this)

To see if we can create one json object to send containing both color and brightness, I would like to see the config topic/payload

ajarzyn:
Some time ago I've followed you advice and migrated all ZWave devices to [zwave-js-ui](https://github.com/zwave-js/zwave-js-ui). Today all zwave MQTT devices go through zwave-js-ui.

On `COMMAND_SET_COLOR`
Topic: `"zwave/Aria_LED/51/1/targetColor/set"`
Payload: `"{\"state\":\"ON\",\"value\":{\"blue\":0,\"green\":18,\"red\":255,\"warmWhite\":0}}"`

I've checked templated for the colour settings:
[wave-js-ui:devices.ts#L286](https://github.com/zwave-js/zwave-js-ui/blob/0b8f33f94bcf230489f6c0e8288bae7554011669/hass/devices.ts#L286)
But there is simply no information about brightness.

I believe this is what you are interested with (retrieved from the node config):
```json
"light_rgb_dimmer_1": {
      "type": "light",
      "object_id": "rgb_dimmer_1",
      "discovery_payload": {
        "state_topic": "zwave/Aria_LED/38/1/currentValue",
        "command_topic": "zwave/Aria_LED/38/1/targetValue/set",
        "brightness_scale": 99,
        "rgb_state_topic": "zwave/Aria_LED/51/1/currentColor",
        "rgb_command_topic": "zwave/Aria_LED/51/1/targetColor/set",
        "brightness_state_topic": "zwave/Aria_LED/38/1/currentValue",
        "brightness_command_topic": "zwave/Aria_LED/38/1/targetValue/set",
        "device": {
          "identifiers": [
            "zwavejs2mqtt_0xc8b07bc8_node10"
          ],
          "manufacturer": "Fibargroup",
          "model": "RGBW CONTROLLER 2 (FGRGBW-442)",
          "name": "Aria LED",
          "sw_version": "5.0"
        },
        "name": "Aria LED_rgb_dimmer_1",
        "unique_id": "zwavejs2mqtt_0xc8b07bc8_10-51-1-currentColor"
      },
      "values": [
        "38-1-currentValue",
        "38-1-targetValue",
        "51-1-currentColor",
        "51-1-targetColor"
      ],
      "discoveryTopic": "light/Aria_LED/rgb_dimmer_1/config",
      "persistent": false,
      "ignoreDiscovery": false
    },
```

I was digging around and couldn't find the command to set both colour and level at the same time.

gizmocuz:
You still did not answer if it is working with the latest beta version without your modification.

I am afraid this is not what I was looking for.

Please use MQTT Explorer and navigate to your MQTT broker
In here you should be able to find your light config topic, it probably starts with

homeassistant/light/xx/config

Please copy/paste the contents of the config (JSON object) here

hardware/MQTTAutoDiscover.cpp
```diff
@@ -3909,7 +3949,7 @@ bool MQTTAutoDiscover::SendSwitchCommand(const std::string& DeviceID, const std:
 			if (!pSensor->rgb_command_topic.empty())
 				command_topic = pSensor->rgb_command_topic;
 
-			if ((bCouldUseBrightness) && (pSensor->bBrightness))
+			if (bCouldUseBrightness && pSensor->bBrightness)
```
ajarzyn, line 3952: @gizmocuz check that bCouldUseBrightness is already in the code.

gizmocuz:
Thank you for this PR and your work!

Issue #5841, opened 2023-10-17
P1 Meter - Voltage swell and sag devices are not created unless there are swells or lags


robk358:
[2023.2] (build 15603)
"P1 Smart meter" swell and sag devices are created only if there are swells or lags (and device discovery is on).
A zero value of a swell or sag is also a valid value..

Tests for sag/swell values in P1MeterBase::MatchLine() should compare with "-1" instead of "> 0"

gizmocuz:
Thanks for your PR!

It was indeed my idea to not add them if there was no data... But happy to accept this patch if you think it's better

robk358:
i think it's better to add them: if you want to trigger an alert, the device(s) needs to be present. (and if you don't want to use the devices, you can disable them)

gizmocuz:
Okey will do! Thanks!

PR #5842, opened 2023-10-17
Fix for #5841

robk358:
fix: P1 Meter - Voltage swell and sag devices are not created unless there are swells or lags


gizmocuz:
Thank you for the PR!

Issue #5630, opened 2023-03-08
Error messages in new plugin system when calling Update


Krzysztof1234534:
I've added question on forum:
https://www.domoticz.com/forum/viewtopic.php?t=40037
But without response. I suspect that there is a bug, so I'm creating issue here. 
From forum description:

Hi, I'm creating new python plugin, I've decided to use new DomoticzEx interface. Sometimes it seems that in logs there is such message:
Error: WMBus: Update to 'UnitEx' failed to update any DeviceStatus records for key 32/RFM69Temperature/1
I'm using 2022.2 version.
I've decided to write multithread plugin, I've seen that no it is possible and there was examples for that. I need async, because I'm using spidev and RPi.GPIO.
But I'm not sure if it is save to call Unit.Update() function from another thread (not onHeartbeat), documentations says nothing about this. I thought that this is possible, because multiple plugins has now running on different threads, so probably Domoticz is not single thread.

I've search the code and it seams that this message is logged in PythonObjectEx when m_sql.execute_sql returns 0 rows. But I think that bug is in SQLHelper, because there are 2 statements:
sqlStatement.Execute();
result = m_sql.safe_query("SELECT changes();");
There are no lock and no transaction, so between this first and second command some another thread can execute own query.

gizmocuz:
Are you calling/using database queries in your plugin or doing everything via the python script?
(Not that I can help you with the python part, but directly using the database is not a good idea!)

Krzysztof1234534:
Of course not, I'm calling DomoticzEx api, more less something like this:
```
deviceUnit.sValue = total_m3
deviceUnit.Update()
```
I've search code what doing this update method, and this is probably in PythonObjectEx.cpp (CUnitEx_update function).
There is:
```
if (!iRows)
{
    pModState->pPlugin->Log(LOG_ERROR, "Update to 'UnitEx' failed to update any DeviceStatus records for key %d/%s/%d", pModState->pPlugin->m_HwdID, sDeviceID.c_str(), self->Unit);
    Py_RETURN_NONE;
}
```

This is the error I can see in logs. 
Do you need to repeat this or this explanation is enough? My plugin using hardware, so it will not work for you, but I can try to create simplified version of plugin. I have a lot of devices (max idx is 258), probably this will not occur in clean database, because it looks like race condition (I only observing this error 1-2 times per day).

Krzysztof1234534:
[...]

PR #5849, opened 2023-10-23
Sqlhelper locks and cWebem isnew connection

Krzysztof1234534:
Added more locks to sqlhelper and fixed problem with multiple incomming connection in cWebem.

Changes to fix issue: https://github.com/domoticz/domoticz/issues/5630 are in SQLHelper. I used recursive_mutex for simplicity to not care about multiple locks from the same thread, I've didn't spot any performance issue. It is possible to use mutex, but it will require much more code refactoring, so my proposition as simple fix is to use recursive_mutex. 

I've also fixed problem with multiple logs "Status: [web:80] Incoming connection from: ...". This was in specific case: when AreWeInTrustedNetwork returns true and browsers cookie is expired (for example first opening the browser in the day). In some cases browsers sends a lot of requests to domoticz in parallel (browser has cached html page and all resources, but if this is first using of domoticz that day it assumes that resources needs to be refreshed, so this is why there is a lot of requests). Because we are in trusted network, the code in CheckAuthentication (if (AreWeInTrustedNetwork... sets session.rights = 2). At the end of the CheckAuthentication there is:
if ((session.rights == URIGHTS_ADMIN) && (session.id.empty()))
This is the place where session.isnew = true; 
Code goes to this place, because there are no cookie, so parse_cookie returns false. And there is a lot of requests without cookie (something around 60 - this is the number of icons, pages and so on). In handle_request after executing CheckAuthentication  there is:
if ((session.isnew == true) &&
				(session.rights == URIGHTS_ADMIN) &&
				(req.uri.find("json.htm") != std::string::npos) &&
				(req.uri.find("logincheck") == std::string::npos)
				)
I've removed this json.htm check, because we also want to go there in all cases in Trusted Network (except logincheck, but if will probably not be used in trusted network). Note that code "session.isnew = true" is in 3 places:
- logincheck (this is excluded)
- in CheckAuthentication at the end (this is Trusted Network)
- in CheckAuthentication when parse_cookie returns true and session.rights == 2 - if rights == 2 it means that it is trusted network, but if there is cookie (parse_cookie return true), it means that it is not expired yet. This case looks like some fallback, it is not used frequently. I've removed private method authorize, because it was not used and there was "session.isnew = true".

Note also that I'm not sure if handle_request is synchronous or async method (probably it is async). Because there is incorrect lock mechanism in cWebem. "std::unique_lock<std::mutex> lock(m_sessionsMutex);" is in GetSession, AddSession, RemoveSession, CountSessions, GetExpiredSessions, but for example look at code:

```
WebEmSession* memSession = myWebem->GetSession(session.id);
if (memSession != nullptr)
{
  time_t now = mytime(nullptr);
  // Renew session expiration date if half of session duration has been exceeded ("dont remember me" sessions, 10 minutes)
  if (memSession->expires - (SHORT_SESSION_TIMEOUT / 2) < now)
  {
    memSession->expires = now + SHORT_SESSION_TIMEOUT;
    memSession->auth_token = generateAuthToken(*memSession, req); // do it after expires to save it also
    send_cookie(rep, *memSession);
  }
  // Renew session expiration date if half of session duration has been exceeded ("remember me" sessions, 30 days)
  else if ((memSession->expires > SHORT_SESSION_TIMEOUT + now) && (memSession->expires - (LONG_SESSION_TIMEOUT / 2) < now))
  {
    memSession->expires = now + LONG_SESSION_TIMEOUT;
    memSession->auth_token = generateAuthToken(*memSession, req); // do it after expires to save it also
    send_cookie(rep, *memSession);
  }
}
```
memSession is edited outside of lock guard (it means that this modification is not protected by lock). Also there is a lot of GetSession and after that RemoveSession - this also should be atomic operation. But I've didn't observe any problems with incorrect locks behaviour, so my fix for now should be enough.

### DIFF
--- a/main/SQLHelper.cpp
+++ b/main/SQLHelper.cpp
@@ -3599,7 +3599,7 @@ bool CSQLHelper::OpenDatabase()
 
 void CSQLHelper::CloseDatabase()
 {
-	std::lock_guard<std::mutex> l(m_sqlQueryMutex);
+	std::lock_guard<std::recursive_mutex> l(m_sqlQueryMutex);
 	if (m_dbase != nullptr)
 	{
 		OptimizeDatabase(m_dbase);
@@ -4307,7 +4307,7 @@ bool CSQLHelper::safe_UpdateBlobInTableWithID(const std::string& Table, const st
 	if (!m_dbase)
 		return false;
 
-	std::lock_guard<std::mutex> l(m_sqlQueryMutex);
+	std::lock_guard<std::recursive_mutex> l(m_sqlQueryMutex);
 
 	sqlite3_stmt *stmt = nullptr;
 	char* zQuery = sqlite3_mprintf("UPDATE %q SET %q = ? WHERE ID=%q", Table.c_str(), Column.c_str(), sID.c_str());
@@ -4336,6 +4336,7 @@ bool CSQLHelper::safe_UpdateBlobInTableWithID(const std::string& Table, const st
 
 int CSQLHelper::execute_sql(const std::string &sSQL, std::vector<std::string> *pValues, bool bLogError)
 {
+	std::lock_guard<std::recursive_mutex> l(m_sqlQueryMutex);
 	CSQLStatement sqlStatement(m_dbase, sSQL);
 	std::vector<std::vector<std::string>> result;
 	for (unsigned int i = 0; (i < pValues->size()) && (!sqlStatement.Error()); i++)
@@ -4404,7 +4405,7 @@ std::vector<std::vector<std::string> > CSQLHelper::query(const std::string& szQu
 		std::vector<std::vector<std::string> > results;
 		return results;
 	}
-	std::lock_guard<std::mutex> l(m_sqlQueryMutex);
+	std::lock_guard<std::recursive_mutex> l(m_sqlQueryMutex);
 
 	sqlite3_stmt* statement;
 	std::vector<std::vector<std::string> > results;
@@ -4471,7 +4472,7 @@ std::vector<std::vector<std::string> > CSQLHelper::queryBlob(const std::string& 
 		std::vector<std::vector<std::string> > results;
 		return results;
 	}
-	std::lock_guard<std::mutex> l(m_sqlQueryMutex);
+	std::lock_guard<std::recursive_mutex> l(m_sqlQueryMutex);
 
 	sqlite3_stmt* statement;
 	std::vector<std::vector<std::string> > results;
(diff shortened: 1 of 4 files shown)


gizmocuz:
Thank you for this PR!

I am not sure why you need to use **recursive_mutex** if there is a double lock somewhere, that should be fixed

Can you tell me make a callstack that goes wrong?

Krzysztof1234534:
Look at execute_sql method:
```
int CSQLHelper::execute_sql(const std::string &sSQL, std::vector<std::string> *pValues, bool bLogError)
{
	std::lock_guard<std::recursive_mutex> l(m_sqlQueryMutex);
	CSQLStatement sqlStatement(m_dbase, sSQL);
	std::vector<std::vector<std::string>> result;
	for (unsigned int i = 0; (i < pValues->size()) && (!sqlStatement.Error()); i++)
	{
		sqlStatement.AddParameter((*pValues)[i]);
	}

	if (!sqlStatement.Error())
	{
		sqlStatement.Execute();
	}

	if (!sqlStatement.Error())
	{
		result = safe_query("SELECT changes();");
	}
	else
	{
		if (bLogError)
		{
			_log.Log(LOG_ERROR, "Error performing operation: '%s'", sqlStatement.ErrorText());
		}
	}

	if (result.empty() || result[0][0] == "0")
		return 0;
	else
		return atoi(result[0][0].c_str());
}
```
First look at "**sqlStatement.Execute();**" (this is **CSQLStatement::Execute()**) there is no lock, but it doing sqlite3_step (this is operation on database). 
Next look at **safe_query**, it using **query** and there is lock in **query**. 
Note also that we cannot add lock only in: "**sqlStatement.Execute();**" this is because when there will be some another sql operation between "**sqlStatement.Execute();**" and "**safe_query("SELECT changes();");**" than SELECT changes(); will return not our changes, but some another thread's changes, so both sql operations needs to be locked. 
As alternate way we can write second query method (for example **query_nolock**), and second safe_query (for example **safe_query_nolock**), but it will complicate code even more (note that current SQLHelper contains 10k lines, it would be better to extract data access to some another class, extract thread safe to another and keep only public methods in SQLHelper, but this is very big task).

gizmocuz:
Okey and who is calling this function 'execute_sql' ?
Ahhh someone implemented this for Python

Python should call 'safe_query', like is done everywhere else in the code

That same person also added something like 'CSQLStatement' .... That should also be moved to the python code
It is also only used in the python code.
(No idea what this class is for and why it should be needed... just saying that 100 other classes just use m_sql.safe_query)

Could you try to move this and change the python code to call m_sql.safe_query like everywhere else?

gizmocuz:
I had a look at the python code that calls execute_sql....
Pfffff

I know it is not your or mine code, (PythonObjectEx.cpp , line 625). this has to be changed m_sql.safe_query, there is no need for stacking values and passing that....this is exactly why safe_query was made

I will make the modification right now....

There is a large section of remarked code that could be removed (line 898), I made some changes some time ago, those are working, and forgot about this...

gizmocuz:
I made the patch in the commit above. This should solve the problem. Beta 15619, will take some time before available.

gizmocuz:
Is it possible you could make a separate PR for issue #5630 ?

Krzysztof1234534:
Thanks for fixing Sqlhelper, I've didn't test it yet, but I hope that it solve the problems. 
Regarding cWebem, I've created another PR: https://github.com/domoticz/domoticz/pull/5853

PR #5852, opened 2023-10-24
Do not update blinds on non numeric state change.

ajarzyn:
### States
Controllers state may contain intermediate levels.
In general my controller supports following states:
```
closing
closed
stopped
opening
open
```

### Logs with MQTT value
```
2023-10-24 23:47:33.962 domoticz-mqtt: MQTT received: entry_gate (value: opening)
```

```
2023-10-24 23:48:40.720 domoticz-mqtt: MQTT received: entry_gate (value: closing)
2023-10-24 23:48:47.874 domoticz-mqtt: MQTT received: entry_gate (value: closing)
2023-10-24 23:48:48.109 domoticz-mqtt: MQTT received: entry_gate (value: closed) 
```

```
2023-10-24 23:47:33.301 domoticz-mqtt: MQTT received: entry_gate (value: stopped)
2023-10-24 23:47:36.302 domoticz-mqtt: MQTT received: entry_gate (value: stopped)
2023-10-24 23:48:40.146 domoticz-mqtt: MQTT received: entry_gate (value: stopped) 
```

### Explanation
While `open`, `opening`, `closing`, `closed` are not causing problems. Value is updated level after short period of time or on the limit reached.

`stopped` unfortunately resets the value of the switch to 0 because of this `else` statement:
https://github.com/domoticz/domoticz/blob/0fa4ed5bcf83ba1e828da78895e6f422e00b69da/hardware/MQTTAutoDiscover.cpp#L4287

As value is not json here we were expecting to receive numerical value, when we didn't we do not change level value which was declared earlier https://github.com/domoticz/domoticz/blob/0fa4ed5bcf83ba1e828da78895e6f422e00b69da/hardware/MQTTAutoDiscover.cpp#L4195

It will set `szSwitchCmd`  to value of `"Close"` here:
https://github.com/domoticz/domoticz/blob/0fa4ed5bcf83ba1e828da78895e6f422e00b69da/hardware/MQTTAutoDiscover.cpp#L4299

And later on `nValue` to 0:
https://github.com/domoticz/domoticz/blob/0fa4ed5bcf83ba1e828da78895e6f422e00b69da/hardware/MQTTAutoDiscover.cpp#L4322

### Presentation
On the left is how the instance works with the provided fix.
On the right is presentation of stable release
```
Version: 2023.2 (build 15457)
Build Hash: 25624ca14
```

gate_controller

### Additional information
I've also tested in on the freshly built instance of `development` branch with 0fa4ed5bcf83ba1e828da78895e6f422e00b69da and the behaviour is same as on the right side of the presentation animation.
Verified if Fibaro ZWave Blinds controlled by [zwave-js-ui](https://zwave-js.github.io/zwave-js-ui/) works correctly and it does.

### DIFF
--- a/hardware/MQTTAutoDiscover.cpp
+++ b/hardware/MQTTAutoDiscover.cpp
@@ -4308,6 +4308,10 @@ void MQTTAutoDiscover::UpdateBlindPosition(_tMQTTASensor* pSensor)
 			level = atoi(szSwitchCmd.c_str());
 			szSwitchCmd = "Set Level";
 		}
+		if (pSensor->last_topic == pSensor->state_topic)
+		{
+			return;
+		}
 	}
 
 	if (level == pSensor->position_open)
--- a/hardware/MQTTAutoDiscover.cpp
+++ b/hardware/MQTTAutoDiscover.cpp
@@ -4308,8 +4308,9 @@ void MQTTAutoDiscover::UpdateBlindPosition(_tMQTTASensor* pSensor)
 			level = atoi(szSwitchCmd.c_str());
 			szSwitchCmd = "Set Level";
 		}
-		if (pSensor->last_topic == pSensor->state_topic)
+		else if (pSensor->last_topic == pSensor->state_topic)
 		{
+			// Value for state_topic is not a number.
 			return;
 		}
 	}


gizmocuz:
Thank you for this PR and your research!

Could your patch maybe also block a last position update for sensors that do not send 'stopped' ?
Updated to position could be send to the state topic.
Could you post your MQTT 'config' / json object?
Maybe your device is using a position_topic, in this case, your patch is correct, BUT also check if pSensor->position_topic is not empty like:

```
if (
    (pSensor->last_topic == pSensor->state_topic)
    && (!position_topic.empty())
    ) {
		return;
}
```

If your device does not use a position topic, another way that you could test is to also query the LastLevel on the two select queries (line 4211 and 4223) and set the level to this value?

Or if you are in the //not json part, and the szSwitchCmd is not Open or Close and it is not a numeric value, you exit the function with 'return' as we don't know how to interpratate it

ajarzyn:
Hello @gizmocuz sorry for delay, few tough days.

Frankly sometimes it's hard for me to understand clearly what is on your mind.

> Could your patch maybe also block a last position update for sensors that do not send 'stopped' ?
> Updated to position could be send to the state topic.

Am I understanding that correctly that you are worried that this change will brake devices which send `position value` in `state topic`?
If that's the case you are right it would introduce side defect.

Instead of following your suggestion I propose different approach:

```cpp
		szSwitchCmd = pSensor->last_value;
		if (is_number(szSwitchCmd))
		{
			//must be a level
			level = atoi(szSwitchCmd.c_str());
			szSwitchCmd = "Set Level";
		}
		else if (pSensor->last_topic == pSensor->state_topic)
		{
			// Value for state_topic is not a number.
			return;
		}
```
Why I do not want to bind this code to position_topic?
Code that follows after line 4315 is based on levels.
If we won't set level before then it will always fall through to:
```cpp
	else if (level == pSensor->position_closed)
	{
		szSwitchCmd = "Close";
	}
```
Thus every non-inteager state format will set the device as `Closed`, as it will have `position_topic` empty.
In my opinion 
```cpp
else if (pSensor->last_topic == pSensor->state_topic)
```
is a better option as it will only return in following case:
```cpp
if (last_topic == state_topic && is_number(last_value) == false)
```
That way we make sure that later on we process only numerical values.
***
My config is as follows:
```json
{
    "position_topic": "j3tech/003002202000/cover/position/state",
    "set_pos_t": "j3tech/003002202000/cover/position/set",
    "stat_t": "j3tech/003002202000/cover/state",
    "cmd_t": "j3tech/003002202000/cover/set",
    "pl_cls": "CLOSE",
    "pl_open": "OPEN",
    "avty": [{
            "topic": "j3tech/003002202000/state"
        }
    ],
    "name": "gate-estate-controller",
    "uniq_id": "j3tech_0x6A3374656368003002202000",
    "dev_cla": "gate",
    "dev": {
        "ids": ["j3tech_0x6A3374656368003002202000"],
        "name": "entry_gate",
        "mdl": "J3TECHEGCv1",
        "sw": "48",
        "mf": "j3tech"
    }
}
```

***
My device uses position topic and does not use json.
It's `//not json part`.
And currently the code does not recognize Open/Close - it's based on the level anyway:
```cpp
	if (level == pSensor->position_open)
	{
		szSwitchCmd = "Open";
	}
	else if (level == pSensor->position_closed)
	{
		szSwitchCmd = "Close";
	}
```

But (as I mentioned above) with current implementation, any non numerical value will be interpreted as `Close` state.

Updated in commit https://github.com/domoticz/domoticz/commit/79fc71bd747cfa5501d82f85c1196d306612d203

gizmocuz:
It seems you understand me correctly 😂
And no, English is not my native language, but I try to translate my mind as best as I can.

Is this PR ready for testing, or do you want to make some additional changes to it?

ajarzyn:
> It seems you understand me correctly 😂

Glad to hear that. I was scraching my head aroud this for some time.
Don't you worry I understand "casting" thoughts to outher language is not an easy task, but I think you are doing ok.

I think this PR is ready for testing.

Let me know if I may help with testing.

gizmocuz:
Thanks for the PR. I merged your change!

gizmocuz:
@ajarzyn , In the latest beta version I am now ignoring the state_topic/stat_t when a position topic is configured in the config.
Everything 'should' still work... But I'm not able to verify this.
If possible is it possible to check this?

ajarzyn:
Hello @gizmocuz 

I've tested:
Windows | 2024.2.15855 |   | 2024-01-15 |   | 13,48 MB

1. Checked LED control from Switches :white_check_mark:
   1. Turn LED off :white_check_mark:
   2. Turn LED on :white_check_mark:
   3. Colour transition :white_check_mark:
2. Checked Scenes control
   1. From state off -> Turn on the LED with colour value 1 (blue) :white_check_mark:
   2. From state off -> Turn on the LED with colour value 2 (green) :white_check_mark:
   5. From state on LED colour value 1 -> transition to colour value 2 :white_check_mark:
   6. Transition colour value 2 -> 1 :white_check_mark:
   7. Transition to off state :white_check_mark:

Quick test shows that your change is "a ok".
In case of any problems I will update.

gizmocuz:
@ajarzyn , super thanks for checking!!!

PR #5861, opened 2023-11-02
Update domoticzdevices.js to add Current Graph display and Water Meter Graph Display when activated from a FloorPlan

mealgi:
Add Current Graph display when activated from a Floorplan


mealgi:
Add Water Meter Graph display when activated from a Floorplan

gizmocuz:
Thank you for your PR and work!!

PR #5865, opened 2023-11-04
go to settings again after save

ml17950:
Most times, there are some changes in the settings, and every time after save, you have go from dhashboard to settings again.
Maybe it's better, to return to settings page and if all changes are done, manually go to dhashboard or any other page.


gizmocuz:
Thank you for your PR.

This is as designed. You adjust your settings, and when you are done, you click save.
No need to go back to settings again.

PR #5874, opened 2023-11-15
Update domoticzdevices.js - show current value as status for custom sensor

martin-g-it:
To show current value as status on custom sensor (while hovering over the icon)

### DIFF
--- a/www/js/domoticzdevices.js
+++ b/www/js/domoticzdevices.js
@@ -1541,6 +1541,7 @@ function Custom(item) {
             this.image = "images/Custom.png";
         }
         this.LogLink = this.onClick = "window.location.href = '#/Devices/" + this.index + "/Log'";
+	this.status = this.data;
         this.data = '';
     }
 }


gizmocuz:
Could you try beta 15653?

martin-g-it:
oeh, no, this makes the P1 meter looks terrible on the floorplan:
image

gizmocuz:
Why is your P1 meter looking so much different than mine?
Besides the P1 meter, everything else is OK?
You are using the default theme right?

Without delivery:
image

With return delivery:
image

gizmocuz:
I don't think you are using the default theme looking at your screenshots

gizmocuz:
Ahhh I am looking at the popup when you hover your mouse above it... you are looking at the text underneath it... 
The popup's seems fine now

gizmocuz:
Seems the text beneath is an option as I don't have this here... have to dig into this

image

Anyway, the popups are now correct... that's part 1

gizmocuz:
Could you test beta 15654, it should now be correctly display the values as well

martin-g-it:
> I don't think you are using the default theme looking at your screenshots

It's the Elemental theme with a different background (some carbon look)..

martin-g-it:
> Could you test beta 15654, it should now be correctly display the values as well

Will check tomorrow; busy day and plans tonight

martin-g-it:
Partially fixed..

The P1 icon shows the current usage (instead of CounterToday)
image
Not sure if this is better, since you also have a dedicated device for the actual usage..

The Current icon seems OK now:
image

Although for the popup when you hoover over it becomes unreadable when the device name is longer than 18 characters.
image
Would be better to show it like this:
image

Which can be achieved by adding below case in the `switch (this.type) {:`
```
>             case "Current":
>                 this.LogLink = this.onClick = "ShowCurrentLog('#" + Device.contentTag + "','" + Device.backFunction + "','" + this.index + "','" + this.name + "', '" + this.switchTypeVal + "');";
>                 this.smallStatus = this.data;
>                 this.status = this.data;
>                 this.data = '';
>                 break;
```

gizmocuz:
Thank you for your feedback. I consider this fixed. (And fixed for you as well as you rename your device)

martin-g-it:
Hmm, let me create a new PR, because the P1 device is now even worse when power is being delivered:
image

Really better to show the CounterToday..

gizmocuz:
What do I see above? What is your current usage, don't you want to see your house usage? Like it is using 1000 W, or it is delivering 300 Watt (-300 W)?

martin-g-it:
delivering 300 Watt... I got it all sorted out now (incl. some more fixes w/o impacting others), will make a PR in a few

martin-g-it:
see https://github.com/domoticz/domoticz/pull/5878

PR #5876, opened 2023-11-15
Update domoticzdevices.js - Set 3 phases of a 'Current' device type on the status

martin-g-it:
Set 3 phases of a 'Current' device type on the status

image
instead of 
image

### DIFF
--- a/www/js/domoticzdevices.js
+++ b/www/js/domoticzdevices.js
@@ -864,7 +864,9 @@ Device.create = function (item) {
 	  ) {
         type = item.SwitchType.toLowerCase()
     } else if (item.Type === 'RFXMeter') {
-        type = 'counter'; 
+        type = 'counter';
+    } else if (item.Type === 'Current') {
+        type = 'current';
     } else {
         type = item.TypeImg.toLowerCase();
         if (item.CustomImage !== 0) {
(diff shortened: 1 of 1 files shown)


gizmocuz:
I am not sure if this PR is correct

Codewise I would change things like


```
this.smallStatus = this.data;
this.smallStatus = this.smallStatus.split(', ')[0];
```
to

`this.smallStatus = this.data.split(', ')[0];`

But you changed a lot of code just for that one 3 phase current sensor, you also changed others

I would expect that in the 'data' field all three are present when they are used and that this will be set from within Domoticz
(As it is also used in the normal GUI/Devices tab)

gizmocuz:
Looking at my Current meter, it displays correctly already like on the utility/dashboard page

image

martin-g-it:
On the dashboard it looks fine but on the Floorplan it doesn't (only the current of the first phase is shown)
Thats why I removed the ` this.smallStatus = this.smallStatus.split(', ')[0];`

To avoid impact on other sensor types, I put that line in the other branches of the `switch` 
image

Indeed it can be done more efficient now by `this.smallStatus = this.data.split(', ')[0];`

gizmocuz:
Well, what you do above is split the 3 values and take only the fist value... that does not seem correct?
Seems the oposite of what you want

Domoticz itself will decide if the other two values are used or not, depending if there are used internally.
The same 'data' you see on the dashboard is the same 'data' that is send to the floorplan (same device object)

martin-g-it:
Indeed, this is not what I want for my 3 phase current sensor hence why I didn't put this split in the newly added section for device type Current:
image

gizmocuz:
Could you review this PR as you made changes for lots of devices which should not be needed.
Did you see my screenshot above? This is from the floorplan and seems correct

PR #5878, opened 2023-11-17
Update domoticzdevices.js - bugfix on floorplan visualization

martin-g-it:
Bugfix on floorplan visualization

### DIFF
--- a/www/js/domoticzdevices.js
+++ b/www/js/domoticzdevices.js
@@ -1451,21 +1451,22 @@ function Counter(item) {
         }
         this.LogLink = this.onClick = "window.location.href = '#/Devices/" + this.index + "/Log'";
 
+	this.smallStatus = this.data;
         if (typeof item.CounterToday != 'undefined') {
 			this.status = this.data;
 			this.data = item.CounterToday;
+			this.smallStatus = item.CounterToday;			
         }
         if (typeof item.CounterDeliv != 'undefined') {
 			this.data = this.status;
 			this.status = $.t("Usage") + ': ' + item.CounterToday;
             if (item.CounterDeliv != 0) {
                 if (item.UsageDeliv.charAt(0) != 0) {
-                    this.data += '-' + item.UsageDeliv;
+                    this.data += ' -' + item.UsageDeliv;
                 }
                 this.status += ', ' + $.t("Return") + ': ' + item.CounterDelivToday;
             }
         }
-		this.smallStatus = this.data;
     }
 }
 Counter.inheritsFrom(UtilitySensor);
@@ -1527,6 +1528,12 @@ function Current(item) {
                         break;
                 }
                 break;
+            case "Current":
+                this.LogLink = this.onClick = "ShowCurrentLog('#" + Device.contentTag + "','" + Device.backFunction + "','" + this.index + "','" + this.name + "', '" + this.switchTypeVal + "');";
+                this.smallStatus = this.data;
+                this.status = this.data;
+                this.data = '';
+                break;
             default:
                 this.LogLink = this.onClick = "ShowCurrentLog('#" + Device.contentTag + "','" + Device.backFunction + "','" + this.index + "','" + this.name + "', '" + this.switchTypeVal + "');";
                 break;
@@ -1544,7 +1551,8 @@ function Custom(item) {
             this.image = "images/Custom.png";
         }
         this.LogLink = this.onClick = "window.location.href = '#/Devices/" + this.index + "/Log'";
-        this.data = '';
+        this.status = this.data;
+	this.data = '';
     }
 }
 Custom.inheritsFrom(UtilitySensor);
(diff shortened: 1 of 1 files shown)


martin-g-it:
**P1 meter**
image
image

**Custom sensor:**
image
image

**Current sensor:**
image
image

martin-g-it:
Last one item also correct; General kWh meter:
image

image

martin-g-it:
And added another fix; Dimmer switch with custom icon didn't have the slider:

**Dashboard:** 
image

**Floorplan:**
image

**Floorplan popup:**
image

**After fix:**
image

gizmocuz:
The P1 meter can not be 0 Watt - 344 Watt ... that is odd... it should be -344 Watt.
And that is what it currently displays before this PR correct?

Why do you want to see the total kWh value of today? (And not the actual usage like 614 Watt)
The later is much more important for most users?

The Enphase production has a Watt and a kWh counter value, I can't see the actual Watt value now?

Your names are also to big...

martin-g-it:
Actually you can have usage and delivery at the same time (with 3 phases).

Originally, the counterToday value was shown on the floorplan label. On the popup (and dashboard), the actual usage like -344 was shown along with the daily counters etc..

Like I said, there is Actual usages sensors provided by the P1 hardware type, so if you want to show the actual usage (Wattage), you can bring in such sensor.

The Enphase sensor has the wattage on the floorplan icon, so I created a custom sensor with today's total production.

For your reference; this is the floorplan overview I created with all utility stuff:
image

gizmocuz:
I really have no idea why you need to create a custom sensor for Enphase as the today kWh counter is already present

image

So the popup in the floorplan could show this.

The P1 meter can have delivery+usage but the result must be 'usage-delivery', and that is what is also visible on the dashboard/utility tab correct?

What you are trying to do is maybe also possible by using a custom template page, see the example
Or using Influx in combination with Grafana

To recap:
Displaying the current P1 usage on the floorplan is preferred over a counter.
If you want to see the counter, just use the popup

image

(Why... because you will see directly a change when you turn on/off something, and you won't see a kWh increase unless you stare at it for a long time)

The current L1/L2/L3 display is perfect:
image

image

That you use a very large name is something you only can solve (..... hint: shortening that name!)

Let me remind you that the Floorplan is a way different representation than the Dashboard/Tabs where you have widgets that tells you everything in one view. (But that needs a large display, and you don't have this in floorplans)

You could make a new PR for the Volume only if you like, where it should display 34% as text underneath, and when popup, it should look exactly like the dashboard/utility tab

martin-g-it:
> I really have no idea why you need to create a custom sensor for Enphase as the today kWh counter is already present

On the floorplan, the 'standard' sensor gives the actual production as small status; which is good information.
I also want to see the production of today on that same overview (without the need to hover or click on the sensor). Thats why I've created the custom sensor (along with some others).

> The P1 meter can have delivery+usage but the result must be 'usage-delivery', and that is what is also visible on the dashboard/utility tab correct?

Indeed, but this seems OK now.
 
> What you are trying to do is maybe also possible by using a custom template page, see the example Or using Influx in combination with Grafana

I know, but I'd like to keep everything at one place; thats why I use a floorplan functionality for this overview/dashboard.

> To recap: Displaying the current P1 usage on the floorplan is preferred over a counter. If you want to see the counter, just use the popup

Thats the thing; seeing all information in an overview/floorplan without the need to click/hover on those sensors.
Plus the fact that there is a different sensor already with the actual usage/wattage provided by the P1 hardware 
The kWh usage of today has been on the floorplan initially (in the stable version). 

> The current L1/L2/L3 display is perfect

Yes indeed :-)

> That you use a very large name is something you only can solve (..... hint: shortening that name!)

It's not a big thing, but cosmetically I think this looks a bit nicer due to icon scaling/size on the floorplan and this only applies to the floorplan
image
On the utility tab it still looks as-is which is OK:
image


> Let me remind you that the Floorplan is a way different representation than the Dashboard/Tabs where you have widgets that tells you everything in one view. (But that needs a large display, and you don't have this in floorplans)

I use it all; to me, the floorplan feature works perfectly for (additional) dashboards like utility overview / server information easily accessible on the phone.

gizmocuz:
Cosmetically the L1/L2/L3 is looking correct as it is.

The actual usage/return (in Watt) for the P1 meter is way more important than the total usage/return of the day.
This information is available in the popup

So that being said, there is probably indeed something wrong with the volume sensor...
Could you make a new PR for that one and close this one?

On second guess.. I just tried the sound device on the floorplan it and displays perfectly:

image

image

There is no need for a change.....?

No changes are needed, please close this PR

gizmocuz:
(Except for the icon state, it is displaying the off state icon, I will fix this)

gizmocuz:
Changed it and now it looks at the Data (0 dB = Off state else On state)

image

martin-g-it:
Let's agree to disagree ;-)

I will make my preferred setup locally and update along new Domoticz (stable) versions..

Will also make a PR for the Volume percentage slider fix (its actually a Dimmmer switch, not Sound).

PR #5884, opened 2023-11-19
Update domoticzdevices.js - fix dimmer switch on Floorplan

martin-g-it:
Fix dimmer switch on Floorplan

### DIFF
--- a/www/js/domoticzdevices.js
+++ b/www/js/domoticzdevices.js
@@ -859,7 +859,8 @@ Device.create = function (item) {
 	} else if ((item.Type === 'General') && (item.SubType === 'Percentage')) {
 		type = 'percentage';
     } else if (
-        (item.SwitchType === 'Dusk Sensor') ||
+        (item.SwitchType === 'Dimmer') ||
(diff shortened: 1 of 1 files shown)


gizmocuz:
Could you display here how it looks with your patch?

Here, using a dimmer on a floorplan without your patch it looks like:

image

Which seems OK

The popup looks like:
image

martin-g-it:
Thats because it has the **default** Icon I guess... try it with a custom one.

martin-g-it:
Hmm, the P1 meter (with your setting on smallStatus) is still looking odd once there is delivery:
image

gizmocuz:
Gotha... Why is that last change on line 1540 needed in your PR?

martin-g-it:
To show the status/value of a Custom sensor on the popup when you hover over it.

Type = General, SubType  = Custom Sensor

Before:
image

After:
image

gizmocuz:
Ahh okey, because this PR is about "fix dimmer switch on Floorplan"

Thanks for the PR

PR #5889, opened 2023-11-21
Update domoticzdevices.js - fix P1 meter on floorplan when meter has usage delivery (return usage)

### LINKED ISSUE #5888: P1 meter smallStatus on Floorplan is still looking odd once there is delivery/return
P1 meter smallStatus on Floorplan is still looking odd once there is delivery/return usage
image

_Originally posted by @martin-g-it in https://github.com/domoticz/domoticz/issues/5884#issuecomment-1819111255_

Same applies also to the dashboard:
image

With just return it looks like this on the dashboard:
image

And like this on the floorplan:
image

martin-g-it:
Fix P1 meter (counter) on floorplan when meter has usage delivery (return usage)

See issue https://github.com/domoticz/domoticz/issues/5888

### DIFF
--- a/www/js/domoticzdevices.js
+++ b/www/js/domoticzdevices.js
@@ -1461,7 +1461,7 @@ function Counter(item) {
             this.status = $.t("Usage") + ': ' + item.CounterToday;
             if (item.CounterDeliv != 0) {
                 if (item.UsageDeliv.charAt(0) != 0) {
-                    this.data += ' -' + item.UsageDeliv;
+                    this.data += ', -' + item.UsageDeliv;
                 }
                 this.status += ', ' + $.t("Return") + ': ' + item.CounterDelivToday;
             }
--- a/www/js/domoticzdevices.js
+++ b/www/js/domoticzdevices.js
@@ -860,7 +860,7 @@ Device.create = function (item) {
 		type = 'percentage';
     } else if (
         (item.SwitchType === 'Dimmer') ||
-        (item.SwitchType === 'Dusk Sensor') ||	    
+        (item.SwitchType === 'Dusk Sensor') ||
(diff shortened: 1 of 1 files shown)


martin-g-it:
Added the comma to the status

image
in line with dashboard:
image

gizmocuz:
Thanks for the PR. When this happens, how does the P1 widget on the dashboard looks like?

martin-g-it:
> Thanks for the PR. When this happens, how does the P1 widget on the dashboard looks like?

With both actual usage and actual return it looks like the above screenshot
>  in line with dashboard: 
image

Device status like this:
image

-----------------

With 0 usage and eg. 456 Watt return like this:
image

Device status like this:
image

-----------------

To complete; with 123 usage and 0 return:
image

Device status like this:
image

-----------------
I just made one more adjustment to align the **floorplan popup** with the dashboard in case of 0 usage, 456 return (also added the various smallstatus for the floorplan for completeness:
image
image

-----------------

With 123 / -456
image
image

-----------------

With 123 / 0
image
image

gizmocuz:
Thanks for this ! Looking great now!

gizmocuz:
Thank you for this PR and your time! Going to solve this also in the Dashboard/Utility tab

martin-g-it:
I think for the dashboard it looks OK (it is aligned with the popup).
On that popup/dashboard there is enough space to show both the usage and return in case both apply.

Dashboard:
image

Floorplan popup:
image

For the floorplan smallstatus, it makes more sense to balance them like done (and get the full info when you hover over it).

gizmocuz:
true, but it looks a bit wierd, but these are indeed the actual values reported by the meter.
Okey leaving it like it is ;)

PR #5853, opened 2023-10-25
cWebem isnew connection problem

Krzysztof1234534:
Extracted cWebem changes from https://github.com/domoticz/domoticz/pull/5849 pull request as SQLHelper problem was fixed different way.

PR #5904, opened 2023-11-30
Refactor Webserver on incomingconnection and more

kiddigital:
A PR that refactors parts of the webserver code handling the sessions, cookies and auth checking.

PR #5853 from @Krzysztof1234534 was the reason to look at this piece of the code and start the refactoring.

It does NOT (yet) solve the problem of the many 'Incoming connection' messages when in local network with no active session and the browser firing a lot of simultaneous requests (which do not have a session cookie either as they are fired in parallel).

Some older pieces of code from the old implementation of local networks (when the session was set to Admin rights but no real admin user was used) have been removed.

### DIFF
--- a/main/Helper.cpp
+++ b/main/Helper.cpp
@@ -1626,7 +1626,7 @@ char* make_web_time(const time_t rawtime)
 	if (gmtime_r(&rawtime, &gmt) == nullptr)
 #endif
 	{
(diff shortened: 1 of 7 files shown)


kiddigital:
@gizmocuz , ready for review and merge :) 

When merging, please do not squash the commits so we can restore individual commits in the unlikely event that somehow something broke.

gizmocuz:
Thank you!

PR #5905, opened 2023-11-30
Added specific notification handler for alert sensor

### LINKED ISSUE #5879: Feature Request: Addition of $text Variable for Custom Message Notifications of  Alert sensors
**In fact for the sensor alert, the $value injection is not good. We should have the reason (text) of the alert instead. We don't care about having the level number, since we can correspond to a level in the notification settings.**

### References

https://github.com/domoticz/domoticz/blob/dcded44e5787e6b25eb2a4c605275473125d7853/notifications/NotificationHelper.cpp#L234-L240

https://github.com/domoticz/domoticz/blob/dcded44e5787e6b25eb2a4c605275473125d7853/notifications/NotificationHelper.cpp#L465-L466

### Issue
_The custom message notification feature allows the injection of variables like $name and $value. However, for more versatile and detailed alerts, it would greatly enhance functionality to introduce a new variable, $text, which could be utilized in notifications for alert sensors and similar widgets._

### Proposal
_The addition of the $text variable would enable users to create more informative and context-rich custom messages, allowing for better communication of sensor alerts._

### Use Case
_For instance, when an alert sensor is triggered, having access to a $text variable would enable users to include specific details or additional context related to the alert in their custom notification messages. This could significantly improve the usefulness and relevance of notifications, aiding users in taking prompt and appropriate actions based on the received information._

bierdok:
This PR addresses an issue related to the sensor alert functionality. The $value parameter injected and used for sensor alerts has been replaced with a more informative approach instead of focusing on the level associated.

This adjustment eliminates the useless need for displaying the level number since it can be mapped to a corresponding level in the notification settings. This modification ensures that alert notifications provide a clearer understanding of the issue at hand by focusing on the descriptive text rather than a numeric value, thereby enhancing the overall system's usability and comprehension.

**Test**

1. Create a Dummy Hardware.
2. Create an Alert called "House".
3. Add the notification:

```
Type: Usage
When: Greater or Equal
Value: 0
Pririty: Emergency
Ingore Interval: true
Recovery Notification: false
Custom Message: House is $value
Active Systems: browser
```

4. Add the following script: _dev-domoticz/scripts/dzVents/scripts/test.lua_

```
return {
  active = true,
    on = {
      ['timer'] = {
        'every minute'
      }
    },
    execute = function(domoticz)
      domoticz.devices('House').updateAlertSensor(domoticz.ALERTLEVEL_RED, 'burning!')
    end
}
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -175,7 +175,6 @@ push/FibaroPush.cpp
 push/GooglePubSubPush.cpp
 push/HttpPush.cpp
 push/InfluxPush.cpp
-push/MQTTPush.cpp
 push/WebsocketPush.cpp
 httpclient/HTTPClient.cpp
 httpclient/UrlEncode.cpp


kiddigital:
@bierdok , can you please rebase your PR to the updated Development branch so the PR check will succeed? Thx.

PR #5906, opened 2023-12-01
MQTT-AD: Fix autodiscovery with prefix that contains slashes

dwmw2:
Because I cargo-culted my esphome MQTT setup from https://esphome.io/components/mqtt.html#tls-with-esp-idf-esp32 it ended up using:

```
discovery_prefix: ${mqtt_prefix}/homeassistant
```

I only just updated to a version of Domoticz which can do HA autodiscovery, and tried to get it working.

And Domoticz doesn't cope. Not because of the weird bit that looks like a variable expansion but isn't _(it literally is using `${mqtt_prefix}`._ But because of the slash.

The prefix matching in `on_message()` does a direct comparison of `m_TopicDiscoveryPrefix` against the first element (up to the first `/`) of the topic, which doesn't work if the `m_TopicDiscoveryPrefix` contains more than one path element.

### DIFF
--- a/hardware/MQTTAutoDiscover.cpp
+++ b/hardware/MQTTAutoDiscover.cpp
@@ -76,7 +76,7 @@ void MQTTAutoDiscover::on_message(const struct mosquitto_message* message)
 			return;
 
 		if (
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thank you!

PR #5908, opened 2023-12-01
Added specific notification handler for alert sensor

bierdok:
This PR addresses an issue related to the sensor alert functionality. The $value parameter injected and used for sensor alerts has been replaced with a more informative approach instead of focusing on the level associated.

This adjustment eliminates the useless need for displaying the level number since it can be mapped to a corresponding level in the notification settings. This modification ensures that alert notifications provide a clearer understanding of the issue at hand by focusing on the descriptive text rather than a numeric value, thereby enhancing the overall system's usability and comprehension.

**Test**

1. Create a Dummy Hardware.
2. Create an Alert called "House".
3. Add the notification:

```
Type: Usage
When: Greater or Equal
Value: 0
Priority: Emergency
Ignore Interval: true
Recovery Notification: false
Custom Message: House is $value
Active Systems: browser
```

4. Add the following script: _dev-domoticz/scripts/dzVents/scripts/test.lua_

```
return {
  active = true,
  on = {
    ['timer'] = {
      'every minute'
    }
  },
  execute = function(domoticz)
    domoticz.devices('House').updateAlertSensor(domoticz.ALERTLEVEL_RED, 'burning!')
  end
}
```

### DIFF
--- a/notifications/NotificationHelper.cpp
+++ b/notifications/NotificationHelper.cpp
@@ -462,7 +462,6 @@ bool CNotificationHelper::CheckAndHandleNotification(const uint64_t DevRowIdx, c
 					return CheckAndHandleNotification(DevRowIdx, sName, cType, cSubType, NTYPE_PERCENTAGE, fValue);
 				case sTypeSoilMoisture:
 				case sTypeLeafWetness:
-				case sTypeAlert:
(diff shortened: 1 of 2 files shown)


gizmocuz:
Thank you for this PR!!

PR #5910, opened 2023-12-03
Correction of tomorow information

lcha78:
Hello,
tomorow information was not right. Tomorow (witch is not realy a good name for this information but this is the name used !) information is set at 20h and valid till 6h. Then this is invalid (next tempo period unknown) from 6h to 20h. 
N.B. : "----" is the way to display unknown on linky electric meter.


gizmocuz:
Thank you! Feel free to correct tomorrow

PR #5918, opened 2023-12-05
Refactor of cWebem and adding webserver session gherkin tests

kiddigital:
Refactor of some parts of the cWebem.cpp and removal of obsolete code.

- Removed deprecated code
- Parsing of cookie and checking of Trustednetwork now only performed once (instead of multiple times)
- Added inline documentation for better readability of what is going on
- Session and cookie creation only done when needed (not when calling the API or when in Trusted Network)
- Code moved to better/scoped locations and removed when redundant
- and more cleanup/refactoring

Also added some automated functional tests for the webserver session.

This is a replacement of parts of #5904 (other pieces of that PR are done in PR #5917).

### DIFF
--- a/webserver/cWebem.cpp
+++ b/webserver/cWebem.cpp
@@ -176,8 +176,8 @@ namespace http {
 		// Show a Debug line with the registered functions, actions, includes, whitelist urls and commands
 		void cWebem::DebugRegistrations()
 		{
-			_log.Debug(DEBUG_WEBSERVER, "cWebEm Registration: %d pages, %d actions, %d whitelist urls, %d whitelist commands",
-				(int)myPages.size(), (int)myActions.size(), (int)myWhitelistURLs.size(), (int)myWhitelistCommands.size());
+			_log.Debug(DEBUG_WEBSERVER, "cWebEm Registration: %zu pages, %zu actions, %zu whitelist urls, %zu whitelist commands",
+				myPages.size(), myActions.size(), myWhitelistURLs.size(), myWhitelistCommands.size());
 		}
 
 		std::istream & safeGetline(std::istream & is, std::string & line)
@@ -1804,7 +1804,7 @@ namespace http {
 
 		bool cWebemRequestHandler::CheckAuthentication(WebEmSession & session, const request& req, reply& rep)
 		{
-			session.rights = -1; // no rights
+			session.rights = URIGHTS_NONE; // no rights
 			session.id = "";
 			session.username = "";
 			session.auth_token = "";
@@ -1813,6 +1813,7 @@ namespace http {
 			if (myWebem->m_userpasswords.empty())
 			{
 				_log.Log(LOG_ERROR, "No (active) users in the system! There should be at least 1 active Admin user!");
+				session.reply_status = reply::internal_server_error;
 				return false; // No users in the system!
 			}
 			else if (AreWeInTrustedNetwork(session.remote_host))
@@ -1826,7 +1827,7 @@ namespace http {
 						break;
 					}
 				}
-				if (session.rights == -1)
+				if (session.rights == URIGHTS_NONE)
 				{
 					_log.Debug(DEBUG_AUTH, "[Auth Check] Trusted network exception detected, but no Admin User found!");
 					//If the User database table is without an Admin, we will create a temporary Admin user (we are in trusted network anyway)
@@ -1846,7 +1847,7 @@ namespace http {
 					session.isnew = false;
 					session.rememberme = false;
 					session.username = _ah.user;
-					session.rights = std::atoi(_ah.qop.c_str());
+					session.rights = static_cast<_eUserRights>(std::atoi(_ah.qop.c_str()));
 					return true;
 				}
 				else if (_ah.method == "BASIC")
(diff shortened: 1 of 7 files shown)


gizmocuz:
Thank you vert much for this!!!

Is it possible to name the operator"" function? like _sh ? (StringHash)
It looks (for me) a bit weird to call a function with the name '_'

webserver/cWebem.cpp
```diff
@@ -2195,13 +2067,22 @@ namespace http {
 
 			// Initialize session
 			WebEmSession session;
+			session.id = "";
```
gizmocuz: No need to initialize a string with empty

webserver/cWebem.cpp
```diff
@@ -2195,13 +2067,22 @@ namespace http {
...
+			session.username = "";
```
gizmocuz: No need to initialize a string with empty

webserver/cWebem.cpp
```diff
@@ -2195,13 +2067,22 @@ namespace http {
...
+			session.rights = -1;
+			session.reply_status = reply::ok;
+			session.isnew = false;
```
gizmocuz: No need to initialize variables that are already initialized in the constructor of the object. (also below)

webserver/cWebem.cpp
```diff
@@ -199,70 +180,6 @@ namespace http {
 				(int)myPages.size(), (int)myActions.size(), (int)myWhitelistURLs.size(), (int)myWhitelistCommands.size());
```
gizmocuz: Perhaps we should use %zu so we do not have to cast to an integer?

webserver/cWebem.cpp
```diff
@@ -1886,16 +1804,16 @@ namespace http {
 
 		bool cWebemRequestHandler::CheckAuthentication(WebEmSession & session, const request& req, reply& rep)
 		{
-			bool bTrustedNetwork = false;
-
 			session.rights = -1; // no rights
 			session.id = "";
 			session.username = "";
 			session.auth_token = "";
+			session.istrustednetwork = false;
```
gizmocuz, line 1811: No need to initialize this to false as this is done in the constructor of the object, same for the above strings

webserver/cWebem.cpp
```diff
@@ -1886,16 +1804,16 @@ namespace http {
 
 		bool cWebemRequestHandler::CheckAuthentication(WebEmSession & session, const request& req, reply& rep)
 		{
-			bool bTrustedNetwork = false;
-
 			session.rights = -1; // no rights
```
gizmocuz: Maybe we should create a new (first) enum entry in **_eUserRights** for this and use this instead of -1
Should this be the default in the object's constructor? If yes, then no need to initialize it here

webserver/cWebem.cpp
```diff
@@ -1886,16 +1804,16 @@ namespace http {
...
 			session.id = "";
 			session.username = "";
 			session.auth_token = "";
+			session.istrustednetwork = false;
 
 			if (myWebem->m_userpasswords.empty())
 			{
 				_log.Log(LOG_ERROR, "No (active) users in the system! There should be at least 1 active Admin user!");
+				return false; // No users in the system!
```
gizmocuz, line 1817: I am just thinking aloud here...
At some parts in the function we return false, because there is an error, not because of the authentication check.
Maybe before the 'return false', add a line like

`	session.reply_status = reply::internal_server_error;`

Then at the point after where '**CheckAuthentication**' is called, we check if the reply_status is anything other than 'reply::ok', and if that is the case, exit, so the user received this error

gizmocuz (commented):
Thanks for the PR. Just a few comments/suggestions.
I know some already existed in the current code, but since you are making changes to it, maybe consider taking them into account as well.

I will test the working next...

Also, when merging, make it a Squash-Merge because of the many additions and forces

gizmocuz:
When I try to login for the first time, and am presented with the login screen (username/password), I receive about **142** ! log lines with:

Incoming connection from: xx.yy.xx.yy

It seems for every resource (image/html/js) this line is logged

Suggestion:
If the user needs to login, only add this line at the place where the login is requested (where a 401 is send, but 1 time)

webserver/cWebem.cpp
kiddigital: done

webserver/cWebem.cpp
kiddigital: done

webserver/cWebem.cpp
kiddigital: done

gizmocuz:
Is this PR ready for merge?

kiddigital:
From my perspective, yes. All my testing was positive, but might be a good idea to build the Windows version and give that a quick spin just to be sure.

gizmocuz:
Maybe we need to change two lines of code in 'WebServerCmds.cpp'

CWebServer::Cmd_GetMyProfile

if (session.rights > 0)	// Viewer cannot change his profile

and

CWebServer::Cmd_UpdateMyProfile

if (req.method == "POST" && session.rights > 0)	// Viewer cannot change his profile

This because the 'none' rights is 254, it used to be -1 ?

gizmocuz:
Thank you for this PR and all the work!!!! 🙌👌

Issue #5916, opened 2023-12-05
Domoticz does not get the Python Errors occured during the handling of the OnStopCallback message


mealgi:
Hello,

Context:
Version: 2023.2 (build 15670)
Build Hash: 70ed681d6-modified
Compile Date: 2023-11-24 08:59:28
dzVents Version: 3.1.8
Python Version: 3.9.2 (default, Mar 12 2021, 04:06:34) [GCC 10.2.1 20210110]

Issue:
Domoticz does not get the Python Errors occured during the handling of the OnStopCallback message

This is due to the fact the function PyErr_Occurred() is not called at the end of the  OnStopCallback message

Fixing proposition:
In Plugins.cpp before the line 2300 //Stop Python
insert the following lines:
			if (m_PyInterpreter)
			{
				if (PyErr_Occurred()) // get the errors occured during onStopCallback message handling
				{
					Log(LOG_NORM, "Python error was set during onStopCallback for '%s'",  m_PluginKey.c_str());
					LogPythonException();
					PyErr_Clear();
 				}
			}

I will create a dedicated PR
BR

gizmocuz:
Thank you for your PR. It's merged!

mealgi:
Fixed and merged

PR #5919, opened 2023-12-05
Domoticz does not get the Python Errors occured during the handling o…

mealgi:
Domoticz does not get the Python Errors occured during the handling of the OnStopCallback message #5916


gizmocuz:
Thank you for this PR!

PR #5935, opened 2023-12-17
Removed some double semicolon

Olifant1990:


gizmocuz:
Thanks!!

gizmocuz:
Found one more in EvohomeWeb.cpp (already made the patch)

PR #5942, opened 2023-12-21
hardware/plugins/DelayedLink.h: Fix memleak

thertp:
From python git history, looks like the _Py_Dealloc macro is gone but not the function in the library. So, it has to be defined. Moreover, not defining it leads to Py_*DECREF not freeing any memory when the object refcount is reaching 0. For instance, this leads to a memory leak, for instance everytime CDevice_refresh().

Tested with domoticz HEAD "MQTT-Push, only start when enabled" with python 3.9 (Debian bullseye) and 3.11 (Debian bookworm).


gizmocuz:
Thank you for this PR!!

pipiche38:
_Py_Dealloc is part of  interpreter core, I don't think you have to define.

gizmocuz:
@pipiche38  what do you mean?

PR #5943, opened 2023-12-21
main/EventsPythonModule.cpp: Workaround scope issue

thertp:
The switch to python stable ABI switched to Py_CompileString+PyEval_EvalCode to run code but this leads to NameErrors at run time. For instance, simple code now fails:

```
import DomoticzEvents as DE

def print_msg(msg):
    DE.Log(msg)

def foo():
    print_msg(foo)

foo()
```

As a quick workaround, pass globals dict as locals dict to PyEval_EvalCode. I'm no python embedding expert so while it's working:
- I'm not sure it's 100% correct
- the refcount semantics of PyEval_EvalCode is not totally clear to me but it seems to be fine.

### DIFF
--- a/main/EventsPythonModule.cpp
+++ b/main/EventsPythonModule.cpp
@@ -499,7 +499,7 @@ namespace Plugins
 				PyNewRef	pCode = Py_CompileString(PyString.c_str(), filename.c_str(), Py_file_input);
 				if (pCode)
 				{
-					PyNewRef	pEval = PyEval_EvalCode(pCode, global_dict, local_dict);
+					PyNewRef	pEval = PyEval_EvalCode(pCode, global_dict, global_dict);
 				}
 				else
 				{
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thank you for your PR! I think you pass "locals dict" as "globals dict" to PyEval_EvalCode 😊

I am no expert as well, but in a lot of places I see both local and global dicts are passed like

`PyNewRef	pEval = PyEval_EvalCode(pCode, global_dict, local_dict);`

Maybe this is the solution?

gizmocuz:
PyEval_EvalCode has 3 arguments , the last two are for global and local dict, your are now twice passing global_dict

thertp:
Passing the global_dict as locals is the way I'm workarounding/fixing the issue. Otherwise the code won't find the functions. I don't have links at hand but other people have the same issue (I managed to reproduce it in C) and their solution was to do what this patch is doing. Nevertheless, as said in commit message, I'm not sure in 100% correct.

gizmocuz:
But why are you using the Domoticz object like

import DomoticzEx as Domoticz
Domoticz.Log("Hello World")

or

import Domoticz as Domoticz
Domoticz.Log("Hello World")

gizmocuz:
By accident I found out what part of Domoticz and Python you are referring to...
This is used in Setup->Events->Python and not in an external python (hardware) script.

waltervl:
This PR got accidentally closed by github actions. Perhaps reopen?

gizmocuz:
No feedback or review... closing issue. Feel free to request to re-open if needed

PR #5944, opened 2023-12-23
Enphase: force reading from 'production[0]' with Envoy-S Metered

berkrub:
Added a setting that forces reading production data from production[0] even if a production[1] exists, and 'whLifetime' > 0.
Mode3 has to be set to 'true' to make this work. When set to 'false' nothing changes current behaviour.


gizmocuz:
Thank you for your PR.
Is it possible to make some changes? I don't know what editor you use, but maybe use Visual Studio if you're on windows

If you look in the 'Files changed' tab above, you will see at some point you added a extra tab which should not be there, and a space as well

And you also changed the degrees sign in the code that should not be done

berkrub:
Hi Gizmocuz,

I have fixed the errors you mentioned.
I guess that due to some kind of font / localization mistake the degrees sign where malformed.

gizmocuz:
@berkrub , thank you for this!
It's weird, but there is still a change in mainworker line 12970/12971
Maybe you could checkout the original file, and make this change manually?
You can use Visual Studio or Notepad2 for instance:

https://www.flos-freeware.ch/notepad2.html

berkrub:
Hi @gizmocuz, that is weird indeed.
I am using Visual Studio Code on Linux Mint. Normally I don't have any issues with Unicode.

I have done a `git clone` on both development branch and my own branch.
When I do a diff on the mainworker.cpp file, actually the **original** file is showing a "�" sign before the "C" and "F" in lines 12970 and 12971.

So I ran `file` on both files, with the following outcome:
`dev-domoticz/main/mainworker.cpp: C++ source, ISO-8859 text, with very long lines`

`dev-domoticz-berkrub/main/mainworker.cpp: C++ source, UTF-8 Unicode text, with very long lines`

So apparently VS Code changed the file encoding to UTF-8?

What I did now is convert the character encoding of the file back to ISO-8859-1.
Now, VS Code is showing the question marks again, but VIM for example doesn't.

I hope it is alright now...!?

berkrub:
Hi @gizmocuz,

I was wondering if you have found any time too look to merge this pull request?
It would be great if it is merged, then I can update Domoticz again from the upgrade channel :-)

Best regards,
Rob

gizmocuz:
Sure! I was away on a trip ;) ... Thanks for the PR!!

berkrub:
It was my pleasure :-)
Thanks a lot from a happy Domoticz user!

Rob

Issue #5896, opened 2023-11-26
Shelly RGBW support in MQTT-AD


enesbcs:
I am wondering if it is possible to support Shelly RGB & RGBW devices in MQTT Autodiscovery?

I created a configuration payload for ShellyTeacher script integration. Possible it is not perfect, it is based on the previous dimmer device, and some parts may be not needed.

topic:
`homeassistant/light/shellybulb-ID-0-color`

Payload:
```
{
    "~": "shellies/shellybulb-ID/",
    "name": "shellybulb-ID Color 0",
    "uniq_id": "shellybulb-ID-color-0",
    "command_topic": "~color/0/set",
    "payload_on": "{\"turn\":\"on\",\"mode\":\"color\"}",
    "payload_off": "{\"turn\":\"off\",\"mode\":\"color\"}",
    "state_topic": "~color/0/status",
    "val_tpl": "{{value_json.ison}}",
    "stat_val_tpl": "{{value_json.ison}}",
    "state_on": "true",
    "state_off": "false",
    "color_mode": true,
    "supported_color_modes": [
        "rgbw",
        "rgb",
        "brightness",
        "white"
    ],
    "white_scale": 100,
    "white_command_topic": "~color/0/status",
    "red_template": "{{ value_json.red }}",
    "green_template": "{{ value_json.green }}",
    "blue_template": "{{ value_json.blue }}",
    "white_value_template": "{{ value_json.white }}",
    "brightness": true,
    "brightness_scale": 100,
    "brightness_command_topic": "~color/0/set",
    "brightness_state_topic": "~color/0/status",
    "brightness_template": "{{ value_json.gain }}",
    "brightness_value_template": "{{ value_json.gain }}",
    "brightness_command_template": "{\"gain\": {{ value }} }",
    "rgb_state_topic": "~color/0/status",
    "rgb_command_topic": "~color/0/set",
    "rgb_command_template": "{{ {'mode': 'color', 'red': red, 'green': green, 'blue': blue}|to_json }}",
    "rgb_value_template": "{{ value_json.red }},{{ value_json.green }},{{ value_json.blue }}",
    "rgbw_state_topic": "~color/0/status",
    "rgbw_command_topic": "~color/0/set",
    "rgbw_command_template": "{{ {'mode': 'color', 'red': red, 'green': green, 'blue': blue, 'white': white, 'gain': brightness}|to_json }}",
[...]

PR #5945, opened 2023-12-27
Added Shelly RGB device support to AutoDiscovery #5896

pagocs:
I added Shelly RGB devices support to auto discovery modul.

### DIFF
--- a/hardware/MQTTAutoDiscover.cpp
+++ b/hardware/MQTTAutoDiscover.cpp
@@ -1118,6 +1118,7 @@ void MQTTAutoDiscover::on_auto_discovery_message(const struct mosquitto_message*
 			{
 				pSensor->supported_color_modes[itt.asString()] = 1;
 			}
+
 			if (pSensor->supported_color_modes.find("brightness") != pSensor->supported_color_modes.end())
 			{
 				pSensor->bBrightness = true;
@@ -1125,6 +1126,16 @@ void MQTTAutoDiscover::on_auto_discovery_message(const struct mosquitto_message*
 				if (pSensor->supported_color_modes.size() == 1)
 				{
 					//we only support brightness, so it is a normal dimmer and does not support setting a color
+					pSensor->bColor_mode = false;
+					pSensor->supported_color_modes.clear();
+				}
+			}
+
+			if (pSensor->supported_color_modes.find("onoff") != pSensor->supported_color_modes.end())
+			{
+				if (pSensor->supported_color_modes.size() == 1)
+				{
+					//we only support onoff, so it is a normal switch and does not support setting a color
 					pSensor->bColor_mode = false;
 					pSensor->supported_color_modes.clear();
 				}


gizmocuz:
Thank you for this PR!

Firstly, are you using shellyteacher4domo ? Could you try with this first?

https://github.com/enesbcs/shellyteacher4domo

Could you post the MQTT configuration topic + payload as well as a few status examples and also some example where you set the lights colour ?

The reason for this is that the MQTT-AD code is pretty clear and I'm not a fan of hacking this just for one company's strange implementation.
We support lot's of RGB bulbs (zigbee/zwave) without this funky need

pagocs:
> Firstly, are you using shellyteacher4domo ? Could you try with this first?
> https://github.com/enesbcs/shellyteacher4domo

Yes, I am using that and I already working with the author, he was raised the original issue here:
[https://github.com/domoticz/domoticz/issues/5896](url)

> Could you post the MQTT configuration topic + payload as well as a few status examples and also some example where > you set the lights colour ?

Here is the shellyteacher4domo template line:
`payload={"name": "%shelly_id% Color 0", "cmd_t": "~color/0/set", "payload_on": "{\"turn\":\"on\",\"mode\":\"color\"}","payload_off": "{\"turn\":\"off\",\"mode\":\"color\"}", "enabled_by_default": true, "stat_t": "~color/0/status", "value_template": "{{value_json.ison}}","brightness_command_topic": "~color/0/set","brightness_state_topic": "~color/0/status","brightness": true,"color_mode": true,"supported_color_modes": ["onoff", "brightness" , "color_temp", "rgbcct"],"brightness_scale": 100, "rgb_state_topic": "~color/0/status","rgb_command_topic": "~color/0/set","rgbw_cmd_jtpl": "{ \"r\": \"red\",\"g\": \"green\",\"b\": \"blue\", \"w\": \"white\" ,\"state\":\"turn\",\"brightness\": {\"color\": \"gain\",\"white\": \"brightness\"},\"mode\": {\"3\": \"color\",\"2\": \"white\"},\"color_temp\":\"temp\"}","avty_t": "~online","pl_avail": "true","pl_not_avail": "false","uniq_id": "%shelly_id%-rgbw-0","qos": 0,"min_mirs": 6500,"max_mirs": 2700,"~": "shellies/%shelly_id%/"}`

The main difference is this part:

`"rgbw_cmd_jtpl": "{ \"r\": \"red\",\"g\": \"green\",\"b\": \"blue\", \"w\": \"white\" ,\"state\":\"turn\",\"brightness\": {\"color\": \"gain\",\"white\": \"brightness\"},\"mode\": {\"3\": \"color\",\"2\": \"white\"},\"color_temp\":\"temp\"}"`

which is describe all the special settings mapping: mode values, color names, brightness name (whicah is different in color os white mode (if the bulb is regbw)) and so... 

MQTT payload samples:
- Set Level Color mode:
`{"gain":65,"mode":"color","turn":"on"}`
- Set Level White mode:
`{"brightness":65,"mode":"white","temp":3862,"turn":"on"}`
- Set Color:
`{"blue":69,"gain":81,"green":46,"mode":"color","red":255,"turn":"on","white":0}`
- Command off:
`{"turn":"off","mode":"color"}`

> The reason for this is that the MQTT-AD code is pretty clear and I'm not a fan of hacking this just for one company's 
> strange implementation.
> We support lot's of RGB bulbs (zigbee/zwave) without this funky need

I agree with you, I do not support for the hacking also, that is why I tried to add this support based on a JSON descriptor.

gizmocuz:
I'm sorry but you might want to ask the developers from Shelly to have a look into this, as this is a large hack to get it working.

'rgbw_cmd_jtpl' is also not defined by MQTT-AD

I would vote for shelly to add some compatibility for controlling the lights in their next firmware like other manufacturers are doing.

Another option could be that shellyteacher4domo will run as a service and listens for set commands and translate them to the correct output commands

pagocs:
> 'rgbw_cmd_jtpl' is also not defined by MQTT-AD

This what my  PR was added... 
Can you explain what means hack for you in this case?

gizmocuz:
Sure

With a 'hack' I mean adding more code to a class that makes it unreadable (and not clean) just to support one manufacturers MQTT implementation.

 Remember, this class is NOT about Shelly... it is about all devices that implements the basic auto discovery protocol.

So, for shelly I see 4 solutions:
- Derive a new Shelly class based on the MQTT class to support all Shelly devices
- Ask Shelly to implement using a more basic MQTT protocol for setting/retrieving device states (I volunteer to assist if needed!)
- Extend 'shellyteacher4domo', so it will run as a background service to help translating some MQTT commands for the few Shelly devices where it's needed
- Have a look at a Zigbee device (I know, not really a solution but....)

pagocs:
Sorry, but I cannot agree with you. I think the autodisco is just a platform which can make more easy to adopt new hardwares.
Currently the domoticz autodisco implementation is not support the special template phrases so impossible to add any hw support which have special modes or just working differently than others. 

So I guess that is why there are some special code  in the implementation already, for example:

```
this
if (root["value"].isObject() && (!root["value"]["red"].empty() && root["value"]["r"].empty()))
			{
				// Color values are defined in "value" object instead of "color" as expected by domoticz (e.g. Fibaro FGRGBW)
				root["color"] = root["value"];
				root.removeMember("value");

or this:
if (root["state"].empty() && root["color"].isObject())
			{
				// The on/off state is omitted in the message, so guess it from the color components (e.g. Fibaro FGRGBW)
				int r = root["color"]["r"].asInt();

or this:
	// Check if the rgb_command_template suggests to use "red", "green"... instead of the default "r", "g"... (e.g. Fibaro FGRGBW)
				if (!pSensor->rgb_command_template.empty() && pSensor->rgb_command_template.find("red: red") != std::string::npos)
				{
					// For the Fibaro FGRGBW dimmer:
					//  "rgb_command_template": "{{ {'red': red, 'green': green, 'blue': blue}|to_json }}",  
					//	"rgb_value_template": "{{ value_json.value.red }},{{ value_json.value.green }},{{ value_json.value.blue }}",
					//  -> variables are red, green and blue, but white is missing entirely, so the template can't be used as is.
					Json::Value colorDef;
					colorDef["red"] = root["color"]["r"];

and so...
```

But my approach was different, because the modifications are based on a json template. If you think the modification is too much I will check how I can reduce it (I have some idea already). But I do not want to put more effort into it, if could not be approved...

About your other suggestion, I am afraid to achieve the protocol change at Shelly could not be possible  based on Fibaro example. All other alternatives is not a good idea if you look the pro and cons. Do not forget all the other functionality is working perfectly (except some little thingy, but I already have some other PR in the queue :) ) just the nice color and temperature control are missing...

gizmocuz:
Well, these are actually zwavejs-ui specific.... Implemented by someone with a Fibaro RGBW
I am thinking (loudly) how we can keep this clean as in the feature I can expect other implementations for other hardware

How about in the MQTTAutoDiscover.h we add a section for 'Shelly':
**ShellyHandleDiscovery(.....)**
this will read all shelly specific objects, so we also need to some extra variables for the _tMQTTASensor so in MQTTAutoDiscover.h we make this structure look something like:

`struct _tMQTTASensor : ShellySensor`

**ShellySendSwitchCommand**

This is called inside **SendSwitchCommand**, returns true if handled (for RGBW for example), if false, we continue to use the code in **SendSwitchCommand**

Then we make two new files

MQTTAutoDiscover_Shelly.h and cpp

This way the base class stays clean, and any special code for shelly can be written in the shelly class?

kiddigital:
Maybe we can extend the base hardware class to allow for custom MQTT discovery handling? Or have a base MQTT discovery 'hardware' class. That way we can add custom handling in hardware specific modules and keep the generic part clean.

gizmocuz:
@kiddigital , that is already done in the MQTT base class. As Shelly depends on the MQTT-AD it makes more sense to derive the new class from that one

pagocs:
OK, I tthink about it and I will get back in the new year.
I am busy just now...

On the other hand, I have some other suggestion/modification for the autodisco. It should create PR those also or can we discuss somewhere in advance?

Last but not least Happy New Year!

gizmocuz:
Sure, please join our slack channel, you can send me a PM!

Happy new Year!!

gizmocuz:
Sorry forgot the invite link 🤭

pagocs:
Sorry, I was busy in the past days. So get back to the origin al topic...

Currently I did not find any good solution because if we create a **ShellySendSwitchCommand**  function (if I understand corretcly what you was proposed) it should be duplicate nearly the whole code because all the different functions (ON/OFF, SET_LEVEL, SET_COLOR) are affected. I am afraid this is not a graceful approach and in this case any further maintenance will be more complicated also.

So I would like to propose the original approach but I will reduce the code affectedness as much as possible if this could be accepted.

gizmocuz:
I think it depends on where you would call this detect shelly and call ShellySendSwitchCommand
The logic should be 99% the same indeed, it's just a switch

pagocs:
I did some research about Homeassitant auto discovery which bring me an another idea. It could be acceptable for you if I add some more parsing to the template handling? Because as I think using  some existing autodisco config tags ( for example:   'brightness_command_template', 'brightness_value_template' , 'rgbw_command_template', 'rgbw_value_template',...) could be cover the Shelly json format issue. But it should need some more investigations if this could be a direction...

gizmocuz:
Could you give some examples?

Maybe we should use a library parsing these templates?

pagocs:
Sorry I was busy int the past few days. 
So here is a quick sample bout my idea:
For the Shelly RGB devices could be define the following auto disco descriptor which is could be same what shellyteacher already uses except the "rgbw_command_template" and "command_template" fields.  These fields could be processed in the template parser and there the #values.... values could be resolved (the naming just an example so it could be use anything else of course). The other issue is the mode change what could be handle if the auto disco use different field for template parsing like the command_template if the mode is "ColorModeTemp" and the other one as  "rgbw_command_template" if the mode is "color.mode == ColorModeRGB || color.mode == ColorModeCustom", but just if both ( "rgbw_command_template" and command_template" ) are defined on a device.

```{"name": "%shelly_id% Color 0", 
"cmd_t": "~color/0/set", 
"payload_on": "{\"turn\":\"on\"}",
"payload_off": "{\"turn\":\"off\"}", `

"rgbw_command_template": "{
\"turn\":\"#value.state\"
\"mode\":\"color\"
\"gain\":\"#value.brightness\"
\"red\":\"#value.color.red\"
\"green\":\"#value.color.green\"
\"blue\":\"#value.color.blue\"
\"white\":\"#value.color.cw\"
}", 

"command_template" : "{
\"turn\":\"#value.state\"
\"mode\":\"white\"
\"brightness\":\"#value.brightness\"
\"temp\":\"#value.color_temp\"
}",

"enabled_by_default": true,
"stat_t": "~color/0/status", 
"value_template": "{{value_json.ison}}",
"brightness_command_topic": "~color/0/set",
"brightness_state_topic": "~color/0/status",
"brightness": true,
"color_mode": true,"supported_color_modes": ["onoff", "brightness" , "color_temp", "rgbcct"],
"brightness_scale": 100, 
"rgb_state_topic": "~color/0/status",
"rgb_command_topic": "~color/0/set",
"avty_t": "~online","pl_avail": "true",
"pl_not_avail": "false",
"uniq_id": "%shelly_id%-rgbw-0",
"qos": 0,
"min_mirs": 6500,
"max_mirs": 2700,
"~": "shellies/%shelly_id%/",
"device": {"identifiers": ["%shelly_mac%"],"manufacturer": "Shelly","model": "%shelly_model%","name": "%shelly_id%"}
} 
```

PR #5962, opened 2024-01-05
Re-implement OpenZWave

gizmocuz:


dwmw2:
Oooh, shiny. Why did you do this? Not that I'm complaining; now I can upgrade my Domoticz again!

I did a bunch of work on trying to get the huge node.js thing working, and even bought new hardware (because node.js doesn't work on MIPS any more) but never got round to deploying it all. So I'm pleased to see OpenZWave support back.

gizmocuz:
Well 7 months ago 🤣😂

dwmw2:
Well yeah, but I stopped even looking because I knew I had to move to a non-MIPS box with at least an extra 128MiB of RAM for a huge node.js runtime, before I could upgrade :)

gizmocuz:
It's called an oDroid/Raspberry Pi or you can install it as docker on a larger machine 😂

dwmw2:
Sure, but all my network, phones and HA run off a single 12v battery, and I wasn't keen on the extra power draw just for additional hardware required to host a JavaScript monstrosity :)

PR #5969, opened 2024-01-07
Optionally use libuuid to generate UUID's

arjendekorte:
There isn't a whole lot of randomness in the rand() function (usually 31 bits or less) and definitly not enough to provide for 120+ bits of randomness in GenerateUUID(). The libuuid library does provide enough randomness and may already be installed on the system. In these cases, it may be beneficial to use that instead.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,6 +30,7 @@ option(USE_PYTHON "Use Python for Plugins and Event-Scripts" YES)
 option(INCLUDE_LINUX_I2C "Include I2C support" YES)
 option(INCLUDE_SPI "Include SPI support" YES)
 option(WITH_LIBUSB "Enable libusb support" YES)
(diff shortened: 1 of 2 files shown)


gizmocuz:
Thank you for the PR! We have a function 'GenerateRandomNumber' which is quite OK.

And for the UUID we now use other methods then rand/srand from the random here:

**std::random_device, std::mt19937,std::uniform_int_distribution**

Which are already much better.

Introducing a new library is not needed and is another dependency we don't want.

Besided that, for the functions we request a unique id it is more then random enough ;)

arjendekorte:
The GenerateRandomNumber is a fine example of how to degrade a PRNG to make rand() a lot worse than it already is. One really shouldn't reseed a PRNG more than once. Doing it more often from sources with *very* low entropy like a counter or the clock time, will actually make the output fairly predictable between instances.

PR #5970, opened 2024-01-08
Add Stale Workflow

waltervl:
Checks issues with a PendingFeedback Label and add a stale label when inactive for more then 14 days and close them after 7 days stale. PR's could also be checked but should be ignored now.

Currently in debug only so we can check the impact. Replaces the NoResponse action https://github.com/domoticz/domoticz/blob/development/.github/no-response.yml


gizmocuz:
Thanks, great idea!

Issue #5795, opened 2023-08-23
1wire - disconnected sensor still being reported as a live


blmpl:
Hello,

Regardless of "Error: 1-wire: 1-Wire system : error reading value from /sys/bus/w1/devices/28-xxxxx" error, the last correct reading is constantly send back to the core  which makes you thing that the sensor is still working (see the 
![Screenshot 2023-08-23 at 15 47 56](https://github.com/domoticz/domoticz/assets/10145557/672b6ef6-3d12-4697-8c3c-4208c68baa7e)
graph)

This is happening on the project build from the latest develop branch (last commit 30041b1ee0aff12d1731c5d8fc643173917e9963) using kernel 1wire support

gizmocuz:
That is true. We take the last value received and use this in the short log

blmpl:
I'm not really concerted about the graph, but the last value is send over and over to the core and reported by mqtt. Also sensor is NOT marked as "HaveTimeout" on api and doesn't change its color to red in the gui.

Maybe not a big issue... unless this is a boiler temperature read by PLC (mqtt) that starts 4.5kW 3phase water heater based on it.

my findings are that in 1WireByKernel.cpp => ReadStates, should have some ``device->m_Temperature = -1000.0F;`` in OneWireReadErrorException

gizmocuz:
If you have time and could make a patch/PR that would be great!

blmpl:
can you take a look? it seams to be working

gizmocuz:
Are you compiling Domoticz yourself? How about changing the function 

**C1WireByKernel::ThreadReadRawDataHighPrecisionDigitalThermometer**

At the bottom, instead of throwing an exceptionreturn -1000.0F?

Another option, add a 'valid' boolean to the device object, and at the start of the reading set this to 'false'.
When the reading is correct, set it to true
(Do this everywhere, not only for the kernel method)

In all report methods, check if the device is valid, if not, don't report

I think that is a better method?

gizmocuz:
Thank you for your PR. I had a small comment about checking if a device needs to be created.
This is not needed at this is run at startup and you can be very sure this is not created yet 😊
At the moment you fixed it for 1wire-kernel, if you use OWFS this needs to be done as well?

blmpl:
> Are you compiling Domoticz yourself? 

Yes :)

>  How about changing the function
> **C1WireByKernel::ThreadReadRawDataHighPrecisionDigitalThermometer**
> 
[...]

PR #5988, opened 2024-01-15
reset temperature on OneWireReadErrorException, don't create device m…

blmpl:
this is in reference to https://github.com/domoticz/domoticz/issues/5795

### DIFF
--- a/hardware/1Wire/1WireByKernel.cpp
+++ b/hardware/1Wire/1WireByKernel.cpp
@@ -94,13 +94,13 @@ void C1WireByKernel::ReadStates()
 {
 	for (auto & it : m_Devices)
 	{
+		// Priority to changes asked by Domoticz
+		ThreadProcessPendingChanges();
+
+		DeviceState* device = it.second;
 		// Next read one device state
 		try
 		{
-			// Priority to changes asked by Domoticz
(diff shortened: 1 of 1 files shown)


hardware/1Wire/1WireByKernel.cpp
```diff
@@ -241,8 +244,10 @@ void C1WireByKernel::ThreadBuildDevicesList()
 						case _8_channel_addressable_switch:
 						case programmable_resolution_digital_thermometer:
 						case Temperature_memory:
-							m_Devices[device.devid] = new DeviceState(device);
-							m_p1WireBase->Log(LOG_STATUS, "1Wire: Added Device: %s", sLine.c_str());
+							if (m_Devices.count(device.devid) == 0) {
```
gizmocuz: This is not needed as it is created only at startup
Instead of 'count' I would also advice to use if **(m_Devices.find(device.devid) == m_Devices.end())**
But it's not needed

blmpl: yes, you are right - it's a leftover from my attempt to call this more frequent. 
I will remove that, and thanks for the tip!

gizmocuz:
Thank you for your PR and time!

PR #6004, opened 2024-01-26
cWebm: Fixing high CPU load caused by time jumps.

Morolem:
This fix addresses the issue of high CPU load when there are time jumps. If, at the time of launch, the system time was set to 01.01.1970 and was subsequently changed to the current time, creating a waiting point using 'm_session_clean_timer.expires_at(m_session_clean_timer.expires_at() + boost::posix_time::minutes(15));' would result in returning 1970 plus 15 minutes on the first iteration, another +15 on the second, and so on. In this case, the call to 'cWebem::CleanSessions()' will occur without delays until the value returned from 'm_session_clean_timer.expires_at()' equals the current time.

### DIFF
--- a/webserver/cWebem.cpp
+++ b/webserver/cWebem.cpp
@@ -974,7 +974,7 @@ namespace http {
 				mySessionStore->CleanSessions();
 			}
 			// Schedule next cleanup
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thank you for the PR!

Issue #6005, opened 2024-01-26
Evohome keeps logging "fetch data from server"


fwieringen:
Hi,

I've been running Evohome for a while now with Domoticz. The device is set to log only Errors. But still I see a lot of log messages stating "(Evohome) fetch data from server".

![image](https://github.com/domoticz/domoticz/assets/4232879/5cca4f30-4457-406d-ad3e-ecac5b195418)

Any idea how to solve this?

Looked it up in the code, but looks like a normal log line:

`_log.Log(LOG_NORM, "(%s) fetch data from server", m_Name.c_str());`

gizmocuz:
Seems you have a problem with the data from the server.
Please use the forum as there could by other users using evohome.

fwieringen:
I don't understand how this is a problem with the data from the server. When I look at the source, the only place this log entry is created is in hardware/EvohomeWeb.cpp line 276. And it's basically created when two basic checks are performed without errors. 

This appears to be a normal log line, but it ignores the logging setting for Evohome Hardware. In my case this is set to Error only. I cannot explain why this LOG_NORM log entry slips through. 

Data from Evohome server is received and properly reflected in the domoticz dashboard by the way.

fwieringen:
Apparently, all other log messages use LOG_STATUS or LOG_ERROR. I would propose to change this LOG_NORM line to LOG_STATUS as well.

gizmocuz:
PR merged, thank you !

fwieringen:
Reverted the earlier change and reworked all logging in the file. Instead of invoking _log.Log(), I now refer to Log(), Submitted this in pull request #6008

PR #6007, opened 2024-01-29
Updated some EvoHome logging messages to LOG_STATUS

fwieringen:
Most of the log messages are logging to LOG_STATUS or LOG_ERROR, except for two. These logged to LOG_NORM, causing those log messages to popup in the log even when no logging is selected. This update solves issue #6005 .


gizmocuz:
Thank you for this PR!!

PR #6008, opened 2024-01-29
Reworked EvohomeWeb logging to standard

fwieringen:
Removed all the _log.Log() references and created regular Log() calls.


fwieringen:
Tested it with various logging settings and behaviour is now as expected. The three logging settings in the Hardware screen are now properly processed and reflected in the log.

gizmocuz:
Perfect! Thank you for this!!

PR #6010, opened 2024-01-30
Reworking logging of remaining Evohome files

fwieringen:
All references to _log.Log() removed and replaced by Log(). Redundant logging text ("Evohome: ") removed.


gizmocuz:
Thank you again!!

PR #6012, opened 2024-01-31
Show column in counter day graph between hours

waltervl:
This for better understanding the graph data. The counter column value eg Energy kWh is from data between hour x and hour y.
PS I do not know where commit change in the last line is coming from.....

Example before the change
Before_between
After change
after_between


gizmocuz:
Thank you!!

Issue #6014, opened 2024-02-03
GPIO pinnumber for raspberry Pi 5


ericstaal:
The raspbery Pi 5 has a complete new pin nummering.

GPIO 26 (pin 37) on the Pi4 is mapped to pin nummer GPIO 425 on the pi5
(See file /sys/kernel/debug/gpio)

The GPIO hardware limits thist number:
```
#define GPIO_PIN_MIN		0
#define GPIO_PIN_MAX		31
```

Max on the Pi5 is 511.
Can this number be increased to 511?

gizmocuz:
I'm afraid that won't work.
The reason is that the current code is using this 'number' as unit.

```
			result = m_sql.safe_query("SELECT nValue FROM DeviceStatus WHERE (HardwareID==%d) AND (Unit==%d)",
				m_HwdID, gpio_pin);
```

The unit can range from 1 to 255, not higher

What you could try is to try to find only the PIN items, and use the number beging the PIN as unit?

 gpio-460 (PIN3                )
 gpio-461 (PIN5                )
 gpio-462 (PIN7                )
 gpio-463 (PIN29               )
 gpio-464 (PIN31               )
 gpio-465 (PIN26               )
 gpio-466 (PIN24               )
 gpio-467 (PIN21               )
 gpio-468 (PIN19               )
 gpio-469 (PIN23               )
 gpio-470 (PIN32               )
 gpio-471 (PIN33               )
 gpio-472 (PIN8                )
 gpio-473 (PIN10               )
 gpio-474 (PIN36               )
 gpio-475 (PIN11               )
 gpio-476 (PIN12               )

But they are named differently on a RPI4 (and probably below)

 gpio-16  (GPIO16              )
 gpio-17  (GPIO17              )
 gpio-18  (GPIO18              )
 gpio-19  (GPIO19              )
 gpio-20  (GPIO20              )

But what hardware have you connected to the GPIO?
Could you use SysFS ? (https://www.domoticz.com/forum/viewtopic.php?t=22501)

Or if they are switches/buttons/lights, move to zigbee2mqtt?

ericstaal:
PIN number can still be used, bit still it will prevent access all available IO. RPI5 exposed quite a lot (which can be very nice ;) )

I connected a doorbell (12 vac) to some gpio.
sysfs is deprecated in bookworm: https://forums.raspberrypi.com/viewtopic.php?t=343514 It is migrated to libgpiod
[...]

PR #6015, opened 2024-02-04
Increase gpio numbering for bookworm

ericstaal:
Solves PR #6014

GPIO numbers are changed since bookworm, therefore the range must be extended to make it work with the existing code.


gizmocuz:
This won't work as currently the number is used as device 'unit', which has a range from 1-255
Thanks for the PR!

PR #5815, opened 2023-09-10
Netatmo OAUTH2 change to auth_code grant (from password)

kiddigital:
Small PR addressing issue #5766 

Not the most user friendly (_option 1_ as discussed in the issue) but not to difficult to set up. Should be done only once as long as domoticz is able to keep refreshing the tokens.

### step 1:
Setup an _App_ under __MyApps__ in the Netatmo [developers portal](https://dev.netatmo.com)
- The _Redirect_URI_ has to contain `http://localhost/netatmo`

Once configured correctly, the _Client_ID_ and _Client_Secret_ will be generated and shown.

### step 2:
In the domoticz hardware setup:
- Create/modify the Netatmo hardware
- the _username_ should contain `<client_id>:<client_secret>`
- while the password should contain the retrieved `<code>`

To retrieve the _code_ (a.k.a. authorisation_code), use the browser and type the following URL. Make sure to fill in the CLIENTID with your clientid!

```
https://api.netatmo.com/oauth2/authorize?client_id=<YOURCLIENTID>&redirect_uri=http%3A%2F%2Flocalhost%2Fnetatmo&state=teststate&scope=read_station%20read_thermostat%20write_thermostat%20read_homecoach%20read_smokedetector%20read_presence%20read_camera
```
This URL brings you to the My Netatmo page where you have to identify yourself and are asked for permission to grant access to certain devices and types (based on the _scope_ provided).

After confirmation, your browser will try to redirect you to a page it cannot find so you will get an error. But the URL of that page is visible in the browser and looks like `http://localhost/netatmo?state=teststate&code=<SOMECODE>`.

You need to pass this _SOMECODE_ as the _password_ in the domoticz hardware screen.

__NOTE__:
- At the moment, the _scopes_ are fixed to `read_station read_thermostat write_thermostat read_homecoach read_smokedetector read_presence read_camera`
- The _\<SOMECODE\>_ you retrieved is only valid for a few minutes, so if you take too long and domoticz says (check the logs) it cannot login, retrieve a new code and be a little quicker :)
- Proper documentation (on the [Wiki](https://www.domoticz.com/wiki/Netatmo)) has to be created/updated

PR #6016, opened 2024-02-04
 Remaster API Netatmo

JurgenLB:
With this PR I’m willing to share my work so far.
Some off the issues addressed;
- Homecoach and weather station at the same account
- battery  level
- Thermostat live data instead off historical data
- door sensors

TO DO
the Send to hardware for thermostat/camera’s
Events from smoke/co2/camera’s/door-sensors in Domoticz

with this PR my system is from 16 used devices to  +50 with the same account

kind regards

### DIFF
--- a/History.txt
+++ b/History.txt
@@ -1,4 +1,5 @@
 Version 2024.x
+- Implemented: Support for Honeywell Series 5/PIR
 - Changed: Google Firebase Cloud Messaging (FCM) alternative way to pass a Domoticz device ID through the extra field ('|Device=<devidx>')
 - Changed: dzVents, possible to pass a Domoticz device ID in notify() through the extra parameter
 - Changed: Energy Dashboard now also available for non-admin users
--- a/hardware/RFXBase.cpp
+++ b/hardware/RFXBase.cpp
@@ -164,6 +164,8 @@ bool CRFXBase::CheckValidRFXData(const uint8_t *pData)
 		return (pLen == 0x0C);
 	case pTypeDDxxxx:
 		return (pLen == 0x0C);
+	case pTypeHoneywell_AL:
+		return (pLen == 0x0A);
 	default:
 		return false;//unknown Type
 	}
--- a/main/Helper.cpp
+++ b/main/Helper.cpp
@@ -1058,6 +1058,7 @@ bool IsLightOrSwitch(const int dType, const int dSubType)
 	case pTypeFS20:
 	case pTypeHunter:
 	case pTypeDDxxxx:
+	case pTypeHoneywell_AL:
 		bIsLightSwitch = true;
 		break;
 	case pTypeRadiator1:
--- a/main/RFXNames.cpp
+++ b/main/RFXNames.cpp
@@ -521,6 +521,7 @@ const char* RFX_Type_Desc(const unsigned char i, const unsigned char snum)
 		{ pTypeSOLAR, "Solar", "solar" },
 		{ pTypeHunter, "Hunter", "Hunter" },
 		{ pTypeDDxxxx, "Blinds", "blinds" },
+		{ pTypeHoneywell_AL, "Honeywell", "doorbell" },
 		{ 0, nullptr, nullptr },
 	};
 	if (snum == 1)
@@ -945,6 +946,10 @@ const char* RFX_Type_SubType_Desc(const unsigned char dType, const unsigned char
 		{ pTypeGeneralSwitch, sSwitchTypeFaber, "Faber" },
 		{ pTypeGeneralSwitch, sSwitchTypeDrayton, "Drayton" },
 		{ pTypeGeneralSwitch, sSwitchTypeV2Phoenix, "V2Phoenix" },
+
+		{ pTypeHoneywell_AL, sTypeSeries5, "Honeywell Series 5" },
+		{ pTypeHoneywell_AL, sTypePIR, "Honeywell PIR" },
+
 		{ 0, 0, nullptr },
 	};
 	return findTableID1ID2(Table, dType, sType);
@@ -2254,6 +2259,9 @@ void GetLightStatus(
 			lstatus = "unknown/unsupported!";
 			break;
 		}
+		break;
+	case pTypeHoneywell_AL:
+		lstatus = "On";
 		break;
 	}
 
--- a/main/SQLHelper.cpp
+++ b/main/SQLHelper.cpp
@@ -5477,6 +5477,7 @@ uint64_t CSQLHelper::UpdateValueInt(
 	case pTypeRadiator1:
 	case pTypeHunter:
 	case pTypeDDxxxx:
+	case pTypeHoneywell_AL:
 		if ((devType == pTypeRadiator1) && (subType != sTypeSmartwaresSwitchRadiator))
 			break;
 		m_LastSwitchID = ID;
@@ -9033,7 +9034,7 @@ void CSQLHelper::CheckDeviceTimeout()
 	result = safe_query(
 		"SELECT ID, Name, LastUpdate FROM DeviceStatus WHERE (Used!=0 AND LastUpdate<='%04d-%02d-%02d %02d:%02d:%02d' "
 		"AND Type!=%d AND Type!=%d AND Type!=%d AND Type!=%d AND Type!=%d AND Type!=%d AND Type!=%d AND Type!=%d AND Type!=%d "
-		"AND Type!=%d AND Type!=%d AND Type!=%d AND Type!=%d AND Type!=%d AND Type!=%d AND Type!=%d AND Type!=%d AND Type!=%d AND Type!=%d AND Type!=%d AND Type!=%d AND Type!=%d AND Type!=%d) "
+		"AND Type!=%d AND Type!=%d AND Type!=%d AND Type!=%d AND Type!=%d AND Type!=%d AND Type!=%d AND Type!=%d AND Type!=%d AND Type!=%d AND Type!=%d AND Type!=%d AND Type!=%d AND Type!=%d AND Type!=%d) "
 		"ORDER BY Name COLLATE NOCASE ASC",
 		ltime.tm_year + 1900, ltime.tm_mon + 1, ltime.tm_mday, ltime.tm_hour, ltime.tm_min, ltime.tm_sec,
 		pTypeLighting1,
@@ -9058,7 +9059,8 @@ void CSQLHelper::CheckDeviceTimeout()
 		pTypeHomeConfort,
 		pTypeFS20,
 		pTypeHunter,
-		pTypeDDxxxx
+		pTypeDDxxxx,
+		pTypeHoneywell_AL
 	);
 	if (result.empty())
 		return;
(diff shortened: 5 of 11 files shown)


gizmocuz:
Thank you for this PR and time! Let us know when it's ready to merge!

hardware/Netatmo.h
```diff
@@ -2,29 +2,49 @@
 
 #include "DomoticzHardware.h"
 #include "../main/BaroForecastCalculator.h"
+#include <mutex>
```
gizmocuz: This should not be needed at it is already included in the precompiled header file (stdafx.h)

JurgenLB:
I changed the headers like you suggested.

The Thermostat gives a Log-message; 
`Status: Warning: Notification NOT handled (Hardware: 2 - Netatmo, ID: 3557328368, Unit: 0, Type: 40 - Thermostat 1, SubType: 0 - Digimax), please report on GitHub!`
so there is a problem with my code.

For the rest I think it is ready.
With my setup I have all of my devices
The "Events" and the "send to hardware" still need some work.

Kind regards

JurgenLB:
I now have the "Last Events" in the code as 'Text sensor'.
I found my fault with the 'Unknown? Hardware' and I'm correcting this

waltervl:
Is the change of Netatmo API Auth on 29 may taken into consideration in this change?

JurgenLB:
I believe the change off 29 may is already done by #5815 
Currently I'm working on the setpoint/mode off the thermostat for the other devices I think all the measurements are in this PR.
Kind regards

JurgenLB:
are you willing to accept this addition to "mineworker.cpp"

`std::string id;

int idx;

std::stringstream ss;

id = sd[1];

ss << std::hex << sd[7];

ss >> idx;`

because now it is returning a int of "DeviceID" and not idx as the naming suggest.
with this addition my code for changing the setpoint works on my local setup.

`pGateway->SetSetpoint(id, TempValue);`

Kind regards

gizmocuz:
Could you undo all formatting and rearranging in your PR.
This way it is hard to read what has changed, and also the class constructor/deconstructor declararion needs to stay at the top (as it is with all classes)

JurgenLB:
Only the selector switch is giving me trouble.
Are there more formatting changes you wan't to be addressed ?
Kind regards

gizmocuz:
I would like you to undo the formatting changes

image

Please start fresh, and ONLY add your changing, and do not perform any formatting or shifting at all

JurgenLB:
it was because off error message
`/hardware/Netatmo.h:20:36: error: ‘_eNetatmoType’ does not name a type; did you mean ‘_eMeterType’?`
`void Get_Respons_API(const _eNetatmoType& NType, std::string& sResult, std::string& home_id , bool& bRet, Json::Value& root );`

this is the reason that I put the public declaration after the private


kind regards

gizmocuz:
Okey, thank you for this!
Please let me know when you would like your PR to be merged!

JurgenLB:
Do you have any pointers or advice on the use off   selector switch for schedule ?
That part is not working for the moment, when that is working I believe we have the same behavior as from the old implementation.

Kind regards

gizmocuz:
Hi Jurgen, I am not quite sure what you mean with 'selector switch for schedule'
You mean set a timer on the selector switch that will set a specific state? This is already working
Or you mean when a user makes a selection, how you can handle this in your hardware class?

JurgenLB:
I'm sorry for all the mix-up and trouble.

F858EF48-D5C5-4A92-A7FC-35BA99BB09EE_4_5005_c

With the old integration there is a selector switch for every thermostat but not a valve, and if you change 1 the others are also changing.
With my new integration I have them separated but the status information for the switch is for me not possible to change  with `UpdateValue(...`
with `SendSelectorSwitch(...` I think it will work but this must be verified.
Apparently there is some difference between a local running instance and the checks off GitHub 

Kind regards

JurgenLB:
Do you want me to remove the sTypeTextStatus with " - MAC-adres" ?
I believe it Is helpful to identify the devices.

Kind regards

gizmocuz:
It's totally up to you. But I see there are some merge issues below

JurgenLB:
I ask because the merge conflict is that the sTypeTextStatus is marked as "development"
For me only the Floodlight from the outdoor-camera is not working.
But if you want to close this its Ok for me.

gizmocuz:
It's not me having a issue, but github with merging.
Best to fix it.

One thing I am not to sure about, is the change you made in MainWorker::SetSetPointInt(

I verified this and it works like this, it is a UTF16 compare not a unicode... did you set a breakpoint here and tested this?
Else revert it

JurgenLB:
I put some debug-lines in MainWorker:: SetSetPointInt( to find a solution for the function SetSetPoint
I removed the debug-lines.
the code necessary for a correct `int idx` I leave it there.
the old code had a `std::hex sd[1]` but this dit not result in a correct idx, the correct idx is in sd[7]
because off the 'int' constructor I made the code like this.
with the SetProgramState there is the use of 'const int id' to identify the device

Kind regards

Jurgen

JurgenLB:
most of the warning from appVeyor are from debug lines.
These can be removed after everything is working.
The callback from the selector-switch are not complete correct but for the rest all is working with my devices.
As mentioned in #6036  there is a change necessary for the login / scopes

Kind regards

gizmocuz:
Could you undo the changes in MainWorker::SetSetPointInt

This should not be necessary as you have 32 bits for an ID and also an 8 bit for Unit. 
That should be more than enough to identify the setpoint?

Also, the changes for the degrees are not necessary as they are currently working. Changing this could(is) give issues

JurgenLB:
with the return to original mainworker.cpp the function setsetpoint does not work.
The output from the log is;
`Debug: SQLH UpdateValueInt Verwarming - MAC-adres HwID:3  DevID:1906536938 Type:243  sType:19 nValue:48 sValue:04:00:00:1e:84:a8 
Debug: SQLH UpdateValueInt  Verwarming Bat. Lvl HwID:3  DevID:1906536938 Type:243  sType:31 nValue:48 sValue:37 
Debug: SQLH UpdateValueInt  Verwarming - Sig. + Bat. Lvl HwID:3  DevID:1906536938 Type:243  sType:6 nValue:48 sValue:86.6667 
Debug: SQLH UpdateValueInt Verwarming - Bridge HwID:3  DevID:1906536938 Type:243  sType:19 nValue:48 sValue:70:ee:50:1e:89:96 Relay 
Debug: SQLH UpdateValueInt Verwarming - Heating Status HwID:3  DevID:1906536938 Type:242  sType:1 nValue:48 sValue:14 

Debug: SQLH UpdateValueInt Verwarming HwID:3  DevID:1906536938 Type:80  sType:5 nValue:48 sValue:24
Debug: SQLH UpdateValueInt Verwarming - Heating Status HwID:3  DevID:1906536938 Type:244  sType:73 nValue:0 sValue:0 


setsetpoint
Status: User: domoticz initiated a SetPoint command
Debug: IDX 208 , HardwareID 3 - 1906536938
Debug: Netatmo: ID = 106129720
Debug: Netatmo: reverseID = 946426630
Debug: Netatmo: IDX = 0
Debug: Netatmo: NetatmoThermostat = 
Debug: Netatmo: Netatmo Room ID =  in Home: 
Error: Netatmo: NetatmoThermostat : No thermostat found in online devices!
Debug: Netatmo: NetatmoThermostat idx = 0, MAC = , Home = 
Debug: Netatmo: Netatmo RoomID =  / 127 
Debug: SQLH UpdateValueInt Unknown HwID:3  DevID:6536938 Type:242  sType:1 nValue:0 sValue:15.00
`
I was expecting the '1906536938' as reverseID but that is not wat happened.
This was the reason I wanted to use the DeviceRowIdx as identifier.
But it is your project. 

I even put this is my local test setup
`
int id1 = (id >> 24) & 0x000000FF;

int id2 = (id >> 8) & 0x0000FF00;

int id3 = (id << 8) & 0x00FF0000;

int id4 = (id << 24) & 0xFF000000;

Debug(DEBUG_HARDWARE, "ID1 = %d", id1);

Debug(DEBUG_HARDWARE, "ID2 = %d", id2);

Debug(DEBUG_HARDWARE, "ID3 = %d", id3);

Debug(DEBUG_HARDWARE, "ID4 = %d", id4);
`

Debug: Netatmo: ID1 = 6
Debug: Netatmo: ID2 = 21248
Debug: Netatmo: ID3 = 6881280
Debug: Netatmo: ID4 = 939524096

kind regards

gizmocuz:
in

MainWorker::SetSetPointInt

The sd[1] = a string representing the DeviceID. You can put anything there that you want.
To see how to use a string as deviceID and still keep everything working correctly (like the internal event system and notifications) have a look in the MQTTAutoDiscovery.cpp file

But the reason why it goes wrong is that if you use the 'SendSetPointSensor' call (which is highly preferred and good), the first number (ID1) is always 0x00 and the Unit is always 1

This is caused by:

**CDomoticzHardwareBase::SendSetPointSensor**

I will change this to:

void CDomoticzHardwareBase::SendSetPointSensor(const uint8_t ID1, const uint8_t ID2, const uint8_t ID3, const uint8_t ID4, const uint8_t Unit, const float Value, const std::string& defaultname)

gizmocuz:
I have changed the base hardware SendSetpointSensor in beta #16110
Hope you can work with this now?

JurgenLB:
Thank for the quick reply.
I will look in to this.
The MAC-adres is per apparatus witch result in multiple devices in Domoticz

jcbraem:
Hello,
Any news about your remaster coding ?
Thanks in advance.

JurgenLB:
Yes for the moment I'm working with the setpoint of the thermostat.
It is only updating the last room instead off the correct thermostat-room.
And the return function for the selector switch off the thermostat is also not working (On, Away, Frost Guard)

Kind regards

jcbraem:
Thanks for your feedback and your working.

Of course if you have need one test device or other, if I can, I make :)

JurgenLB:
I can now confirm that I have the setpoint working.
For the status selector switch; when I klik on 'Away' there is a error shown in the webpage, but when I look to my thermostat he is in away mode. Will look into this.

@gizmocuz I keep using the full MAC-adres because whitin the same family off devices (for example Temperature) the first segments off the MAC-adres are the same but for other family's their different
image

Kind regards

jcbraem:
For MAC of Netatmo weather station :
image

JurgenLB:
Sorry that it took so long but
I believe I have everything working for the thermostat setpoint and status.
also the conflict I use 'unit = 7' instead off 1
so let me know if you want me to change something 

for the login I believe the hardware-page must be changed to have a changeable scope.
If we use a scope that is different than the devices in the Nettamo-account then the first login is effective
but after some time that App-id is blocked by Netatmo because the scope is bigger than the devices.

Kind regards

JurgenLB:
Do you think it is wise for the 'SendSetPointSensor' and 'SendSelectorSwitch' to also include the battery and Rf- level.
`const int BatteryLevel = 255, const int RssiLevel =12`

because the thermostat is battery operated and then there is a battery an rf- level for all necessary devices.

thermostat

For the text-sensor and the individual batterij level I believe it is not necessary to include the battery and RF- level
But it can be better to do it for all.

kind regards

LoukKamp:
Just tested: The refresh token doesn't expire after 24 hours. Now testing if it will expire within 7 days.

PR #6019, opened 2024-02-05
Add mapping from GPIO pin to UNIT

### LINKED ISSUE #6014: GPIO pinnumber for raspberry Pi 5

ericstaal:
UNIT have the range [0-255] GPIO pins can be in a higher range for a raspberry Pi5 (400-426)

Added a unit in the data so it will become more robust.

Solves: #6014

### DIFF
--- a/hardware/Gpio.cpp
+++ b/hardware/Gpio.cpp
@@ -70,7 +70,7 @@ Connection information:
 #define GPIO_FALLING		2
 #define GPIO_BOTH			3
 #define GPIO_PIN_MIN		0
-#define GPIO_PIN_MAX		512
+#define GPIO_PIN_MAX		31
 #define GPIO_MAX_VALUE_SIZE	16
 #define GPIO_MAX_PATH		64
 #define GPIO_PATH			"/sys/class/gpio/gpio"
--- a/hardware/Gpio.cpp
+++ b/hardware/Gpio.cpp
@@ -70,7 +70,7 @@ Connection information:
 #define GPIO_FALLING		2
 #define GPIO_BOTH			3
 #define GPIO_PIN_MIN		0
-#define GPIO_PIN_MAX		31
+#define GPIO_PIN_MAX		512
 #define GPIO_MAX_VALUE_SIZE	16
 #define GPIO_MAX_PATH		64
 #define GPIO_PATH			"/sys/class/gpio/gpio"
--- a/hardware/SysfsGpio.cpp
+++ b/hardware/SysfsGpio.cpp
@@ -109,7 +109,7 @@
 	24-jun-2017	HvB		Add hardware debounce parameter, range 10..750 milli sec.
 	04-jul-2017	HvB		Poll after an interrupt received to recover missed interrupts
 	06-jul-2017 HvB		Removed log message for interrupt state change, forum request
-	05-feb-2023 Es    Add unit id so GPIO outside [0-255] can also be used 
+	05-feb-2023 Es		Add unit id so GPIO outside [0-255] can also be used 
 */
 
 #include "stdafx.h"
@@ -508,7 +508,7 @@ void CSysfsGpio::FindGpioExports()
 			gi.id3 = 0;
 			gi.id4 = 0;
 			gi.request_update = -1;
-      gi.unit = ConvertGpioPin(gpio_pin);
+			gi.unit = ConvertGpioPin(gpio_pin);
 
 			m_saved_state.push_back(gi);
 			close(fd);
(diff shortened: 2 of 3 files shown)


gizmocuz:
Thanks, but this is not really a good solution. (%255)

As mentioned, can't you map the end of the GPIO string (the number part) when you detect it's a RPI5?
This way it stays compatible with previous versions/users as well

ericstaal:
> Thanks, but this is not really a good solution. (%255)
> 
> As mentioned, can't you map the end of the GPIO string (the number part) when you detect it's a RPI5? This way it stays compatible with previous versions/users as well

Well as you mentioned the unit is between [0-255], if it is currently higher it will no we visible as a device (the current bug).
current valid pin number should remain the same => so they should be from [0-255] which is the case.

Different hardware eg the PI5 has the gpio in the range 400-423. This means module 255 will all have unique numbers in the range of [0-255]. So for PI5 this does work. 
If you migrate to a new hardware the GPIO number are different again and you will get new devices, this is the thing I cannot prevent (its impossible to map [0-1023] to a unique [0-255]).

Create a mapping for PI5 number to prev pi4-1 number will solve this single time (migrating to pi5) but not for other hardware and platform (the ones I don't know about like orange PI, odriod etc). So creating a fixed mapping for pi5 will end up in problems on other single board systems

This solution gives always the same unit number per hardware platform gpio pin. Which I think, the best there is to offer. 

Problem is that PI[1-4] mapped the GPIO to sysfs number [0-32], this is for busters/bullseye/bookworm. From pi5 this is 400-423] in bookworm (since pi5 only support bookworm)

ericstaal:
Tested it on a Pi 4 with bookworm, IO are the same units:
2024-02-06 20:06:58.151  Status: TESTER: gpio26 (unit 26) new state = on
2024-02-06 20:06:58.158  TESTER: Lighting 2 (Input)
2024-02-06 20:07:01.435  Status: TESTER: gpio26 (unit 26) new state = off
2024-02-06 20:07:01.442  TESTER: Lighting 2 (Input)
2024-02-06 20:07:02.387  Status: TESTER: gpio26 (unit 26) new state = on

gizmocuz:
Great, thank you for testing!

akamming:
@gizmocuz  will this PR be merged in the near future? would be great cause now GPIO functionality is not working anymore on pi4 on bookworm, so now i have several workarounds in place. Would be great if i can remove them again..

gizmocuz:
Sure!! Thanks for the PR!

akamming:
Great, tx!

akamming:
@ericstaal  since it's in beta i could now test. I have a PI4.   i use gpio14 and gpio15. So i have the following script to  export the pins before starting domoticz:
```
#!/bin/sh
##export pins for domoticz
# 14 and 15 (see /sys/kernel/debug/gpio to get the numbers)
echo 526 > /sys/class/gpio/export
echo out > /sys/class/gpio/gpio526/direction
echo 1 > /sys/class/gpio/gpio526/active_low
echo 527 > /sys/class/gpio/export
echo out > /sys/class/gpio/gpio527/direction
echo 1 > /sys/class/gpio/gpio527/active_low
```

I would expect  the unit numbers to be 14 and 15 in domoticz, however they are 16 and 17:

image

Can you check?

ericstaal:
Why do you expect 14 and 15?

What was it before the update to the latest beta?

Op do 30 mei 2024 17:36 schreef akamming ***@***.***>:

akamming:
before the update it wasn't working anymore (as of kernel 6.6 also on pi4 the old implementation didn't work anymore).

Before i upgrade to kernel 6.6 on my pi4 the unitnumbering in domoticz was the same as the gpio numbering on the pi.

ericstaal:
So the fixed picks the gpio number and takes module 255 of it since the numbering with latest kernel are different than before. 
This fixed will have the gratest possible way of mapping a gpio pin number to an unique uin8_t [0-255] (which is the range of the unit in domoticz)

526 % 255 is 16 and 527 % 255 is 17, so this is as expected.

It is impossible to create a mapping between all those numberings since each platform has a different numbering. Just update the devices in domoticz and you should be OK. When I created the pinnumber still was the same for Pi 4 with old kernel, the new kernel broke the unit numbering.

akamming:
A clear, tx!

mgugu10:
Quick test on a Pi2 / Bookworm with GPIO 526 out-exported
 The command `cat /sys/kernel/debug/gpio` gives 
`gpio-526  (GPIO14              |sysfs               ) out lo`
 This debug command reports GPIO 14 for gpio-526 while in domoticz Unit 16 is detected
 Confusing …

ericstaal:
If you tead all the comments you should new the implementation uses modulo
255. 526 mod 255 is 16.

If it is too confusing you can always stop using gpio or create an better
implementation yourself ;)

Op ma 3 jun 2024 18:18 schreef mgugu10 ***@***.***>:

Issue #6027, opened 2024-02-13
Domoticz Extended create duplicate entry in DeviceStatrus when updating Type, SubType, SwitchType


pipiche38:
As described in the forum https://www.domoticz.com/forum/viewtopic.php?p=313962#p313962 there is an issue with the Domoticz Extended framework which doesn't exist with the legacy.

If the plugin try to update Type, SwitchType, SubType, it looks like the Extended framework create a new entry in the DeviceStatus table !

CC: @

Here is the plugin to duplicate the problem

[plugin.py.zip](https://github.com/domoticz/domoticz/files/14270410/plugin.py.zip)

cc: @dnpwwo

pipiche38:
What I found is that Legacy use m_sql.UpdateDeviceValue() while Extended use m_sql.UpdateValue()

pipiche38:
From the source code, looks like there is some comment

```
#ifdef ENABLE_PYTHON
		//TODO: Plugins should perhaps be blocked from implicitly adding a device by update? It's most likely a bug due to updating a removed device..
		if (pHardware != nullptr && pHardware->HwdType == HTYPE_PythonPlugin)
		{
			_log.Debug(DEBUG_NORM, "CSQLHelper::UpdateValueInt: Notifying plugin %u about creation of device %u", HardwareID, unit);
			Plugins::CPlugin* pPlugin = (Plugins::CPlugin*)pHardware;
			pPlugin->DeviceAdded(ID, unit);
		}
#endif

```

pipiche38:
As you can see with the plugin , the problem occurs only when changing the (Type, SubType, SwitchType) to a new value, the plugin is updating back and forth every 30 seconds, in the 2nd round there is no more device added.

So it looks like the DomoticzEx stack beleive that is a new device because the (Type, SubType, SwitchType) has changed to a non existing one !!! Which is not the case when using the Legacy Framework

And I guess when using the native Virtual Devices, we can change Type, Subtype ... without creating a new device !!!

gizmocuz:
Okey, but why change the type/subtype? This should never happen in code by an update, only by a create
And when you created something with a wrong type/subtype, delete the device, so it get's created with the right type/subtype the next time the plugin in launched?

pipiche38:
The use case is linked to native zigbee groups, which depending on the devices attached to the group could change.

[...]

PR #6041, opened 2024-03-19
handle change of type subtype switchtype when TypeName is specified

badzz:
As in legacy python framework, when we update the python device and a typename is specified, we need to udpate the device type,subtype and switchtype of the current device if it is not the same as before.
it should address (partly) : https://github.com/domoticz/domoticz/issues/6027


gizmocuz:
Thanks for the PR!!

PR #6044, opened 2024-03-26
code cleaning : replace C style string to numbers conversions by C++ style

efdeb:
Replaced C style string to number conversion functions by their C++ style equivalents:
e.g. atoi(sValue.c_str()) -> stoi(sValue), and same for other similar functions : atol, atof, strtol, strtoll, strtoul, strtoull, strtod.

Removed unnecessary cast specifiers:
e.g. (float) stof(str) -> stof(str) or static_cast<float>stof(str) -> stof(str)

Replaced stof by stod when appropriate.

Replaced signed functions by unsigned when appropriate.

Checked (and rechecked, and rechecked again) the changes made.

### DIFF
--- a/hardware/1Wire/1WireByKernel.cpp
+++ b/hardware/1Wire/1WireByKernel.cpp
@@ -396,7 +396,7 @@ float C1WireByKernel::ThreadReadRawDataHighPrecisionDigitalThermometer(const std
 	if (bFoundCrcOk)
 	{
 		if (is_number(data))
-			return stof(data) / 1000.0F; // Temperature given by kernel is in thousandths of degrees
+			return static_cast<float>(stod(data) / 1000.0); // Temperature given by kernel is in thousandths of degrees
 	}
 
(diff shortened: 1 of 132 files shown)


keryell:
There has been a somewhat related discussion in the past: https://github.com/domoticz/domoticz/pull/5086#discussion_r781911690

efdeb:
OK @keryell, let's @gizmocuz decide.
I would prefer using C++ Standard Library, but domoticz is in production and has to keep running !

Another option could be to encapsulate C++ style functions and manage exceptions returning a valid value, but with a debug message in the log that could be used to fix the problem.

gizmocuz:
Aldo this is a valid PR (Thank you!), it is very dangerous as with this, we need to add exception handling everywhere.
As users can theoretically add anything in fields via the API, which is of course not correct, wrong data (like text, or illegal values) will trigger exceptions.
There are no problems with the current implementation

PR #6045, opened 2024-03-27
fix STYPE_END value in RFXNames.h

efdeb:
set to 23, but should be 22 according to the comment
 => fixed to be STYPE_BlindsPercentageWithStop + 1

(could also just not be manually assigned if other enum values are in the ascending order)


gizmocuz:
Thank you!

PR #6049, opened 2024-03-29
update setup script for pi4

akamming:
should fix the libssl1.1 error for new installs on bookworm

unfortunately i have no spare pi to test... so looking for someone who will test...


waltervl:
It is probably not only a Pi issue .. every Linux version on any architecture that uses new kernel based on libssl3 will be affected and needs to be tested.

gizmocuz:
Thank you for your PR and time! Hope we can migrate soon to 3.0

raibisch:
...same issue on actual debian bookworm (this is the base for all actual arm64 distros)
so please update to libssl3. (Docker is - like mentioned above - not a real solution for hardware access on embedded systems)

gizmocuz:
@raibisch , sure because this platform also uses the new libssl.
Depends on the hardware you are using... simple things like usb devices should have no issues.

But when using the current installer, it should install libssl1.1, so there should be no issue?

Issue #5336, opened 2022-09-11
OpenWeatherMap failed to retrieve data, onecall version 3.0 not supported (for new OWM accounts)


gsagsa99:
Domoticz Version: 2022.1 (build 14477)
Platform : Windows 10

OpenWeatherMap is not supporting version 2.5 of the 'onecall' API.  

I specified the city ID = 6455259 (Paris).  The first call (version 2.5) worked fine .  The pair Lat/Lon is retrieved.
This pair is used to build version 2.5 of 'onecall' API.   This is failing.

Below the log

2022-09-10 11:08:24.727  Debug: Meteo_Local: Got location parameter 6455259
2022-09-10 11:08:24.727  Debug: Meteo_Local: Starting with setting 0, 0, 0, 0
2022-09-10 11:08:24.727  Status: Meteo_Local: Using specified location (Station ID 6455259)!
2022-09-10 11:08:24.728  Debug: Meteo_Local: Get data from https://api.openweathermap.org/data/2.5/weather?id=6455259&appid=f4bc1306c8c0bda6307d0dc8941437a4&units=metric&lang=fr
2022-09-10 11:08:24.903  Status: Meteo_Local: City (6455259)-> Lat/Long = 48.8565,2.3524
2022-09-10 11:08:24.904  Status: Meteo_Local: Started
2022-09-10 11:08:27.907  Debug: Meteo_Local: Get data from https://api.openweathermap.org/data/2.5/onecall?lat=48.8565&lon=2.3524&exclude=minutely&appid=f4bc1306c8c0bda6307d0dc8941437a4&units=metric&lang=fr
2022-09-10 11:08:28.005  Debug: HTTP 401: Unauthorized
2022-09-10 11:08:28.007  Error: Meteo_Local: Error getting http data!

I checked both using a browser :  the first one (weather) is OK and the second one fails.

Instead using the version 3.0 of 'onecall' is OK.  Below the working URL (**just replace 2.5 with 3.0**)

https://api.openweathermap.org/data/3.0/onecall?lat=48.8565&lon=2.3524&exclude=minutely&appid=f4bc1306c8c0bda6307d0dc8941437a4&units=metric&lang=fr

Regards
gsagsa99

kiddigital:
@gsagsa99 , thx for reporting. I will take a look asap...

Did not get/see any deprecation message/e-mail coming by. Also no message on their website as far as I can see.

[...]

PR #6053, opened 2024-04-06
Handle OpenWeatherMap API 2.5 and 3.0

pipiche38:
Address issues #5336 and propose an implementation which is backward compatible.

If you have an old/legacy OpenWeatherMap API based on 2.5 you don't have to change anything.
If you have a new API  Key based on 3.0, then you just have to prefix your key with "V3.0:" like

Screenshot 2024-04-06 at 11 41 34


gizmocuz:
Thank you for your PR and time! I made the changes and as discussed also changed the web part

PR #6058, opened 2024-04-19
Release Python GIL in case of blocking IO

pipiche38:
PR #4518 was made to release GIL when doing update, but SQL read could also be blocked and GIL need to be released to leave other threads to run.

### DIFF
--- a/hardware/plugins/PythonObjects.cpp
+++ b/hardware/plugins/PythonObjects.cpp
@@ -677,7 +677,9 @@ namespace Plugins {
 		{
 			// load associated devices to make them available to python
 			std::vector<std::vector<std::string> > result;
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thank you for this PR!

pipiche38:
@gizmocuz I'm not use to the updatebeta script, but we have just installed Version: 2024.4 (build 15945) and wonder how I can check if the PR is included or not ?

Screenshot 2024-04-23 at 12 08 44

kiddigital:
@pipiche38 You can check using the buildnumber which commits on Github are included.

This is described on the Wiki 😁

gizmocuz:
It's included in this beta

pipiche38:
I'll push a PR for the DomoticzEx which present the same issue

PR #6075, opened 2024-05-08
[BleBox] rewrite, fixing, add RSSI, less requests

szczukot:


gizmocuz:
Thank you for this PR!!!

Is it possible to make one change in SQLHelper.cpp

```
else
						if (deviceId.size() == 7)

```

combine these two on one line

`else if (deviceId.size() == 7)`

szczukot:
done

gizmocuz:
Thanks again!!

PR #6077, opened 2024-05-10
Show selector level instead of index

szczukot:
image

### DIFF
--- a/hardware/BleBox.cpp
+++ b/hardware/BleBox.cpp
@@ -908,13 +908,13 @@ Json::Value BleBox::SendCommand(const std::string& IPAddress, const std::string&
 
 	if (!ParseJSon(result, root))
 	{
-		Log(LOG_ERROR, "Invalid json received!");
+		Log(LOG_ERROR, "(IP: %s, Command: %s) Invalid json received!", IPAddress.c_str(), command.c_str());
 		return root;
 	}
 
 	if (root.empty())
 	{
-		Log(LOG_ERROR, "Json is empty!");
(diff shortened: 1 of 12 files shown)


szczukot:
@gizmocuz is it ok, or I need to change somethong?

gizmocuz:
I am afraid this is way too risky; others might depend on the level value. Also, internally we work with levels (dzVents)
I think this is changing the sValue?

szczukot:
sValue is not changed.
This will change the data (status) returned only by one function: CWebServer::GetJSonDevices
And only for gswitch_sSetLevel
Instead of 
`Set Level: 10 % (where 10 is index)`
we will get
`Set Level: SelectorNameForIndexEqual10`

This is only returned status string.
Capture

Capture

Because for selector 10,20,30 etc are not percentage level or something like this. This is only index for selector value

gizmocuz:
Correct, and the 'level' (index) is also still in the JSON output? Third party developers need to have the (level)index
But this is only a problem in the devices overview right?

szczukot:
Yes, level is still returned, without changes.

gizmocuz:
Thanks for checking.
I also think this is a good idea, but it should be in the 'Data' field generated only inside the function '**CWebServer::GetJSonDevices**'
All the other code is not necessary ...

gizmocuz:
I made the necessary changes...

This was all that was needed

```
							std::vector<std::string> strarray;
							StringSplit(levelNames, "|", strarray);
							int isLevel = llevel / 10;
							if (static_cast<int>(strarray.size()) >= isLevel)
							{
								lstatus = strarray[isLevel];
							}
```

Thanks for the PR and your time!

PR #6078, opened 2024-05-10
Add On/Off TypeName

guillaumezin:
Because "Switch" is generic and doesn't allow to revert back to SwitchType = STYPE_OnOff when calling CUnitEx_update (Plugin Framework extended) if you created for instance a "Contact" switch from plugin


guillaumezin:
I added the fan device to TypeName too

gizmocuz:
Thanks!!!

PR #6079, opened 2024-05-11
Add properties and options update to unit in python framework extended

guillaumezin:
To be on par with normal python plugin framework, add UpdateProperties and UpdateOptions to Update method, to be able to update properties and options of unit

If PR is accepted, I will update the Wiki.


gizmocuz:

PR #6088, opened 2024-05-31
Let CounterReport graph show negative values

waltervl:
Screenshot from 2024-06-01 00-34-19


gizmocuz:
Thanks for the PR!
Just a quick question, could you post a screenshot without april/may, does it start at 0?
What are you counting?

waltervl:
I do not understand your question. But before my change the months April and May had no value shown in the graph.
I am counting my total Electric energy usage (I have an old fashioned turning electricity meter). 
Currently abroad so cannot make further screenshots until next week.

PR #6093, opened 2024-06-07
Add Custom Manual Switch configuration to python plugins

ultrasuperpingu:
This probably needs more testing, but the idea is to allows plugin to be able to use the "Add Manual Switch" interface.

### DIFF
--- a/hardware/DomoticzHardware.cpp
+++ b/hardware/DomoticzHardware.cpp
@@ -19,14 +19,24 @@ bool CDomoticzHardwareBase::CustomCommand(const uint64_t /*idx*/, const std::str
 	return false;
 }
 
-std::string CDomoticzHardwareBase::GetManualSwitchesJsonConfiguration() const
+bool CDomoticzHardwareBase::HasManualSwitchesSupport()
+{
+	return false;
+}
+
+std::string CDomoticzHardwareBase::GetManualSwitchesJsonConfiguration()
 {
 	return std::string("");
 }
 
-void CDomoticzHardwareBase::GetManualSwitchParameters(const std::multimap<std::string, std::string> & /*Parameters*/, _eSwitchType & /*SwitchTypeInOut*/, int & /*LightTypeInOut*/,
-	 int & /*dTypeOut*/, int &dSubTypeOut, std::string & /*devIDOut*/, std::string & /*sUnitOut*/) const
+bool CDomoticzHardwareBase::AddManualSwitch(const std::string Name, _eSwitchType SwitchType, int Type, const std::multimap<std::string, std::string>& Parameters)
 {
+	return false;
+}
+
+bool CDomoticzHardwareBase::TestManualSwitch(_eSwitchType SwitchType, int Type, const std::multimap<std::string, std::string>& Parameters)
+{
+	return false;
 }
 
 bool CDomoticzHardwareBase::Start()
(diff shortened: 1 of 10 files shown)


gizmocuz:
Why would this be needed? The hardware base class has the options to add/update a switch already

ultrasuperpingu:
The idea is for the python plugins to be able to use the "Add Manual Switch/Light" form to add a Switchs/Lights (and to be able to configure the data they need in that form).

gizmocuz:
Okey, thank you for this PR but this is not needed. Python plugins can create switches the normal way and that is perfect.

ultrasuperpingu:
I think there is a misunderstanding of what this PR is doing (this is my fault, my PR comment should have been more clear about it). The way the device are created (ie inserted into database) is still the same. This is just 3 new callbacks which informs plugin when a user use the "Add Switch" Form:
 - GetManualSwitchesJsonConfiguration: When the form is open to requests the list of device types and associated parameters
 - TestManualSwitch: When the user push the Test button
 - AddManualSwitch: When the user push the Add button (and the plugin need to call Device.Create() then)

For now, plugins needs to use a config file or something to create those devices. This config is usually only used when the plugin runs for the first time.

I provided an example plugin "ManualSwitches.py" to show the concept.

With this PR, plugins will be able to use an existing part of the GUI to configure their devices.

gizmocuz:
Do you need this for your plugin?
Did you had a look at a RFXCom tranciever or RFLink?
For now I do not see the benefit for these changes, sorry.

ultrasuperpingu:
`Did you had a look at a RFXCom tranciever or RFLink?`

Those are native no? And you can use the "Add Manual Switch" form to add devices?

That's exactly what I would like to be able to do in my plugins :)

But Ok, I understand.
Thanks for you time

gizmocuz:
Yes, these are natively supported by Domoticz and seem to do the same as the hardware you are using.
It might be a good idea to use one of the two instead of spending a lot of ours getting your hardware to work

ultrasuperpingu:
I guess you're saying that because the PR modify the CZiBlueBase class. But it's not to make this hardware work, it's because this device use the possibility to customize the "Add Manual Switch" form and since it's the feature I was porting to the python plugins, I had to modify it.
But ok, has I said, I understand, no problem. Sorry for spending your time.

gizmocuz:
No, I thought I had seen you work on a plugin on Github to support a 433Mhz transceiver like RFXCom/RFLink/ZiBlue
I could be wrong about this...
But I really think this PR is not needed.
Thank you for the PR!

ultrasuperpingu:
You're right on that, I just coded a python plugin to manage "AirSend Duo" which support both 433MHz and 867MHz.

First of all, to be clear, I do not send this response to convince you to integrate this PR. I understand this is a no go for you and I'm ok with it, it's just for you to know :)

Sadly, AFAIK, RFXCom and RFLink does not support all the protocols I need, especially Somfy IO (I know RTS is ok, but it seems Homecontrol IO is not). And I think there is no Dual Band RFXCom/RFLink (so I'd need to buy 2 of them :(). Maybe I'm wrong about it but that was my conclusion when I tried to find a solution that met my needs...

But the good news is domoticz python plugin is just great. I manage to make the module works in domoticz in just a couple hours. And it allows the code to be really concise and clean.

So thank you very much for this good piece of software :)

PR #6128, opened 2024-07-21
Ensure AllowPlainBasicAuth settings is set to both HTTP and HTTPS webserver (Fixes #6064)

### LINKED ISSUE #6064: I can't send json.htm by HTTP
I can send something like
`https://192.168.1.2/json.htm?type=command&param=udevice&idx=164&nvalue=0&svalue=12`
and of course, in header I have 
Authorization: Basic xxx
Bu when I try send it by HTTP
`http://192.168.1.2:8080/json.htm?type=command&param=udevice&idx=164&nvalue=0&svalue=12`
I always get 401 - unauthorized

ps.  have enabled also "Allow Basic-Auth authentication over plain HTTP (API only)."

kiddigital:
PR _fixes_ issue #6064 .

When toggling the __API Protection__ (_Allow Basic-Auth authentication over plain HTTP (API only)._) on the _Settings/Security_ tab, the setting was only applied to the webserver process (Either HTTP or HTTPS) where the settings was changed instead of both.

Not really a big issue, but annoying when trying to secure your domoticz environment as currently it requires a restart of domoticz to make the setting active for both HTTP and HTTPS server.

So in case one was using a HTTPS connection to domoticz to toggle the setting and in parallel testing the API over HTTP, there was no noticeable effect until domoticz was restarted (to make the setting active on the HTTP server).

Now the setting is active immediately for both HTTP and HTTPS regardless if the setting is changed over a HTTP or HTTPS connection.

### DIFF
--- a/main/WebServerCmds.cpp
+++ b/main/WebServerCmds.cpp
@@ -2574,7 +2574,7 @@ namespace http
 				bool AllowPlainBasicAuth = (request::findValue(&req, "AllowPlainBasicAuth") == "on" ? 1 : 0);
 				m_sql.UpdatePreferencesVar("AllowPlainBasicAuth", AllowPlainBasicAuth);
 
-				m_pWebEm->SetAllowPlainBasicAuth(AllowPlainBasicAuth);
+				m_webservers.SetAllowPlainBasicAuth(AllowPlainBasicAuth);
 				cntSettings++;
 
 				std::string WebLocalNetworks = CURLEncode::URLDecode(request::findValue(&req, "WebLocalNetworks"));


gizmocuz:
I think the idea was to only allow this for HTTPS for safety reasons
Still think this is the best idea for such a weak method
Using an API key is maybe better but still for HTTPS

kiddigital:
> I think the idea was to only allow this for HTTPS for safety reasons Still think this is the best idea for such a weak method Using an API key is maybe better but still for HTTPS

The setting is specifically introduced to override the default safety of allowing only Basic-Auth over HTTPS. So it is meant to make it possible to allow Basic-Auth over HTTP although it is considered less secure.

An (OAuth2/OIDC) API key/token is much safer indeed and that is already possible now levering the build in IAM service :)

I think our [Security_Setup.md](https://github.com/domoticz/domoticz/blob/development/SECURITY_SETUP.md) clearly explains it ;)

gizmocuz:
I read think we should stop supporting the basic authentication method for plain HTTP... It's a real bad method for authentication and much worse on a unencrypted line.
If the current behaviour is that it is not working for HTTP than the current implementation is correct
If it is partly working for HTTP then maybe a patch to remove it?

kiddigital:
> I read think we should stop supporting the basic authentication method for plain HTTP... It's a real bad method for authentication and much worse on a unencrypted line. If the current behavior is that it is not working for HTTP than the current implementation is correct If it is partly working for HTTP then maybe a patch to remove it?

It is fully working in the current code base :)
It is just the 'instant' activation of the setting on both the HTTP and HTTPS webserver threads that isn't. A restart is needed currently to make it active on both threads. This PR _fixes_ that.

I do agree that we want to move away from Basic-Auth all together sometime... 

But I guess that there are still quite a few users that use _less safe_ methods like not disabling the HTTP webserver and only use the HTTPS webserver, using self-signed SSL certificates, have IP-ranges in the _trusted network_, etc. Which is ok'isch if you are running fully internal/locally and are confident that your perimeter defenses are sufficient :)

And you could consider Basic-Auth over HTTP _safer_ than using anonymous access with in the _trusted network_ :)

In the end, this 1-line code change of this PR doesn't change much :) It just removes a _nuisance_ for some during testing/configuring the security of Domoticz.

gizmocuz:
By default HTTPS is enabled with a self signed certificate.
It is no different to call HTTPS instead of HTTP, it is just the letter S that makes it different
Those users you are referring to that are 'able' to call a API with basic auth, are also able to add an additional letter to the protocol I suppose... 

Some of those users use basic auth remotely... Like with IFTTT... I do consider a fixed IP address much more secure than basic auth over HTTP

And when it's in a local network, for sure

gizmocuz:
Thank you!!

PR #6132, opened 2024-08-03
Change to improve token handling of the Netatmo hardware device

LoukKamp:
When the worker (in hardware/Netatmo.cpp) is restarted or stopped (by disabling the hardware device or stopping Domoticz), the worker will use an expired refresh token or tries to log in with an expired authorization code. In some cases, a valid refresh token is lost, caused by the web server, not supporting this field.

This is partly caused by a change in the authorization process since May 29, 2024:

'Starting May 29, 2024, this behaviour will change to comply with the [OAuth2 Authorization Framework RFC recommendations (section 10.4)] (https://www.rfc-editor.org/rfc/rfc6749) and improve the security of our users' data.
When refreshing the tokens, the Access Token and Refresh Token will be different from the previous ones and the old tokens will be invalidated.'

This requires that the refresh token needs to be updated in its storage, so the new code can be used after a restart of the worker.

This patch will include 3 functional changes and some additional text in log and debug messages:

1.  Store the new refresh token after each token refresh request (about every 3 hours).
2. Refresh the tokens before stopping or restarting the hardware device, so the user has 3 hours to fix the reason why the hardware device needed to be stopped or restarted.
3. Don't let the web server update the refresh token in its storage, because it has an uninitialized value (0).

I've tested this changes, and it all works as expected, so It's ready to review and/or merged

### DIFF
--- a/hardware/Netatmo.cpp
+++ b/hardware/Netatmo.cpp
@@ -221,7 +221,7 @@ bool CNetatmo::Login()
 		}
 	}
 
-	Log (LOG_STATUS, "Requesting new tokens using authorization code");
+	Log (LOG_NORM, "Requesting new tokens using authorization code");
 
 	//Loggin on the API
 	std::stringstream sstr;
@@ -302,7 +302,7 @@ bool CNetatmo::RefreshToken(const bool bForce)
 			return true; //no need to refresh the token yet
 	}
 
-	Log (LOG_STATUS, "Requesting refreshed tokens");
+	Log (LOG_NORM, "Requesting refreshed tokens");
 
 	// Time to refresh the token
 	std::stringstream sstr;
--- a/hardware/Netatmo.cpp
+++ b/hardware/Netatmo.cpp
@@ -196,6 +196,7 @@ void CNetatmo::Do_Work()
 			}
 		}
 	}
+	RefreshToken(true);
 	Log(LOG_STATUS, "Worker stopped...");
 }
 
@@ -243,7 +244,7 @@ bool CNetatmo::Login()
 	//Check for returned data
 	if (!ret)
 	{
-		Log(LOG_ERROR, "Error connecting to Server...");
+		Log(LOG_ERROR, "Error connecting to Server (login)...");
 		return false;
 	}
 
@@ -252,14 +253,14 @@ bool CNetatmo::Login()
 	ret = ParseJSon(sResult, root);
 	if ((!ret) || (!root.isObject()))
 	{
-		Log(LOG_ERROR, "Invalid/no data received...");
+		Log(LOG_ERROR, "Invalid/no data received (login)...");
 		return false;
 	}
 	//Check if access was granted
 	if (root["access_token"].empty() || root["expires_in"].empty() || root["refresh_token"].empty())
 	{
-		Log(LOG_ERROR, "No access granted, check credentials...");
-		Debug(DEBUG_HARDWARE, "No access granted, check credentials...(%s)(%s)", httpData.c_str(), root.toStyledString().c_str());
+		Log(LOG_ERROR, "No access granted, check credentials (Login)...");
+		Debug(DEBUG_HARDWARE, "No access granted, check credentials (Login)...(%s)(%s)", httpData.c_str(), root.toStyledString().c_str());
 		return false;
 	}
 
@@ -320,7 +321,7 @@ bool CNetatmo::RefreshToken(const bool bForce)
 	//Check for returned data
 	if (!ret)
 	{
-		Log(LOG_ERROR, "Error connecting to Server...");
+		Log(LOG_ERROR, "Error connecting to Server (refresh tokens)...");
 		return false;
 	}
 
@@ -329,7 +330,7 @@ bool CNetatmo::RefreshToken(const bool bForce)
 	ret = ParseJSon(sResult, root);
 	if ((!ret) || (!root.isObject()))
 	{
-		Log(LOG_ERROR, "Invalid/no data received...");
+		Log(LOG_ERROR, "Invalid/no data received (refresh tokens)...");
 		//Force login next time
 		m_isLogged = false;
 		return false;
@@ -339,7 +340,7 @@ bool CNetatmo::RefreshToken(const bool bForce)
 	if (root["access_token"].empty() || root["expires_in"].empty() || root["refresh_token"].empty())
 	{
 		//Force login next time
-		Log(LOG_ERROR, "No access granted, forcing login again...");
+		Log(LOG_ERROR, "No access granted, forcing login again (Refresh tokens)...");
 		m_isLogged = false;
 		return false;
 	}
@@ -351,6 +352,7 @@ bool CNetatmo::RefreshToken(const bool bForce)
 	//Store the duration of validity of the token
 	m_nextRefreshTs = mytime(nullptr) + expires;
 
+	StoreRefreshToken();
 	return true;
 }
 
@@ -750,7 +752,7 @@ void CNetatmo::GetMeterDetails()
 		httpUrl = MakeRequestURL(NETYPE_ENERGY);
 		if (!HTTPClient::GET(httpUrl, ExtraHeaders, sResult))
 		{
-			Log(LOG_ERROR, "Error connecting to Server...");
+			Log(LOG_ERROR, "Error connecting to Server (GetMeterDetails)...");
 			return;
 		}
 
@@ -758,7 +760,7 @@ void CNetatmo::GetMeterDetails()
 		bRet = ParseJSon(sResult, root);
 		if ((!bRet) || (!root.isObject()))
 		{
-			Log(LOG_ERROR, "Invalid data received...");
+			Log(LOG_ERROR, "Invalid data received (GetMeterDetails)...");
 			return;
 		}
 		if (!root["error"].empty())
@@ -782,7 +784,7 @@ void CNetatmo::GetMeterDetails()
 	httpUrl = MakeRequestURL(m_weatherType);
 	if (!HTTPClient::GET(httpUrl, ExtraHeaders, sResult))
 	{
-		Log(LOG_ERROR, "Error connecting to Server...");
+		Log(LOG_ERROR, "Error connecting to Server (GetMeterDetails)...");
 		return;
 	}
 
@@ -790,7 +792,7 @@ void CNetatmo::GetMeterDetails()
 	bRet = ParseJSon(sResult, root);
 	if ((!bRet) || (!root.isObject()))
 	{
-		Log(LOG_ERROR, "Invalid data received...");
+		Log(LOG_ERROR, "Invalid data received(GetMeterDetails)...");
 		return;
 	}
 	if (!root["error"].empty())
@@ -812,7 +814,7 @@ void CNetatmo::GetMeterDetails()
 			httpUrl = MakeRequestURL(NETYPE_HOMECOACH);
 			if (!HTTPClient::GET(httpUrl, ExtraHeaders, sResult))
 			{
-				Log(LOG_ERROR, "Error connecting to Server...");
+				Log(LOG_ERROR, "Error connecting to Server (ParseStationData)...");
 				return;
 			}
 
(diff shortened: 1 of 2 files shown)


gizmocuz:
Thank you for your PR!!

Question, why do you refresh the token when the hardware is stopped? (This also happens when Domoticz is stopped)
Why not refresh the token at the top of Do_Work() ? So everytime the hardware is (re)started, the token will be refreshed

LoukKamp:
> Thank you for your PR!!
> 
> Question, why do you refresh the token when the hardware is stopped? (This also happens when Domoticz is stopped) Why not refresh the token at the top of Do_Work() ? So everytime the hardware is (re)started, the token will be refreshed

Because both the access and refresh tokens expire in 3 hours. Rather strange, because I expect that the refresh token would have a much longer expiration time.

When you post a token request, the reply contains two expiration times, both 10800 seconds.

I put this call at the end of the worker methode to be sure that the tokens won't expire after a (very) short time, because the tokens where last refreshed almost 3 hours ago. In that case the user has very little time left to restart.

gizmocuz:
How about delete the internal token so it gets refreshed when the hardware is started (Domoticz, or after a disable/enable)
It makes no use to refresh a token when Domoticz is stopped (or the hardware is stopped)

It will also cause the shutdown of Domoticz/system to become slower

LoukKamp:
> How about delete the internal token so it gets refreshed when the hardware is started (Domoticz, or after a disable/enable) It makes no use to refresh a token when Domoticz is stopped (or the hardware is stopped)
> 
> It will also cause the shutdown of Domoticz/system to become slower

I see your point, but what do you mean by the internal token? When this is the access token, then this token isn't stored at all. When you mean the refresh token, then we are back at where it all started about: The only way to get e new refresh token, is posting a login request, using the authorization code, which is only valid a view seconds (30 .. 60?). To get a new authorization code the user has to request one from the Netatmo site manually.

Maybe a compromise: Only refresh when the token expires within an hour?

gizmocuz:
Okey, so it is already being refreshed/loaded/renewed at startup?
Sure, when you cannot get any data because of a token issue, refresh the token, retry again, else wait for the normal poll interval and retry it again

JurgenLB:
I also believe that line 199 `RefreshToken(true);` is not necessary at all.
On my local setup it is working without that line and the refresh is active after restart and/or disabling 

kind regards

LoukKamp:
Ok, I'll do another test: stop the device, wait 3 hours until the refresh token is expired, and then restart again.

I expect it will fail to refresh.

kiddigital:
It is in general good practice to renew an Access Token before it expires. Like after more than 50-80% of the Access Token lifespan period has gone by.

This is done either by re-login or using a valid Refresh Token.

As in this case the lifespan of the Access Token and Refresh Token are the same (3 hours), which normally isn't the case by the way as most Refresh Token have a lifespan of twice or more of the lifespan of the Access Token, a refresh of the Access Token (and also the Refresh Token that comes with it) needs to be done (way) before expiration.

An Access Token should never be stored/persisted and only kept in memory. This means at startup a new Access Token always has to be retrieved.

A Refresh Token can be persisted and used at startup for example to get a new Access Token, saving the user to go through the (re)login procedure.

As the Refresh Token has a relatively short lifespan, it makes sense to get a new Refresh Token, and store it, when stopping the process. Otherwise changes are high that when restarting the process, the stored Refresh Token actually expired.

So it makes sense to me to get a new Refresh Token when stopping the process.

(And get a new Access Token when the process starts of course)

LoukKamp:
> It is in general good practice to renew an Access Token before it expires. Like after more than 50-80% of the Access Token lifespan period has gone by.
> 
> This is done either by re-login or using a valid Refresh Token.
> 
> As in this case the lifespan of the Access Token and Refresh Token are the same (3 hours), which normally isn't the case by the way as most Refresh Token have a lifespan of twice or more of the lifespan of the Access Token, a refresh of the Access Token (and also the Refresh Token that comes with it) needs to be done (way) before expiration.
> 
> An Access Token should never be stored/persisted and only kept in memory. This means at startup a new Access Token always has to be retrieved.
> 
> A Refresh Token can be persisted and used at startup for example to get a new Access Token, saving the user to go through the (re)login procedure.
> 
> As the Refresh Token has a relatively short lifespan, it makes sense to get a new Refresh Token, and store it, when stopping the process. Otherwise changes are high that when restarting the process, the stored Refresh Token actually expired.
> 
> So it makes sense to me to get a new Refresh Token when stopping the process.
> 
> (And get a new Access Token when the process starts of course)

I totally Agree!

However, with a shorter interval, network traffic will increase, and I'm not sure Netatmo will appreciate this.

So we also could decrease the internal to 2 hours (1/3 of 3 hours. in that case, the user has at least 1 hour to restart the device and/or Domoticz.

kiddigital:
> > It is in general good practice to renew an Access Token before it expires. Like after more than 50-80% of the Access Token lifespan period has gone by.

...

> 
> 
> 
> I totally Agree!
> 
> 
> 
> However, with a shorter interval, network traffic will increase, and I'm not sure Netatmo will appreciate this.
> 
> 
> 
> So we also could decrease the internal to 2 hours (1/3 of 3 hours. in that case, the user has at least 1 hour to restart the device and/or Domoticz.

Yes, internally requesting a new Access Token (which comes with a new Refresh Token) after 2/3 of the lifespan is acceptable (also for Netatmo I assume as they choose 3 hours as the TTL).

But I would think that 1 additional request which happens sooner when stopping the process, is no issue at all. As stopping of course does not happen frequently:) Stopping with a Refresh Token that 'survives' almost 3 hours is to be preferred over stopping with a Refresh Token that could only be valid for an hour or a little more.

Don't think this any issue at all for Netatmo.

gizmocuz:
No that should not be an issue.

Anyway, requesting a new access token at closure/shutdown does, however, make no sense.

So please move this RefreshToken call from the bottom of do_work to the top of do_work

LoukKamp:
This afternoon, I disabled the device for about 5 hours, and after enabling again, it turns out that the refresh token was still valid! :)

2024-08-04 13:52:13.445 Status: NetAtmoWeerEnLuchtkwaliteit: Worker stopped...
2024-08-04 15:45:01.619 Status: Notification: Sensor Slaapkamer Oostzijde - CO2 Last Update: 2024-08-04 14:45:00 [>= 60 minutes]
2024-08-04 15:45:01.621 Status: Notification: Sensor Woonkamer - CO2 Last Update: 2024-08-04 14:45:00 [>= 60 minutes]

... 

2024-08-04 19:19:41.515 Status: NetAtmoWeerEnLuchtkwaliteit: Worker started...
2024-08-04 19:20:01.513 Status: Notification: [Recovered] Sensor Slaapkamer Oostzijde - CO2 Last Update: 2024-08-04 14:45:00 [>= 60 minutes]
2024-08-04 19:20:01.514 Status: Notification: [Recovered] Sensor Woonkamer - CO2 Last Update: 2024-08-04 14:45:00 [>= 60 minutes]


Note the absence of login failures: This means that the last received refresh token was still valid, otherwise the request to refresh  the access and refresh tokens would have leaded to an Invalid Grand exception. 

So this means that the refresh token is much longer valid than the expiration time returned by the response of the refresh token request (10800 seconds = 3 hours).  So there is no reason to refresh the tokens on exit of the worker.

There is also no need to move this call to the start of the worker, because the token is loaded from the database, and then used to request the tokens right away.

I followed the advice from @kiddigital and reduced the interval time to 2/3 of the expiration time of the access token. Before, this was only 15 seconds. However, I left this unchanged, because this might influence the initial startup.

LoukKamp:
I think I applied the change in this pull request.

Is it OK now?

gizmocuz:
Maybe you could make another small code cleanup change

Currently there is:

```
				if ((sec_counter % 900 == 0) || (bFirstTimeWS))
				{
					bFirstTimeWS = false;
					if ((m_bPollWeatherData) || (sec_counter % 1200 == 0))
						GetMeterDetails();
				}

				//Thermostat data is updated every 10 minutes
				if ((sec_counter % 900 == 0) || (bFirstTimeTH))
				{
					bFirstTimeTH = false;
					if (m_bPollThermostat)
						GetThermostatDetails();
				}
```

to:

```
				if ((sec_counter % 900 == 0) || (bFirstTimeWS) || (bFirstTimeTH))
				{
					bFirstTimeWS = false;
					bFirstTimeTH = false;
					if (m_bPollWeatherData)
						GetMeterDetails();
					if (m_bPollThermostat)
						GetThermostatDetails();
				}

```

Not sure why the %1200 was there, maybe before it was polled every 20 minutes?

And maybe better to add somewhere at the top of the code the following

#define NETAMO_POLL_INTERVALL 900

and use that instead of %900

This way it is also easier later to change it


Maybe later have a look at when/how often and why 'GetThermostatDetails' is called maybe too much (m_tSetpointUpdateTime)

LoukKamp:
> Maybe you could make another small code cleanup change
> 
> Currently there is:
> 
> ```
> 				if ((sec_counter % 900 == 0) || (bFirstTimeWS))
> 				{
> 					bFirstTimeWS = false;
> 					if ((m_bPollWeatherData) || (sec_counter % 1200 == 0))
> 						GetMeterDetails();
> 				}
> 
> 				//Thermostat data is updated every 10 minutes
> 				if ((sec_counter % 900 == 0) || (bFirstTimeTH))
> 				{
> 					bFirstTimeTH = false;
> 					if (m_bPollThermostat)
> 						GetThermostatDetails();
> 				}
> ```
> 
> to:
> 
> ```
> 				if ((sec_counter % 900 == 0) || (bFirstTimeWS) || (bFirstTimeTH))
> 				{
> 					bFirstTimeWS = false;
> 					bFirstTimeTH = false;
> 					if (m_bPollWeatherData)
> 						GetMeterDetails();
> 					if (m_bPollThermostat)
> 						GetThermostatDetails();
> 				}
> ```
> 
> Not sure why the %1200 was there, maybe before it was polled every 20 minutes?
> 
> And maybe better to add somewhere at the top of the code the following
> 
> #define NETAMO_POLL_INTERVALL 900
> 
> and use that instead of %900
> 
> This way it is also easier later to change it
> 
> Maybe later have a look at when/how often and why 'GetThermostatDetails' is called maybe too much (m_tSetpointUpdateTime)

Done this.

However, I'n not able to test this, because I don't have all the devices.

gizmocuz:
Thank you for all your work!!

LoukKamp:
Another` (dummy) question: Since yesterday, each time I do a (local) commit, I get:

```
On branch development
Your branch is ahead of 'origin/development' by 3 commits.
  (use "git push" to publish your local commits)

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   extern/fmtlib (new commits)
        modified:   extern/jsoncpp (new commits)
        modified:   extern/minizip (new commits)
        modified:   extern/mosquitto (new commits)
        modified:   extern/sqlite-amalgamation (new commits)
        modified:   hardware/Netatmo.cpp

no changes added to commit (use "git add" and/or "git commit -a")

```
After adding my own changed file, the others files stay in the list, even when a do a git restore.

These are all directory files and when I do a git diff, I get:

```
diff --git a/extern/fmtlib b/extern/fmtlib
index a1c6bfd77..58aba5a3d 160000
--- a/extern/fmtlib
+++ b/extern/fmtlib
@@ -1 +1 @@
-Subproject commit a1c6bfd77b409f21e234cc7042795ca4d31fa020
+Subproject commit 58aba5a3defb6e46031392f1a58156d093efe5e0
```

(this repeats for all 5 folders)

Where does this came from, and how do I get rid of this?

Edit: Hence that those folders aren't touched for about 4 years now...

gizmocuz:
I donno, but try doing this

git pull
git reset --hard origin/development

gizmocuz:
Your changes are now in the latest beta

LoukKamp:
> Your changes are now in the latest beta

That's great; thanks!

LoukKamp:
> I donno, but try doing this
> 
> git pull git reset --hard origin/development

No, still there...

I also tried:
   git pull extern/sqlite-amalgamation
   git restore -3 extern/fmtlib


Did you see all those messages above from github-actions?  I thought it was about the sentence about touch files that were not mine, so I removed that. However, I'm still accused of using bad language... :(

LoukKamp:
According to Lesly from Netatmo, there is no expiration time on the refresh token:

> The refresh_token doesn't have an expiration date. It will always remain valid even if you use it several days later. Are you sure you don't have any routine sending a request to the /token endpoint that invalidates the tokens ? 
> 
> You can also try - if you don't already do it - to refresh the access_token a bit before the 10800 seconds expiration to see if it can help 
> 
> Have a good day,
> 
> Leslie - Community Manager
>  

So, as long as we keep the refresh token at a save place, we don't need to save the Client secret and authorization code.

PR #6133, opened 2024-08-06
Frank Energie missing as gas supplier

arjendekorte:
The Enever JSON output also lists Frank Energie as gas supplier. Add it so the users which use then as their supplier to use the correct data.


gizmocuz:
Thanks!!!

PR #6139, opened 2024-08-18
Make update and delete buttons on hardware page really disabled, not only grey: INCORRECT PR: DON'T MERGE!!!

LoukKamp:
The delete buttons, which are really just link elements (<a>), are made gray when there is no device select: Only addition of new devices should be possible.

However, when the user clicks on this button, the last referenced device will be updated of deleted: The "button" seems to be disabled (because it is gray), but actually, it is still active.

Adding a disabled attribute, doesn't help much for these elements.

So, I added a line with the removeAttr("href") function for each button after the line that will make the button gray, so the link with the "on click handler" is removed


LoukKamp:
### This is not what I want!!!
I opened a Pull Request to add the update and delete button change only, but this also contains the changes I'm still working on for the Netatmo authorization process.

So I have two options: I change the title of this PR, or someone deletes this PR and I start all over.

@gizmocuz: Can you help me out with this? I don't seem to be able to create another branch for this simple patch (because I'm not a member of this organization, maybe)?

gizmocuz:
@LoukKamp , sure.
Please make a new Fork if that is possible, or in your checked-out fork, you should create a new branch like

`git checkout -b hardware-buttons-fix`

(-b creates a new branch)

now you could make your changes and commit and make a PR

you can switch to another branch with you do not use the -b like

`git checkout development`

you could make a copy of your current changes (or everything), and FORCE all your changes to be gone with

`git reset --hard origin/development`

now you can create a new branch with

`git checkout -b netamo`

Now copy your .cpp/h/html/js files back

Now you can develop patches for netamo and make a PR etc

But make a good backup before so you can always go back

It's best to get uses of making branches when you work on multiple things

LoukKamp:
So I need to do this on my local repository. I created a new branch, saved my changes in patch files, and did a reset to the Domoticz/development branch. This all seems to work on my local repository only: The online repository in GitHub still contains all the changes and history. So, I think, I need to reset that repository too, or maybe remove it all if that is possible.

I can't make a new fork from either the main development branch, or my own fork/branch somehow.

gizmocuz:
If you can't fork twice, than you indeed need to backup all your changes, delete your repository (online/local), and make a new fork and two branches locally
Reapply the changes for each of them

LoukKamp:
Thanks,

To be honest, I already did this yesterday evening (deleted the online repro, and created a new one with a different name). I cloned the new repo to my local system and made a new brach for the delete button change.

Now I have to rebuild Domoticz on my local system to be able to test the change, but I forgot that this will take several hours, so enough time to read/learn about using GitHub... 😉

Note: I'm an experienced software support engineer (I worked for over 20 years on railway security soft- and hardware. And as a hobby project, I grew from user to main developer on Centrafuse Auto (Domoticz for in the car 😉). However I never used a repository of GitHub alike software, so this explains my unexperienced and stupid mistakes... 🙁

Thanks for  help and patience, so far!

Louk.
Van: gizmocuz ***@***.***>
Verzonden: dinsdag 20 augustus 2024 07:50
Aan: domoticz/domoticz ***@***.***>
CC: Louk ***@***.***>; State change ***@***.***>
Onderwerp: Re: [domoticz/domoticz] Make update and delete buttons on hardware page really disabled, not only grey: INCORRECT PR: DON'T MERGE!!! (PR #6139)


If you can't fork twice, than you indeed need to backup all your changes, delete your repository (online/local), and make a new fork and two branches locally
Reapply the changes for each of them

LoukKamp:
@gizmocuz : Sorry to border you again, but I still have 2 questions that border me:

I need to transfer a value (the returned refresh token) and a flag from the button handler to the refresh and update Hardware functions in Hardware.js, but the refresh function does not have any parameters and is used very often, so I don't want to add a parameter just for this single purpose (except when it can be an optional parameter with a default value?) The solution I found was to use a hidden input field in the HTML to transport the value and the flag. **Now I wonder if this is the proper way or not?**

The second question is about the Update and Delete button change I made: For the oAuth2 change, I needed this function also for a login button, but I found an even more simple solution by merging the two functions (enable and disable) into one single function: Just one enable function with a flag (enable = true, disable = false). 

**Is it OK if I do the same for the update and delete buttons?** This is more logical, because both functions are defined next to each other. If so, I will do this after I pushed my oAuth2 changes, and hopefully before the next release.  

**Or can I add the change to the original PR, although this one is already merged and closed?**

gizmocuz:
you can push things in the scope with

$scope.my_value = 123;
$scope.my_other_value = 234;

And then you could look if values are set?

This is better than a hidden input

PR #6141, opened 2024-08-20
Hardware buttons fix

LoukKamp:
Make update and delete buttons on hardware page really disabled, not only gray

### DIFF
--- a/www/app/hardware/Hardware.js
+++ b/www/app/hardware/Hardware.js
@@ -1,5 +1,5 @@
 define(['app'], function (app) {
-	
+
 	/**
 	 * Table of hardware using the new structure.
 	 * This should really be auto-generated.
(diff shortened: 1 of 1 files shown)


LoukKamp:
Note: The first commit is white space changes only: checked with:

`git diff -w --ignore-blank-lines www/app/hardware/Hardware.js`

gizmocuz:

LoukKamp:
Question: Who is responsible for maintaining the history file; Am I supposed to add a line to it to describe this patch?

gizmocuz:
Yes, if you could do that... Before a new stable release I also check the complete commit history and add missing lines when needed

LoukKamp:
I will also use this fork to update the Netatmo oAuth2 patch, I think...
Is it OK when I make the change in the history file in the new PR for the NetAtmooAuth2 update, or should I do this in a separate PR?

gizmocuz:
Thats OK... but please create a new local branch

PR #6142, opened 2024-08-22
Netatmo oAuth2 improvements

LoukKamp:
Changed: Improved and automated login process for Netatmo devices (to obtain client ID and Password with user selectable scopes)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,7 +49,7 @@ install_manifest.txt
 /www/images/floorplans/*.jpg
 .vscode
 sqlite3_config.h
-/extern/
+extern/
 /node_modules/
 /domoticz.iml
 __pycache__
--- a/.gitignore
+++ b/.gitignore
@@ -49,7 +49,7 @@ install_manifest.txt
 /www/images/floorplans/*.jpg
 .vscode
 sqlite3_config.h
-extern/
+/extern/
 /node_modules/
 /domoticz.iml
 __pycache__
--- a/History.txt
+++ b/History.txt
@@ -1,5 +1,6 @@
 Version 2024.x
 - Changed: Improved and automated login process for Netatmo devices (to obtain client ID and Password with user selectable scopes)
+- Implemented: Support for Honeywell Series 5/PIR
 - Changed: Google Firebase Cloud Messaging (FCM) alternative way to pass a Domoticz device ID through the extra field ('|Device=<devidx>')
 - Changed: dzVents, possible to pass a Domoticz device ID in notify() through the extra parameter
 - Changed: Energy Dashboard now also available for non-admin users
--- a/History.txt
+++ b/History.txt
@@ -1,4 +1,5 @@
 Version 2024.x
+- Changed: Improved and automated login process for Netatmo devices (to obtain client ID and Password with user selectable scopes)
 - Changed: Google Firebase Cloud Messaging (FCM) alternative way to pass a Domoticz device ID through the extra field ('|Device=<devidx>')
 - Changed: dzVents, possible to pass a Domoticz device ID in notify() through the extra parameter
 - Changed: Energy Dashboard now also available for non-admin users
--- a/History.txt
+++ b/History.txt
@@ -1,4 +1,6 @@
 Version 2024.x
+- Changed: Improved and automated login process for Netatmo devices (to obtain client ID and Password with user selectable scopes)
+- Implemented: Support for Honeywell Series 5/PIR
 - Changed: Google Firebase Cloud Messaging (FCM) alternative way to pass a Domoticz device ID through the extra field ('|Device=<devidx>')
 - Changed: dzVents, possible to pass a Domoticz device ID in notify() through the extra parameter
 - Changed: Energy Dashboard now also available for non-admin users
--- a/hardware/Hardware.html
+++ b/hardware/Hardware.html
@@ -1,0 +1,1824 @@
+<div id="hardwarecontent" class="container">
+</div>
+
+<div id="opentherm" style="display:none;">
+	<center>
+		<form id="openthermform">
+			<br>
+			<center>
+				<table border="0" cellpadding="10" cellspacing="10">
+					<tr>
+						<td align="right" style="width:310px">
+							<span data-i18n="Outside Temperature Sensor">Outside Temperature Sensor</span>:
+						</td>
+						<td>
+							<select id="combooutsidesensor" name="combooutsidesensor" style="width:300px" class="combobox ui-corner-all"></select>
+						</td>
+					</tr>
+					<tr>
+						<td align="right">
+							<span data-i18n="Command">Command</span>:
+						</td>
+						<td>
+							<input type="input" id="otgwcmnd" name="otgwcmnd" style="width: 120px; padding: .2em;" class="text ui-widget-content ui-corner-all">  <button data-i18n="Send" class="btn btn-info" id="buttonsendotgwcmnd">Send</button>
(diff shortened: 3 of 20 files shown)


gizmocuz:
Thank you for this PR.
Please do not include the 'extern' folder in your commit

There seems to be way to many values for the scope, can this not be limited to 2 or 3?

LoukKamp:
Sorry about those 'Extern' files (I was just happy to get rid of those files in the list from files to be committed ;) ); what can I do about it; shell I redo the PR? I don't seem to be able to remove them.

No, these are the choices a user has, depending on the devices the user owns. In addition, we promised Netatmo to keep the scope as small as possible. Maybe we can reduce this by bundling read, write, and access scopes for the same device, but that would take a quite complex addition and time; I need to think about this and discuss this with @JurgenLB.

gizmocuz:
Not sure why you have included the extern folder, it is also in the .gitignore file
If you are using windows, you could have a look at using tortoisegit to make things a bit easier

LoukKamp:
No, I'm using my PI 4, because that is the only system I own running Linux. For editing, I'm using Nano. I'm a complete dummy, when it comes to Linux; sorry about that...

The reason why I included the _extern_ (folder) files, is that I now use '_git commit -a_' to commit. Before, I used _git add_ before a 'commit' without the '-a' option, but that didn't work for the _extern_ files. I also tried to use git restore on those files, but that didn't resolve anything too. 

Maybe this is caused by adding a global function to _hardware/Netatmo.h_. Strange to define a function global when it is only used locally, but this was done for all other local functions too, so I didn't want to break that tradition (partly, because I didn't understand why this was done this way and, it was my first change in this repository).

LoukKamp:
I found a simple way to reduce the number of scope options to about 13 for the moment (for 13 different Netatmo devices). I defined local scope names for this that we can use in the front- and back-end and store them in the database. When I need them for communication with Netatmo, I expand the local names to a set of the original scope names via a translation table. In this way we only need to do a one way translation, making things a lot easier.

When you agree to this, I'll commit the changes, and we can make a lot of Netatmo users happy with the next beta. 

Yesterday, we had a power outage (again), and this time I didn't need to aquire a new authotization code from Netatmo to get new tokens: It worked without any intervention when the power came back!

LoukKamp:
@gizmocuz : Are you waiting for my to do something about those "extern" files, maybe?

In that case, I think it's better to start all over (again), but then I will lose all steps in between (the edits with white space only), which makes reviewing quite a job.

LoukKamp:
@JurgenLB: Do you agree on this combining of scopes? Note that the user won't be able to select his own choices: It is all or nothing per device type. E.g. when he wants to have his doorbell to be read only, we need to make an extra scope for this.

I don't see this as a big issue, for the moment.

JurgenLB:
I also don't see a problem if the scopes are combined.
This shall bee documented so that the user knows what he gets.

Then when selected for example 'Presence'
The scopes are ;
'read_presence, write_presence, access_presence'

Kind regards

gizmocuz:
@LoukKamp , yeah the 'extern' folder needs to be excluded... you can probably do this with a git command
And I also need a 'go' from you to merge this PR when it is ready

I am totally for simplifying the scope, indeed like the example from Jurgen, maybe checkboxes what equipment your have
- temperature
- presence (PIR)
- ...
- ...

LoukKamp:
Ok, I will exclude them for future commits, what do you want me to do with the current PR? Leave it this way, or start a new PR. In that case, I will save my changed files reset to HEAD, and replace the changed files.

At this moment, I have the following layout:

afbeelding


When you really want check boxes in the combo select box, I can do this later.

Edit: Little correction in  the combo box options: removed READ from several items

gizmocuz:
Looking great! Is it an idea to remove the access rights Read/Write/Access words?
So, you just have  Weather Station, Thermostat, Camera etc?
It will confuse the user

gizmocuz:
We can't merge this PR when the 'extern' folder is also part of the changes. I think you should be able to exclude it 
Not sure how it became part of the changes as it is excluded in the **.gitignore** file

LoukKamp:
> Looking great! Is it an idea to remove the access rights Read/Write/Access words? So, you just have Weather Station, Thermostat, Camera etc? It will confuse the user

I was thinking of making the user aware about the possibilities/consequences of selecting an item and keep the possibility open to let the user choose for e.g. a read only version of his camera's. Maybe not that important right now?

But when @JurgenLB agrees, I'll take it out, so we can merge this PR ASAP (the user will benefit more from the other feature of this PR).

LoukKamp:
> We can't merge this PR when the 'extern' folder is also part of the changes. I think you should be able to exclude it Not sure how it became part of the changes as it is excluded in the **.gitignore** file

Ok, that's clear: I will work on this today.

Is there a way to get them out of this PR, or do I need to redo this in a new PR?

LoukKamp:
> We can't merge this PR when the 'extern' folder is also part of the changes. I think you should be able to exclude it Not sure how it became part of the changes as it is excluded in the **.gitignore** file

Yes, indeed it is present in this file, but as "/extern/"; I'm absolute not familiar with the format of the file, but shouldn't this be "extern/" (without the leading slash)?

JurgenLB:
> > Looking great! Is it an idea to remove the access rights Read/Write/Access words? So, you just have Weather Station, Thermostat, Camera etc? It will confuse the user
> 
> I was thinking of making the user aware about the possibilities/consequences of selecting an item and keep the possibility open to let the user choose for e.g. a read only version of his camera's. Maybe not that important right now?
> 
> But when @JurgenLB agrees, I'll take it out, so we can merge this PR ASAP (the user will benefit more from the other feature of this PR).

When @gizmocuz find its better whiteout the read/write information I'm OK with that.

I was just testing this on my setup and can you change this line;
`Press and hold the control key (Windows) or the Command key (Apple) to select multiple scope items. Please select only those scope items that are approiate for your devives.`

now it is 1 long line, I think it is better for having this in 2 lines in case someone is using a small screen.

I have made a new setup for testing but when I fill in Client ID & Client Secret, then selecting the scopes

it shows `Tokens are invalid or expired; please login to obtain new tokens!`
but when I press LOGIN -> There is nothing happening 
Also in the log file white `sudo tail -f  /var/log/domoticz.log` there is no entry.

and when I press ADD there is a message `Plese login before adding a new Netatmo device`

I think the we must first ADD and then LOGIN
so the block for the ADD button is not correct I think

Kind regards

JurgenLB:
I believe you have mixt-up some off the files
`hardware/Hardware.html
hardware/Hardware.js
www/app/hardware/Netatmo.cpp
www/app/hardware/Netatmo.h`

where non existing before.
can you check this.

LoukKamp:
> I believe you have mixt-up some off the files `hardware/Hardware.html hardware/Hardware.js www/app/hardware/Netatmo.cpp www/app/hardware/Netatmo.h`
> 
> where non existing before. can you check this.

Yes, you are probably right; I'm working on this (may also explain the issue in your previous post)

LoukKamp:
In the meanwhile, I opened a new pull request  (~V2 of this request) to get rid of the extern trees.

Sadly enough, I made a mistake while restoring the changed files from a save storage outside the repository, so that messed up this PR. 

However, I corrected this mistake and all files are now up-to-date and in the right places. 

So this PR can be closed.

LoukKamp:
This PR is replaced by [V2 of this PR](https://github.com/domoticz/domoticz/pull/6148)

PR #6149, opened 2024-08-29
New Netatmo-API v2

### LINKED ISSUE #6144: Battery levels of Netatmo modules are displayed incorrectly (much too wide)
The battery levels of additional Netatmo inside modules are displayed much too wide:

afbeelding

The levels of other parameters with battery levels include are displayed well.

Side issue: It seems that these values are only updated at the first connection, but this may not be a problem of Domoticz (not sure yet)

JurgenLB:
I dit a rewrite off my code.
With my setup the devices are working with multiple homes in my account.

The devices in my Home;
- Weather station - Main unit, Indoor sensor, outdoor sensor, wind sensor, rain sensor
- Netatmo Thermostat
- Netatmo Valves
- Homecoach
- Indoor Camera
- Doortags
- Presence (Outdoor Camera)
- Smoke detector
- Smart Co2 detector

I changed the code so all devices using the 'UpdateValueInt' , only the selector switch is still on 'SendSelectorSwitch'
the code also declares first 'Private' and then 'Public' so the Public-function can use the privates in declaration.

I have some `SaveJson2Disk` in the code so when testing the other scopes the output off the JSON is possible to share
and the code can be adapted.

Kind regards

jurgen

@track53 the bubendorff scope is now also possible but Not Tested, try with caution in a separated setup.

### DIFF
--- a/hardware/Netatmo.cpp
+++ b/hardware/Netatmo.cpp
@@ -1741,21 +1741,21 @@ bool CNetatmo::ParseDashboard(const Json::Value& root, const int DevIdx, const i
 		if (!root["time_utc"].empty())
 			tNetatmoLastUpdate = root["time_utc"].asUInt();
 		Debug(DEBUG_HARDWARE, "Module [%s] last update = %s", name.c_str(), ctime(&tNetatmoLastUpdate));
-		// check if Netatmo data was updated in the past 10 mins (+1 min for sync time lags)... if not means sensors failed to send to cloud
+		// check if Netatmo data was updated in the past NETAMO_POLL_INTERVALL (+1 min for sync time lags)... if not means sensors failed to send to cloud
 		int Interval = NETAMO_POLL_INTERVALL + 60;
 		Debug(DEBUG_HARDWARE, "Module [%s] Interval = %d", name.c_str(), Interval);
 		if (tNetatmoLastUpdate > (tNow - Interval))
 		{
 			if (!m_bNetatmoRefreshed[ID])
 			{
-				Log(LOG_STATUS, "cloud data for module [%s] is now updated again", name.c_str());
+				Log(LOG_STATUS, "cloud Dashboard for module [%s] is now updated again", name.c_str());
 				m_bNetatmoRefreshed[ID] = true;
 			}
 		}
 		else
 		{
 			if (m_bNetatmoRefreshed[ID])
-				Log(LOG_ERROR, "cloud data for module [%s] no longer updated (module possibly disconnected)", name.c_str());
+				Log(LOG_ERROR, "cloud Dashboard for module [%s] no longer updated (module possibly disconnected)", name.c_str());
 			m_bNetatmoRefreshed[ID] = false;
 			return false;
 		}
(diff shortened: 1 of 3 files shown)


gizmocuz:
Thank you for the PR, could you please add the class declaration,functions back on top again and variables at the bottom?

JurgenLB:
it was because off error message

/hardware/Netatmo.h:20:36: error: ‘_eNetatmoType’ does not name a type; did you mean ‘_eMeterType’?

void Get_Respons_API(const _eNetatmoType& NType, std::string& sResult, std::string& home_id , bool& bRet, Json::Value& root );


this is the reason that I put the public declaration after the private

kind regards

JurgenLB:
just tested and when I put
`enum _eNetatmoType`
also in the public declaration then the code also works.
I'm just not convinced that it is a good situation.
I thought it is best to keep the public declaration as small as possible.

But if you want I can make the change.

kind regards

gizmocuz:
In all code (and good practice), we declare the constructor/destructor of the class at the top of the class.
You can see this if you open any other header file.
You can always add public/private/protected on any line that you like.
No problem having private structures/enums declared at the very top, but from then on, the constructor/destructor, public functions, private functions, and then member variables (and good practice to start them with m_ so you know in the class it's a member variable and not a private function variable)

JurgenLB:
because LoukKamp did adapt my code this is to restore the desired format off public function before local variables.

JurgenLB:
my changes use `updatevalueint` so they show up as new devices.
I can't keep the old ID because the old way uses a conversion to HEX  `DeviceID=='%08X' NodeID`  and the new code only support INT

do you wan't to keep the Old Netatmo for existing users, so they can keep their history
and for the New API set a Netatmo_V2 Hardware ?

JurgenLB:
I think this is ready for now.

gizmocuz:
Thank you for this PR and all work you put into it!!!

track53:
Hi JurgenLB,
Today my domoticz was down I should reboot and my netmo dosen't work again.
I don't understand how to install your netatmo plugin version .... It's
possible with latest domoticz stable version ?

Thx

JurgenLB:
Hello,
My work is not in the Stable Version yet.
If your setup is on a Linux PC, you can build the latest version from Github;
```
cd domoticz
git pull https://github.com/domoticz/domoticz.git
cmake -DCMAKE_BUILD_TYPE=Release CMakeLists.txt
make
sudo nano /etc/systemd/system/domoticz.service
```

Make Changes to comply with your setup!
And save. (ctrl + x)

```
sudo systemctl daemon-reload
sudo systemctl enable domoticz.service

sudo service domoticz stop
sudo service domoticz start
```

track53:
Ok, it's not possible update only plugin netatmo. :(

nipsgur:
With the new version 16371 the rain sensor showed up as new sensor and with a changed subtyp from WWW to TFA.

Is there some workaround to get the historical data  from old sensor to the new sensor within domoticz, when the typ has change?

(I guess I have to really learn SQL better, always feel uncomfortable manipulating the domoticz db)

gizmocuz:
> With the new version 16371 the rain sensor showed up as new sensor and with a changed subtyp from WWW to TFA.
> 
> Is there some workaround to get the historical data from old sensor to the new sensor within domoticz, when the typ has change?
> 
> (I guess I have to really learn SQL better, always feel uncomfortable manipulating the domoticz db)

Can't you press 'Edit' on the old sensor, and choose 'Replace' and select the new one?

nipsgur:
> > With the new version 16371 the rain sensor showed up as new sensor and with a changed subtyp from WWW to TFA.
> > Is there some workaround to get the historical data from old sensor to the new sensor within domoticz, when the typ has change?
> > (I guess I have to really learn SQL better, always feel uncomfortable manipulating the domoticz db)
> 
> Can't you press 'Edit' on the old sensor, and choose 'Replace' and select the new one?

That is what I thought
I added the new one but on the old it says: "No devices to Transfer too!"

nipsgur:
Ok, just changed the subtype from WWW to TFA in the domoticz.db and now I could replace the devices in the web interface.

Is there a full list, which numbers in the db correspond to which description in the web interface , or would I have to look into the src code?

gizmocuz:
No, most is based on type/subtype
But in the upcoming beta version the subtype is now ignored for Rain types

PR #6153, opened 2024-09-03
Fix refresh token

mce35:
Fix sha256raw to return the full (32 bytes) SHA256 of the input instead of truncating it at the first null byte. In the case where the 1st byte of the SHA256 is 0, it is returning an empty string and the refresh token used for authentication is empty, making it unusable and requiring a new authentication.

### DIFF
--- a/main/Helper.cpp
+++ b/main/Helper.cpp
@@ -1645,7 +1645,7 @@ std::string sha256raw(const std::string &input)
 {
     unsigned char digest[33] = {0};
     SHA256((const unsigned char *)input.c_str(), input.length(), digest);
-	return std::string((const char *)digest, 32);
+    return std::string((const char *)digest, 32);
 }
 
 #ifdef _WIN32
--- a/main/Helper.cpp
+++ b/main/Helper.cpp
@@ -1645,7 +1645,7 @@ std::string sha256raw(const std::string &input)
 {
     unsigned char digest[33] = {0};
     SHA256((const unsigned char *)input.c_str(), input.length(), digest);
-	return std::string((const char *)digest);
+	return std::string((const char *)digest, 32);
 }
 
 #ifdef _WIN32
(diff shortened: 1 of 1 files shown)


gizmocuz:
Wouldn't it return a SHA string? Or when it has an error 0x0?
How would it return 0x00 at the beginning?, all std::string(...) does is convert a char array to std string

mce35:
> Wouldn't it return a SHA string?

SHA256 computes the SHA256 of the buffer and puts the result in the output buffer (`digest`). The result is a binary buffer of 32 bytes, not a string, and it can contain bytes with a value of 0. For example you can try hashing ` AAA` , it contains a 0 bytes at position 22.

```c++
include <iostream>
#include <openssl/sha.h>

std::string sha256raw(const std::string &input)
{
  unsigned char digest[33] = {0};
  SHA256((const unsigned char *)input.c_str(), input.length(), digest);
  return std::string((const char *)digest);
}

int main()
{
  std::string t1(1, 'A');
  std::string hash = sha256raw(t1);
  std::cout << "size of SHA256(\"A\"): " << hash.size() << std::endl;
  std::string t3(3, 'A');
  hash = sha256raw(t3);
  std::cout << "size of SHA256(\"AAA\"): " << hash.size() << std::endl;

  return 0;
}
```

```
$ g++ -Wall -Werror /tmp/t.cpp -o /tmp/t2 -lssl -lcrypto && /tmp/t2 
size of SHA256("A") 32
size of SHA256("AAA") 21
```

> Or when it has an error 0x0?

SHA256 cannot fail

> How would it return 0x00 at the beginning?

It depends on the buffer to hash, I was not able to quickly find an example but when refreshing token during a few hours it happens.

> all std::string(...) does is convert a char array to std string

When length is not specified, it stops at the first null byte, and in this case it could stop at the first byte resulting in an empty string.

gizmocuz:
Thanks again!!! Going to build new beta's now

mce35:
FYI I noticed this issue with dashticz, it was asking to re-authenticate almost each day when using oauth.

kiddigital:
> FYI I noticed this issue with dashticz, it was asking to re-authenticate almost each day when using oauth.

Thank you and nice catch 👍

PR #6154, opened 2024-09-03
Make is_number more robust

mce35:
Check number of dots in is_number to return false in the case of an IP address (so that MQTT sensors for IP addresses are created as text sensors)


gizmocuz:
Thank you for this PR!

Can't you assign your device a fixed IP address via your internal router? I can't see the point of having IP addresses in Domoticz?
The overview is in your router's DHCP leases overview
It is also a sensor that does not change, flooding the database and making the system slower

Domoticz even tries to filter out most of these devices (for instance when the object_id contains _address , _ssid, etc)

mce35:
> Can't you assign your device a fixed IP address via your internal router? I can't see the point of having IP addresses in Domoticz?
>  The overview is in your router's DHCP leases overview
>  It is also a sensor that does not change, flooding the database and making the system slower

Yes I agree with you and I don't really care about having the IP in domoticz, but it is added by some sensors using MQTT and the current behavior is to add a numeric sensor which truncates the IP address, which is misleading.

Anyway I though it might be worth adding this check in `is_number` as it could be useful in some other circumstances maybe.

> Domoticz even tries to filter out most of these devices (for instance when the object_id contains _address , _ssid, etc)

Yes and I have a sensor with `alarm` in its name and it was not appearing at all in domoticz, it took me sometime to figure out that this keyword was actually filtered out by domoticz, I guess it should be documented in the [MQTT wiki](https://www.domoticz.com/wiki/MQTT)

gizmocuz:
Could you give me the config topic/payload for both the IP and the alarm sensor?

mce35:
Sorry for the delay. For the IP:

config (topic `homeassistant/sensor/ecodan-heatpump/esp_ip/config`)
```
{"name":"ESP IP","ic":"mdi:ip","ent_cat":"diagnostic","stat_t":"ecodan-heatpump/sensor/esp_ip/state","avty_t":"ecodan-heatpump/status","uniq_id":"806599fd673c-wifiinfo-ip","dev":{"ids":"806599fd673c","name":"Ecodan heatpump","sw":"2024.8.1 (Sep 10 2024, 20:03:56)","mdl":"lolin_s2_mini","mf":"Espressif","cns":[["mac","806599fd673c"]]}}
``` 

payload (topic `ecodan-heatpump/sensor/esp_ip/state`)
```
192.168.6.103
```

(This is from the esphome ecodan component from [here](https://github.com/gekkekoe/esphome-ecodan-hp/blob/022aeb2eda2af9c0eea90cede66e1e164a633c27/confs/base.yaml#L106) where the default config is publishing the IP address)

Regarding the alarm sensors, it is a custom sensor based on esphome and I tried to include `alarm` in its name, so the topic names contains `alarm`, which is filtered out by domoticz. After realizing this (but it took me some time and reading domoticz code), I just changed the name of the device in the esphome config :smile:

gizmocuz:
Thank you for this. How about I check if the icon is an IP, and block this sensor?

kiddigital:
My 2 cents..
In general extending the `is_number` function to check for multiple dots is a sane check. It is not a number. So PR is valid I think.

The reason though for this change might not be the right reason and needs a different/better solution to handle the problem/symptom.

kiddigital:
@mce35 , @gizmocuz 
Looking at this PR again, checking for only 1 `.` (dot) and not more is a valid check.
But maybe the same should be applied to the `-` sign? Actually it should only exists at the first position and not anywhere else.
And ` ` (space) should only exist at the beginning or end, not in the middle. So better to first `trim()` the input and then not allow ` ` anymore?
And the 'terminating' `0x00` should only be at the last position.
Maybe improve this PR to make the function even more robust?

gizmocuz:
I think specifying a 'unit' (text) in the config payload also solves this issue

gizmocuz:
I think specifying a 'unit' (text) in the config payload also solves this issue

mce35:
Hi @gizmocuz @kiddigital sorry for the late reply, I modified is_number to perform more checks on the input string:

- Allow whitespaces only at the begining/end
- Allow a single optional sign (+/-)
- Allow a single dot
- Allow a single terminating NULL byte

Let me know if you have any comment.

gizmocuz:
Thanks for the changes.

My 2 cents:

- This function is called from various places, could be called very frequently.
- This function has now become slow... much more checks than previously
- The above problem (with IP address, which I think should not be displayed in Domoticz, but that's besides the issue), can be solved to specify a unit (Text), so it will not call this function

mce35:
>     * This function is called from various places, could be called very frequently.
> 
>     * This function has now become slow... much more checks than previously

I don't think it is significantly slower than the original function, it only goes through the string once as previously. Depending on the input it can do a few more checks or a few less. And I don't think this function is called often enough to make any visible difference. And the additional checks are needed to really check that the input is an actual valid number.

>     * The above problem (with IP address, which I think should not be displayed in Domoticz, but that's besides the issue), can be solved to specify a unit (Text), so it will not call this function

I agree with you, the IP is of no interest, my point was more to solve the general issue related to is_number which may cause other useful sensors to not be displayed correctly.

gizmocuz:
This function is called and designed for variable values like yes/no/true/false, and numbers
Your initial issue post is about IP addresses.
First, I am no fan of having these static devices being created at all in Domoticz, write them down, or name your devices correct, and use a DHCP manager aka your your router for this info
If you do want these to be shown, use the text unit

This function is called a lot... Specially if you have a lot of nodes that generates lots of devices

For this reason, I am not a fan of changing this function 

As this issue is active quite a while, with no backup/support or need for by users, I will close it as the solution to your specific.problem can be solved with the provided solutions above

Sorry!

Issue #6036, opened 2024-03-01
Netatmo plugin -- Auth update  May 29, 2024


Phil353556:
Hi, today everything is working perfectly with Netatmo, thank you so much.
However as a netatmo customer, I just received this mail, I copy paste, fyi,  I don't know if it is already taken into account in the plugin or not. Thank you.

----
Authentication update (second notice)

This is a reminder regarding our authentication update following the release initially planned for last fall which we decided to postpone in order to give our users more time.
 
Dear Netatmo developer,

Today, when you refresh an Access Token using the associated endpoint https://api.netatmo.com/oauth2/token, the Netatmo servers respond with a pair of tokens: an Access Token and a Refresh Token.

If the previous Access Token is still valid, the newly returned access token is identical but has an expiration time extended by 3 hours.

The Refresh Token is not renewed.

Starting May 29, 2024, this behavior will change to comply with the [OAuth2 Authorization Framework RFC recommendations (section 10.4)] (https://www.rfc-editor.org/rfc/rfc6749) and improve the security of our users' data.
When refreshing the tokens, the Access Token and Refresh Token will be different from the previous ones and the old tokens will be invalidated.

What does this mean for you?

If you already store the tokens returned by the API when refreshing your tokens, this change will not affect you.
If you do not update the tokens when refreshing them, your users will be logged out because the old tokens will be invalidated.
If for a given user, you have stored the same token in different places (for example: an application and its widget), you must have a way to synchronize them because the new token will be applied for the entire application.

Sincerely,

Legrand - Netatmo - Bticino
----

Mrhugues:
Hi, every body.

Same question , all run fine , but at 29/05/2024 what happen ?

waltervl:
Someone seems to work on this already: https://github.com/domoticz/domoticz/pull/6016

waltervl:
Should already be implemented by https://github.com/domoticz/domoticz/pull/5815

Phil353556:
Hi @waltervl 
Ok,
"When refreshing the tokens, the Access Token and Refresh Token will be different from the previous ones and **the old tokens will be invalidated**."
will see soon :-)
Thank you.

Salan54:
@Phil353556  It appears that your predictions were accurate. After following the method outlined on the Domoticz wiki to update my Netatmo configuration, everything seems to be functioning correctly for a few hours. However, I've noticed an error appearing in my logs around this time:

[...]

PR #6155, opened 2024-09-04
Patch device discovery

JurgenLB:
There was a check for live Cloud data but this does not function correctly for other setups
so I comment out the line `connected = false;`
`bool connected = true;` will continu be set so `if (connected)` will continu to run.

Sorry for this mistake.

kind regards


JurgenLB:
related to #6151 and #6036

gizmocuz:
Thank you!!

PR #6157, opened 2024-09-07
Update Netatmo

JurgenLB:
this is to adapt the problems other users have with the code.

### DIFF
--- a/hardware/Netatmo.cpp
+++ b/hardware/Netatmo.cpp
@@ -81,9 +81,7 @@ CNetatmo::CNetatmo(const int ID, const std::string& username, const std::string&
 	m_ActHome = 0;
 	Debug(DEBUG_HARDWARE, "Netatmo Actif Scopes %s ", m_scopes.c_str());
 
-	m_bPollWeatherData = false;
 	m_bPollWeatherData = (m_scopes.find("station_R") != std::string::npos);      //read_station
-	m_bPollHomecoachData = false;
 	m_bPollHomecoachData = (m_scopes.find("homecoach_R") != std::string::npos);  //read_homecoach
 
 	m_bPollHomeStatus = true;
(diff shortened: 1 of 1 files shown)


JurgenLB:
#6151
#6036

gizmocuz:
Thank you for your PR!

Isn't this much better for readability ?

```
m_bPollWeatherData = (m_scopes.find("read_station") != std::string::npos);
m_bPollHomecoachData = (m_scopes.find("read_homecoach") != std::string::npos);

```

gizmocuz:
Thanks! But you do not need to set the variables to false the line before, where the line after it is set to either true or false

JurgenLB:
With your proposal there is no log line possible.
`
m_bPollWeatherData = false;
m_bPollWeatherData = (m_scopes.find("station_R") != std::string::npos);      //read_station

m_bPollHomecoachData = false;
m_bPollHomecoachData = (m_scopes.find("homecoach_R") != std::string::npos);  //read_homecoach`

there is a placeholder used for the scopes from @LoukKamp, this is why I use the placeholder instead off the actual scope.
so this is ready.

JurgenLB:
Ok I thought that it needs a declaration in front.
I will change that.

gizmocuz:
If you like, you can make a variable m_bFirstRun = true, in the header
And when is true, you make some logging, and set the variable to false
Is this an idea?

JurgenLB:
I now have `Debug(DEBUG_HARDWARE, "Netatmo Actif Scopes %s ", m_scopes.c_str());`
so in debug-mode we can check this, is it necessary for the end user to known?
I think it can be useful but it is not necessary.
because the new HomeStatus does also include the weather station sensors.

I will later make a separated PR for extending the HomeStatus. Because this is Recently changed by Netatmo.

kind regards

gizmocuz:
Thank you once again!!!

PR #6159, opened 2024-09-09
missing image clock48_Off.png

ml17950:
Hi.

If you have a switch with tha clock icon and it ist Off, the image was missing.

Best regards


gizmocuz:
Thank you for the PR!!

PR #6161, opened 2024-09-10
Move the functional code of CMD_DoTransferDevice() 

LoukKamp:
From WebServerCmds to SQLHelper, so it can be used elsewhere too.

### DIFF
--- a/main/SQLHelper.cpp
+++ b/main/SQLHelper.cpp
@@ -10276,8 +10276,7 @@ void CSQLHelper::RefreshActualPrices()
 
 bool CSQLHelper::TransferDevice(const std::string& sOldIdx, const std::string& sNewIdx)
 {
-	std::vector<std::vector<std::string>> result;
-	result = m_sql.safe_query("SELECT HardwareID, OrgHardwareID, DeviceID, Unit, Type, SubType FROM DeviceStatus WHERE (ID == '%q')", sNewIdx.c_str());
+	auto result = m_sql.safe_query("SELECT HardwareID, OrgHardwareID, DeviceID, Unit, Type, SubType FROM DeviceStatus WHERE (ID == '%q')", sNewIdx.c_str());
 	if (result.empty())
 		return false;
 
--- a/main/SQLHelper.cpp
+++ b/main/SQLHelper.cpp
@@ -10276,11 +10276,8 @@ void CSQLHelper::RefreshActualPrices()
 
 bool CSQLHelper::TransferDevice(const std::string& sOldIdx, const std::string& sNewIdx)
 {
-	const int oldIdx = std::atoi(sOldIdx.c_str());
-	const int newIdx = std::atoi(sNewIdx.c_str());
-
 	std::vector<std::vector<std::string>> result;
-	result = m_sql.safe_query("SELECT HardwareID, OrgHardwareID, DeviceID, Unit, Type, SubType FROM DeviceStatus WHERE (ID == '%d')", newIdx);
+	result = m_sql.safe_query("SELECT HardwareID, OrgHardwareID, DeviceID, Unit, Type, SubType FROM DeviceStatus WHERE (ID == '%q')", sNewIdx.c_str());
 	if (result.empty())
 		return false;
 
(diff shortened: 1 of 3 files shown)


LoukKamp:
Ready to be reviewed and tested!

gizmocuz:
Is it possible to keep the changes to a minimum, also keep the %q instead of %d as these are 64bit indexes.

LoukKamp:
Of course; I'll change this.

LoukKamp:
Done and tested

gizmocuz:
Thank you!!

Once small request, could you declare 'result' as auto at the beginning of the function

`auto result = m_sql.safe_query("SE....`

LoukKamp:
I thought this was common practice, but now I see this was done before (using auto).

With the commit coming up, I also added a parameter bFourceAllDateCopied (default false). I need this to merge data from previous test sessions with older versions of the interface. Without this, I'm left with holes in my data.

gizmocuz:
Please do not change the transfer function as it is working as it should.

kiddigital:
> ...
> 
> With the commit coming up, I also added a parameter bFourceAllDateCopied (default false). I need this to merge data from previous test sessions with older versions of the interface. Without this, I'm left with holes in my data.

@LoukKamp , if you need some, temporary(?), additional code, it is better to create a new wrapper-function you can call which calls the `CMD_DoTransferDevice()` function AND performs the additional work. This way the original function remains the same for all existing callers and for this particular case you also have a separate function to call to do a) the same work and b) the additional work.

Do you also expect that in due time this migration code will be obsolete and can be removed? If so, please add that to the comments so people know it is depricated code.

LoukKamp:
I see your point, however the fix was minimal:

```
diff --git a/main/SQLHelper.cpp b/main/SQLHelper.cpp
index 11844f5e5..e01170e6d 100644
--- a/main/SQLHelper.cpp
+++ b/main/SQLHelper.cpp
@@ -10274,10 +10274,9 @@ void CSQLHelper::RefreshActualPrices()
 }


-bool CSQLHelper::TransferDevice(const std::string& sOldIdx, const std::string& sNewIdx)
+bool CSQLHelper::TransferDevice(const std::string& sOldIdx, const std::string& sNewIdx, const bool bForceAllDataCopied)
 {
-       std::vector<std::vector<std::string>> result;
-       result = m_sql.safe_query("SELECT HardwareID, OrgHardwareID, DeviceID, Unit, Type, SubType FROM DeviceStatus WHERE (ID == '%q')", sNewIdx.c_str());
+       auto result = m_sql.safe_query("SELECT HardwareID, OrgHardwareID, DeviceID, Unit, Type, SubType FROM DeviceStatus WHERE (ID == '%q')", sNewIdx.c_str());
        if (result.empty())
                return false;

@@ -10295,6 +10294,9 @@ bool CSQLHelper::TransferDevice(const std::string& sOldIdx, const std::string& s

        std::string szLastOldDate = result[0][0];

**+       if (bForceAllDataCopied)
+               szLastOldDate = "1970-01-01 00:00:00";
+**
        _log.Log(LOG_STATUS, "Replace old device %s to new device %s from %s.", sOldIdx.c_str(), sNewIdx.c_str(), szLastOldDate.c_str());

        m_sql.safe_query("UPDATE DeviceStatus SET HardwareID = %d, OrgHardwareID = %d, DeviceID = '%q', Unit = %d, Type = %d, SubType = %d WHERE ID == '%q'",
```

I put this change on hold, because I found that the latest version of the new Netatmo interface adds data to the new devices as well as the old devices, for some curious reason. For this reason, I can't use this change to copy all data i.s.o. just the data since the last update of the old device, because this will lead to duplicate data rows after the merge.

So for the early adaptors among us, who were helpful enough to test the first versions, I need to find another solution to merge the devices.

For all other users, migrating to the new Netatmo interface (by updating to the next master release), the function TransferDevice(), called from Netatmo.cpp will do the job.

For now, I will withdraw the change with this extra parameter/option.

LoukKamp:
Ready to merge, I guess.

LoukKamp:
Just to be sure: The change, mentioned some posts ago, is NOT part of this PR anymore; I removed this code and just changed the definition of result to auto.

So, as far as it concerns me, it is ready to merge

LoukKamp:
@gizmocuz: I noticed that this PR was not merged; is there still a problem to be solved?

I want to do a PR that includes my migration tool for Netatmo devices, so we can discuss the code and the options.

But it will also include the changes from this PR, as long as this PR is not merged.

gizmocuz:
Sorry, did not had time to review yet (busy weeks)
All looks perfect, thanks for this PR!

LoukKamp:
No problem; thanks too!

kiddigital:
@LoukKamp , sorry for being late to this party but I see in this PR you first set the API return message to OK and then call the new SQL helper function.

This function returns a boolean but you don't check the return value and set the return message accordingly.

Better to initialize the return message to ER, check to see if the function completed successfully and then set it to OK.

PR #6162, opened 2024-09-10
keys are all lowercase

JurgenLB:
this PR is to adres a typo in ten keys and some small code fix and detailing the temperature

I tested this wil on my test setup.

### DIFF
--- a/hardware/Netatmo.cpp
+++ b/hardware/Netatmo.cpp
@@ -2074,7 +2074,7 @@ bool CNetatmo::ParseHomeStatus(const std::string& sResult, Json::Value& root, st
 				{
 					float room_measured = room["therm_measured_temperature"].asFloat();
 					std::stringstream room_str;
-					room_str << std::setprecision(2) << room_measured;
+					room_str << std::fixed << std::setprecision(2) << room_measured;
 					room_str >> room_temp;
 					m_Room_Temp[roomNetatmoID] = room_temp;
 					//Debug(DEBUG_HARDWARE, "room %s %stemp = %s °C - %f ", roomName.c_str(), roomNetatmoID.c_str(), room_temp.c_str(), room_measured);
@@ -2165,11 +2165,11 @@ bool CNetatmo::ParseHomeStatus(const std::string& sResult, Json::Value& root, st
 				float Temp = 0, SP_temp = 0;
 				int hum = 0;
 				int baro = 0;
-				int abs_baro = 0;
+				float abs_baro = 0;
 				int co2 = 0;
 				int rain = 0;
-				int rain_1 = 0;
-				int rain_24 = 0;
+				float rain_1 = 0;
+				float rain_24 = 0;
 				int sound = 0;
 				std::stringstream t_str;
 
@@ -2177,7 +2177,7 @@ bool CNetatmo::ParseHomeStatus(const std::string& sResult, Json::Value& root, st
 				float wind_strength = 0;
 				float wind_gust = 0;
 				float wind_gust_angle = 0;
-				int Temp_outdoor = 0;
+				std::string Temp_outdoor = "0";
 				int WindChill_c = 0;
 
 				//uint64_t DeviceRowIdx;
@@ -2448,42 +2448,46 @@ bool CNetatmo::ParseHomeStatus(const std::string& sResult, Json::Value& root, st
 					if (!module["temperature"].empty())
 					{
 						bHaveTemp = true;
+						std::stringstream bt;
 						Temp = module["temperature"].asFloat();
 						t_str << std::setprecision(2) << Temp;
 						temp = static_cast<int>(Temp);
 						if(type == "NAModule1")
-							Temp_outdoor = temp;
+						{
+							bt << std::setprecision(2) << Temp;
+							Temp_outdoor = bt.str();
+						}
 						//Debug(DEBUG_HARDWARE, "HomeStatus Module Temperature [%s]", t_str.str().c_str());
 					}
 					if (!module["co2"].empty())
 					{
 						bHaveCO2 = true;
-						co2 = module["CO2"].asInt();
+						co2 = module["co2"].asInt();
 						//Debug(DEBUG_HARDWARE, "HomeStatus Module CO2 [%d]", co2);
 					}
 					if (!module["humidity"].empty())
 					{
 						bHaveHum = true;
-						hum = module["Humidity"].asInt();
+						hum = module["humidity"].asInt();
 						//Debug(DEBUG_HARDWARE, "HomeStatus Module hum [%d]", hum);
 					}
 					if (!module["noise"].empty())
 					{
 						bHaveSound = true;
-						sound = module["Noise"].asInt();
+						sound = module["noise"].asInt();
 						//Debug(DEBUG_HARDWARE, "HomeStatus Module Noise [%d]", sound);
 					}
 					if (!module["pressure"].empty())
 					{
 						bHaveBaro = true;
-						baro = module["Pressure"].asInt();
+						baro = module["pressure"].asInt();
 						//Debug(DEBUG_HARDWARE, "HomeStatus Module Pressure [%d]", baro);
 					}
 					if (!module["absolute_pressure"].empty())
 					{
 						bHaveBaro = true;
-						abs_baro = module["Pressure"].asInt();
-						//Debug(DEBUG_HARDWARE, "HomeStatus Module Pressure [%d]", baro);
+						abs_baro = module["absolute_pressure"].asFloat();
+						//Debug(DEBUG_HARDWARE, "HomeStatus Module Absolute Pressure [%f]", baro);
 					}
 					if (!module["rain"].empty())
 					{
@@ -2494,19 +2498,19 @@ bool CNetatmo::ParseHomeStatus(const std::string& sResult, Json::Value& root, st
 					if (!module["sum_rain_1"].empty())
 					{
 						bHaveRain = true;
-						rain_1 = module["sum_rain_1"].asInt();
-						//Debug(DEBUG_HARDWARE, "HomeStatus Module Rain [%d]", rain_1);
+						rain_1 = module["sum_rain_1"].asFloat();
+						//Debug(DEBUG_HARDWARE, "HomeStatus Module Rain [%f]", rain_1);
 					}
 					if (!module["sum_rain_24"].empty())
 					{
 						bHaveRain = true;
-						rain_24 = module["sum_rain_24"].asInt();
-						//Debug(DEBUG_HARDWARE, "HomeStatus Module Rain [%d]", rain_24);
+						rain_24 = module["sum_rain_24"].asFloat();
+						//Debug(DEBUG_HARDWARE, "HomeStatus Module Rain [%f]", rain_24);
 					}
 					if (!module["wind_strength"].empty())
 					{
 						bHaveWind = true;
-						wind_strength = module["wind_strength"].asFloat() / 3.6F;
+						wind_strength = module["wind_strength"].asFloat(); // / 3.6F;
 						//Debug(DEBUG_HARDWARE, "HomeStatus Module Wind strength [%d]", wind_strength);
 					}
 					if (!module["wind_angle"].empty())
@@ -2518,20 +2522,20 @@ bool CNetatmo::ParseHomeStatus(const std::string& sResult, Json::Value& root, st
 					if (!module["wind_gust"].empty())
 					{
 						bHaveWind = true;
-						wind_gust = module["wind_gust"].asFloat() / 3.6F;
+						wind_gust = module["wind_gust"].asFloat(); // / 3.6F;
 						//Debug(DEBUG_HARDWARE, "HomeStatus Module Wind gust [%d]", wind_gust);
 					}
 					if (!module["wind_gust_angle"].empty())
 					{
 						bHaveWind = true;
-						wind_gust_angle = module["wind_gust_angle"].asFloat() / 3.6F;
+						wind_gust_angle = module["wind_gust_angle"].asFloat(); // / 3.6F;
 						//Debug(DEBUG_HARDWARE, "HomeStatus Module Wind gust angle [%d]", wind_gust_angle);
 					}
 					//Data retrieved create / update appropriate domoticz devices
 					if (bHaveTemp && bHaveHum && bHaveBaro)
 					{
 						int nforecast = m_forecast_calculators[crcId].CalculateBaroForecast(Temp, baro); //float temp, double pressure
-						//Debug(DEBUG_HARDWARE, "%s name Temp & Hum & Baro %d [%s] %d %d %d - %d / %d ", Hardware_int, Hardware_ID.c_str(), name.c_str(), temp, hum, baro, batValue, rssiLevel);
+						//Debug(DEBUG_HARDWARE, "%d name Temp & Hum & Baro %s [%s] %f %d %d - %d / %d ", Hardware_int, Hardware_ID.c_str(), name.c_str(), Temp, hum, baro, mrf_status, batteryLevel);
 						// Humidity status: 0 - Normal, 1 - Comfort, 2 - Dry, 3 - Wet
 						const char status = '0';
 						std::stringstream r;
@@ -2546,7 +2550,7 @@ bool CNetatmo::ParseHomeStatus(const std::string& sResult, Json::Value& root, st
 						r << nforecast;
 
 						// sValue is string with values separated by semicolon: Temperature;Humidity;Humidity Status;Barometer;Forecast
-						//Debug(DEBUG_HARDWARE, "(%d) %s (%s) [%s] Temp & Humidity & Baro %s %s %d %d", Hardware_int, str_ID.c_str(), pchar_ID, name.c_str(), r.str().c_str(), m_Name.c_str(), rssiLevel, batValue);
+						//Debug(DEBUG_HARDWARE, "(%d) %s (%s) [%s] Temp & Humidity & Baro %s %s %d %d", Hardware_int, str_ID.c_str(), pchar_ID, name.c_str(), r.str().c_str(), m_Name.c_str(), mrf_status, batteryLevel);
 						UpdateValueInt(Hardware_int, ID.c_str(), 0, pTypeTEMP_HUM_BARO, sTypeTHBFloat, mrf_status, batteryLevel, '0', r.str().c_str(), moduleName,  0, m_Name);
 					}
 					else if (bHaveTemp && bHaveHum)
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thank you!

LoukKamp:
Thanks too.

I noticed that you changed the deviceID parameter to from decimal to hex. However, is it possible to force is to uppercase and fill with leading 0's?

In my test version (based on an older version of your work), I had the following:

```
        iStrID << deviceID;
        int iDeviceID;
        iStrID >> iDeviceID;
        std::stringstream strID;
        strID   << std::uppercase
                << std::setfill ('0')
                << std::hex
                << iDeviceID;
        std::string sDeviceID(strID.str());

```

JurgenLB:
I think where speaking not about the same thing.

This PR is because the Netatmo-API-respons from GetWeatherDetails has a leading uppercase in the Keys 
And the new API-respons from HomeStatusDetails has all lowercase in the keys.

I think you mean something like this;
`
std::stringstream hardware;

hardware << std::hex << std::setfill ('0') << std::uppercase << ID;

hardware >> str_ID;`

but what is the purpose for this ?

LoukKamp:
The purpose of this was to get the DeviceID in the DeviceStatus table all in uppercase HEX, like it was before, and for most other devices. I was using this conversion to uppercase also to be sure to get everything in the same case (upper) when testing the old devices for a match.

But shouldn't we read the statement from right to left? So it would be:

```
std::stringstream hardware;

hardware << std::setfill ('0') << std::uppercase << std::hex << ID;

hardware >> str_ID;`
```
B.t.w.: I don't think I'll need the leading zero's any more

JurgenLB:
I have tested this an with `moduleid << std::hex << std::setfill ('0') << std::uppercase << crcId;`
It generate DeviceId with Uppercase but they show up as unused devices!

Is it not better just use the lowercase HEX for the new devices and Uppercase for the Old Devices, 
because users already did change their scripts.
maybe it is also better to identify Old API devices and New API devices ?

kind regards.

JurgenLB:
@LoukKamp in Hardware.html you have
`

						<option value="station_R">Weather Station</option>

						<option value="thermostat_RW">Thermostat </option>

						<option value="camera_RWA">Indoor Camera</option>

						<option value="doorbell_RA">Doorbell *</option>

						<option value="presence_RWA">Presence</option>

						<option value="carbonmonoxidedetector_R">CO2 Detector</option>

						<option value="smokedetector_R">Smoke Detector</option>

						<option value="homecoach_R">Home Coach</option>

						<option value="magellan_RW">Magellan *</option>

						<option value="bubendorff_RW">Bubendorff *</option>

						<option value="smarther_RW">Smarther *</option>

						<option value="mx_RW">MX *</option>

						<option value="mhs1_RW">My Home Server 1 *</option>
`
and in Hardware.js

` 

				station_R :			'read_station',

				thermostat_RW :			'read_thermostat write_thermostat',

				camera_RWA :			'read_camera write_camera access_camera',

				doorbell_RA :			'read_doorbell access_doorbell',

				presence_RWA :			'read_presence write_presence access_presence',

				carbonmonoxidedetector_R :	'read_carbonmonoxidedetector',

				read_smokedetector_R :		'read_smokedetector',

				read_homecoach_R :		'read_homecoach',

				read_magellan_RW :		'read_magellan write_magellan',

				read_bubendorff_RW :		'read_bubendorff write_bubendorff',

				read_smarther_RW :		'read_smarther write_smarther',

				read_mx_RW :			'read_mx write_mx',

				read_mhs1_RW :			'read_mhs1 write_mhs1'
`

must they not all be the same placeholders (keys)?
Because when using the Homecoach I get;
`Error data ...  {"error":{"code":13,"message":"Application does not have the good scope rights"}}`

kind regards

LoukKamp:
Yes, you are right, they should, but they aren't...

I will correct this ASAP.

LoukKamp:
And done....

LoukKamp:
> I have tested this an with `moduleid << std::hex << std::setfill ('0') << std::uppercase << crcId;` It generate DeviceId with Uppercase but they show up as unused devices!
> 
> Is it not better ust use the lowercase HEX for the new devices and Uppercase for the Old Devices, because users already did change their scripts. maybe it is also better to identify Old API devices and New API devices ?
> 
> kind regards.

I wonder why users use the deviceID in their scripts; shouldn't they use the Domoticz ID from the DeviceSatus table?

Personally I don't see the use in a difference between old and new ID's, and I hate lower case letters in HEX codes, but I think it's your choice.

While checking the results of the migration code, I found another différance in the subtypes: My old (broken) Rain Sensor had a type of 85, and a subtype of 3. the new device has a subtype of 113; is this correct?

Is there also a new subtype for the wind meter? When it was still alive, it had a subtype of 49 (type was 86)

JurgenLB:
The temperatuur from '82 - 5'
,   to  '82 - 160'
The Wind sensor from '86 - 49'
,  to  '86 - 48'
The Rainsensor from '85 - 3'
, to  '85 - 113'

For the thermostat the '- Mode' is changed
, to 'Schedule'
and the '- setpoint' is changed
, to 'Heating Status'
But these Last 2 uses the same code as before, because I could not get the selector switch working correctly for the moment.

JurgenLB:
I think users use the IDX in their scripts but if the Hardware makes a new device from the new DeviceID then the IDX is also changed.

gizmocuz:
Correct, but if you click 'edit' on the old hardware and choose 'replace' , the old idx is kept

LoukKamp:
Yes, correct indeed, and that's what the migration code, I'm working on, will do for you.

However, because of bad WiFi circumstances here on my vacation address in France, I won't make much progress the next days.

Maybe, I'll commit a test/dummy version one of these days, to aquire some logging's from people who are willing to test with other Netatmo devices than the weather station I own.

One question: why do you add a leading space in the device name, like with "Bat. Lvl" and "Sig.+Bat. Lvl"; is this intended behavior?

JurgenLB:
I think you mean this;

`batName = " " + moduleName + " Bat. Lvl";`
`pName = " " + moduleName + " - Sig. + Bat. Lvl";`

yes this was intended, this is for the Log, some times there was no space between the text from the function and the name.
But with the "UpdateValueInt" function this is no longer necessary 

Then maybe with your migration code set the HEX to uppercase?

apparently I format a schedule-selectorswitch, so I going to use the Old Name for the selectorswitch.
but this is a bit tricky because this schedule-selector is not always the same

LoukKamp:
Yes, that was what I meant. I don't think the migration tool will depend on this, but I think you better remove the leading spaces if not required anymore.

If you don't have any objections, I leave the conversion to uppercase in the migration code.

I'm not sure if I can follow you on the selector switch, because I don't have a Netatmo thermostat and don't know how the old and new shedule-selectorswitch  looks like.

LoukKamp:
> Then maybe with your migration code set the HEX to uppercase?

Just to be sure: Do you mean that I will convert it to HEX and store it that way, or just convert it for own usage and store it in lowercase HEX?

Note: The migration code is part of UpdateValueInt(), where the new/existing devices are stored.

B.t.w. Do you know the new subtype for the wind meter? I assume this will change too, as the rain meter changed from 3 to 113. The type (85) and unit number (0) stayed the same.

LoukKamp:
While searching for a suitable test case for another patch, I found that multiple devices are updated with the same data at the same time...??? This does not only destroy my theory that I could merge data from old and new devices (created by the original Netatmo interface) and the new one (from @JurgenLB), because there should only one of them active at any time, so data should not overlap. Makes it impossible to merge data from multiple devices created during development and testing.

However when searching for the cause of this, I found the following code in SQLHelper.cpp that really worries me:

```
       if (bDeviceUsed)
        {
                m_mainworker.m_eventsystem.ProcessDevice(HardwareID, ulID, unit, devType, subType, signallevel, batterylevel, nValue, sValue);

                if (OrgHardwareID == 0)
                {
                        //Send to connected Sharing Users
                        m_mainworker.m_sharedserver.SendToAll(HardwareID, ulID, nullptr);
                }
        }
        return ulID;

```
This worries me, because of the use of the original hardware ID: 0 seems to have a special meaning. In the old Netatmo interface Devices had an original hardware ID of 0 for some reason, but in the new interface all the devices get a original hardware ID of -1.

Can someone explain why this have to be changed and what the impact might be?

gizmocuz:
No idea why it is -1, but it should be 0
The OrgHardwareID is used when using multiple Domoticz instances that is sharing devices between them.

Maybe it is best to let the user transfer the old devices to the new one? This is easy to do, and should cost only a few minutes?

JurgenLB:
I looked at my database and the "OrgHardwareID" are all 0 or positive numbers.
The one with 0 are the selectorswitch and the setpointsensor.
All other have positive numbers, but maybe this need some fine-tuning.
Where do you have negative numbers?

LoukKamp:
All (or almost) of the new Netatmo devices have -1 (all old devices have 0).

B.t.w.: My car broke down on our journey home from a short vacation in France and now we are stuck in Belgium...  Sorry for the delayed responses.

LoukKamp:
I checked on my database and all the devices created by the new interface have an OrgHardwareID of -1, except for the devices that have a type and subtype of 243 and 6 (Name  - Sig. + Bat. Lvl). I suspect that this device type is added later (after the setting of OrgHardwareID was corrected).

But Just tested with a new hardware device, and it created all devices with an OrgHardwareID  of -1:
```
ID    HardwareID  DeviceID  Name                                     OrgHardwareID  Type  SubType
----  ----------  --------  ---------------------------------------  -------------  ----  -------
1990  17          48a57370  Eetkamer/Keuken                          -1             82    160
1991  17          48a57370  Eetkamer/Keuken                          -1             249   1
1988  17          48a57370  Eetkamer/Keuken - MAC-adres              -1             243   19
1989  17          48a57370  Eetkamer/Keuken RF. Lvl                  -1             243   6
2003  17          48a57370   Eetkamer/Keuken Bat. Lvl                -1             243   31
2002  17          48a57370   Eetkamer/Keuken - Sig. + Bat. Lvl       -1             243   6
2004  17          48a57370  Eetkamer/Keuken - Bridge                 -1             243   19
1978  17          51c5678c  Binnen                                   -1             84    16
1980  17          51c5678c  Binnen                                   -1             243   24
1979  17          51c5678c  Binnen                                   -1             249   1
1976  17          51c5678c  Binnen - MAC-adres                       -1             243   19
1977  17          51c5678c  Binnen RF. Lvl                           -1             243   6
1995  17          7b2ba3cf  Regenmeter                               -1             85    113
1992  17          7b2ba3cf  Regenmeter - MAC-adres                   -1             243   19
1993  17          7b2ba3cf   Regenmeter Bat. Lvl                     -1             243   31
1994  17          7b2ba3cf  Regenmeter - Bridge                      -1             243   19
1983  17          87f710f5  Slaapkamer Oostzijde                     -1             82    160
1984  17          87f710f5  Slaapkamer Oostzijde                     -1             249   1
1981  17          87f710f5  Slaapkamer Oostzijde - MAC-adres         -1             243   19
1982  17          87f710f5  Slaapkamer Oostzijde RF. Lvl             -1             243   6
1997  17          87f710f5   Slaapkamer Oostzijde Bat. Lvl           -1             243   31
1996  17          87f710f5   Slaapkamer Oostzijde - Sig. + Bat. Lvl  -1             243   6
1998  17          87f710f5  Slaapkamer Oostzijde - Bridge            -1             243   19
1987  17          e221c362  Buiten - oost                            -1             82    160
1985  17          e221c362  Buiten - oost - MAC-adres                -1             243   19
1986  17          e221c362  Buiten - oost RF. Lvl                    -1             243   6
2000  17          e221c362   Buiten - oost Bat. Lvl                  -1             243   31
1999  17          e221c362   Buiten - oost - Sig. + Bat. Lvl         -1             243   6
2001  17          e221c362  Buiten - oost - Bridge                   -1             243   19
```

LoukKamp:
Just tested with the latest beta merges (there were a lot of changes in NetAtmo.cpp, but can't see what was actually changed since the former sync/pull to my local system), but the results were all the same (see previous post).

I also noted that all devices are updated twice every 5 minutes: once as part of parsing the Station Data and once as part of parsing the HomeStatus.

The messages about the cloud data being updated again, are also there (on each HomeStatus update)

JurgenLB:
Yes it is just recently that the WeatherStation is also accessible by HomeData/HomeStatus
This result in the double update/message you mention. But that is not a problem I think.

Within HomeStatus there is a block possible for creating/updating the sensor if their data is older than 16 min. but this is comment out for the moment. Now its only a Log-line, but it is possible to set this to debug.

for the 'OrgHardwareID' I'm looking for making it to 0.

Whit a lot off devices there is sometimes the Error; 'Error: Netatmo hardware (2) thread seems to have ended unexpectedly'
The code for this is;
`

        //Check hardware heartbeats
        for (const auto& pHardware : m_hardwaredevices)
        {
                if (!pHardware->m_bSkipReceiveCheck)
                {
                        //Check Thread Timeout
                        bool bDoCheck = (pHardware->HwdType != HTYPE_Dummy) && (pHardware->HwdType != HTYPE_EVOHOME_SCRIPT);
                        if (bDoCheck)
                        {
                                double diff = difftime(now, pHardware->m_LastHeartbeat);
                                //_log.Log(LOG_STATUS, "%d last checking  %.2lf seconds ago", iterator->first, dif);
                                if (diff > 60)
                                {
                                        result = m_sql.safe_query("SELECT Name FROM Hardware WHERE (ID='%d')", pHardware->m_HwdID);
                                        if (result.size() == 1)
                                        {
                                                std::vector<std::string> sd = result[0];
                                                _log.Log(LOG_ERROR, "%s hardware (%d) thread seems to have ended unexpectedly", sd[0].c_str(), pHardware->m_HwdID);
                                        }
                                }
                        }
`
but is it necessary 60 seconds or can this be changed to 90 seconds ?

LoukKamp:
As @gizmocuz said, this old hardware ID is only used in multiserver setups. As far as I understand, we don't use that with Netatmo devices, so just set it hard to 0, I would think.

I can't find the exact code from the previous post, but something like this exists in main/MainWorker.cpp. 

However, in both cases, this does not mean that the tread is crashed, but just that the thread wasn't able to update the heartbeat time in time (60 seconds).  In Netatmo.cpp this heartbeat field is updated every 12 seconds. I don't think there is any interactive communication in this thread, so as long as you don't use a debugger with active breakpoints, the only reasons I can think of is a long delay in getting the data from Netatmo (over internet), or a database lock.

JurgenLB:
yes I now have the OrgHardwareId hardcoded to 0 in my test setup, but this result in new unused devices with the same type and subtype.

so for the early adopters this is some trouble to re-merg their devices again. 
But it is not the same DeviceId as with the Old API.
I did read that OrgHardwareId as something different, so I think this need to be fixt.

LoukKamp:
I also found that the e.g. temperature logs are updated by UpdateTemperatureLog(). This function will update ALL temperature parameters with some restrictions on the type and subtype. The last update time is the current time, but the values come from DeviceStatus, so they may be outdated.

And now I wonder: Why does this cause just a view (temperature) devices to add to the logs and not all. And why doesn't this happen at all with the original version of the Netatmo interface? Or is this outdated or dead code perhaps?

LoukKamp:
To show my observations:

```
DeviceRowID  Temperature  Chill  Humidity  Barometer  DewPoint  SetPoint  Date
-----------  -----------  -----  --------  ---------  --------  --------  -------------------
<History removed>
14           16.7         0.0    76        10155      12.45     0.0       2024-09-24 22:05:00
14           16.6         0.0    76        10156      12.35     0.0       2024-09-24 22:10:00
14           16.6         0.0    76        10156      12.35     0.0       2024-09-24 22:15:00
14           16.6         0.0    76        10156      12.35     0.0       2024-09-24 22:20:00
14           16.6         0.0    76        10156      12.35     0.0       2024-09-24 22:25:00
14           16.6         0.0    76        10156      12.35     0.0       2024-09-24 22:30:01
14           16.6         0.0    76        10156      12.35     0.0       2024-09-24 22:35:00
14           16.6         0.0    76        10156      12.35     0.0       2024-09-24 22:40:00
14           16.6         0.0    76        10156      12.35     0.0       2024-09-24 22:45:00
14           16.6         0.0    76        10156      12.35     0.0       2024-09-24 22:50:00
sqlite> select * from temperature where deviceRowId=1814;
sqlite> select * from temperature where deviceRowId=1707;
sqlite> select * from temperature where deviceRowId=1912;
DeviceRowID  Temperature  Chill  Humidity  Barometer  DewPoint  SetPoint  Date
-----------  -----------  -----  --------  ---------  --------  --------  -------------------
1912         16.6         0.0    76        10150      12.35     0.0       2024-09-24 22:20:00
1912         16.5         0.0    76        10150      12.25     0.0       2024-09-24 22:25:00
1912         16.5         0.0    76        10150      12.25     0.0       2024-09-24 22:30:01
1912         16.5         0.0    76        10150      12.25     0.0       2024-09-24 22:35:00
1912         16.5         0.0    76        10150      12.25     0.0       2024-09-24 22:40:00
1912         16.5         0.0    76        10150      12.25     0.0       2024-09-24 22:45:00
1912         16.5         0.0    76        10150      12.25     0.0       2024-09-24 22:50:00
sqlite> select id,DeviceID,Name,unit,type,subtype,lastupdate from devicestatus where  (deviceId like "%8c" or Id = 14) and type =84 and subtype=16 and hardwareId=5;
ID    DeviceID  Name                 Unit  Type  SubType  LastUpdate
----  --------  -------------------  ----  ----  -------  -------------------
14    **26508**     Woonkamer - NetAtmo  140   84    16       2024-09-24 22:07:46
1814  51C5678C  Binnen               0     84    16       2024-09-09 23:25:23
1912  51c5678c  Binnen               0     84    16       2024-09-24 22:51:26

```

After switching to the development branch, the values of ID 14 are still added, but don't change anymore. The values of the active device with unit 1912 are added and updated. The values of other devices created by earlier test versions with the same type and subtype are not added...

After switching back to the master (last official release) with the original Netatmo interface, only the original device (14) was updated in the short log, but the values stayed the same until a new request was done at the Netatmo server.

Then I noticed that the device ID of the original device in the database was a decimal number (26508) and the devices page showed an uppercase HEX number (678C), so I wondered how (and where) this was done? After 2 days of searching (:shame:), I found the following line in main/WebServer.cpp:

```
root["result"][ii]["ID"] = is_number(sd[1]) ? std_format("**%04X**", (unsigned int)atoi(sd[1].c_str())) : sd[1];
```
This is about the same conversion I used for matching devices in the migration code., and it explains the difference, but there is no relation to the multiple updates in the short log.

So, I still don't see a cause for this multiple updates in the short log, so I still don't see a solution to solve this. 

However, Am I the only one experiencing this issue (as usually), or does this happen on all updated systems? 

In the last case, I think we need a solution for this (ASAP).

LoukKamp:
I was finally able to successfully migrate / merge all my Netatmo devices to the new devices created by the interface. As far as I can check this, without losing any data.

As I mentioned before, I only own a Netatmo weather station with 3 extra modules (2 inside, 1 outside), a broken rain sensor, and a dead wind meter. So I wonder what have been changed for other Netatmo devices (except for the device ID), like changed unit numbers, changed types, or subtypes?

I found the changes for the weather station by using the following SQL statement to analyze the situation:

```
SELECT ID, hardwareID, OrgHardwareID, deviceID, SUBSTR(iif(deviceID == CAST(deviceId AS Decimal), printf("%08.8X", deviceID), deviceID),-4,4) device_ID, unit, name, used, type, subtype FROM DeviceStatus WHERE hardwareID = 5 ORDER BY iif(deviceID == CAST(deviceId AS Decimal), printf("%08.8X", deviceID), deviceID), unit, type; 

```
So, If anyone with a lot of Netatmo devices can send my the result of this, I would be very appreciated.

When you don't want this information on a public forum like this, you may send it to me as a private mail on the Domoticz forum

JurgenLB:
Thanks for your work en effort.
gizmocuz has ask to use the HardwareBase instead off the updatevalueint for the standard sensors.
so I will change that and then we can see about the migration.

Kind regards

PR #6163, opened 2024-09-12
Fix MQTT AutoDiscovery electricity usage meter sensors

dwmw2:
• Fix the `total_increasing` support so that meter readings don't go backwards.
 • Support pairing of kW sensors.
 • Fix loss of precision in JSON/float conversions.

Tested with both ESPHome devices, and data from a smart electricity meter.


gizmocuz:
Thank you for this PR!!
One small change request, could you remove the FLOAT declaration here

double multiply = 1000.0F;

to

double multiply = 1000.0;

gizmocuz:
Thanks!!!

PR #6165, opened 2024-09-12
MQTT: Support 'onoff' colour mode

dwmw2:
I have a simple switch in ESPHome which doesn't work in Domoticz.
```yaml
light:
  - platform: binary
    name: "Blue LED"
    id: blue_led
    output: led_output

```
It appears like this in MQTT autodiscovery:
```json
{
   "avty_t" : "bathroom/status",
   "clrm" : true,
   "cmd_t" : "bathroom/light/blue_led/command",
   "dev" : {
      "cns" : [
         [
            "mac",
            "782184bb134c"
         ]
      ],
      "ids" : "782184bb134c",
      "mdl" : "esp32-gateway",
      "mf" : "Espressif",
      "name" : "bathroom",
   },
   "name" : "Blue LED",
   "obj_id" : "bathroom_blue_led",
   "schema" : "json",
   "stat_t" : "bathroom/light/blue_led/state",
   "supported_color_modes" : [
      "onoff"
   ],
   "uniq_id" : "782184bb134c-light-6a29e1c2"
}
```

### DIFF
--- a/hardware/MQTTAutoDiscover.cpp
+++ b/hardware/MQTTAutoDiscover.cpp
@@ -1126,6 +1126,15 @@ void MQTTAutoDiscover::on_auto_discovery_message(const struct mosquitto_message*
 					pSensor->supported_color_modes.clear();
 				}
 			}
+			if (pSensor->supported_color_modes.find("onoff") != pSensor->supported_color_modes.end())
+			{
+				if (pSensor->supported_color_modes.size() == 1)
+				{
+					//we only support onoff, so it is a normal switch and does not support setting a color
+					pSensor->bColor_mode = false;
+					pSensor->supported_color_modes.clear();
+				}
+			}
 		}
 		else
 		{


gizmocuz:
Why does it report 'onoff' as a color mode? Should this not be solved in ESPHome?

gizmocuz:
Could you post the complete config topic?

dwmw2:
The topic is `${mqtt_prefix}/homeassistant/light/bathroom/blue_led/config`

dwmw2:
> Why does it report 'onoff' as a color mode? Should this not be solved in ESPHome?

That seems correct to me. https://developers.home-assistant.io/docs/core/entity/light says

New integrations must implement both color_mode and supported_color_modes. If an integration is upgraded to support color mode, both color_mode and supported_color_modes should be implemented.

...

ColorMode.ONOFF: The light can be turned on or off. This mode must be the only supported mode if supported by the light.

gizmocuz:
Yeah, but that is "If an integration is upgraded to support color mode," ... a simple on/off switch should not support color mode
Could it be a ESPHome configuration issue? Did you create a normal switch (as the example) or did you add some other parameters?

dwmw2:
It literally says, "New integrations must implement both `color_mode` and `supported_color_modes`," and it's just reminding us to implement *both* those fields, not just one.

I think that color_mode *wasn't* mandatory in the past, and now it is. A device should implement color_mode even if it's an `onoff` type, in precisely the way that this device does. The sentence you cited should probably say "*when* an implementation is upgraded...", not "*if*..."

The YAML example I gave was exactly the configuration for this switch; no added parameters.

gizmocuz:
"should probably say..." , well, it should not implement any color mode if the device does not support color.
That is why I asked how your ESPHome definition/yaml file looks like, and if you based this on the normal on/off switch example

dwmw2:
Yes, this is an entirely simple switch. The YAML above is all there is.

It provides `onoff` as irs colour mode because the documentation says that is mandatory now for new integrations.

gizmocuz:
And you can't use 'light_output' ? Is the light controlled by a GPIO high/low? Then you can use 'pin: GPIOXX'
Or is it a color led?

dwmw2:
> Is the light controlled by a GPIO high/low? Then you can use 'pin: GPIOXX' Or is it a color led?

It's just a simple on/off LED (which happens to be blue). Controlled by a GPIO, yes:

```
light:
  - platform: binary
    name: "Blue LED"
    id: blue_led
    output: led_output

output:
  - id: led_output
    platform: gpio
    pin: GPIO23
```

> And you can't use 'light_output' ? 

What's that? I can't find any reference to it on the ESPHome documentation; the only `light_output` is in the example at https://esphome.io/components/light/binary.html where it's just the name of the corresponding GPIO output. In my case I've used the name `led_output` but — except for matching one thing to the other, of course — the names don't make any difference.

Honestly, the correct thing in my case is to stop advertising this light over MQTT at all; it's just the status LED telling me *whether* it's correctly connected to MQTT. I have no need to see or especially control it remotely. It's just that having spotted Domoticz getting it wrong, I figured I should submit the PR to fix that before doing so.

gizmocuz:
If you do not want it controlled, then you could remove the light part from the ESPHome config
I don't think there is anything wrong with the current code in Domoticz as it is being used by many and for quite a long time now.
That a normal light advertises 'color' is not correct

https://esphome.io/components/light/status_led.html

dwmw2:
> That a normal light advertises 'color' is not correct

Are you saying that the HA documentation which explicitly says that *all* new implementations shall have `supported_color_modes`, and defines an `onoff` mode that would otherwise have no reason to exist, is wrong? Shall we file a bug to get it fixed, and another to fix ESPHome not to conform to that documentation?

Seems more likely that the standard changed to make the colour modes mandatory even for`onoff` lights, and I'm the first person to offer a PR to fix Domoticz to cope with the change?

gizmocuz:
We can keep this PR and see if there is a need for this in the feature 
As it is now, you seem to use a wrong switch type in the sketch as it's not a colour led
It also seems easily solvable by changing this

ESPHome != Hass != Domoticz 

If every normal switch in ZWaveJS / Zigbee2MQTT is using it, then yes, but until then, this advertisement seems very strange and wierd to say the least!
Color mode OnOff has absolutely nothing to do with color

dwmw2:
>  As it is now, you seem to use a wrong switch type in the sketch as it's not a colour led

Hm? It's configured as a plain binary light. This is what I'm using: https://esphome.io/components/light/binary.html

Screenshot from 2024-09-16 09-37-22

If that's wrong, what *should* I be using instead?

> ESPHome != Hass != Domoticz

Well yes, but they're all attempting to be compatible with the MQTT AutoDiscovery standard which comes from HASS and is documented at https://developers.home-assistant.io/docs/core/entity/light/

> If every normal switch in ZWaveJS / Zigbee2MQTT is using it, then yes,

They should be. It's been part of the standard for a while. It was added to the binary light component in 2021 as part of 
https://github.com/esphome/esphome/pull/2012 for example. I'll reproduce for ZWave-JS and file a bug there because they still seem to be using the now-deprecated model.

> but until then, this advertisement seems very strange and wierd to say the least!
>Color mode OnOff has absolutely nothing to do with color

Seems sensible to me. A light has a `supported_color_modes` property, which tells you what colour modes, if any, it supports. The documentation calls it out really clearly, including the fact that it includes both plain dimmable, and on/off, lights:

- Start with an empty set
- If SUPPORT_COLOR_TEMP is set, add ColorMode.COLOR_TEMP
- If SUPPORT_COLOR is set, add ColorMode.HS
- If SUPPORT_WHITE_VALUE is set, add ColorMode.RGBW
- If SUPPORT_BRIGHTNESS is set and no color modes have yet been added, add ColorMode.BRIGHTNESS
- If no color modes have yet been added, add ColorMode.ONOFF

I'm not sure why you *wouldn't* want Domoticz to support this.

gizmocuz:
You have a different logic maybe. Does not matter. When it is needed, of course I add support for it. But lights/switches should not report anything about color when it doesn't do anything with color.

dwmw2:
The HA documentation explicitly says that lights should report "onoff" as their 'colour mode' if that is all they support, and in this case Domoticz is trying to be compatible with the HA discovery method.

The next release of ZWave-JS fixes this, and they all will.

dwmw2:
I believe this is now fixed in ZWave-JS-UI 9.19.0, which will report `supported_color_modes` for all lights *(including those with only `brightness` and `onoff` modes)* as required by the standard.

https://github.com/zwave-js/zwave-js-ui/releases/tag/v9.19.0

gizmocuz:
Okey, I just added the code to remove this mode again directly (will be in beta 16265 and up... building now)

dwmw2:
Looks like it's working. Thanks.
```
2024-09-19 08:51:30.736  Status: mqtt: Discovered: bathroom/bathroom (Blue LED) (unique_id: 782184bb134c-light-6a29e1c2)
```

Issue #6164, opened 2024-09-12
MQTT Temp/Hum devices choose strange name.


dwmw2:
These devices combine to make a single Temp/Hum device:

```
{
   "avty_t" : "bathroom/status",
   "dev" : {
      "cns" : [
         [
            "mac",
            "782184bb134c"
         ]
      ],
      "ids" : "782184bb134c",
      "mdl" : "esp32-gateway",
      "mf" : "Espressif",
      "name" : "bathroom",
      "sw" : "2024.9.0-dev (Sep  6 2024, 22:54:39)"
   },
   "dev_cla" : "temperature",
   "name" : "Bathroom Temperature",
   "obj_id" : "bathroom_bathroom_temperature",
   "stat_cla" : "measurement",
   "stat_t" : "bathroom/sensor/bathroom_temperature/state",
   "uniq_id" : "782184bb134c-sensor-fb9dd5d9",
   "unit_of_meas" : "°C"
}
```

```
{
   "avty_t" : "bathroom/status",
   "dev" : {
      "cns" : [
         [
            "mac",
            "782184bb134c"
         ]
      ],
      "ids" : "782184bb134c",
      "mdl" : "esp32-gateway",
      "mf" : "Espressif",
      "name" : "bathroom",
      "sw" : "2024.9.0-dev (Sep  6 2024, 22:54:39)"
   },
   "dev_cla" : "humidity",
   "name" : "Bathroom Humidity",
   "obj_id" : "bathroom_bathroom_humidity",
   "stat_cla" : "measurement",
   "stat_t" : "bathroom/sensor/bathroom_humidity/state",
   "uniq_id" : "782184bb134c-sensor-b74eb23e",
   "unit_of_meas" : "%"
}
```

That much is sensible. Including this device, however, is less sensible...

```
{
   "avty_t" : "bathroom/status",
   "clrm" : true,
   "cmd_t" : "bathroom/light/blue_led/command",
   "dev" : {
      "cns" : [
         [
            "mac",
            "782184bb134c"
         ]
      ],
      "ids" : "782184bb134c",
      "mdl" : "esp32-gateway",
      "mf" : "Espressif",
      "name" : "bathroom",
      "sw" : "2024.9.0-dev (Sep  6 2024, 22:54:39)"
   },
   "name" : "Blue LED",
   "obj_id" : "bathroom_blue_led",
   "schema" : "json",
   "stat_t" : "bathroom/light/blue_led/state",
   "supported_color_modes" : [
[...]

PR #6167, opened 2024-09-17
Fix MQTT AutoDiscover device names

dwmw2:
In HA MQTT AutoDiscovery, many sensors can be attached to a single device. The device doesn't have an MQTT topic of its own; it is reported as a property of each sensor, e.g.:
```
   "dev" : {
      "cns" : [
         [
            "mac",
            "782184bb134c"
         ]
      ],
      "ids" : "782184bb134c",
      "mdl" : "esp32-gateway",
      "mf" : "Espressif",
      "name" : "bathroom",
   },
```
During sensor discovery, we look for an existing device with the same 'ids' and use that if it exists, otherwise we create a new one. We keep track of which sensors are attached to which device, for matching up things like electricity power/usage and temp/hum/baro sensors.

However, as we discover each sensor, we currently *overwrite* the device name to include the sensor name, then blindly set the new sensor name to the current device name. Leaving the device with the name of whatever sensor happens to have been discovered last.

Clean that up, to set the device name using only information about the device itself, and then set the sensor name as before but using its own variable instead of messing with the device name.

Now my combined Temp/Hum sensors won't get such strange names.

Fixes: 6164

### DIFF
--- a/hardware/MQTTAutoDiscover.cpp
+++ b/hardware/MQTTAutoDiscover.cpp
@@ -736,6 +736,7 @@ void MQTTAutoDiscover::on_auto_discovery_message(const struct mosquitto_message*
 
 		pDevice->identifiers = device_identifiers;
 
+		// This is the name of the *device* to which this sensor happens to be attached
 		std::string dev_name("");
 		if (!root["device"]["name"].empty())
 		{
@@ -745,32 +746,35 @@ void MQTTAutoDiscover::on_auto_discovery_message(const struct mosquitto_message*
 		{
 			dev_name = root["dev"]["name"].asString();
 		}
+		else
+		{
+			dev_name = pDevice->identifiers;
+		}
+		if (pDevice->name.empty())
+			pDevice->name = dev_name;
 
+		std::string sensor_name("");
+		if (root["name"].empty())
+		{
+			root["name"] = sensor_unique_id;
+		}
 		if (!dev_name.empty())
 		{
-			if (root["name"].empty())
-			{
-				root["name"] = (dev_name.empty()) ? sensor_unique_id : dev_name;
-			}
 			std::string subname = root["name"].asString();
 			if (subname.find("0x") != 0)
 			{
-				pDevice->name = dev_name;
 				if (dev_name != subname)
-					pDevice->name += " (" + subname + ")";
+					sensor_name = dev_name + " (" + subname + ")";
 			}
 			else
 			{
-				pDevice->name = dev_name;
+				sensor_name = subname;
 			}
-		}
-		else if (!root["name"].empty())
+                }
+		else
 		{
-			pDevice->name = root["name"].asString();
+			sensor_name = root["name"].asString();
 		}
-
-		if (pDevice->name.empty())
-			pDevice->name = pDevice->identifiers;
 
 		if (!root["device"]["sw_version"].empty())
 			pDevice->sw_version = root["device"]["sw_version"].asString();
@@ -853,7 +857,7 @@ void MQTTAutoDiscover::on_auto_discovery_message(const struct mosquitto_message*
 		pSensor->config = qMessage;
 		pSensor->component_type = component;
 		pSensor->device_identifiers = device_identifiers;
-		pSensor->name = pDevice->name;
+		pSensor->name = sensor_name;
 
 		if (!root["enabled_by_default"].empty())
 			pSensor->bEnabled_by_default = root["enabled_by_default"].asBool();


gizmocuz:
Our naming scheme has been changed a few times over time.

With your patch my naming will be

image

What we have at the moment works best. Thanks for the PR!

gizmocuz:
This is how it looks with our current implementation

image

dwmw2:
Are those *sensor* names? I didn't think I'd changed those significantly; I was only trying to stop it from messing with the *device* names. Must be a bug in what I did; I'll take a look and fix it. Thanks for testing.

dwmw2:
I'm not quite sure how the code is coming up with your "current implementation"  version. Are you sure none of those were edited by hand after discovery? Please could you show the JSON from the MQTT discovery topic?

gizmocuz:
Yes, you changed them like this. I deleted the hardware, used your patch, and delete the hardware again and used the original code.

As said, we tried to find the best of both world here and the current implementation works best. (zigbee2mqtt/zwavejs/others)
Not sure how the current implementation looks with ESPHome, but I bet that you can find 'a' name in here.

gizmocuz:
here are a few from different transport types

Minimum:

`{"stat_t":"~/basenvt_voltage/state","dev":{"manufacturer":"PA1DVB","ids":["0x8000000011"]},"~":"my_victron","name":"BasenVT Voltage","uniq_id":"0x8000000011","unit_of_meas":"V","val_tpl":"{{ value_json.basenvt_voltage }}"}
`

Shelly:

```
{
  "dev": {
    "ids": [
      "485519DABF59"
    ]
  },
  "name": "shellyem-1 Total Returned 0",
  "stat_t": "~emeter/0/total_returned",
  "avty_t": "~online",
  "pl_avail": "true",
  "pl_not_avail": "false",
  "enabled_by_default": true,
  "uniq_id": "485519DABF59_0_totalreturned",
  "qos": 0,
  "~": "shellies/shellyem-1/",
  "dev_cla": "energy",
  "la_te": true,
  "stat_cla": "total_increasing",
  "unit_of_meas": "Wh",
  "val_tpl": "{{value|round(2)}}"
}
```

Zigbee2MQTT:

`{"availability":[{"topic":"zigbee2mqtt/bridge/state"}],"device":{"identifiers":["zigbee2mqtt_0x00158d000217ac86"],"manufacturer":"Innr","model":"Smart plug (SP 120)","name":"WCD_Master_Bedroom","sw_version":"2.0","via_device":"zigbee2mqtt_bridge_0x00124b0021cc4112"},"device_class":"power","enabled_by_default":true,"entity_category":"diagnostic","json_attributes_topic":"zigbee2mqtt/WCD_Master_Bedroom","object_id":"wcd_master_bedroom_power","origin":{"name":"Zigbee2MQTT","sw":"1.40.1-dev","url":"https://www.zigbee2mqtt.io"},"state_class":"measurement","state_topic":"zigbee2mqtt/WCD_Master_Bedroom","unique_id":"0x00158d000217ac86_power_zigbee2mqtt","unit_of_measurement":"W","value_template":"{{ value_json.power }}"}
`

ZWaveJS Power switch with 6 end-points (2 shown here):

`{"payload_off":false,"payload_on":true,"value_template":"{{ value_json.value }}","command_topic":"zwave/Slof_James/37/1/targetValue/set","state_topic":"zwave/Slof_James/37/1/currentValue","device":{"identifiers":["zwavejs2mqtt_0x161fc52_node3"],"manufacturer":"GreenWave Reality Inc.","model":"Multi-socket PowerNode (6-plug) (NP240 / NP242)","name":"Slof James","sw_version":"4.21"},"name":"Slof James_switch_1","unique_id":"zwavejs2mqtt_0x161fc52_3-37-1-currentValue"}`

`{"payload_off":false,"payload_on":true,"value_template":"{{ value_json.value }}","command_topic":"zwave/Slof_James/37/4/targetValue/set","state_topic":"zwave/Slof_James/37/4/currentValue","device":{"identifiers":["zwavejs2mqtt_0x161fc52_node3"],"manufacturer":"GreenWave Reality Inc.","model":"Multi-socket PowerNode (6-plug) (NP240 / NP242)","name":"Slof James","sw_version":"4.21"},"name":"Slof James_switch_4","unique_id":"zwavejs2mqtt_0x161fc52_3-37-4-currentValue"}
`

ESP32:

`{"device":{"identifiers":["WATERMETER-3157A3"],"manufacturer":"PA1DVB","model":"WATERSENSOR","name":"WATERMETER-3157A3","sw_version":"2022.02.25 rev 1.4"},"availability_topic":"esp8266-watermeter-sensor/WATERMETER-3157A3/status","state_topic":"esp8266-watermeter-sensor/WATERMETER-3157A3/state","name":"WATERMETER-3157A3","unit_of_measurement":"L","value_template":"{{value_json.pulses}}","unique_id":"WATERMETER-3157A3_pulses","icon":"mdi:counter-inc"}
`

dwmw2:
Thanks. The only thing I'm trying to change is the way that we change the device name over and over again, every time we find a new sensor attached to that device. Leaving the device with the name of the last sensor that happens to be discovered on it.
And then using *that* name for any combined Temp+Hum sensor we create on it.

I'll take another look at the code and see what I did wrong.

gizmocuz:
It's used by many and I have not heard a report of a problem yet

dwmw2:
Except for #6164? And we can see that that's what the code is doing. Is there a reason to keep changing the *device* name, that I'm missing?

gizmocuz:
Well, that is 'just' you reporting. If there was a big issue, we should have lots of reports.
So, I consider this not a valid issue until others are reporting this as well.
As said, we tried to find a solution that works for all. As mentioned before, you should be able to find your sensor
And I also consider sensors with a temp+hum+switch weird sensors you cannot buy off the shelf

dwmw2:
That combination seems fairly standard for a device designed to control a bathroom fan and/or heated mirror. You can buy dumb bathroom fans with a humidity switch built in; why not smart ones?

I guess other people may have just corrected the name and moved on, rather than diving into the code and trying to help improve Domoticz for everyone?

gizmocuz:
I don't think you see the 'big' picture... Let's assume 100 devices in a zigbee/zwavejs network and naming...
If there was something wrong, we know, and that is why it has been tweaked a few times
As it works now for the majority and has been for quite some while, I am also not eager to change it.
If you have one device, it is indeed easier to press rename on the one

dwmw2:
Yes, that's exactly why I'm trying to fix it rather than just working around the problem by renaming the devices for myself.

Remember, the only sensors affected by this bug are the "combined" sensors we build out of separate e.g. Temp and Hum sensors. That's when it makes a name based on the "device" name, and that's when it matters that the "device" gets renamed after the last sensor we happened to discover. So the Temp+Hum sensor ends up being named after some other random sensor.

I didn't mean to change the other sensor names; as I said that was a bug in my PR. I'll fix it.

gizmocuz:
Okey, but just don't spend too much time on it as it's working for 99% of the discovered devices

dwmw2:
Sure. It only affects these combined Temp+Hum / Temp+Hum+Baro sensors, as that's the only time that Domoticz ends up using the *device* names. But I'm always happy to try to improve Domoticz to get a little bit closer to 100% :)

gizmocuz:
Probably this goes also for Watt/kWh or other devices with multiple endpoints?

dwmw2:
Watt/kWh meters are OK because we don't synthesise a *new* sensor name for those; we just add the instantaneous power into the "existing" kWh sensor using its original name. I do have a few fixes for the kWh/W sensors  in #6163 but the *names* were fine.

I think it's *only* the temp/hum/baro sensors where we just make up a new name based on the parent *device* name. So that's the only place it matters that the *device* names are wrong.

Perhaps if we made the Temp/Hum/Baro sensors do the same thing as kWh, and just use the name of the Temperature sensor, then we wouldn't care about *device* names at all?

gizmocuz:
We should take care of device names. I think you are to focused on the comment of another function above it.
There is no check for temp/hum in the naming code

dwmw2:
>  There is no check for temp/hum in the naming code

Not in the code which names devices, indeed. That code is around line 760 in the `on_auto_discovery_message()` function. What *that* code does is calculate the right name for the *sensor*, but also changes the *device* name to that sensor's name as a side-effect. Leaving the *device* with the name of the last *sensor* to have been discovered on it. That's what I was trying to fix.

The special naming for Temp/Hum/Baro is about line 2503 where we do:
```
	if (bIsTemp || bIsHum || bIsBaro)
	{
		if (m_discovered_devices.find(pSensor->device_identifiers) == m_discovered_devices.end())
			return; //not discovered yet!?

		std::string szDeviceID = pSensor->device_identifiers;
		std::string sDeviceName = m_discovered_devices[pSensor->device_identifiers].name;
```
And that's the *only* place where we actually use the **device** names, AFAICT.

dwmw2:
Oh, I see what I did wrong. I was just trying to untangle the creation of the *device* vs. *sensor* names, by setting the device name first and *then* generating the sensor name from it. That way, we leave the device name in a sensible state while still getting the *same* sensor names as before.

Or such was the theory.

But in the case where the device doesn't have a *pretty* name, because we fall back to `pDevice->identifiers`, that shouldn't get used as the base for the sensor name. Testing the fix now...

dwmw2:
Pushing to the branch no longer works; fixed version in https://github.com/domoticz/domoticz/pull/6174

PR #6172, opened 2024-09-18
fix change Schedules

JurgenLB:
There was a missing selector switch for the schedules
and the naming is returned as before to '- Mode' and '- schedule'

This also blocks the GetEvents so there is no "User usage" Error.


gizmocuz:
Thank you!!

PR #6175, opened 2024-09-22
MQTT: Fix typo in total_increasing kWh meter handling

dwmw2:
It was supposed to read the previous value from the database if the sensor *hasn't* been received yet. A last-minute cosmetic change to make the if() condition look more like the existing code accidentally inverted the logic.


dwmw2:
No wait, that *should* be correct. It's ugly, and I'll fix it, but it's not actually *wrong*. Nevertheless, when I reset a sensor earlier today I did see it go back to zero for some reason. Perhaps I ran a wrong version of Domoticz without the total_increasing support.

PR #6177, opened 2024-09-26
Correcting Rain Sensor

JurgenLB:
Now I have changed the OrgHardwareId hardcoded to 0, this give again new 'unused devices' the type and subtype are the same this time.

I comment out some Log-lines to have a better overview 

I have also found a fix for the rain sensor.
The rain sensor gives 3 parameters

- rain
- rain_1
- rain_24
now with using rain_24 it is the acumuled data from the rain sensor.
I'm now waiting to see what the next rain wil do with the sensor data in Domoticz.

Kind regards

### DIFF
--- a/hardware/Netatmo.cpp
+++ b/hardware/Netatmo.cpp
@@ -488,7 +488,7 @@ bool CNetatmo::WriteToHardware(const char* pdata, const unsigned char /*length*/
 
 	if (packettype == pTypeLighting2)
 	{
-		Debug(DEBUG_HARDWARE, "Packettype pTypeLighting2 ");
+		//Debug(DEBUG_HARDWARE, "Packettype pTypeLighting2 ");
 		return true;
 	}
 	//This is the selector switch for setting the thermostat schedule
@@ -510,7 +510,7 @@ bool CNetatmo::WriteToHardware(const char* pdata, const unsigned char /*length*/
 	}
 	if (packettype == pTypeGeneralSwitch)
 	{
-		Debug(DEBUG_HARDWARE, "Packettype pTypeGeneralSwitch ");
+		//Debug(DEBUG_HARDWARE, "Packettype pTypeGeneralSwitch ");
 		//
 		// Recast raw data to get switch specific data
 		const _tGeneralSwitch* xcmd = reinterpret_cast<const _tGeneralSwitch*>(pdata);
@@ -565,7 +565,7 @@ bool CNetatmo::WriteToHardware(const char* pdata, const unsigned char /*length*/
 		int switchType = STYPE_Selector;
 		int devType = packettype;  //unsigned char
 		int subType = sSwitchTypeSelector;
-		Debug(DEBUG_HARDWARE, "Netatmo uid %08X", uid);
+		//Debug(DEBUG_HARDWARE, "Netatmo uid %08X", uid);
 
 		return SetProgramState(uid, xcmd->level);
 		//return true;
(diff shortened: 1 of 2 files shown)


JurgenLB:
With this the rain sensor is updated.

Foto 28-09-2024 om 15 48

and the Wind sensor is also corrected, an uses now 'sTypeWINDNoTempNoChill'
This is type 86 subtype 49.

so I think this is ready.

JurgenLB:
for the moment the Error; 'Error: Netatmo hardware (2) thread seems to have ended unexpectedly' is not more happening for me,
the highest Netatmo last checking is 42.00 seconds ago.

Now there is a double sensor update in this code, 1 time for getWeatherDetails(parsing the StationData)
 and 1 for getHomestatus(parsing the HomeStatus)

Its possible to block the GetWeatherDetails
and let all the sensors come from GetHomeStatus.

gizmocuz:
Thank you again!!

gizmocuz:
There are compile warnings.... 

hardware\Netatmo.cpp(1959,38): warning C4244: '=': conversion from 'float' to 'int', possible loss of data
hardware\Netatmo.cpp(1965,40): warning C4244: '=': conversion from 'float' to 'int', possible loss of data
hardware\Netatmo.cpp(2104,46): warning C4244: 'argument': conversion from 'int' to 'float', possible loss of data
hardware\Netatmo.cpp(2754,50): warning C4244: 'argument': conversion from 'int' to 'float', possible loss of data

You should also not call **UpdateValueInt** directly but use the normal update functions like 'SendAirQualitySensor'

Reason is that otherwise you can't use this in scripts/notifications etc...

JurgenLB:
Sorry not all is correct in first instance.
But I was thinking you wanted all the sensors by updateValueint, because then they all have the RF-level and the battery level.
Also when I change to 'sendairquality' there is again the change off IDX.


Kind regards

gizmocuz:
It is best to use the CDomoticzHardwareBase functions.

SendRainRateSensor, SendWind, SendPercentageSensor, etc

Indeed, some do not have RSSI, but that could be added later.

It does not matter if IDX is changing as this is still in beta version

@LoukKamp is working on a migration, but this can also be done manual.

UpdateValueInt should only be used directly under specific circumstances, and should also be used differently via an additional function. This hardware class can use the normal  CDomoticzHardwareBase functions.
I am pretty sure this was also the case before all modifications?

LoukKamp:
In addition to the bad timing of changing this back to what it was (I just finished the migration tool; see [this post](https://github.com/domoticz/domoticz/pull/6162#issuecomment-2389945081)), this will mean a lot of extra work.

The advantaged of Jurgen his approach was unity and centralized handling of storing data (all done in one function). I also made use of this by handling the migration in this function. 

But Gismocuz has a point too: When scripting and notifications don't work with this implementation, that would have a negative impact on the overall design of Domoticz. 

We probably need some extra send functions, like sendBatteryLevel() and sendSignalStrengthLevel().

This will probably mean the migration needs to have a reversed conversion too (a way back for the early adapters).

gizmocuz:
Correct, and 99% of all hardware are using these functions, and they are meant to be used....
No, you do not need extra send functions; you can use custom sensors for those.
Maybe tweak some send functions to include them. But i suspect they are already there for most of the sensors anyway, so not a bit deal if on a few of them they are not existing? (If you think it is, then we need to modify some send functions)
Anyway, I think the idea was to get this expensive hardware working again, and that seems to work. With the current (stable) implementation, you probably also had some sensors that did not include rssi/battery, and most of them did.

PR #6178, opened 2024-10-04
Netatmo migration tool to migrate old and new devices

LoukKamp:
First beta release for testing and discussion.

Merged with Netatmo rework by @JurgenLB.

Note: This version is only tested with the Netatmo Weather Station. Other devices still need testing or even adjusting when device units, types, or subtypes are changed.

### DIFF
--- a/hardware/Netatmo.cpp
+++ b/hardware/Netatmo.cpp
@@ -3012,7 +3012,7 @@ bool CNetatmo::ParseHomeStatus(const std::string& sResult, Json::Value& root, st
 							nDevice.BatteryLevel = batteryLevel;
 							SendCustomSensor(crcId, 2, batteryLevel, mrf_percentage, pName, "  ", mrf_status);   // RF-level
 							//UpdateValueInt(0, ID.c_str(), 2, pTypeGeneral, sTypePercentage, mrf_status, batteryLevel, '0', sigValue.c_str(), pName,  0, m_Name);  // RF- level
-							CNetatmo::MigrateDevices(ID.c_str(), 2, pTypeGeneral, sTypePercentage, pName);
+							CNetatmo::MigrateDevices(ID.c_str(), 2, pTypeGeneral, sTypeCustom, pName);
 						}
 					}
 					if (!module["wifi_state"].empty())
--- a/hardware/Netatmo.cpp
+++ b/hardware/Netatmo.cpp
@@ -265,7 +265,7 @@ void CNetatmo::Do_Work()
 					}
 					if (m_bPollHomeStatus)
 					{
-						// GetHomeStatusDetails
+						// GetHomesDataDetails
 						GetHomeStatusDetails();
 						Log(LOG_STATUS,"Status %d",  m_isLogged);
 					}
--- a/hardware/Netatmo.cpp
+++ b/hardware/Netatmo.cpp
@@ -135,7 +135,7 @@ void CNetatmo::Init()
 	m_ScheduleHome.clear();
 	m_DeviceModuleID.clear();
 	m_LightDeviceID.clear();
-
+	m_homeid.clear();
 	m_wifi_status.clear();
 	m_DeviceHomeID.clear();
 	m_PersonsNames.clear();
@@ -1549,7 +1549,17 @@ void CNetatmo::GetHomesDataDetails()
 				// Home ID from Homesdata
 				homeID = home["id"].asString();
 				m_homeid.push_back(homeID);
-				//Debug(DEBUG_HARDWARE, "Get Homes ID %s", homeID.c_str());
+				//Debug(DEBUG_HARDWARE, "Get Home ID %s", homeID.c_str());
+				std::stringstream stream_homeid;
+				for(size_t i = 0; i < m_homeid.size(); ++i)
+				{
+					if(i != 0)
+						stream_homeid << ",";
+					stream_homeid << m_homeid[i];
+				}
+				std::string st_homeid = stream_homeid.str();
+				Debug(DEBUG_HARDWARE, "Get Homes ID %s", st_homeid.c_str());
+				
 				m_Home_ID = home["id"].asString();
 				std::string Home_Name = home["name"].asString();
 				// home["altitude"];
@@ -1828,11 +1838,13 @@ void CNetatmo::GetHomeStatusDetails()
 	std::string locale;
 	std::string home_data;
 	std::string home_id;
+	m_homeid.clear();
 
 	GetHomesDataDetails();                 //Homes Data
 
 	Debug(DEBUG_HARDWARE, "Home Status Details");   // Multiple Homes possible
 	size = (int)m_homeid.size();
+	Log(LOG_STATUS, "Home_ID size = %d", size);
 	for (int i = 0; i < size; i++)
 	{
 		home_id = m_homeid[i];
@@ -2442,7 +2454,7 @@ bool CNetatmo::ParseDashboard(const Json::Value& root, const int DevIdx, const i
 		sp_temp = static_cast<int>(SP_temp);
 		std::string sValue = sp_str.str();
 		//Debug(DEBUG_HARDWARE, "(%d) %s (%s) [%s] parsedashboard  %s %s %d %d", Hardware_int, str_ID.c_str(), pchar_ID, name.c_str(), sValue.c_str(), m_Name.c_str(), rssiLevel, batValue);
-		SendSetPointSensor(ID, (uint8_t)((ID & 0x00FF0000) >> 16), (ID & 0XFF00) >> 8, ID & 0XFF, Unit, SP_temp, sName);    // No RF-level and Battery level
+		SendSetPointSensor(ID, (uint8_t)((ID & 0x00FF0000) >> 16), (ID & 0XFF00) >> 8, ID & 0XFF, Unit, batValue, SP_temp, sName);    // No RF-level
 	}
 
 	//Rain meter
@@ -3203,7 +3215,7 @@ bool CNetatmo::ParseHomeStatus(const std::string& sResult, Json::Value& root, st
 						float SP_temp = std::stof(room_setpoint);
 						int uid = crcId;
 
-						SendSetPointSensor(crcId, (uint8_t)((crcId & 0x00FF0000) >> 16), (crcId & 0XFF00) >> 8, crcId & 0XFF, Unit, SP_temp, moduleName); // No RF-level and Battery-level
+						SendSetPointSensor(crcId, (uint8_t)((crcId & 0x00FF0000) >> 16), (crcId & 0XFF00) >> 8, crcId & 0XFF, Unit, batteryLevel, SP_temp, moduleName); // No RF-level and Battery-level
 
 						// thermostatModuleID
 						uint64_t mid = convert_mac(module_id);
--- a/hardware/Netatmo.cpp
+++ b/hardware/Netatmo.cpp
@@ -53,6 +53,7 @@ std::string ReadFile(std::string filename)
 }
 #endif
 
+#ifdef _DEBUG
 std::string prettifyJson(const Json::Value value)
 {
 	using namespace Json;
@@ -69,6 +70,7 @@ std::string prettifyJson(const Json::Value value)
 
 	return result.str();
 }
+#endif
 
 
 CNetatmo::CNetatmo(const int ID, const std::string& username, const std::string& password)
@@ -960,7 +962,7 @@ bool CNetatmo::MergeDevices(const uint64_t ipOldDeviceId, const uint64_t ipNewDe
 /// <summary>
 /// Merge logs in all log tables
 /// </summary>
-bool CNetatmo::MergeDeviceLogs (const std::string& spTableName, const std::string& spOldDeviceId, const std::string& spNewDeviceId)
+bool CNetatmo::MergeDeviceLogs(const std::string& spTableName, const std::string& spOldDeviceId, const std::string& spNewDeviceId)
 {
 	auto logResult = m_sql.safe_query
 	(
--- a/hardware/Netatmo.cpp
+++ b/hardware/Netatmo.cpp
@@ -209,7 +209,7 @@ void CNetatmo::Do_Work()
 	if (!m_sql.GetPreferencesVar(m_sNetatmoProtVersionPrefName, m_iNetatmoProtVersion))
 		m_iNetatmoProtVersion = 0;
 
-	if (m_iNetatmoProtVersion < 1)
+	if (m_iNetatmoProtVersion < 2)
 	{
 		m_bMigrationFlag = true;
 		Log(LOG_STATUS, "New naming protocol detected; start migration old and new mathing devices.");
@@ -265,7 +265,7 @@ void CNetatmo::Do_Work()
 					}
 					if (m_bPollHomeStatus)
 					{
-						// GetHomesDataDetails
+						// GetHomeStatusDetails
 						GetHomeStatusDetails();
 						Log(LOG_STATUS,"Status %d",  m_isLogged);
 					}
@@ -567,33 +567,35 @@ uint64_t CNetatmo::UpdateValueInt(int HardwareID, const char* deviceID, unsigned
 /// When enabled, this function will match the same (old and new) devices based on the
 /// (HEX formatted) deviceID
 /// </summary>
-void CNetatmo::MigrateDevices(const char* deviceID, unsigned char unit, unsigned char devType, unsigned char devSubType, std::string& devname, uint64_t DeviceRowIdx)
+void CNetatmo::MigrateDevices(const char* deviceID, const unsigned char unit, const unsigned char devType, const unsigned char devSubType, const std::string& devname, const uint64_t pDeviceRowIdx)
 {
 	if (m_bMigrationFlag) {
 		std::string sDeviceID = deviceID;
 		std::transform(sDeviceID.begin(), sDeviceID.end(), sDeviceID.begin(), ::toupper);
+		uint64_t deviceRowIdx = pDeviceRowIdx;
+		std::string lastUpdate;
 
-//		Log(LOG_STATUS, "UpdateValueInt: DeviceRowIdx=%d, ID=%s (%s), unit=%d, devType=%d, devSubType=%d, devname=%s", 
-//			(int)DeviceRowIdx,
-//			deviceID, 
-//			sDeviceID.c_str(), 
-//			unit, 
-//			devType, 
-//			devSubType, 
-//			devname.c_str()
-//		);
-//
+		Log(LOG_STATUS, "MigrateDevices: pDeviceRowIdx=%d, ID=%s (%s), unit=%d, devType=%d, devSubType=%d, devname=%s", 
+			(int)deviceRowIdx,
+			deviceID, 
+			sDeviceID.c_str(), 
+			unit, 
+			devType, 
+			devSubType, 
+			devname.c_str()
+		);
+
 
 		std::vector<std::vector<std::string> > result;
 
 		result = m_sql.safe_query(
-			"SELECT ID "\
+			"SELECT ID, LastUpdate "\
 			"FROM DeviceStatus "\
 			"WHERE (HardwareID=%d "\
-				"AND DeviceID == %q "\
+				"AND DeviceID == '%q' "\
 				"AND Type == %d "\
 				"AND SubType == %d "\
-				"AND unit == %d) "
+				"AND unit == %d) ",
 			m_HwdID,
 			sDeviceID.c_str(),
 			devType,
@@ -605,18 +607,26 @@ void CNetatmo::MigrateDevices(const char* deviceID, unsigned char unit, unsigned
 				sDeviceID.c_str());
 			return;
 		}
-		if (result.size() <> 1) {
-			Log(LOG_ERROR, "Expected number of devices found: %d matches found for %s (0x%s):",
+		if (result.size() != 1) {
+			Log(LOG_ERROR, "Unexpected number of devices found: %d matches found for %s (0x%s):",
 				result.size(),
 				deviceID,
 				sDeviceID.c_str());
 			return;
 		}
-		uint64_t DeviceRowIdx = result[0][0];
 
+		if (pDeviceRowIdx == -1)
+			deviceRowIdx = std::stoll(result[0][0]);
+		lastUpdate = result[0][1];
 
+		if (std::all_of(sDeviceID.begin(), sDeviceID.end(), ::isdigit) && sDeviceID.length() < 8) {
+			int iDeviceID = std::stoi(sDeviceID);
+			std::stringstream str;
+			str << std::uppercase << std::hex << iDeviceID;
+			str >> sDeviceID;
+		}
 		std::string searchString = sDeviceID.substr(sDeviceID.length()-2,2);
-
+		//Log(LOG_STATUS, "Search for devices with ID ending with: %s and devType = %d ... ", searchString.c_str(), devType);
 		result = m_sql.safe_query(
 			"SELECT ID, DeviceID, Name, Type, SubType, Unit, Used, LastUpdate "\
 			"FROM DeviceStatus "\
@@ -629,15 +639,17 @@ void CNetatmo::MigrateDevices(const char* deviceID, unsigned char unit, unsigned
 				"AND Type = %d) "\
 			"ORDER BY ID DESC",
 			m_HwdID, 
-			DeviceRowIdx,
+			deviceRowIdx,
 			"%X", 
 			searchString.c_str(), 
 			devType);
 		if (!result.empty()) {
-//			Log(LOG_STATUS, "Possible conversion found: %d matches found for %s (0x%s):",
-//				result.size(), 
-//				deviceID, 
-//				sDeviceID.c_str());
+			Log(LOG_STATUS, "Possible conversion(-s) found for device with ID = %" PRIu64 ": %d matches found for %s (0x%s):",
+				deviceRowIdx,
+				result.size(), 
+				deviceID, 
+				sDeviceID.c_str());
+
 			for (const auto& device : result)
 			{
 				if (!device.empty()) {
@@ -650,7 +662,7 @@ void CNetatmo::MigrateDevices(const char* deviceID, unsigned char unit, unsigned
 					unsigned char oldUsed = std::stoul(device[6], nullptr, 10);
 					std::string oldLastUpdate = device[7];
 
-					if (oldID == DeviceRowIdx)
+					if (oldID == deviceRowIdx)
 					{
 						Log(LOG_STATUS, "Case 0 (New device): Device: ID=%" PRIu64 ", DeviceID=0x%s, Name=%s, Type=%d, devSubType=%d, unit=%d, used=%d, lastUpdate=%s",
 							oldID,
@@ -667,6 +679,8 @@ void CNetatmo::MigrateDevices(const char* deviceID, unsigned char unit, unsigned
 						oldUnit == unit
 						&& oldDevType == devType
 						&& oldDevSubType == devSubType
+						&& m_iNetatmoProtVersion <= 1
+						//&& oldLastUpdate < lastUpdate
 					)
 					{
 						Log(LOG_STATUS, "Case 1 (full match): Device: ID=%" PRIu64 ", DeviceID=0x%s, Name=%s, Type=%d, devSubType=%d, unit=%d, used=%d, lastUpdate=%s",
@@ -679,7 +693,7 @@ void CNetatmo::MigrateDevices(const char* deviceID, unsigned char unit, unsigned
 							oldUsed,
 							oldLastUpdate.c_str()
 						);
-						MergeDevices(oldID, DeviceRowIdx);
+						MergeDevices(oldID, deviceRowIdx);
 						break; 	//One device per cycle
 					}
 					else if (
@@ -688,6 +702,8 @@ void CNetatmo::MigrateDevices(const char* deviceID, unsigned char unit, unsigned
 						&& oldDevType == devType
 						&& oldDevSubType == sTypeSystemTemp
 						&& devSubType == sTypeTH_LC_TC
+						&& m_iNetatmoProtVersion <= 1
+						//&& oldLastUpdate < lastUpdate
 					)
 					{
 						Log(LOG_STATUS, "Case 2 (new subtype (82:5->82:160): Device: ID=%" PRIu64 ", DeviceID=0x%s, Name=%s, Type=%d, devSubType=%d, unit=%d, used=%d, lastUpdate=%s",
@@ -700,7 +716,7 @@ void CNetatmo::MigrateDevices(const char* deviceID, unsigned char unit, unsigned
 							oldUsed,
 							oldLastUpdate.c_str()
 						);
-						MergeDevices(oldID, DeviceRowIdx);
+						MergeDevices(oldID, deviceRowIdx);
 						break; 	//One device per cycle
 					}
 					else if (
@@ -709,6 +725,8 @@ void CNetatmo::MigrateDevices(const char* deviceID, unsigned char unit, unsigned
 						&& oldDevType == devType
 						&& oldDevSubType == sTypeSoilMoisture
 						&& devSubType == sTypeRAINByRate
+						&& m_iNetatmoProtVersion <= 1
+						//&& oldLastUpdate < lastUpdate
 					)
 					{
 						Log(LOG_STATUS, "Case 3 (new subtype (85:3->85:113): Device: ID=%" PRIu64 ", DeviceID=0x%s, Name=%s, Type=%d, devSubType=%d, unit=%d, used=%d, lastUpdate=%s",
@@ -721,15 +739,18 @@ void CNetatmo::MigrateDevices(const char* deviceID, unsigned char unit, unsigned
 							oldUsed,
 							oldLastUpdate.c_str()
 						);
-						MergeDevices(oldID, DeviceRowIdx);
+						MergeDevices(oldID, deviceRowIdx);
 						break; 	//One device per cycle
 					}
 					else if (
 						unit == 0
+						&& oldUnit == 1
 						&& oldDevType == pTypeGeneral	// (from HardwareTypes.h)
 						&& oldDevType == devType
 						&& oldDevSubType == sTypeSoundLevel
 						&& oldDevSubType == devSubType
+						&& m_iNetatmoProtVersion <= 1
+						//&& oldLastUpdate < lastUpdate
 					)
 					{
 						Log(LOG_STATUS, "Case 4 (new unit number for sound): Device: ID=%" PRIu64 ", DeviceID=0x%s, Name=%s, Type=%d, devSubType=%d, unit=%d, used=%d, lastUpdate=%s",
@@ -742,16 +763,18 @@ void CNetatmo::MigrateDevices(const char* deviceID, unsigned char unit, unsigned
 							oldUsed,
 							oldLastUpdate.c_str()
 						);
-						MergeDevices(oldID, DeviceRowIdx);
+						MergeDevices(oldID, deviceRowIdx);
 						break; 	//One device per cycle
 					}
 					else if (
-						oldUnit != unit
+						unit == 0
 						&& oldUnit == 140
 						&& oldDevType == devType
 						&& oldDevType == pTypeTEMP_HUM_BARO	// (from RFXtrx.h)
 						&& oldDevSubType == sTypeTHBFloat
 						&& oldDevSubType == devSubType
+						&& m_iNetatmoProtVersion <= 1
+						//&& oldLastUpdate < lastUpdate
 					) 
 					{
 						Log(LOG_STATUS, "case 5 (New unit number for weather station): Device: ID=%" PRIu64 ", DeviceID=0x%s), Name=%s, Type=%d, devSubType=%d, unit=%d, used=%d, lastUpdate=%s",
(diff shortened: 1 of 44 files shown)


hardware/Netatmo.cpp
```diff
@@ -488,14 +526,146 @@ uint64_t CNetatmo::convert_mac(std::string mac)
 }
 
 
+uint64_t getValueFromString(const std::string& value)
```
gizmocuz: Instead of this function please use **std::stoll**

hardware/Netatmo.cpp
```diff
@@ -488,14 +526,146 @@ uint64_t CNetatmo::convert_mac(std::string mac)
...
+{
+	uint64_t result;
+	std::istringstream iss( value );
+	iss >> result;
+	return result;
+}
+
+std::string getStringFromValue(const uint64_t value)
```
gizmocuz: Instead of this function, please use **std::to_string**

hardware/Netatmo.cpp
```diff
@@ -488,14 +526,146 @@ uint64_t CNetatmo::convert_mac(std::string mac)
...
+	return string;
+}
+
+/// <summery>
+/// Merge old and new devices, bu deleting duplicate log records and
+/// copy all other log records from the new to the old device before
+/// updating the old device to the new devices and finaly remove the new device.
+/// Note: the old device will become the new device.
+/// </summary>
+bool CNetatmo::MergeDevices (const uint64_t ipOldDeviceId, const uint64_t ipNewDeviceId)
```
gizmocuz: Please remove the space after the function name

hardware/Netatmo.cpp
```diff
@@ -488,14 +526,146 @@ uint64_t CNetatmo::convert_mac(std::string mac)
...
+{
+//	Log(LOG_STATUS, "MergeDevices: Merge old device %" PRIu64 " with new device %" PRIu64 ".", ipOldDeviceId, ipNewDeviceId);
+
+	std::string sOldDeviceId = getStringFromValue(ipOldDeviceId);
+	std::string sNewDeviceId = getStringFromValue(ipNewDeviceId);
+
+	Log(LOG_STATUS, "MergeDevices: Merge old device %s with new device %s", sOldDeviceId.c_str(), sNewDeviceId.c_str());
+
+	auto result = m_sql.safe_query
```
gizmocuz: No need to apply the end semicolon, and it could be a simple one line

hardware/Netatmo.cpp
```diff
@@ -488,14 +526,146 @@ uint64_t CNetatmo::convert_mac(std::string mac)
...
+	(
+		"SELECT id FROM DeviceStatus  WHERE Id='%q' OR Id = '%q';",
+		sOldDeviceId.c_str(),
+		sNewDeviceId.c_str()
+	);
+	if (result.empty())
```
gizmocuz: Here you check if the result is empty, and the next check you check if the result contains at least two values.
This function is only called on certain conditions below, and only by this class, so I think the first check can be removed, even the second check i guess... 

Also please keep the variable names case sensitive as how there are created. Id --> ID

hardware/Netatmo.cpp
```diff
@@ -488,14 +526,146 @@ uint64_t CNetatmo::convert_mac(std::string mac)
...
+}
+
+/// <summary>
+/// Merge logs in all log tables
+/// </summary>
+bool CNetatmo::MergeDeviceLogs (const std::string& spTableName, const std::string& spOldDeviceId, const std::string& spNewDeviceId)
+{
+	auto logResult = m_sql.safe_query
+	(
+		"SELECT COUNT(*) FROM %q as a  WHERE a.deviceRowId='%q' AND  EXISTS (SELECT date FROM %q as b WHERE b.DeviceRowID = '%q' AND b.date == a.date);",
```
gizmocuz: Please keep the database column names case sensitive, how they are defined. like DeviceRowID, Date etc.

gizmocuz: Also for the remaining code please

hardware/Netatmo.cpp
```diff
@@ -504,8 +674,167 @@ uint64_t CNetatmo::UpdateValueInt(int HardwareID, const char* ID, unsigned char
                 Log(LOG_NORM, szLogString);
         }
         m_mainworker.sOnDeviceReceived(m_HwdID, DeviceRowIdx, devname, nullptr);
-        m_notifications.CheckAndHandleNotification(DeviceRowIdx, m_HwdID, ID, devname, unit, devType, subType, nValue, sValue);
+        m_notifications.CheckAndHandleNotification(DeviceRowIdx, m_HwdID, sDeviceID.c_str(), devname, unit, devType, subType, nValue, sValue);
```
gizmocuz: no need for the c_str() here

hardware/Netatmo.cpp
```diff
@@ -504,8 +674,167 @@ uint64_t CNetatmo::UpdateValueInt(int HardwareID, const char* ID, unsigned char
...
+//			sDeviceID.c_str(), 
+//			unit, 
+//			devType, 
+//			subType, 
+//			devname.c_str());
+
+		std::vector<std::vector<std::string> > result;
+		std::string searchString = sDeviceID.substr(sDeviceID.length()-2,2);
+
+		result = m_sql.safe_query(
```
gizmocuz, line 615: Please keep the MySQL functions case sensitive, like most of them are in CAPITAL letters, see the documentation

https://www.w3schools.com/sql/func_mysql_substr.asp

Also check this throughout this PR

hardware/Netatmo.cpp
```diff
@@ -504,8 +674,167 @@ uint64_t CNetatmo::UpdateValueInt(int HardwareID, const char* ID, unsigned char
...
+							oldDevType,
+							oldSubType,
+							oldUnit,
+							oldUsed,
+							oldLastUpdate.c_str()
+						);
+						MergeDevices(oldID, DeviceRowIdx);
+						break; 	//One device per cycle
+					}
+					else if (oldUnit == unit && oldDevType == 82 && oldDevType == devType && oldSubType == 5 && subType == 113)
```
gizmocuz: As these if/then/else lines can become quite large, I use another coding formatting throughout the Domoticz code like
```

    else if (
        (oldUnit == unit)
        && (oldDevType == 82)
        && (oldDevType == devType)
        && (oldSubType == 5)
        && (subType == 113)
        )
    {
        ....
    }

```

I think this also looks a lot cleaner

And instead of these numbers, please use the real variable/defines, like defined in **hardwaretypes.h** or **RFXtrx.h**

82 -> pTypeTEMP_HUM

Numbers 'can' change, and nobody knows whats happening by looking at just numbers

hardware/Netatmo.cpp
```diff
@@ -1872,10 +2202,19 @@ bool CNetatmo::ParseDashboard(const Json::Value& root, const int DevIdx, const i
...
+//			Log(LOG_STATUS, "No time stamp in received data; using current time");
+//			tNetatmoLastUpdate = tNow;
+//		}
+		//Debug(DEBUG_HARDWARE, "Module [%s] last update = %s (%d)", name.c_str(), ctime(&tNetatmoLastUpdate), tNetatmoLastUpdate);
+		Log(LOG_STATUS, "Module [%s] last update = %s (%d)", name.c_str(), ctime(&tNetatmoLastUpdate), tNetatmoLastUpdate);
+
 		// check if Netatmo data was updated in the past NETAMO_POLL_INTERVALL (+1 min for sync time lags)... if not means sensors failed to send to cloud
 		int Interval = NETAMO_POLL_INTERVALL + 60;
 		//Debug(DEBUG_HARDWARE, "Module [%s] Interval = %d", name.c_str(), Interval);
+		Log(LOG_STATUS, "Module [%s] Interval = %d", name.c_str(), Interval);
```
gizmocuz: I think this should be a DEBUG_HARDWARE ?

hardware/Netatmo.h
```diff
@@ -166,4 +168,7 @@ class CNetatmo : public CDomoticzHardwareBase
 
 	uint64_t UpdateValueInt(int HardwareID, const char* ID, unsigned char unit, unsigned char devType, unsigned char subType, unsigned char signallevel, unsigned char batterylevel, int nValue, const char* sValue, std::string& devname, bool bUseOnOffAction, const std::string& user);
 	bool ParseDashboard(const Json::Value &root, int DevIdx, int ID, std::string& name, const std::string &ModuleType, int battery_percent, int rf_status, std::string& Hardware_ID, std::string& home_id);
+
+	bool MergeDevices (const uint64_t ipOldDeviceId, const uint64_t ipNewDeviceId);
```
gizmocuz: Please remove the space behind the function name

www/app/hardware/Hardware.html
```diff
@@ -1485,6 +1485,10 @@ <h2><span data-i18n="Hardware">Hardware</span>:</h2>
 					</div>
 				</td>
 			</tr>
+			<tr>
+				<td align="right" style="width:110px"><span data-i18n="Migration Tool"></span>:</label></td>
```
gizmocuz: I guess the migration is a onetime only thing to do?
If yes, this should not be here.
Maybe you could use a Preference version value (NetatmoHWVersion) and if this does not exists, or is a specific value, call the migration at startup of the Netatmo hardware?

gizmocuz: This way also no changes in the web part needs to be made

gizmocuz (changes requested):
Sorry, quite some remarks here

JurgenLB:
I just wonder why you go from 'DEBUG_HARDWARE' to 'LOG_STATUS' between line 2204 to 2217

kiddigital:
@LoukKamp , is this code intended to be run only once? After migration, this code will never be used again by migrated users?

gizmocuz:
Guys, these are all questions I already asked above

hardware/Netatmo.cpp
LoukKamp: OK

hardware/Netatmo.cpp
LoukKamp: OK

hardware/Netatmo.cpp
LoukKamp: OK

hardware/Netatmo.cpp
LoukKamp: OK

hardware/Netatmo.cpp
LoukKamp: Investigating....

hardware/Netatmo.cpp
LoukKamp: OK

hardware/Netatmo.cpp
LoukKamp: This is probably Jurgen's part of the code, but I'll take a look.

hardware/Netatmo.cpp
LoukKamp: OK (I suppose you mean SUBSTR)

hardware/Netatmo.cpp
LoukKamp: OK, but I looked in hardwaretypes.h, but didn't find matching names for the numbers. Maybe in RFXtrx.h; I'll take a look.

hardware/Netatmo.cpp
LoukKamp: Sorry, but I (miss-)use LOG_STATUS for logging/debugging, because when I enable debugging, I get a lot of debug messages from other devices that will distract my attention of the important message to understand what is happening.

I will remove them later (when I get some results from other beta testers with devices I don't own).

hardware/Netatmo.h
LoukKamp: OK (sorry, was not my intention ;)).

www/app/hardware/Hardware.html
LoukKamp: Not really: It needs to run several times, depending on the moment that the "Early Adaptor" started using the new software (there were several changes over time in the device ID (incl. case changes), unit numbers, types, and subtypes during the development process). I planned to do this in one run, but realized that the "New" device (that is triggering the migration for this device) was changing during the process, and also, I needed to work backwards in time. So I needed to split up. Users that will step in now, probably will need only one run, but users that stepped in after the first update, probably need to run it 3 or even 4 times (as in my case).

I also think that I should give the user an option to use the tool or not (the default setting of the flag is OFF).

There is no changed hardware or firmware version to test on; All changes are caused by the new interface within Domoticz (in hardware/Netatmo.cpp).

The only way to test if a migration is useful, is the very complex SQL statement that matches old and new devices, based on their Device ID (or a part of it). That is why I wanted to introduce the flag, so it can be switched off after the conversion to avoid unnecessary database activity.

I also think that the whole tool, including the flag on the website, can be removed after one or two master releases.

LoukKamp:
> I just wonder why you go from 'DEBUG_HARDWARE' to 'LOG_STATUS' between line 2204 to 2217

No, not really (see comments above). I will remove them after I'm completed the tool.

www/app/hardware/Hardware.html
gizmocuz: Yes, that's why I mentioned the idea of using a version number in the preferences so you can migrate on startup of the hardware

hardware/Netatmo.cpp
gizmocuz: if you want you can also add

#ifdef _DEBUG
... do something
#endif

Then this will only be displayed when you build in debug mode (which you do while developing)

hardware/Netatmo.cpp
gizmocuz: correct and also some date/time functions, most things are case sensitive, like SELECT instead of Select for instance (not that you did this... )

hardware/Netatmo.cpp
LoukKamp: On the second hand, I don't see what you mean here: There is only one semicolon in the (single) line with the log statement, and it is required by the C++ standard...

hardware/Netatmo.cpp
LoukKamp: These lines were meant as general check on all kinds of (external caused) errors (like e.g. a database lock) and a sanity check. I was learned to test always on the unexpected...

I'll remove them, because there is no need for this query anymore.

hardware/Netatmo.cpp
LoukKamp: Done

hardware/Netatmo.cpp
LoukKamp: Done

hardware/Netatmo.cpp
LoukKamp: Done

hardware/Netatmo.cpp
LoukKamp: Done (removed)

hardware/Netatmo.cpp
LoukKamp: Done

hardware/Netatmo.cpp
LoukKamp: In the end, it was me: I've changed this because of the upper case conversion needed to compare the device IDs. Indeed, the c_str() function is not needed, as the parameter for the called function is also std::string.

But now I realize that this converted uppercase HEX value is also stored in the database, and this was not the intention of the author (@JurgenLB). So when he wants this to be corrected, I need to see if this has any consequences on the matching process: Original device IDs were in the format "04X" (4 uppercase HEX digits).

hardware/Netatmo.cpp
LoukKamp: The example above is not a SQL syntax, but C++, so it needs to be in lower case in this case. However, I modified all other cases where it was part of a SQL statement.

hardware/Netatmo.cpp
LoukKamp: Done (both requests)

hardware/Netatmo.cpp
LoukKamp: That would not help, because after merging to the beta release, it will be built without the _DEBUG value defined. I will need this information to check the logs from (BETA-) users that experience issues or have devices I don't own. 

By the way, I don't have a proper debug environment (with a decent symbolic) setup yet, because the first trials to do so failed, and I didn't have the time to figure that out. So no useful need for debug mode so far... Sorry!

hardware/Netatmo.h
LoukKamp: Done

www/app/hardware/Hardware.html
LoukKamp: But, in that case, I need to add an option to the preferences' webpage to enable the migration. While in BETA, I don't want to run this tool without permission and knowledge of the user.

hardware/Netatmo.cpp
gizmocuz: It's in the sql query

hardware/Netatmo.cpp
gizmocuz: Maybe try to install/use visual studio? Community version is free and great!

www/app/hardware/Hardware.html
gizmocuz: No, there are functions like

`CSQLHelper::GetPreferencesVar`

```
m_netamo_hw_version = 1;
if (!m_sql.GetPreferencesVar("NetamoHWVersion", m_netamo_hw_version)) m_netamo_hw_version = 0;

```

hardware/Netatmo.cpp
LoukKamp: Thanks for the tip: I used this for several years to development in C# on a Windows system. But can this be used to debug on a Raspberry Pi, running Linux? In need to figure this out, but too late for this project, I guess.

www/app/hardware/Hardware.html
LoukKamp: That still leave me with two questions: How should the user enable the migration tool, and what if the user has more versions of the device running (e.g. one for production and one for testing like me, or one for the weather station and one for the alarm system)?

Can we please agree that I use this flag on the hardware page for beta testing, and that I replace it with your suggestion (or something like that) before the next master release?

hardware/Netatmo.cpp
LoukKamp: I removed the query, so problem solved. 

(I also removed the semicolons from the other SQL statements)

hardware/Netatmo.cpp
LoukKamp: I moved the uppercase translation, so the updates are done without it, but found that @JurgenLB already changed this in his part of the code. So, in fact, this change has no effect on the results: I'm happy with it.

www/app/hardware/Hardware.html
gizmocuz: The user should 'not' be bothered with the migration tool. You as developer knows this.
Version 2 (this new version), we need to convert some sensors.
Later some more changes, version 3 (#define at the top of the class), you need to convert more

LoukKamp: OK, I rest my case: see the last commit (removed migration flag)

LoukKamp:
Who is waiting for whom??? This PR is still waiting for requested changes, although I thought all requested changes were implemented.

The last updates from @JurgenLB are also still in "pre-beta", so I can't merge jet.

I'm running the code that is based on the current beta, including my changes from this PR.  However, today I noticed errors in my logging: 

```
2024-10-13 07:55:15.866 Error: NetAtmoWeerEnLuchtkwaliteit hardware (5) thread seems to have ended unexpectedly 
2024-10-13 07:55:29.868 Error: NetAtmoWeerEnLuchtkwaliteit hardware (5) thread seems to have ended unexpectedly 
2024-10-13 07:55:43.872 Error: NetAtmoWeerEnLuchtkwaliteit hardware (5) thread seems to have ended unexpectedly 
2024-10-13 08:11:47.414 Error: NetAtmoWeerEnLuchtkwaliteit hardware (5) thread seems to have ended unexpectedly 
2024-10-13 08:12:01.417 Error: NetAtmoWeerEnLuchtkwaliteit hardware (5) thread seems to have ended unexpectedly 
....

2024-10-13 13:02:03.374 Error: NetAtmoWeerEnLuchtkwaliteit hardware (5) thread seems to have ended unexpectedly 
2024-10-13 13:17:57.546 Error: NetAtmoWeerEnLuchtkwaliteit hardware (5) thread seems to have ended unexpectedly 
2024-10-13 13:18:11.549 Error: NetAtmoWeerEnLuchtkwaliteit hardware (5) thread seems to have ended unexpectedly 
2024-10-13 13:18:17.365 Status: NetAtmoWeerEnLuchtkwaliteit: HomeStatus parsed 
2024-10-13 13:18:17.692 Status: NetAtmoWeerEnLuchtkwaliteit: HomeStatus parsed 
2024-10-13 13:18:18.006 Status: NetAtmoWeerEnLuchtkwaliteit: HomeStatus parsed 
2024-10-13 13:18:18.334 Status: NetAtmoWeerEnLuchtkwaliteit: HomeStatus parsed 
2024-10-13 13:18:18.694 Status: NetAtmoWeerEnLuchtkwaliteit: HomeStatus parsed 

...

2024-10-13 13:18:52.594 Status: NetAtmoWeerEnLuchtkwaliteit: HomeStatus parsed 
2024-10-13 13:18:52.946 Status: NetAtmoWeerEnLuchtkwaliteit: HomeStatus parsed 
2024-10-13 13:18:53.229 NetAtmoWeerEnLuchtkwaliteit: General/Text (Binnen - MAC-adres) 
2024-10-13 13:18:53.232 NetAtmoWeerEnLuchtkwaliteit: Temp + Humidity + Baro/Weather Station (Woonkamer - NetAtmo) 
2024-10-13 13:18:53.236 NetAtmoWeerEnLuchtkwaliteit: Air Quality/Voc (Woonkamer - CO2) 
2024-10-13 13:18:53.239 NetAtmoWeerEnLuchtkwaliteit: General/Sound Level (Geluid WK) 
2024-10-13 13:18:53.241 NetAtmoWeerEnLuchtkwaliteit: General/Text (Regenmeter - MAC-adres) 
2024-10-13 13:18:53.242 NetAtmoWeerEnLuchtkwaliteit: General/Custom Sensor ( Regenmeter Bat. Lvl) 
2024-10-13 13:18:53.244 NetAtmoWeerEnLuchtkwaliteit: General/Text (Regenmeter - Bridge) 
2024-10-13 13:18:53.248 NetAtmoWeerEnLuchtkwaliteit: Rain/RainByRate (Buiten: Regen) 
2024-10-13 13:18:53.250 NetAtmoWeerEnLuchtkwaliteit: General/Percentage ( Slaapkamer Oostzijde - Sig. + Bat. Lvl) 
2024-10-13 13:18:53.251 NetAtmoWeerEnLuchtkwaliteit: General/Text (Slaapkamer Oostzijde - MAC-adres) 
2024-10-13 13:18:53.253 NetAtmoWeerEnLuchtkwaliteit: General/Custom Sensor ( Slaapkamer Oostzijde Bat. Lvl) 
2024-10-13 13:18:53.255 NetAtmoWeerEnLuchtkwaliteit: General/Text (Slaapkamer Oostzijde - Bridge) 
2024-10-13 13:18:53.258 NetAtmoWeerEnLuchtkwaliteit: Temp + Humidity/LaCrosse TX3 (Slaapkamer Oostzijde:) 
2024-10-13 13:18:53.261 NetAtmoWeerEnLuchtkwaliteit: Air Quality/Voc (Slaapkamer Oostzijde - CO2) 
2024-10-13 13:18:53.262 NetAtmoWeerEnLuchtkwaliteit: General/Percentage ( Buiten - oost - Sig. + Bat. Lvl) 
2024-10-13 13:18:53.264 NetAtmoWeerEnLuchtkwaliteit: General/Text (Buiten - oost - MAC-adres) 
2024-10-13 13:18:53.266 NetAtmoWeerEnLuchtkwaliteit: General/Custom Sensor ( Buiten - oost Bat. Lvl) 
2024-10-13 13:18:53.268 NetAtmoWeerEnLuchtkwaliteit: General/Text (Buiten - oost - Bridge) 
2024-10-13 13:18:53.271 NetAtmoWeerEnLuchtkwaliteit: Temp + Humidity/LaCrosse TX3 (Buiten: Temp. en vocht) 
2024-10-13 13:18:53.274 NetAtmoWeerEnLuchtkwaliteit: General/Percentage ( Eetkamer/Keuken - Sig. + Bat. Lvl) 
2024-10-13 13:18:53.276 NetAtmoWeerEnLuchtkwaliteit: General/Text (Eetkamer/Keuken - MAC-adres) 
2024-10-13 13:18:53.279 NetAtmoWeerEnLuchtkwaliteit: General/Custom Sensor ( Eetkamer/Keuken Bat. Lvl) 
2024-10-13 13:18:53.286 NetAtmoWeerEnLuchtkwaliteit: General/Text (Eetkamer/Keuken - Bridge) 
2024-10-13 13:18:53.290 NetAtmoWeerEnLuchtkwaliteit: Temp + Humidity/LaCrosse TX3 (Eetkamer/Keuken -Temp. en vocht) 
2024-10-13 13:18:53.294 NetAtmoWeerEnLuchtkwaliteit: Air Quality/Voc (Eetkamer/Keuken - CO2) 
2024-10-13 13:18:53.295 Status: NetAtmoWeerEnLuchtkwaliteit: HomeStatus parsed 
2024-10-13 13:18:53.558 Error: NetAtmoWeerEnLuchtkwaliteit hardware (5) thread seems to have ended unexpectedly 
2024-10-13 13:18:53.564 NetAtmoWeerEnLuchtkwaliteit: General/Text (Binnen - MAC-adres) 

```

I never saw this before, so I think this is new since the current beta.

Note: The first part of the log shows only errors about unexpectedly ending thread. The second part it seems to start parsing, but no data is found. The last part is like I expect it to look (except for the last error, but this was really the last error).

Also note: This all seems to happen around 13:18:53. This was about the moment I refreshed the log page on my browser, so I'm not sure if this is a case of some kind of deadlock in the logging system and is not related to the issue you mentioned before. So, take it for the record and don't spill any time on it.

gizmocuz:
Yes that's what's happening when multiple persons are working on the same files
I asked Jurgen if he would like his PR merged

LoukKamp:
Yes, that's the risk (I took), but I'm sure Jurgen will solve this thread issue. 

In the meanwhile, I will take a look what is needed to trigger the migration tool (now a separate function) when it is not called from the local function UpdateValueInt() in Hardware/Netatmo.cpp. The send functions only take an integer value for the ID of the device, but I also need the unit, type, and subtypes to match old and new devices.

I'll try to do some kind of pull to merge Jurgens's changes and mine in advance, so I can test the migrations before merging to the beta.

JurgenLB:
With the HardwareBaseSensors I put a Exel with the differences for my setup.
I believe there are only differences for the Thermostat.
[Changes API Netatmo Domoticz.xlsx](https://github.com/user-attachments/files/17445768/Changes.API.Netatmo.Domoticz.xlsx)

Kind regards

LoukKamp:
I just committed my last changes for the "migration tool". I've tested this migrated with the development tree about 2 weeks ago (1 November 2024 01:52). Note: I only have a weather station with some extra modules and a wind meter, so I can't test other devices.

With some of the test runs, I experienced something that seems like a (database) deadlock situation: Several threads started to time out, and I needed to restart Domoticz several times. 

Because of this, I actually want the enable switch back in the UI, so the user has a choice to use this tool and be able to disable it when it causes problems. I also think that this tool should be removed before this beta release will be merged with the master.

Because of persenal issues, I won't be able to work as much on this tool as I would wish to at least for the comming 2 months. Sorry about that.

LoukKamp:
@gizmocuz : As far as I know, I did solve all review changes you requested for before, but it still seems to be waiting for something; do you know what?

JurgenLB:
why dit you change;
````
					if (m_bPollHomeStatus)
					{
						// GetHomesDataDetails
						GetHomeStatusDetails();
						Log(LOG_STATUS,"Status %d",  m_isLogged);
					}
````
to // GetHomesStatusDetails ???
This was a reminder that GetHomeStatusDetails makes a call to GetHomesDataDetails.
But now you have made it the same.

The 'boiler_boost' and some other values are in the respons from Netatmo but not used in the code.
I first had a sensor for everything but this is changed.

what do you mean by;
````

						//create update / domoticz device
						//??? Why using a different module here???
						SendSelectorSwitch(Hardware_int, 2, ssv.str(), moduleName + " - Schedule", 15, true, allSchName, allSchAction, true, m_Name); // No RF-level and Battery level
````

There is a selectorswitch for the schedule and the mode off the thermostat.

**Comment by LoukKamp** [comment]
> why dit you change;
> 
> ```
> 					if (m_bPollHomeStatus)
> 					{
> 						// GetHomesDataDetails
> 						GetHomeStatusDetails();
> 						Log(LOG_STATUS,"Status %d",  m_isLogged);
> 					}
> ```
> 
> to // GetHomesStatusDetails ??? This was a reminder that GetHomeStatusDetails makes a call to GetHomesDataDetails. But now you have made it the same.
> 

This was not my intention and will change this back.

> The 'boiler_boost' and some other values are in the respons from Netatmo but not used in the code. I first had a sensor for everything but this is changed.
> 
Are you still referring to the change above, or is this something else I need to change?
>
> what do you mean by;
> 
> ```
> 
> 						//create update / domoticz device
> 						//??? Why using a different module here???
> 						SendSelectorSwitch(Hardware_int, 2, ssv.str(), moduleName + " - Schedule", 15, true, allSchName, allSchAction, true, m_Name); // No RF-level and Battery level
> ```
>
 In other instances of _SendGeneralSwitch()_ you use _crcId_ for the device ID, but here _Hardware_int_; I just wondered why, or do I miss the point of this difference?
> There is a selectorswitch for the schedule and the mode off the thermostat.

There are more places in the code where I had questions about, all marked with 3 question marks as a reminder. Maybe you can take a look. When it is intended code, you can remove them or tell my to remove them, if you want.

**Comment by JurgenLB** [comment]
The ‘boiler_boost’ was a separate item.I will have a look at your questionsKind regards JurgenOp 17 nov 2024 om 09:46 heeft Louk ***@***.***> het volgende geschreven:﻿

why dit you change;
					if (m_bPollHomeStatus)
					{
						// GetHomesDataDetails
						GetHomeStatusDetails();
						Log(LOG_STATUS,"Status %d",  m_isLogged);
					}

to // GetHomesStatusDetails ??? This was a reminder that GetHomeStatusDetails makes a call to GetHomesDataDetails. But now you have made it the same.

This was not my intention and will change this back.

The 'boiler_boost' and some other values are in the respons from Netatmo but not used in the code. I first had a sensor for everything but this is changed.

Are you still referring to the change above, or is this something else I need to change?

what do you mean by;

						//create update / domoticz device
						//??? Why using a different module here???
						SendSelectorSwitch(Hardware_int, 2, ssv.str(), moduleName + " - Schedule", 15, true, allSchName, allSchAction, true, m_Name); // No RF-level and Battery level


In other instances of SendGeneralSwitch() you use crcId for the device ID, but here Hardware_int; I just wondered why, or do I miss the point of this difference?

There is a selectorswitch for the schedule and the mode off the thermostat.

There are more places in the code where I had questions about, all marked with 3 question marks as a reminder. Maybe you can take a look. When it is intended code, you can remove them or tell my to remove them, if you want.

**Comment by LoukKamp** [comment]
I think I forgot to rewind some patches I've made to test multiple percentage sensors.

At line 2515:
```

	//??? SendCustomSensor(ID, 2, batValue, static_cast<float>(rssiLevel), name + " - RF-level, ", " ", rssiLevel); // RF Percentage
	SendPercentageSensor(ID, 2, batValue, static_cast<float>(rssiLevel), name + " - RF-level, "); // RF Percentage

```
And at line 2992:
```

		//SendCustomSensor(crcId, 2, batteryLevel, mrf_percentage, pName, "  ", mrf_status);   // RF-level
		SendPercentageSensor(crcId, 2, batteryLevel, mrf_percentage, pName);

```
I can undo this changes, but wasn't it the result of a discussion to remove them at all? And the same for the battery levels? In that case, I wonder if you already implied that? And in that case, why didn't those changes end up in my code after the merge with the latest beta, I did this morning?

**Comment by LoukKamp** [comment]
About the Battery and RF levels: I see the point that the battery level and in some case also RF levels, are parameters of other send functions, but where are they shown? And can I show some kind of history? Maybe not that important for most people (not like my with a lot of RF trouble / disturbance in the house), but fort batteries this can be very useful information.

**Comment by JurgenLB** [comment]
for //??? Why using a different module here???
						SendSelectorSwitch(Hardware_int, 2, ssv.str(), moduleName + " - Schedule", 15, true, allSchName, allSchAction, true, m_Name); // No RF-level and Battery level


In other instances of SendGeneralSwitch() you use crcId for the device ID, but here Hardware_int; I just wondered why, or do I miss the point of this difference?

==> There is a selectorswitch for the schedule and the mode off the thermostat. so there is a need to have different ID because otherwise the mode and schedule are overwriting each other.

My last changes for deleting the device are not yet merged so they are not in your code.
But I do see that I forgot the Percentagesensor for the Battery-level.
The other unwanted sensors are removed in my GitHub-clone.

I do find a percentage-sensor for the battery-level useful, If gizmocuz  wants to accept them.

your comments
"//??? No send function to replace the line above???"
"//??? Why using a different module here???"
are not needed in the code. I believe that they are explained.

**Comment by LoukKamp** [comment]
> for //??? Why using a different module here??? SendSelectorSwitch(Hardware_int, 2, ssv.str(), moduleName + " - Schedule", 15, true, allSchName, allSchAction, true, m_Name); // No RF-level and Battery level
> 
> In other instances of SendGeneralSwitch() you use crcId for the device ID, but here Hardware_int; I just wondered why, or do I miss the point of this difference?
> 
> ==> There is a selectorswitch for the schedule and the mode off the thermostat. so there is a need to have different ID because otherwise the mode and schedule are overwriting each other.
>
Ok, clear.
> 
> My last changes for deleting the device are not yet merged so they are not in your code. But I do see that I forgot the Percentagesensor for the Battery-level. The other unwanted sensors are removed in my GitHub-clone.
> 
> I do find a percentage-sensor for the battery-level useful, If gizmocuz wants to accept them.
>
Me too...
> 
> your comments "//??? No send function to replace the line above???" "//??? Why using a different module here???" are not needed in the code. I believe that they are explained.

OK, I will remove them after the merge with your code.

**Comment by JurgenLB** [comment]
I believe line 2413 and 2418 are best in DEBUG_HARDWARE log not in 'LOG_STATUS' or comment out.

if you want to be consistent with my changes you must remove the following lines;
2528 to 2532
2537 to 2538
3013 to 3015
3032 to 3035

also comment out lines;
3082
3083
3473
3585
3586

kind regards

**Comment by LoukKamp** [comment]
> I believe line 2413 and 2418 are best in DEBUG_HARDWARE log not in 'LOG_STATUS' or comment out.
> 
Removed them (these were copies of the DEBUG_HARDWARE lines right above them). I'll add them again when it is required for logging of other users (who don't run in debug mode).
> if you want to be consistent with my changes you must remove the following lines; 2528 to 2532 2537 to 2538 3013 to 3015 3032 to 3035
> 
Done.
> also comment out lines; 3082 3083 3473 3585 3586
> 
Also done.
> kind regards

File: hardware/Netatmo.cpp [hunk 13]
```diff
@@ -53,6 +53,23 @@ std::string ReadFile(std::string filename)
 }
 #endif
 
+std::string prettifyJson(const Json::Value value)
```
gizmocuz, line 57: Could you uncomment this function as well as it is not used?
Or add a #define at the top to enable/disable debugging, or make it a member function of the hardware class please

hardware/Netatmo.cpp
```diff
@@ -504,12 +576,445 @@ uint64_t CNetatmo::UpdateValueInt(int HardwareID, const char* ID, unsigned char
...
+
+	return m_sql.TransferDevice(sOldDeviceId, sNewDeviceId);
+	 Log(LOG_STATUS, "Merge completed");
+}
+
+
+/// <summary>
+/// Merge logs in all log tables
+/// </summary>
+bool CNetatmo::MergeDeviceLogs (const std::string& spTableName, const std::string& spOldDeviceId, const std::string& spNewDeviceId)
```
gizmocuz: Could you remove the space behind the function name, or in visual studio press control-k followed by normal 'd'

gizmocuz (approved):
Looking great!

hardware/Netatmo.cpp
LoukKamp: Added #ifdef _DEBUG as it is only used for debugging purposes

hardware/Netatmo.cpp
LoukKamp: Done

LoukKamp:
@gizmocuz : 

Great, thanks. 

Can we agree on this migration flag being present while in Beta (not committed yet!!)? 

The migration tool is only useful for early adapters who used the beta versions in the beginning. So I want to remove the migration tool, or minimize it, before merging with the master release. Please let me know what you think.

Now I wait for JurgenJB to commit his changes when he is ready to do so, so I can merge his changes into my branch.

JurgenLB:
why the check is failing I don't know.

But can you change this line;
`//m_bPollGetEvents = false;  // Blocking GetEvents because off Error "User Usage"`

to match with my code.
`//m_bPollGetEvents = false;  // Blocking GetEvents because off Error`
or make it
`//m_bPollGetEvents = false;  // Possible to Block GetEvents because off Error`

line 438 must that be Flag ?
and line 2905 must that comment be //Not used ?

on line 3396 you marked 8??? this is the Temperature from the Thermostat?
it is now send by `SendTempSensor(crcId, batteryLevel, Temp, moduleName, mrf_status);`

LoukKamp:
> why the check is failing I don't know.

What check you mean??
> 
> But can you change this line; `//m_bPollGetEvents = false; // Blocking GetEvents because off Error "User Usage"`
> 
> to match with my code. `//m_bPollGetEvents = false; // Blocking GetEvents because off Error` or make it `//m_bPollGetEvents = false; // Possible to Block GetEvents because off Error`
> 
Although it's just comment, I choose for the second variant.

> line 438 must that be Flag ? and line 2905 must that comment be //Not used ?

Yes, changed that to flag at Line 438 and 2905  (thanks for noticing this typo in fiag (took me a while to see it myself ;) ))

> 
> on line 3396 you marked 8??? this is the Temperature from the Thermostat? it is now send by `SendTempSensor(crcId, batteryLevel, Temp, moduleName, mrf_status);`

I initially marked this, because 8 was not on your list of possible unit / sub-device numbers. But now you use _SendTempSensor()_ without specifying a unit number. Now I wonder what the unit number will be (0 or 1). This unit number should replace the 0 as the second parameter of _MigrateDevices()_  (line 3396?) .

JurgenLB:
with check I mean this;
                 check Failing


For the Temperature of the thermostat the unit number is not always the same ??;
unit thermostat
I have 2 thermostat and 3 valves in my House.

Kind regards

LoukKamp:
I needed to undo 27 commits to allow a merge: Now this PR is closed....

As this was absolute not my intention, I will reopen it now.

LoukKamp:
Thanks for the explanation: I thought the check was in the code :SHAME:

I resolved the merge conflicts and hope I did it all the right way (not undoing your recent changes); please check the last 2 commits.

LoukKamp:
> For the Temperature of the thermostat the unit number is not always the same ??; unit thermostat I have 2 thermostat and 3 valves in my House.

Far, Far, inside Domoticz, I found an explanation for this (some devices have a unit number and others not (0, or even fixed to 1):

Within the sendTemp*sensor functions, the device ID is split over two fields (ID1 andID2). Further on in the decode functions, this ID is reassembled to a single ID of four digits when presented in HEX..

However, in some of the decode functions the second part of the split ID, is also used to set the unit number:
```
uint8_t Unit = pResponse->TEMP_HUM_BARO.id2;
```

This is done in _decode_Temp()_, _decode_TempHumBaro()_, and _decode_TempRain()_.

But **not** in _decode_Hum()_, and _decode_TemBaro()_ (even set to a fixed value of 1), _decode_TempHum()_ (set fixed to 0).

So the call of _SendTempSensor()_ will use the last two digits (when presented in HEX) from the device ID as the unit number.

JurgenLB:
some remarks;
you have 'initializing': conversion from 'unsigned long' to 'unsigned char' error on line 683, 684, 685, 686

On line 3028 there is a 'SendPercentageSensor' but on the next line you migrate to 'sTypeCustom'

on line 3396 I would remove the ???

LoukKamp:
> some remarks; you have 'initializing': conversion from 'unsigned long' to 'unsigned char' error on line 683, 684, 685, 686

Strange, my compiler doesn't complain. However, I added some casting (unsigned char): I hope this will be sufficient.
> 
> On line 3028 there is a 'SendPercentageSensor' but on the next line you migrate to 'sTypeCustom'

I already marked this with question marks (I probably wasn't sure what to do). Changed this.
> 
> on line 3396 I would remove the ???

Removed.

JurgenLB:
sorry but now I notice on line 1686 you have a second `m_homeid.clear();`
this is already done on line 1985. I believe Your second one will block the multiple Homes in GetHomeStatusDetails

and on line 3100 you can remove that because after a conversion the SendAlertSensor is called this is the replacement off that previous UpdateValueInt call.

LoukKamp:
> sorry but now I notice on line 1686 you have a second `m_homeid.clear();` this is already done on line 1985. I believe Your second one will block the multiple Homes in GetHomeStatusDetails

Yes, I see you've added the same line too. But no, it won't block multiple homes in because the stack is filled just after the clear at line 1868: 
```
1695:         m_homeid.push_back(homeID);

```
Actually, the two clears are done almost right after each other, so there is no use for the second one anymore. I will remove it.
> 
> and on line 3100 you can remove that because after a conversion the SendAlertSensor is called this is the replacement off that previous UpdateValueInt call.

When you mean that the call to _SendAlertSensor ()_ replaces both instances of _UpdateValueInt()_, it is clear to me (I'll remove the question).

But there is still one question left at line 3083: There is no corresponding send-function (not even commented out) for the boiler comfort boost status.

JurgenLB:
yes the boiler comfort boost status I have not used anymore,
if someone need it it can be `SendSwitch(crcId, 9, batteryLevel, bIsActive, Level, bName, m_Name, mrf_status);`
But I do not find it useful.

LoukKamp:
Ok, I drop that one too.

LoukKamp:
Also checked the logging: The removal of line 1686 doesn't seem to have any impact (as expected), so this change is OK.

gizmocuz:
Thank you very much for the PR!

LoukKamp:
TOO FAST!!!

Sorry, but I was waiting for an answer of @gizmocuz about a temporary addition of a migration tool enable flag in the UI.

The only way to enable the tool now, is to set the mod2 flag of the Netatmo device in the database by hand.

> @gizmocuz :
> 
> Great, thanks.
> 
> Can we agree on this migration flag being present while in Beta (not committed yet!!)?
> 
> The migration tool is only useful for early adapters who used the beta versions in the beginning. So I want to remove the migration tool, or minimize it, before merging with the master release. Please let me know what you think.
> 
> Now I wait for JurgenJB to commit his changes when he is ready to do so, so I can merge his changes into my branch.

I've prepared a change to add the flag to the UI, when you agree to this. Otherwise, we have to think about another solution.

My suggestion still is: Add it now and remove (all of) it before merging to the main branch.

gizmocuz:
How many devices are we talking about? If it's about 10, the user can do the migration manually.
I think I already mentioned you can use a version number in the preferences table (netamo hardware version), and with this see if there is any migration to be done.
Nothing to be added in the user interface, and no confusion users

LoukKamp:
Yes, I know, we had this discussion before. However, because I can't test all the devices involved (I only own a weather station), I don't want to start this tool for every user. Just the early adaptors who end up with a discontinued logging and who really want to make use of the tool. And, for the case something goes wrong, I want the user to have the option to disable the tool.

Depending on the number of submodules, the weather station has about 10 to 15 devices, although not all of them need migration. I don't know about the other devices (security and thermostats).

gizmocuz:
So, there are not many devices to migrate.
It's not that hard to transfer old to new devices in Domoticz... maybe just leave it to the user?

LoukKamp:
As you wish; I'll remove the software.

JurgenLB:
OK we leave this migration to the users,
but the PR is merged at the moment ?

PR #6179, opened 2024-10-05
Netatmo with HardwareBase Sensors

JurgenLB:
I changed the sensors to use the HardwareBasesensors.

Only the MAC-adres identifiers and the Rain-sensor is still on Updatevalueint
The Rainsensor is just to verify the correct working of the data.

Also I enabled back the Events and for now the error 'User usage' is not happening anymore 
If this error comes back the events need to be block again.

### DIFF
--- a/hardware/Netatmo.cpp
+++ b/hardware/Netatmo.cpp
@@ -2126,7 +2126,7 @@ bool CNetatmo::ParseDashboard(const Json::Value& root, const int DevIdx, const i
 		y << ";";
 		y << Temp_outdoor;
 		y << ";";
-		y << WindChill_c;
+		y << wind_Chill;
 		std::string sValue = y.str().c_str();
 		// sValue: "<WindDirDegrees>;<WindDirText>;<WindAveMeterPerSecond*10>;<WindGustMeterPerSecond*10>;<Temp_c>;<WindChill_c>"
 		SendWind(ID, batValue, wind_angle, wind_strength, wind_gust, wind_Temp, wind_Chill, false, false, name, rssiLevel);
--- a/hardware/Netatmo.cpp
+++ b/hardware/Netatmo.cpp
@@ -1871,11 +1871,11 @@ bool CNetatmo::ParseDashboard(const Json::Value& root, const int DevIdx, const i
 			m_bNetatmoRefreshed[ID] = true;
 		// Check when dashboard data was last updated
 		if (!root["time_utc"].empty())
-			tNetatmoLastUpdate = root["time_utc"].asUInt();
-		//Debug(DEBUG_HARDWARE, "Module [%s] last update = %s", name.c_str(), ctime(&tNetatmoLastUpdate));
+			tNetatmoLastUpdate = root["time_utc"].asUInt(); //["dashboard_data"]["time_utc"].asUInt();
+		Debug(DEBUG_HARDWARE, "Module [%s] last update = %s", name.c_str(), ctime(&tNetatmoLastUpdate));
 		// check if Netatmo data was updated in the past NETAMO_POLL_INTERVALL (+1 min for sync time lags)... if not means sensors failed to send to cloud
 		int Interval = NETAMO_POLL_INTERVALL + 60;
-		//Debug(DEBUG_HARDWARE, "Module [%s] Interval = %d", name.c_str(), Interval);
+		Debug(DEBUG_HARDWARE, "Module [%s] Interval = %d", name.c_str(), Interval);
 		if (tNetatmoLastUpdate > (tNow - Interval))
 		{
 			if (!m_bNetatmoRefreshed[ID])
@@ -1887,7 +1887,7 @@ bool CNetatmo::ParseDashboard(const Json::Value& root, const int DevIdx, const i
 		else
 		{
 			if (m_bNetatmoRefreshed[ID])
-				Log(LOG_ERROR, "cloud Dashboard for module [%s] no longer updated (module possibly disconnected)", name.c_str());
+				Log(LOG_ERROR, "cloud Dashboard for module [%s] no longer updated (module possibly disconnected) since %s", name.c_str(), ctime(&tNetatmoLastUpdate));
 			m_bNetatmoRefreshed[ID] = false;
 			return false;
 		}
@@ -1989,7 +1989,9 @@ bool CNetatmo::ParseDashboard(const Json::Value& root, const int DevIdx, const i
 	RF_level << rssiLevel;
 	RF_level >> sValue;
 	std::string module_name  = name + " RF. Lvl";
-	UpdateValueInt(0, str_ID.c_str(), 2, pTypeGeneral, sTypePercentage, rssiLevel, batValue, '0', sValue.c_str(), module_name, 0, m_Name); // RF Percentage
+	SendCustomSensor(ID, 3, batValue, rssiLevel, name + " - RF-level, ", "% ", rssiLevel);
+        SendPercentageSensor(ID, 3, batValue, batValue, name + " - Bat. Level");
+	//UpdateValueInt(0, str_ID.c_str(), 2, pTypeGeneral, sTypePercentage, rssiLevel, batValue, '0', sValue.c_str(), module_name, 0, m_Name); // RF Percentage
 
 	//Temperature and humidity sensors
 	if (bHaveTemp && bHaveHum && bHaveBaro)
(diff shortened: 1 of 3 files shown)


LoukKamp:
Ok, this leads to a conflict between our changes: I used the function Updatevalueint() to trigger the migration tool (is part of this function.

Now I will remove the code from Updatevalueint() and create a new function for this and call this from Updatevalueint().
However, it should also be called after any call to a send* function. Maybe the commented out calls to Updatevalueint() will be a mark for me to add a call to the migration tool, so please leave them there until I'm done with the merge...

JurgenLB:
Yes I leave the 'updatevalueint' for reference for the moment.

The block for the 'gegevenst' was only removed on my local installation, that is not in this code. But the 'user usage error ' is returning.
So no change there.

The error from the thread was not happening for a few days but now it's back.

That error will need a closer look.

JurgenLB:
Yes I leave the 'updatevalueint' for reference for the moment.

The block for the 'gegevenst' was only removed on my local installation, that is not in this code. But the 'user usage error ' is returning.
So no change there.

The error from the thread was not happening for a few days but now it's back.

That error will need a closer look.

gizmocuz:
Please let me know when you would like to merge this

JurgenLB:
I'm just waiting on some more days of rain so I can compare the different rain-sensors.
If you want to merg this it is to prevent the errors from the hardware thread.

LoukKamp:
While waiting, can you please check this out:

 ```
               SendPercentageSensor(ID, 3, batValue, static_cast<float>(batValue), name + " - Bat. Level");
                //UpdateValueInt(0, str_ID.c_str(), 2, pTypeGeneral, sTypePercentage, rssiLevel, batValue, '0', sValue.c_str(), module_name,  0, m_Name); // RF Percentage

               CNetatmo::MigrateDevices(str_ID.c_str(), 3, pTypeGeneral, sTypePercentage, module_name.);
 
```
In the previous version you called UpdateValueInt() with unit = 2, but in the call to SendPercentageSensor() you used 3 for the ChildID parameter. I think this is not intended.

UPDATE: I think this is the intention:

```
        SendCustomSensor(ID, **2**, batValue, static_cast<float>(rssiLevel), name + " - RF-level, ", " ", rssiLevel);
        CNetatmo::MigrateDevices(str_ID.c_str(), 2, pTypeGeneral, sTypeCustom, name  + " - RF-level");
        if (batValue != 255)
        {
                SendPercentageSensor(ID, **2**, batValue, static_cast<float>(batValue), name + " - Bat. Level");
                CNetatmo::MigrateDevices(str_ID.c_str(), 2,  pTypeGeneral, sTypePercentage, name + " - Bat. Level");
        }
        //UpdateValueInt(0, str_ID.c_str(), 2, pTypeGeneral, sTypePercentage, rssiLevel, batValue, '0', sValue.c_str(), module_name, 0, m_Name);  // RF Percentage

```
However, I'm still confused: In the previous version (using UpdateValueInt() in all cases), The Bat. Lvl was a Custom type (31) and the RF-Lvl a Percentage type (6). In your latest update, it is the other way around...

**Comment by JurgenLB** [comment]
The 'custemsensor' is the RF-level and the 'percentagesensor' is the battery-level
There is indeed a incorrect number for the child ID.
I have the following logic;
- 0 Sensor data
- 1 MAC-adres
- 2 RF-Level
- 3 Battery-Level
- 4 Bridge MAC-adres
- 5 Doorsensor Switch (Not in use anymore)
- 6 Doorsensor Alert
- 7 Setpoint

The following is not in use for the moment;
- 8 Temperature Thermostat
- 9 Boiler Status Switch
- 10 Selector Switch
- 11 Events Text
- 12 Events Alert

Kind regards

File: hardware/Netatmo.cpp [hunk 1]
```diff
@@ -2090,11 +2090,11 @@ bool CNetatmo::ParseDashboard(const Json::Value& root, const int DevIdx, const i
 		v << ";";
 		v << rain_24;
 		std::string sValue = v.str().c_str();
-		SendRainSensor(ID, batValue, m_RainOffset[ID] + m_OldRainCounter[ID], name, rssiLevel);
+		SendRainSensor(5, batValue, m_RainOffset[ID] + m_OldRainCounter[ID], name, rssiLevel);
```
LoukKamp: Why do you use a fixed ID for this device?

hardware/Netatmo.cpp
```diff
@@ -2090,11 +2090,11 @@ bool CNetatmo::ParseDashboard(const Json::Value& root, const int DevIdx, const i
...
 		std::string Name = name + "-updatevalueint";
                 ///Debug(DEBUG_HARDWARE, "(%d) %s (%s) [%s] rain %s %s %d %d", Hardware_int, str_ID.c_str(), pchar_ID, name.c_str(), sValue.c_str(), m_Name.c_str(), rssiLevel, batValue);
 		//UpdateValueInt(0, str_ID.c_str(), 0, pTypeRAIN, sTypeRAINByRate, rssiLevel, batValue, '0', v.str().c_str(), Name, 0, m_Name);
-		SendRainSensor(ID, batValue, rain_1 + rain_24, name + "- rain", rssiLevel);
+		SendRainSensor(ID, batValue, rain_24 + rain_1, name + "- rain", rssiLevel);
```
LoukKamp: Why do you use the ID and not a fixed ID of 5 as above (for the same device)?

Note: This is a device ID, not a type...

hardware/Netatmo.cpp
```diff
@@ -2766,8 +2765,8 @@ bool CNetatmo::ParseHomeStatus(const std::string& sResult, Json::Value& root, st
 
 						//Debug(DEBUG_HARDWARE, "(%d) %s (%s) [%s] rain %s %s %d %d", Hardware_int, str_ID.c_str(), pchar_ID, name.c_str(), v.str().c_str(), m_Name.c_str(), mrf_status, batteryLevel);
 						//UpdateValueInt(0, ID.c_str(), 0, pTypeRAIN, sTypeRAINByRate, mrf_status, batteryLevel, '0', v.str().c_str(), moduleName, 0, m_Name);
-						SendRainSensor(crcId, batteryLevel, m_RainOffset[crcId] + m_OldRainCounter[crcId], moduleName, mrf_status);
-						SendRainSensor(crcId, batteryLevel, rain_1 + rain_24, moduleName + "-rain", mrf_status);
+						SendRainSensor(7, batteryLevel, m_RainOffset[crcId] + m_OldRainCounter[crcId], moduleName, mrf_status);
```
LoukKamp: Same as comments above: Why the different (device) ID's?

JurgenLB: the First 'SendRainSensor' is from the old_API but this gives incorrect data compared to the data from NetatmoApp.
That is also the reason I temporally set the fixt-ID value.
When I have correct data this will be removed.

JurgenLB:
Now this is ready with the correct rain data.

JurgenLB:
In Mainworker the Log line;
`_log.Log(LOG_STATUS, "%d last checking  %.2lf seconds ago", iterator->first, dif);`
is not correct,
my proposal is to change this in;
`_log.Log(LOG_STATUS, "%s last checking  %.2lf seconds ago", pHardware->m_Name.c_str(), diff);`

gizmocuz:
> In Mainworker the Log line; `_log.Log(LOG_STATUS, "%d last checking %.2lf seconds ago", iterator->first, dif);` is not correct, my proposal is to change this in; `_log.Log(LOG_STATUS, "%s last checking %.2lf seconds ago", pHardware->m_Name.c_str(), diff);`

Yes please, but make sure it stays commented out

Also, is the retrieval of the name from the database needed a few lines below? I think not...?

JurgenLB:
I only changed 13808 to 13812, but apparently there is a big difference in mainworker.cpp

on my local setup the changed lines are working.

main/mainworker.cpp
```diff
@@ -12988,8 +12988,8 @@ bool MainWorker::SetSetPointInt(const std::vector<std::string>& sd, const float
 
 		if (
 			(value_unit.empty())
-			|| (value_unit == "�C")
-			|| (value_unit == "�F")
+			|| (value_unit == "?C")
```
gizmocuz: Could you undo your change here? Try using an editor that supports UTF8, also for the line below here

main/mainworker.cpp
```diff
@@ -12988,8 +12988,8 @@ bool MainWorker::SetSetPointInt(const std::vector<std::string>& sd, const float
 
 		if (
 			(value_unit.empty())
-			|| (value_unit == "�C")
-			|| (value_unit == "�F")
+			|| (value_unit == "°C")
```
gizmocuz: Thank you for reverting this.
But github still thinks there is something changed otherwise it would not show the delete/addition

JurgenLB: I will look into this.

JurgenLB:
I think now it is fine

gizmocuz:
It is! Let me know when you want this merged!

JurgenLB:
for me you can merge this.

gizmocuz:
Thank you!!

PR #6183, opened 2024-10-16
Fix for Meter computed summation

badzz:
When we update a meter widget that is compute mode, domoticz does the computation based on the lastupdate of the widget However if some options of the widget are modified as well (eg: signallevel), then the widget will be updated before we update the meter widget, hence generating a wrong summation as the lastupdate has just been updated.
Hence this fix revert the logic to update the widget main value first then update the options.
I suppose the same fix is needed on the extended api but I have no way to test it thouroughly.

### DIFF
--- a/hardware/plugins/PythonObjects.cpp
+++ b/hardware/plugins/PythonObjects.cpp
@@ -979,137 +979,6 @@ namespace Plugins {
 
 			std::string sID = std::to_string(self->ID);
 
-			// Name change
-			if (Name)
-			{
-				sName = Name;
-				Py_BEGIN_ALLOW_THREADS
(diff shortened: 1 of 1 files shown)


waltervl:
Good you have found this. I now run a dummy device and dzvents script to calculate the Meter Energy (kWh) values myself because I could not find the root cause of the sometimes strange deviations. It is a little bit difficult to see what you have changed as the delta is a complete code block.....

badzz:
Display the fix in split mode. It is maybe more readable. I only move a
block of code in a different place. No real code change by itself.

https://github.com/domoticz/domoticz/pull/6183/commits/84c37987668adb235dd4c72ad277622027d1c410?diff=split&w=0

gizmocuz:
Thank you!

PR #6189, opened 2024-10-24
Blebox patch app veyor

JurgenLB:
If you like,
This is to reduce the AppVeyor Warnings.
The SQL function returns a  'uint64_t' and the variable IDX is 'int'

kind regards

### DIFF
--- a/notifications/NotificationKodi.cpp
+++ b/notifications/NotificationKodi.cpp
@@ -308,6 +308,8 @@ bool CNotificationKodi::SendMessageImplementation(
 			_log.Log(LOG_ERROR, "%s", logline.str().c_str());
 			return false;
 		}
+
+		closesocket(_Sock);
 	}
 	return true;
 }


gizmocuz:
Thank you for the PR.
Yeah, it's bugging me for quite some time as well and hoped the author was going to solve this eventually

The problem is not that it needs to be casted to an 'int' but the variable should be a uint64_t, maybe it's possible to change that?

JurgenLB:
I have changed it to 'uint64_t' and then '%n'

kind regards

gizmocuz:
Thank you for this!

But %n is not the right way to print a 64bit number, better use:

for int64_t type:

```
int64_t t;
printf("%" PRId64 "\n", t);
```

for uint64_t type:

```
uint64_t t;
printf("%" PRIu64 "\n", t);
```

(You do not need the \n at the end)

JurgenLB:
i will change the Log line.
and the `m_sql.safe_query("UPDATE DeviceStatus SET StrParam1='%d', StrParam2='%q' WHERE (ID==%n)", deviceTypeID, IPAddress.c_str(), idx);`
is that right for you?

gizmocuz:
don't use %n, it's not used through the entire code.

it would become something like:

`m_sql.safe_query("UPDATE DeviceStatus SET StrParam1='%d', StrParam2='%q' WHERE (ID==%" PRIu64 ")", deviceTypeID, IPAddress.c_str(), idx);`

If idx is a uint64_t

JurgenLB:
The warnings from AppVeyor about Netatmo are resolved by #6179 I believe.

Kind regards

Issue #6176, opened 2024-09-22
Visual Crossing doesn't connect anymore with newest version


Olifant1990:
Last week I updated to version 2024_7.16253 (running on a fully updated Windows 10 PC).
I got an error with Visual Crossing: "2024-09-22 21:57:45.811 Error: Visual Crossing: Error getting http data!."
So, mark the dot at the end, it brings me to https://github.com/domoticz/domoticz/blob/03baca33331c03c5cfd295ec546a2eca151b17c8/hardware/VisualCrossing.cpp#L133

But... When I run the latest code on my development PC (Windows 11), build with Visual Studio, I do not get this error and all is working fine.
Any ideas where to look for this?
Problem is still present in current beta 2024_7.16265

waltervl:
Seems an internet connection issue on your W10 machine.
What do you get on this link when used on the browser on that W10 machine. replace szLoc  with the contents of your location used in intergration or Domoticz settings.
And change API field
`https://weather.visualcrossing.com/VisualCrossingWebServices/rest/services/timeline/" << szLoc << "?key=" << m_APIKey << "&unitGroup=metric&include=current`

Olifant1990:
Sorry, I forgot to mention that the direct JSON link IS working in a browser.
Same as that version 2024_7.16187 is working, but newer aren't.

kiddigital:
@Olifant1990 , as the connection is using HTTPS (so SSL), there is a good chance that your W10 PC does not has the right SSL library available for Domoticz.

The browser has often a internal SSL library.

Olifant1990:
Then somewhere between version 2024_7.16187 and 2024_7.16253 that must have been changed.
As I stick to 2024_7.16187 for now and it is working for now.

Olifant1990:
Any ideas on how to fix this?

gizmocuz:
The last code change for this hardware class has been made on 9 September 2023... (after this only one change that has no impact)
So maybe there has been a change on your side?

It is also a very small class with not many code lines... Maybe you can have a look at it: VisualCrossing.cpp

Olifant1990:
Well, as said, It is not a problem of VisualCrossing itself. I know the code as I am the writer of it :P
But... Also as said, see pointed line in first post.
if (!HTTPClient::GET(sURL.str(), sResult)) is giving false in later versions, so I saw that error comming from visual crossing.
I went back to 2024_7.16187 and it is still working for now on that version.
[...]

PR #6201, opened 2024-11-03
Update VisualCrossing.cpp

Olifant1990:
Added used URL to get more info why it is failing on win10 with error "Bad API Request:Invalid location parameter value"

https://github.com/domoticz/domoticz/issues/6176

### DIFF
--- a/hardware/VisualCrossing.cpp
+++ b/hardware/VisualCrossing.cpp
@@ -125,7 +125,7 @@ void CVisualCrossing::GetMeterDetails()
 #else
 	std::stringstream sURL;
 	std::string szLoc = CURLEncode::URLEncode(m_Location);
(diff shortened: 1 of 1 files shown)


gizmocuz:
Sorry, there is a hardcoded number in your PR (location)
And as there is only one call, you know the complete URL i suppose?

Olifant1990:
Whoops, fixed!
Yes, I should know the URL, but Windows 10 is making it a bit different somewhere I guess...

gizmocuz:
I don't think it is wise to log the complete url/api key.
Everyone can read the log if it is written to disk, or via the GUI
So you can add a deeper log message via DEBUG_HARDWARE if you wish, but as mentioned, you already know the complete URL, it is only one URL, so it seems not needed to log this?

Olifant1990:
Yeah right again, still trying to get a finger on the problem xD.
Indeed I want to know what my PC is making of that URL, but not wise to do this for everyone in the log sorry!
Will have another look on how to debug this further

gizmocuz:
Hope you will find the issue! Thanks for looking!

PR #6206, opened 2024-11-11
Netatmo Events active

JurgenLB:
some code adjusting to have all possible data sensors


gizmocuz:
Is the battery/RF level not available when updating a temp sensor?

JurgenLB:
yes they are as arguments by a temperature sensor, only not by all other sensors for example setpoint from the termostat.
Those arguments are not available to a script as far that I know, this is useful for the battery-level.
Can I just comment them out or do you want me to remove them?

gizmocuz:
Yes, please remove them.

I added battery level for SetPoint sensors in the below PR, it is already merged into the codebase but no beta's have been build.
Could you see if this is working out for you? When you 'set' a setpoint, the battery level will become invalid/NA (255), but it should be received again shortly from the real hardware

https://github.com/domoticz/domoticz/pull/6208

JurgenLB:
I have rebuild my test setup and the setpoint sensors have now battery levels.
I will check the rest of the sensors in the coming days.

JurgenLB:
I believe this can be merged.

kind regards

gizmocuz:
Thank you again!

PR #6207, opened 2024-11-12
Fix a typo on initial sValue for General/kWh typename

FlyingDomotic:
Suppress a space after semi-column


gizmocuz:
Thank you for this PR!

It is included in beta 16328

PR #6209, opened 2024-11-15
Allow to optionally force lastUpdate for SendSwitch and use it for Enocean Smoke Detector

ludos78:
Smoke Detector status nearly never change, but sends a periodic status as a freshness info (for enocean F6-05-02, During "Alarm OFF" state, the "Alarm OFF" message shall be repeated every 20 minutes). 
This allows to update the associated 'lastUpdate' field of the db to know that the detector is still alive.

note : revival of a part of pretty old PR (#5143) i never took time to update..

### DIFF
--- a/hardware/DomoticzHardware.cpp
+++ b/hardware/DomoticzHardware.cpp
@@ -675,7 +675,7 @@ void CDomoticzHardwareBase::SendSwitchUnchecked(int NodeID, uint8_t ChildID, int
 
 
 void CDomoticzHardwareBase::SendSwitch(const int NodeID, const uint8_t ChildID, const int BatteryLevel, const bool bOn, const double Level, const std::string &defaultname, const std::string &userName,
-				       const int RssiLevel /* =12 */)
+				       const int RssiLevel /* =12 */, const bool bForceLastUpdate /* =false */)
 {
 	double rlevel = (16.0 / 100.0) * Level;
 	int level = int(rlevel);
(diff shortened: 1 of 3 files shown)


gizmocuz:
Thanks for the PR.
Another approach (maybe better) is to call m_sql.UpdateLastUpdate ?

ludos78:
This clearly would be better to call m_sql.UpdateLastUpdate, i didn't know it, i'll try this.

Just another question, i see that the Battery_level is not taken into account for the change logic detection in the SendSwitch function. That mean in my case with smoke sensor, as the bOn value is nearly always the same, i will never see the change in battery level. I would suggest to also consider it for testing if there is a change, but i don't know if it could be harmful for other part of code that also use SendSwitch, even if a lot of caller seem to use a fixed value. Do you think i can also propose this change in this PR ?

gizmocuz:
Maybe you could call UpdateLastUpdate as well?

Could you undo the code changes in the base hardware class?

gizmocuz:
Do you still want to make the change?

ludos78:
i need to change this a bit to also include the battery level update for the smoke detector. Will try to test and update this soon

ludos78:
Just pushed a new commit, now using UpdateDeviceValue to update at least BatteryLevel along with lastupdate if nothing else has changed and the optional parameter bForceDbUpdate is set (only set in smokesensor code for enocean). This should not break any other code as the parameter is optional and default to false. 
Is this acceptable like this ?

gizmocuz:
Maybe you can use the 'SendSwitchUnchecked' ?

ludos78:
That could be an option, but in the former PR (https://github.com/domoticz/domoticz/pull/5143#issuecomment-1037062131) you said this will trigger On/Off event. Is it ok if this happen at each message received (every 20 minutes) ? If yes, i think i'll drop this PR and open another one with only the change to use SendSwitchUnchecked instead of SendSwitch for the EnOcean Smoke Meter.

gizmocuz:
Yes, this indeed causes an update every time.
But also updating the lastupdate will cause this

The difference is, if you use the smoke detector to trigger other devices/groups/scenes they will get triggered if you use 'SendSwitchUnchecked'
Updating the 'lastupdate' will cause the web gui/websocket to be updated

There are however exceptions for this...

SQLHelper.cpp, line 5412, the status is previous state is ignored if the type is a type contact,doorlock.
I think if we add STYPE_SMOKEDETECTOR to this list, all will be well

I made a change in the latest code to allow this

You should now be able to use 'SendSwitchUnchecked'

It should not fire any on/off actions or trigger groups/scenes if the status did not change

PR #6213, opened 2024-11-20
Enocean use baseid as senderid

ludos78:
With this change, one can optionally choose to use the base id of the enocean dongle as sender id, instead of the chip id.
This can be usefull for dongle replacement without the need to re-associate every controller (see https://www.enocean.com/en/faq-knowledge-base/what-is-difference-between-base-id-and-chip-id/).

Note :
- changing the base id of the dongle is out of scope (i did it only once using some python script, and it can be done only a limited number of times). 
- this allowed me to use domoticz along to an existing solution provided in my flat by the builder.

### DIFF
--- a/hardware/EnOceanESP3.cpp
+++ b/hardware/EnOceanESP3.cpp
@@ -316,6 +316,7 @@ CEnOceanESP3::CEnOceanESP3(const int ID, const std::string &devname, const int t
 	m_Type = type;
 	m_id_base = 0;
 	m_id_chip = 0;
+	m_id_src = 0;
 }
 
 bool CEnOceanESP3::StartHardware()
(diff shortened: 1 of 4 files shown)


gizmocuz:
Thank you for this PR.

The default is to use the base_id as source_id? If this is the case, add a default checked option in the hardware setup.
Just wanting to make sure that existing users have no issues when they leave everything as-is

gizmocuz:
@ludos78, Could you tell if there won't be any changes for existing users?

ludos78:
In the PR, in the second commit, there is a new checkbox 'usebaseid' in hardware setup for Enocean ESP3 that is unchecked by default, linked to the 'Mode1' parameter that was unused up to now for this hardware type.
So by default for existing users, behaviour is unchanged : m_type is 0 and m_id_chip is used for sending frames as before those changes.

gizmocuz:
Thank you ! And thanks for the PR!!

PR #6224, opened 2024-12-13
Update EvohomeWeb.cpp

Olifant1990:
Honeywell endpoint has been replaced by Resideo endpoints.

Honeywell endpoint is still working, but SSL certificate doesn't contain the domain of honeywell.com anymore


gizmocuz:
Thank you!!!

PR #6225, opened 2024-12-14
Update JWT-CPP lib (ready for openSSL3)

kiddigital:
This PR updates the included [jwt-cpp](https://github.com/Thalhammer/jwt-cpp) lib.

For (future) testing purposes some basic tests of the JWT library as used in domoticz are included in the `domoticztester` and a few _gherkin_ tests are added.

Before, the library used _picosjon_ internally, but now it uses the jsoncpp lib as part of domoticz.

This should also make the library ready for switching to OpenSSL3.

Reviewing and (more) testing needed.


gizmocuz (commented):
Looking good. One question, instead of adding the JWT-CPP code to the Domoticz branch, we could use it like the other 'extern' projects?
And for Windows, it is available in the VCPKG packages
Would that be an idea?

kiddigital:
> Looking good. One question, instead of adding the JWT-CPP code to the Domoticz branch, we could use it like the other 'extern' projects? And for Windows, it is available in the VCPKG packages Would that be an idea?

@gizmocuz , definitely a good idea. I think we tried that initially when we added the lib the first time, but than there were a few issues as far as I remember. But seems a good idea to retry and see if we can keep it external.

Would you give it a try as I don't have a Windows build machine at hand :) ?
You could use the `domoticztester`, with the additions in this PR branch to see if the lib works as expected.

gizmocuz:
I will give it a go asap... hope tomorrow, else it's going to be after Christmas I'm afraid

gizmocuz:
Thanks for the PR, I rebuild the code to use the external library

PR #6227, opened 2024-12-23
Update Netatmo

JurgenLB:
This is to solve some trouble with undefined selectorswitch and history off rainsensor.

PS. Do you want me to update the wiki also?

### DIFF
--- a/hardware/Netatmo.cpp
+++ b/hardware/Netatmo.cpp
@@ -110,6 +110,7 @@ void CNetatmo::Init()
 	m_Module_Bat_Level.clear();
 	m_Module_RF_Level.clear();
 	m_Types.clear();
+	m_Device_types.clear();
 	m_Module_category.clear();
 	m_thermostatModuleID.clear();
 	m_ScheduleNames.clear();
@@ -665,7 +666,8 @@ bool CNetatmo::SetProgramState(const int uid, const int newState)
 	std::string home_data;
 	bool bRet;              //Parsing status
 	bool bHaveDevice;
-	std::string module_id = m_DeviceModuleID[uid];
+	std::string module_id = m_DeviceModuleID[uid];            // mac-adres
+	std::string type_module = m_Device_types[module_id];
 	std::string type_module = m_thermostatModuleID[uid];
 	std::string name = m_ModuleNames[module_id];
 	std::string roomNetatmoID = m_RoomIDs[module_id];
@@ -706,15 +708,24 @@ bool CNetatmo::SetProgramState(const int uid, const int newState)
 			return false;
 		}
 		bHaveDevice = true;
+		// Energy        {NAPlug, OTH, BNS}
+		// Home+control  {NLG,    OTH, BNS, NBG,              BNMH}
+		// Home+security {NACamera, NOC, NDB, NSD, NCO, BNCX, BNMH}
+
 		// NAPlug = Netatmo Thermostat
-		// OTH = Opentherm Thermostat Relay
 		// BNS = Smarther with Netatmo Thermostat
+		// NLG = Gateway
+		// OTH = Opentherm Thermostat Relay
+		// NBG = 
+		// BNMH = Bticino MyHome server 1
 		//
+		std::string Type = "NAPlug";
+		Debug(DEBUG_HARDWARE, "Device Type %s =  %s ??", Type.c_str(), type_module.c_str());
 		// When a thermostat mode is changed all thermostat/valves in Home are changed by Netatmo
 		// So we have to update the corresponding devices       Device_Types {NAPlug, OTH, BNS}
 		// https://api.netatmo.com/api/homestatus?home_id=xxxxx&device_types=NAPlug
-		std::string Type = "NAPlug";
-		std::string _data = "home_id=" + Home_id + "&device_types=" + Type ;
+		//
+		std::string _data = "home_id=" + Home_id + "&device_types=" + type_module ;
 		//Debug(DEBUG_HARDWARE, "Home_Data = %s ", _data.c_str());
 		Get_Respons_API(NETYPE_STATUS, sResult, _data, bRet, root, "");
 		//Parse API response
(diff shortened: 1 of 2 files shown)


hardware/Netatmo.cpp
```diff
@@ -1363,7 +1383,9 @@ void CNetatmo::GetHomesDataDetails()
 						m_ScheduleNames[index] = schedule["name"].asString();
 						m_ScheduleIDs[index] = schedule_id;
 						m_DeviceHomeID[schedule_id] = homeID;
-						if (!schedule["selected"].empty() && schedule["selected"].asBool())
+						Debug(DEBUG_HARDWARE, "Schedule empty %d", !schedule["selected"].empty());
+						Debug(DEBUG_HARDWARE, "Schedule Bool %d", schedule_selected);
```
kiddigital: @JurgenLB , I guess these 2 debug lines can be removed?
And/or add a Debug line within the If or combine all 3 values after the if?

hardware/Netatmo.cpp
```diff
@@ -2114,21 +2137,21 @@ bool CNetatmo::ParseDashboard(const Json::Value& root, const int DevIdx, const i
...
 		v << rain_1;
 		v << ";";
 		v << rain_24;
 		std::string sValue = v.str().c_str();
                 ///Debug(DEBUG_HARDWARE, "(%d) %s (%s) [%s] rain %s %s %d %d", Hardware_int, str_ID.c_str(), pchar_ID, name.c_str(), sValue.c_str(), m_Name.c_str(), rssiLevel, batValue);
 		//UpdateValueInt(0, str_ID.c_str(), 0, pTypeRAIN, sTypeRAINByRate, rssiLevel, batValue, '0', v.str().c_str(), Name, 0, m_Name);
-		SendRainSensor(ID, batValue, rain_24 + rain_1, name, rssiLevel);
-		//SendRainSensorWU(ID, batValue, rain_24, rain_1, name + "- WU", rssiLevel);
+		//SendRainSensor(ID, batValue, m_RainOffset[ID] + m_OldRainCounter[ID], name, rssiLevel);
+		SendRainSensorWU(ID, batValue, rain_24, rain_1, name, rssiLevel);
```
kiddigital, line 2036: @JurgenLB , why switch (back) to WU? What 'issue' does that solve?

hardware/Netatmo.cpp
```diff
@@ -2561,7 +2585,9 @@ bool CNetatmo::ParseHomeStatus(const std::string& sResult, Json::Value& root, st
 						b << ";";
 						b << boiler_boost;
 						std::string sValue = b.str().c_str();
+						const double Level = module["boiler_valve_comfort_boost"].asDouble();
 						//UpdateValueInt(0, ID.c_str(), 0, pTypeGeneralSwitch, sSwitchGeneralSwitch, '0', 255, '0', sValue.c_str(), bName,  bIsActive, m_Name);
+						//SendSwitch(crcId, 9, batteryLevel, bIsActive, Level, bName, m_Name, mrf_status);
```
kiddigital: @JurgenLB , add a line but commented out? Remove it?

hardware/Netatmo.cpp
```diff
@@ -2561,7 +2585,9 @@ bool CNetatmo::ParseHomeStatus(const std::string& sResult, Json::Value& root, st
 						b << ";";
 						b << boiler_boost;
 						std::string sValue = b.str().c_str();
+						const double Level = module["boiler_valve_comfort_boost"].asDouble();
```
kiddigital: @JurgenLB , add a variable and value that is not used? Or was, but now commented out.. remove as well?

hardware/Netatmo.cpp
JurgenLB: Because off a Legrand Gateway there was some undefined behavior,
adding " && schedule_type == "therm"" did resolve that.
So I can remove the Log lines or just keep 1.

hardware/Netatmo.cpp
JurgenLB: when I use "SendRainSensor(ID, batValue, rain_24 + rain_1, " the daily data seems correct but the historical data is not.
the original code was "SendRainSensor(ID, batValue, m_RainOffset[ID] + m_OldRainCounter[ID], " but this result in a very odd number for the current rain. 

image

because the WU sensor uses also rain_24 and rain_1 as parameter I think this will work better.

hardware/Netatmo.cpp
JurgenLB: this belongs to the //SendSwitch

hardware/Netatmo.cpp
kiddigital: Why add a line and comment it out? Adding code that is not used does not help.
Remove it? (And the variable created above as well)

hardware/Netatmo.cpp
kiddigital: See above. Adding a variable and not use it does not make things clearer.

JurgenLB:
do you want me to remove the unneeded code in the complete file?

JurgenLB:
For the AirQualitysensor using the index resulted in different (incorrect) sensors.
Error Air Quality

so I made the unit number in case off the 'Netatmo weatherstation' and 'Netatmo additional indoorsensor' fix to 0
This is consistent with the Old implementation.

kind regards

hardware/Netatmo.cpp
JurgenLB: The past 3 days the Domoticz panel and history are corresponding to the Netatmo website and history.
With using the.  `SendRainSensorWU(ID, batValue, rain_24, rain_1, name, rssiLevel)`

image

JurgenLB:
additional;
The DevID in the debug line off SQLH UpdateValueInt is not always the same as visible in Domoticz.
```
2024-12-26 11:41:47.104  Debug: SQLH UpdateValueInt Weather Station HwID:2  DevID:204 Type:249  sType:1 nValue:932 sValue: 
2024-12-26 11:41:47.105  subtype       = Voc
CO2 = 932 ppm
Quality = Fair
```
image
so 204 in log and 00CC in the devices tab.

in my local setup I have added the IDX to the log-line;
`_log.Debug(DEBUG_NORM, "SQLH UpdateValueInt %s HwID:%d  DevID:%s Type:%d  sType:%d nValue:%d sValue:%s IDX: %" PRIu64 , devname.c_str(), HardwareID, ID, devType, subType, nValue, sValue, ulID);`

I believe this is more helpful to find the corresponding device.

Kind regards

gizmocuz:
Let me know when it's ready to be merged!

JurgenLB:
I don't have a lot off rain on my location for the moment but the last 5 days the rainsensorWU is consisted with the Netatmo Website.

The other sensors are also working in my opinion.
so I think its ready to be merged.

Kind regards

gizmocuz:
Thanks once again!!

PR #6228, opened 2024-12-24
Update about.html

skywalker1979:
corrected link to forum


gizmocuz:
Thank you for this, could you add a second change... the wiki URL? (Just change, commit and push will update this PR)

skywalker1979:
the other pull request should be correct....

PR #6230, opened 2024-12-25
Code Refactor of the WebServerCommands file/function (10% less code)

kiddigital:
This PR refactors the `HandleCommandParam` function of the _WebServerCommands.cpp_ file.

The goal is to totally get rid of this function (and the file) once the (still 70!) command options are rewritten to proper `Cmd_<commandoption>` methods (in either _WebServerCmds.cpp_ or even separate files).

Things that have changed:
- Authorization checking is now done at the top instead of per each command options (saves a few hundred lines of code!)
- Instead of a very long `if else (cparam == "<someoption>")` structure, now a `switch/case` is used (by hashing the strings to integers)
- Setting the command title property at the top, so in case of failure it is at least known in which command options the failure occurred.
- The status is only set to 'OK' at the very end of each command routine, so we know for sure that it went correctly (instead of failing somewhere after that the OK was already set).

NOTE: Although fairly some testing has been done, not all 70! commands have been individually tested. As the actual code of these commands have not been altered, the changes are minimal that something broke (but hey, nobody is perfect ;) )


kiddigital:
@gizmocuz , changed the `_` Operator to `_sh` for readability

gizmocuz:
Thank you once again!!!

lokonli:
Can it be that this commit introduces some changes to the websocket interface?
When I open the websocket from Dashticz an error is returned:
image

Any idea where to start debugging?
Domoticz debug logging doesn't show any error or warning.

kiddigital:
@lokonli , run domoticz with debugflags `webserver,auth,received` for more detailed logging of what is going on.
But it is true that if some internal command sets a reply status 500 (Internal Server Error), this is now also returned, which wasn't the case before.

lokonli:
I had tried already with those debug flags: no feedback why the websocket request gets rejected.

Did the authentication mechanism for the websocket change?
The websocket from the Domoticz UI still is working, so something on Dashticz side is wrong.

kiddigital:
> I had tried already with those debug flags: no feedback why the websocket request gets rejected.
> 
> Did the authentication mechanism for the websocket change? The websocket from the Domoticz UI still is working, so something on Dashticz side is wrong.

It did not.. but what did _change_ is that the session cookie is now only sent after successful login (and not always with any request).. do you sent authentication data with each request? The debug log should tell you if the request was authorized or not.

lokonli:
For the websocket connection I did not use any authentication method. I can't set the correct cookie for a websocket coming from a different ip address.

From the logs:
```
2025-01-02 13:11:01.671  Debug: [web:8080] Host:192.168.1.22 Uri:/json
2025-01-02 13:11:01.671  Debug: [web:8080] IP (192.168.1.22) is within Trusted network range!
2025-01-02 13:11:01.671  Debug: [web:8080] isPage 0 isAction 0 isUpgrade 1 needsAuthentication 1 isAuthenticated 1 (admin)
2025-01-02 13:11:01.671  Debug: Web ACLF: 192.168.1.22 - - [02/Jan/2025:13:11:01.671 +0100] "GET /json HTTP/1.1" 101 0 - "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/131.0.0.0 Safari/537.36"
2025-01-02 13:11:01.679  Debug: CWebServer::GetJSonPage :getdevices :/json.htm?type=command&param=getdevices&filter=all&used=true&order=Name 
```

kiddigital:
@lokonli , from the piece of log everything looks fine :)

- Request comes in (`/json`')
- Get's authenticated as you are in the Trusted network
- It is an ' upgrade' request (and needs Auth and is Authenticated)
- And a proper 101 response is returned (so switched to ..)

Maybe some more logging with consecutive calls?
Try adding `received` debug flag to see all incoming HTTP headers as well

PR #6231, opened 2024-12-25
Remove mapquestapi address to lon/lat from Setup screen (API does not work anymore)

kiddigital:
Small PR which removes the Address to Lon/Lat translation in the setup screen from the 'Find your Geo Location' dialog.

It seems that the used MapQuestAPI no longer works.

So only the browsers `navigator.geolocation` functions remains.


gizmocuz:
Best to look for an alternative solution instead of removing it

gizmocuz:
I modified the code to use OpenStreetMap now, thanks for pointing out this was not working anymore!

PR #6232, opened 2024-12-25
Remove Thermosmart hardware support as the product is no longer maintained/supported

kiddigital:
The Thermosmart smart thermostat is not working anymore as the manufacturer has quit. The product does not work anymore (via the Cloud) and therefor is no longer usable from Domoticz.

This PR removes all the code involved.


gizmocuz:
Thank you!

PR #6236, opened 2024-12-31
Changed links to wiki, added Netatmo Wiki

JurgenLB:
I made the Netamo wiki to reflect for all the changes.
I have also changed the wiki URL to the wiki.domoticz instead off domoticz.com/wiki

If you don't want this wiki on GitHub, the Netatmo wiki can be copied.

Kind regards and Happy Newyear


gizmocuz:
Thank you very much for this!!! I will copy the Netatmo wiki to the Wiki

Replaced the following page with it's content

https://wiki.domoticz.com/Netatmo

PR #6238, opened 2025-01-02
Support for Multicast DNS so User can browse to domoticz-domoticz.local (no need to know IP-address)

kiddigital:
This PR adds Multicast DNS support (mDNS).

Domoticz now announces itself on the local netwerk that it can be found at _domoticz.local_ and listens to mDNS requests for queries that request details.

mDNS support in Domoticz is also needed for feature support for _Matter_ as it utilizes mDNS (over IPv6) a lot.

So by default, the mDNS name will be `domoticz-domoticz.local`.
The user can specify something different under Settings / System / Location / Name. Then it will be `domoticz-<thenamespecified>.local`.
The reason why we not use the shorter `domoticz.local` is because often machines broadcast their names on the network as well. And an obvious name for any domoticz machines is ... ;) And if you run more than 1 Domoticz installation in your (larger) network, these needs to be distinguishable from each other.

It is possible to disable mDNS by specifying `-nomdns`.

To see more details of what is going on, enable the `received` debug flag via the commandline (`-debuglevel received`).


gizmocuz:
could you change the hostname to

domoticz_ (domoticz and a underscore), then take the instance name that the user can configure in the settings, and past this behind this?
This way we always know it's a domoticz instance, and which one

gizmocuz:
Maybe you could remove the include in mainworker.h and the declaration, and just place this at the top of mainworker ?
There are also some other declarations there

As you noticed, quite a few classes include mainworker.h, and this leads to slow compilation
Also use the correct path, probably add ../ infront of it

kiddigital:
> could you change the hostname to
> 
> domoticz_ (domoticz and a underscore), then take the instance name that the user can configure in the settings, and past this behind this? This way we always know it's a domoticz instance, and which one

Done, except that the `_` (underscore) is not allowed in dns names :) So instead it uses a `-` (hyphen).

So by default, the mDNS name will be `domoticz-domoticz.local`. 
Unless the user specify something different under _Settings / System / Location / Name_. Then it will be `domoticz-<thenamespecified>.local`.

pipiche38:
How will it behave if you have 2 domoticz server on the same lan ?

kiddigital:
See an earlier comment 😁
You can set the name of the system and it will use that 👍

gizmocuz:
Is it OK to merge this PR?

gizmocuz:
Thank you for the changes.
I don't think the option 'USE_MDNS' is necessary, as without it, there will probably be compile errors
Maybe best to just always use it instead?

koenkooi:
> [...] And if you run more than 1 Domoticz installation in your (larger) network, these needs to be distinguishable from each other.

Since mDNS doesn't get routed, mDNS name clashes are limited to the same broadcast domain. Having multiple hosts with the same name on the same broadcast domain won't be fixed by doubling up the mDNS name with the internal domoticz name, this proposal would just hide the fallout, but only for mDNS.

kiddigital:
> > [...] And if you run more than 1 Domoticz installation in your (larger) network, these needs to be distinguishable from each other.
> 
> Since mDNS doesn't get routed, mDNS name clashes are limited to the same broadcast domain. Having multiple hosts with the same name on the same broadcast domain won't be fixed by doubling up the mDNS name with the internal domoticz name, this proposal would just hide the fallout, but only for mDNS.

Fair point. In case multiple hosts within the same broadcast domain have the same hostname, that could already lead to network issue but only if these hostnames are _visible_ on the network through certain protocols (like DHCP, mDNS/bonjour, NetBios, WINS, etc.). The solution in this PR at least can prevent mDNS issues, but not others. It might be better/common practice to avoid using the same hostnames at all in any network. And if so, there is no (more) need to prevent this. It also makes the broadcasted name shorter and easier.

I guess, maybe the PR should be modified to use the actual hostname of the device and let that be overwritten by the Domoticz setting. That would remove the need for the 'domoticz-' prefix and shorten the mDNS name. I will have a look at that.

gizmocuz:
I agree with the above. Having multiple hosts with the same name inside your network is already a point of failure.
In case of multiple Domoticz instances in your network (perfectly possible), you can name your instance in the Domoticz settings
Using this name makes it also unique for mDNS

kiddigital:
Closed in favor of PR #6327

PR #6244, opened 2025-01-08
Weather widget update

badzz:
Change the log link on the weather widget to use the angular component and allow zooming on daily chart

### DIFF
--- a/www/app/log/RainLog.js
+++ b/www/app/log/RainLog.js
@@ -254,7 +254,6 @@ define(['app', 'lodash', 'RefreshingChart', 'DataLoader', 'ChartLoader', 'log/Ch
 					highchartTemplate: {
 						chart: {
 							type: 'column',
-							zoomType: false,
 							marginRight: 10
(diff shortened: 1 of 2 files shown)


gizmocuz:
Thank you for this PR.
However, this does not work. When you have for example a Temp+Hum+Baro sensor and display the log, it shows the Temp+Hum log, not the barometer log.
That is why the internal function is called to display the correct log.

badzz:
Ok thanks. I was suspecting that there was a good reason for this but could
not find one.
Is there a way, as is, to have the nicer component for rain log ?

gizmocuz:
What do you suggest for the rain log? Does it not show the amount of rain during the day?

badzz:
Hi
I can see the rain info ok but I don't see the zoomdays. 3d, 1d, 3h ,1h,..
Thanks

gizmocuz:
Maybe you could implement the zooming on the individual log device type views/code?

PR #6251, opened 2025-01-20
Boost 1.87 fixes + a bit more

danielwallner:
All fixes required to build with latest Boost.
This also contains a few bug and warning fix commits that was done along the way.

A few related lines that seemed unused were also removed.
I did keep a blocking resolve followed by an asynchronous resolve that looked a bit odd at line 855 in PluginTransports.cpp.
m_Endpoint could be assigned in the lambda if it isn't used before that callback.

Note that this is largely untested.
Domoticz works for me with these commits but that is all testing I have done.
I wasn't able to run the automated tests.

### DIFF
--- a/hardware/ASyncSerial.cpp
+++ b/hardware/ASyncSerial.cpp
@@ -54,7 +54,7 @@ public:
   {
   }
 
-    boost::asio::io_service io; ///< Io service object
+    boost::asio::io_context io; ///< Io service object
     boost::asio::serial_port port; ///< Serial port object
     boost::thread backgroundThread; ///< Thread that runs read/write operations
     bool open{ false };		    ///< True if port open
@@ -117,10 +117,10 @@ void AsyncSerial::open(const std::string& devname, unsigned int baud_rate,
 		throw;
 	}
 
-	pimpl->io.reset();
+	pimpl->io.restart();
 
-	// This gives some work to the io_service before it is started
-	pimpl->io.post([this] { return doRead(); });
+	// This gives some work to the io_context before it is started
+	boost::asio::post(pimpl->io, [this] { return doRead(); });
 
 	boost::thread t([p = &pimpl->io] { p->run(); });
 	pimpl->backgroundThread.swap(t);
@@ -149,10 +149,10 @@ void AsyncSerial::openOnlyBaud(const std::string& devname, unsigned int baud_rat
 		throw;
 	}
 
-	pimpl->io.reset();
+	pimpl->io.restart();
 
-	//This gives some work to the io_service before it is started
-	pimpl->io.post([this] { return doRead(); });
+	//This gives some work to the io_context before it is started
+	boost::asio::post(pimpl->io, [this] { return doRead(); });
 
 	boost::thread t([p = &pimpl->io] { p->run(); });
 	pimpl->backgroundThread.swap(t);
@@ -176,9 +176,9 @@ void AsyncSerial::close()
     if(!isOpen()) return;
 
     pimpl->open = false;
-    pimpl->io.post([this] { doClose(); });
+    boost::asio::post(pimpl->io, [this] { doClose(); });
     pimpl->backgroundThread.join();
-    pimpl->io.reset();
+    pimpl->io.restart();
     if(errorStatus())
     {
         throw(boost::system::system_error(boost::system::error_code(),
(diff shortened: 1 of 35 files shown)


gizmocuz:
Thank you very much for this great piece of work!!!

Regarding the blocking resolve, where you able to find a solution for this, or is it something we still need to look at?
This issue as probably also before the migration to Boost 1.87?

All looks good, I will do some testing with this

gizmocuz:
I have one issue compiling under Windows, and that is in 

ASyncTCP::connect_start

It is the second argument passed to '**cb_connect_done**'

message:

`'void ASyncTCP::cb_connect_done(const boost::system::error_code &,const boost::asio::ip::tcp::endpoint &)': cannot convert argument 2 from 'Iterator' to 'const boost::asio::ip::tcp::endpoint &'
`

To solve it, I have removed the second parameter as it is not used... but I failed at solving this by changing the argument type

gizmocuz:
I now switched to Boost 1.87 and recompiled.

I do have some more errors now:

Actually it seems only in **tcpproxy_server.cpp**

'query': is not a member of 'boost::asio::ip::basic_resolver<boost::asio::ip::tcp,boost::asio::any_io_executor>'

'from_string': is not a member of 'boost::asio::ip::address'

danielwallner:
The first windows build problem was caused by the second parameter being a reference I think.
It shouldn't be and isn't in the other places I modified.
Removing it does not build on MacOS.

tcpproxy_server.cpp wasn't included in the build here so I didn't notice that problem.
This build for me:
```
1:
		boost::asio::ip::tcp::resolver resolver(ios);
		auto endpoints = resolver.resolve(upstream_host, upstream_port);
		if (endpoints.size() == 0)
		{
			end=boost::asio::ip::tcp::endpoint(
				boost::asio::ip::make_address_v4(upstream_host),
				(unsigned short)atoi(upstream_port.c_str()));
		}
		else
			end=*endpoints.cbegin();

2:
		, localhost_address(boost::asio::ip::make_address_v4(local_host))

3:
		boost::asio::post(io_context_, [this] { handle_stop(); });
```

Again untested.

danielwallner:
The double resolve probably isn't a real issue but looks a bit weird and was there before.
Someone should have a look eventually.

gizmocuz:
Thanks for point out that your build did not include tcp_proxy, this was an old left-over in the visual studio project.
(Which can be removed)

I am now able to compile and test.

I am trying to connect to a P1 smart meter on TCP port 23.
It stops at 'attempt connect to 172.16.x.x:23)

When I manually connect this from the command line it works

`ASyncTCP::cb_resolve_done` is not called (ASyncTCP::connect, line 78)

It seems the below call is working

`auto endpoints = boost::asio::ip::tcp::resolver(mIoc).resolve(mIp, port_str);`

Where you able to connect any TCP hardware (or Serial)

gizmocuz:
What I noticed is that when i place the async_resolve code 'above' the mIoc.restart() all works!

image

gizmocuz:
Serial port seems to work perfect

danielwallner:
Serial port works for me too but that is also the only part of this that I use.

That mIoc.restart() and the following recreation of the work_guard is the part I was least confident about in this.
That you have to put it in an optional instead of a shared_ptr now to recreate it makes me think that this is not the way to do it.

It's just very hard to find examples and documentation of the "new" way to do things.
Seems most projects have been using the deprecated stuff until it now.

gizmocuz:
If you have time, you could use 'ser2net' (raw mode) to make your serial port available via TCP
Depending on the hardware, you could then try out the TCP code. (For example P1 meter, RFXCom)

The only part I changed was putting the async_resolve call about mIoc.restart() ... (I hope, because I have been fiddling to try to get it working)

I totally agree about the boost examples.... it seems someone has modified the asio code, but did not bother migrating the documentation examples. There are lots of examples missing that were part of the old boost libraries (1.45+)

It seems there is more to find here:

https://github.com/boostorg/asio/tree/develop/example

still looking for async client examples

Edit: Found a HTTP client example... I will tweak the code and probably make a new branch with all your changes as I am probably not allowed to push into this PR (or don't know how 🤣)
Will report back shortly

gizmocuz:
@danielwallner , I made some modifications, if it is possible, could you have a look at

https://github.com/domoticz/domoticz/pull/6252

PR #6252, opened 2025-01-24
Boost 1.87 fixes + a bit more

### LINKED ISSUE #6246: Domoticz does not build with Boost 1.87
Boost 1.87 removes deprecated functionality in boost::asio that currently is used by Domoticz.

gizmocuz:


gizmocuz:
@kiddigital , Is it possible to also have a look at this PR. I have tested TCP/Serial and the changes in the webserver also looks OK

@dnpwwo , I know your busy, but if you have time:

One remark from Daniel:

```
I did keep a blocking resolve followed by an asynchronous resolve that looked a bit odd at line 855 in PluginTransports.cpp.
m_Endpoint could be assigned in the lambda if it isn't used before that callback.
```

What I find double here... is that we call 'resolve' twice, sync and later async_resolve

I have no idea how to test this, but it seems to be that m_Endpoint should be set in

`handleAsyncResolve`

I tested the code with the Python 'Pinger' example. (Fixed the plugin as I had a crash here in this branch as well (probably caused by a timeout)

**m_Resolver.async_resolve** is called in handleConnect and handleListen.
The Pinger plugin is always closing the connection, so it is always called.
This means 'm_bConnected' is also always false.

I did make a change where 'm_Endpoint' is now set in **CPluginTransportICMP::handleAsyncResolve**

But when used with the Pinger example, there is actually no reason to keep 'm_Endpoint' and even 'm_bConnected'

danielwallner:
I have had a look at the changes here and can also at least confirm that it still works with my rather small setup.
I'm still not able to be really helpful in running any real testing beyond that unfortunately.

kiddigital (approved):
@gizmocuz , as far as I can see and (automatically) test, building and testing against boost 1.87 and older versions, I did not see or find any issue.
Difficulty (also for @danielwallner ) is that it is not possible to test the hardware changes.
I would suggest to merge this __after__ the next stable, so there is a longer time-window for user with relevant hardware to test the Beta's after that.

NOTE: we need to update the build pipelines as well, not part of this PR.

gizmocuz:
Relevant hardware is serial or TCP, which is used by quite a lot of users....
Going to merge this in development as my build system is not usable anymore
Maybe the next beta version should also address the latest openssl soon, if that is working make a stable?

gizmocuz:
The current build systems should have no issues with the boost changes as they use a recent boost version

PR #6254, opened 2025-01-26
Added support for Simon 100 thermostat.

JordiAndujar:
Added support for Simon 100 thermostat.

Please, I'm new into this, check at least:

- My Simon 100 thermostat has the identifier 1000081**0**-13X, while the only similar device I've found in Z-Wave alliance has the identifier 1000081**X**-13X (changes the middle X) in here: https://products.z-wavealliance.org/products/3904 . I've put into the XML this link, do you think it would be a problem?
- The product ID is set to "0000" and my Domoticz has recognized the device properly. If I set the ID shown in the ZWave alliance page (https://products.z-wavealliance.org/products/3904), the device is not recognized. Is it correct to leave it as "0000"?

The configurations have been tested with my Simon 100 thermostat.

Thank you in advance.


gizmocuz:
Thank you for this PR! Looking good!

But please consider moving to ZWaveJS-UI (MQTT-AD)

PR #6267, opened 2025-02-23
Fix  #6264 issue : Motherboard Sensors - inconsistent order of disks

### LINKED ISSUE #6264: Motherboard Sensors - inconsistent order of disks
Key facts about Linux:

- The order of `/dev/sd*` devices is not specified. It depends on the order of hardware detection, the order of loading kernel modules. It is usually unchanging when the hardware does not change, but according to the documentation the order is unspecified.
- Modern versions of udev (device manager) cannot change the kernel assigned name (it worked a long time ago). There is practically no way to change the order of `/dev/sd*`.
- Also, the order of entries in the list printed by the `df` tool is not specified.

The problem is that Domoticz relies on the order printed by the `df` tool. If the order of disks in the system does not change, then it somehow works.

**But when the order of disks in the system changes, the approach used in Domoticz logic is completely wrong.**

Examples of problems:

Three loop devices arrived. Domoticz stopped reporting disk usage correctly. The essence of the problem was not solved, the issue was closed.
https://github.com/domoticz/domoticz/issues/2822

Devices swapped order. Domoticz stopped reporting disk usage correctly. The essence of the problem was not solved, the issue was closed.
https://github.com/domoticz/domoticz/issues/3375

Now my problem. After updating the system on the raspberrypi to the latest Raspberry Pi OS (based on Debian 12, bookworm) the disk order has completely changed. Domoticz stopped reporting disk usage correctly.

Moreover, there are reports on the internet that in Debian 12 in some cases the disk order went crazy. It can even change on every boot.

**In this case Motherboard Sensors is useless because it relies on the order of devices when first booted.**

The solutions found to the problem are ineffective. One of them involves removing and adding sensors again. This is nonsense, the entire sensor history will be removed. Another suggests manually modifying the names in domoticz so that they match the physical disks. But then the sensor history will not match.

In domoticz you cannot change the sensor idx. Too bad. I could constantly export the database, change the idx there and import it again. Don't you think it's a crazy idea? All this to fix the wrong logic in Domoticz code.

**It's high time to finally fix this logic and ensure that the motherboard sensors remembers mountpoints names. Then if the order of disks in the system changes, the data will still be assigned to the correct sensor.**

mealgi:
Fix  #6264 issue : Motherboard Sensors - inconsistent order of disks

### DIFF
--- a/hardware/HardwareMonitor.cpp
+++ b/hardware/HardwareMonitor.cpp
@@ -937,21 +937,34 @@ void CHardwareMonitor::FetchUnixDisk()
 				std::string hddname = "HDD " + dusage.MountPoint;
 				dindex=0;
 				std::vector<std::vector<std::string> > listOfHdd;
-				listOfHdd = m_sql.safe_query("SELECT ID, DeviceID, Name FROM DeviceStatus WHERE (HardwareID=%d AND DeviceID>'0000044D' AND DeviceID<'000004B0')", m_HwdID);
+				listOfHdd = m_sql.safe_query("SELECT ID, DeviceID, Name, Options FROM DeviceStatus WHERE (HardwareID=%d AND DeviceID>'0000044D' AND DeviceID<'000004B0')", m_HwdID);
 				if (!listOfHdd.empty())
 				{
 					for (const auto& sd : listOfHdd)
 					{
-						std::string idx = sd[0];
+						std::string szIdx = sd[0];
 						std::string szDeviceId = sd[1];
 						std::string Name = sd[2];
 						int deviceId;
 						sscanf(szDeviceId.c_str(), "%x", &deviceId);
-						if (!strcmp(Name.c_str(),hddname.c_str()))
+						std::string sOptions = sd[3];
+						std::map<std::string, std::string> optionsMap = m_sql.BuildDeviceOptions(sOptions);
+						if (!optionsMap.empty() and !strcmp(optionsMap["hdd"].c_str(),hddname.c_str()))
 						{	
 							dindex=deviceId;
 							break;
 						}	
+						else
+						{
+							if (!strcmp(Name.c_str(),hddname.c_str()))
+							{
+								dindex=deviceId;
+								uint64_t idx = std::stoull(szIdx);
+								optionsMap["hdd"] = hddname;
+								m_sql.SetDeviceOptions(idx, optionsMap);
+								break;
+							}	
+						}
 					}
 				}
 				if (dindex==0)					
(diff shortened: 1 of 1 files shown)


gizmocuz:
Does this fix your system? I see quite a few hard coded device ids en indexes, is this working for everyone?
Will everyone get new devices?

mealgi:
@gizmocuz 
Yes it fixes my system.
Regarding the hardcoded indexes I wanted only to proceed to few changes.
The base index for the "load" devices is already hardcoded and the value is 1100 (see line 297 of HardwareMonitor.cpp file).
Memory Usage is attached to index 1100
CPU usage is attached to index 1101
The HDD usages are allocated from index 1102 up to 1199.

mealgi:
To complement the comment:

is this working for everyone? Yes as long as the device name has not been modified and is the one provided by the "df" command.
If the device name has been modified domoticz will create a new device with the name of the HDD provided by the "df" command. 

Will everyone get new devices? Yes as long as the setting "Accept new device" is checked in domoticz

jan-adam:
You use the "Name" field to compare disk mountpoints, but the "Name" field can be changed by the user in the Web UI. Of course, this is better than the previous solutions where domoticz would give incorrect data after changing the order of disks.

For example in the manual for writing python plugins for domoticz (https://wiki.domoticz.com/Developing_a_Python_plugin) you can read "Name - This should not be used in Python because it can be changed in the Web UI."

And indeed. This can be changed by the Web UI. Can this field be used like that? I wonder if there is a better way. Does the domoticz architecture allow to save the mountpoint name permanently in some private field, so that it is independent when the user changes names in the Web UI? That would be the best solution.

mealgi:
@jan-adam 
In the fix the name is written in c++ source code, not in Python.
An alternate solution would be to use the field Description but it is also possible to change it in the Web UI.

@gizmocuz 
What do you suggest : Name or Description (or other field) to keep track of the hdd name returned from the "df" command ?

jan-adam:
It's not python, but the database is the same. The field value can be easily changed in the WEB UI and that has to be stated.

So that there are no bug reports to domoticz from uninformed users that "they change the sensor name and domoticz automatically creates a new one with the old name".

Let's look for another place where we can save the mountpoint name. In my opinion the description would be better, but still not ideal.

mealgi:
@jan-adam 
I have developed a new solution using the Option field which is not accessible from the Web UI
The PR will be updated accordingly

mealgi:
@gizmocuz 
The PR has been updated accordingly using the Option field

mealgi:
@gizmocuz
Did you have time to review the PR prior to merging it ?
BR

gizmocuz:
Thank you for this PR!

PR #6272, opened 2025-03-04
Added some new icons

### LINKED ISSUE #6257: Custom icons
Hi. 
1. is it possible to add some useful icons to the repository, so they are immediately available to all domoticz users without the need to search for the needed icons and install them by  hand?
2. if I can, is it sufficient to submit a pull request with the 3 icons file in www/images   and add info on www/switch_icons.txt ?
Thanks.

CreasolTech:
I've added some icons, and modified the switch_icons.txt file adding info about icons.


gizmocuz:
Thank you for this PR.
I added some of these icons. 'Pump' would not have worked in this PR as the icon starts with 'Pump1'
A Fan did already exist
And some icons are very specific and do not match the coloring scheme of the default theme

I added them amongst some of the icons I made/use in beta 16490

PR #6273, opened 2025-03-04
Fix Floorplan protected selector issue 6268

### LINKED ISSUE #6268: Problem on floorplan when activating selector device protected by code
Hello,
On floorplan, when I try to use a selector switch protected by code, the message "wrong code" is returned.
On switches tab, the same operation works fine (code is asked and the right action is triggered when I select a value with the protected switch).
I use the latest version of domoticz on raspberry pi (raspbian 11).
Thanks for help, I will try to give more details if needed.
Stan

waltervl:
Fix Floorplan protected selector issue https://github.com/domoticz/domoticz/issues/6268

### DIFF
--- a/www/js/domoticzdevices.js
+++ b/www/js/domoticzdevices.js
@@ -1965,7 +1965,7 @@ function Selector(item) {
         };
         this.onSelectorValueChange = function (idx, levelname, level) {
             // Send device command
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thank you for this PR and fix! Will be available in beta 16488 and up

PR #6274, opened 2025-03-09
Netatmo Change selector and Add Blinds

JurgenLB:
BUGFIX;
In 1 Home the Schedules are the same for every Thermostat/valve.
The Schedules can be different for multiple Homes .
With this PR I have mapped the SelectorLevels With the Home_ID.

ADDED;
Now with working open/close Blinds from Legrand.

Kind Regards


gizmocuz:
Thank you again for this PR!!

PR #6276, opened 2025-03-15
Return message when data is invalid.

JurgenLB:
With the most recent (again) changes from Netatmo.
There are a lot of users with dealing with the 'invalid login', this PR shown the return message in that case so it is more clear to where the problem is situating.
When running in a normal installation the Error log shown the return HTTP header code.
And when running in Debug the compleet return message is displayed in the Debug LOG.

The authentication is now situating in Hardware.js and `_log.Debug(DEBUG_HARDWARE, "Debug : $(_url)");` is not working there.
it is send to a different console.log

This is work from Loukkamp, I reached out to him with a negative respons.

kind regards


hardware/Netatmo.cpp
```diff
@@ -331,7 +331,7 @@ bool CNetatmo::RefreshToken(const bool bForce)
 			return true; //no need to refresh the token yet
 	}
 
-	Log (LOG_STATUS, "Requesting refreshed tokens");
+	Log (LOG_STATUS, "%s Requesting refreshed tokens", m_Name.c_str());
```
gizmocuz: This is not necessary, it will already add the hardware name Infront of each log message

hardware/Netatmo.cpp
```diff
@@ -355,7 +355,7 @@ bool CNetatmo::RefreshToken(const bool bForce)
 	//Check for returned data
 	if (!ret)
 	{
-		Log(LOG_ERROR, "Error connecting to Server (refresh tokens): %s", ExtractHtmlStatusCode(returnHeaders).c_str());
+		Log(LOG_ERROR, "%s Error connecting to Server (refresh tokens): %s", m_Name.c_str(), ExtractHtmlStatusCode(returnHeaders).c_str());
```
gizmocuz: The hardware name is already logged by default

hardware/Netatmo.cpp
```diff
@@ -364,7 +364,8 @@ bool CNetatmo::RefreshToken(const bool bForce)
 	ret = ParseJSon(sResult, root);
 	if ((!ret) || (!root.isObject()))
 	{
-		Log(LOG_ERROR, "Invalid/no data received (refresh tokens)...");
+		Debug(DEBUG_HARDWARE, "Netatmo Invalid ... %s", sResult.c_str());
+		Log(LOG_ERROR, "%s Invalid/no data received (refresh tokens)... %s", m_Name.c_str(), ExtractHtmlStatusCode(returnHeaders).c_str());
```
gizmocuz: The hardware name is already logged by default

JurgenLB:
Sorry I missed the other 2. Now resolved.

Kind regards

gizmocuz:
Thank you once again!!! 👍👍

PR #6281, opened 2025-03-23
Scenarios from Gateway

JurgenLB:
With this PR the scenarios are running is a separated thread and working for the Legrand Gateway.

I have also added a timer to block the multiple request to the server witch leeds to a lock-out from the Netatmo server.

This timer does not solve the block, only a manual stop wait 20 minutes and start keeps the plugin Live.


gizmocuz:
Thank you for this PR! Changes are merged into beta 16584

gizmocuz:
@JurgenLB , I think there is a possible issue in the **find_scopes** function, after each 'if' there is a ; (semicolon), I don't think it should be there.

For example this:

```
	if (m_scopes.find("thermostat_RW") != std::string::npos);            //
		return true;

```

should be?

```
	if (m_scopes.find("thermostat_RW") != std::string::npos)
		return true;

```

JurgenLB:
Sorry, I will adapt this when I'm home.

jcbraem:
Hello,

Any news about this correction ?
Is it possible that this is a cause of connection issue? 
(https://forum.domoticz.com/viewtopic.php?t=38968&sid=b7994c74664644f8a68cc93c4caf2431&start=340à

Thanks,

Jean-Charles

gizmocuz:
A new beta has been created (16594) not 100% if the changes are already in this beta

PR #6282, opened 2025-03-23
Fixed getting device id (website changed). More clear debuginfo.

kiddigital:
Small PR that fixes the AtagOne hardware module that does not work anymore due to changes in the website it scrapes (after login).

Now it works again. Interval has been lowered from 60 seconds to 120 seconds due to login block occurring when to many logins (attempted) in a short period. 

Added some Debugging info (Hardware level) so it is easier to see if/what is going on.


gizmocuz:
Thank you for this PR!

PR #6288, opened 2025-03-29
Fix/linuxcompilewarnings

kiddigital:
Fix compiler warnings on Linux.


gizmocuz:
Thanks for the PR. I corrected them myself. The first one needed a string, and the second one is a int64

PR #6290, opened 2025-03-31
Use SendSwitchUnchecked to update unconditionnally Enocean Smoke Detector

ludos78:
This replace previous proposal in #6209, following discussions


gizmocuz:
Thank you for this PR! Quick question, did you check all works well with your patch and my previous change?

ludos78:
It seems OK on my side for this change, but i have a strange issue with the hardware and devices pages : the tab list is no more visible since i rebased my builds on development branch.

gizmocuz:
Try deleting the browser cache + appcache
Also open the browsers development console (F12) and dock in at the bottom
Next open the console to see if there are any errors when you refresh the page

ludos78:
it's ok today, probably some cache issue, thanks !

PR #6292, opened 2025-04-03
Fix AVATTO dimmers not switching on

arjendekorte:
Some Tuya dimmers report different manufacturer names, like 'AVATTO' in this case. Tested with both single and double version.


arjendekorte:
AVATTO dimmers (and other Tuya dimmers might as well) apparently ignore changes in 'brightness' value in case the 'state' is OFF. So when changing from OFF->ON, the brightness will be set to the value it had at the time it was switched off. This can be fixed by changing the order of the commands. So instead of sending the command
`zigbee2mqtt/dimmer/set {"brightness":76,"state":"ON"}`
it would be better to send
`zigbee2mqtt/dimmer/set {"state":"ON","brightness":76}`
instead.

gizmocuz:
Ehhh, this is JSON, it should not matter what order they are. It could even be so that the jsoncpp library sorts all fields alphabetical.

What you did in the patch is treat 'AVATTO' the same as TuYa, meaning, it now includes the 'state' field.

Thank you for the patch!

PR #6293, opened 2025-04-05
User agent added for Netatmo API call

Chrominator:
Added user agent to netatmo API call to ensure better compatibility.
user agent is made of "User-Agent: Domoticz/Version " and VERSION_STRING from main/appversion.h

### DIFF
--- a/hardware/Netatmo.cpp
+++ b/hardware/Netatmo.cpp
@@ -1,5 +1,6 @@
 #include "stdafx.h"
 #include "Netatmo.h"
+#include "../main/appversion.h"
 #include "../main/Logger.h"
 #include "../main/SQLHelper.h"
 #include "../main/mainworker.h"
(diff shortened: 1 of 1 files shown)


JurgenLB:
Did it resolve the connection block ?
I'm working on a code review.

Chrominator:
> Did it resolve the connection block ? I'm working on a code review.

It's working like a charm this morning, issuing a refreshed token after multiple starts, restarts, reboots and new connections issued from hardware page (tested only with thermostat scope).

JurgenLB:
yes and I stated also that using a separated cUrl in Netatmo.cpp is also a temporarily solution.
Because off the Fan-base off Domoticz when multiple users are going to use the same user-agent, then where blocked again.

So I believe we must make a solution to have the user-agent different for each installation off domoticz, even inside Docker.
kind regards

Chrominator:
One version of an Internet browser has only one user agent. 
And users are not blocked on a server because of that.
Users use different IP address, which should make the trick for the server.

gizmocuz:
Thanks for the PR
Could you undo all changes in the extern folder and only make the changes in the Netatmo class?

Chrominator:
Don't know how to cancel the changes on extern directory files.
Will redo my changes on Netatmo.cpp from a new branch

PR #6295, opened 2025-04-05
Enocean changes

ludos78:
Current A5-12-00, Automated Meter Reading, Counter implem is a bit wrong : it does not take into account CH and DT fields, mixing unrelated measure (at least for my device that provides several measurement channels and data types).
Use CH field to distinguish devices, and also DT that indicate if we've got a counter (using SendMeterSensor) or instant value measurement (using SendWattMeter).

note : 
- A5-12-00 profile is not explicitly for electric measurements (see https://tools.enocean-alliance.org/EEPViewer/profiles/A5/12/00/A5-12-00.pdf), but i don't know how to deal with this.
- modification in SendWattMeter is probably discussable (don't know the impact on other part of code), but i find it strange to restrict the nodeID to uint8_t where other functions use int ?

### DIFF
--- a/hardware/DomoticzHardware.cpp
+++ b/hardware/DomoticzHardware.cpp
@@ -502,7 +502,7 @@ bool CDomoticzHardwareBase::GetWindSensorValue(const int NodeID, int& WindDir, f
 	return bExists;
 }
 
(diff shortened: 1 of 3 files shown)


gizmocuz:
Thank you for this PR. I also see no issue in changing it to an 'int'.
I made a small patch in another class to make sure it keeps a uint8_t (ziBlueBase.cpp)

gizmocuz:
Changes are available in beta 16625 and up

PR #6296, opened 2025-04-05
User agent added for Netatmo API call

Chrominator:
Added user agent to netatmo API call to ensure better compatibility.
user agent is made of "User-Agent: Domoticz/Version " and VERSION_STRING from main/appversion.h

ex: User-Agent: Domoticz/Version 2025.1 (build 16611)


gizmocuz:
Thank you for this PR, merged with the other PR and available in beta 16619 and up

PR #6297, opened 2025-04-07
Refresh Login before token expires

JurgenLB:
this is to resolve the issue from {"code":3,"message":"Access token expired"}}

### DIFF
--- a/hardware/Netatmo.cpp
+++ b/hardware/Netatmo.cpp
@@ -221,7 +221,9 @@ void CNetatmo::Do_Work()
 		if (sec_counter % 12 == 0) {
 			m_LastHeartbeat = mytime(nullptr);
 		}
-
+		if ((mytime(nullptr) - 15) < m_nextRefreshTs)
+			m_isLogged = false;
+		
 		if (sec_counter % NETAMO_ERROR_INTERVALL == 0)
 		{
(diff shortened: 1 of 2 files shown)


gizmocuz:
Thanks for the PR.

One question, when the below check is true, it should just get a new token based on the refresh token instead of logging in again?

if (mytime(nullptr) > m_nextRefreshTs)

JurgenLB:
But the check for the refreshtoken is on line 314 in function login.
Maybe the function name is misleading.

Kind regards

gizmocuz:
Maybe you could change this? It is indeed not clear what is happening

- Either the refresh token expires, and we just need to get a new refresh token
For this, no variable has to be set, just call the RefreshToken function
In this function, if it fails, with a 'real' error from the server, not a connection issue?, then we indeed need to login again

- Or we need to login (because we never logged in before, or there was an error)

JurgenLB:
with this change my system is refreshing tokens `2025-04-11 17:35:19.074 Status: Netatmo: Refresh token 1`

this is ready to be merged if you want.

JurgenLB:
I did a recheck off all levels and now alle the write to hardware levels are working on my setup.

kind regards

gizmocuz:
Thank you for this PR!

PR #6301, opened 2025-04-08
bugfix for typename in PythonObjects.cpp

fleinze:
in PR #6183 in PythonObjects.cpp the code blocks where moved around in the CDevice_update function.
Writing default values did not matter prior to PR #6183 because the nValue and sValue were written to the database later in the function.
Calling CDevice_update (from python) with a TypeName but no nValue and sValue results in an error message. Therefore it seems unnecessary to write default values to the database.


gizmocuz:
Thank you for this PR. 
As I'm not into python plugins that much, maybe @badzz could also have a look at this?

fleinze:
I thought about this change overnight and I want to make it more save. So for now I will close this and reopen later with slightly improved code.

PR #6303, opened 2025-04-09
bugfix for typename change in PythonObjects.cpp

fleinze:
in PR https://github.com/domoticz/domoticz/pull/6183 in PythonObjects.cpp the code blocks where moved around in the CDevice_update function.
Writing default values did not matter prior to PR https://github.com/domoticz/domoticz/pull/6183 because the nValue and sValue were written to the database later in the function.

Second try, this way it should be more safe to not break any existing plugins
If the Update function is called with `SupressTriggers = false`, the nValue and sValue from function call is written to the DB and the default values are not written. If called with `SupressTriggers = true`, the default nValue and sValue is written to the DB

### DIFF
--- a/www/app/app.js
+++ b/www/app/app.js
@@ -423,8 +423,19 @@ define(['angularAMD', 'app.routes', 'app.constants', 'app.notifications', 'app.p
 							accessibility: {
 								enabled: false
 							},
+/*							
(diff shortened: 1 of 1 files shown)


fleinze:
my bad....

PR #6304, opened 2025-04-10
bugfix for default values when updating typename

fleinze:
in PR https://github.com/domoticz/domoticz/pull/6183 in PythonObjects.cpp the code blocks where moved around in the CDevice_update function.
Writing default values did not matter prior to PR https://github.com/domoticz/domoticz/pull/6183 because the nValue and sValue were written to the database later in the function.

Second try, this way it should be more safe to not break any existing plugins
If the Update function is called with SupressTriggers = false, the nValue and sValue from function call is written to the DB and the default values are not written. If called with SupressTriggers = true, the default nValue and sValue is written to the DB


gizmocuz:
Thank you for this PR and your time! Could you undo all changes to the extern folder in this PR?

fleinze:
Sorry about that. Still learning git. Should be clean now.

gizmocuz:
Thank you for this PR!

PR #6317, opened 2025-04-27
Pressure: allow 2 decimals for pressure sensors

koenkooi:
Plugwise Adam reports boiler pressure in 2 decimals, e.g. 1.53 bar, but domoticz internally lops off the second decimal. To make it worse, SQLHelper them drops *all* decimals.

In the context of Plugwise, the difference between '2.49' and '2.00' is significant, presenting and storing both as '2' is very misleading. I noticed this when wanting to use Domoticz to see if the repair to the leak in the central heating was successful.

Increase precision by one decimal everywhere: "%.2f", except for SQLHelper, that changes to "%.1f".

The net result is that data push to e.g. InfluxDB will have actually useful values.

### DIFF
--- a/main/SQLHelper.cpp
+++ b/main/SQLHelper.cpp
@@ -7012,8 +7012,8 @@ void CSQLHelper::UpdateMeter()
 			}
 			else if ((dType == pTypeGeneral) && (dSubType == sTypePressure))
 			{
-				double fValue = atof(sValue.c_str()) * 10.0F;
-				sprintf(szTmp, "%.1f", fValue);
+				double fValue = atof(sValue.c_str()) * 1000.0F;
+				sprintf(szTmp, "%.0f", fValue);
 				sValue = szTmp;
 			}
 			else if (dType == pTypeUsage)
--- a/main/SQLHelper.cpp
+++ b/main/SQLHelper.cpp
@@ -7013,7 +7013,7 @@ void CSQLHelper::UpdateMeter()
 			else if ((dType == pTypeGeneral) && (dSubType == sTypePressure))
 			{
 				double fValue = atof(sValue.c_str()) * 10.0F;
-				sprintf(szTmp, "%.0f", fValue);
+				sprintf(szTmp, "%.1f", fValue);
 				sValue = szTmp;
 			}
 			else if (dType == pTypeUsage)
--- a/main/WebServer.cpp
+++ b/main/WebServer.cpp
@@ -3342,7 +3342,7 @@ namespace http
 						}
 						else if (dSubType == sTypePressure)
 						{
-							sprintf(szData, "%.1f Bar", atof(sValue.c_str()));
+							sprintf(szData, "%.2f Bar", atof(sValue.c_str()));
 							root["result"][ii]["Data"] = szData;
 							root["result"][ii]["TypeImg"] = "gauge";
 							root["result"][ii]["HaveTimeout"] = bHaveTimeout;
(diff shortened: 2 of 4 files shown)


kiddigital:
@koenkooi , SQLhelper does not drop _all_ decimals but limits the amount of digits.
It does so by multiplying the value by 10 or a multitude of 10 (100, 1000, etc). This way, only real integer values have to be stored within the domoticz database and no floating values.
For 'pressure' subtype the value is multiplied by 10 (for 1 digit). So instead of using `%.1f`, please multiply by 100 in SQLhelper.
And make sure that when the values is used, it is not divided by 10 but by 100 also.

koenkooi:
> @koenkooi , SQLhelper does not drop _all_ decimals but limits the amount of digits. It does so by multiplying the value by 10 or a multitude of 10 (100, 1000, etc). This way, only real integer values have to be stored within the domoticz database and no floating values. For 'pressure' subtype the value is multiplied by 10 (for 1 digit). So instead of using `%.1f`, please multiply by 100 in SQLhelper. And make sure that when the values is used, it is not divided by 10 but by 100 also.

That explains the `%.1f` vs `%.2f` ! The current code in `WebServerHandleGraph.cpp` handles 10 and 1000 already, should I add clauses for 100 or just make pressure use 1000?

It's for sections like this:
```C++
if (((dType == pTypeGeneral) && (dSubType == sTypeVoltage)) || ((dType == pTypeGeneral) && (dSubType == sTypeCurrent)))
{
        sprintf(szTmp, "%.3f", fValue1);
        root["result"][ii]["v_min"] = szTmp;
        sprintf(szTmp, "%.3f", fValue2);
        root["result"][ii]["v_max"] = szTmp;
        if (fValue3 != 0)
        {
                sprintf(szTmp, "%.3f", fValue3);
                root["result"][ii]["v_avg"] = szTmp;
        }
}
else
{
        sprintf(szTmp, "%.1f", fValue1);
        root["result"][ii]["v_min"] = szTmp;
        sprintf(szTmp, "%.1f", fValue2);
        root["result"][ii]["v_max"] = szTmp;
        if (fValue3 != 0)
        {
                sprintf(szTmp, "%.1f", fValue3);
                root["result"][ii]["v_avg"] = szTmp;
        }
}
```

gizmocuz:
Thank you for the changes.

I am getting a bit confused with %.3f and %.2f ....

And why didn't you go to for a multiply of 100?

koenkooi:
> Thank you for the changes.
> 
> I am getting a bit confused with %.3f and %.2f ....

I can easily make everything %.3f if you want
 
> And why didn't you go to for a multiply of 100?

To avoid adding a host of `else if (type==general && type==foo)` blocks, I'm not familiar enough with this code to get all the exceptions right, so being allow to add to the current and voltage logic make it easier to track for me.

I also converted this to 'draft' while I'm playing around with all this. Keep the feedback coming!

gizmocuz:
As mentioned, %.1f does not 'chop' off the second digit, but rounds it. (1.37 -> 1.4)
Next, what is the accuracy of the pressure sensor? 
And, for a boiler, the pressure should be between 1.5 and 2 bar. , 1.6, 1.7, all fine. It really has no use to know if it is 1.52 bar.
Just program an alarm if it is outside the accepted range to notify you (for example with Blockly or dzVents)

As also mentioned above, you are now doing a multiply of 1000 instead of 100 and using %.2f

But I see no reason why to convert this to two digits, the sensor used is probably not that accurate at all

gizmocuz:
No feedback for over a month, closing issue

PR #6318, opened 2025-05-01
Netatmo Log-levels

JurgenLB:
With this I get the code ready for a new stable.
There are some loglines that where repetitive and very big so they are comment out.

Kind regards


gizmocuz:
Thank you!

PR #6323, opened 2025-05-08
Fix issue 6310

mealgi:
During the shutdown of TCPServer shared server make sure that Thread 2 (CTCPServerInt::handle_stop()) has completed its job before Thread 1 (Domoticz) deletes the TCP Server object.


gizmocuz:
Thank you for this PR!!

PR #6327, opened 2025-05-11
Support for Multicast DNS so User can browse to domoticz.local:<port> (no need to know IP-address)

kiddigital:
_This PR is a better/re-implementation replacing PR #6238 !_

This PR adds initial Multicast DNS support (mDNS), see [RFC6762](https://datatracker.ietf.org/doc/html/rfc6762)

Domoticz now announces itself on the local netwerk that it can be found at `domoticz.local:<port>` and listens to mDNS requests for mDNS queries that request details.

Next to A (IPv4) and AAAA (IPv6) resolving records, also domoticz is discoverably as an `_http._tcp`-service. The SRV record gives the resolvable name AND port. The extra TXT records report the `app` (domoticz), `path` (/) and `version` (for example 2025.1 (Build 16701)).

So by default, the mDNS name will be `domoticz.local`.
The user can specify something different under Settings / System / Location / Name. Then it will be `<thenamespecified>.local` (after restart!).

It is possible to disable mDNS by specifying `-nomdns` as a startup option.

To see more details of what is going on, enable the received debug flag via the commandline (`-debuglevel normal, received`).

mDNS support in Domoticz is also needed for future support for Matter as it utilizes mDNS (over IPv6) a lot. So this is a start.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -515,7 +515,7 @@ include_directories(${CMAKE_SOURCE_DIR}/main)
 ADD_CUSTOM_TARGET(revisiontag ALL)
 
 # creates appversion.h using cmake script
-ADD_CUSTOM_COMMAND(TARGET revisiontag COMMAND ${CMAKE_COMMAND}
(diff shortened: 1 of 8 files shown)


gizmocuz:
Thank you for this PR and time!!!

PR #6336, opened 2025-05-21
[MQTT] Repair Zwave RGB + WW

ericstaal:
Zwave js ui from version  10 uses colormode white is device as a warmwhite and or coldwhite channel.
This color mode can have warmWhite and or coldWhite. If it has both is also supports color temp mode.

This PR will use this white command to determine if it is RGBW or RGFBWW device.
Based on the device with will sue the correct color from domoticz to provide to the zwave device. Domoticz uses coldwhite for white channel and the zwave device can be warmWhite or coldWhite.

Tested with zipato E2 Bulb (RGBWW)
and using chinse RGBW(armhite) device.

Part of discovery message (which is used)
```
"color_temp_command_template": "{{ {'warmWhite': ((value - 245)|round(0)), 'coldWhite': (255 - (value - 245))|round(0))}|to_json }}",
"color_temp_value_template": "{{ '%03d%03d' | format((value_json.value.warmWhite || 0), (value_json.value.coldWhite || 0)) }}",
  
```

### DIFF
--- a/hardware/MQTTAutoDiscover.cpp
+++ b/hardware/MQTTAutoDiscover.cpp
@@ -520,6 +520,7 @@ void MQTTAutoDiscover::FixCommandTopic(std::string& command_topic, std::string& 
 void MQTTAutoDiscover::on_auto_discovery_message(const struct mosquitto_message* message)
 {
 	std::string topic = message->topic;
+
 	std::string org_topic(topic);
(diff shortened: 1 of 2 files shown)


gizmocuz:
Thank you for this PR!

I tested it on two Zigbee RGBWW devices and they also continue to work

PR #6345, opened 2025-06-02
Proposed PR to fix #6343

### LINKED ISSUE #6343: Issue with Domoticz.Configuration in Beta Versions
Hi Developers,

Has there been a change in the behavior of Domoticz.Configuration in the Beta version?

I use Domoticz.Configuration to store information about devices in Domoticz at the time of their creation. However, since the first Beta release after the stable 2025.1 version, I’ve been encountering the following error when attempting to write a dictionary to Domoticz.Configuration:

```
Error: (JSONtoPython) failed to add item '-1', to list for string.
Error: Tuya Cloud: Domoticz.Configuration operation failed: <built-in function Configuration> returned a result with an exception set
```

The data is stored once at the start of the plugin and read back on every heartbeat. I also added a routine to purge the Domoticz.Configuration, but that hasn’t resolved all the error messages.

I’m using the subroutine provided under "Developing a Python plugin."

Thanks in advance for any insights or suggestions.

Best regards,
Xenomes

pipiche38:
@dnpwwo  if you pass here your feedbacks/comments would be more than welcome.

As suggested also by @gizmocuz PY_SSIZE_T_CLEAN  could be the missing defined to enable '#'

### DIFF
--- a/hardware/plugins/PluginProtocols.cpp
+++ b/hardware/plugins/PluginProtocols.cpp
@@ -207,10 +207,16 @@ namespace Plugins {
 				}
 				else if (pRef.isConvertibleTo(Json::stringValue))
 				{
-					std::string sString = pRef.asString();
-					PyBorrowedRef pObj = Py_BuildValue("s#", sString.c_str(), sString.length());
-					if (!pObj || (PyList_SetItem(pRetVal, Index++, pObj) == -1)) // steals the ref to pObj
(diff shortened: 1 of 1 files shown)


pipiche38:
no memory leak found after 4 hours with heartbeat at 1s

gizmocuz:
Thank you for this PR!!

gizmocuz:
It's available in beta 16717

PR #6355, opened 2025-06-29
Preventing Netatmo plugin Crash

JurgenLB:
When a thermostat runs out of battery
the Rooms are not the same JSON and not all values are filled in.

This is to prevent a crash on that occasion.

### DIFF
--- a/main/SQLHelper.cpp
+++ b/main/SQLHelper.cpp
@@ -6072,9 +6072,6 @@ void CSQLHelper::ScheduleShortlog()
 		UpdateMultiMeter();
 		UpdatePercentageLog();
 		UpdateFanLog();
-		//Removing the line below could cause a very large database,
-		//and slow(large) data transfer (specially when working remote!!)
-		CleanupShortLog();
 	}
 	catch (boost::exception& e)
 	{
--- a/main/SQLHelper.h
+++ b/main/SQLHelper.h
@@ -387,6 +387,7 @@ public:
 	void CheckSceneStatusWithDevice(const std::string &DevIdx);
 
 	void ScheduleShortlog();
+	void CleanupShortLog();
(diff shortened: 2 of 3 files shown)


gizmocuz:
Thanks for the PR!

There is a debug variable enabled (NETATMO_SAVE_HOME)

Maybe a good idea to change them to #define and #ifdef like in all other hardware is possible?

JurgenLB:
Do you agree with my latest 'ifndef' change ?

gizmocuz:
Hmmm not 100% ;)

Do something like this



#ifdef _DEBUG
//	#define DEBUG_NetatmoWeatherStationR
//	#define NETATMO_CHECK_TOKEN
//	#define NETATMO_CHECK_SERVER
//	#define NETATMO_SAVE_HOME
//	#define NETATMO_SAVE_MODULE
#endif

then in your code

#ifdef NETATMO_SAVE_HOME
... code
#endif

gizmocuz:
Could I ask how you edit and test the code changes?
Are you using Visual Studio?

Did you test your changes? You could have almost copied the suggested code block in my previous email, but you forgot one essential part in your latest changes.... 😁 (the // in the front).
In your latest commit you enabled all debug flags

JurgenLB:
I'm testing this on a separated system.

Maybe I'm to quick to put it into master.

Will report back.

Kind regards

gizmocuz:
I would have 'remarked' them instead of deleting them....

JurgenLB:
remark or not I could not see a change in the output.

I can do a rebase so you have less commits.

gizmocuz:
Of course you don't see a change in the output, but now for other developers it is not easy to see which debug flags they can enable.
When you remark them, we could

Sure, please feel free to rebase

PR #6361, opened 2025-07-19
Preventing Netatmo plugin Crash

JurgenLB:
When a thermostat runs out of battery
the Rooms are not the same JSON and not all values are filled in.

This is to prevent a crash on that occasion.


gizmocuz:
Thank you for this PR!!

PR #6369, opened 2025-08-05
add net current meter to P1 base

### LINKED ISSUE #6367: Type in the code?
Hi,

Trying to add a a net ampere sensor (i will make a PR when it's ready), i think i saw an small bug in the SendCurrentSensor code:

The code in SendCurrentSensor  mixes up a high byte for Current with a low byte for temperature.  If your amperage stay below 25,6A (which in 99% of the househoulds will the case, since they have 3x25A) you will not notice weird behaviour, cause "tsen" is initalized at zero's, so tsen.TEMP.temperatureh will always be zero when this function is called

This does explain why I (while testing if my loadbalancing worked) did not see amperages enige 25 in domoticz when my load was about 30A. Fortunately i noticed it another way, so my fuse didn't blow while testing.

The correction would be I think: 

```
--- a/hardware/DomoticzHardware.cpp
+++ b/hardware/DomoticzHardware.cpp
@@ -772,17 +772,17 @@ void CDomoticzHardwareBase::SendCurrentSensor(const int NodeID, const int Batter
 
        int at10 = ground(std::abs(Current1 * 10.0F));
        tsen.CURRENT.ch1h = (BYTE)(at10 / 256);
-       at10 -= (tsen.TEMP.temperatureh * 256);
+       at10 -= (tsen.CURRENT.ch1h * 256);
        tsen.CURRENT.ch1l = (BYTE)(at10);
 
        at10 = ground(std::abs(Current2 * 10.0F));
        tsen.CURRENT.ch2h = (BYTE)(at10 / 256);
-       at10 -= (tsen.TEMP.temperatureh * 256);
+       at10 -= (tsen.CURRENT.ch2h * 256);
        tsen.CURRENT.ch2l = (BYTE)(at10);
 
        at10 = ground(std::abs(Current3 * 10.0F));
        tsen.CURRENT.ch3h = (BYTE)(at10 / 256);
-       at10 -= (tsen.TEMP.temperatureh * 256);
+       at10 -= (tsen.CURRENT.ch3h * 256);
        tsen.CURRENT.ch3l = (BYTE)(at10);
```

Unfortunately i cannot test, so i also did not make a PR, cause i cannot verify my working assumption. In my house amperages are at max 25A, So if I increase the amperage to more than 25A, i risk melting my main fuse  :-(   

I did notice that in other SendXXXSensor functions, are like the corrections i made, e.g. this is what's in the SendTempSensor :
```
int at10 = ground(std::abs(temperature * 10.0F));
tsen.TEMP.temperatureh = (BYTE)(at10 / 256);    
at10 -= (tsen.TEMP.temperatureh * 256);         
tsen.TEMP.temperaturel = (BYTE)(at10); 
```

akamming:
Adds net current meter (can be used for EVCC use case) to P1 base

also fixes https://github.com/domoticz/domoticz/issues/6367

Don't merge, yet, cause i have to wait till the sun comes up to really test (now only simulated tests), but would be nice if the code could be reviewed:  I had to make a change to RFCtrx.h. SO i don't know the impact on RFX devices (and can't test, cause i don't have one)

### DIFF
--- a/hardware/DomoticzHardware.cpp
+++ b/hardware/DomoticzHardware.cpp
@@ -771,19 +771,22 @@ void CDomoticzHardwareBase::SendCurrentSensor(const int NodeID, const int Batter
 	tsen.CURRENT.battery_level = BatteryLevel;
 
 	int at10 = ground(std::abs(Current1 * 10.0F));
-	tsen.CURRENT.ch1h = (BYTE)(at10 / 256);
-	at10 -= (tsen.TEMP.temperatureh * 256);
+	tsen.CURRENT.ch1h = (BYTE)(at10 / 256 & 0x7f);
+	at10 -= (tsen.CURRENT.ch1h * 256);
 	tsen.CURRENT.ch1l = (BYTE)(at10);
+	tsen.CURRENT.temp1sign = (Current1 >= 0) ? 0 : 1;
 
 	at10 = ground(std::abs(Current2 * 10.0F));
-	tsen.CURRENT.ch2h = (BYTE)(at10 / 256);
-	at10 -= (tsen.TEMP.temperatureh * 256);
+	tsen.CURRENT.ch2h = (BYTE)(at10 / 256 & 0x7f);
+	at10 -= (tsen.CURRENT.ch2h * 256);
 	tsen.CURRENT.ch2l = (BYTE)(at10);
-
+	tsen.CURRENT.temp2sign = (Current2 >= 0) ? 0 : 1;
+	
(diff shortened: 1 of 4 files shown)


gizmocuz:
I have seen your post on the Forum, and as mentioned, you can do this via dzVents

The main reason this is closed, is because you are NOT allowed to change the RFXtrx.h file as this is copyrighted by RFXCom and could be overwritten by any API change

akamming:
ah that's a bummer...  sun just started to shine and i noticed it worked ;-)

but understandable...wasn't aware of this.. tx!

PR #6373, opened 2025-08-13
make temperature series invisible in Humidity only device log

waltervl:
Solves reported issue for Humidity only devices where the charts still show Temperature labels. https://forum.domoticz.com/viewtopic.php?f=6&t=43860

### DIFF
--- a/www/app/log/TemperatureLog.js
+++ b/www/app/log/TemperatureLog.js
@@ -257,6 +257,7 @@ define(['app', 'RefreshingChart', 'DataLoader', 'ChartLoader', 'log/Chart', 'log
         return {
             id: 'temperature',
             dataItemKeys: ['te'],
+            showWithoutDatapoints: false,
             label: 'Te',
             template: {
                 name: $.t('Temperature'),
@@ -316,6 +317,7 @@ define(['app', 'RefreshingChart', 'DataLoader', 'ChartLoader', 'log/Chart', 'log
         return {
             id: 'setpointrange',
             dataItemKeys: ['sm', 'sx'],
+            showWithoutDatapoints: false,
             dataItemIsComplete: function (dataItem) {
                 return dataItem.se !== undefined;
             },
@@ -342,6 +344,7 @@ define(['app', 'RefreshingChart', 'DataLoader', 'ChartLoader', 'log/Chart', 'log
         return {
             id: 'prev_setpoint',
             dataItemKeys: ['se'],
+            showWithoutDatapoints: false,
             useDataItemsFromPrevious: true,
             showWithoutDatapoints: false,
             label: 'Sp',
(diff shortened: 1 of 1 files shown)


gizmocuz:
What is the type of this sensor in the devices tab? (Humidity) ? In the database it should be 0x51

We probably would need another graph display for this type if it is not already there

What type of sensor is it? Is it received by RF/Zigbee/ZWave ? What is the brand?

waltervl:
I tested this on a Dummy Humidity sensor. As I am not at home now I cannot see the exact type. It works fine now with this change besides of the compare function (giving no data). Also for the THB or TH sensors it still works fine.

As there is a code line
`chartName: ctrl.device.Type === 'Humidity' ? $.t('Humidity') : $.t('Temperature'),`  I presume TemperatureLog.js is also used for Humidity devices.

waltervl:
It is the Lacrosse TX3 as shown in the screenshot. Visible in the Temperature Tab.
Screenshot_20250815-134215 Firefox

gizmocuz:
So it seems a dummy sensor, probably made with data from Buienradar? Why the need for a Humidity only sensor and not a combination with temp and/or baro?

I will see if I can solve this, for now I suggest to invest in a sensor that is just a few dollars more

waltervl:
To test the issue that someone mentioned in the forum I made a dummy, indeed filled with buienradar data. But that is not the issue.

You can make a dummy humidity sensor so it is then also better to have it shown correctly on the log page. So you only have to merge the PR :-) If you want to fix the compare function too that can be done separately whenever you feel like it...

gizmocuz:
@waltervl , I will do so.... Yes, it's best to do complete fix... 
Your solution is one way... another way is to 'hide' it... so temp sensors that have not received any data for the last week still show a temp axis

gizmocuz:
I am going to add some small changes. In your PR, for setpoint the patch was not needed (as it's not a temp/humidity devicetype), but the ShowWithoutDatapoints was already set to false there, so it was set twice

gizmocuz:
Is it possible to test the latest beta to see if this solves the issue?

waltervl:
Yes, already tested yesterday, looks good now.

RuiPereira:
Hi, I’ve been using Domoticz under Docker for a while. I updated to the “latest” image (16759) a few days ago, and none of my temperature and temperature/humidity sensors is showing the Temp axis on the logs anymore. Has this been reported? Thanks.

waltervl:
Yes, you are right, I did not test enough, indeed the temperature axis on the left has disappeared for all temperature devices....

gizmocuz:
I did not test good enough! ;) ... Found the issue... standby for a fix

gizmocuz:
Found it, and also places the Y-Axis on the left for a Humidity only sensor

Should be fixed in beta 16762 which is available now

RuiPereira:
Thank you, everything's back to normal.

PR #6375, opened 2025-08-17
Change default MQTT press payload to "PRESS"

### LINKED ISSUE #6374: ESPHome MQTT Button press doesn't work. Default press_payload should really be "PRESS" instead of "1"
ESPHome buttons do not set payload_press or pl_prs in their autodiscovery message. As a result domoticz uses its own default, which is "1" (from hardware/MQTTAutoDiscover.h). This doesn't work, because ESPHome buttons expect a payload of "PRESS".

```
1 blinds-b76d10/button/reboot/command	1
102 blinds/debug	[W][mqtt.button:023][async_tcp]: 'Reboot': Received unknown status payload: 1
100 blinds/debug	[W][component:167][async_tcp]: Component mqtt set Warning flag: unspecified
```

Changing the default payload_press to "PRESS" in hardware/MQTTAutoDiscover.h fixes the problem:

```
5 blinds-b76d10/button/reboot/command	PRESS
70 blinds/debug	[D][button:010][async_tcp]: 'Reboot' Pressed.
81 blinds/debug	[I][restart.button:012][async_tcp]: Restarting device...
69 blinds/debug	[I][app:149][async_tcp]: Rebooting safely...
```

On a side note, the condition on line 3496 of hardware/MQTTAutoDiscover.h is always true, because payload_press can never be empty due to the default assignment above.

```
        else if (pSensor->component_type == "button")
        {
                switchType = STYPE_PushOn;
                if (!pSensor->payload_press.empty())
                        szSwitchCmd = pSensor->payload_press;
                else
                        szSwitchCmd = "on";
        }
```

ndilieto:
Fixes https://github.com/domoticz/domoticz/issues/6374

### DIFF
--- a/hardware/MQTTAutoDiscover.h
+++ b/hardware/MQTTAutoDiscover.h
@@ -55,7 +55,7 @@ class MQTTAutoDiscover : public MQTT
 		std::string payload_open = "OPEN";
 		std::string payload_close = "CLOSE";
 		std::string payload_stop = "STOP";
-		std::string payload_press = "1";
+		std::string payload_press = "PRESS";
 		int position_open = 100;
 		int position_closed = 0;
 


gizmocuz:
Thanks for the PR and your time! Maybe this has been changed some time ago from 1 to PRESS?

If the config object had a '**payload_press**' field, it would have used it

PR #6386, opened 2025-09-10
Build fixes for Boost 1.89.0+

arjendekorte:
Since version 1.69.0, Boost.System is header-only. A stub library was still built for compatibility, but linking to it was no longer necessary. Version 1.89.0 removed the stub library, so require 1.69.0+ and get rid of the library. It also removed deadline_timer, basic_deadline_timer and time_traits from the convenience header boost/asio.hpp, so we need to #include <boost/asio/deadline_timer.hpp> for deadline_timer where used.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,7 +13,7 @@ INCLUDE(TestBigEndian)
 #
 #
 ## required min. libBoost version
-SET(DOMO_MIN_LIBBOOST_VERSION 1.66.0)
+SET(DOMO_MIN_LIBBOOST_VERSION 1.69.0)
 ##
 
 ### USER-SETTABLE OPTIONS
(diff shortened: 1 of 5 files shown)


gizmocuz:
Thank you for this PR!

One question, could you remove the comment on the lines of the inclusion of the deadline_timer? I think it is quite obvious that this file is included for this timer by the filename 😁

arjendekorte:
I can. The reason that I added this, is that I copied this line from line 5 in `hardware/ASyncTCP.h`. You'll find a lot of obvious comments there too. Do you want to remove those too? Most of these #include's would be superfluous anyway, had line 3 in `hardware/ASyncTCP.cpp` been moved there instead.

gizmocuz:
If you could, yes please.

arjendekorte:
> If you could, yes please.

I removed the additional comments already, do you mean to cleanup `hardware/ASyncTCP.h` and `hardware/ASyncTCP.cpp` as well?

gizmocuz:
Thank you very much for this PR and work!
I will cleanup these two files ;)

PR #6388, opened 2025-09-13
Feature/ModelContextProtol support for use with LLM clients (domoticz LLM Agent)

kiddigital:
This PR adds support for the [Model Context Protocol](https://modelcontextprotocol.io/specification/2025-06-18) (MCP)  to domoticz so domoticz can be used as an agent by LLM clients.

For safety reasons, it is not enabled by default. It has to be enabled explicitly.
To enable it, start domoticz with the `-mcp` flag (A status message should tell that support is enabled).

_NOTE_: Make sure that your LLM client (like Claude or any other) is either running within the private network range of domoticz (configure it in the security settings) or supports OAuth2/OIDC authentication. Otherwise the client gets rejected due to no authorization. When using OAuth2, please create an _application_ with a name and secret.

This PR currently implements some basic stuff.
_Only specification version 2025-06-18 is supported!_
_Only basic Streamable HTTP support (no STDIO or SSE)_ (And not really _streamable_, just stateless HTTP calls)

__resources__
Available devices are offered as `resources`.
Same for the floorplans.

__tools__
At the moment the following `tools` are offered:

- get_switch_state
- toggle_switch_state
- get_sensor_value
- get_logging
- get_floorplan
- set_thermostat_setpoint

More will follow if needed (which is to be expected :) )

__prompts__
The following `prompts` are suggested:
- _housesummary_
- _systemanalysis_

**Example use : Claud Desktop**
Configure local MCP Servers in Claude desktop. As Claude Desktop only supports STDIO for standard users (Pro/MAX users should get remote MCP support), I used [MCP-Proxy](https://github.com/sparfenyuk/mcp-proxy) as an intermediate to do the STDIO<->Streamable HTTP translation.

```JSON
    "dom-mcp-proxy": {
        "command": "<location of the proxy software on the filesystem>/mcp-proxy.exe",
        "args": ["--transport", "streamablehttp", "http://domoticz.local:8080/mcp"],
        "env": {
          "API_ACCESS_TOKEN": "<A valid JWT or put your client in the Private Network range>"
        }		
    },
```

**Example use : VS Code**
In Visual Studio Code, you can configure MCP servers. Via the _command palette_  (See _View_ / CTRL+SHIFT+P) you can __MCP: Add Server...__. Or at _Extensions_ (at the bottom their is a _MCP Servers - Installed_ section).
```
{
    "servers": {
        "dommcp": {
            "type": "http",
            "url": "http://127.0.0.1:8080/mcp"
        }
    }
}
```
When you press _start server_, it will try to connect and the first time it will go through the Authorization process. You have to register an _Application_ within domoticz (for example 'domoticzMCP' and an _Application Secret_) which is needed as ClientID and ClientSecret as asked for by VSCode to complete the Authorization process.

__NOTE:__ If Authorization is successful or failed in VSCode, you can see the information under the _Account_ info (bottom left of the screen). Here you can also delete that info so it you can re-enter it if needed.

### DIFF
--- a/mcpserver/McpService.cpp
+++ b/mcpserver/McpService.cpp
@@ -1,0 +1,61 @@
+/*
+ * McpService.cpp
+ *
+ *  Created on: 4 April 2025
+ *      Author: kiddigital
+ * 
+ * 
+ * It contains routines that are part of the WebServer class, but for sourcecode management
+ * reasons separated out into its own file so it is easier to maintain the MCP related functions
+ * of the WebServer. The definitions of the methods here are still in 'main/Webserver.h'
+ *  
+*/
+
+#include "stdafx.h"
+#include <iostream>
+#include <json/json.h>
+#include "../main/Logger.h"
+#include "../main/SQLHelper.h"
+#include "../httpclient/UrlEncode.h"
+#include "../main/WebServer.h"
+#include "../webserver/Base64.h"
+
+#define __STDC_FORMAT_MACROS
+#include <inttypes.h>
+
+namespace http
+{
+	namespace server
+	{
+
+		void CWebServer::GetMcpSse(WebEmSession &session, const request &req, reply &rep)
+		{
+			_log.Debug(DEBUG_RECEIVED, "GetMcpSse (%d): %s (%s)", req.content_length, req.content.c_str(), req.uri.c_str());
+			// Check if the request is valid
+			if (req.get_req_header(&req, "Accept") != nullptr)
+			{
+				std::string accept = req.get_req_header(&req, "Accept");
+				if (accept.find("text/event-stream") == std::string::npos)
+				{
+					rep = reply::stock_reply(reply::bad_request);
+					return;
+				}
+			}
+			
+			// Set headers for SSE
+			reply::add_header(&rep, "Content-Type", "text/event-stream");
+			//reply::add_header(&rep, "Cache-Control", "no-cache");
+			//reply::add_header(&rep, "Connection", "keep-alive");
+		
+			Json::Value jsonRPC;
+			jsonRPC["jsonrpc"] = "2.0";
+			jsonRPC["result"] = "Welcome to the Domoticz Model Context Protocol ServerSideEvents\n\n";
+			jsonRPC["id"] = 1;
+
+			// Set response content
+			//rep.content = jsonRPC.toStyledString();
+			rep.status = reply::ok;
+		}
+
+	} // namespace server
+} // namespace http
--- a/mcpserver/McpService.cpp
+++ b/mcpserver/McpService.cpp
@@ -244,6 +244,18 @@ namespace http
     "nextCursor": "next-page-cursor"
   }
 }
+
+{
+  "type": "resource_link",
+  "uri": "file:///project/src/main.rs",
+  "name": "main.rs",
+  "description": "Primary application entry point",
+  "mimeType": "text/x-rust",
+  "annotations": {
+    "audience": ["assistant"],
+    "priority": 0.9
+  }
+}
 */
 			// Prepare the result for the tools/list method
 			jsonRPCRep["result"]["tools"] = Json::Value(Json::arrayValue);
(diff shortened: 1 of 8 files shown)


mcpserver/McpService.cpp
```diff
@@ -0,0 +1,847 @@
...
+#include <inttypes.h>
+
+extern http::server::CWebServerHelper m_webservers;
+extern CLogger _log;
+
+namespace http
+{
+	namespace server
+	{
+
```
gizmocuz: Please remove empty line

mcpserver/McpService.cpp
```diff
@@ -0,0 +1,847 @@
...
+					rep = reply::stock_reply(reply::bad_request);
+					return;
+				}
+			}
+			// Check if the request is a POST request
+			if (req.method != "POST")
+			{
+				_log.Debug(DEBUG_WEBSERVER, "MCP: Invalid method: %s", req.method.c_str());
+				rep = reply::stock_reply(reply::bad_request);
+				// VScode MCP client does sends GET's
```
gizmocuz: Please move the comment above the log line

mcpserver/McpService.cpp
```diff
@@ -0,0 +1,847 @@
...
+				_log.Debug(DEBUG_WEBSERVER, "MCP: Handling notification %s (do nothing).", sReqMethod.c_str());
+				rep = reply::stock_reply(reply::no_content);
+				return;
+			}
+
+			Json::Value jsonRPCRep;
+			jsonRPCRep["jsonrpc"] = "2.0";
+
+			// Check if the request has an ID
+			std::string sReqID;
```
gizmocuz: Could you declare these two variables in the scope below after

`if (jsonRequest.isMember("id"))`

mcpserver/McpService.cpp
```diff
@@ -0,0 +1,847 @@
...
+		jsonRPCRep["result"]["tools"].append(tool);
+	}
+
+	void McpToolsCall(const Json::Value &jsonRequest, Json::Value &jsonRPCRep)
+	{
+		// Check if the required parameters are present
+		if (!jsonRequest.isMember("params") || !jsonRequest["params"].isMember("name"))
+		{
+			_log.Debug(DEBUG_WEBSERVER, "MCP: Missing required tool parameter 'name' in tools/{tool} request.");
+			jsonRPCRep["error"]["code"] = -32602; // Invalid params
```
gizmocuz: Maybe better for the feature to use "#defines" instead of hard coded numbers?

`#define MCP_INVALID_PARAMETER -32602`

mcpserver/McpService.cpp
```diff
@@ -0,0 +1,847 @@
...
+				}
+			}
+		}
+		// Add any available floorplans as resources too
+		auto result = m_sql.safe_query("SELECT ID, Name FROM Floorplans");
+		if (!result.empty())
+		{
+			for (const auto &row : result)
+			{
+				if (row.size() >= 2)
```
gizmocuz: This line is not needed, as you debug anyway you know you have the correct query above

mcpserver/McpService.cpp
```diff
@@ -0,0 +1,847 @@
...
+	}
+
+	void McpResourcesTemplatesList(const Json::Value &jsonRequest, Json::Value &jsonRPCRep)
+	{
+		_log.Debug(DEBUG_WEBSERVER, "MCP: Handling resources/templates/list request.");
+
+		// Prepare the result for the resources/templates/list method
+		jsonRPCRep["result"]["resourceTemplates"] = Json::Value(Json::arrayValue);
+		// Currently we do not have any resource templates to offer
+
```
gizmocuz: Please remove empty line

gizmocuz (changes requested):
Looking great! Just a few remarks here and there....

mcpserver/McpService.cpp
kiddigital: Removed

mcpserver/McpService.cpp
kiddigital: Good idea... added defines

mcpserver/McpService.cpp
kiddigital: Rewritten so variables not needed anymore

PR #6389, opened 2025-09-17
Added some sections and headings

ml17950:
Added some sections and headings on different pages.

Example:
example

### DIFF
--- a/www/app/LogController.js
+++ b/www/app/LogController.js
@@ -119,10 +119,11 @@ define(['app'], function (app) {
 
 		$scope.ResizeLogWindow = function () {
 			var pheight = $(window).innerHeight();
-			$("#logcontent #logdata").height(pheight - 160);
-			$("#logcontent #logdata_status").height(pheight - 160);
-			$("#logcontent #logdata_error").height(pheight - 160);
-			$("#logcontent #logdata_debug").height(pheight - 160);
+			var poffset = 210;
+			$("#logcontent #logdata").height(pheight - poffset);
+			$("#logcontent #logdata_status").height(pheight - poffset);
+			$("#logcontent #logdata_error").height(pheight - poffset);
+			$("#logcontent #logdata_debug").height(pheight - poffset);
 		}
 
 		init();
--- a/www/app/hardware/Hardware.html
+++ b/www/app/hardware/Hardware.html
@@ -450,1228 +450,1230 @@
 	<div class="page-header-small">
 		<h3 data-i18n="Add/Update Hardware">Add/Update Hardware</h3>
 	</div>
-	<table class="display" id="hardwareparamstable" border="0" cellpadding="0" cellspacing="20">
-		<tr>
-			<td align="right" style="width:80px"><span data-i18n="Enabled"></span>:</td>
-			<td><input type="checkbox" id="enabled" checked><label for="enabled"></label></td>
-		</tr>
-		<tr>
-			<td align="right" style="width:110px"><label for="hardwarename"><span data-i18n="Name"></span>:</label></td>
-			<td><input type="text" id="hardwarename" style="width: 250px; padding: .2em;" class="text ui-widget-content ui-corner-all"></td>
-		</tr>
-		<tr>
-			<td align="right" style="width:110px"><label for="combotype"><span data-i18n="Type"></span>:</label></td>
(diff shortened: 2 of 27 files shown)


gizmocuz:
Thank you for this PR.

Some comments:

Add/Update text, could this be a separate data-i18n span, as otherwise we introduce new strings to be translated, better to have only one string to be translated 'Add/Update' and 'Hardware', 'Users' etc are already present

But, it is not consistent, 'Add/Update Hardware' and 'Edit Uservariable'

It is also not Add/Update, but Add/Update/Delete

Do we need this text? Is it not clear?

I am personally not a fan of this white divider line... maybe give it another color? 
Maybe something in the trend of RGB(0,162,232) ? (Not tested, but could be nice?)

ml17950:
> I am personally not a fan of this white divider line... maybe give it another color? Maybe something in the trend of RGB(0,162,232) ? (Not tested, but could be nice?)

Just used the existing class "page-header-small".
But here are some other colors for the line.
1,2 is the color from the settings tab
3 is your RGB(0,162,232)
4 is the background of the inputs.

v1
v2
v3
v4

ml17950:
> Add/Update text, could this be a separate data-i18n span, as otherwise we introduce new strings to be translated, better to have only one string to be translated 'Add/Update' and 'Hardware', 'Users' etc are already present

ok, better idea
 
> But, it is not consistent, 'Add/Update Hardware' and 'Edit Uservariable'

Just used already existing wordings.

> It is also not Add/Update, but Add/Update/Delete
> 
> Do we need this text? Is it not clear?

Maybe here an "Edit Hardware" or "Change Hardware" instead of "Add/Update/Delete"?

ml17950:
Maybe also with a border around the lower content:

Screenshot 2025-09-19 164315

gizmocuz:
Looking great!

Some things I noticed

- Custom Icons, when you select an icon, it looks 3 sections are 'clued' together:

image

Maybe some more spacing between them

Are there any new strings that need to be translated? Could you name them here? (I will then add them to the file/transifex system)

ml17950:
> Looking great!

Thanks

> Custom Icons, when you select an icon, it looks 3 sections are 'clued' together:
> 
> Maybe some more spacing between them

I'll look at it
 
> Are there any new strings that need to be translated? Could you name them here? (I will then add them to the file/transifex system)

Just layout until now. Text will follow.

ml17950:
> Add/Update text, could this be a separate data-i18n span, as otherwise we introduce new strings to be translated, better to have only one string to be translated 'Add/Update' and 'Hardware', 'Users' etc are already present

It's a little bit tricky, I think.

For example in Custom Icons:

It's now "Edit Custom Icon" as one translation string.
This could be "Edit" and "Custom Icon" and that's fine in english.

In i.e. german the translations where "Bearbeiten" and "Benutzerdefiniertes Symbol", but the right translation would be "Benutzerdefiniertes Symbol bearbeiten" (order swapped and capitalisation).

So I think, it's better to add some new translations in some locations without splitting the words.

gizmocuz:
Okey no problem... if you can write the lines here that have to be translated I will put them in the system

ml17950:
Morning.

Here are the new strings:

```
"Edit Icon": "Edit icon",
"Add/Edit Hardware": "Add/Edit hardware",
"Add/Edit Application": "Add/Edit application",
"Add/Edit User": "Add/Edit user",
"Add/Edit Variable": "Add/Edit variable",
"Add/Edit Timerplan": "Add/Edit timerplan",
```

gizmocuz:
Do we need actually these strings? Is it not obvious that you edit the settings there?

ml17950:
> Do we need actually these strings? Is it not obvious that you edit the settings there?

Or maybe the headings of the sections will all be "General settings"?

gizmocuz:
I think it is pretty clear when you select a user that you can 'edit' it
All these texts are quite obvious and not really needed

ml17950:
Morning.

I changed the translations strings to existing ones.

gizmocuz:
At a lot of places we now see only 'Add/Edit', instead of 'Add/Edit Hardware'

Maybe it is best if you undo all the these changes... leave them as they where....

ml17950:
Headings are removed

gizmocuz:
Looking great!!! Thank you very much again!!!

PR #6390, opened 2025-09-18
Improvement/oauth2appsupport allow custom URI scheme's and loopback devices in redirect URI's

kiddigital:
This PR extends/expands the IamService to allow for custom/private URI scheme's and loopback device's as in [RFC 8252](https://datatracker.ietf.org/doc/html/rfc8252#section-7) as redirect URI's in an OAuth2/OIDC authorization flow.

With this extensions, in is now possible that mobile app's can register their own _redirect URI scheme_ allowing to go from the mobile APP to the domoticz authorization screen (on the web in a browser) and then being redirect back to the App (using the registered URI scheme).

Another improvement is that the _Security Realm_ is now in sync with the mDNS name so that the `<your domoticz server>/.well-known/openid-configuration` offers the correct OAuth2/OIDC endpoints (as needed by client that discover the endpoints as intended).

### DIFF
--- a/iamserver/IamService.cpp
+++ b/iamserver/IamService.cpp
@@ -43,7 +43,7 @@ namespace http
 			std::string code_challenge = request::findValue(&req, "code_challenge");
 			std::string code_challenge_method = request::findValue(&req, "code_challenge_method");
 
-			if (!redirect_uri.empty() && redirect_uri.substr(0,8) == "https://")	// Absolute and (TLS)safe redirect URI expected
+			if (!redirect_uri.empty() && ValidRedirectUri(redirect_uri))
 			{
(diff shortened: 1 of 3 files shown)


galadril:
If i compare this with for example Okta, then you need to configure the allowed redirect_url as part of your application/client setup. So you then configure the allowed redirect_url for that client id.

Would that not be better then a validation method?

kiddigital:
@galadril , that would indeed be even better.

But would _complicate_ the Application registration process a bit more as the user needs to know these redirect URI('s).

And still these URI should pass the test to comply to the specifications 😃

(And more work in domoticz to allow maintenance and storing. That was also a reason to not implement this initially 😉)

gizmocuz:
Thanks!

PR #6408, opened 2025-10-22
Update the smoke detector device history when ResetSecurityStatus in invoked (either via JSON or User Interface)

mealgi:
Update the smoke detector device history when ResetSecurityStatus in invoked (either via JSON or User Interface)

The objective is to record in smoke device history the ResetSecurityStatus action processed by the user after fixing physically the issue.

Prior to the current PR the state was recorded directly in DeviceStatus database but the information about who processed the ResetSecurityStatus action and when was lost


gizmocuz:
Thank you for this PR!!

PR #6411, opened 2025-10-24
adding Room mode switch and code refactoring.

### LINKED ISSUE #6396: Feature request : Netatmo Thermostat add endtime in each room (and correction on the room mode)
Hello,

Is it possible to add some device for the endtime parameter on each room ?
I use Domoticz to lower the required temperature when windows are opened, but I need to set a timer more important than 1 hour (I return to the home mode only when windows are closed).

I have also a problem with the mode device on each room : at this time it's the same mode than the thermostat. It should be the mode of the room : manual, max, home (as described in the API documentation :

> /setroomthermpoint Control the heating of a room
> This endpoint permits to control the heating of a specific room. A room can be set in 3 differents modes:
> 
> "manual" mode in which setpoints can be applied for a specific duration (use endtime)
> 
> "max" mode which will put the setpoint at max (30°) for a given duration (default is xx and can be changed by the user in its settings)
> 
> "home" mode which basically makes the room following the Home mode.

Thanks

JurgenLB:
This is to close #6396 so the user can return to the automatic schedule before the end time is reached.

Also some code refactoring.

I tested this with every scope possible for myself separately.

Kind Regards

Jurgen

### DIFF
--- a/hardware/Netatmo.cpp
+++ b/hardware/Netatmo.cpp
@@ -546,9 +546,9 @@ std::string CNetatmo::WindDirection(float dDirection)
 /// </summary>
 std::string CNetatmo::bool_as_text(bool b)
 {
-                std::stringstream converter;
-                converter << std::boolalpha << b;   // flag boolalpha calls converter.setf(std::ios_base::boolalpha)
-                return converter.str();
+	std::stringstream converter;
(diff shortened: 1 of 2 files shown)


gizmocuz:
Then you very much for this PR again! 👌

PR #6414, opened 2025-10-30
Adjusting status of Room Switch

JurgenLB:
This is to fix a issue for the state off the Room switch,
so it follows the status off the thermostat.


gizmocuz:
Thank you again!

PR #6416, opened 2025-10-31
Match the poll interval to the refresh interval

Jan-nl:
Because the poll interval can be increased now, the token refresh interval/loop does not match the refresh token cycle, resulting in errormessages and a spoiled poll cycle. This fix computes the token refresh cycle base on the poll interval to match the 600 seconds Tado token refresh cycle.

### DIFF
--- a/hardware/Tado.cpp
+++ b/hardware/Tado.cpp
@@ -239,7 +239,7 @@ bool CTado::GetAccessToken()
 	else
 	{
 		//Mainly for debug reasons
-		Log(LOG_STATUS, "AccessToken refreshed.");
+		Log(LOG_STATUS, "Access Token refreshed.");
 	}
 
 	Json::Value root;
--- a/hardware/Tado.cpp
+++ b/hardware/Tado.cpp
@@ -52,7 +52,14 @@ CTado::CTado(const int ID, const int PollInterval)
 		m_szRefreshToken = result[0][0];
 	}
 	if ((PollInterval > 10) && (PollInterval < 3600))
+	{
 		m_iPollInterval = PollInterval;
+		//Compute new maxloops based on pollinterval, if maxloops < 1, set to 1 to be save
+		//Basically the 500 should be a define I guess, being the refresh intervall
+	    m_iTADO_TOKEN_MAXLOOPS = 500 / m_iPollInterval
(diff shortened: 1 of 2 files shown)


hardware/Tado.cpp
```diff
@@ -229,6 +236,11 @@ bool CTado::GetAccessToken()
 		Log(LOG_ERROR, "Failed to get a Refresh Token");
 		return false;
 	}
+	else
+	{
+		//Mainly for debug reasons
+		Log(LOG_STATUS, "Access Token refreshed.");
```
kiddigital: @Jan-nl , please make it a debug log statement.

gizmocuz:
Thank you for this PR.

Question, what's up with the 500 number?

Is it not way better to store the current time in a variable when the token was received, check the refresh interval (timeout) of this token, and use this to calculate when the token needs to be refreshed? (minus a few minutes to be sure it is refreshed on time)

You also added a log line that the token is refreshed every time it is refreshed. Could you make it a debug log message?

Jan-nl:
The 500 is the ´'to be safe´' number of seconds for the refresh token action, compared to the 12 in the MAXLOOPS in the tado.h. In the present code 600seconds is mentioned to be the token refresh interval. I will add a variable/constant for the 500 secs.

I also will change the log line to debug. That was the idea, but I forgot.

gizmocuz:
@Jan-nl , is the time to refresh the token not advertised in the returning result when you request a new token?
What if they change it to 15 minutes? or 2 hours, of 1 day?
Better to use this reported interval and check against that? (and with a time_t value instead of 'loops' counter)?

Jan-nl:
@gizmocuz basically I agree your approach is the better one. For now I proceeded with the present implementation, also because of my not so great knowledge of cpp.

Nevertheless, you give me a challenge. Is it wise to accept this pr for now while I dive into the solution you propose?

gizmocuz:
Sure I'll accept it , thanks!!! But for safeness maybe rewrite that part in the feature

somethink like (psuedio)

time_t atime = time(nullptr)
if (atime - last_received_token > token_refresh_time - 60) {
 //1 minute earlier... maybe do 3 or 5
 refresh_token_function() -> last_received_token = time(nullptr)
}

PR #6417, opened 2025-11-02
Add support for Indoor Camera Advance

JurgenLB:
This is to have support for the Events from the New Indoor Camera Advance
I tested this with a family members account.

I hope they don't make anymore new scopes.

Kind Regards


JurgenLB:
What do you think off the overlay in DASHBOARD, SWITCHES and UTILITY tab with a long device name?
should the names be shorter or a fixed length ?

kind regards

Domoticz Switch 1
Domoticz Switch 2
Domoticz Utility 3

gizmocuz:
Thanks for the PR! I hope the same for you! 😂

I think the end user could rename the devices themself and the initial (long) device name helps them to find the right device name.

PR #6418, opened 2025-11-03
Oauth2 fixes for Alexa

dwmw2:
I've been working on getting Alexa integration working with our OAuth2 support: https://github.com/dwmw2/alexa_domo

1. Our default refresh lifetime for a token is only a single day, and I think that means the user would have to relink the Alexa skill if they go a day without invoking it. So I've made it a per-client setting.

2. Alexa Smart Home skills also recommend using Basic auth, which is mandatory in RFC6749. Alexa did work with the 'credentials in request' method, and I've documented that the user should do so, but it's one more thing for a user to trip over, so it's better to fix it and simplify the docs.

3. The [password grant flow](https://datatracker.ietf.org/doc/html/rfc6749#section-4.3) was previously looking for user creds in the Basic auth header, when they should be in the body of the request. Fixed that to get them from the right place.

4. **[SECURITY]** The password flow also wasn't even checking the client creds at all! Pull the client cred checking right up to the start of the function and do it once, instead of leaving it to be checked in each individual grant flow and potentially forgotten.

5. **[SECURITY]** Signing JWTs with the client_secret means that the client can just _make up_ tokens for any user it wants, without actually having to log in. So for example I've carefully set up an `alexa` user with only the permissions I want it to have, but setting up a client cred for Alexa to do OAuth2 has given it a way to fake tokens for *any* admin user in my system! Fix this by generating a new signing secret for each client, which is known *only* to Domoticz. The old tokens signed with the client secret are accepted only for 24 hours after the upgrade, after which they will have expired anyway, and any *refreshed* tokens are signed with the new signing secret. So we upgrade, and don't even break any existing tokens in the process.

6. I've made it easier to get the correct URIs in as the JWT issuer, and in the `/.well-known/openid-configuration` result. By using the content of the `Host:` header in the request, *if* the hostname is still set to the default 'domoticz.local'. This fixes OAuth2 for people whose Domoticz server is accessible only through port forwarding on a hostname and port number that Domoticz has no way of knowing.

I'm slightly unsure about the last one, as it dilutes the meaning of the 'issuer' field in the JWT. A client can put *anything* in the Host: header, and that means we can ask Domoticz to issue tokens with anything in the `iss` field. I don't *think* that matters, as these tokens are only for use with this one Domoticz server anyway. Similarly, Domoticz does currently check the issuer when accepting a token. And now a client can just present a Host: header which matches the issue from the token, diluting the issuer check... but who cares? If that token is correctly signed and accepted by the local Domoticz server, checking the 'issuer' string doesn't do anything anyway.

Having the correct URI in the issuer of the JWT is really helpful because the Alexa Lambda function can then use it and not have to be explicitly configured with the server's hostname and port. My current documentation just tells users to ensure Domoticz is running with the right `-vhostname` argument, but that doesn't solve the port issue, for those who are using port forwarding. And since Lambda can only use Legacy IP and not IPv6, *most* people are likely to be using port forwarding for external access, with Domoticz itself not even knowing which *external* port number is used.

@kiddigital what do you think?

### DIFF
--- a/iamserver/IamService.cpp
+++ b/iamserver/IamService.cpp
@@ -386,69 +386,60 @@ namespace http
 
 						// Maybe we should only allow this when in a safe 'local' network? So check if request comes from local networks?
 						int iUser = -1;
-						int iClient = -1;
-						if (request::get_req_header(&req, "Authorization") != nullptr)
+						int iClient = FindClient(client_id.c_str());
+
+						if (iClient != -1)
 						{
-							std::string auth_header = request::get_req_header(&req, "Authorization");
-							// Basic Auth header
-							size_t npos = auth_header.find("Basic ");
-							if (npos != std::string::npos)
+							std::string username = request::findValue(&req, "username");
+							std::string password = request::findValue(&req, "password");
+							if (!username.empty() && !password.empty())
 							{
-								std::string decoded = base64_decode(auth_header.substr(6));
-								npos = decoded.find(':');
-								if (npos != std::string::npos)
+								_log.Debug(DEBUG_AUTH, "OAuth2 Access Token: Password grant for user (%s)", username.c_str());
+								iUser = FindUser(username.c_str());
+								if(iUser != -1)
 								{
-									std::string user = decoded.substr(0, npos);
-									std::string passwd = decoded.substr(npos + 1);
-									_log.Debug(DEBUG_AUTH, "OAuth2 Access Token: Found a Basic Auth Header for User (%s)", user.c_str());
-
-									iUser = FindUser(user.c_str());
-									if(iUser != -1)
+									if (GenerateMD5Hash(password).compare(m_users[iUser].Password) == 0)
 									{
-										if (GenerateMD5Hash(passwd).compare(m_users[iUser].Password) == 0)
+										Json::Value jwtpayload;
+										jwtpayload["preferred_username"] = m_users[iUser].Username;
+										jwtpayload["name"] = m_users[iUser].Username;
+										jwtpayload["roles"][0] = m_users[iUser].userrights;
+										std::string issuer = GetIssuerFromRequest(req, m_pWebEm->m_DigistRealm);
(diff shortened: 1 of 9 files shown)


dwmw2:
Btw, I think I want to move the Alexa endpoint into Domoticz itself, like HA does. Then the Lambda function is *only* a bare proxy _(with a grumpy comment about Smart Home skills not allowing a custom endpoint URI like normal skills do, because the Lambda function shouldn't even have to exist, and only adds latency!)_.

This will mean that Domoticz is in much better control over the way that devices are exposed according to their capabilities, rather than translating through the JSON API. And it also means that Domoticz will be able to do proactive notifications *to* Alexa. I've got that working in testing; we get given a refresh token for the push notifications when the account is linked (but what am I going to do with that from Lambda?).

dwmw2:
New commit on top to improve security by using a secret only known to Domoticz, to sign the actual authorization tokens. Otherwise, a client can use its client secret to forge a token for *any* Domoticz user, without actually logging in as that user.

dwmw2:
As requested, combined the SigningSecret and refresh token lifetime changes into a single database schema bump.

mealgi:
Hello,

I am currently using Google Home and Google Assistant to vocally control my domoticz devices.
The solution is based on the DZGA-Flask add on ( see https://github.com/DewGew/DZGA-Flask) which uses the oauth concept.
My question is does this PR have an impact on the DZGA-Flask add on ?

BR

dwmw2:
> Hello,
> 
>  My question is does this PR have an impact on the DZGA-Flask add on ?

No, it should not. Unless the DZGA-Flask add on was using the nonstandard password authentication having based its implementation only on Domoticz and not the actual standard. Everything should work fine.

We should merge https://github.com/domoticz/domoticz/pull/6424 as soon as possible though, as upgrading the database fails in the current development branch and I think it leaves the database with the DB_Version incremented but without the new fields.

gizmocuz:
PR 6424 is merged, thank you!

PR #6419, opened 2025-11-05
Tado: Made expire time of refresh token dynamic

Jan-nl:
Made the expire time of the refresh token of the Tado API dynamic, based on the expire time returned when a token is refreshed. Max loops is abs(expire time / poll interval) - 1, to be safe.

Remark: the first expire time is 599 seconds, the first refresh token returns 600 secs


gizmocuz:
Is one second not to risky? Maybe make it 30 seconds?
I don't see anywhere where you use a time_t to measure the actual time... Please remove the loop count thing and just store the time when you got the token, and then it is just a matter of subtracting it

if (acttime-token_get_time >= token_refresh_time) ...

Jan-nl:
It is not one second, the 599/600 is something Tado delivers. The actual time is at least the poll intervall.

Could be an issue if the poll intervall is set to 1 or so, but in that case the request limit is reached way too soon.

In case the poll intervall is greater compared to expire time, there also is a problem, but in that case the loops is set to one.

I know, not fool proof.

gizmocuz:
Sorry that is not what I meant. You receive 600 from Tado. Instead of checking the time and see if 600 seconds have elapsed, be a bit more safe, like 590 seconds (minus 10 seconds)

Jan-nl:
I made use of the loopcheck already present in the code. 

The time should be at least the pollinterval.

Or do you mean substract 10secs from the expire time Tado sends and compute the nr of loops based on that and the pollinterval?

Jan

gizmocuz:
Yes, and I suggest to not use the loop for time checking, 

	while (!IsStopRequested(1000))
	{

time_t atime=time(nullptr)
if (atime >= m_token_expire_time) {
  RefreshToken();
}

bool  RefreshToken()
{
-   get token
-  get token timeout
-  m_token_expire_time = time(nullptr) + (result of get token timeout) - 10

tado.h
private:
  time_t m_token_expire_time = 0

gizmocuz:
@Jan-nl , is it possible to implement the above feature request?

Jan-nl:
Yes, sure. But it will take some time. Due to my c++ knowledge and the actual time I have.

If you can cancel the pr for now that is fine.

gizmocuz:
No that's fine, take your time!

gizmocuz:
Is the "expries_in" in Seconds?

Could you make the m_iTokenExpiresIn  a time_t type in the header

and when it is in seconds you could do

m_iTokenExpiresIn = time(nullptr) + std::stoi(root["expires_in"].asString());

Is it a string or a number? Maybe you could directly to integer?

m_iTokenExpiresIn = time(nullptr) + root["expires_in"].asInt();

Delete the 'Set_TokenRefresh' function, this is not needed

m_iTADO_TOKEN_MAXLOOPS can also be deleted

In Do_Work above this line

```
	if (!(iSecCounter % m_iPollInterval == 0))
			continue;
```

do something like

if (time(nullptr)>=m_iTokenExpiresIn ) {
  //refresh the token (m_iTokenExpiresIn should get a new value in the feature, in case of an error, you can also set this to actualtime+30 seconds, to retry in 30 seconds in that function of course)
}

Maybe clean up some code, and move code from Do_Work in it's own function  (Own poll function)

If no token, no need to poll

Hope that makes a bit sense?

Jan-nl:
Fix should be complete now

gizmocuz:
Thankyou for the changes, looking good!

One minor change, this could be rewritten

if (
	    (!RefreshAccessToken())
           )
	{
	   //Problem with refreshing access token
	   return false;
	}

to

if (!RefreshAccessToken())
	   return false;

On two places

Jan-nl:
Updated and tested (shortly)

gizmocuz:
Thank you for the changes! I do see one more

if ( 
		(m_token_expire_time <= time(nullptr))
           )
	{

(double ( ) )

Jan-nl:
Updated

gizmocuz:
Thank you for this PR and all your time!

JurgenLB:
There is a compiler warning;
/home/runner/work/domoticz/domoticz/hardware/Tado.cpp: In member function ‘bool CTado::RefreshAccessToken()’:
/home/runner/work/domoticz/domoticz/hardware/Tado.cpp:214:32: warning: too many arguments for format [-Wformat-extra-args]
  214 |                 Log(LOG_ERROR, "Pollinterval > token expire time.", m_iPollInterval);

gizmocuz:
Thanks for reporting... fixed in latest code

PR #6421, opened 2025-11-07
Use IsLightOrSwitch() helper consistently

dwmw2:
There are a bunch of places in the code where we have a hard-coded list of pType/pSubType which are actually a switch, and for which SwitchType is STYPE_xxx (as opposed to being a meter type or something else).

Many of them seem to have *slightly* different criteria, which seems more by accident than design. Especially the one in the `clearlightlog` handler which checked for `pTypeThermostat4` twice, and not `pTypeThermostat3`. They mostly differ in their opinions about whether to include `pTypeRego6XXValue`/`sTypeRego6XXStatus`, and the EvoHome types.

This PR makes them all use IsLightOrSwitch() consistently, but *doesn't* actually make any functional changes except for the obvious pTypeThermostat3 typo. If some of the original code was wrong, we can fix the actual behaviour in subsequent commits but let's make the safe cleanup as a first step without functional changes.

### DIFF
--- a/main/WebServer.cpp
+++ b/main/WebServer.cpp
@@ -1733,31 +1733,12 @@ namespace http
 
 					bool bHasTimers = false;
 
-					if (
-						(dType == pTypeLighting1)
-						|| (dType == pTypeLighting2)
-						|| (dType == pTypeLighting3)
-						|| (dType == pTypeLighting4)
-						|| (dType == pTypeLighting5)
-						|| (dType == pTypeLighting6)
-						|| (dType == pTypeFan)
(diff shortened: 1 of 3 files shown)


gizmocuz:
Thank you for this PR! This cleans up the code quite a lot ! ;)

dwmw2:
I deliberately didn't make any actual changes here. We should look at the differences and see if there are bugs to be fixed. And in a related observation I believe I spotted that GetLightStatus doesn't handle pTypeLighting3 either?

PR #6423, opened 2025-11-10
Native Alexa Smart Home API support

dwmw2:
This adds a native Alexa Smart Home API endpoint to Domoticz, with support for the most common device types.

Providing a native API endpoint gives the most control over how devices appear in Alexa. It lets us implement specific features and utterances on the devices, and do proactive push notifications to Alexa (not yet implemented).

It does currently require the user to set up a Lambda function which *just* proxies the HTTP request to the Domoticz server, like the HA one (https://www.home-assistant.io/integrations/alexa.smart_home/). I've worked on simplifying and automating that process, and will probably pull in the docs and scripting that I've implemented for https://github.com/dwmw2/alexa_domo/blob/master/README.md 

My documentation says that you do need to open a port to Domoticz from the outside, although it *only* needs to do OAuth2 and not the normal API or authentication and perhaps we could have a mode which restricts all other access, to improve security.

And it should actually be possible to run the Lambda function in a VPC (private network) and have the Domoticz server [VPN in](https://docs.aws.amazon.com/vpc/latest/userguide/vpn-connections.html) to that same network, so there's no need for a public-facing port at all. Or we could provide the Lambda with a client SSL cert, and allow *only* connections with that cert from the outside, perhaps?

Since HA also exposes a native Alexa endpoint, I'm hoping that Alexa will eventually learn to discover those automatically and avoid the need for external access.

There are other ways we can expose devices to Alexa, even with automatic local discovery. But the simpler it gets, the more it's like playing "[game of telephone](https://en.wikipedia.org/wiki/Telephone_game)" with the device models. Some stuff can be exposed as a Philips Hue light bulb, or other types of device for which the manufacturer has created a skill with local discovery.

A decent compromise in the future for those who really don't want an external port (or to wait for local discovery to be added) might be to expose devices via Matter; the Alexa support for discovering those is improving. There's still a bit of 'game of telephone' as we translate Domoticz devices to Matter, and then we're at the mercy of Alexa's own interpretation of the Matter devices into its own model. We can add that later if we want, and this Alexa support can serve as a model for how we do so.

### DIFF
--- a/ALEXA.md
+++ b/ALEXA.md
@@ -1,0 +1,300 @@
+# Domoticz Alexa Smart Home Integration
+
+This integration allows you to control your Domoticz devices through Amazon Alexa using the Alexa Smart Home API v3.
+
+Domoticz includes native Alexa Smart Home API support and a built-in OAuth2 server that works with Alexa's account linking. A minimal AWS Lambda function acts as a passthrough, forwarding Alexa requests to your Domoticz instance.
+
+## Features
+
+- Control lights (on/off, dimming, color, color temperature)
+- Control scenes and groups
+- Control selector switches (mode controllers)
+- Monitor temperature and humidity sensors
+- Control thermostats
+- Control blinds and locks
+- Monitor various sensors (weight, pH, absolute humidity)
+
+## Requirements
+
+- **Domoticz home automation system** - Your Domoticz instance must be accessible from the Lambda function over Legacy IP, with a valid SSL certificate.
+- **Amazon Developer Account** - Sign up at [developer.amazon.com](https://developer.amazon.com) to create your Alexa Smart Home skill
+- **AWS Account** - Needed to host the Lambda function that bridges Alexa and Domoticz. AWS Lambda offers a free tier with 1 million requests per month, which is more than sufficient for typical home automation use.
+
+## Installation
+
+### Prepare Domoticz
+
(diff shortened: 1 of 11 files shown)


gizmocuz:
Thank you very much for this PR and time!

Could you undo the changes in the 'extern' folder?

You made two database patches, but only one is enough (171)?

If possible, functions for Alexa could be in Alexa.cpp like

void CWebServer::GetAlexaPage(WebEmSession& session, const request& req, reply& rep)

(needs the web namespace around it, for an example see MQTTAutoDiscovery.cpp, line 6174+)

All looks perfect! @kiddigital , do you agree with the OAuth changes?

dwmw2:
Argh, I *keep* reverting the extern changes; will fix them and also fix the Windows build (appveyor shows failures that didn't happen locally). I can happily squash the database changes, those are actually part of #6418 which I'd submitted separately, and one of them was the one I'd submitted even *more* separately because it's a security issue. Happy to squash them together; let's handle that in the other PR and I'll just rebase this one on top.

gizmocuz:
Thank you for the changes!

I don't think the changes in SQLHelper.cpp need to be in the PR as it's already being merged in another PR.
If you bring this file up2date with the latest version in the dev branch it should not change

gizmocuz:
Thank you for the PR and all the work you put into this!! (And the patience testing it with me 🤣)
Hope Amazon will support local discovery/skills soon

PR #6426, opened 2025-11-11
MQTT: Reinstate 'total_increasing' support for kWh sensors

### LINKED ISSUE #6160: MQTT "total_increasing" energy counter goes backwards.
I have a power meter device running ESPHome: https://devices.esphome.io/devices/Tongou-TO-Q-SY1-JWT

It reports its power usage (kWh sensor) as `"stat_cla": "total_increasing"`:
```
${mqtt_prefix}/homeassistant/sensor/power-lights-upstairs/upstairs_light_circuit_energy/config 
{
   "avty_t" : "power-lights-upstairs/status",
   "dev" : {
      "cns" : [
         [
            "mac",
            "18de50e14433"
         ]
      ],
      "ids" : "18de50e14433",
      "mdl" : "cbu",
      "mf" : "Beken",
      "name" : "power-lights-upstairs",
      "sw" : "2024.9.0-dev (Sep  9 2024, 14:39:50)"
   },
   "dev_cla" : "energy",
   "name" : "Upstairs light circuit Energy",
   "obj_id" : "power-lights-upstairs_upstairs_light_circuit_energy",
   "stat_cla" : "total_increasing",
   "stat_t" : "power-lights-upstairs/sensor/upstairs_light_circuit_energy/state",
   "uniq_id" : "18de50e14433-sensor-96f75e13",
   "unit_of_meas" : "kWh"
}
```
See documentation for `total_increasing` in https://developers.home-assistant.io/docs/core/entity/sensor/ :

> Similar to total, with the restriction that the state represents a monotonically increasing positive total which periodically restarts counting from 0, e.g. a daily amount of consumed gas, weekly water consumption or lifetime energy consumption. Statistics of the accumulated growth of the sensor's value since it was first added is updated every 5 minutes. A decreasing value is interpreted as the start of a new meter cycle or the replacement of the meter.

When Domoticz auto-discovers this device, it should probably set it up as an RFXMeter Counter device? It doesn't, and when the unit power cycles and goes back to zero, I see a massive negative energy usage.
Screenshot from 2024-09-10 22-51-52

dwmw2:
This was reverted due to a ZWave-JS-UI bug where it advertised sensors with 'total_increasing' even when they shouldn't be. So this commit reinstates it except *not* for ZWave devices.

A state class of "total_increasing" indicates "a monotonically increasing increasing positive total which periodically restarts counting from 0" … "A decreasing value is interpreted as the start of a new meter cycle or the replacement of the meter".

Handle this by storing an "epoch" for the measurement; this is the base value added to the raw measurement, to reach the actual total to be recorded.

If the result of adding a measurement to this epoch is ever *lower* than the previous recorded value, that means the counter must have restarted, so the epoch is increased to match the previous value and we keep counting from there.

The epoch is stored in "StrParam1" of the DeviceStatus record.

Clean up the matching of power and usage sensors to use the *actual* value of the paired sensor instead of its raw last_value. This keeps knowledge about one type of sensor away from the other type. So the code handling the Watt sensor no longer needs to care about different multipliers for kWh/Wh/Wm sensors, etc. — and, more to the point, doesn't need to learn about the epoch handling. It was already failing to throw away bogus (< -1000000) readings as the paired sensor should.

Fixes: #6160

### DIFF
--- a/hardware/MQTTAutoDiscover.cpp
+++ b/hardware/MQTTAutoDiscover.cpp
@@ -2055,34 +2055,27 @@ bool MQTTAutoDiscover::GuessSensorTypeValue(_tMQTTASensor* pSensor, uint8_t& dev
 			return false;
 		}
 
+		sValue = std_format("%.3f", fUsage);
+
 		float fkWh = 0.0F;
 		_tMQTTASensor* pkWhSensor = get_auto_discovery_sensor_unit(pSensor, "kwh");
-		if (pkWhSensor)
-			fkWh = static_cast<float>(atof(pkWhSensor->last_value.c_str())) * 1000.0F;
-		else
-		{
+		if (!pkWhSensor)
 			pkWhSensor = get_auto_discovery_sensor_unit(pSensor, "wh");
-			if (pkWhSensor)
-				fkWh = static_cast<float>(atof(pkWhSensor->last_value.c_str()));
-			else
-			{
-				pkWhSensor = get_auto_discovery_sensor_unit(pSensor, "wm");
-				if (pkWhSensor)
-					fkWh = static_cast<float>(atof(pkWhSensor->last_value.c_str())) / 60.0F;
-			}
-		}
+		if (!pkWhSensor)
+			pkWhSensor = get_auto_discovery_sensor_unit(pSensor, "wm");
+
 		if (pkWhSensor)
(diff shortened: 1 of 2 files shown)


gizmocuz:
Thank your for this PR and your time again!!

PR #6427, opened 2025-11-13
Add detection for deleting scene/device from devices page

rhuijben:
Every few years I trigger this same stupid issue, where I delete devices when I try to update scenes or vice versa.

Devices and scenes share index numbers, so code should be careful to find the right device type before operating on them. This fixes the devices screen.


gizmocuz:
Thank you for this PR!!

PR #6429, opened 2025-11-14
Fix CounterHelper race condition on simultaneous restart

dwmw2:
When both Domoticz and a meter device restart simultaneously, the m_nLastCounterValue was not persisted, causing rollover detection to fail. The meter would reset to 0 and the monotonic counter would incorrectly report the old offset value instead of continuing from the last reported total.

Fix by:
- Adding NodeID and ChildID parameters to Init()
- Querying DeviceStatus to restore m_nLastCounterValue from the device's last reported total value on restart
- This allows proper rollover detection even when both systems reset

Note: This fix is entirely untested.


gizmocuz:
Closed in favor of #6430

gizmocuz:
Closed in favor of #6430

PR #6430, opened 2025-11-16
Making CounterHelper ready to be used in more locations

gizmocuz:


hardware/MQTTAutoDiscover.cpp
```diff
@@ -2094,62 +2097,28 @@ bool MQTTAutoDiscover::GuessSensorTypeValue(_tMQTTASensor* pSensor, uint8_t& dev
 		else if (szUnit == "wm")
 			multiply = 1.0 / 60.0;
 
-		double dkWh = atof(pSensor->last_value.c_str()) * multiply;
+		double dkWh = (!pSensor->last_value.empty()) ? (atof(pSensor->last_value.c_str()) * multiply) : pSensor->prev_value;
```
dwmw2: Don't we have to subtract the counterhelper offset, if we're using pSensor->prev_value here? The dkWh value is treated as a raw input from the device, which gets the offset added to it (again).

dwmw2:
This seems to go wrong when the offset gets too high?
```
2025-11-18 17:40:50.570  Error: mangle-ad: Exception (on_message): stoi! (topic: power-ashp3/sensor/heat_pump_3_circuit_energy/state, message: 1)
```
One test I tried was just sending alternate values 1, 15102 for the energy topic. Which should be seen as multiple resets, jumping by ~15MWh each time. It only works if I repeat the submissions, then breaks with the stoi exception above. 

The existing code in the development branch works, if I just fix the zwave check to check for std::string::npos

hardware/MQTT.cpp
```diff
@@ -672,72 +682,39 @@ bool MQTT::ConnectIntEx()
...
 	bool bFirstTime = true;
-	int msec_counter = 0;
 	int sec_counter = 0;
 
 	set_callbacks();
 
-	while (!IsStopRequested(100))
+	time_t last_time = time(nullptr);
+
+	while (!IsStopRequested(1000))
```
dwmw2, line 692: Let's not do this as part of this PR; it doesn't belong here. But this still isn't right. I have a proper fix in #6438 (will use CI to make sure the Windows version works).

The problem with the MQTT thread was that it spends a lot of its time not actually *working*. Most of its lifetime is spend waiting to die. Each time around the loop, it pauses for a whole second, ignoring any work that comes in, waiting in IsStopRequested(). Only then does it proceed to do its job, before going back round again and waiting to die for another second.

A worker thread should sit in select() on the file descriptors on which its work arrives (serial ports, sockets, etc.), and it can wake up the moment a packet is received and process it without delay. In #6438 I make the stop signal arrive on a socket that the threads can include in their select(), so they can do the right thing and just wait for *either* work, *or* a stop signal to arrive (or a timeout, if they have something to do at a certain time in the future).

With that, #6433 is solved and I don't think we even need the added complexity of the queue.

PR #6431, opened 2025-11-16
Fix MQTT AutoDiscover device vs. sensor name confusion

### LINKED ISSUE #6164: MQTT Temp/Hum devices choose strange name.

dwmw2:
During MQTT AutoDiscovery, we were overwriting the name of the parent MQTT device with the name of each sensor we discovered that was attached to it.

Each time we discovered a sensor, we get the device information again in that sensor's JSON information, and construct a name using what we find in the JSON, e.g. "device (sensor_name)".

But then we put that sensor-specific name into pDevice->name which we use to represent the *device*, not just that sensor.

This mostly didn't matter, as pDevice->name is *mostly* a write-only string, because sensor naming just uses the device name from the JSON each time.

The only case where we care about pDevice->name is when we build up combined Temperature/Humidity/Barometer devices from multiple MQTT sensors. In *that* case, we were constructing the name of the combined Temp+Hum+Baro device from pDevice->name, which contained the name of the last *sensor* that we happened to find attached to that device!

Which explains why we ended up with Temp+Hum+Baro devices with weird names like "bathroom (Blue LED)".

Clean up the device naming, use a local variable for sensor_name instead of messing with pDevice->name, and make it a bit easier to follow and better documented.

Fixes: #6164

### DIFF
--- a/hardware/MQTTAutoDiscover.cpp
+++ b/hardware/MQTTAutoDiscover.cpp
@@ -824,6 +824,7 @@ void MQTTAutoDiscover::on_auto_discovery_message(const struct mosquitto_message*
 
 		pDevice->identifiers = device_identifiers;
 
+		// This is the name of the *device* to which this sensor happens to be attached
 		std::string dev_name("");
 		if (!root["device"]["name"].empty())
 		{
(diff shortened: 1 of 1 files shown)


gizmocuz:
Super thanks for this PR! You've done it!!

GravityRZ:
does this mean the complete name of the device is now included when discovering and creating a device

i discovered the following issue.
MQTT AD is only using the name ESPsensor.... or ESPlight...  instead of the complete path

e.g when you rename a device it still works(which gives you problems if you have multiple sensors with the same name but in units with a different name

eg unit called esp8266 sends out this message

homeassistant/sensor/esp8266-co2/mh-z19_co2_value
esp8266-co2/sensor/mh-z19_co2_value

now when i move that sensor into a complete new unit called esp32-co2 i expected the devices to be recreated but they are not

this is send out
homeassistant/sensor/esp32-co2/mh-z19_co2_value
esp32-co2/sensor/mh-z19_co2_value

but the same previous created ESPsensormh-z19_co2_value is being used.

so this means if you use the same naming convention in all of your esp's( which is preferred) you get into trouble if you have more devices with the same sensor name/light

PR #6432, opened 2025-11-16
Fix setpoint-only thermostats in Alexa

dwmw2:
The permissions check required that *all* of the main endpoint, the setpoint, and the mode (if present) are available.

But for a setpoint without an associated thermostat, the main endpoint index *is* the setpoint index, which means it'll only be return once. Fix the permissions check to understand that, and only expect as many results as it had *unique* indices.


gizmocuz:
Thank you!

PR #6435, opened 2025-11-18
Add data-i18n attribute for localization in links

Tixav:


gizmocuz:
This should be translated by

{{:: 'Log' | translate }}

gizmocuz:
Translation is also working

image

Maybe the language that you are using is not 100% translated? Which language is this?

Tixav:
It's primarily a way to have a nearly unique CSS selector (also using the parent element, of course).

gizmocuz:
But now they are double translated? via data-i18m and the | translate method

`<a class="btnsmall" data-i18n="Log" ng-href="#/Devices/{{ ::item.idx }}/Log">{{:: 'Log' | translate }}</a>`

Tixav:
I revert to the defaut theme and there are not double translation for both "data-i18n" additions
image
image

Tixav:
I have just tested in Portuguese (I am French), that is working well too
image
image

gizmocuz:
Sorry, that is not what I meant. It is already translated with the | translate option
Because you now add data-i18n as well, it is double. It should work with the original code as well.

See the image below:

image

Tixav:
I understood, but now I don't know what to do. Because everywhere in the other html page, options buttons can be identify by data-i18n by css
I am develloping a theme for domoticz and just these 2 options buttons of temperature page prevent to do a generic css code. 
https://forum.domoticz.com/viewtopic.php?p=329622#p329622 and https://github.com/Tixav/LittleTheme
Another method would be adding an Id for each options buttons in all html page...
Your point of view ?

gizmocuz:
change:

`<a class="btnsmall" ng-href="#/Devices/{{ ::item.idx }}/Log">{{:: 'Log' | translate }}</a>`

to

`<a class="btnsmall" data-i18n="Log" ng-href="#/Devices/{{ ::item.idx }}/Log"></a>`

gizmocuz:
And you have this only in the temperature tab? Not in Lights/Utility/Weather etc?

I made a new branch (i18n), feel free to contribute to this

https://github.com/domoticz/domoticz/tree/i18n-tweaks

I already made a PR with some changes

https://github.com/domoticz/domoticz/pull/6457

Take care about the double points for labels : : :

Tixav:
whaoo, You made all these modifications even in dialogbox ? yes I confirm I saw this only in the temperature_widget.html all the other "|translate" are in "dialogbox" like timer.html, device_light_edit.html, notification.html and scenes.html (from I can see in the www\views: grep -rn translate)

gizmocuz:
Yes, but more things can be rewritten.... Going to do this in a larger PR... so one more day patiance please

gizmocuz:
It's merged! Hope your theme can work with the recent changes/beta

PR #6436, opened 2025-11-18
Implement delay for highlight effect on update Temperature

Tixav:
Add a delay before highlighting updated temperature widget. Must be delay in another way effect is finished before angular finished to draw the widget

### DIFF
--- a/www/app/TemperatureController.js
+++ b/www/app/TemperatureController.js
@@ -102,7 +102,10 @@ define(['app', 'livesocket'], function (app) {
 					oldarray[oldindex] = item;
 					if (!document.hidden) {
 						if ($scope.config.ShowUpdatedEffect == true) {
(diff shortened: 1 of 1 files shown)


gizmocuz:
Is this with the default Domoticz theme? I did not notice any issue since the last years.
If this is an issue, it is also on all other tabs (Dashboard, Switches, Utility and Weather)

Tixav:
Yes with the default theme also. I saw this problem is only on temperature and weather widget in their own tab (that works on dashboard and stwiches and utility tabs). I think the problem is due to .directive only use for these 2 tabs.(line 508 .directive('dztemperaturewidget', function ($rootScope,$location) )
.directive in these tab refresh totaly each widget and should take more time so the variable #cssvar of the line #cssvar.effect does not exist or not reachable because not yet created. (jquery var.length = 0)
Si I also make a PR for weather

gizmocuz:
Thank you very much!

PR #6437, opened 2025-11-19
Fix total_increasing check for zwave

dwmw2:
Oops, need to check for std::string::npos not zero.


gizmocuz:
Thank you for this fix! Aldo the unique ID probably starts with zwavejs2mqtt, it is handled in #6430

Issue #6433, opened 2025-11-18
MQTT-AD scaling issues


dwmw2:
I have about 430 sensors reporting to MQTT-AD, maybe or 10 updates a second in the steady state.

Domoticz doesn't keep up. When I click to switch a light, I can see the MQTT command message sent immediately, and with a separate mosquitto_sub I can see the state change immediately, but Domoticz takes a very long time (perhaps even minutes) to show the state change.

Running `netstat -tn` shows a receive backlog because it isn't even reading the data from the socket after it comes in from the MQTT server:
```
tcp6   57466      0 2001:8b0:10c:1::213:49112 2001:8b0:10c:1:843:1883 ESTABLISHED
```

Sometimes Domoticz reconnects with a keepalive failure:
```
Error: mangle-ad: Disconnected, restarting (rc=19/Keepalive exceeded)
```

I ran `strace` on the MQTT-AD thread to try to see what it's doing...

It's only managing to process messages every few hundred milliseconds:
```
15:16:14.264364 write(1, "2025-11-17 15:16:14.264  mangle-ad: General/Custom Sensor (faikin3 (ESP Uptime))\n", 81) = 81
15:16:14.592103 write(1, "2025-11-17 15:16:14.592  mangle-ad: General/Custom Sensor (power-ashp2 (Heat pump 2 circuit Frequency))\n", 104) = 104
15:16:14.803208 write(1, "2025-11-17 15:16:14.803  mangle-ad: General/Custom Sensor (softener (HX711 value))\n", 83) = 83
15:16:15.143804 write(1, "2025-11-17 15:16:15.143  mangle-ad: General/Custom Sensor (faikin4 (Compressor Frequency))\n", 91) = 91
15:16:15.250609 write(1, "2025-11-17 15:16:15.250  mangle-ad: General/Custom Sensor (landing-blind-2 (Heap Free))\n", 88) = 88
```

Other times, it looks like it's actually sending notififications to web clients? Shouldn't the MQTT-AD thread just be submitting into the mainworker queue rather than doing that for itself?
```
[...]

PR #6438, opened 2025-11-19
Fix StoppableTask to allow for proper select() based event loops, fix MQTT

dwmw2:
Worker threads need to wake up quickly and respond to three event types:
 - When work comes in on a socket or serial port
 - When they're asked to stop
 - When a certain time is reached (for periodic events, heartbeats, etc.)

Domoticz currently doesn't provide a way for threads to act responsively on all three. Mostly we use `StoppableTask::IsStopRequested()` which will wait for a period of time and return immediately if the thread is asked to stop _(which isn't the common case we should be optimising for)_. And then we process the actual work of the thread mostly as an afterthought, often up to a second after it actually arrived.

The alternative approach, like in `MainWorker::Do_Work_On_Rx_Messages()`, is to prioritise the *work* instead. It mostly spends its time sitting in `m_rxMessageQueue.timed_wait_and_pop()`, waking every five seconds just to check if it should stop with an `IsStopRequested(0)` call.

The current `StoppableTask` implementation means that either the work, or the stop check, is done as an 'afterthought' and not handled promptly. And it means that threads are waking up often when they don't need to, as well as not waking up as fast as they should when they *do*.

This change to `StoppableTask` makes it possible for a thread to react quickly to all three. It changes the implementation to use a socketpair which means that we can add a `GetStopFd()` method, and then a worker thread can use a standard `select()` loop waiting on that fd as well as its file descriptors for real I/O, with whatever timeout it desires.

Then we can fix MQTT to use a proper select-based loop, so #6433 is fixed and MQTT updates are seen fairly much instantly without needing the hack of an extra thread (although I've left that fix out of this PR).

The EventSystem and Scheduler are OK with the existing `IsStopRequested()` interface because they don't have separate file descriptors to poll on, but they were waking up far more often than they needed to. EventSystem was waking twice a second just to see if it had reached the time that it needed to do something. So we give it a hint: if it doesn't have anything to do for 12 seconds, then it can sleep for 12 seconds in `IsStopRequested()`. Aside from fewer wakeups, this actually improves the precision of when the actions happen anyway; they're now within a millisecond of the top of the minute, instead of up to half a second thereafter.

Scheduler was similar although for now I've left it waking up every second because although most scheduler entries only have minute precision, I can't prove to myself that there's no way to schedule something at a precise second _(e.g. sunset/sunrise based?)_ But the cleanup does improve the precision still, so events are run at the start of the second, not at some random point within each second.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -176,6 +176,7 @@ main/RFXNames.cpp
 main/Scheduler.cpp
 main/SignalHandler.cpp
 main/SQLHelper.cpp
+main/StoppableTask.cpp
 main/SunRiseSet.cpp
 main/TrendCalculator.cpp
 main/WebServer.cpp
--- a/main/EventSystem.cpp
+++ b/main/EventSystem.cpp
@@ -316,32 +316,34 @@ void CEventSystem::Do_Work()
 	m_python_Dir = szUserDataFolder + "scripts/python/";
 #endif
 #endif
-	time_t lasttime = mytime(nullptr);
-	struct tm tmptime;
+	time_t atime = mytime(nullptr);
 	struct tm ltime;
 
-	localtime_r(&lasttime, &tmptime);
-	int _LastMinute = tmptime.tm_min;
+	localtime_r(&atime, &ltime);
+	int _LastMinute = ltime.tm_min;
 
 	_log.Log(LOG_STATUS, "EventSystem: Started");
-	while (!IsStopRequested(500))
+	while (true)
 	{
-		time_t atime = mytime(nullptr);
+		// Calculate sleep time until next 12-second heartbeat
+		// Handle leap seconds (tm_sec can be 60 or 61) by clamping to 59
+		int current_sec = (ltime.tm_sec > 59) ? 59 : ltime.tm_sec;
+		int next_wake_sec = ((current_sec / 12) + 1) * 12;
+		int sleep_ms = (next_wake_sec - current_sec) * 1000;
 
-		if (atime != lasttime)
+		if (IsStopRequested(sleep_ms))
+			break;
+
+		atime = mytime(nullptr);
+		localtime_r(&atime, &ltime);
+
+		if (ltime.tm_sec % 12 == 0) {
+			m_mainworker.HeartbeatUpdate("EventSystem");
+		}
+		if (ltime.tm_min != _LastMinute)
 		{
-			lasttime = atime;
-
-			localtime_r(&atime, &ltime);
-
-			if (ltime.tm_sec % 12 == 0) {
-				m_mainworker.HeartbeatUpdate("EventSystem");
-			}
-			if (ltime.tm_min != _LastMinute)
-			{
-				_LastMinute = ltime.tm_min;
-				ProcessMinute();
-			}
+			_LastMinute = ltime.tm_min;
+			ProcessMinute();
 		}
 	}
 	_log.Log(LOG_STATUS, "EventSystem: Stopped...");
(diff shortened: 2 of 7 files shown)


gizmocuz:
Thank you for this PR... Could you undo the changes to MQTT as they conflict with my changes in #6430 ...

dwmw2:
I suspect the MQTT thread changes probably didn't live in a CounterHelper PR? I also don't actually think the extra queue is necessary at all if we fix the worker thread to actually wake up when it has work to do :)

I'll keep it here for my testing for now, and I can rebase on top if you do merge the other changes, OK?

dwmw2:
Moving some comments from Slack so they don't get lost..

> Would it be a good idea to move (later) everything to the header, or move all code to the .cpp? (like constructor)

Happy either way. I moved it all out to a .cpp file so I could play with the implementation without having to rebuild every C++ file in the project, but once I'm done playing with it that's no longer needed. 

> You do not need all those if/defs, unix should/could also use closesocket/send/recv
> Also please include stdafx.h like in all .cpp files, then there should be no need for all these includes as most of them are already in that header file

Not closesocket unless we have a wrapper? I'll take a look. Mostly I threw that in as known good code and went to concentrate on fixing MQTT, so I'll look at making it more "Domoticzy". And I'll use the CI here to check the Windows build.

dwmw2:
Reworked and cleaned up the Windows code in StoppableTask.cpp. Now it *only* uses szUserDataFolder as the directory for its AF_UNIX sockets. That's going to be writeable. It doesn't leave anything behind because it deletes the listening socket as soon as it's created the connected socketpair that it needs. And it keeps the fallback to AF_INET so *if* you run Domoticz so it can't write to use szUserDataFolder... well, it probably won't get far but at least *this* code will cope :)

Test case which shows the difference in speed of worker threads actually waking up for *work*: https://david.woodhou.se/MagWeg2-work.zip

Compile with and without `-DUSE_GETSTOPFD` to see the difference.
```
 ● I will run the following shell command: 
 ./MagWeg2
 ⋮ 
 ↳ Purpose: Run test with USE_GETSTOPFD

Sending ping 1...
Entered thread function
Pong received, latency: 126 µs
Sending ping 2...
Pong received, latency: 132 µs
Sending ping 3...
Pong received, latency: 76 µs
Sending ping 4...
Pong received, latency: 173 µs
Sending ping 5...
Pong received, latency: 152 µs
main app stopping thread....
Stop requested
Leaving thread function
main app thread stopped....
 ⋮ 
 ● Completed in 3.513s





🛠️  Using tool: execute_bash (trusted)
 ⋮ 
 ● I will run the following shell command: 
 ./MagWeg2-poll
 ⋮ 
 ↳ Purpose: Run test without USE_GETSTOPFD

Sending ping 1...
Entered thread function
Pong received, latency: 1001233 µs
Sending ping 2...
Pong received, latency: 299991 µs
Sending ping 3...
Pong received, latency: 300785 µs
Sending ping 4...
Pong received, latency: 301022 µs
Sending ping 5...
Pong received, latency: 300776 µs
main app stopping thread....
Leaving thread function
main app thread stopped....
 ⋮ 
 ● Completed in 5.719s


> ## Test Results Summary

With USE_GETSTOPFD (event-driven with select()):
• Ping 1: 126 µs
• Ping 2: 132 µs
• Ping 3: 76 µs
• Ping 4: 173 µs
• Ping 5: 152 µs
• Average: ~132 µs (0.132 milliseconds)

Without USE_GETSTOPFD (polling with 1-second IsStopRequested()):
• Ping 1: 1,001,233 µs (1.001 seconds)
• Ping 2: 299,991 µs (0.300 seconds)
• Ping 3: 300,785 µs (0.301 seconds)
• Ping 4: 301,022 µs (0.301 seconds)
• Ping 5: 300,776 µs (0.301 seconds)
• Average: ~441 milliseconds

Performance improvement: ~3,300x faster with the event-driven approach using GetStopFd(). The polling version must wait up to 1 second in IsStopRequested() before checking for
work, while the event-driven version responds immediately via select().

```

gizmocuz:
Thank you for this PR, time and patience 😁

PR #6439, opened 2025-11-19
Introduce delay for highlight effect on weather widgets

Tixav:
Add a delay before highlighting updated weather widget names. otherwise the effect finished before the object creation by angular

### DIFF
--- a/www/app/WeatherController.js
+++ b/www/app/WeatherController.js
@@ -80,7 +80,10 @@ define(['app', 'livesocket'], function (app) {
 					oldarray[oldindex] = item;
 					if (!document.hidden) {
 						if ($scope.config.ShowUpdatedEffect == true) {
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thank you for this PR. Is this with the default theme?

Tixav:
commented
3 days ago
• 
Yes with the default theme also. I saw this problem is only on temperature and weather widget in their own tab (that works on dashboard and stwiches and utility tabs). I think the problem is due to .directive only use for these 2 tabs.(line 508 .directive('dztemperaturewidget', function ($rootScope,$location) )
.directive in these tab refresh totaly each widget and should take more time so the variable #cssvar of the line #cssvar.effect does not exist or not reachable because not yet created. (jquery var.length = 0)
Si I also make a PR for temperature

gizmocuz:
Thank you very much!

PR #6442, opened 2025-11-21
Clean up MQTT to remove unneeded extra thread, reduce CPU and memory usage

dwmw2:
A previous commit (hidden in with unrelated CounterHelper changes) changed MQTT to use *two* threads, to work around the problem that our worker threads are always obsessed with their own death: they spend their entire time just waiting to die in `IsStopRequested()` even while there is actual work for them to do.

Only when their `IsStopRequested()` call times out without them being asked to die, do they do some work just kind of as an afterthought. This is why MQTT was so slow (#6433) — because the thread was spending 100ms at a time waiting to see if it should die, while it had a backlog of message to process!

The previous commit worked around this by introducing yet another thread, which runs the `mosquitto_loop_forever()` and actually *does* wake up promptly when it has work to do. So the original `MQTT::Do_Work()` was now free to obsess only about death, as its primary job now is *just* to react to the stop signal by calling `mosquitto_disconnect()`.

This worked because mosquitto's own loop function *does* have a proper worker thread model which can wake up promptly on *either* work *or* the stop/disconnect request.

Now that our `StoppableThread` implementation has been fixed to allow threads to wait for both work *and* stop signal efficiently, there's no need for spawning a separate thread just to work around its limitations. Remove that, and use our own worker thread for its intended purpose with the improved `StoppableThread` implementation.

### DIFF
--- a/hardware/MQTT.cpp
+++ b/hardware/MQTT.cpp
@@ -678,8 +678,9 @@ void MQTT::Do_Work()
 
 	set_callbacks();
 
-	while (!IsStopRequested(0))
+	while (true)
 	{
+		ClearSelectFds();
 		time_t now = mytime(nullptr);
 
 		// Handle periodic tasks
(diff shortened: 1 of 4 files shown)


gizmocuz:

Aldo this probably works, it is a little bit hard to read/use/messy.

Is it somehow possible to make a (helper)structure/class inside the StopableTask class that can deal with all this?

```
		// Setup select with both mosquitto socket and stop fd
		fd_set rfds, wfds;
		FD_ZERO(&rfds);
		FD_ZERO(&wfds);

		if (mosq_fd >= 0)
		{
			FD_SET(mosq_fd, &rfds);
			if (want_write())
				FD_SET(mosq_fd, &wfds);
			if (mosq_fd > maxfd)
				maxfd = mosq_fd;
		}

		// Wait up to 1 second for activity
		struct timeval tv;
		tv.tv_sec = 1;
		tv.tv_usec = 0;

		int ret = select(maxfd + 1, &rfds, &wfds, nullptr, &tv);
		
		if (FD_ISSET(stop_fd, &rfds))
				break;

		last_ping = now;
				
```

Maybe that could cleanup the code and it made it more easier to use in other hardware classes?

dwmw2:
It's just a standard select loop. This is normal POSIX event handling that's been around for decades. You can wrap little helpers around it to hide the FD_SET and automatically increment `nfds` maybe... but it's mostly more hassle than it's worth to do so as it hides something that's familiar for decades, and turns it into something that the user has to look up and understand.

dwmw2:
Maybe something simple like
  - `clear_select_fds()` would do so, zero `nfds` and reset timeout.
  - `set_select_fd(int fd, bool wantRead, bool wantWrite, bool wantExcept)`
  - `select_timeout(...)` converts to timeval and ensures the timeout is set no higher than its argument?
 
Reduces the typing of the standard method but doesn't obfuscate it?

gizmocuz:
Looking already way better ;) thanks again!

PR #6444, opened 2025-11-22
Add combined Temperature/Setpoint for thermostats

dwmw2:
We have lovely combined devices for Temp/Hum/Baro, and even for Evohome zones we have a combined setpoint/temp with a lovely graph that shows them together. Let's have the setpoint support for non-Evohome too.

This adds a pTypeThermostat6 type, with subtypes for the different hum/baro combinations. Uses the existing graph support for those in the UI and just has to enable it for the appropriate types, as well as the setpoint control.

Hook it up to Alexa, allow users to make them as virtual sensors, make MQTT 'climate' devices register a single combined Domoticz device where they can instead of the separate setpoint/temp/etc, and we have nice shiny thermostat support...

Screenshot from 2025-11-22 01-34-44
Screenshot from 2025-11-22 01-36-35

### DIFF
--- a/hardware/MQTTAutoDiscover.cpp
+++ b/hardware/MQTTAutoDiscover.cpp
@@ -1519,6 +1519,14 @@ void MQTTAutoDiscover::on_auto_discovery_message(const struct mosquitto_message*
 			pSensor->current_temperature_template = root["current_temperature_template"].asString();
 		if (!root["curr_temp_tpl"].empty())
 			pSensor->current_temperature_template = root["curr_temp_tpl"].asString();
+		if (!root["current_humidity_topic"].empty())
+			pSensor->current_humidity_topic = root["current_humidity_topic"].asString();
+		if (!root["curr_hum_t"].empty())
+			pSensor->current_humidity_topic = root["curr_hum_t"].asString();
+		if (!root["current_humidity_template"].empty())
+			pSensor->current_humidity_template = root["current_humidity_template"].asString();
+		if (!root["curr_hum_tpl"].empty())
+			pSensor->current_humidity_template = root["curr_hum_tpl"].asString();
 		if (!root["preset_modes"].empty())
 		{
 			for (const auto& ittMode : root["preset_modes"])
@@ -3481,6 +3489,24 @@ void MQTTAutoDiscover::handle_auto_discovery_climate(_tMQTTASensor* pSensor, con
 		bIsJSON = root.isObject();
 	}
 
+	// Check for legacy devices once on first message
+	if (pSensor->devType == 0)
+	{
+		// If both topics exist, check database for old devices
+		if (!pSensor->temperature_command_topic.empty() && !pSensor->current_temperature_topic.empty())
+		{
+			std::vector<std::vector<std::string>> old_dev;
+			old_dev = m_sql.safe_query("SELECT COUNT(*) FROM DeviceStatus WHERE (HardwareID==%d) AND (DeviceID=='%q') AND (Unit==%d) AND (Type==%d OR Type==%d)",
+				m_HwdID, pSensor->unique_id.c_str(), CLIMATE_TEMP_SETPOINT_UNIT, pTypeTEMP, pTypeSetpoint);
+			pSensor->bUseLegacyClimate = !old_dev.empty() && atoi(old_dev[0][0].c_str()) > 0;
+		}
+		else
+		{
+			// Without both topics, always use legacy standalone handlers
+			pSensor->bUseLegacyClimate = true;
+		}
+	}
+
 	// Create/update Selector device for config and update payloads 
 	bool bValid = true;
 	bool isNull = false;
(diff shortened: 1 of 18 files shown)


dwmw2:
PXL_20251122_103959231

JurgenLB:
There is now a warning from `'=': conversion from 'double' to 'float', possible loss of data ` in main\mainworker.cpp 8410 & 8412

Great work for the rest.
Kind regards

dwmw2:
Thanks, will fix. Also going to make MQTT keep feeding the old devices if they exist rather than always creating the combined one. Users might be using those; they can delete them and rediscover to convert if they *want* to.

hardware/MQTTAutoDiscover.cpp
```diff
@@ -3467,6 +3475,24 @@ void MQTTAutoDiscover::handle_auto_discovery_climate(_tMQTTASensor* pSensor, con
...
 	}
 
+	// Check for legacy devices once on first message
+	if (pSensor->devType == 0)
+	{
+		// If both topics exist, check database for old devices
+		if (!pSensor->temperature_command_topic.empty() && !pSensor->current_temperature_topic.empty())
+		{
+			std::vector<std::vector<std::string>> old_dev;
+			old_dev = m_sql.safe_query("SELECT COUNT(*) FROM DeviceStatus WHERE (HardwareID==%d) AND (DeviceID=='%q') AND (Unit==%d) AND (Type==%d OR Type==%d)",
```
gizmocuz: please use 'auto old_dev = m_sql....' and drop the above vector declaration.

hardware/MQTTAutoDiscover.cpp
```diff
@@ -3467,6 +3475,24 @@ void MQTTAutoDiscover::handle_auto_discovery_climate(_tMQTTASensor* pSensor, con
...
+				m_HwdID, pSensor->unique_id.c_str(), CLIMATE_TEMP_SETPOINT_UNIT, pTypeTEMP, pTypeSetpoint);
+			pSensor->bUseLegacyClimate = !old_dev.empty() && atoi(old_dev[0][0].c_str()) > 0;
```
gizmocuz: Does this mean that when a user already had 'any' old thermostats, it will never get to enjoy the new type?

gizmocuz:
Thanks once again for your time and PR!!

waltervl:
@dwmw2 How to use this as a virtual device?
What is the API call to update the device (setpoint and/or temperature)?
What is the python framework way to create and update this device?
Why is the device only visible as temperature device in the temperature tab and not in the Utility tab as any other setpoint device?
Edit: How can you edit from gui the setpoint device settings (min,max,uom, step)?
Edit2: why does the widget ribbon change color? It normally indicates a non working device or low battery (red/yellow). If it is a temperature range indication, the icon itself changes accordingly.

dwmw2:
It's creatable like any other virtual devices through the UI.
Screenshot_20251203-191651

The `setsetpoint` JSON call works as you'd expect, as does `udevice` with semi-colon separated fields in the sValue like many devices. It's mostly modelled on the behaviour of the Evohome devices. Showing colour-coded for temperature, and the way you interact with it to set the setpoint, are deliberately set up to be the same as the combined Evohome setpoint+temp devices. Including the fact that they are temperature devices.

It's possible that we are missing the ability to edit those options; I'll look into that. Thanks for pointing it out.

dwmw2:
For python example see https://github.com/tuk90/RemehaHome-Domoticz/issues/40

waltervl:
Thanks for the references.
What would be the API sValue format for this device? Just the same as for a normal temp or temp/hum or temp/hum/baro device?

And I do not have Evohome, so what does these ribbon colors mean? They are not documented in the evohome wiki as far as I can see.

waltervl:
And the setpoint edit function is probably not working because the device is not in the Utility tab. So either it should also be visible there or copy the Edit setpoint javascript function to the temperature view.

dwmw2:
The sValue is something like `setpoint;temp;hum;hum_status;baro` for fields that actually exist (for the subtype). In `udevice` you can leave some empty if you don't want to update those (so you don't have to update the setpoint when you're updating the actual *measurements*, etc.

dwmw2:
For the editing part, please test https://github.com/domoticz/domoticz/pull/6477

Setting the *Unit* is hosed, because it behaves precisely like the Setpoint, which are also hosed. I don't even know how this is supposed to work. If we set the unit does that mean the incoming numbers for that device *are* in that unit, and should then be converted to the user's preference for display? Or does it mean not change the actual temperature, but only change how it's *displayed* to the user? A wild stab in the dark at https://github.com/dwmw2/domoticz/commit/b4ef5bcb9883525a1b9bdb3a90eec9b9fc439a35 but without knowing what it's *supposed* to do it's hard to fix that part.

#6477 at least makes it behave the same as the existing standalone setpoints though, for any sane person who lives in the 21st century and would never use a unit *other* than °C for any of this stuff :)

Happy to take a closer look if there's a clear explanation of what it should be doing.

PR #6453, opened 2025-11-24
Fix flash of unfiltered devices on live updates

### LINKED ISSUE #6447: Discovered MQTT sensors flash on filtered devices tab
When I got to e.g. the Temperature tab and filter for something like 'kitchen', it shows me only the appropriate devices.

However, when MQTT updates come in for other Temperature devices, those seem to flash up in my filtered list for a moment and then disappear. Here's a screenshot captured at the moment one of them appears, before it goes again...

Image

dwmw2:
When devices are updated via websocket (e.g., MQTT autodiscovery), they would briefly flash on screen even when a filter was active, before the filter logic would hide them again.

Fixed by checking the active filter in RefreshItem() and returning early without updating items that don't match. This prevents Angular from re-rendering non-matching items, eliminating the flash.

Fixes #6447

### DIFF
--- a/www/app/LightsController.js
+++ b/www/app/LightsController.js
@@ -314,6 +314,21 @@ define(['app', 'livesocket'], function (app) {
 				return; //idx not found
 			}
 
+			var bigtext = TranslateStatusShort(item.Status);
+			if (item.UsedByCamera == true) {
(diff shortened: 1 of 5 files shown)


gizmocuz:

PR #6454, opened 2025-11-24
Fix db update when updating options for custom device with Python plugin Ex Framework

guillaumezin:
Prevent the update to be propagated to every devices managed by a plugin


gizmocuz:
Thank you! Next time please compile/test 👍

guillaumezin:
> Thank you! Next time please compile/test 👍

My bad, I did tests and saw the compiling problem but forget to push the fix immediately afterwards.

PR #6455, opened 2025-11-25
Revert "Fix flash of unfiltered devices on live updates"

gizmocuz:
Reverts domoticz/domoticz#6453


dwmw2:
Why? Is there a problem? Fairly sure that it isn't causing https://github.com/domoticz/domoticz/issues/6447#issuecomment-3571380833 and anyway, I can't even *tell* if I have that issue while the devices are flashing up and disappearing all the time.

PR #6456, opened 2025-11-25
Revert "Clean up MQTT to remove unneeded extra thread, reduce CPU and memory usage"

gizmocuz:
Reverts domoticz/domoticz#6442


gizmocuz:
Going back to just two threads and see if this solves #6458

PR #6459, opened 2025-11-25
Adding extra validations since Hue Bridge firmware 1.74.1974063020 crashes Domoticz in certain cases

gbonny1982:
I've added some extra validations to Philips Hue Bridge API v1 code since several Philips Hue v2 Bridges are experiencing issues with the new firmware 1.74.1974063020. 

The root cause is unclear at this moment as I cannot reproduce it. However I hope and suspect this commit will eliminate Domoticz crashes or at least it could give some more information regarding the issue.

I've also changed the Hue v2 API LOG_ERROR messages into LOG_DEBUG_INT as these are not real errors, which could be confusing.


gizmocuz:
Thank you very much! PR is merged in beta 16940

PR #6462, opened 2025-11-27
Update scripts changes to use up to date version.

lym70:
Rework update to, after uploading new domoticz archive, 1st extract archive up to date upgrade[beta|release] script and go ahead with it.
This allows better management of update restrictions that, for now, screw users when a compatibility change is needed. A few changes suggested by shellcheck also.

### DIFF
--- a/updatebeta
+++ b/updatebeta
@@ -7,26 +7,90 @@ lowercase(){
     echo "$1" | sed "y/ABCDEFGHIJKLMNOPQRSTUVWXYZ/abcdefghijklmnopqrstuvwxyz/"
 }
 
-OS=`lowercase \`uname -s\``
-MACH=`uname -m`
-ARCH=`dpkg --print-architecture`
+# 1st determine if we run script extracted from update archive or current installed version?
+# Also print CRC to allow check if updated upgrade script differs from current...
+CALLER=$(ps -o args= $PPID)
+CALLEE=$(basename "$0")
+SCRIPTCRC=$(crc32 "$0")
+SCRIPTDIR="$( cd -- "$(dirname "$0")" >/dev/null 2>&1 || exit ; pwd -P )"
+[ "$SCRIPTDIR" != "$PWD" ] && echo "Script must be started from Domoticz directory!" && exit 1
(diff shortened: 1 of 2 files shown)


gizmocuz:
Thank you for this PR!

Why this 'printf' change at the end instead of the 'echo' ?

It is not only the openssl version that could cause an incompatibility, it is also the Debian version

 cat /etc/debian_version
 
 Not I think the output from this should be compared to a list.
 
 Accepted at the moment are:
bookwork
trixie

Maybe you could add this check as well?

lym70:
Hello,

Change for printf was one of shellcheck suggestions as echo does not handle "\n"...

Can add Debian version check, but I'm not sure all users use Debian so this may be a bit too specific? Is there a more specific way to target this incompatibility as I'm not aware of this.

Something else that may IMO be considered is shebang, looking at execution current version script call, before extracting updated one and going ahead with it shows:
 
>> Running 'updaterelease' (from '/home/yann/test_update', crc32=4a3abbe0)
>> Caller: 'bash'

Going ahead with up to date same script from archive then shows:

>> Running 'updaterelease' (from '/home/yann/test_update', crc32=4a3abbe0)
>> Caller: '/bin/sh ./updaterelease'

=> We 1st use default user shell (so bash in my case and probably many others) for current version script call, but after this updated one use the exact shebang sh.

Did not observed side effect myself, but shells have incompatibilities... Anyway, as this'll be used by beta users first this'll allow various test configurations.

gizmocuz:
@lym70 , updating this script was on my wish list quite some time.
Thanks to your ideas I created a new script (and enhanced it with AI), which took me quite some debugging as it did not work the way I originally wanted it to work

https://github.com/domoticz/domoticz/blob/development/updatebeta

Feel free to make any modifications if you want, and if possible test it...

Edit: beta 16989 already includes this script

gizmocuz:
Might have an issue:

https://forum.domoticz.com/viewtopic.php?f=6&t=44190

Could be related to the fact that the script is now launched twice, or could be a buggy internet

gizmocuz:
Found the issue, should work correctly again

PR #6465, opened 2025-11-28
Additional validations in PhilipsHue to prevent crashes and generate nice errors

gbonny1982:
I've added a lot more JSON parsing validations and safe checks. 
Polling interval must be 5 or more (HTTPS requires more cycles from Domoticz host and Hue Bridge).
Code cleanup can be done later, when this helps.
There are probably more additional fixes needed in httpclient.cpp to make it more robust and not only for PhilipsHue.

### DIFF
--- a/hardware/PhilipsHue/PhilipsHue.cpp
+++ b/hardware/PhilipsHue/PhilipsHue.cpp
@@ -26,6 +26,13 @@
 #define SensorTypeZLLLightLevel "ZLLLightLevel"
 #define SensorTypeGeofence "Geofence"
 
+#define SAFE_TYPE_CHECK(obj, check, fieldname) \
+    do { \
+        if (!((obj).check())) { \
+            _log.Log(LOG_DEBUG_INT, "PhilipsHue: JSON field '%s' has unexpected type", (fieldname)); \
+        } \
+    } while (0)
+
 //#define DEBUG_PhilipsHue
 
 #ifdef DEBUG_PhilipsHue
@@ -73,10 +80,11 @@ CPhilipsHue::CPhilipsHue(const int ID, const std::string& IPAddress, const unsig
 	m_use_v2_sensors = (effectiveOptions & HUE_USE_V2_SENSORS) != 0;
 
 	// Catch uninitialised Mode1 entry.
-	if (m_poll_interval < 1)
+	// if (m_poll_interval < 1)
+	if (m_poll_interval < 5)
 	{
 		m_poll_interval = HUE_DEFAULT_POLL_INTERVAL;
-		Log(LOG_STATUS, "Using default poll interval of %d secs.", m_poll_interval);
+		Log(LOG_STATUS, "Invalid value. Using default poll interval of %d secs.", m_poll_interval);
 	}
 	else
 	{
@@ -311,6 +319,78 @@ bool CPhilipsHue::WriteToHardware(const char* pdata, const unsigned char /*lengt
 	return true;
 }
 
+// -------------------------------------------------------------------
+// Safe wrapper for ParseJSon with truncation detection + exception protection
+// -------------------------------------------------------------------
+static bool SafeParseJson(const std::string& sResult, Json::Value& root)
+{
+    // Trim right
+    auto trim_right = [](const std::string &s)->std::string {
+        size_t i = s.find_last_not_of(" \t\r\n");
+        if (i == std::string::npos) return std::string();
+        return s.substr(0, i+1);
+    };
+
+    // Empty response?
+    if (sResult.empty()) {
+        _log.Log(LOG_ERROR, "PhilipsHue: Empty JSON response from bridge.");
+        return false;
+    }
+
+    // Detect likely truncated JSON
+    std::string sTrimmed = trim_right(sResult);
+    char last = sTrimmed.empty() ? '?' : sTrimmed.back();
+
+    if (sTrimmed.empty() || (last != '}' && last != ']'))
+    {
+        _log.Log(LOG_ERROR,
+            "PhilipsHue: Malformed or truncated JSON from Hue bridge (len=%zu).",
+            sResult.size());
+
+#ifdef DEBUG_PhilipsHue
+        SaveString2Disk(sResult, "/tmp/hue_bad_response.json");
+#endif
+        return false;
+    }
+
+    // Attempt parsing with full exception protection
+    bool ret = false;
+    try {
+        ret = ParseJSon(sResult, root);
+    }
+    catch (const std::exception& ex) {
+        _log.Log(LOG_ERROR,
+            "PhilipsHue: JSON parser threw exception: %s (len=%zu).",
+            ex.what(), sResult.size());
+#ifdef DEBUG_PhilipsHue
+        SaveString2Disk(sResult, "/tmp/hue_bad_response.json");
+#endif
+        return false;
+    }
+    catch (...) {
+        _log.Log(LOG_ERROR,
+            "PhilipsHue: JSON parser threw unknown exception (len=%zu).",
+            sResult.size());
+#ifdef DEBUG_PhilipsHue
+        SaveString2Disk(sResult, "/tmp/hue_bad_response.json");
+#endif
+        return false;
+    }
+
+    if (!ret) {
+        _log.Log(LOG_ERROR,
+            "PhilipsHue: JSON parser returned failure for response len=%zu.",
+            sResult.size());
+#ifdef DEBUG_PhilipsHue
+        SaveString2Disk(sResult, "/tmp/hue_bad_response.json");
+#endif
+        return false;
+    }
+
+    return true;
+}
+
+
 bool CPhilipsHue::SwitchLight(const int nodeID, const std::string& LCmd, const int svalue, const int svalue2 /*= 0*/, const int svalue3 /*= 0*/)
 {
 	std::vector<std::string> ExtraHeaders;
@@ -459,17 +539,31 @@ bool CPhilipsHue::SwitchLight(const int nodeID, const std::string& LCmd, const i
 
 	Json::Value root;
 
-	bool ret = ParseJSon(sResult, root);
-	if (!ret)
-	{
-		Log(LOG_ERROR, "Invalid data received (Switch Light/Scene), or invalid IPAddress/Username!");
+	// bool ret = ParseJSon(sResult, root);
+	// if (!ret)
+	// {
+	// 	Log(LOG_ERROR, "Invalid data received (Switch Light/Scene), or invalid IPAddress/Username!");
+	// 	return false;
+	// }
+	if (!SafeParseJson(sResult, root))
 		return false;
-	}
 
-	if (sResult.find("error") != std::string::npos)
-	{
-		//We had an error
-		Log(LOG_ERROR, "Error received: %s", root[0]["error"]["description"].asString().c_str());
+
+	// if (sResult.find("error") != std::string::npos)
+	// {
+	// 	//We had an error
+	// 	Log(LOG_ERROR, "Error received: %s", root[0]["error"]["description"].asString().c_str());
+	// 	return false;
+	// }
+	if (sResult.find("\"error\":") != std::string::npos) {
+		// normalized handling: support both array-of-errors and object error
+		if (root.isArray() && root.size() > 0 && root[0].isObject() && root[0].isMember("error")) {
+			Log(LOG_ERROR, "Error received: %s", root[0]["error"]["description"].asString().c_str());
+		} else if (root.isObject() && root.isMember("error")) {
+			Log(LOG_ERROR, "Error received: %s", root["error"]["description"].asString().c_str());
+		} else {
+			Log(LOG_ERROR, "Hue bridge returned an error but JSON layout was unexpected");
+		}
 		return false;
 	}
 	return true;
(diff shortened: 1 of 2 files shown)


gizmocuz:
Thank you for this PR.
Aldo AI is very nice, I have some recommendations for you.

First, use the already available string functions in helper.h, like 

stdstring_rtrim
stdstring_ltrim
stdstring_trim

etc

To check if the JSON is valid, just use what is used everywhere

```
	Json::Value root;
	bool ret = ParseJSon(sResult, root);
	if ((!ret) || (!root.isObject()))
	{
		Log(LOG_ERROR, "Invalid data received! (login/json)");
		return false;
	}
```

Next check if a key exists that should always be there, like:

```
	if (root["data"].empty())
	{
		Log(LOG_ERROR, "Invalid data received! (no session_id)");
		return false;
	}
```

If you look at most hardware classed that are using HTTP, there are helpers like

#ifdef _DEBUG
//#define DEBUG_EnphaseAPI_R
//#define DEBUG_EnphaseAPI_W
#endif

I noticed in the PhilipsHUE code there are no read and write defines, but that makes it easier to just safe, or load, or both, maybe you could add this

When a HTTP returns, the JSON cannot be parsed, it is sufficient to just scan the string for 'error' and print there was an error.
This code is not so nice:

if (sResult.find("\"error\":") != std::string::npos) {
		// normalized handling: support both array-of-errors and object error
		if (root2.isArray() && root2.size() > 0 && root2[0].isObject() && root2[0].isMember("error")) {
			Log(LOG_ERROR, "Error received: %s", root2[0]["error"]["description"].asString().c_str());
		} else if (root2.isObject() && root2.isMember("error")) {
			Log(LOG_ERROR, "Error received: %s", root2["error"]["description"].asString().c_str());
		} else {
			Log(LOG_ERROR, "Hue bridge returned an error but JSON layout was unexpected");
		}
		return false;
	}
	
You could make one function that extracts the error

I will merge this PR and make some changes and let you know when done so others could test both our modifications

gizmocuz:
The crashes where also not there with the original code, there was no need to make modifications to them (Like SwitchLight, RegisterUser etc)

gizmocuz:
Question, it seems the polling of the V2 Lights/contacts/etc/etc seems to fill some array vectors and they are cleared on every new poll.
It seems nothing is done with it, is there anything done with the V2 polling?
If not, maybe only enable this in debug mode, otherwise we have 10 more HTTP requests to the hub that are not used/needed

gbonny1982:
You've reverted quiet a lot I see. Regarding the string functions in helper.h: I was not aware as I'm not a Domoticz code expert yet :)

What I was wondering, isn't there a "safe JSON parser" in helper.cpp which can be used/benefit throughout Domoticz?

A couple of questions I have by some reverts. As I mentioned earlier I'm not an C++ expert either. But maybe I can learn something from your answers/decisions.

Before calling: 
`bool ret = ParseJSon(sResult, root);`

Why wouldn't you check _sResult.empty()_ first? Its not done in httpclient either. It would help with debuging and could give you more precise error messages.

Why wouldn't you put _bool ret = ParseJSon(sResult, root);_ between a try/catch?
ParseJSon() does not protect against invalid UTF-8
ParseJSon() does not protect against truncated JSON
JsonCpp throws exceptions when downstream code calls .asString(), .operator[], etc.

regarding SafeParseJson removal:
sResult may be empty, binary garbage, HTML, partial JSON, etc.
If the Hue Bridge is overloaded, returns 503, times out, or returns partial TLS frames:
```
sResult = ""         (empty)
sResult = "<html>"   (bridge error page)
sResult = "{... tra" (truncated)
sResult = "\x16\x03" (TLS junk)
```

I see:
```
hscene.name = scene.get("name", "").asString()
hscene.lastupdated = scene.get("lastupdated", "").asString()
```

changed back into (line 1175):
```
hscene.name = scene["name"].asString();
hscene.lastupdated = scene["lastupdated"].asString();
```

Name or scene can be missing and get() provides a safe default value. Which is safer right?


Here you assume it is a boolean, its safer to check it first:
`tlight.on = lightstate["on"].asBool();`
Same for integers along the way.


Here you assume its size is 2, its safer to check it first:
```
if (!lightstate["xy"].empty())
	{
		//Lamp with color control
		hasHueSat = true;
		tlight.x = lightstate["xy"][0].asDouble();
		tlight.y = lightstate["xy"][1].asDouble();
```


A couple of typos:
Retuned > Returned
json > JSON
Poll interval to short > Poll interval too short


`if ((!ret) || (!root.isObject()))`
Does not prevent:
invalid JSON that still "parses"
null root used later → .operator[] crash
accessing non-object fields
JsonCpp internal errors


Possible inconsistency? Or is it on purpose?
`sResult.find("error") vs sResult.find("\"error\":")`


regarding your "_sResult.find("error")_" solution after "_if ((!ret) || (!root.isObject()))_" (line 500 for example):

`[ { "error": { "description": "Invalid value" } } ]`

→ find("error") works
→ But root.isObject() == false, so your code rejects ALL these legitimate error responses incorrectly.

Compared to the suggested solution. I agree it could have been in one function.

As it was uncertain where the crashes exactly happened, in GetGroups or in SwitchLight or RegisterUser etc. As always there is always room for improvement. 
And as you can read in the issue 6451 from matmar29, current error message comes from SwitchLight() line 496. So it seems?
So what I see are several assumptions which make code trivial, right? I'm glad it seems that it doesn't crash Domoticz anymore :)

I also see you've made more corrections/additions which are great, also for v2. Thanks for helping out. WTG!

gizmocuz:
Correct, I changed quite a lot, and reverted some changes as well.
(Just corrected the \"error\": search on the two locations, so they are now consistent with the rest.

You can assume a lot. Even for the light "on" state we expect a Boolean, but what if it is a Int or a String?
Most things you added because an error was introduced and you start debugging where this issue was, which is OK,
but it is unnecessary to double/tripple check API implementations.
I added also a try/catch, so if this ever happens (quite sure not for V1), it is catched and can be tracked.

Thanks for the great work!

kiddigital:
Maybe first check the HTTP resultcode AND content-type before assuming it is a JSON object that is returned and ready to be parsed?

And that should be a generic way for all kind of HTTP API requests and not specific to HUE.

gizmocuz:
@kiddigital , correct! That's also done, did we miss something?
Every return HTTPClient::GET/POST/PUT should be checked first, next parse the json (ParseJSon, also a boolean)

gbonny1982:
Another possible improvement for HTTPClient.

In SetGlobalOptions right after:
`curl_easy_setopt(curl, CURLOPT_COOKIEJAR, domocookie.c_str());`
add:
```
// ---- TCP keepalive ---- 
curl_easy_setopt(curl, CURLOPT_TCP_KEEPALIVE, 1L); 
curl_easy_setopt(curl, CURLOPT_TCP_KEEPIDLE, 30L); 
curl_easy_setopt(curl, CURLOPT_TCP_KEEPINTVL, 15L); 

// ---- Allow HTTP/2 where available ---- 
curl_easy_setopt(curl, CURLOPT_HTTP_VERSION, CURL_HTTP_VERSION_2TLS);
 
// ---- Let curl reuse existing connections ---- 
curl_easy_setopt(curl, CURLOPT_PIPEWAIT, 1L); 
curl_easy_setopt(curl, CURLOPT_MAXCONNECTS, 20L);
```

https://curl.se/libcurl/c/CURLOPT_TCP_KEEPALIVE.html
https://curl.se/libcurl/c/CURLOPT_HTTP_VERSION.html
https://curl.se/libcurl/c/CURLOPT_PIPEWAIT.html
https://curl.se/libcurl/c/CURLOPT_MAXCONNECTS.html

gizmocuz:
CURLOPT_TCP_KEEPALIVE, I think this would only work if the HTTPClient was a member of your class.
Else it is called statically, and you could have 5 different types of hardware using it

It is also enabled by default:
https://everything.curl.dev/usingcurl/connections/keepalive.html

CURLOPT_PIPEWAIT would introduce a possible long wait?

CURL_HTTP_VERSION_2TLS, will first try a HTTP 2 connection, and will fallback to HTTP 1.1 ...
Assuming not much web servers serve with HTTP 2, this would introduce a extra handshake/delay/call/data

gbonny1982:
> CURLOPT_TCP_KEEPALIVE, I think this would only work if the HTTPClient was a member of your class. Else it is called statically, and you could have 5 different types of hardware using it
> 
> It is also enabled by default: https://everything.curl.dev/usingcurl/connections/keepalive.html

Each hardware could have its own connection with keep-alive enabled (or disabled in other cases). There is no need to close the TCP connection: for example when polling every 5 seconds, multiple calls possibly per poll, depending on the hardware.

Quote from the documentation link that I provided "_Set to 0 (default behavior) to disable keepalive probes._ "
I don't know what is true. It could be relevant for HTTP/1.1 connections that support it. Hue doesn't support keepalive for API v1 using HTTP/1.1 for instance. However my HomeWizard water meter supports it for HTTP/1.1.


> CURLOPT_PIPEWAIT would introduce a possible long wait?

I believe it checks for an existing connection and/or waits for a short time and then multiplexes multiple calls over one connection when supported, instead of creating new connection per call. It might add a very small delay, I don't know. No experience here.

> CURL_HTTP_VERSION_2TLS, will first try a HTTP 2 connection, and will fallback to HTTP 1.1 ... Assuming not much web servers serve with HTTP 2, this would introduce a extra handshake/delay/call/data

I thought it was an enabler, however when you do not set it, curl sees what fits best.
One third of the internet is using HTTP/2.0 last year. Hue API v2 supports it too, see the output of a manual curl below. HTTP/2 supports keepalive native.
```
* ALPN: curl offers h2,http/1.1
..
* ALPN: server accepted h2
* using HTTP/2
* [HTTP/2] [1] OPENED stream for https://..
```

You would see: "* Connection #0 to host a.b.c.d left intact"

Try:
curl -vk https://<hue-ip>/api/<api-key>/lights/  https://<hue-ip>/api/<api-key>/lights/
curl -v http://<hue-ip>/api/<api-key>/lights/  http://<hue-ip>/api/<api-key>/lights/


So nevermind about the http/2, the other 2 could be useful still I guess. Maybe there's someone with CURL experience around here?

```
// watermeter 2 calls
$ curl -v http://<ip>/api/v1/data http://<ip>/api/v1/data
*   Trying <ip>:80...
* Connected to <ip> (<ip>) port 80
> GET /api/v1/data HTTP/1.1
> Host: <ip>
> User-Agent: curl/8.5.0
> Accept: */*
>
< HTTP/1.1 200 OK
< Content-Type: application/json
< Content-Length: 124
< Access-Control-Allow-Origin: *
<
* Connection #0 to host <ip> left intact
{json-output}
* Found bundle for host: 0x5871c2041791 [serially]
* **Can not multiplex**, even if we wanted to
* **Re-using existing connection** with host <ip>
> GET /api/v1/data HTTP/1.1
> Host: <ip>
> User-Agent: curl/8.5.0
> Accept: */*
>
< HTTP/1.1 200 OK
< Content-Type: application/json
< Content-Length: 124
< Access-Control-Allow-Origin: *
<
* Connection #0 to host <ip> left intact
{json-output}
```
You see can not multiplex and re-using existing connection here.

When I enable CURL verbose in Domoticz code, I don't see "re-using existing connection with host.." for my watermeter.
With or without the above options, and including the with following extra code:
```
struct curl_slist* HTTPClient::m_defaultHeaders = nullptr;

// Combine default headers with extra headers, returning a new curl_slist*
static struct curl_slist* BuildMergedHeaders(struct curl_slist* default_headers,
                                             const std::vector<std::string>& ExtraHeaders)
{
    struct curl_slist* headers = nullptr;

    // Copy default headers
    for (struct curl_slist* p = default_headers; p; p = p->next)
        headers = curl_slist_append(headers, p->data);

    // Add extra headers
    for (const auto &h : ExtraHeaders)
        headers = curl_slist_append(headers, h.c_str());

    return headers;
}

...in setglobaloptions
if (!m_defaultHeaders)
	{
  m_defaultHeaders = curl_slist_append(m_defaultHeaders, "Connection: keep-alive");
  m_defaultHeaders = curl_slist_append(m_defaultHeaders, "Keep-Alive: timeout=30, max=50");
  m_defaultHeaders = curl_slist_append(m_defaultHeaders, "Expect:");  // disables 100-continue
	}
curl_easy_setopt(curl, CURLOPT_HTTPHEADER, m_defaultHeaders);
...
.. in getbinary, postbinary, putbinary, patchbinary GETBinarySingleLine
		// struct curl_slist *headers = nullptr;
		// if (!ExtraHeaders.empty())
		// {
		// 	for (const auto &header : ExtraHeaders)
		// 	{
		// 		headers = curl_slist_append(headers, header.c_str());
		// 	}
		// 	curl_easy_setopt(curl, CURLOPT_HTTPHEADER, headers);
		// }
		// Merge default headers + plugin headers
		struct curl_slist* headers = BuildMergedHeaders(m_defaultHeaders, ExtraHeaders);

//added in HTTPClient.h
private:
    	static struct curl_slist* m_defaultHeaders;
```
I might be missing something here.

gbonny1982:
I've just tested my Hue Secure door/window contacts, Hue temp / lux / motion sensor and Hue light (with dim function). They still work/function with build 16988 on my development machine. 

Didn't try a long run yet (like a day or so). Woody1 seems to have issues still, see Forum post. https://forum.domoticz.com/viewtopic.php?t=44133&start=20

PR #6466, opened 2025-11-29
Support compressed .gz files in PresentOauth2LoginDialog

dwmw2:
Try to load iam_auth.html.gz first, falling back to uncompressed iam_auth.html if not found. Decompress gzipped content using CGZIP2AT before setting reply content.


gizmocuz:
Thank you for this PR. But there is no compressed version? ;)

PR #6467, opened 2025-11-29
Honour ValueStep/Min/Max/Unit for combined Thermostat devices

dwmw2:
The UI was already looking for these fields; we just needed to export them.


gizmocuz:
Thank you!!

PR #6469, opened 2025-11-30
Make it possible to force building without telldus-core

dwmw2:
Just because you can find it, doesn't mean we want to link directly against it. It can be found with dlopen() anyway.


gizmocuz:

PR #6470, opened 2025-12-01
MQTT-AD: Drop duplicate messages

dwmw2:
ESPHome sends duplicate messages to MQTT every time it polls the hardware sensor, even if the value hasn't changed. Drop them in MQTT-AD if the previous one is less than ten minutes old, rather than wasting a lot of time processing the 'updates'.


dwmw2:
Closing for now as apparently there are cases where we do want to run automations even when nothing has changed.

gizmocuz:
Only for specific devices like buttons/scenes

PR #6472, opened 2025-12-01
Spread data spikes over missing days in counter graphs

dwmw2:
When there are gaps in the data (e.g., device offline for multiple days), the accumulated usage appears as a large spike on the day data resumes. This preprocessing function detects gaps and spreads the accumulated value evenly across the missing days, preventing spikes from affecting graph scale.

It used to look like this:
Screenshot from 2025-12-01 11-32-14

... now it looks like this:

Screenshot from 2025-12-01 11-53-24

### DIFF
--- a/main/SQLHelper.cpp
+++ b/main/SQLHelper.cpp
@@ -7893,13 +7893,6 @@ void CSQLHelper::AddCalendarUpdateMeter()
 				}
 			}
 		}
-		else
-		{
-			//no new meter result received in last day
-			result = safe_query("INSERT INTO Meter_Calendar (DeviceRowID, Value, Price, Date) "
-					    "VALUES ('%" PRIu64 "', '%.2f', '%.4f', '%q')",
-					    ID, 0.0F, 0.0F, szDateStart);
-		}
 	}
 }
 
--- a/www/app/log/CounterLogCounter.js
+++ b/www/app/log/CounterLogCounter.js
@@ -48,6 +48,9 @@ define(['app', 'lodash', 'log/Chart', 'log/CounterLogParams', 'log/CounterLogCou
                     .concat(counterLogCounterSeriesSuppliers.counterTrendlineSeriesSuppliers(deviceTypeIndex, chart.valueMultipliers.m1000))
                     .concat(counterLogCounterSeriesSuppliers.counterPreviousSeriesSupplier(deviceTypeIndex, chart.valueMultipliers.m1000));
             },
+            preprocessMonthYearData: function (data) {
+                counterLogSeriesSupplier.fillMissingDays(data);
+            },
             extendDataRequestCompare: function (dataRequest) {
                 return dataRequest;
             },
@@ -111,6 +114,9 @@ define(['app', 'lodash', 'log/Chart', 'log/CounterLogParams', 'log/CounterLogCou
                     .concat(counterLogCounterSeriesSuppliers.counterTrendlineSeriesSuppliers(deviceTypeIndex, chart.valueMultipliers.m1))
                     .concat(counterLogCounterSeriesSuppliers.counterPreviousSeriesSupplier(deviceTypeIndex, chart.valueMultipliers.m1))
 					.concat(counterLogCounterSeriesSuppliers.counterPriceSeriesSuppliers());
+            },
+            preprocessMonthYearData: function (data) {
+                counterLogSeriesSupplier.fillMissingDays(data);
             },
             extendDataRequestCompare: function (dataRequest) {
                 return dataRequest;
@@ -176,6 +182,9 @@ define(['app', 'lodash', 'log/Chart', 'log/CounterLogParams', 'log/CounterLogCou
                     .concat(counterLogCounterSeriesSuppliers.counterPreviousSeriesSupplier(deviceTypeIndex, chart.valueMultipliers.m1, times1000, 0))
 					.concat(counterLogCounterSeriesSuppliers.counterPriceSeriesSuppliers());
             },
+            preprocessMonthYearData: function (data) {
+                counterLogSeriesSupplier.fillMissingDays(data);
+            },
             extendDataRequestCompare: function (dataRequest) {
                 return dataRequest;
             },
@@ -190,6 +199,7 @@ define(['app', 'lodash', 'log/Chart', 'log/CounterLogParams', 'log/CounterLogCou
             },
             preprocessMonthYearData: function (data) {
                 preprocessData.call(this, data);
+                counterLogSeriesSupplier.fillMissingDays(data);
             },
             chartParamsDayTemplate: {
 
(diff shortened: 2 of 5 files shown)


dwmw2:
For some reason this isn't working on my production box even though it was working in testing on the same type of meter device.

dwmw2:
Ah, I figured it out. My *test* server had gaps where Domoticz itself hadn't been running. My production server had been *running* but the meter data hadn't been coming in, so there were actual zeroes in the data and not just gaps:
```
    {
      "d": "2025-03-02",
      "v": "97.703"
    },
    {
      "d": "2025-03-03",
      "v": "87.305"
    },
    {
      "d": "2025-03-04",
      "v": "0.000"
    },
    {
      "d": "2025-03-05",
      "v": "0.000"
    },
```
I'm confused by the presence of zeroes; surely we shouldn't be writing to `Meter_Calendar` for that day if the device hasn't actually updated at all?

We can tweak the graph code to treat those like empty data, but I'm less sure that we should. What do you think?

dwmw2:
Fixed the code which introduced the zeroes. We deliberately don't introduce explicit zeroes in the `Meter` table, and we shouldn't in `Meter_Calendar` either. Then the JS code which smooths over *empty* days will be fine.

0crap:
> Fixed the code which introduced the zeroes. We deliberately don't introduce explicit zeroes in the `Meter` table, and we shouldn't in `Meter_Calendar` either. Then the JS code which smooths over _empty_ days will be fine.

Just to be sure, this does not alter my solar power "Meter_Calendar" table?
When the solar panels are fully covered with snow there are days with zeroes in that table.
Which is totally legit.

Sorry if unrelated, just asking...

dwmw2:
Good question. No, it should not affect those days because they should have an *explicit* zero rather than just an absence of data.

0crap:
Thx, good to know.

I do understand your use case, but e.g. `Meter` and `Meter_Calendar` should reflect the truth as close as possible.
Let code automagically alter it feels a bit uncomfortable.
Maybe I'm just overly concerned.

I do appreciate your time!

0crap:
If I may suggest, another way of looking at your use case might be to change the autoscaler behavior?
Let the autoscaler ignore one or two stupid high readings...

gizmocuz:
As you know, there is still an issue with Enphase/Envoy and the latest beta/counterhelper.
This needs to be addressed and solved.
If this was masked, we have no idea of a spike at all.. not sure if that is desired?

0crap:
I don't think @dwmw2 is specifically having an issue with his Enphase/Envoy. That is not stated in his opening post.
I personally did have an issue in that regard and got confronted with a spike in my Generated graph.

I do agree that I would not like to have that spike get masked. That also would have masked the underlying problem.
Which in my case, i guess, is somehow related to new Envoy firmware.

My personal opinion is that Domoticz should not alter `Meter` or `Meter_Calendar` data on it's own.
Having code in place that can do that, makes me feel a bit uncomfortable.
I am not a coder, so I can't comment on the OP's code, I'm just trying to give some feedback by reading his comments.

(It is fair to say that I manually needed to edit the database to fix my issue.
Not by changing the column `Value` but by making the column `Counter` consistent again.)

dwmw2:
> My personal opinion is that Domoticz should not alter Meter or Meter_Calendar data on its own.

It's Domoticz which *creates* those tables. But yes, again: the 'smoothing' here is only in the *display*, and the tables aren't modified to contain the averaged-out data.

That does need a bug fix to avoid writing false explicit zeroes to the Meter_Calendar when there are no new data during that day, which is the only change to the Meter_Calendar handling.

gizmocuz:
@0crap , ("do it in the database") , you should not do this and can be handled by shift-click / delete the datapoint in the month/year graph

gizmocuz:
Thank you for this contribution, David — the idea of spreading accumulated values over gap days is exactly right and we have included it in development.

We cherry-picked your JS changes and corrected them in two ways before merging:

**1. Zero-value rows were not handled**
Your `fillMissingDays` only spread values when date entries were entirely *missing* from the result (`daysDiff > 1`). However, when a device goes offline Domoticz writes a row to `Meter_Calendar` for each day with `Value=0` and an unchanged counter. Those rows *are* present in `data.result`, so `daysDiff` between consecutive dates is always 1 and the spike was never spread. The fix: buffer zero-value items and compute the gap from the last *non-zero* date, which naturally spans both missing and zero-value days.

**2. The price field was not spread**
The accumulated price (`p` field / "Earned" line) was not divided across the gap days, leaving a spike in the cost graph on the return day. Fixed by dividing `p` by `daysDiff` the same way `v` is.

We also included your `SQLHelper` change (removing the zero-entry insertion) and added spike highlighting for anomalous meter values in the year/month charts.

Closing this PR as the changes are now in `development`. Thanks again!

0crap:
> @0crap , ("do it in the database") , you should not do this and can be handled by shift-click / delete the datapoint in the month/year graph

Done it many times with this Enphase Envoy-S nonsense, nothing to be afraid of. Just make sure you have backups.
(Shift-click / delete does not always solve the issue.)
I hope the root cause can be found. All this patching up should not be needed.

dwmw2:
> Your fillMissingDays only spread values when date entries were entirely missing from the result (daysDiff > 1). However, when a device goes offline Domoticz writes a row to Meter_Calendar for each day with Value=0 and an unchanged counter.

I thought those explicit zero entries were what the SQLHelper change was fixing? We shouldn't have them any more, so a day when we *measured* zero is different from a day when we didn't measure anything at all?

I had deliberately *not* excluded explicit zeroes from the graph.

gizmocuz:
@dwmw2 , that is correct, it is dead code, but kept

The graphs now display correctly with your patches applied among other things

gizmocuz:
@0crap, is it still happening? I restart/update my system many times without issue. Yes, when the envoy restarts this can happen. This is now catched?

gizmocuz:
You raise a valid point — after the SQLHelper fix, explicit zero rows should represent genuine zero-consumption days and not stale offline entries. Spreading them is technically incorrect in that case.

However, the reason we kept the zero-handling in `fillMissingDays` is for existing users upgrading with historical data already in their DB. Those databases have stale zero rows inserted by the old code path, and without handling them, those users would still see spikes on the return day after an offline period — even after upgrading.

Genuine zero-consumption days on energy/gas meters are rare in practice, so the practical impact of incorrectly spreading them is low. The historical-data case affects many existing users.

If you have a suggestion for distinguishing the two cases (e.g. checking whether `c` is identical to the previous row's `c` value, which would indicate a stale entry vs. a real measurement), we're open to a cleaner approach.

dwmw2:
My logic was that my combined approach would do the right thing for all future data, while existing historical data would still be displayed as it is displayed today. And if users want to delete the stray explicit zeroes, they can.

0crap:
> My logic was that my combined approach would do the right thing for all future data, while existing historical data would still be displayed as it is displayed today. And if users want to delete the stray explicit zeroes, they can.

Snow covered solar panels is still working as normal?
That generates days with 0 production, which is totally legit.

gizmocuz:
@0crap, That was totally not the issue here. But when it snows and that happens, look at the graph.

0crap:
> @0crap, is it still happening? I restart/update my system many times without issue. Yes, when the envoy restarts this can happen. This is now catched?

Yes, but I'm still running 2025.2 build 17060.
To be honest I'm waiting for some kind of stable release.
Do you have a suggestion for a beta build I might consider?
Doesn't need to be the latest, but stability is priority.

gizmocuz:
@0crap , I think the current beta is stable enough for you. What is not stable for you about it?

0crap:
> @0crap , I think the current beta is stable enough for you. What is not stable for you about it?

I don't know, never tried anything beyond build 17060.
Just asking.

0crap:
> @0crap, That was totally not the issue here. But when it snows and that happens, look at the graph.

Fair enough, just asking.
When there is snow my current build shows a day with no production, which is as it should be.

gizmocuz:
@0crap , shutdown Domoticz, make a backup of the complete folder, and upgrade to the latest beta.

Command prompt

./updatebeta

User Interface -> Settings -> Release Channel -> Beta , Settings -> Check for Updates

Docker, use the append :latest to the image

gizmocuz:
This way you can also help with testing the latest version

PR #6473, opened 2025-12-01
MQTT-AD: Calculate temperature trends for climate devices

dwmw2:
Add trend calculation for temperature values in climate devices, matching the pattern from commit e72ac4eb4. Updates both the thermostat (line 4466) and setpoint (line 4567) update paths to:
- Add ID to SELECT queries
- Update result array indices after adding ID field
- Calculate and store temperature trends using trend_calculator


gizmocuz:
Thank you!!

PR #6474, opened 2025-12-01
Do not force re-enable RemoteSharedPort on startup

dwmw2:
The user can currently set RemoteSharedPort to 0 to disable it, which is consistent with the -www and -sslwww command line options and works fine... until Domoticz is restarted, when it turns it back on again.

Stop forcing it back on again at restart, and add a note to the input box saying '0 = Disabled'.


gizmocuz:

PR #6475, opened 2025-12-02
Dashboard: Add setpoint display and editing for Thermostat6 devices

dwmw2:
- Display setpoint value in bigtext (first position) for both mobile and desktop views
- Add override icon button to adjust setpoint via ShowSetpointPopup
- Use itemtablesmalldoubleicon table layout for devices with setpoints
- Support step/min/max parameters from device options
- Exclude Evohome Zone/Hot Water devices (they use different dialogs)
- Mobile view shows setpoint as clickable button
- Desktop view shows override icon next to temperature icon

Screenshot from 2025-12-02 23-14-02

edited_25267603-6c70-48c3-a83e-1bfff4adc1ce4080173468839443533

### DIFF
--- a/www/app/DashboardController.js
+++ b/www/app/DashboardController.js
@@ -995,7 +995,17 @@ define(['app', 'livesocket'], function (app) {
 					if (($scope.config.DashboardType == 2) || (window.myglobals.ismobile == true)) {
 						var status = "";
 						var bHaveBefore = false;
+						if (typeof item.SetPoint != 'undefined') {
+							var step = item.step || 0.5;
+							var min = item.min || -200;
+							var max = item.max || 200;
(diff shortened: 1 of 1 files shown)


dwmw2:
~~Doesn't work for Evohome devices yet, because their SetSetpoint dialog is in TemperatureController.js and would need to be moved somewhere global. If we actually want to do this (and if we can make it pretty enough) then we can do that...~~

Turns out evohome can use the basic setpoint dialog without the schedule support. I think that makes sense from the dashboard page, so I've enabled it.

dwmw2:
Dunno if I should be using `override.png` for that, and I also needed to scale it because it looks weird next to the other icon (just because the visible image part happens to go lower in the .png).

Smanar:
>>Support step/min/max parameters from device options

Hello, just by curiosity, if the widget is used for other hardware, you know if it will be possible to set the min/max using the API ?

dwmw2:
That's in #6477

Smanar:
Hello, I have a question again ^^.
Why this device have the "titlebar" orange ?
Normal is blue, timeout is red, but orange ?

dwmw2:
It's colour coded by (target, IIRC) temperature. Exactly like the Evohome zones

Smanar:
Ok, thx, I will check the code direclty, but I think  it's a "soon timeout".

PR #6477, opened 2025-12-04
Add setpoint field editing support for Thermostat 6 devices, add dzVents support

dwmw2:
Allow editing of ValueUnit, ValueStep, ValueMin, and ValueMax fields for pTypeThermostat6 devices in the temperature device edit dialog, matching the functionality available for setpoint devices in the utility controller.

Also add a dzVents device adapter for it.

### DIFF
--- a/dzVents/runtime/device-adapters/thermostat6_device.lua
+++ b/dzVents/runtime/device-adapters/thermostat6_device.lua
@@ -1,0 +1,65 @@
+local forecastMapping = {
+	['noinfo'] = 0,
+	['sunny'] = 1,
+	['partlycloudy'] = 2,
+	['cloudy'] = 3,
+	['rain'] = 4
+}
+
+local humidityMapping = {
+	['dry'] = 2,
+	['normal'] = 0,
+	['comfortable'] = 1,
+	['wet'] = 3
+}
+
+return {
+
+	baseType = 'device',
+
+	name = 'Thermostat 6 device adapter',
+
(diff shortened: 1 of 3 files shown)


dwmw2:
As noted elsewhere, setting the ValueUnit with this dialog *does* work, but just like with the existing standalone setpoints, the ValueUnit seems to be ignored. With some guidance on how it's actually *supposed* to work, we can fix those separately.

gizmocuz:
The ValueUnit is presented in the bigtext/top right of the widget. Also if you change this to  (degree sign)F it will use it for internal calculation.

dwmw2:
> The ValueUnit is presented in the bigtext/top right of the widget. Also if you change this to (degree sign)F it will use it for internal calculation.

AFAICT it *isn't* though, not even for standalone setpoint devices (and thus not for the combined thermostat, which were based on those). My commit in https://github.com/dwmw2/domoticz/commit/b4ef5bcb9883525a1b9bdb3a90eec9b9fc439a35 tries to change that, but I wasn't sure exactly how it's supposed to behave.

waltervl:
If you create a virtual setpoint device you can change the value unit, even to RPM or A (Ampere), so not temperature related.
Screenshot_20251204-150310 Firefox

Edit: so the value sent is interpreted for the user and scripts as the value to use. No other calculation done. That could be different though for C/F temperature conversion.  Not sure how that works.

gizmocuz:
Thank you for this PR!!

PR #6481, opened 2025-12-16
Set selector device active/inActive based on level

Tixav:
Without this,  domotice.devices("MySelector").active is always false and .inActive always true and it cannot be tested in the same way as boolean 'on/off' switches althought it can. Tested during 3 days without issue


gizmocuz:
There is no inActive/Active...

PR #6482, opened 2025-12-17
Fix update process

Onshuis:
Do not look if domoticz is running, just stop it anyway.


gizmocuz:
could you please undo all other changes not related to this fix? They where fine and also served as an example

PR #6486, opened 2025-12-20
Fix handle_stop call to use io_context_

arjendekorte:
Could it be this is missing? Up to (and including) Boost 1.89.0 this didn't seem to be a problem, but starting with Boost 1.90.0 I get compilation errors on this line, which seem to be fixed with this patch.


arjendekorte:
Wrong branch

PR #6487, opened 2025-12-20
Fix post call to use io_context_ for handle_stop

arjendekorte:
Could it be this is missing? Up to (and including) Boost 1.89.0 this didn't seem to be a problem, but starting with Boost 1.90.0 I get compilation errors on this line, which seem to be fixed with this patch.


gizmocuz:
Thanks for spotting. This is also used on other places in the code base, and I tested it with boost 1.89

arjendekorte:
I did indeed base my fix on other occurrences in the code. FWIW, later I found this is also what Copilot suggested as improvement for this line of code.

PR #6488, opened 2025-12-23
Replace deprecated deadline_timer with steady_timer

arjendekorte:
The deadline_timer was deprecated in Boost 1.87. The steady_timer was introduced in Boost 1.49 and is the most suitable replacement as it is guaranteed to be monotonic (unlike the system_timer and deadline_timer).


gizmocuz:
Thanks again Arjen, and happy new year!

arjendekorte:
Same to you.

PR #6496, opened 2026-01-02
Change updatebeta

Onshuis:
only change not check service, just stop service.


gizmocuz:
Thank you for the patch.
If this is the only change needed, could you undo all other changes in the script?

Issue #6492, opened 2025-12-30
Android app FCM notifications from Domoticz not working anymore since 18 or 19 december 2025


waltervl:
After investigation in Issue https://github.com/domoticz/domoticz-android/issues/723 it seems google revoked the FCM key.

🔔 FCM notifications stopped working – service account key revoked
Google has disabled the Firebase Admin SDK service account key used by Domoticz to send Firebase Cloud Messaging (FCM) notifications.

What happened
Google detected that a Firebase SDK FCM key was publicly exposed.
The exposed key was found in the domoticz/domoticz repository, and subsequently propagated via publicly available Docker images.

As a security measure, Google automatically revoked the compromised key.
link to the leaked key and why Google disabled it:
https://github.com/domoticz/domoticz/blob/cb3b96b72e5699eee726b33d9995fa374626841c/notifications/NotificationFCM.cpp#L65C43-L65C3214

@galadril can create a new key but there should be another method chosen to use the key without making it public

gizmocuz:
maybe the key was a year old?

galadril:
No i actually received a security policy violation risk email by Google that the key was exposed ;) 

This key was found at the following URL: https://europe-docker.pkg.dev/v2/artifact-registry-docker-cache/remote-dockerhub-mirror/domoticz/domoticz/blobs/sha256:b2d3fd0cbc9fa54efd613d5ce38841479063f11cc1459c4a42f3dd0939e05507

<img width="620" height="303" alt="Image" src="https://github.com/user-attachments/assets/ef8ace84-f66a-4ca5-a3dc-bd2c3e148983" />

galadril:
What could be an option, just [like with Home Assistant](https://developers.home-assistant.io/docs/android/tips/fcm_push_notification/), is that people setup their own FCM firebase project, and set the configuration of firebase in the app. i just created this page to allow users to bring their own FCM config

<img width="945" height="2048" alt="Image" src="https://github.com/user-attachments/assets/1016cde0-a2fb-41ba-b4c5-55ca14b7c285" />

So then Domoticz dont need to store the FCM Key anymore, just have a input field and store the key in the database.
So people need to do the setup themselfs if they like notaifications

galadril:
Ive added a quick sample on how we can describe the process of setting up FCM notifications on domoticz/Android app here
https://github.com/domoticz/domoticz-android/blob/beta-development/FIREBASE_SETUP.md

We can setup something like this on the wiki right @waltervl

[...]

PR #6498, opened 2026-01-03
Add FCM Service Account JSON configuration support

galadril:
- Introduced a new field for FCM Service Account JSON in NotificationFCM class.
- Updated IsConfigured method to validate the provided JSON.
- Enhanced SetupController to handle file uploads and JSON input for the service account.
- Modified setup.html to include a file upload option and a textarea for pasting JSON.


gizmocuz:
It could be a bit complicated for some users to do this. 
Is our current key not just a year old and needs to be refreshed?

waltervl:
No, it is has been rejected by Google for security issues: https://github.com/domoticz/domoticz/issues/6492#issuecomment-3704997547

waltervl:
Creating the google-services.json seems do-able for a user with the instructions in https://github.com/domoticz/domoticz-android/blob/6c4bca89d2cba96c391de9f3e9f465133f5411ff/FIREBASE_SETUP.md (and later put in the wiki)

gizmocuz:
I am not into Android application development or sorts, but I do have quite some few applications that notifies me (like Eufy, Ring, etc) without me needing to create a FCM JSON file...
Is it not possible to use one (private) application wide API key? Not sure how it works...
I am using PushOver wish is only a few $ lifetime

waltervl:
Ring, Eufy etc probably do not have public server and/or firmware repositories where the key is visible....

galadril:
@gizmocuz
I understand the comparison, but the architecture is fundamentally different.
Notifications are sent by a company-owned backend server that authenticates with FCM using private credentials that are never exposed. 



Domoticz works the opposite way:
- It is self-hosted
- Its not 1 privately owned backend, but people install their own Domoticz server in their homes
- The code and binaries are publicly distributed

Because of that, a single application-wide FCM key cannot be kept secure, it will inevitably leak and be revoked again (which is exactly what happened).

PushOver is a completely different model entirely: it’s a paid third-party notification service where key management and security are outsourced. So they keep their company owned private key also a secret, but they have an underlaying authentication/user/credential model. Also this is not locally installed.



So lets compare Domoticz FCM implementation with a similar application Home Assistant, which is also self-hosted and a better comparisson then Ring or Eufy.

They implemented FMC the exact same way:
https://developers.home-assistant.io/docs/android/tips/fcm_push_notification/

waltervl:
Here is how another opens source application with an android app has done it
https://github.com/nextcloud/notifications/blob/master/docs/push-v2.md

> Why is push-notifications.nextcloud.com necessary?
> 
> The Nextcloud mobile apps from the Google Play and Apple App Store are signed with Nextcloud developer keys or certificates. Push notifications sent to those devices need to be signed with a generated push key or certificate from the same developer account. The keys and certificates can not be shipped with the Nextcloud server as otherwise everyone would have our developer key and could manipulate releases or push to any random Nextcloud device. The Firebase Cloud Messaging (Google) and Apple Push Notification Service are not made for something like a federated project like Nextcloud and still assume there is a single entity behind them like with all the other services.
> 
> So we created the push proxy push-notifications.nextcloud.com to protect our users and their data. We took some extra efforts and reduced the available information to a bare minimum for each of the sections.

gizmocuz:
Pffff complicated.... But OK, i guess it is what it is.....
Thanks all for the patience 😊

PR #6505, opened 2026-01-12
Refactor FCM config to use separate fields instead of json and increasing pref svalue threshold to 5000

galadril:
This pull request refactors the Firebase Cloud Messaging (FCM) notification configuration to simplify setup and improve user experience. Instead of requiring users to paste or upload the entire Firebase Service Account JSON, the UI and backend now ask for only the essential fields: Client Email, Private Key, and Project ID. Additional improvements include better error handling.

### DIFF
--- a/main/SQLHelper.cpp
+++ b/main/SQLHelper.cpp
@@ -3301,7 +3301,7 @@ bool CSQLHelper::OpenDatabase()
 	UpdatePreferencesVar("DB_Version", DB_VERSION);
 
 	//Check preferences table for extreme sized sValues
-	result = safe_query("SELECT Key FROM Preferences WHERE LENGTH(sValue) > 1000");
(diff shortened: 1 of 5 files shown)


kiddigital:
Maybe @galadril , still ask to upload the full JSON, and still parse it, but only store the required fields.

Makes it less error prone for users to copy/paste the details. You can still verify if the JSON is valid, etc.

galadril:
With the latest commit there is still an upload option, and there is an option for them to fill in the fields themselfs.
The parse validates, but there is also a 'Test' butten to validate the contents of the fields.

So i guess it already contains what you mean?

galadril:
@gizmocuz what do you think?

gizmocuz:
Looks great! Thank you for this PR!!

waltervl:
Settings page seems malformed by this change: https://github.com/domoticz/domoticz/issues/6512

gizmocuz:
Fixed in latest code/betas

PR #6507, opened 2026-01-13
Rounding fix and support for negative values in MultiMeter table

Olifant1990:
Fixed rounding and use correct types as BIGINT in database is signed, not unsigned. This fixes a problem meter values can be negative

### DIFF
--- a/main/SQLHelper.cpp
+++ b/main/SQLHelper.cpp
@@ -7278,24 +7278,24 @@ void CSQLHelper::UpdateMultiMeter()
 			std::vector<std::string> splitresults;
 			StringSplit(sValue, ";", splitresults);
 
-			uint64_t value1 = 0;
-			uint64_t value2 = 0;
-			uint64_t value3 = 0;
-			uint64_t value4 = 0;
-			uint64_t value5 = 0;
-			uint64_t value6 = 0;
+			int64_t value1 = 0;
+			int64_t value2 = 0;
+			int64_t value3 = 0;
+			int64_t value4 = 0;
+			int64_t value5 = 0;
+			int64_t value6 = 0;
 
 			if (dType == pTypeP1Power)
 			{
 				if (splitresults.size() != 6)
 					continue; //impossible
 
-				uint64_t powerusage1 = 0;
-				uint64_t powerusage2 = 0;
-				uint64_t powerdeliv1 = 0;
-				uint64_t powerdeliv2 = 0;
-				uint64_t usagecurrent = 0;
-				uint64_t delivcurrent = 0;
+				int64_t powerusage1 = 0;
+				int64_t powerusage2 = 0;
+				int64_t powerdeliv1 = 0;
+				int64_t powerdeliv2 = 0;
+				int64_t usagecurrent = 0;
+				int64_t delivcurrent = 0;
 
 				try
 				{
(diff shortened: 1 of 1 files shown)


waltervl:
How can P1 smart meter values be negative? In what use case? For a normal kWh meter it can indeed be negative but for a P1 with splitted usage and return values there can not be a negative value in the database ( I think...).

Olifant1990:
It's not only a P1 meter, but an (dummy) Ampere device, which in my case could be negative.
But, not the case. BIGINT supports that, so use valid types in code too.

waltervl:
Ok, I would expect that multimeter was only used for a multimeter device like P1, not for a normal sensor like Ampere.

gizmocuz:
you just halved the possible counter values of all counters...
can you not just support this for the current sensor type? Counters should increase

Olifant1990:
No I did not? As I am understanding it correctly BIGINT in mysql is the same as int64_t. If you still want to support the uint64_t, the database should have BIGINT UNSIGNED.

So:

SQLite BIGINT: https://www.dofactory.com/sql/bigint: -9,223,372,036,854,775,808 to 9,223,372,036,854,775,807
int64_t C++: is a signed integer https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-dtyp/cfb2030b-0d7a-4808-8539-32f35bd9325f, values from –9223372036854775808 through 9223372036854775807.

I already was able to save negative values in the DB directly, but code did not accept it. So no change or halving of possible values.

And besides that, I did not touch the counters, the values only.

gizmocuz:
actually SQLite does not really care about the types...

Olifant1990:
Can you explain that?
If I insert a larger value then above stated values, indeed SQLite accepts it, but precicion is lost.
For example 9223372036854775808 becomes 9.2233720368548E+18 as stored in SQLite.
This will be received as 9223372036854800000. Ok, bigger then the max of int64_t, but not correct anymore.
And negative values are correctly possible in SQLite at the moment, so it really correspondents with int64_t

gizmocuz:
What device are you using that needs negative values? You don't have a P1 meter? This has different sensors for usage and delivery (also ampere)

"bigger then the max of int64_t,"

Yep, way bigger 😎

Olifant1990:
The hardware/device type I use is a dummy device (virtual sensor), the Ampere (3 Phase) device, which uses the MultiMeter table.
The real hardware is a custom build ESP with 3 amp meters, between 3 battery inverters and the power net. It can be negative if the batteries are powering my house.

About the values:
Be real about precicion, yes, we can store numbers larger then 9223372036854775807, as my tiny larger value told.
But they will be stored as floats with an exponent and cannot be loaded from the DB correctly.
I will do it with other examples:

18446744073709551615 (max of uint64_t) -> 1.844674407371E+19
18446744073709551616 (max of uint64_t + 1) -> 1.844674407371E+19

Both will get the same saved value and back from db to Domoticz, it will be showed as 18446744073710000000.

All values from 18446744073705000000 to 18446744073714999999 will have the same stored value 1.844674407371E+19

So we could safely agree that we better could use int64_t, which max value still is a veeeeerrry big number for a counter ;)
And we then can use the negative already supported values too

gizmocuz:
You're right, thanks for the fix!

Note that there are also several read paths that still use **uint64_t** / **std::stoull** for MultiMeter values
These would silently corrupt negative values read back from the database....

I have prepared some complementary changes covering:
  - CalcMultiMeterPrice() in SQLHelper.cpp
  - P1Power parsing in WebServer.cpp (GetJSonDevices)
  - Day/Month/Year graph code in WebServerHandleGraph.cpp

 Will merge both together to make the int64_t migration complete

Olifant1990:
Thank you too!

PR #6528, opened 2026-02-06
Document breaking update change for libmosquitto1 dependency

waltervl:
Added breaking change regarding system dependency update. You have to update twice.


gizmocuz:
Why do we need update twice? I updated once.. This is for users that are not on the latest beta's and do not have to new update scripts?
Something for the release post maybe... not the history, other packages might be included later.

waltervl:
If it is valid for the next stable or it is valid for beta users running an older beta it is a breaking change......
Perhaps the wording could be different eg You have to update twice when you are on a stable version (or old beta)

waltervl:
I see it is implemented with different wording. Thanks!

Issue #6531, opened 2026-02-07
SolarEdgeAPI "PV Power" is incorrect after going from Active to Idle


robk358:
The SolarEdgeAPI assigns the value of currentPower to "PV Power" when "active" (GetBatteryDetails())
if the PV Array status changes from Active to Idle, "PV Power" will contain the last received value, which is incorrect, because at "Idle", the currentPower is zero.

Checking if status is "Active"  or "Idle" solves this (SolarEdgeAPI.cpp, line 508)
Alternative solution: remove lines 507-509 and 512

gizmocuz:
Thank you for reporting!

I think it is better to use 0 when it's not Active? See the fix/commit here:

https://github.com/domoticz/domoticz/commit/e884256863a2fceb1c6e4cfc43d7668f191e8c2a

robk358:
One of the many options, just checked the API docs, "Disabled" is also one of the status values, this is now also handled.

gizmocuz:
Thanks for the feedback!

PR #6532, opened 2026-02-07
Fixed SolarEdge PV Power for Idle

robk358:
fix for issue #6531


robk358:
fixed in Commit e884256

PR #6541, opened 2026-02-10
Fix OpenZWave setpoint devices

UCIS:
It appears that, after [the original OpenZWave code](https://github.com/domoticz/domoticz/blob/e4a6cba30777ee70a5d473091d9769ae941ddae6/hardware/ZWaveBase.cpp#L903) was removed, [the names of some constants/definitions were changed](https://github.com/domoticz/domoticz/commit/3c85c85f25491eeebb7b0355780e2145b3dd9020#diff-8c13f76f87236ff537a02af181abdca582e3923dbcb04bb9f17f658e9052c6f7L47) [resulting in two very similar constants](https://github.com/domoticz/domoticz/blob/9eae2a6912437263a47792f626dc7d91fd8a21b8/hardware/hardwaretypes.h#L38-L51) being in use, and then when the OpenZWave code was restored [the wrong constant was used](https://github.com/domoticz/domoticz/blob/b0eadaa2ec6d6340df7eeba0e6e373b71ba393fc/hardware/ZWaveBase.cpp#L898) to identify packets for Setpoint type devices.

This commit changes the constant used back to the original value and fixes OpenZWave setpoint devices. I confirmed this works with the Horstmann SRT323 thermostat.

### DIFF
--- a/hardware/ZWaveBase.cpp
+++ b/hardware/ZWaveBase.cpp
@@ -732,7 +732,7 @@ void ZWaveBase::SendDevice2Domoticz(_tZWaveDevice* pDevice)
 	else if (pDevice->devType == ZDTYPE_SENSOR_SETPOINT)
 	{
 		_tSetpoint tmeter;
-		tmeter.subtype = sTypeSetPoint;
+		tmeter.subtype = sTypeSetpoint;
 		tmeter.id1 = ID1;
 		tmeter.id2 = ID2;
 		tmeter.id3 = ID3;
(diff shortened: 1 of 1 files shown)


gizmocuz:
Ohh my... good find! They look so similar! Thank you for this PR. (Aldo it is deprecated and you should move to ZWaveJS-UI!!!)

It seems pTypeSetpoint/sTypeThermTemperature is not used anywhere (maybe in the eventsystem), so this could be removed, and the best thing to do is to migrate all to use pTypeGeneral/sTypeSetPoint or pTypeThermostat6 eventually

UCIS:
Thanks! I know it's been deprecated, although I personally do not agree with the way this was done and the chosen migration path, I appreciate the continued development of Domoticz. I'm happy to contribute this fix and may be able to help out with future issues.

Issue #6593, opened 2026-02-21
SolarEdge API is polling 24/7 even when there is no battery attached


robk358:
**The Issue**
GetBatteryDetails() is called even when there is no battery connected. This causes unnecessary calls (24/7) to monitoringapi.solaredge.com.
The detection of a battery is based on the response of "Site Power Flow": empty-> no battery.
According to the API docs: In case the site does not support this information, the response should be an empty object. The site can support this information, even if there is no battery connected.

**Expected behavior**
Correct detection of connected batteries, to avoid unnecessary calls to monitoringapi.solaredge.com.

**Possible Implementation**
Use "Inventory" API call to detect batteries.

PR #6594, opened 2026-02-21
SolarEdge API Battery detection (see #6593)

robk358:
Detection of connected batteries for SolarEdgeAPI (see #6593) using "inventory" API call

### DIFF
--- a/hardware/SolarEdgeAPI.cpp
+++ b/hardware/SolarEdgeAPI.cpp
@@ -110,6 +110,7 @@ void SolarEdgeAPI::Do_Work()
 			{
 				if (!GetSite())
 					continue;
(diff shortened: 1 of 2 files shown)


gizmocuz:
Thank you for your PR. So now you introduced a second API call to GetBatteryFromInventory all the time...
Adding this API query all the time results in the same amount of API calls.

If you add this call only after do_work / worker started log message, then you solve the issue

robk358:
GetBatteryStatus is only called if m_SiteID == 0 (line 109), where m_SiteID is set within GetSite(), unless there is an error getting site data.
As far as i can see, it is called just once..

gizmocuz:
Sorry, that's correct! Thanks for the PR!

PR #6599, opened 2026-02-22
Add shared="true" plugin attribute for main interpreter execution

gizmocuz:
## Summary
- PyO3-based Python modules (e.g. `cryptography`, `jsonschema` via `rpds-py`) refuse to load in Python subinterpreters. Adding `shared="true"` to the plugin XML lets plugin authors opt into running in the main Python interpreter where these modules work.
- Shared plugins reuse the main interpreter via `PythonThread()` instead of `Py_NewInterpreter()`, with `sys.path` managed per-plugin so multiple shared plugins coexist correctly
- `RestoreThread()` updates `module_state->pPlugin` so `Domoticz.Log()` and other callbacks route to the correct shared plugin while the GIL is held
- Fixes a latent hang in `StopHardware()` when plugins fail to initialise (`m_bIsStopped` was never set, causing `Do_Work` loop to block indefinitely)
- Fixes early returns in `Stop()` that skipped flag cleanup, causing hangs when module state lookup fails

### Usage
Plugin authors add `shared="true"` to their plugin XML header:
```xml
<plugin key="MyPlugin" name="My Plugin" shared="true">
```

## Test plan
- [x] Shared plugin starts and loads PyO3 modules (cryptography, jsonschema) successfully
- [x] Multiple shared plugins run simultaneously with independent state and heartbeat counters
- [x] Stopping a shared plugin while others run does not crash or hang
- [x] Restarting a shared plugin loads the correct plugin.py (not another shared plugin's)
- [x] Non-shared plugins still work correctly (subinterpreter path unchanged)
- [x] Stopping and restarting non-shared plugins works as before
- [x] Plugin stop is instant (no 10-second delay from PythonThreadCount)

### DIFF
--- a/hardware/plugins/PluginManager.cpp
+++ b/hardware/plugins/PluginManager.cpp
@@ -415,6 +415,7 @@ namespace http {
 							ATTRIBUTE_VALUE(pXmlEle, "author", root[iPluginCnt]["author"]);
 							ATTRIBUTE_VALUE(pXmlEle, "wikilink", root[iPluginCnt]["wikiURL"]);
 							ATTRIBUTE_VALUE(pXmlEle, "externallink", root[iPluginCnt]["externalURL"]);
+							ATTRIBUTE_VALUE(pXmlEle, "shared", root[iPluginCnt]["shared"]);
 
 							TiXmlElement* pXmlDescNode = (TiXmlElement*)pXmlEle->FirstChild("description");
 							std::string		sDescription;
--- a/hardware/plugins/Plugins.cpp
+++ b/hardware/plugins/Plugins.cpp
@@ -753,11 +753,13 @@ namespace Plugins
 		, m_ImageDict(nullptr)
 		, m_SettingsDict(nullptr)
 		, m_bDebug(PDM_NONE)
+		, m_bShared(false)
 	{
 		m_HwdID = HwdID;
 		m_Name = sName;
 		m_bIsStarted = false;
 		m_bIsStarting = false;
+		m_bIsStopped = false;
 		m_bTracing = false;
 	}
 
@@ -1012,6 +1014,10 @@ namespace Plugins
 
 			Log(LOG_STATUS, "Stopping threads.");
 
+			// Ensure Do_Work loop can exit even if Stop() was never called
+			// (e.g. plugin failed to initialise or start)
+			m_bIsStopped = true;
+
 			if (m_thread)
 			{
 				m_thread->join();
@@ -1151,8 +1157,31 @@ namespace Plugins
 				}
 			}
 
+			// Parse 'shared' attribute early - needed to decide interpreter type
+			if (!sPluginXML.empty())
+			{
+				TiXmlDocument XmlDocEarly;
+				XmlDocEarly.Parse(sPluginXML.c_str());
+				if (!XmlDocEarly.Error())
+				{
+					TiXmlNode *pXmlNode = XmlDocEarly.FirstChild("plugin");
+					if (pXmlNode)
+					{
+						TiXmlElement *pXmlEle = pXmlNode->ToElement();
+						if (pXmlEle)
+						{
+							const char *pAttributeValue = pXmlEle->Attribute("shared");
+							if (pAttributeValue)
+							{
+								m_bShared = (std::string(pAttributeValue) == "true");
+							}
+						}
+					}
+				}
+			}
+
 			void* pPreserved = CPluginSystem::GetPreservedInterpreter(m_PluginKey);
-			if (pPreserved)
+			if (pPreserved && !m_bShared)
 			{
 				// Reuse preserved interpreter from previous run
 				m_PyInterpreter = (PyThreadState *)pPreserved;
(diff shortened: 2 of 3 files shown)


pipiche38:
Thanks for that work. From a performance stand point, they are some potential impact.
And if it is better to run all plugin in shared mode or not ..

Your view would be interesting

gizmocuz:
100% no!  The trade-off is that plugins using shared="true" share global state with each other (no isolation), so a misbehaving plugin could affect others

If it working?

DGA45:
Hi Gizmocuz,
I just had this PyO3 error message when trying this new plugin:
https://github.com/avriessen/domoticz_lg_thinqconnect_plugin/issues/4

I saw this issue and tried the shared="true" attribute, but now I have a fatal signal 11 (Segmentation fault) at domoticz startup.
```
2026-05-01 14:45:12.568  Error: Domoticz(pid:167931, tid:168257('_Clim XXX')) received fatal signal 11 (Segmentation fault)
2026-05-01 14:45:12.568  Error: siginfo address=0xffff00010101038c, address=0xffff9f8cba84
2026-05-01 14:45:12.591  Error: Failed to start gdb, will use backtrace() for printing stack frame

2026-05-01 14:45:12.593  Error: #0  /home/pi/domoticz/domoticz :  + 0x398c44 [0xaaaade2e8c44]
2026-05-01 14:45:12.593  Error: #1  /home/pi/domoticz/domoticz : signal_handler(int, siginfo_t*, void*) + 0x23c [0xaaaade2e9690]
2026-05-01 14:45:12.593  Error: #2  linux-vdso.so.1 : __kernel_rt_sigreturn + 0 [0xffffa1acc7ac]
2026-05-01 14:45:12.593  Error: #3  /lib/aarch64-linux-gnu/libpython3.13.so : _PyEval_EvalFrameDefault + 0x5dcc [0xffff9f8c87cc]
2026-05-01 14:45:12.593  Error: #4  /lib/aarch64-linux-gnu/libpython3.13.so :  + 0x10b01c [0xffff9f79b01c]
2026-05-01 14:45:12.593  Error: #5  /lib/aarch64-linux-gnu/libpython3.13.so : PyObject_CallNoArgs + 0x58 [0xffff9f799948]
2026-05-01 14:45:12.593  Error: #6  /lib/aarch64-linux-gnu/libpython3.13.so : _PyEval_EvalFrameDefault + 0x87f8 [0xffff9f8cb1f8]
2026-05-01 14:45:12.593  Error: #7  /lib/aarch64-linux-gnu/libpython3.13.so :  + 0x10a9d4 [0xffff9f79a9d4]
2026-05-01 14:45:12.593  Error: #8  /lib/aarch64-linux-gnu/libpython3.13.so : PyObject_CallMethodObjArgs + 0xa8 [0xffff9f79abd0]
2026-05-01 14:45:12.593  Error: #9  /lib/aarch64-linux-gnu/libpython3.13.so : PyImport_ImportModuleLevelObject + 0x2c4 [0xffff9f90e184]
2026-05-01 14:45:12.593  Error: #10 /lib/aarch64-linux-gnu/libpython3.13.so :  + 0x162518 [0xffff9f7f2518]
2026-05-01 14:45:12.593  Error: #11 /lib/aarch64-linux-gnu/libpython3.13.so :  + 0x109dfc [0xffff9f799dfc]
2026-05-01 14:45:12.593  Error: #12 /lib/aarch64-linux-gnu/libpython3.13.so : PyObject_CallFunction + 0x8c [0xffff9f799f50]
2026-05-01 14:45:12.593  Error: #13 /lib/aarch64-linux-gnu/libpython3.13.so : PyImport_Import + 0xd0 [0xffff9f90ee70]
2026-05-01 14:45:12.593  Error: #14 /lib/aarch64-linux-gnu/libpython3.13.so : PyImport_ImportModule + 0x20 [0xffff9f90ca0c]
2026-05-01 14:45:12.593  Error: #15 /home/pi/domoticz/domoticz : Plugins::CPlugin::Initialise() + 0x514 [0xaaaade835544]
2026-05-01 14:45:12.593  Error: #16 /home/pi/domoticz/domoticz : Plugins::CPlugin::Do_Work() + 0x1ec [0xaaaade8379dc]
2026-05-01 14:45:12.593  Error: #17 /home/pi/domoticz/domoticz :  + 0xbb373c [0xaaaadeb0373c]
2026-05-01 14:45:12.593  Error: #18 /lib/aarch64-linux-gnu/libc.so.6 :  + 0x85f38 [0xffffa0dd5f38]
2026-05-01 14:45:12.593  Error: #19 /lib/aarch64-linux-gnu/libc.so.6 :  + 0xede9c [0xffffa0e3de9c]
```

I am using the latest beta version:
2026-05-01 14:26:04.559  Status: Domoticz V2026.1 (build 17811)-beta (c)2012-2026 GizMoCuz
2026-05-01 14:26:04.559  Status: Build Hash: 8b0067bf5, Date: 2026-04-24 12:54:24

gizmocuz:
@DGA45 ,

The crash should be solved in beta 17879 (available within an hour)

Not sure if you need the shared option in your plugin

DGA45:
@gizmocuz, you are amazing ! This is solved.
The "shared" option is required because the LG thinQ python module imports openSSL (for mqtt), which uses this famous PyO3.
Thanks a lot and congratulations for your really great job with Domoticz !

DGA45:
@gizmocuz, 
Unfortunately, I just saw another issue after adding MQTT support to the LG thinQ plugin.
I use 5 plugins and the log displays mixed information (device id or hardware names).
I tried updating only one hardware pluging ("_Clim Haut") with python debug log enabled (see attached log).
The correct init sequence for this pluging should be:
1. _async_init_metadata
2. _async_init_MQTT device:myID1234
3. MQTT create MQTT client device:myID1234
4. connecting to MQTT server... device:myID1234
5. MQTT connection OK device:myID1234

Those steps are coming from "async/await" calls or MQTT callback (ex: "on_connect").

In the log we see that step 1 is correctly performed as "_Clim Haut", but after step 2 it is performed as "_Clim Chambre".
I think this may be due to the "_Clim Chambre" plugin having its "onHeartbeatCallback" in between, changing the main interpreter context, but not restoring it for the next "await" execution of the "_Clim Haut" plugin.

[log.txt](https://github.com/user-attachments/files/27482974/log.txt)

gizmocuz:
@DGA45 ,

I really can't do anything without plugin examples demonstrating the issue

DGA45:
@gizmocuz,
Ok, let me try doing a simple one.
Can you also tell me if there is some limitation using asyncio (coroutines) ? I did not find any example using it and I wonder if this may have some influence.

pipiche38:
No limitations of using Asunción. I have the z4d plugin which is doing an intensive use of Asuncion with zigpy libs.

gizmocuz:
@DGA45 , why do you need to use the 'shared' flag?
I assume a user wants this when they have multiple aircons ? Why not extend your plugin that they can enter multiple IP addresses of the aircons and your single plugin polls/controls them all?
Really nothing fancy there

DGA45:
@pipiche38 @gizmocuz I just found the problem. In "shared" mode, the Domoticz object is common and in CPlugin::RestoreThread, a context switch of pPlugin is done so that Domoticz.Log() is "routed to the correct plugin", meaning displaying the right plugin name for Domoticz.log.
This context switching is done only for events/callbacks managed by Domoticz but of course not for callbacks from other libraries/asyncio (ex: LG thinQ connect library).
Nevertheless, the plugins are correctly isolated so their execution is correct (from what I see), except when they use the Domoticz object, like log.
The solution using queue from the multi-threaded example could be used, which would call Domoticz.log only from the "managed" space.
I think this limitation should be written/explained somewhere.

gizmocuz:
@DGA45 , why do you need to use the 'shared' flag?
I assume a user wants this when they have multiple aircons ? Why not extend your plugin that they can enter multiple IP addresses of the aircons and your single plugin polls/controls them all?
Really nothing fancy there

DGA45:
@gizmocuz , I had the "ImportError: PyO3 modules do not yet support subinterpreters" error and I found this git merge where you added the "shared" flag exactly for this case.
The plugin (domoticz_lg_thinqconnect_plugin, which is not mine) is not dedicated to air conditioners only, LG thinQ supports many types of devices (washing machine, refrigerator, oven, fan, robots, air purifier, etc...).
I just wanted to do a few push requests in this plugin to be able to use it with multiple AC, since it is more recent and rely on the official LG library (my current solution being a quite old reverse-engineered one).
Your proposal could certainly be a solution but is a different approach from the initial plugin, which would be a big "push request" maybe not wanted by the plugin author (yes, I could fork and create my own, but if I have to spend more time developing, I have also other possible solutions). 
My last comment was only to contribute so that others who will use the "shared=true" option are aware of some limitation/design constraints.

PR #6628, opened 2026-03-04
Add Battery Voltage [V] display option to Energy Dashboard

bblaszkiewicz:
## Summary

Added support for displaying **Battery Voltage [V]** as an alternative to **Battery SOC [%]** in the Energy Dashboard. The existing SOC functionality is fully preserved.

## Changes

- **`setup.html`** — added a mode selector (`% SOC` / `V Voltage`) next to the Battery SOC device dropdown
- **`SetupController.js`** — added `General/Voltage` sensor type to the Battery SOC device list; added loading of the new `BatterySocMode` setting
- **`WebServerCmds.cpp`** — added reading and saving of the `BatterySocMode` parameter to `ESettings`
- **`EnergyDashboardController.js`** — added `idBattSocMode` and `fBattVolt` scope variables; updated `handleBattSoc()` to handle both modes; fixed `fBattSoc==100` guard to be mode-aware
- **`energy_dashboard.html`** — battery level bars and SOC text are hidden in voltage mode; voltage value is displayed instead

## How it works

The user selects a Voltage sensor in the existing *Battery SOC* dropdown and switches the new mode selector to `V (Voltage)`. The dashboard then displays the live voltage value (e.g. `48.3 V`) instead of percentage bars. Switching back to `% (SOC)` restores the original behaviour.

No database schema changes — `BatterySocMode` is stored as a new key inside the existing `ESettings` JSON preference. The GET endpoint (`Cmd_GetEnergyDashboardDevices`) required no changes as it returns the full `ESettings` JSON as-is.

### DIFF
--- a/main/WebServerCmds.cpp
+++ b/main/WebServerCmds.cpp
@@ -3852,8 +3852,7 @@ namespace http
 				int ESolar = atoi(request::findValue(&req, "ESolar").c_str());
 				int EBatteryWatt = atoi(request::findValue(&req, "EBatteryWatt").c_str());
 				int EBatterySoc = atoi(request::findValue(&req, "EBatterySoc").c_str());
-				std::string EBatterySocMode = request::findValue(&req, "EBatterySocMode");
-				if (EBatterySocMode != "volt") EBatterySocMode = "soc";
+				int EBatteryVolt = atoi(request::findValue(&req, "EBatteryVolt").c_str());
 				int ETextSensor = atoi(request::findValue(&req, "ETextSensor").c_str());
 				int EOutsideTempSensor = atoi(request::findValue(&req, "EOutsideTempSensor").c_str());
 				int EExtra1 = atoi(request::findValue(&req, "EExtra1").c_str());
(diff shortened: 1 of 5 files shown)


gizmocuz:
@bblaszkiewicz , thank you for this PR!

Maybe it would be better if the user 'can' also display the battery voltage and it will appear 'extra' next to the % (but much smaller font)
This will work if you have one battery, how about when you have 4?

Another option is to push everything via the influxDB pusher and create a Grafana dashboard with all your battery sensors/graphs (that's what I do, I can also ready out the BMS)

gizmocuz:
@bblaszkiewicz , thank you for your changes. Did you test this?

gizmocuz:
There was a error that caused the voltage sensors not go to the voltage dropdown but to the battery soc.
Thank you for the PR, I made a small correction and it is available in beta 17319

bblaszkiewicz:
@gizmocuz Thanks for the info and the bug fix. I didn't have time to test these changes yesterday.

gizmocuz:
Ok, thank you, but next time test before checking in the work please ;)

PR #6633, opened 2026-03-07
Identify FAN type in LOG

JurgenLB:
For knowing witch type off FAN a user is using,
I added a LOG-line.

Kind Regards


gizmocuz:
Thank you!

PR #6635, opened 2026-03-08
MSCV Warnings

JurgenLB:
My problem was about msbuild\Windows Libraries\include\jwt-cpp\ — this is a separate copy of jwt-cpp that's bundled in the Windows libraries archive. The libwebem project uses this path, not the extern/jwtcpp submodule.
The Release|Win32 configuration includes both ../extern/jwtcpp/include and ../msbuild/Windows Libraries/include,
The Windows Libraries bundle version of jwt-cpp, these fixes haven't been applied yet. The bundle is downloaded from https://raw.githubusercontent.com/domoticz/win32-libraries/master/WindowsLibraries.7z.
putting extern/jwtcpp/include before msbuild/Windows Libraries/include in the search path would make libwebem use the fixed jwt-cpp headers instead of the older ones in the Windows Libraries bundle.

For NoticficationFCM;
std::chrono::system_clock::now() was called twice — iat and exp JWT claims were based on slightly different timestamps.

Kind Regards


gizmocuz:
@JurgenLB , we are using the official jwt-cpp library code. For Windows this is taken from vcpkg.
If the official library code has been packed, and also updated in vcpkg, please make a new issue mentioning this and I will update this package.
In the mean time, warnings are 'not' errors and it does not matter. If you want you can add pragma's around the calling functions/headers to mask them as well.

PR #6646, opened 2026-03-14
updated "my profile page"

skywalker1979:
Refreshed the style of the "My Profile" page.

### DIFF
--- a/main/WebServerCmds.cpp
+++ b/main/WebServerCmds.cpp
@@ -2704,6 +2704,10 @@ namespace http
 			root["title"] = "GetActualHistory";
 
 			std::string historyfile = szUserDataFolder + "History.txt";
+			if (szStartupFolder != szUserDataFolder)
+			{
+				historyfile = szStartupFolder + "History.txt";
+			}
 
 			std::ifstream infile;
 			int ii = 0;
--- a/www/views/myprofile.html
+++ b/www/views/myprofile.html
@@ -1,136 +1,176 @@
-<div id="myprofilecontent" class="container">
-	<div class="page-header-small">
-		<h1 data-i18n="My Profile">My Profile</h1><span></span>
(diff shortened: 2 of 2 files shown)


skywalker1979:
feel free to refuse the PR if it's not up to standard.

skywalker1979:
The second part is to fix a problem in docker, where clicking on the top of the screen link regarding the release notes does not work.

gizmocuz:
Please add a screenshot of the new look... I thought I made it already pretty OK compared to the version in the current stable release?

gizmocuz:
Please remove the 'docker' history fix and make a separate PR please

skywalker1979:
image
here is the screenshot.

gizmocuz:
Okey, let's close this PR for now. First we let users get used to this (new stable), next it is time to restyle more.
Maybe best to let me continue doing this to keep it in line. You are now centering everything, while the whole application aligns it nicely left.
The Update button is higher then the cancel and is on a really strange position, it should be at the bottom.
The labels for the passwords also are not on one line
Did you see the room it takes now for these little settings? a whole page?

Did you use AI? Please do not do this.

The fix for the history seems valid, please create a PR for this..

Thanks for the PR and time!

skywalker1979:
well..the button was higher because my mouse was on it.... no problem anyway

skywalker1979:
also... i placed it above the passkeys area because the passkey area does not use update or cancel.... seemed clearer that way.

gizmocuz:
Ahh that indeed makes a difference, did not see the mouse cursor... Nothing will actually be updated/changes in the profile page.. maybe sometimes you change your password/passkey. 
There are lot of pages to redesign. But not at the present. I worked for 3 weeks on a giant angular patch that has now been merged among server fixes and it's time for a new stable first

PR #6649, opened 2026-03-14
feat: Extended plugin settings with new input types, groups, and conditional visibility

### LINKED ISSUE #6644: Extend Python Plugin Settings Beyond Mode1-Mode6 [willing to implement]
### Prerequisites

- [x] I have searched existing issues for this feature request

### Description

Python plugins are currently limited to 6 settings (Mode1-Mode6) with basic input types (text, password, select, textarea). Plugin authors who need more configuration options have to work around this by overloading fields or managing settings programmatically via `Configuration()`.

This feature would add support for **unlimited, flexibly-named plugin settings** with richer input types, while maintaining full backward compatibility with existing plugins.

I have a detailed design ready and am happy to implement this and submit a PR if the approach looks good.

### Use Case

For my own [Luxtronik plugin](https://github.com/Rouzax/luxtronik-domoticz-plugin-v2), I'm running into this limitation — there are more configuration options I'd like to expose to users than 6 Mode fields allow, and the workarounds are clunky.

More broadly, any plugin with a non-trivial configuration surface is forced to either cram settings into Mode1-Mode6 with cryptic labels, or implement custom configuration management via `Configuration()` without any auto-generated UI.

### Storage

A new `PluginSettings TEXT` column in the Hardware table stores custom settings as a flat JSON object. Plugin XML `<param>` fields with reserved names (Mode1-6, Address, Port, etc.) continue to route to their existing columns — **zero breakage** for existing plugins. Any other field name goes to the new column and appears in the `Parameters` dict alongside Mode1-6.

### New input types

In addition to existing text/password/select/textarea:
- `type="number"` — with optional min/max/step
- `type="boolean"` — checkbox, stored as `"true"`/`"false"`
- `type="slider"` — jQuery UI slider with min/max

### Structural improvements

- `<group label="...">` — collapsible sections for organizing related settings
- `visible_when="field=value"` — conditional visibility (e.g., show certificate field only when protocol is HTTPS)

### Example plugin XML

```xml
<params>
    <param field="Address" label="IP Address" required="true"/>
    <param field="Port" label="Port" type="number" default="80"/>

    <group label="Polling">
        <param field="UpdateInterval" type="number" min="5" max="3600" default="30" label="Update Interval (s)"/>
        <param field="RetryCount" type="number" min="0" max="10" default="3" label="Retries"/>
    </group>

    <group label="Advanced">
        <param field="EnableDebug" type="boolean" default="false" label="Debug Logging"/>
        <param field="Protocol" type="select" label="Protocol">
            <options>
                <option label="HTTP" value="http"/>
                <option label="HTTPS" value="https"/>
            </options>
        </param>
        <param field="Certificate" type="text" visible_when="Protocol=https" label="Certificate Path"/>
    </group>
</params>
```

### Python access

Plugin authors access new settings the same way as Mode1-Mode6 — no new API to learn:

```python
def onStart(self):
    # Old-style still works
    debug = Parameters["Mode6"]

    # New custom fields appear alongside them
    interval = Parameters["UpdateInterval"]
    retries = Parameters["RetryCount"]
    enabled = Parameters["EnableDebug"]       # "true" or "false"
```

### What stays the same

- Mode1-Mode6 work identically for existing plugins
- `Configuration()` is unaffected (separate column)
- All values are strings in the `Parameters` dict
- No new Python API to learn

### Long-term benefit

Over time, plugins could migrate from Mode1-Mode6 to descriptive field names (e.g., `UpdateInterval` instead of `Mode3`), improving readability. The Mode columns themselves remain for C++ hardware types.

### Files affected

`SQLHelper.cpp`, `WebServerCmds.cpp`, `Plugins.cpp`, `PluginManager.cpp`, `Hardware.js`, `Hardware.html`, `mainworker.cpp`

### Alternatives Considered

1. **Adding more Mode columns (Mode7-Mode12):** Simple but doesn't solve the flexibility problem — still fixed-name fields, no support for richer types or grouping, and requires ~20 code changes per additional column across 6 files.

2. **Reusing the existing `Configuration` column with namespacing:** Would avoid a new DB column, but risks collision with plugins already using `Configuration()` for runtime state, and mixes two concerns in one column.

### Forum Discussion

_No response_

Rouzax:
## Summary

Adds unlimited, flexibly-named plugin settings with auto-generated UI, new input types, grouping, and conditional visibility to the Python plugin system.

Closes #6644

## Problem

Python plugins are limited to 6 hardcoded settings (Mode1–Mode6) with only basic input types (text, password, select, textarea). Plugin authors need more settings with descriptive names and richer input types. The current system requires touching ~20 code locations across 6 files to add a single new Mode field.

## Solution

A new `Settings` TEXT column in the Hardware table stores custom plugin settings as flat JSON. Reserved field names (`Mode1`–`Mode6`, `Address`, `Port`, `SerialPort`, `Username`, `Password`, `Extra`) continue routing to their existing columns. The plugin XML `<param>` system is extended with new attributes, and the frontend dynamically renders and collects all fields. Python plugins access everything through the existing `Parameters` dict — no new API.

### New input types

| Type | XML | Renders as |
|------|-----|------------|
| `number` | `type="number" min="1" max="100" step="5"` | HTML5 number spinner |
| `boolean` | `type="boolean"` | Toggle switch (styled like Enabled/Log Level) |
| `slider` | `type="slider" min="0" max="100"` | jQuery UI slider with value display |

### Grouping

`<group label="Advanced">` wraps related params into collapsible sections with chevron toggle icons, matching the existing Domoticz UI patterns.

### Conditional visibility

`visible_when="field=value"` hides/shows a param based on another field's value. Hidden fields are excluded from the Settings JSON.

### Example plugin XML

```xml
<params>
    <param field="Address" label="IP" required="true" default="127.0.0.1"/>
    <param field="Interval" type="number" label="Poll Interval" min="5" max="3600" default="30"/>
    <param field="EnableDebug" type="boolean" label="Debug Mode" default="false"/>
    <param field="Brightness" type="slider" label="Brightness" min="0" max="100" default="50"/>
    <param field="Protocol" label="Protocol">
        <options>
            <option label="HTTP" value="http" default="true"/>
            <option label="HTTPS" value="https"/>
        </options>
    </param>
    <param field="Certificate" label="Cert Path" visible_when="Protocol=https"/>
    <group label="Advanced Settings">
        <param field="RetryCount" type="number" label="Retries" min="0" max="10" default="3"/>
        <param field="EnableNotifications" type="boolean" label="Notifications" default="true"/>
    </group>
    <param field="Mode6" label="Legacy Debug">
        <options><option label="None" value="0" default="true"/></options>
    </param>
</params>
```

### Python access (unchanged)

```python
def onStart(self):
    interval = Parameters["Interval"]          # "30"
    debug = Parameters["EnableDebug"]           # "true" or "false"
    legacy = Parameters["Mode6"]                # still works
```

### Database (`main/SQLHelper.cpp`)
- Add `Settings TEXT DEFAULT ''` column to Hardware table
- Bump `DB_VERSION` to 177
- Add `ALTER TABLE` migration for existing databases

### Backend API (`main/WebServerCmds.cpp`)
- Extract and validate Settings JSON in `Cmd_AddHardware` and `Cmd_UpdateHardware` (64KB max, 4KB per value)
- Include Settings in plugin INSERT/UPDATE SQL queries
- Return Settings as parsed JSON object in `Cmd_GetHardware`
- Exclude password-type field values from API responses (uses plugin XML manifest lookup)

### Plugin Python layer (`hardware/plugins/Plugins.cpp`)
- Add `IsReservedFieldName()` helper for routing reserved names to existing columns
- Extend SELECT query to include Settings
- Load Settings JSON into Python `Parameters` dict
- Apply XML defaults for fields missing from stored Settings
- Clean up orphaned keys no longer defined in plugin XML

### XML parsing (`hardware/plugins/PluginManager.cpp`)
- Parse new `<param>` attributes: `type`, `min`, `max`, `step`, `visible_when`
- Support `<group label="...">` elements wrapping child `<param>` elements
- Refactored with `parseParam` lambda to avoid code duplication

### Frontend (`www/app/hardware/Hardware.js`)
- New `InitPluginWidgets()` function for deferred widget initialization (fixes Angular.js DOM duplication)
- Render `number`, `boolean` (toggle switch), and `slider` (jQuery UI) input types
- Collapsible group sections with chevron icons
- Conditional visibility via `visible_when` with change event handlers
- Collect non-reserved field values into Settings JSON on save (both add and update)
- Restore Settings values on edit (checkboxes, sliders, regular inputs)

## Backward compatibility

- Existing plugins using Mode1–Mode6 work identically without modification
- Reserved field names route to existing columns (case-insensitive)
- The `Configuration` column and `Extra` column are untouched
- The `Settings` column name is intentionally generic for future C++ hardware adoption

## Security

- Server-side JSON validation before storage (`ParseJSon`, size limits)
- Parameterized queries (`%q`) — existing pattern
- Password-type fields excluded from API responses
- Client-side `encodeURIComponent()` on all values — existing pattern

### API tests (6 pass)
- Add hardware with Settings JSON — round-trip verified
- Update Settings — persistence verified
- Empty Settings accepted
- Settings returned as JSON object
- Malformed JSON rejected
- Mode1–Mode6 backward compatibility

### Playwright UI tests (6 pass)
- Plugin appears in dropdown
- All field types render (number, boolean, slider, select, password, text)
- Conditional visibility (Certificate shows for HTTPS, hides for HTTP)
- Collapsible group expand/collapse
- Number input min/max/step constraints
- Save and restore round-trip via API

### Migration tests
- Fresh install: Settings column exists
- Upgrade from v176: ALTER TABLE migration runs, DB_VERSION = 177

## Test plan

- [ ] Build passes (`make -j$(nproc)`)
- [ ] Fresh install creates Settings column
- [ ] Existing database upgrades cleanly (v176 → v177)
- [ ] Existing Mode1–Mode6 plugins work unchanged
- [ ] New plugin with all input types: values save and restore
- [ ] Slider is interactive
- [ ] Conditional visibility toggles correctly
- [ ] Groups collapse/expand
- [ ] Password fields not exposed in API response
- [ ] UI works across default and dark themes

## Design document

See `docs/plans/2026-03-14-plugin-settings-design.md` for full design rationale.

---

🤖 Generated with [Claude Code](https://claude.com/claude-code)


gizmocuz:
## Code Review: PR #6649 - Extended Plugin Settings

### Critical Issues

**1. XML re-parsed on every plugin start** (`hardware/plugins/Plugins.cpp` ~1660-1689)
Every plugin load iterates all plugin XMLs, does a linear string search, and re-parses the XML. This is wasteful and could cause startup hangs with many plugins. Should be cached at plugin registration time.

**2. Unnecessary DB write on every plugin start** (`Plugins.cpp` ~1768-1770)
Settings are written back to the database on every plugin load even if nothing changed. Should only write if corruption is actually detected.

**3. SQL parameter mismatch in UPDATE Hardware** (`main/WebServerCmds.cpp` ~1824)
`WHERE (ID == '%q')` — ID should be numeric (`%d`), not a quoted string. Risk of SQL syntax error or lookup failure on hardware updates.

---

### Warnings

**4.** Duplicated Settings validation code in `Cmd_AddHardware` and `Cmd_UpdateHardware` — extract to a shared helper function.

**5.** Inefficient plugin XML lookup in `Cmd_GetHardware` — repeated linear search + full XML re-parse on every API call to strip password fields.

**6.** Missing `field` attribute on a `<param>` element is silently skipped — should log a warning about malformed plugin XML.

**7.** No try-catch around `JSON.stringify` in `Hardware.js` (~line 272, ~1789).

**8.** Slider initialization not idempotent — `$slider.empty()` destroys widget state but the existing guard prevents reinitialization on a second call to `InitPluginWidgets()`.

**9.** Group rendering HTML nesting bug (`Hardware.js` ~5906) — ungrouped → grouped → ungrouped param sequence produces unbalanced HTML, which may break CSS/JS selectors.

---

### What's Done Well
- Backward compatibility via `IsReservedFieldName()` keeps Mode1-Mode6 plugins working unchanged
- Password field stripping from API responses
- Input validation with size limits (64KB total, 4KB per value)
- Comprehensive test coverage (Gherkin + Playwright + migration tests)

Rouzax:
Thanks for the detailed review! Addressing each item:

### Accepted — will fix

**1. XML re-parsed on every plugin start:** Agreed. Will cache the XML string in a `m_PluginXML` member variable, set during `Initialise()`, reused in `Start()`.

**4. Duplicated validation:** Agreed. Will extract `ValidateSettingsJSON()` helper.

**5. Inefficient XML lookup in GetHardware:** Agreed. Will pre-build a password field cache before the hardware loop instead of per-entry XML parsing.

**6. Missing field warning:** Agreed. Will add `LOG_STATUS` warning.

**7. JSON.stringify guard:** Will add defensive fallback.

### Investigated — not a bug

**2. DB write on every plugin start:** `bSettingsChanged` is only set `true` when orphaned keys are removed (line 1751) or the check at line 1757 finds a field not in `settingsJson`. However, defaults are added to `settingsJson` at lines 1705-1728 *before* the cleanup loop, so `settingsJson.isMember(field)` is always `true` for all XML fields — the check at 1757 never triggers on normal starts. On first start with empty Settings, defaults are added in step 1, making step 3's check false. The DB write only occurs when a plugin author removes a param from their XML (orphan cleanup).

**3. SQL `WHERE (ID == '%q')`:** This is the existing codebase pattern — every UPDATE in `WebServerCmds.cpp` uses `WHERE (ID == '%q')` with `idx.c_str()` (lines 1809, 1816, 1833, 1844, 2245, 2262, 4110, 4770, etc.). SQLite handles quoted integers. Changing to `%d` would be inconsistent with the rest of the file.

**8. Slider `$slider.empty()` idempotency:** The `$slider.data("ui-slider")` guard at line 5205 returns early if the widget is already initialized. `$slider.empty()` at line 5214 only executes on first initialization — it clears stale DOM elements left by Angular.js DOM cloning. A second call to `InitPluginWidgets()` on the same table is a no-op.

### Investigated — not a bug (but adding test to verify)

**9. Group nesting:** Traced the HTML output for ungrouped→grouped→ungrouped sequences. The close tag `</table></div></td></tr>` at line 5908 properly closes the inner table, div, td, and tr — returning to the outer `<table>`. Subsequent ungrouped params render as `<tr>` in the outer table. The test plugin already has this pattern (Address → Advanced Settings group → Mode6). Will add a Playwright assertion to verify.

### Additional fix

**Slider value display:** The value `<span>` next to the slider is not clearly visible. Will improve styling.

gizmocuz:
Can you contact me on our Slack channel please

gizmocuz:
**Re: Item #2 — this is still a bug**

Your explanation is correct that the \"Check if defaults were added\" loop never triggers — but that's exactly the problem: it's dead code.

Here's the actual flow:

1. Step 1: For each XML field not in `settingsJson` → `settingsJson[pField] = defaultVal` ← defaults are added here
2. `bSettingsChanged = false`
3. Cleanup loop builds `cleanedSettings`. Orphans set `bSettingsChanged = true`.
4. \"Check if defaults were added\" loop: checks `!settingsJson.isMember(field)` — but step 1 already added all XML fields, so this is **always false**. Dead code.
5. `settingsJson = cleanedSettings`
6. DB write only if `bSettingsChanged` — which is only `true` for orphan cleanup.

**Concrete bug scenario — fresh plugin install:**
- DB has empty Settings
- Step 1 populates `settingsJson` with all XML defaults (`Interval=30`, `EnableDebug=false`, etc.)
- No orphans → `bSettingsChanged` stays `false`
- DB write skipped → DB still has **empty Settings**

Python `Parameters` dict gets the correct defaults (good), but the DB is never updated. When the user opens the Edit Hardware dialog, all custom fields show **blank** instead of the XML defaults. If they save without touching anything, blank values get persisted — wiping out the defaults on next start too.

**Fix — one line, in the first loop:**

```cpp
if (!settingsJson.isMember(pField))
{
    const char *pDefault = pEle->Attribute("default");
    std::string defaultVal = pDefault ? pDefault : "";
    settingsJson[pField] = defaultVal;
    bSettingsChanged = true;  // ← add this (same pattern in the group loop below)
}
```

Apply the same `bSettingsChanged = true` in the group child loop too. After that, the \"Check if defaults were added\" block is dead code and can be removed entirely.

Rouzax:
## Updates since review

All review feedback addressed:

**Fixes:**
- **Item 1:** Cached plugin XML in `m_PluginXML` member — eliminates redundant manifest lookup on every plugin start
- **Item 2:** Fixed defaults persistence bug — `bSettingsChanged` now set when XML defaults are added, removed dead code loop (you were right, my pushback was wrong)
- **Item 4:** Extracted `ValidateSettingsJSON()` helper — eliminates duplicated validation
- **Item 5:** Pre-built password field cache in `GetHardware` — O(n+m) instead of O(n*m)
- **Item 6:** Log warning for missing field attribute (LOG_NORM to avoid noise from legitimate separator params)
- **Item 7:** Defensive `|| {}` guard on `JSON.stringify`
- **Item 9:** Verified HTML nesting is correct — added test coverage

**Additional fixes found during testing:**
- Slider value display: wrapped in `position:relative` container so value is visible
- Slider/checkbox restore on edit: fetch Settings from API (DataTable row doesn't include Settings)
- Conditional visibility re-evaluated after restoring saved values
- `autocomplete="off"` on plugin form inputs
- Extracted `CollectPluginSettings()` JS helper (DRY)
- `m_PluginXML` moved to private scope

**Test coverage added:**
- 7 Playwright edge case tests (collapsed groups, conditional visibility, slider restore, update, delete, autocomplete)
- 6 integration lifecycle tests (fresh defaults, value preservation, field addition/removal, orphan cleanup)
- All 13 Playwright + 6 lifecycle + 6 API tests pass

Rouzax:
[TestExtendedSettings.zip](https://github.com/user-attachments/files/26003650/TestExtendedSettings.zip)
The test plugin

Rouzax:
Updated wiki for the extended settings
[domoticz wiki - revised.txt](https://github.com/user-attachments/files/26003668/domoticz.wiki.-.revised.txt)

waltervl:
Perhaps wait for this implementation after the next stable release (hopefully coming soon?) as it is a major new functional implementation?

gizmocuz:
@waltervl , it would be good to have this in the stable, plugins not using it won't see any difference, but plugin developers could make a use of it and users do not have to update to the beta...
It's not a big change

I just needed to tweak the new dashboard a little bit as there where some inconsistencies with the previous look.... I want to wait a few more ways for the new stable... maybe the weekend... but all looks ok

gizmocuz:
Thank you for this PR. I merged your PR and one more change via another branch

waltervl:
> Updated wiki for the extended settings [domoticz wiki - revised.txt](https://github.com/user-attachments/files/26003668/domoticz.wiki.-.revised.txt)

Thanks, Wiki is updated

PR #6651, opened 2026-03-14
Release notes path in docker is broken

skywalker1979:
The release notes path in Docker is broken: 
the page does not show anything because History.txt resides in wrong path and domoticz only looks in szUserDataFolder (which differs in Docker due to the -userdata flag). 

The fix applies a similar logic to what has been already done in OpenZWave.cpp.


gizmocuz:

PR #6653, opened 2026-03-15
Energy view last 24h patch

guillaumezin:
This patch extracts the most recent data, ideally from current time, otherwise (as fallback) we go back to yesterday at midnight max for pTypeP1Power device when missing data (for managed counter for instance) to prevent this result:

image

### DIFF
--- a/main/WebServerHandleGraph.cpp
+++ b/main/WebServerHandleGraph.cpp
@@ -58,21 +58,10 @@ namespace http
 			if (srange.empty() && sgroupby.empty())
 				return;
 
-			result = m_sql.safe_query("SELECT MAX(Date) FROM %s WHERE DeviceRowID==%" PRIu64, dbasetable.c_str(), idx);
-
 			time_t now = mytime(nullptr);
 			struct tm tm1;
-
-			if (!result.empty() && !result[0][0].empty())
-			{
-				ParseSQLdatetime(now, tm1, result[0][0], -1);
-			}
-			else
-			{
-				now = time(nullptr);
-				localtime_r(&now, &tm1);
-			}
-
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thank you for this PR. I am not understanding it correctly. If there is no data before 19.00pm, we should not display it?

guillaumezin:
Imagine it's 6pm (18:00). As done currently, Domoticz will try to show data from yesterday 19:00 to today 18:00. But there are data only from 19:00 yesterday to 23:00 in database for now. It's what is shown on the image above. The patch changes the logic and chooses instead as end date the date of the most recent data in database (but limits the search down to yesterday 00:00) down to 24h sooner, this way you have more chances to see a complete graph.

guillaumezin:
This is the result with path applied:

image

gizmocuz:
Thank you for taking the time to create this PR and for the clear explanation and screenshots!

After careful consideration, we've decided not to merge this change. The core concern is that it alters the semantics of the "last 24h" graph for all users:

- **Current behavior:** Always shows the window `[now − 24h, now]` — gaps are visible and honest.
- **Patched behavior:** Shows `[lastData − 24h, lastData]` — the window silently shifts to older data without any indication to the user.

A gap in the graph is the correct signal that data collection has a problem. Hiding it by shifting the time window would be misleading — users would think they're seeing current data when they're not.

If this is something you'd like to improve further, a better approach might be to show a visual indicator on the graph (e.g., "last data received at 23:00") so the user is informed without the time window being silently moved.

Thanks again for your contribution!

guillaumezin:
Ok, I understand, I will think about it and try to improve my contribution.

PR #6693, opened 2026-03-26
Add Daikin Heat pump devices (via Modbus TCP)

adrighem:
Daikin has a nice system called the HomeHub. This is meant to manage Daikin Heatpumps via modbus.
This plugin intends to implement management of a Daikin Altherma 3 (and daikin air conditioner) via homehub.
Fully local, no internet connection required.

Please provide feedback, or merge if you're happy. :-D

I also found some warning during the build and added a fix for them. Not sure if I should put that into the same pull request... but the intention is good. :-D

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -236,6 +236,7 @@ hardware/CurrentCostMeterBase.cpp
 hardware/CurrentCostMeterSerial.cpp
 hardware/CurrentCostMeterTCP.cpp
 hardware/Daikin.cpp
+hardware/DaikinModbus.cpp
 hardware/DarkSky.cpp
 hardware/DavisLoggerSerial.cpp
 hardware/DenkoviDevices.cpp
@@ -472,7 +473,7 @@ ELSE(LUA_LIBRARIES AND LUA_INCLUDE_DIRS)
   IF(LUA_FOUND)
     MESSAGE(STATUS "LUA library found at  : ${LUA_LIBRARIES}")
     MESSAGE(STATUS "LUA includes found at : ${LUA_INCLUDE_DIR}")
-    INCLUDE_DIRECTORIES(${LUA_INCLUDE_DIR}) 
+    INCLUDE_DIRECTORIES(${LUA_INCLUDE_DIR})
   ELSE(LUA_FOUND)
     MESSAGE(FATAL_ERROR "LUA 5.3 not found! use sudo apt-get install liblua5.3-dev")
   ENDIF(LUA_FOUND)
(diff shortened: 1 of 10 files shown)


gizmocuz:
@adrighem , thank you very much for this PR!

Please do not include any other changes in this PR , I have no warnings at all (maybe one or 2), so maybe it is your compiler.

adrighem:
I removed the compiler warning changes. It's a fairly pure PR now. Any comments?

gizmocuz:
Thank you for the changes!

I will be checking the PR shortly, but I have to comments 🫣

we have debug logs with DEBUG_HARDWARE, this initial log is probably one of those?

If it is intended for the end user, it should be placed in

bool CDaikinModbus::StartHardware()

(also the debug log)

_log.Log(LOG_STATUS, "DaikinModbus: Initializing (ID: %d, IP: %s, Port: %d, Poll: %d, Unit: %d, Model: %s)",
		m_HwdID, m_szIPAddress.c_str(), m_usIPPort, m_iPollInterval, m_iUnitID, m_bIsAirToAir ? "Air-to-Air" : "Altherma 3");

And the hardware plugin should not call **_log.Log**

It should call just **Log**

you also do not need to specify "DaikinModbus: ", it will use the hardware name

gizmocuz:
Thanks again for the changes. All looks perfect except two things:

- Please change HTYPE_FREE2USE to HTYPE_DaikinModbus (and delete it from the bottom of the list)
This does mean current your devices belonging to the hardware has to change the hardware id from 129 to 26
This is a simple database patch (SQLite Expert Personal Edition)

- You did not add the files to the Visual Studio project (the most important of them all 🫣😁).
I can do this for you after merging, no problem, unless you want to do this for yourself. They should go under the Devices filter in DaikinModbus (new)
It is a risky job so better leave it if you are fine with that

adrighem:
Thanks for the feedback.
* HTYPE adjusted
* I'll leave the Visual Studio project adjustments to the expert. Sounds like a job that's too important to mess up. :wink:

adrighem:
Please do not pull yet. I just noticed that the setpoints are scaled wrong. Let me check.

gizmocuz:
No worries, take your time and tell me when you want it to be merged

adrighem:
Okay, this looks better. It's merge ready now.

gizmocuz (changes requested):
Thanks for the contribution! The driver looks well-structured overall, but there is one issue that needs to be fixed before merging:

**Blocking issue:**

**`UpdateUsageSensor` declared but never implemented** — The header declares `void UpdateUsageSensor(int ChildID, float usage, const std::string& szDefaultName);` but there is no implementation in the .cpp file. This will cause a linker error. Either implement it or remove the declaration.

Once this is fixed, this looks good to go!

gizmocuz:
We will take care of the remaining fixes ourselves — the two blocking issues and the Visual Studio project file integration will be handled on our end. We'll merge the PR once everything is in order. Thanks again for the contribution\!

adrighem:
Sorry for the left over UpdateUsageSensor declaration. Thanks for merging.

PR #6697, opened 2026-03-26
Energy view last 24h patch

guillaumezin:
This is a new try after https://github.com/domoticz/domoticz/pull/6653

This version adds a framework to translate and show a warning message (extracted from root["warningmessage"] json data) or error message (root["errormessage"]) on graph if defined in received json data.

CaptureWarningEnergy

The look of messages can be customized using css .chart-watermark-error .wm-message and .chart-watermark-warning .wm-message


gizmocuz:
Thank you for this PR. Could you remove the 2 translation files updates, and add the string(s) to be translated in the file www/i18n/to_be_translated.json

Before you press the 'Log' button, the widget already displays the last-seen, so you know when it is outdated? And also the widgets color is red
That are two indicators data is not updated already, is there really a need for a third?

For the warning message, how about:

- No recent data available
- Data is outdated - results may not reflect current conditions
- No live data available. Showing the most recent data from [date/time]

guillaumezin:
Some plugin can force pPlugin->HasNodeFailed to never be true, because they manage batch data and retrieve them only by night. But I agree that usually it means the user is looking at a log of a device known to retrieve data of the day before, the warning message is maybe too much.

The framework I made to add warning or error message on charts is still worth keeping I think, you tell me.

I updated the translation.

gizmocuz:
How about just a subtile warning glymp icon? glyphicon glyphicon-warning-sign (48px red or yellow, with a tooltip with the warning message) overlayed on the chart?

If a plugin forces HasNodeFailed to false, then it should not matter, and the warning should not be displayed?

guillaumezin:
Result:

CaptureWarningEnergy2

gizmocuz:
Thank you for the change! I love the icon!... But it is not the right aprouch... I have been thinking about this for 2 days now...

I am currently working (all day) at the charts. Specifically cleaning up when something goes wrong because of counter loops or missing days.
Also I changed the chart rendering behavior, and this also solved your issue i think/hope.

Charts are now always aligned on the right on the current day/hour. In the chart above, if it was for instance 03.00am the next morning, there is now a red plotband visible between 00:00 and 03:00am, In case you miss a few days, in the shortlog/day/year graph you can see this directly with the plot band.

If you miss 2 days, and the shortlog was set to 3 days, you only see one day, followed by 2 days of empty (red plotbands indicating missing data)

It will be available in the upcoming beta in I hope an hour, but for sure tomorrow

gizmocuz:
Beta 17574 is available within the hour

guillaumezin:
Hi,

I see, but in my scenario, this is not completely satisfactory:

image

Time here is 20:20, and the data are grabbed in the morning for yesterday 00:00 to 23:59. Now without patch, I can see only data from 21:00 to 23:00 on Usage / Hour view. And I don't see current day/hour alignment filled with the red plot-band (but I see it on the other views).

Smanar:
In fact, the patch allows you to have a 24-hour view, but not necessarily the 24 hours preceding the current time (for device that don't working in real time)

gizmocuz:
it should be 24 hours from 'now' that was always the idea. 

The title of the chart also mentions 'Last 48 hours', so you should see the last 48 hours, now data that is older than 48 hours

Smanar:
@gizmocuz I m agree, but for me this patch is more an improvement.

There is some data collected by third organisation, like the electricity, or soon at my home for the water (the "water organisation is installing connected counter for water), and they rarely purpose a real time API.
But I have see one for photo voltaic too.

I haven't read the code but I think this patch don't impact other devices and permit something usefull for "not real time counter".

It can be usefull to see on one whole day the evolution, to see for example when the consumption is the highter, even it's more "Graph during 24h" then "Graph during the last 24h since now".

gizmocuz:
The above data won't be displayed anymore when it is older then 24 hours, therefor it does not make sense to show a icon. It is already indicated with a red plotband

PR #6702, opened 2026-03-27
Added PLN (zł) currency option

voyo:
Add Polish Zloty (zł) to the currency selector in the settings view.
Poland is one of the larger EU countries with a significant Domoticz user base, but PLN was missing from the currency list.


gizmocuz:
Dziękuję !

gizmocuz:
If you have time, maybe you could help with the translations

https://app.transifex.com/domoticz/domoticz/dashboard/

PR #6707, opened 2026-03-28
DaikinModbus: Fixed usage sensors and port display

adrighem:
- Fix: Changed power sensors to use proper Usage electric sensors instead of custom sensors. Error log will show if the Max Power limit is hit.
- Fix: Added missing port field in the hardware overview table.

### DIFF
--- a/hardware/DaikinModbus.cpp
+++ b/hardware/DaikinModbus.cpp
@@ -317,38 +317,38 @@ void CDaikinModbus::ProcessInputRegisters(const uint8_t* pData, size_t length)
 	{
 		int alertLevel = (err == 0) ? 1 : 4; // 1=Green (OK), 4=Red (Error)
 		std::string msg = (err == 0) ? "Unit Status: OK" : "Unit Error: " + std::to_string(err) + " (Sub: " + std::to_string(suberr) + ")";
-		UpdateAlertSensor(121, alertLevel, msg, "Unit Status");
+		SendAlertSensor(121, 255, alertLevel, msg, "Unit Status");
 	}
 
 	// Binary States (30-37, 52, 53)
-	if ((val = getReg(30)) != 0x7FFF && val != 32766) UpdateSwitch(130, val != 0, "Circulation Pump");
-	if ((val = getReg(31)) != 0x7FFF && val != 32766) UpdateSwitch(131, val != 0, "Compressor");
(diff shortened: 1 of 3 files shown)


gizmocuz:
Question, why don't you call the base hardware functions? like now you have

void CDaikinModbus::UpdateTemperatureSensor(int ChildID, float temperature, const std::string& szDefaultName)
{
	// Important: Use unique ChildID as NodeID to prevent overwriting
	SendTempSensor(ChildID, 255, temperature, szDefaultName);
}

void CDaikinModbus::UpdateSwitch(int ChildID, bool bOn, const std::string& szDefaultName)
{
	SendGeneralSwitch(m_HwdID, ChildID, 255, bOn ? 1 : 0, 0, szDefaultName, "");
}

All these functions should not be necessary and makes the code longer/harder to read

adrighem:
Good point. I expected more complexity, which isn't there. Time to clean up!

I have refactored the code to call the base hardware functions directly, replacing the Update... calls with direct calls to SendTempSensor, SendGeneralSwitch, SendSetPointSensor, etc.

gizmocuz:
Thanks ! One change please

 SendGeneralSwitch(m_HwdID, 130, 255, val != 0 ? 1 : 0, 0, "Circulation Pump", "");

to

 SendGeneralSwitch(m_HwdID, 130, 255, val != 0 ? 1 : 0, 0, "Circulation Pump", m_Name);

(And for the others as well, m_Name is the hardware name)

gizmocuz:
Let me know when you would like me to merge this

adrighem:
Fixed the username! I also took a deeper look at the Send* calls and fixed a register wrapping issue (the uint8_t overflow) and stabilized the NodeID logic so multiple units won't clash (I know, rare case, but still). Any comment?

adrighem:
If not, then please merge. :smile:

gizmocuz:
Is there a reason you are not using CDomoticzHardwareBase::SendSwitch ?
This function checks if there is any change, if not, it stops. This will prevent unnecessary updates to scripting tools/UI at each poll interval

adrighem:
Uhmm, I plead ignorance. Will update the calls.

gizmocuz:
Thanks, tell me when you want this to be merged

adrighem:
The current state looks good. Thanks for the suggestions. Please merge at your convenience.

gizmocuz:
Thank you!

PR #6714, opened 2026-03-31
Add percentage command template handling in MQTT

heggink:
added percentage command template for a fan device. It was missing


gizmocuz:
Thanks for the PR. 
Unfortunately this is not the way to implement a template. Tomorrow the template could be totally different.
As we already discussed, the one making this config (on a ESP device) could easily work internally with this multiply value

PR #6715, opened 2026-03-31
Plot energy return under the horizontal axis

janvanbesien:
In the Usage Last Month and Usage Last Year charts, returned energy was already plotted as negative values. This commit does the same for the Usage / Hour and Usage Last 7 Days charts.

For the 'day' range API, the power (r, r1, r2) is returned as already-negative values. The negateDatapoints call was flipping them back to positive, so that had to be removed.

### DIFF
--- a/www/app/log/CounterLogEnergySeriesSuppliers.js
+++ b/www/app/log/CounterLogEnergySeriesSuppliers.js
@@ -328,6 +328,7 @@
                     dataIsValid: function (data) {
                         return data.delivered === true;
                     },
+					postprocessDatapoints: negateDatapoints,
 					label: 'C',
 					series: {
 						type: 'column',
@@ -384,7 +385,6 @@
                     },
                     showWithoutDatapoints: false,
                     //convertZeroToNull: true,
-                    postprocessDatapoints: negateDatapoints,
                     label: 'R',
                     template: {
 						type: 'area',
@@ -419,7 +419,6 @@
                         return data.delivered === true;
                     },
                     showWithoutDatapoints: false,
-                    postprocessDatapoints: negateDatapoints,
                     label: 'R',
                     template: {
 						type: 'area',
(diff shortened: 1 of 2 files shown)


gizmocuz:
@janvanbesien , thank you for this PR! Is it possible to add a screenshot of the before/after?

janvanbesien:
Sure, this is before

Screenshot From 2026-04-01 08-29-33

and after

Screenshot From 2026-04-01 08-18-54

gizmocuz:
Okey but the first screenshot is exactly how it should be

janvanbesien:
Why? The green (production) bars are below the axis in the last 2 graphs (month and year overviews). So to me it would make sense if they are also below the axis in the first 2 graphs (last 7 days and hourly overviews)?

It just makes everything consistent (blue/usage above, green/production below).

You could of course also make it consistent in the other direction (change the last two graphs and so that the green bars are also above the axis in those), but I find it easier to read if usage and production do not overlap.

gizmocuz:
This was requested, and me and other also think this looks better. (It was also displayed like this before... so we have display A -> B -> A again)

janvanbesien:
ok.. then we'll have to agree to disagree. Perhaps could make it a choice, but that's more work.

Can't argue about what people think looks better, but the fact that it is inconsistent is still strange?

gizmocuz:
What is your issue with the current implementation? Is it the 7 days shortlog? It displays much better with 3

janvanbesien:
> What is your issue with the current implementation? Is it the 7 days shortlog? It displays much better with 3

My issue is twofold:

- I was used to seeing usage above the axis, and production below the axis, in the the 7 (or 3) days shortlog. I find that easier to read then when they are both above the axis. This is a preference, and other people might have different preferences, fine with that.
- Whatever you choose, I find it strange that you don't make the same choice for all graphs. The month and year overviews still show production below the axis like it used to be. That is inconsistent.

gizmocuz:
The reason is that for 365 values you can not do this, I am sure you see this as well

PR #6723, opened 2026-04-03
improvement Netatmo connection

JurgenLB:
The weatherstation is available in HomesData + HomeStatus.
I leaf the code as is for reference.

Also only call the scenarios with scopes from Gateway's

This result is lesser calls to Netatmo Servers.

Kind Regards


gizmocuz:
Thank you!

PR #6727, opened 2026-04-04
EU number format

ml17950:
Hello.

I added the feature, to display numbers in the EU format (with , as separator).
A long wanted feature from [this forum threat](https://forum.domoticz.com/viewtopic.php?t=28313).

It would be nice to see if the changes found their way into the project.

EU-Numbers-01
EU-Numbers-02

### DIFF
--- a/main/WebServerCmds.cpp
+++ b/main/WebServerCmds.cpp
@@ -2931,6 +2931,11 @@ namespace http
 			root["CurrencySign"] = m_sql.m_currencysign;
 			root["PriceResolution"] = m_sql.m_PriceResolution.load();
 
+			if (m_sql.GetPreferencesVar("EUNumberFormat", sValue))
+			{
+				root["EUNumberFormat"] = sValue;
+			}
+
 			int iUser = -1;
 			if (!session.username.empty() && (iUser = FindUser(session.username.c_str())) != -1)
 			{
@@ -3798,6 +3803,9 @@ namespace http
 
 				std::string sCurrency = request::findValue(&req, "CurrencySymbol");
 				m_sql.UpdatePreferencesVar("Currency", sCurrency); cntSettings++;
+
+				std::string sEUNumberFormat = request::findValue(&req, "EUNumberFormat");
+				m_sql.UpdatePreferencesVar("EUNumberFormat", sEUNumberFormat); cntSettings++;
 
 				bool AllowPlainBasicAuth = (request::findValue(&req, "AllowPlainBasicAuth") == "on" ? 1 : 0);
 				m_sql.UpdatePreferencesVar("AllowPlainBasicAuth", AllowPlainBasicAuth);
@@ -6528,6 +6536,10 @@ namespace http
 				else if (Key == "Currency")
 				{
 					root["Currency"] = sValue;
+				}
+				else if (Key == "EUNumberFormat")
+				{
+					root["EUNumberFormat"] = sValue;
 				}
 				else if (Key == "ESettings")
 				{
--- a/www/app/SetupController.js
+++ b/www/app/SetupController.js
@@ -471,6 +471,7 @@ define(['app'], function (app) {
 						$("#locationtable #Longitude").val(data.Location.Longitude);
 					}
 					$("#locationtable #CurrencySymbol").val(data.Currency);
+					$("#locationtable #EUNumberFormat").val(data.EUNumberFormat);
 					if (typeof data.ProwlEnabled != 'undefined') {
 						$("#prowltable #ProwlEnabled").prop('checked', data.ProwlEnabled == 1);
 					}
--- a/www/app/TemperatureController.js
+++ b/www/app/TemperatureController.js
@@ -574,6 +574,25 @@ define(['app', 'livesocket'], function (app) {
 					var ctrl = this;
 					var item = $scope.item;
 
+					// EU ., replacements
+					if ($.myglobals.EUNumberFormat) {
+						if (item.Temp != null) {
+							item.Temp = formatEUValue(item.Temp);
+						}
+						if (item.Speed != null) {
+							item.Speed = formatEUValue(item.Speed);
+						}
+						if (item.Gust != null) {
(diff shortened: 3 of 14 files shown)


gizmocuz:
This is not the way... 

For now, I assume you are from Germany

Set your system local to german (And I mean the system where you run Domoticz on)
Set your system (desktop) to german
Set Domoticz language to german
You browser should now be in german and the charts should display correctly

ml17950:
Hi.

Yes, that's right. I'm from Germany.

First of all, that's not working...
- Domoticz Server is DE
- Notebook is DE
- Language in Domoticz is German
- Numbers are still with a . (dot)

Bildschirmfoto vom 2026-04-04 13-50-21

___

And secondly, you can choose whether the temperature is displayed in Celsius or Fahrenheit, the weight in kilograms or pounds, or which currency symbol should be shown. All this without changing the language.

However, you can't decide how the numbers are displayed? Even if you want the user interface in a different language, for example?

gizmocuz:
It is Highcharts that is rendering the numbers. I remember this has been asked before and it is a matter of setting up highcharts correctly which we do at the beginning as well for weekdays. And for sure the suggestion you made in this PR is not the way. It was a simple change, and I think it is also prepared and tested. No idea where anymore as it has been quite some while ago

gizmocuz:
If you google you will also find this

Highcharts.setOptions({
    lang: {
        decimalPoint: '.',
        thousandsSep: ','
    }
});

Germany:

Highcharts.setOptions({
    lang: {
        decimalPoint: ',',
        thousandsSep: '.' // or ' ' for a space
    }
});

Ahh yes, and you can also specify the language Highcharts speaks

gizmocuz:
Yes, this was it

```
.setOptions({
    lang: {
        locale: 'de-DE' // Automatically sets comma for decimal
    }
});
```

gizmocuz:
So we need to extend probably the c++ code (or a global JS code as it is for highcharts only) that returns this local code based on the language that is selected in the settings. (for all current languages)

But test first with one (German)

waltervl:
@gizmocuz the screenshots where @ml17950 is pointing to are on the device widgets numbers, not on the highcharts graph numbers. Is highcharts also defining the way numbers are rendered on the widgets?

ml17950:
Yes, @waltervl  is right. I'm not speaking of the charts. The display in the widgets and the info-card above the charts.

gizmocuz:
Ahh okey, but this is a globally highcharts issue, if it is implemented right at the highcharts initialization in domoticz, it is correct on the dashboard as well in the device log screens

gizmocuz:
Thank you for contributing this — EU number formatting is a long-standing request and the feature itself is absolutely valid.

After reviewing the diff, I have some concerns about the implementation that would need to be addressed before this can be merged.

---

**Data mutation bug**

In `TemperatureController.js` and `WeatherController.js`, the code mutates the data model directly:

```js
item.Temp = formatEUValue(item.Temp);
item.Speed = formatEUValue(item.Speed);
```

This replaces numeric values with comma-formatted strings on the actual data object. If those properties are used anywhere else (chart rendering, calculations, comparisons, watchers), they will silently receive a string like `"23,5"` instead of the number `23.5`, which will break things.

---

**Incomplete coverage**

The PR patches around 10 specific places, but the main device views — the Devices list, Temperature/Weather/Utility grid tiles, and Highcharts chart tooltips and axis labels — are not covered. Users enabling this setting would see inconsistent formatting across the UI.

---

**Fragile approach**

`formatEUValue` does a raw `replaceAll('.', ',')` on strings that may already contain units, signs, or spaces. The correct approach would be a centralized locale-aware number formatter using the browser's built-in `Intl.NumberFormat` API, or configuring Highcharts' locale globally. Scattering ad-hoc replacements across every controller is hard to maintain and easy to miss.

---

**Missing i18n strings**

`data-i18n="EU-Numbers"` is referenced in `setup.html` but the translation key is not added to any of the 20+ language files in `www/i18n/`.

---

**Minor UI issue**

The label column width in `setup.html` was changed from `80px` to `120px` for the EU Numbers row only, which shifts the layout of all other rows in that table.

---

**Summary**

The backend preference storage (WebServerCmds.cpp) is fine. The frontend needs a rethink: a single central formatting utility applied at the right layer, full coverage of all device views and charts, and no mutation of data objects. If you are able to rework the frontend along those lines, we would be happy to reconsider.

gizmocuz:
To follow up on the previous comment — there is a better way to implement this that avoids the scattered per-controller patches entirely.

**Option 1: AngularJS filter (best fit for this codebase)**
Create a single `formatNumber` filter that wraps the browser's `Intl.NumberFormat` API with the user's locale. Any template binding using `{{ value | formatNumber }}` would automatically render with the correct decimal separator. No data mutation, no scattered `if (EUNumberFormat)` checks scattered across every controller. One filter, one place to maintain.

**Option 2: Configure Highcharts locale globally**
Highcharts has a built-in `Highcharts.setOptions({ lang: { decimalPoint: ',' } })` global setting. Set it once at app init based on the stored preference and all chart tooltips and axis labels format correctly — zero per-chart changes needed.

**The recommended approach is Option 1 + Option 2 combined.** The AngularJS filter handles all stat card and template display values; the Highcharts global setting handles all charts. That covers the entire UI with two changes instead of 30+, and since the raw numeric values are never mutated, there is no risk of breaking calculations or chart rendering elsewhere.

The backend preference storage you already have is perfectly fine and would not need to change.

gizmocuz:
@ml17950 , I cleaned up the response to this PR a bit. As you see above, I still think we should solve this globally like in my first answer

Edit: I am working on a global implementation as I am not fond of changing a lot of files. Let you know when it is implemented for testing

gizmocuz:
Okey, I was half way, but decided NOT to implement this as this will be a continues change in the development cycle, any new widget/device/print will need to be converted.
Keeping it the way it is... you have to get used to it! (i did not even know it was an issue, and for those few users how hates points or comma's ... get used to it)

PR #6729, opened 2026-04-05
Fix/linux compiler issues and Command http reply status

kiddigital:
Some small code cleanup preventing compiler warning on Linux and better/clearer result statuses from some commands.


gizmocuz:
Thank you for this PR!!!

PR #6730, opened 2026-04-05
Refactor self-sufficiency calculation logicFix self-sufficiency calcu…

frepke:
…lation: use gross grid import instead of net grid

Updated self-sufficiency calculation to use gross grid import instead of net grid balance.


gizmocuz:
Thank you for this PR. I made a few modifications and merged your fix. Will be in a upcoming beta version today

PR #6731, opened 2026-04-05
Fix orcon fan in UI

JurgenLB:
Orcon (SubType="Orcon") was missing from the fan-specific UI checks in dzLightWidget.js, causing Orcon devices to fall through to generic icon/click handling instead of showing the fan icon and popup.

### Changes

Fan icon (dzLightWidget.js): added Orcon to the SubType prefix check so Orcon devices always render Fan48_On.png
Click popup (dzLightWidget.js): extended the Itho popup condition to include Orcon

### Why Orcon shares the Itho popup

The Itho popup sends commands '1', '2', '3', 'timer'. The backend GetLightCommand already has a case sTypeOrcon: that maps these identically to fan_Orconlow, fan_Orconmedium, fan_Orconhigh, and fan_Orcontimer1.

Kind Regards

### DIFF
--- a/main/mainworker.cpp
+++ b/main/mainworker.cpp
@@ -5690,6 +5690,7 @@ void MainWorker::decode_Fan(const CDomoticzHardwareBase* pHardware, const tRBUF*
 	uint8_t OID3;
 	uint8_t OID4;
 	int LastLevel = 0;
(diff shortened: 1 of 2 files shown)


gizmocuz:

PR #6740, opened 2026-04-08
Daikinmodbus: Fix C4244 compiler warnings on Windows MSVC

adrighem:
Sorry for this! This fixes the warnings by:
- Changing childId to uint8_t in the device definition structs.
- Adding an explicit static_cast<float> for dPowerWatt in SendWattMeter (while still using the double internally).


gizmocuz:
Thank you!

Issue #6718, opened 2026-04-02
Time allocated to python plugin


pipiche38:
### Prerequisites

- [x] I have searched existing issues and the forum for this problem
- [x] I am running the latest beta version

### Domoticz Version

2026.1

### Platform

Linux (x86_64)

### OS Version

Fedora - Kernel 6.11.9

### Describe the Bug

I'm currently investigating the possibility to have the plugin stuck for several seconds and getting not time for execution and especially in regards to the AsyncLoop.

In such I have developed a short plugin https://github.com/pipiche38/Monitoring-Loop-Latency

I'm running it on my production environment (running latest Domoticz 2026.1); python3.12

What I observed so far is that time to time the AsyncLoop doesn't get cpu to time for more than 5s, and this seems to concur when an auto-backup is scheduled.

### Steps to Reproduce

Install the plugin https://github.com/pipiche38/Monitoring-Loop-Latency

It will create necessary devices to track the asyncloop latency

### Expected Behavior

Expects the latency to be below 1s

### Log Output

```text

```

pipiche38:
CC: @badzz

 I suspect that GIL could be stuck somewhere (like on a blocking IO) and all plugin (using python framework) are stuck.
I did test with an old 2025.2 domoticz version and it was worst, bigger latence (such as 10 seconds)

Smanar:
Lol, you are realy a bug hunter, how you can found this kind of bugs ?

pipiche38:
@Smanar we are using asyncio loop (via the zigpy library), and there is one type of ZIgbee Coordinator ( EZSP) which required a watchdog feed within 5 seconds to keep the uart connection up.
And time to time we get the connection lost because the plugin was stuck for more than. 5seconds

waltervl:
So one option is to temporarily disable the autobackup in Domoticz. Should it fix it? I know Gizmocuz is not a big fan how this autobackup was implemented by someone else in the past (long time ago....)

Edit: what happens if you just make a manual database backup through menu setup-setting- Tab backup/restore ?

gizmocuz:
Not a big fan of Auto backup at all. We should actually remove this one day. Totally useless. sure once in a while you can do a backup  from the settings page. If you update frequently you already have backups on your disk in the backup folder (the updater does this)

Domoticz gets unresponsive for some time because we need to 'lock' everything during a backup, nobody (Domoticz/hardware/plugins) is allowed to read/write anything in the database

[...]

PR #6741, opened 2026-04-09
Make sure that prior to block I/O SQL statement GIL is released

pipiche38:
This PR is about working on improving time allocation to python framework thread #6718
Every blocking SQL statement must be protected with the 
Py_BEGIN_ALLOW_THREADS
Py_END_ALLOW_THREADS

PR is WIP. It might be also needed to check against other blocking I/O (network, disk, sleep…)


pipiche38:
something wrong to source file

PR #6742, opened 2026-04-09
fix(plugins): release GIL during blocking SQL calls in plugin context

### LINKED ISSUE #6718: Time allocated to python plugin

gizmocuz:
## Problem

Closes #6718.

Python plugin threads hold the Python GIL (Global Interpreter Lock) while executing blocking SQLite queries in `Plugins.cpp`. This starves all other plugins and asyncio loops for the full duration of the query — observed as 5–10 s latency spikes, especially during auto-backup or on slow/busy storage.

The symptom: asyncio-based plugins (e.g. zigpy/EZSP) lose their UART watchdog connection because their event loop gets no CPU time for >5 s.

`PythonObjectEx.cpp` already wraps its SQL calls with `Py_BEGIN_ALLOW_THREADS`/`Py_END_ALLOW_THREADS` correctly. `Plugins.cpp` had no such protection at all.

## Solution

Added a `PyAllowThreads` RAII guard class to `Plugins.h`, consistent with the existing `PyBorrowedRef`/`PyNewRef` pattern:

```cpp
class PyAllowThreads
{
    PyThreadState *m_save;
public:
    PyAllowThreads()  : m_save(PyEval_SaveThread()) {}
    ~PyAllowThreads() { PyEval_RestoreThread(m_save); }
    PyAllowThreads(const PyAllowThreads &) = delete;
    PyAllowThreads &operator=(const PyAllowThreads &) = delete;
};
```

Advantages over the `Py_BEGIN/END_ALLOW_THREADS` macros:
- Exception-safe (destructor always runs)
- No mismatched begin/end pairs possible
- Multiple consecutive SQL calls can share one guard (one acquire/release instead of N)

Applied the guard to all seven blocking SQL call sites in `Plugins.cpp` that were missing it:

| Location | Operation |
|---|---|
| `PyDomoticz_Configuration` | UPDATE config, SELECT config |
| `CPlugin::Start` | `GetPreferencesVar` + SELECT hardware params (grouped) |
| `CPlugin::Start` | UPDATE settings |
| `CPlugin::Start` | SELECT DeviceStatus (both module branches) |
| `CPlugin::Start` | SELECT CustomImages |
| `CPlugin::PopulateModuleDict` | SELECT Preferences |
| `CPluginNotifier::GetIconFile` | SELECT icon |

All Python-derived values (`std::string`, `int`) are captured in plain C++ types **before** each guard scope — no `PyObject*` is accessed inside a guard.

## Testing

The fix can be verified with the latency-monitoring plugin from #6718:
https://github.com/pipiche38/Monitoring-Loop-Latency

Expected: asyncio loop latency stays well below 1 s during normal operation (auto-backup, plugin startup, etc.).

### DIFF
--- a/hardware/plugins/Plugins.cpp
+++ b/hardware/plugins/Plugins.cpp
@@ -485,13 +485,20 @@ namespace Plugins
 				//  Convert to JSON and store
 				std::string sConfig = jsonProtocol.PythontoJSON(pNewConfig);
 
-				// Update database
-				m_sql.safe_query("UPDATE Hardware SET Configuration='%q' WHERE (ID == %d)", sConfig.c_str(), pModState->pPlugin->m_HwdID);
+				// Update database — release GIL during blocking SQL write
+				{
+					PyAllowThreads gil;
+					m_sql.safe_query("UPDATE Hardware SET Configuration='%q' WHERE (ID == %d)", sConfig.c_str(), pModState->pPlugin->m_HwdID);
+				}
 			}
 			PyErr_Clear();
 
-			// Read the configuration
-			std::vector<std::vector<std::string>> result = m_sql.safe_query("SELECT Configuration FROM Hardware WHERE (ID==%d)", pModState->pPlugin->m_HwdID);
+			// Read the configuration — release GIL during blocking SQL read
+			std::vector<std::vector<std::string>> result;
+			{
+				PyAllowThreads gil;
+				result = m_sql.safe_query("SELECT Configuration FROM Hardware WHERE (ID==%d)", pModState->pPlugin->m_HwdID);
+			}
 			if (result.empty())
 			{
 				pModState->pPlugin->Log(LOG_ERROR, "CPlugin:%s, Hardware ID not found in database '%d'.", __func__, pModState->pPlugin->m_HwdID);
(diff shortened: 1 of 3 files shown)


pipiche38:
That is fast. Thanks very much.
Let me know , when a build is available, I'll be able to test it on my prod to validate the result.

gizmocuz:
@pipiche38 , could you test with beta 17686?

pipiche38:
@gizmocuz beta 17686 freshly recompiled on my rpi started. Will keep you posted here if ok ?

pipiche38:
For information, I did test the Vacuum feature. No impact on the UART, so this sound good. I'll monitor during the next 24hours

gizmocuz:
Great thanks!!

pipiche38:
@gizmocuz I do confirm after a full night, with even backup enabled no issue at all.
The Latency is around 1.3 ms and max has been 126 ms!

That do confirm fixing the initial issue.

gizmocuz:
Great! Is your plugin using the Domoticz or the DomoticzEx plugin framework?
As the DomoticzEx already had most of them in place...

pipiche38:
Yes, I didn't mentioned. I'm still using the Legacy framework (Domoticz).
In process of migrating to the new one, this is the next release of the plugin.

Issue #6576, opened 2026-02-17
Blockly UI issue


blooesky:
Hi,

Today i decided to  report an old problem with blockly. 
If you open 2 or more tabs with blocks in events and then navigate for example to the dashboard or another page and then return to events, the blockly display is not correct. The only exception is the last blockly tab opened. I attach a print screen. I tested on the following browsers: Firefox, Chrome, Brave and the display is identical. This issue is older, it is not from the latest versions of domoticz beta.

I hope I explained it well.

![Image](https://github.com/user-attachments/assets/3eb72a5d-e40a-47e1-8598-878aeeb22db0)

Thank you!

gizmocuz:
I looked at it again, and again, and ... again, and can't seem to find a way to solve this (other then using dzVents instead of blockly)
Maybe we need to upgrade blockly to V12 (we are on V8), but there have been made many breaking changes (xml layout etc)
We better not do this
I have the same problem when I have 2 blockly open in the tab, do a browser refresh F5, and switch to tab 1

blooesky:
The inactive tabs..i think not keeping the size...and maybe need resized for the active tab (workspace.resize()). Maybe it will work.

gizmocuz:
Tried... Could not get it to work.... If you could try please do so... After another hour of fiddling with it, and lots of changes, i bailed out..

adrighem:
Verified that the issue is solved. @blooesky Can you confirm, by using a beta release?

blooesky:
Yes looks good. Thanks!

gizmocuz:
Thanks! Closing issue!

PR #6746, opened 2026-04-12
Fix Blockly block sizes corrupted when multiple tabs open

JurgenLB:
**Root cause**: When the page is refreshed with multiple Blockly tabs open, Angular's ng-show hides all inactive event-viewer components (display: none). Blockly initializes in these hidden containers and cannot measure block dimensions (all DOM measurements return 0), so blocks get rendered with incorrect/zero sizes. The existing ResizeObserver called Blockly.svgResize() when the tab became visible, but that only resizes the SVG canvas wrapper — it does not re-measure individual block widths/heights.

**### Fix in www/app/events/EventViewer.js** (initBlockly function):

- needsRerender flag — set to true when container.offsetWidth === 0 at initialization time (i.e., the container is hidden in an inactive tab). This avoids touching the active tab's workspace unnecessarily.

- suppressChangeEvents flag — shared between the change-listener closure and the ResizeObserver closure. Used to suppress the Blockly change listener during the re-render so the script isn't incorrectly flagged as modified (unsaved changes *).

- Re-render on first visible resize — in the ResizeObserver callback, when needsRerender is true and the container first gets a positive width (tab becomes active), the workspace XML is saved, the workspace is cleared, and the XML is reloaded. This forces Blockly to re-measure all block dimensions against the now-visible container, restoring correct block sizes.

CoPilot helped and for my system is seems to work now.

Kind Regards


JurgenLB:
This is to close #6576 

Kind Regards

gizmocuz:
Thank you for the PR!
This looks like a straight forward fix.

There is also another PR created to upgrade blockly which involves a lot more changes and also seems to fix the issue.

Which one do you preffer?

https://github.com/domoticz/domoticz/pull/6756

JurgenLB:
I do love the use off the latest available software versions.
But this one contains a lot less changes.
And by reading your comments on the issue at hand, I thought you did not prefer to update.

If you want a fix you can use this PR and update when there is a need for.

Kind Regards

gizmocuz:
@JurgenLB , You're correct. I will merge this PR. Thank you both!

Issue #6751, opened 2026-04-13
SolarEdge energy sensors introduced in 2026.1 showing incorrect data, lifetime data is not the actual lifetime data


robk358:
### Prerequisites

- [x] I have searched existing issues and the forum for this problem
- [x] I am running the latest beta version

### Domoticz Version

2026.1 (buld17730)

### Platform

Raspberry Pi

### OS Version

Raspbian Bookworm

### Describe the Bug

Issue 1:
The new SolarEdge energy sensors do not show the correct information, the API calls return the correct data (shown in setup->devices). This data is sent to a KwhMeter, which results in wrong outputs:
<img width="793" height="202" alt="Image" src="https://github.com/user-attachments/assets/3554c257-d046-4488-87b1-f312226d0cb8" />
this should be like this:
<img width="797" height="199" alt="Image" src="https://github.com/user-attachments/assets/d102b007-d7d9-4c3b-ba25-7ba04462a56b" />
The log of (for example "energy today", but "energy this month" and "energy this year" have the same issue) shows negative values (as a result of the API returning data from the previous day):
<img width="603" height="387" alt="Image" src="https://github.com/user-attachments/assets/227b330c-377c-40e0-823c-2c2621a0f9e6" />

Issue 2:
All lifetime energy sensors (including optimizers) are calculated from "current values" of the API while there are real lifetime values available. displayed values are not the real values. Log display als doesn't make sense:
<img width="1239" height="793" alt="Image" src="https://github.com/user-attachments/assets/4d6dccf0-5f9c-4657-bc77-eae7fe45531e" />
Should be like:
<img width="1207" height="798" alt="Image" src="https://github.com/user-attachments/assets/4c37f1ea-b02f-4155-ae7d-4997262f4e13" />

### Steps to Reproduce

1. Go to 'Utility'
2. Look at new SolarEdge energy devices
3. See error

### Expected Behavior

See bug description.

### Log Output

```text

```

### Additional Context

Preparing PR for fixes

robk358:
Solved with #6752

gizmocuz:
Thanks once again!

PR #6752, opened 2026-04-13
fix for #6751

robk358:
This PR fixes issues described in #6751


gizmocuz:
Thank you for the PR.
Seems you change quite some more then just the lifetime sensors....
Hope to validate it soon before merging!

gizmocuz:
Thank you!

PR #6767, opened 2026-04-17
fix: commands expect upper case in python-matter-server

fleinze:
Python-matter-server expects upper case commands ("On", "Off").
Without the server returns an error message.
With this fix the on/off-commands work.

### DIFF
--- a/hardware/Matter.cpp
+++ b/hardware/Matter.cpp
@@ -1140,11 +1140,11 @@ bool CMatter::WriteToHardware(const char* pdata, unsigned char length)
 	{
 		std::string cmd;
 		if (cmnd == light2_sOn)
-			cmd = "on";
+			cmd = "On";
 		else if (cmnd == light2_sOff)
-			cmd = "off";
+			cmd = "Off";
 		else
-			cmd = "toggle";
+			cmd = "Toggle";
 
 		Json::Value args;
 		args["node_id"]      = nodeId;


fleinze:
I now see what happened here.
Gizmocuz uses matterjs-server, I used python-matter-server.
The commands work fine with matterjs-server, but not with python-matter-server.
I know that python-matter-server is depricated, but matterjs-server is still officially in alpha.
So the question is, do we want to keep compatibility with python-matter-server in Domoticz?
If we go full matter-js we should change the plugin name in RFXNames.cpp to reflect that.

gizmocuz:
Yes we are going the full matterjs way ! I will make the change!

fleinze:
These are different repositorys:
https://github.com/matter-js/matterjs-server/
the new matter-js server, currently alpha
https://github.com/matter-js/python-matter-server/
the old, stable, but soon-to-be depricated python-matter-server

commands currently work with the matterjs-server, but not with python-matter-server

gizmocuz:
Yes sorry, I was a bit to fast...

So I think we should support both, your fix in this PR works for the old, but will break on the new version...
Maybe we can detect this at startup... I have a look very soon... Hope there are no more suprises.

I noticed (at least with the new matterjs-server version) that my door contact was not received anymore... It is on my Apple Home app...
I will try the old server as well
Hope I can keep my docker data folder and switch instead of including the nodes again ;)

Edit: Could also have been the batteries used... Let's stick to the new matterjs-server as this will replace the old one

PR #6771, opened 2026-04-18
fix: batterypercentage usually lies on endpoint 0 of node

fleinze:
Battery percentage is only displayed in combinded nodes (temp+hum in my case).
This is because the code expects BatPercentRemaining to be in the same endpoint as the data. But it typically is on endpoint 0. This PR fixes this by scanning all endpoints for BatPercentRemaining.

### DIFF
--- a/hardware/Matter.cpp
+++ b/hardware/Matter.cpp
@@ -911,8 +911,13 @@ void CMatter::_DetectAndSend(int nodeId, int endpointId)
 	if (epIt == nodeIt->second.endpoints.end())
 		return;
 	const auto& state = epIt->second;
-	int battery = state.battery_pct;
 	int domoticzID = nodeId * 256 + endpointId;
+
+	// Battery level from any endpoint on this node (PowerSource cluster
+	// typically lives on endpoint 0, not on the sensor endpoint).
+	int battery = 255;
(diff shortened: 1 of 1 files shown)


gizmocuz:
Hi @fleinze, thank you for spotting this and taking the time to submit a PR!

You correctly identified the root cause: `BatPercentRemaining` (and `BatVoltage`) are reported by devices on endpoint 0 (the root endpoint per the Matter spec), while the actual sensor clusters live on endpoints 1, 2, etc. This is why combined sensors like temp+hum worked — they go through a node-level aggregation path that happened to scan ep 0 as well — while per-endpoint sensors (lux, occupancy, door/window) did not.

We took a slightly different approach than scanning all endpoints: instead of keeping battery in `EndpointState` and walking all endpoints to find it, we moved battery entirely to `NodeState` and handle it exclusively in the node-level metadata path (which only runs for endpoint 0, where the PowerSource cluster lives). This is cleaner because battery is inherently a node-level property — there's only one battery per physical device regardless of how many sensor endpoints it exposes. It also captures `BatVoltage` (attr 11) alongside the percentage.

Thanks again for spotting this — we can't wait to see more PRs from you! 😊

Issue #6776, opened 2026-04-18
Matter issue: SwitchType is not correctly set with _ApplySwitchTypeOnCreate


fleinze:
### Prerequisites

- [x] I have searched existing issues and the forum for this problem
- [x] I am running the latest beta version

### Domoticz Version

2026.1 (build 17754)

### Platform

Raspberry Pi

### OS Version

official docker container

### Describe the Bug

When a new matter device is created that has level-control but not color-control, the device is created with switchtype 0 instead of switchtype 7.

In matter.cpp `SendGeneralSwitchInt()`
```
	SendGeneralSwitch(domoticzID, unit, battery, value, level, label, ""); //<-- creates device
	_ApplySwitchTypeOnCreate(domoticzID, unit, isNew, switchType);    //<-- updates switchtype
```

My guess is we have a race condition. `SendGeneralSwitch` calls the database asynchronous, so that the device is not yet created when `_ApplySwitchTypeOnCreate` is called.

sorry, no PR because I couldn't solve this myself

### Steps to Reproduce

add Matter-device with levelcontrol but now colorcontrol

### Expected Behavior

Matter-device is created with correct switchtype

### Hardware/Plugin Type

Matter

### Log Output

```text

```

gizmocuz:
It should be set only at creation time. If you delete the switch from the devices overview and restart the Matter hardware, does it create the switch correctly?

gizmocuz:
Maybe we should check for devType/subType, let me add this

gizmocuz:
Fixed in commit 90ad5c66f.

The `isNew` check in `SendGeneralSwitchInt` was querying `DeviceStatus` without filtering on `Type`/`SubType`, so it could match any existing device sharing the same `HardwareID`, `DeviceID`, and `Unit` — causing `isNew` to be `false` even though the General Switch device hadn't been created yet. As a result `_ApplySwitchTypeOnCreate` would skip the `UPDATE` and the switch type was never set.

The query now includes `AND Type=%d AND SubType=%d` with `pTypeGeneralSwitch`/`sSwitchTypeAC`, so the lookup only matches the correct device type.

fleinze:
The bug still persists. 
`iNew` check returns `true` and `SendGeneralSwitch` will create the device. The problem ist, that `SendGeneralSwitch` calls `sDecodeRXMessage` which - from my understanding - calls the update asynchronous via the boost library.
[...]

PR #6777, opened 2026-04-18
fix race condition that prevents switchtype from beeing set

fleinze:
This is a quick fix for issue #6776 and I don't know if you would want it this way.
But it definitly solves the issue.
The problem:
When a new switch device was created there was a race condition between creation and setting of switchtype, which led to the switchtype not being set. Creation is asynchronous and the device might not be created when switchtype is updated.
With this fix the switchtype is also checked and is set on a later call of `SendGeneralSwitchInt()`.

Another way would be to wait for the actual creation and then set the switchtype.


gizmocuz:
Thanks for the report and PR! Fixed in commit 90ad5c66f via a targeted query fix: the `isNew` check in `SendGeneralSwitchInt` now includes `AND Type=pTypeGeneralSwitch AND SubType=sSwitchTypeAC`, ensuring it only matches the correct device type and won't produce a false negative when another device shares the same HardwareID/DeviceID/Unit coordinates.

PR #6780, opened 2026-04-19
fix(dashboard): counter/energy charts show UTC instead of local time in dynamic dashboard

### LINKED ISSUE #6770: Counter graphs in dynamic dashboard have UTC time
### Prerequisites

### Domoticz Version

v2026.1 (build 17748)

### Platform

Linux (x86_64)

### OS Version

Debian 13.4

### Describe the Bug

Since first trying the dynamic dashboard (about 2 weeks ago) I noticed the graphs for Solar/Electricity etc, all display in UTC time, so are 2 hours off in my case. It is the only place where this happens, time is CEST as expected, and all graphs outside the dynamic dashboard are the correct local time for me.

### Steps to Reproduce

1. At a counter/Energy chart to to dashboard
2. Check the timeline on the x-axe
3. See that is UTC time that is shown

### Expected Behavior

It should have my localtime on the x-axe, just like outside the dynamci dashboard.

### Hardware/Plugin Type

_No response_

### Browser

Firefox 149 (but also happens with other browsers)

### Log Output

```text

```

### Screenshots

_No response_

### Forum Post

_No response_

### Additional Context

_No response_

CodeLine9:
## Problem
Counter/Energy chart widgets in the dynamic dashboard display timestamps in UTC instead of the browser's local timezone. Regular chart views outside the dynamic dashboard show correct local time.

## Root Cause
`baseChartOptions()` in `ddEnergyChart.widget.js` does not set `time.useUTC`, so Highcharts defaults to `useUTC: true`. The data points use `parseDateLocal()` which creates local timestamps, but Highcharts interprets them as UTC when rendering the x-axis labels and tooltips.

The regular log page charts handle this correctly in `RefreshingChart.js` via `time: { timezone: Intl.DateTimeFormat().resolvedOptions().timeZone }`.

## Fix
Add `time: { useUTC: false }` to chart options in `baseChartOptions()`. This makes Highcharts interpret the locally-created timestamps as local time, matching the behavior of charts outside the dynamic dashboard.

1 line change, affects all chart types in the Counter/Energy widget (day, week, month, year, shortlog, compare).

Fixes #6770

### DIFF
--- a/www/app/dashboardDynamic/widgets/ddEnergyChart.widget.js
+++ b/www/app/dashboardDynamic/widgets/ddEnergyChart.widget.js
@@ -296,7 +296,8 @@ define([
                         },
                         legend:    { enabled: false },
                         credits:   { enabled: false },
-                        exporting: { enabled: false }
+                        exporting: { enabled: false },
+                        time:      { useUTC: false }
                     };
                 }
 


gizmocuz:
It's not only in this chart but all dynamic widgets that uses charts.
The refreshing charts in the code base works correctly, and therefor the correct fix is implemented in the commit below:

https://github.com/domoticz/domoticz/commit/efa6b0a58d3e00088e4782f826775835e0bbe93f

Thanks for helping!

Ps. If you are using AI, use claude code, not claude via a browser

PR #6781, opened 2026-04-19
fix(solaredge): add missing return statements in isDaylightWindow()

### LINKED ISSUE #6774: Solaredge keeps displaying received power after sunset
### Prerequisites

### Domoticz Version

2026v1 build 17732

### Platform

Linux (x86_64)

### OS Version

Ubuntu 24

### Describe the Bug

Solaredge panels using optimizers do show generated power for each individual panel. After sunset this should be zero but until sunrise next day current value is maintained.

### Steps to Reproduce

2. Click on '...'
3. See error

### Expected Behavior

no further comment

### Hardware/Plugin Type

_No response_

### Browser

_No response_

### Log Output

```text

```

### Screenshots

_No response_

### Forum Post

_No response_

### Additional Context

_No response_

CodeLine9:
## Root Cause

`isDaylightWindow()` in `SolarEdgeAPI.cpp` had two bare `false;` expressions instead of `return false;` statements. These no-op expressions meant the function always returned `true`, regardless of time of day.

## Impact

Because the daylight window check never actually worked:
- Optimizer data was polled during nighttime hours, keeping stale power readings displayed after sunset
- Inverter details were fetched at night
- Overview data was polled outside daylight hours

SolarEdge panels with optimizers would show the last received power value indefinitely after sunset instead of zeroing out.

## Fix

Added the missing `return` keyword to both conditions in `isDaylightWindow()`.

Fixes #6774

### DIFF
--- a/hardware/SolarEdgeAPI.cpp
+++ b/hardware/SolarEdgeAPI.cpp
@@ -252,9 +252,9 @@ bool SolarEdgeAPI::isDaylightWindow()
 	int sunRise = getSunRiseSunSetMinutes(true);
 	int sunSet = getSunRiseSunSetMinutes(false);
 	if (ActHourMin + 60 < sunRise)
-		false;
+		return false;
 	if (ActHourMin - 60 > sunSet)
-		false;
+		return false;
 	return true;
 }
 


gizmocuz:
Thank you for the PR and your time!!

PR #6783, opened 2026-04-19
fix for race condition in device creation

fleinze:
see issue #6776 
There is a race condition in slower systems (rpi 4 in my case) where `_ApplySwitchTypeOnCreate` fails because the device is not created yet. This PR fixes this by introducing a wait loop that loops for a maximum of 20 time and sleeps 5ms each time.
This way it is ensured that the device that is updated already exists in the database.

Without this PR most switches that are newly created have the default switchtype.
With this PR all switches are created correctly.


fleinze:
1. Added check if new hardware is accepted
2. adopted to match the code style
3. use `sleep_milliseconds` from Helper.h

fleinze:
I see you fixed the problem

gizmocuz:
Yes, I commented on the issue. Did you test if if it was solved?

fleinze:
Yes, I did test it and yes, it is solved!

PR #6784, opened 2026-04-19
fix: correct AirQuality compare-view divisor from 10 to 1

fleinze:
MakeCompareDataSensor was called with divisor 10.0 for pTypeAirQuality, causing monthly averages in the bottom compare panel to be 10x too low (e.g. 1000 ppb shown as 100 ppb). Unlike pTypeWEIGHT which stores values scaled by 10 internally, VOC/AirQuality stores raw integer readings that need no scaling.


gizmocuz:
Thank you for the PR!

PR #6789, opened 2026-04-23
SolarEdge ResetPowerValues: reset optimizer values if we are polling them

robk358:
SolarEdge.cpp ResetPowerValues: only reset optimizer values if we are polling them, otherwise they may create unnecessary devices. Also removed unneeded AcCurrent and Power resets, the API returns 0 before daylight window ends


robk358:
Also reset values for "Voltage" and "Optimizer Voltage" (the behave the same as "Current" and "Power")

gizmocuz:
Thank you!

PR #6794, opened 2026-04-24
fix(mqtt): prevent discovery message loss during startup burst (#6754)

### LINKED ISSUE #6754: MQTT Auto Discovery not triggered for Zigbee Light devices after restart (regression after ~17538)
### Prerequisites

### Domoticz Version

17732

### Platform

Linux (x86_64)

### OS Version

Raspberry Pi 4 Raspbian GNU/Linux 12 (bookworm)

### Describe the Bug

Describe the bug

After upgrading Domoticz, MQTT Auto Discovery for Zigbee2MQTT devices of type Light is no longer triggered automatically after a Domoticz restart.

The issue appears to be a regression introduced after version 17538 (last version where it likely worked correctly).
I first noticed it starting from 17592, and it is still present in the latest version 17732.

This behavior significantly impacts usability, as all Light devices require manual intervention after each restart.

Environment
Domoticz version:
Last known working (likely): ~17538
First observed broken: 17592
Still broken: 17732 (latest as of today)
Zigbee2MQTT: 2.7.2
MQTT broker: Mosquitto
Hardware: Raspberry Pi 4
Zigbee adapter: SONOFF Zigbee 3.0 USB Dongle Plus (EZSP)

### Steps to Reproduce

1. Use Zigbee2MQTT with multiple Light devices (e.g. Philips Hue GU10)
2. Restart Domoticz (tested from 17592 up to 17732)
3. Check logs:
    No MQTT Auto Discovery events for Light devices
4. Rename any Light device in Zigbee2MQTT
5. Observe:
    Immediate discovery in Domoticz logs

Workaround

Rename each Zigbee Light device in Zigbee2MQTT (even temporarily).

This forces Zigbee2MQTT to republish discovery data, and Domoticz correctly detects them.

### Expected Behavior

After restarting Domoticz:

All Zigbee2MQTT devices (including Lights) should be automatically rediscovered via MQTT Auto Discovery
No manual interaction should be required

### Hardware/Plugin Type

Zigbee2mqtt 2.7.2 Hardware: Raspberry Pi 4 Zigbee adapter: SONOFF Zigbee 3.0 USB Dongle Plus (EZSP)

### Browser

_No response_

### Log Output

```text
Logs

Example of discovery logs triggered ONLY after manual rename:

Status: MQTT Auto Discovery: Discovered: Lampada Cucina/Lampada Cucina ...
Status: MQTT Auto Discovery: Discovered: Lampada Sala/Lampada Sala ...

No equivalent entries appear after a normal Domoticz restart.
```

### Screenshots

_No response_

### Forum Post

_No response_

### Additional Context

Issue affects only Light devices
Other Zigbee entities (e.g. sensors) appear to be discovered correctly
Strong indication of a regression in MQTT Auto Discovery handling for Light entities on startup
Likely related to retained MQTT discovery topics not being reprocessed after restart

MQTT Auto Discovery continues to work correctly for other Zigbee devices, such as the IKEA Inspelning smart plug, which are properly rediscovered after a restart.

adrighem:
This PR fixes the issue where some MQTT devices (especially Zigbee2MQTT lights) wouldn't show up after a restart.

  It turns out there is a hardcoded limit of 1,000 messages in the incoming queue. In larger setups—like the one in #6754 which had ~1,100 discovery messages—the initial burst of retained messages on startup would
  overflow the queue and cause configuration messages to be dropped.

  Changes:
   - Increased the queue limit from 1,000 to 10,000.
   - Reduced the poll interval from 1s to 100ms to drain the queue faster.
   - Added a log warning if the queue ever hits the limit again to make future issues easier to spot.

  Fixes #6754

### DIFF
--- a/hardware/MQTTAutoDiscover.cpp
+++ b/hardware/MQTTAutoDiscover.cpp
@@ -77,14 +77,15 @@ MQTTAutoDiscover::MQTTAutoDiscover(const int ID, const std::string& Name, const 
 
 void MQTTAutoDiscover::on_message(const struct mosquitto_message* message)
 {
+	std::string topic = message->topic;
 	std::lock_guard<std::mutex> lock(m_inc_msg_mutex);
-	if (m_incoming_messages.size() > 1000)
+	if (m_incoming_messages.size() > 10000)
 	{
 		//Prevent flooding
+		Log(LOG_ERROR, "MQTT Auto Discover: Incoming message queue full! Dropping message on topic: %s", topic.c_str());
 		return;
 	}
 
-	std::string topic = message->topic;
 	std::string qMessage;
 	if (message->payload != nullptr && message->payloadlen > 0)
 		qMessage = std::string((char*)message->payload, (char*)message->payload + message->payloadlen);
@@ -6575,7 +6576,7 @@ bool MQTTAutoDiscover::StopHardware()
 
 void MQTTAutoDiscover::Do_Work()
 {
-	while (!IsStopRequested(1000))
+	while (!IsStopRequested(100))
 	{
 		if (m_bDisconnected)
 		{


gizmocuz:
Thanks for the PR.

You could also ask yourself, how is it possible I have 1100 retained messages?

They should only contain config objects.

So, I guess your setup is not correct. There is a discovery topic and a state topic, both have to be different
Also, you probably have retained enabled on state messages?
Both are not by default set in Zigbee2MQTT

gizmocuz:
Thanks again for the contribution — we've accepted this fix! We cherry-picked your commit and kept the queue limit increase (1,000 → 10,000) and the log warning, but reverted the polling interval back to 1,000ms.

The reason: the retained message burst arrives within milliseconds of connecting, well before either a 100ms or 1,000ms timer fires. So the reduced interval doesn't help with the startup drop — it only causes `Do_Work()` to wake up 10× more often during normal idle operation for no benefit. The queue limit increase alone is what actually fixes the issue.

We're planning to follow up with a `condition_variable`-based approach that wakes `Do_Work()` immediately when messages arrive, which solves the latency side cleanly without any CPU overhead.

gizmocuz:
Closing as the fix has been cherry-picked and merged into the development branch. Thanks again for the contribution!

Issue #6755, opened 2026-04-14
Since the use of /etc/systemd/system/domoticz.service the redirection of PYTHON scripts/plugins errors to std_out_err_log file is not working any longer


mealgi:
### Prerequisites

- [x] I have searched existing issues and the forum for this problem
- [x] I am running the latest beta version

### Domoticz Version

2026.1 V17730

### Platform

Linux (x86_64)

### OS Version

bookworm

### Describe the Bug

Since the use of /etc/systemd/system/domoticz.service the redirection of PYTHON scripts/plugins errors to std_out_err_log file is not working any longer.

When installing domoticz the preferred way (domoticz.service) domoticz is no more daemonized.

Thus the redirection of Stdin, Stdout, Stderr which is part of the daemonize method in domoticz.cpp is not performed.

The solution is to insert in domoticz.cpp line 1196 the following piece of source code to enable the redirection when it is required in the domoticz command line:

	else
	{
		if (!m_sz_std_out_err_log_file.empty())
		{
			int i;
			/* Close out the standard file descriptors */
			close(STDIN_FILENO);
			close(STDOUT_FILENO);
			close(STDERR_FILENO);

			/* Route I/O connections */

			/* Open STDIN */
			i = open(m_sz_std_out_err_log_file.c_str(), O_RDWR | O_APPEND | O_CREAT, S_IRUSR | S_IWUSR | S_IRGRP | S_IROTH);

			/* STDOUT */
			int dret = dup(i);
			if (dret == -1)
			{
				_log.Log(LOG_ERROR, "Could not set STDOUT descriptor !");
			}

			/* STDERR */
			dret = dup(i);
			if (dret == -1)
			{
				_log.Log(LOG_ERROR, "Could not set STDERR descriptor !");
			}
		}
 

### Steps to Reproduce

2. Click on '...'
3. See error

### Expected Behavior

Allow the redirection of PYTHON scripts/plugins errors when domoticz is installed the preferred way using domoticz.service

### Log Output

```text

```

mealgi:
@gizmocuz 
Fixed in PR #6758

mealgi:
@gizmocuz 

Hello
Could you consider committing the PR #6758 fixing the issue #6755 ?
Many thanks in advance.
BR

gizmocuz:
@mealgi , domoticz is installed in /etc/init.d ?

mealgi:
@gizmocuz 

[...]

PR #6799, opened 2026-04-28
Implement I/O redirection when not in daemon mode

mealgi:
Implement I/O redirection when not in daemon mode
Fix Issue #6755

### DIFF
--- a/main/domoticz.cpp
+++ b/main/domoticz.cpp
@@ -180,7 +180,6 @@ int pidFilehandle = 0;
 
 time_t m_LastHeartbeat = 0;
 
-
 #ifndef WIN32
 
 /* redirectIOconnections */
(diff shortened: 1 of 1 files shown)


gizmocuz:
domoticz.cpp(187,41): error C2146: syntax error: missing ')' before identifier 'or'

gizmocuz:
best to compile before committing

mealgi:
@gizmocuz 
PR is ready for merging

gizmocuz:
Great! Thank you for this. If possible undo that extra empty line on line 183

mealgi:
@gizmocuz 

Empty Line 183 has been removed
BR

gizmocuz:
The feature from this PR has been implemented directly on `development` in commit 5e844f981 with the following fixes applied over this approach:

- Log file is opened **before** closing stdin/stdout/stderr, so a failed `open()` returns cleanly without leaving the process with no I/O
- Uses `dup2()` instead of `dup()` for reliable FD targeting regardless of ordering
- Closes the opened FD after `dup2()` to avoid a file descriptor leak
- Removes the unused variable `i` from `daemonize()`

Closing in favour of that commit.

PR #6826, opened 2026-05-26
Implementation of Robot Vacuum Cleaner features in domoticz matter protocol

mealgi:
In Domoticz Matter protocol implementation 
Implement rechargeable batteries in cluster POWER_SOURCE
Implement cluster SERVICE_AREAS
Implement cluster RVC_RUN_MODE
Implement cluster RVC_CLEAN_MODE
Implement cluster RVC_OPERATIONAL STATE

### DIFF
--- a/hardware/Matter.cpp
+++ b/hardware/Matter.cpp
@@ -146,10 +146,10 @@ static constexpr int CHILD_FREQ         = 16;
 // rechargeable batteries
 static constexpr int CHILD_BAT_PERCENT_REMAINING  = 1;
 static constexpr int CHILD_BAT_CHARGE_STATE  = 2;
-#define  BatChargeState_kUnknown 0
-#define  BatChargeState_kIsCharging 1
-#define  BatChargeState_kIsAtFullCharge 2
-#define  BatChargeState_kIsNotCharging 3
+static constexpr int  BatChargeState_kUnknown = 0;
+static constexpr int  BatChargeState_kIsCharging = 1;
+static constexpr int  BatChargeState_kIsAtFullCharge = 2;
+static constexpr int  BatChargeState_kIsNotCharging = 3;
 
 // service areas
 static constexpr int CHILD_SUPPORTED_AREAS = 3;
@@ -161,47 +161,45 @@ static constexpr int CHILD_RVC_START_PAUSE_STOP  = 6;
 static constexpr int CHILD_RVC_RUN_MODE = 7;
 static constexpr int CHILD_RVC_OPERATIONAL_STATE = 8;
 static constexpr int CHILD_RVC_OPERATIONAL_ERROR = 9;
-static constexpr int CHILD_RVC_UPTIME  = 10;
-static constexpr int CHILD_RVC_TOTAL_OPERATIONAL_HOURS  = 11;
 
 // Selector switch RVC START_PAUSE_STOP
-#define rvcStartOrResume	10
-#define rvcPause			20
-#define rvcStop				30
+static constexpr int rvcStartOrResume = 10;
+static constexpr int rvcPause = 20;
+static constexpr int rvcStop = 30;
 
 // Operational State List in RVC
-#define OperationalState_kStopped 0
-#define OperationalState_kRunning 1
-#define OperationalState_kPaused 2
-#define OperationalState_kError 3
-#define OperationalState_kSeekingCharger 64
-#define OperationalState_kCharging 65
-#define OperationalState_kDocked 66
-#define OperationalState_kEmptyingDustBin 67
-#define OperationalState_kCleaningMop 68
-#define OperationalState_kFillingWaterTank 69
-#define OperationalState_kUpdatingMaps 70
+static constexpr int OperationalState_kStopped = 0;
+static constexpr int OperationalState_kRunning = 1;
+static constexpr int OperationalState_kPaused = 2;
+static constexpr int OperationalState_kError = 3;
+static constexpr int OperationalState_kSeekingCharger = 64;
+static constexpr int OperationalState_kCharging = 65;
+static constexpr int OperationalState_kDocked = 66;
+static constexpr int OperationalState_kEmptyingDustBin = 67;
+static constexpr int OperationalState_kCleaningMop = 68;
+static constexpr int OperationalState_kFillingWaterTank = 69;
+static constexpr int OperationalState_kUpdatingMaps = 70;
 
 // Operational Error List in RVC
-#define OperationalError_kNoError 0
-#define OperationalError_kUnableToStartOrResume 1
-#define OperationalError_kUnableToCompleteOperation 2
-#define OperationalError_kCommandInvalidInState 3
-#define OperationalError_kFailedToFindChargingDock 64
-#define OperationalError_kStuck 65
-#define OperationalError_kDustBinMissing 66
-#define OperationalError_kDustBinFull 67
-#define OperationalError_kWaterTankEmpty 68
-#define OperationalError_kWaterTankMissing 69
-#define OperationalError_kWaterTankLidOpen 70
-#define OperationalError_kMopCleaningPadMissing 71
-#define OperationalError_kLowBattery 72
-#define OperationalError_kCannotReachTargetArea 73
-#define OperationalError_kDirtyWaterTankFull 74
-#define OperationalError_kDirtyWaterTankMissing 75
-#define OperationalError_kWheelsJammed 76
-#define OperationalError_kBrushJammed 77
-#define OperationalError_kNavigationSensorObscured 78
+static constexpr int OperationalError_kNoError = 0;
+static constexpr int OperationalError_kUnableToStartOrResume = 1;
+static constexpr int OperationalError_kUnableToCompleteOperation = 2;
+static constexpr int OperationalError_kCommandInvalidInState = 3;
+static constexpr int OperationalError_kFailedToFindChargingDock = 64;
+static constexpr int OperationalError_kStuck = 65;
+static constexpr int OperationalError_kDustBinMissing = 66;
+static constexpr int OperationalError_kDustBinFull = 67;
+static constexpr int OperationalError_kWaterTankEmpty = 68;
+static constexpr int OperationalError_kWaterTankMissing = 69;
+static constexpr int OperationalError_kWaterTankLidOpen = 70;
+static constexpr int OperationalError_kMopCleaningPadMissing = 71;
+static constexpr int OperationalError_kLowBattery = 72;
+static constexpr int OperationalError_kCannotReachTargetArea = 73;
+static constexpr int OperationalError_kDirtyWaterTankFull = 74;
+static constexpr int OperationalError_kDirtyWaterTankMissing = 75;
+static constexpr int OperationalError_kWheelsJammed = 76;
+static constexpr int OperationalError_kBrushJammed = 77;
+static constexpr int OperationalError_kNavigationSensorObscured = 78;
 
 
 // Non-contiguous SystemModeEnum values in selector-level order (level = index * 10)
@@ -476,9 +474,6 @@ void CMatter::HandleNodeEvent(const Json::Value& node_event)
 	int nodeId = -1;
 	int endpointId = -1;
 	int clusterId = -1;
-	int eventId = -1;
-	int eventNumber = 0;
-	int priority = 0;
 	if (node_event.isObject())
 	{		
 		// Event Error management
@@ -488,12 +483,6 @@ void CMatter::HandleNodeEvent(const Json::Value& node_event)
 			endpointId = node_event["endpoint_id"].asInt();
 		if (node_event.isMember("cluster_id"))
 			clusterId = node_event["cluster_id"].asInt();
-		if (node_event.isMember("event_id"))
-			eventId = node_event["event_id"].asInt();
-		if (node_event.isMember("event_number"))
-			eventNumber = node_event["event_number"].asInt();
-		if (node_event.isMember("priority"))
-			priority = node_event["priority"].asInt();
 		if (node_event.isMember("data") && node_event["data"].isObject())
 		{
 			const Json::Value& data = node_event["data"];
@@ -506,7 +495,7 @@ void CMatter::HandleNodeEvent(const Json::Value& node_event)
 					int operationalError=errorState["errorStateId"].asInt();
 					std::string szOperationalErrorLabel=Getrvc_OperationalErrorLabel(operationalError);
 					if (operationalError!=0)
-						Log(LOG_ERROR, "Node %d Endpoint %d rvc_OperationalError %s", nodeId, endpointId,szOperationalErrorLabel.c_str());
+						Log(LOG_ERROR, "Node %d Endpoint %d Cluster %d rvc_OperationalError %s", nodeId, endpointId, clusterId, szOperationalErrorLabel.c_str());
 				}
 			}	
 		}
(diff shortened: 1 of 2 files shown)


gizmocuz:
Thank you for this PR — RVC support in the Matter driver is a useful addition and the direction is right. Before we can merge it, a few things need to be addressed:

## French log strings

"Valeur" (French for "Value") appears 4 times in log format strings and must be changed to English:

- `"RVC_CURRENT_RUN_MODE : Attrib %d Valeur %d "`
- `"RVC_CURRENT_CLEAN_MODE : Attrib %d Valeur %d "`
- `"RVC_OPERATIONAL_STATE : Attrib %d Valeur %d "`
- `"RVC_OPERATIONAL_ERROR : Attrib %d Valeur %d "`

## Code quality issues

**`#define` instead of `constexpr`**
`BatChargeState_k*`, `rvcStartOrResume`, `rvcPause`, `rvcStop`, and all `OperationalState_k*` / `OperationalError_k*` are `#define`s. The rest of the file uses `static constexpr int` — please make these consistent.

**Unused declared constants**
`CHILD_RVC_UPTIME = 10` and `CHILD_RVC_TOTAL_OPERATIONAL_HOURS = 11` are defined but never used anywhere in the PR. Please remove them or implement them.

**Dead code in `HandleNodeEvent`**
`eventNumber` and `priority` are read from JSON but never used. Please remove them.

**`ReadAttribute` is never called**
The new `ReadAttribute(int nodeId, const std::string& attributePath)` method is declared in the header and implemented, but never invoked anywhere. Is this intended for a follow-up PR? If so, please leave it out for now.

## Potential bugs

**CHILD offset collision risk**
`CHILD_BAT_PERCENT_REMAINING = 1` and `CHILD_BAT_CHARGE_STATE = 2` start at 1, which overlaps with child IDs used by other device types. Since `hasRechargeableBattery` is set whenever *any* endpoint has the `batChargeState` attribute (not just vacuums), this could create spurious extra sensors for non-vacuum Matter devices. Please verify there is no overlap with existing `CHILD_*` values for thermostats, HVAC, etc.

**`rvc_CurrentRunMode` index-vs-mode assumption**
The check `state.rvc_RunModeEntries[state.rvc_CurrentRunMode].mode != state.rvc_CurrentRunMode` assumes mode values equal their 0-based vector index. The Matter spec does not guarantee this. A robust implementation should search the entries by mode value rather than indexing directly.

**`nbSelectedAreas` hardcoded to 1**
There is a comment about "multi selection" but `nbSelectedAreas` is always 1. Either support multi-select properly or remove the dead loop.

**Area offset math**
`levelArea = (state.currentArea + 1) * 10` is non-obvious and assumes area IDs are sequential integers starting at 1. A comment explaining the mapping (or a cleaner approach) would help reviewers.

---

Once the above points are addressed this should be in good shape to merge. Thanks again for working on this!

mealgi:
@gizmocuz 

Regarding the point CHILD offset collision risk
"CHILD_BAT_PERCENT_REMAINING = 1 and CHILD_BAT_CHARGE_STATE = 2 start at 1, which overlaps with child IDs used by other device types. Since hasRechargeableBattery is set whenever any endpoint has the batChargeState attribute (not just vacuums), this could create spurious extra sensors for non-vacuum Matter devices. Please verify there is no overlap with existing CHILD_* values for thermostats, HVAC, etc."

In Matter protocol implementation for a given sensor or switch the DeviceID is NodeID*256+endpointID.
Regarding more specifically  a Percentage sensor it is created (as a General Type) with the unit 1.
This is the reason why I use unit 1 for the BatPercentageRemaining.
It has also a side effect because it is not possible to have more than one percentage sensor allocated to an nodeID/endpointID. 

Regarding the other points do you prefer to address them on your side ?

or do you want me to do it ?

BR

mealgi:
@gizmocuz 
I will update the source codes as mentioned shortly
BR

mealgi:
@gizmocuz

Source code updated to fix the Code quality issues and potential bugs
BR

gizmocuz:
Thanks for the updates — the easy code-quality issues (French log strings, `#define` → `constexpr`, dead code, unused constants) are all properly fixed. A few things still need attention before this can merge:

## Still outstanding

**1. `ReadAttribute` is still present but never called**
`ReadAttribute(int nodeId, const std::string& attributePath)` is declared in the header and implemented in `Matter.cpp`, but has no call sites. Please remove it for now; it can be added in a follow-up PR when it is actually wired up.

**2. Stale bounds check still assumes mode == index**
```cpp
if (state.rvc_CurrentRunMode < 0 || state.rvc_CurrentRunMode >= state.rvc_RunModeEntries.size())
```
This guard treats `rvc_CurrentRunMode` as a vector index. The actual Matter spec uses arbitrary mode tag values (e.g. `16384` for Idle), so a device reporting a valid mode number ≥ the entry count would silently skip the send. The check should be replaced with `if (state.rvc_RunModeEntries.empty())` (or removed), since the real lookup is now correctly done by value in the loop below.

**3. CleanMode selector level still assumes sequential 1-based mode IDs**
```cpp
int levelCleanMode = state.rvc_CurrentCleanMode * 10;
```
And the adjacent error log:
```cpp
if (i+1 != CleanModeEntry.mode)
    Log(LOG_ERROR, "... CleanMode incorrect %d <> %d", CleanModeEntry.mode, i);
```
If a device uses non-sequential mode IDs the selector level will be wrong (and the error log confirms the assumption). Please look up the current clean mode by value (same pattern as the corrected RunMode lookup) and derive the level from the found entry's position in the list.

**4. Area write-back sends positional index instead of actual `areaId`**
In `WriteToHardware`:
```cpp
int selectedArea = pSwitch->level / 10 - 1;
// ...
payload["newAreas"].append(static_cast<Json::UInt>(selectedArea));
```
`selectedArea` is the selector's 0-based position, but the Matter command expects the real `areaEntry.areaId`. For devices with non-sequential area IDs (e.g. 5, 12, 7) this will select the wrong area. Please change this to look up `state.areaEntries[selectedArea].areaId` and send that value.

**5. Idle-mode branch hardcodes `0` instead of using `rvc_IdleRunMode`**
```cpp
if (rvc_CurrentRunMode == 0)
{
    // case idle
```
The idle mode value is device-specific and stored in `rvc_IdleRunMode`. This condition should be `if (rvc_CurrentRunMode == rvc_IdleRunMode)` — otherwise Start/Resume will silently do nothing on any device whose idle mode is not `0`.

---

Items 4 and 5 would result in incorrect commands being sent to the device, so they do need to be fixed. Thanks again for working on this.

mealgi:
@gizmocuz 
All Still outstanding issues have been tackled in the latest source code changes

gizmocuz:
Thanks for the updates — items 1, 2, 4, and 5 are all properly fixed. Two things still need attention before this can merge:

## Still outstanding

**1. CleanMode selector level still uses mode value instead of position**

The code now does a value-based lookup (good), but still derives the selector level from `CleanModeEntry.mode * 10`:

```cpp
levelCleanMode = CleanModeEntry.mode * 10;   // still mode-value based
```

The `Rvc_CleanMode_Names` string is built positionally (`"Off|label0|label1|..."`), so Domoticz selector levels 10, 20, 30 correspond to positions 1, 2, 3. Using `mode * 10` as the level will mismatch for any device with non-sequential mode IDs (e.g., modes 0, 5, 12 → levels 0, 50, 120, none of which align with the `Off|A|B|C` label positions).

The same assumption is in `WriteToHardware`:
```cpp
if (pSwitch->level/10 == CleanModeEntry.mode)
payload["newMode"] = pSwitch->level/10;
```

Please derive the level from the **entry's position in the list** (same as the area selector does with `(i+2)*10`), and in `WriteToHardware` look up the actual mode ID via `rvc_CleanModeEntries[level/10 - 1].mode`.

---

**2. CHILD ID collision with existing values**

The new CHILD constants overlap with existing ones defined earlier in the same file:

```
CHILD_OCCUPANCY = 1   ← conflicts with CHILD_BAT_PERCENT_REMAINING = 1
CHILD_ONOFF     = 2   ← conflicts with CHILD_BAT_CHARGE_STATE = 2
CHILD_CONTACT   = 3   ← conflicts with CHILD_SUPPORTED_AREAS = 3
CHILD_CO2       = 4   ← conflicts with CHILD_RVC_CLEAN_MODE = 4
CHILD_CO        = 5   ← conflicts with CHILD_RVC_GO_HOME = 5
CHILD_PM25      = 6   ← conflicts with CHILD_RVC_START_PAUSE_STOP = 6
CHILD_PM10      = 7   ← conflicts with CHILD_RVC_RUN_MODE = 7
CHILD_FLOW      = 8   ← conflicts with CHILD_RVC_OPERATIONAL_STATE = 8
CHILD_LOCK      = 9   ← conflicts with CHILD_RVC_OPERATIONAL_ERROR = 9
```

The CHILD value is the `Unit` column in the database and must be unique **within the same endpoint**. `hasRechargeableBattery` is set whenever any endpoint advertises `batChargeState` — not only for robot vacuums — so a thermostat or air quality sensor with a rechargeable battery would get battery sensors at unit=1/2 that overwrite its `CHILD_OCCUPANCY`/`CHILD_ONOFF` entries.

Please renumber all new CHILD constants to start at **17** (after the existing `CHILD_FREQ = 16`).

mealgi:
@gizmocuz 
Regarding the point 2 as I mentioned previously the Bat_percent_remaining device will always be assigned the unit 1 because it is a general device type even if it is declared as CHILD_BAT_PERCENT_REMAINING 17

mealgi:
@gizmocuz 
For the text sensors the unit is also 1 but the deviceid in domoticz db is (nodeid *256 + endpointId ) <<8 + CHILD_SENSOR_ID

gizmocuz:
Thanks for the explanation on CHILD IDs — you're right. For percentage sensors the ChildID does not affect the DB key, and for text sensors the ChildID is folded into the DeviceID, so there is no real collision. I'll withdraw that point.

The one remaining issue is still **CleanMode level** from my last comment. `levelCleanMode = CleanModeEntry.mode * 10` ties the selector level to the mode's numeric value, but Domoticz selector levels must match label positions (10 = first label, 20 = second, etc.). For non-sequential mode IDs the selected level will point past the end of the label list. Please derive the level from the entry's position in `rvc_CleanModeEntries` instead, and update `WriteToHardware` accordingly to look up the actual mode ID from the entry at `level/10 - 1`.

gizmocuz:
Here is exactly what needs to change.

### In `_DetectAndSend` — use positional index for the level

Replace:
```cpp
int levelCleanMode=0;
std::string Rvc_CleanMode_Names   = "Off";
std::string Rvc_CleanMode_Actions = "";
bool flCurrentCleanModeFound=false;
for (const auto& CleanModeEntry : state.rvc_CleanModeEntries)
{
    if (state.rvc_CurrentCleanMode==CleanModeEntry.mode)
    {   
        flCurrentCleanModeFound=true;
        levelCleanMode=CleanModeEntry.mode*10;
    }
    else if (!flCurrentCleanModeFound)
        levelCleanMode=CleanModeEntry.mode*10;
    Rvc_CleanMode_Names  += "|" + CleanModeEntry.label;
    Rvc_CleanMode_Actions += "|";
}
```

With:
```cpp
int levelCleanMode=10;
std::string Rvc_CleanMode_Names   = "Off";
std::string Rvc_CleanMode_Actions = "";
bool flCurrentCleanModeFound=false;
for (int i = 0; i < (int)state.rvc_CleanModeEntries.size(); i++)
{
    const auto& CleanModeEntry = state.rvc_CleanModeEntries[i];
    if (state.rvc_CurrentCleanMode==CleanModeEntry.mode)
    {
        flCurrentCleanModeFound=true;
        levelCleanMode=(i+1)*10;
    }
    Rvc_CleanMode_Names  += "|" + CleanModeEntry.label;
    Rvc_CleanMode_Actions += "|";
}
if (!flCurrentCleanModeFound)
    Log(LOG_ERROR, "rvc CleanMode Node %d Endpoint %d DomoticzID %d Label %s CleanMode %d not consistent", nodeId, endpointId, domoticzID, state.label.c_str(), state.rvc_CurrentCleanMode);
```

The selector labels are always at positions 10, 20, 30… so the level must be `(index + 1) * 10`, not `mode * 10`.

---

### In `WriteToHardware` — look up mode ID by position, not by value

Replace:
```cpp
bool flCurrentCleanModeFound=false;
for (const auto& CleanModeEntry : state.rvc_CleanModeEntries)
{
    if (pSwitch->level/10==CleanModeEntry.mode)
    {   
        flCurrentCleanModeFound=true;
        break;
    }
}
if (!flCurrentCleanModeFound)
{   
    Log(LOG_ERROR, "WriteToHardware: node %d endpoint %d rvc clean mode %d not defined", nodeId, endpointId, pSwitch->level/10);
    return false;
}
```
```cpp
payload["newMode"] = pSwitch->level/10;
```

With:
```cpp
int posIndex = pSwitch->level/10 - 1;
if (posIndex < 0 || posIndex >= (int)state.rvc_CleanModeEntries.size())
{
    Log(LOG_ERROR, "WriteToHardware: node %d endpoint %d rvc clean mode index %d out of range", nodeId, endpointId, posIndex);
    return false;
}
int selectedCleanMode = state.rvc_CleanModeEntries[posIndex].mode;
```
```cpp
payload["newMode"] = selectedCleanMode;
```

`pSwitch->level / 10 - 1` converts the selector level back to a 0-based index into `rvc_CleanModeEntries`, then `.mode` gives the actual mode ID to send to the device.

mealgi:
@gizmocuz 
Hello
Many thanks for your reply
I will not be able to update the source code before the end of june 2026
You may proceed to the changes on your side if you feel like it.
Still I have one comment

The first time _DetectAndSend is called the data rvc_ CurrentCleanMode is 0 and you will get the error message which is not needed

once domoticz receives the clean mode attribute message the value of rvc_ CurrentCleanMode is set to a value >0 and in that case the error message is relevant

BR

gizmocuz:
Thanks @mealgi, and good catch on the initial state.

You're right: on the first `_DetectAndSend` pass `rvc_CurrentCleanMode` is still `0` (the attribute hasn't been received yet), so logging "not consistent" there would be a false alarm. I've guarded the error so it only fires once a real clean-mode value has arrived.

I've gone ahead and applied both changes to the PR branch myself so you don't have to wait until June:

**`_DetectAndSend` — positional level + guarded error**
```cpp
int levelCleanMode=10;
std::string Rvc_CleanMode_Names   = "Off";
std::string Rvc_CleanMode_Actions = "";
bool flCurrentCleanModeFound=false;
for (int i = 0; i < (int)state.rvc_CleanModeEntries.size(); i++)
{
    const auto& CleanModeEntry = state.rvc_CleanModeEntries[i];
    if (state.rvc_CurrentCleanMode==CleanModeEntry.mode)
    {
        flCurrentCleanModeFound=true;
        levelCleanMode=(i+1)*10;
    }
    Rvc_CleanMode_Names  += "|" + CleanModeEntry.label;
    Rvc_CleanMode_Actions += "|";
}
// rvc_CurrentCleanMode == 0 means the attribute hasn't been reported yet — not an error
if (!flCurrentCleanModeFound && state.rvc_CurrentCleanMode != 0)
    Log(LOG_ERROR, "rvc CleanMode Node %d Endpoint %d DomoticzID %d Label %s CleanMode %d not consistent", nodeId, endpointId, domoticzID, state.label.c_str(), state.rvc_CurrentCleanMode);
```

**`WriteToHardware` — index-based mode lookup**
```cpp
int posIndex = pSwitch->level/10 - 1;
if (posIndex < 0 || posIndex >= (int)state.rvc_CleanModeEntries.size())
{
    Log(LOG_ERROR, "WriteToHardware: node %d endpoint %d rvc clean mode index %d out of range", nodeId, endpointId, posIndex);
    return false;
}
selectedCleanMode = state.rvc_CleanModeEntries[posIndex].mode;
...
payload["newMode"] = selectedCleanMode;
```

The selector level now maps positionally to `rvc_CleanModeEntries` (level 10 = first entry), and `WriteToHardware` converts the level back to a 0-based index and sends the entry's actual `.mode`, so non-sequential mode IDs work correctly in both directions.

Thanks again for the work on this — RVC support is a great addition.

PR #6842, opened 2026-06-05
CustomChart widget - Add support for Voltage and Current sensor types

bblaszkiewicz:
The CustomChart widget only displays the message "No data to display" when selecting General/Voltage or General/Current devices. This change adds support for these types.


gizmocuz:
Thanks for the fix, @bblaszkiewicz! Verified: General/Voltage and General/Current devices return their data in the `d.v` field (not `d.te`), so the previous fall-through to the temperature handler produced an empty chart. Your branches using `sensor: 'counter'` and `field: d.v` with units V/A correctly match both the day-graph output and the counter routing for week/month/year ranges. Merging.

PR #6843, opened 2026-06-05
CustomChart widget - time range fix

bblaszkiewicz:
When selecting Time range: Last 24h, the CustomChart widget was showing data for the last 7 days. This change fixes this issue.


gizmocuz:
Thank you for the fix! Confirmed the root cause: the graph API returns the full short-log window with no date filter, and the CustomChart widget was missing the client-side `xAxis.min`/`max` windowing that the other chart widgets (e.g. `ddTemperatureGraph`) apply. Your change follows the established pattern. Merging.

PR #6848, opened 2026-06-11
gauge widget - color display fix

bblaszkiewicz:
Despite the value ranges set, the widget always displayed the color `'var(--dz-widget-energy-export)'` because the `ranges` object does not contain a `.status` field, which is checked.
```javascript
if (matched) {
            if (matched.status === 'critical') { return 'var(--dz-accent-red)'; }
            if (matched.status === 'warning')  { return 'var(--dz-widget-amber)'; }
            return 'var(--dz-widget-energy-export)';
        }
        return 'var(--dz-widget-amber)';
```
```
[
  {
    "from": 0,
    "to": 40,
    "color": "#04bbbb"
  },
  {
    "from": 41,
    "to": 60,
    "color": "#05bb04"
  },
  {
    "from": 61,
    "to": 70,
    "color": "#bbaa01"
  },
  {
    "from": 71,
    "to": 100,
    "color": "#bb1301"
  }
]
```
image

This change adds setting `matched.color`, which fixes the problem.
```javascript
if (matched) {
            if (matched.color) { return matched.color; }
            if (matched.status === 'critical') { return 'var(--dz-accent-red)'; }
            if (matched.status === 'warning')  { return 'var(--dz-widget-amber)'; }
            return 'var(--dz-widget-energy-export)';
        }
        return 'var(--dz-widget-amber)';
```

### DIFF
--- a/www/app/dashboardDynamic/widgets/ddGauge.widget.js
+++ b/www/app/dashboardDynamic/widgets/ddGauge.widget.js
@@ -112,6 +112,7 @@ define([
                             }
                         }
                         if (matched) {
+                            if (matched.color) { return matched.color; }
                             if (matched.status === 'critical') { return 'var(--dz-accent-red)'; }
                             if (matched.status === 'warning')  { return 'var(--dz-widget-amber)'; }
                             return 'var(--dz-widget-energy-export)';


matmar29:
Temp_Gauge
The Temperature/Humidity gauge of the Weather Station (Netatmo in my setup) shows the same problem as reported above.

The configured temperature range colors are ignored. For example, I assigned purple to temperatures between 19°C and 26°C, but the gauge always remains green instead of using the configured color.

gizmocuz:
Thank you for the fix, @bblaszkiewicz! You correctly identified that the `range-list` editor stores ranges as `{ from, to, color }` and never sets a `.status` field, so the gauge's per-range colors were always being ignored. This also brings the gauge widget in line with the sibling `ddDial` widget, which already uses the same `matched.color` check. Merging now.

PR #6849, opened 2026-06-11
CustomChart widget

kovainfo:
Add support for Percentag sensor types

### DIFF
--- a/www/app/dashboardDynamic/widgets/ddCustomChart.widget.js
+++ b/www/app/dashboardDynamic/widgets/ddCustomChart.widget.js
@@ -17,6 +17,13 @@ define([
         if (type.indexOf('Temp') >= 0 || sub.indexOf('Temp') >= 0) {
             return { sensor: 'temp', field: function(d){ return d.te !== undefined ? d.te : d.v; }, unit: '°' };
         }
+        if (type === 'Percentage' || sub === 'Percentage' || type.indexOf('Percent') >= 0) {
+            return {
+                sensor: 'Percentage', 
+                field: function(d) { return d.v !== undefined ? d.v : d.v1; },
+                unit: '%'
+            };
+        }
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thanks for the contribution, @kovainfo! 🙏 Verified against the backend graph handlers: `sensor="Percentage"` is correct and the `v` field matches the day-range output. Merging now. As a small follow-up we'll extend the field function to also read `v_avg` so the month/year ranges (which the backend returns as `v_min`/`v_max`/`v_avg`) render too — consistent with how `temp` (`te`) and `counter`/voltage (`v`) reuse a single key across ranges.

gizmocuz:
Follow-up: merged and applied a small refinement on `development` so the Percentage branch works across **all** time ranges:

```javascript
if (sub === 'Percentage') {
    // Day range returns `v`; month/year aggregates return v_min/v_max/v_avg.
    return {
        sensor: 'Percentage',
        field: function(d) { return d.v !== undefined ? d.v : d.v_avg; },
        unit: '%'
    };
}
```

What changed vs. the original PR:
- **`d.v1` → `d.v_avg`** — the month/year graph handler returns `v_min`/`v_max`/`v_avg` (not `v`/`v1`), so this makes the month and year ranges render instead of showing a blank line. This mirrors how `temp` (`te`) and `counter`/voltage (`v`) reuse a single key across ranges.
- **Condition tightened to `sub === 'Percentage'`** — Percentage devices are always `Type="General"`, `SubType="Percentage"`, so the `type === 'Percentage'` and `type.indexOf('Percent')` clauses were dead code and were removed.

Thanks again for adding Percentage support, @kovainfo!

gizmocuz:
The follow-up refinement is now committed and pushed to `development`: https://github.com/domoticz/domoticz/commit/e46e424bc

PR #6850, opened 2026-06-11
Fix casing in JSON key references for Buienradar

vespino:


gizmocuz:
Thank you for this PR and for spotting the problem — much appreciated! We verified it against the live API and you are right: the main feed (`data.buienradar.nl/2.0/feed/json`) has indeed switched to PascalCase keys, so the Buienradar hardware is currently broken for everyone.

However, while investigating we found a few issues that prevent merging the PR as-is:

1. **The individual-station fallback endpoint still uses lowercase keys.** `observations.buienradar.nl/1.0/actual/weatherstation/<id>` still returns `temperature`, `stationname`, `lat`, `regio`, `winddirectiondegrees`, etc. `ParseMeterDetails()` parses data from *both* sources, so converting all keys to PascalCase breaks the fallback path. The parser needs to accept both casings.

2. **`winddirectiondegrees` → `WindDirection` is not just a casing change.** The new feed contains both `WindDirection` (a compass index, e.g. `11`) and `WindDirectionDegrees` (actual degrees, e.g. `251`). The correct key is `WindDirectionDegrees`, otherwise the wind direction would be reported wrong.

3. The guard in `ParseMeterDetails` now checks `root["StationId"]` twice (`StationId && StationId`), losing the dual-casing fallback that was there intentionally.

4. **A deeper problem:** the new main feed no longer provides station coordinates or region — `Latitude`/`Longitude` are `0` and `Region` is `null` for all 40 stations. This silently breaks the nearest-station auto-discovery (and the rain prediction that depends on the station coordinates), so a key rename alone is not enough; the coordinates now have to be retrieved from the observations endpoint.

Because of point 4 the fix requires some restructuring, so we are going to implement the correction ourselves on the `development` branch. We will close this PR once that fix lands.

Thanks again for the report and the effort — it helped us catch this quickly!

gizmocuz:
The fix has landed on development in af75c37d7e294a8c (fix(buienradar): support new PascalCase API keys and restore station coordinate lookup) - it supports both key casings, uses WindDirectionDegrees, and restores the station coordinate lookup via the observations endpoint. Closing this PR as discussed. Thanks again for the report!

PR #6854, opened 2026-06-15
WindChart - time range fix

bblaszkiewicz:
Similar to CustomChart, for the selected `Last 24h` range, the WindChart showed data from the last week, because `range=day` returns data from 7 days. The `Last 24h` filter was fixed and the `Last 7 days` option was added.


gizmocuz (approved):
Thanks for the fix! This is the correct approach — using `xAxis.min/max` to clip the displayed time window to 24h for `shortlog` and 7 days for the new `week` type, exactly as done in the CustomChart and EnergyChart widgets.

PR #6855, opened 2026-06-15
Add Blockly blocks for device LastUpdate age

### LINKED ISSUE #6853: Expose device LastUpdate / last updated age in Blockly
### Prerequisites

- [x] I have searched existing issues for this feature request

### Description

Blockly currently does not appear to expose a device’s LastUpdate value, or the age of that timestamp, as something that can be used in conditions.

Domoticz already tracks and displays this information, and it can be useful for e.g. detecting stale or unresponsive devices. It would be helpful if Blockly could use this value directly, for example as a numeric
“time since last update” block.

### Use Case

I have a temperamental piece of hardware, a Steinel IS 140-2, that crashes every few weeks or months. It is connected through a switch, so recovery is simple: turn the switch off, wait briefly, then turn it
back on.

Today I can detect this kind of stale-device situation through Domoticz notifications or by writing scripts, but I would like to handle it directly in Blockly. For example:

- If last update of "Steinel IS140-2" is more than 24 hours ago
- Then turn "Steinel power switch" Off
- Wait 10 seconds
- Turn "Steinel power switch" On

This would make simple self-recovery automations possible without requiring Lua, dzVents, or external scripts.

### Proposed Solution

Add one or more Blockly blocks that expose a selected device’s last update age as a number, preferably in seconds, minutes, or hours.

For example:
[minutes] since last update of [device]
This could then be used with the existing comparison blocks.

### Alternatives Considered

- Use Domoticz notifications for stale devices, but that only alerts; it does not let Blockly take recovery action.
- Use dzVents/Lua or an external script, but this is more complex for a simple automation that otherwise fits Blockly well.

### Forum Discussion

https://forum.domoticz.com/viewtopic.php?f=62&t=44710

adrighem:
Adds Blockly Time blocks for checking how many seconds, minutes, or hours have passed since a device's `LastUpdate`. The blocks are populated from all used devices and generate Blockly Lua using the `device_lastupdate` table, allowing Time/All events to detect stale devices.

Closes #6853

Tested with `cmake --build . --target domoticz -j2` and verified the blocks in the UI.

Blockly device last update blocks

### DIFF
--- a/main/EventSystem.cpp
+++ b/main/EventSystem.cpp
@@ -36,12 +36,6 @@ struct ScriptNameReport {
 	std::string name;
 };
 
-struct LastUpdateTableEntry
-{
-	uint64_t id = 0;
-	std::string lastUpdate;
-};
-
 static int l_reportScriptName(lua_State *L)
 {
 	auto *report = static_cast<ScriptNameReport *>(lua_touserdata(L, lua_upvalueindex(1)));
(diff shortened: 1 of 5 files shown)


gizmocuz:
Thank you very much for this pr. but considering the changes vs dzVents, it is faster to do this in dzVents

PR #6860, opened 2026-06-22
Auto-refresh of CustomChart widget

bblaszkiewicz:
This PR adds automatic refresh handling to the Dynamic Dashboard Custom Chart widget.
Previously, CustomChart only loaded data on initialization, configuration changes, or refresh event. Time-based ranges such as Last 24h used Date.now() when rendering the x-axis, but the chart did not react to live device updates, so the visible time window could become stale while the dashboard remained open.


gizmocuz:
Thanks @bblaszkiewicz for the fix! This correctly adds the missing `device_update` listener and follows the same debounced-refresh pattern already used by the EnergyChart, TemperatureGraph and RainChart widgets. Merging.

PR #6861, opened 2026-06-25
Add ability to clear selected devices from widget settings

bblaszkiewicz:
**Problem**
`device-picker` fields currently do not provide a way to remove a previously selected device.
For example, when configuring a Custom Chart widget, users can select multiple devices to display. If one of those devices should later be removed from the chart, the only available workaround is to create a new widget and reselect all remaining devices.
The widget implementation already supports empty device slots, but the settings dialog does not expose a way to clear a selection.

**Solution**
Add an empty (`-- None --`) option to all `device-picker` dropdowns in the widget settings dialog.
Selecting this option clears the corresponding configuration value, allowing users to remove devices without recreating the widget.

### DIFF
--- a/www/app/dashboardDynamic/ddWidgetSettings.controller.js
+++ b/www/app/dashboardDynamic/ddWidgetSettings.controller.js
@@ -146,9 +146,16 @@ define([
                             var lb = filter ? b.Name : deviceLabel(b);
                             return la.localeCompare(lb);
                         });
-                        $scope.pickerOptions[field.key] = $scope.deviceListByField[field.key].map(function(d) {
-                            return { value: String(d.idx), label: field.deviceFilter ? d.Name : deviceLabel(d) };
-                        });
+                        $scope.pickerOptions[field.key] =
+                        [{ value: '', label: '-- None --' }]
+                        .concat(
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thanks for the contribution and for spotting this gap — being unable to clear a device-picker slot was a real limitation.

We've gone with a slightly different approach for this. Instead of injecting a `-- None --` entry into every `device-picker` dropdown, each picker now gets an explicit **remove button (trash icon)** next to it, shown only when the picker is *not required* and currently *has a value*.

Rationale:
- The `-- None --` option would appear in **every** device-picker dropdown — including required single-device widgets (Dial, Gauge, DzDevice, Setpoint, …), where clearing the only device makes no sense.
- An explicit trash button reuses the same idiom already used by the `action-list`, `device-list`, and `range-list` fields, so the "remove" affordance is consistent and visible without opening the dropdown.
- The button-based clear also drops any tied per-device metric override (`metricKey`), keeping the config clean.

This targets exactly the multi-device widgets (Custom Chart's Device 1–10, Battery Status, Weather) on their optional slots, while leaving required pickers untouched.

Closing this PR in favour of that implementation — but thank you, the underlying issue was a good catch.

Issue #5806, opened 2023-08-31
Battery status not correctly updated from Z-Wave JS UI for FGSD002


Rouzax:
Domoticz Version: 2023.2 (build 15512)
zwave-js-ui: 8.23.1.77058e8
zwave-js: 11.13.0

The state of the batteries in my Fibaro devices do not seem to get updated correctly, as an example, the smoke detectors
This keeps triggering low battery notifications which are not correct.

In Domoticz
![image](https://github.com/domoticz/domoticz/assets/4103090/f5e86e91-6b4c-4883-af07-8f84749f0a98)

In Zwave JS UI
![image](https://github.com/domoticz/domoticz/assets/4103090/fa0b201f-4186-45bd-81d1-c48c27f2f85d)

## MQTT Explorer:
zwave/Game_Room/Server_Sensor/128/0/level
```json
{"time":1693465158550,"value":65,"nodeName":"Server Sensor","nodeLocation":"Game Room"}
```
zwave/Game_Room/Server_Sensor/128/0/isLow
```json
{"time":1693465158554,"value":false,"nodeName":"Server Sensor","nodeLocation":"Game Room"}
```

homeassistant/sensor/Game_Room-Server_Sensor/battery_level/config
```json
{
    "value_template": "{{ value_json.value }}",
    "unit_of_measurement": "%",
    "device_class": "battery",
    "state_topic": "zwave/Game_Room/Server_Sensor/128/0/level",
    "json_attributes_topic": "zwave/Game_Room/Server_Sensor/128/0/level",
    "device": {
        "identifiers": [
            "zwavejs2mqtt_0xc15d8aa6_node37"
        ],
        "manufacturer": "Fibargroup",
        "model": "Smoke Detector (FGSD002)",
        "name": "Game Room-Server Sensor",
        "sw_version": "3.3"
    },
    "name": "Game Room - Server Sensor - battery_level",
    "unique_id": "zwavejs2mqtt_0xc15d8aa6_37-128-0-level"
}
```

homeassistant/binary_sensor/Game_Room-Server_Sensor/hardware_status/config
```json
{
    "payload_on": 1,
    "payload_off": 0,
    "value_template": "{{ value_json.value }}",
    "device_class": "problem",
    "state_topic": "zwave/Game_Room/Server_Sensor/113/0/System/Hardware_status",
    "json_attributes_topic": "zwave/Game_Room/Server_Sensor/113/0/System/Hardware_status",
    "device": {
        "identifiers": [
[...]

PR #6864, opened 2026-06-26
fix(EventSystem): keep dzVents battery and signal level in sync with the database

Rouzax:
## Problem

dzVents scripts read a stale battery level (and signal level) for any device updated through the generic update path (MQTT Auto Discovery, Python plugins, scripts). `device.batteryLevel` returns `nil` even though the `DeviceStatus.BatteryLevel` column and the JSON API (`getdevices`) hold the correct value. This is the dzVents side of #5806.

## Root cause

dzVents builds its Lua device table from the in-memory `CEventSystem::m_devicestates` map, reading `_tDeviceStatus.batteryLevel` directly (`main/dzVents.cpp`). That field is populated from the database only once, at startup (`GetCurrentStates`). The generic runtime refresh inside `CEventSystem::UpdateSingleState` was left commented out by the cleanup in 4eb71a55e ("Cleanup Energy Dashboard"):

```cpp
//replaceitem.batteryLevel = batteryLevel;
...
//newitem.batteryLevel = batteryLevel;
```

So for every non-RFXCom device the DB column updates but the value dzVents reads never does. Only the RFXCom path refreshes it, via the dedicated `UpdateBatteryLevel()` stopgap. The in-memory `255` ("no battery") sentinel is then surfaced to scripts as `nil` by the dzVents adapter.

`signalLevel` has the same latent issue: it is loaded at startup and never refreshed (it never had a parameter on `UpdateSingleState`).

## Fix

- Restore the two `batteryLevel` assignments in `UpdateSingleState`, **unconditionally**. The DB columns are written unconditionally with the same values `ProcessDevice` passes to `UpdateSingleState` (`SQLHelper.cpp`), so the in-memory struct must mirror the DB; a guard would let the two diverge.
- Add a `signalLevel` parameter to `UpdateSingleState` and assign it alongside, so signal level is kept in sync too.

dzVents is the only consumer of these in-memory struct fields, and assigning them does not trigger scenes/scripts (that is driven by `nValue`/`sValue`), so there is no behavior change beyond dzVents seeing live values.

## Testing

- Builds cleanly.
- Verified on a local instance: pushing runtime battery/signal updates via `udevice` makes a dzVents script read the live values immediately:

```
batteryLevel=nil signalLevel=12   (before any update)
batteryLevel=50  signalLevel=8    (after udevice battery=50 rssi=8)
batteryLevel=77  signalLevel=5    (after a second update)
```

Before this change the first state persisted (`nil`) regardless of later updates.

Part of #5806 (the dzVents side). Follow-ups for the MQTT Auto Discovery battery attribution are planned as separate PRs.

### DIFF
--- a/main/EventSystem.cpp
+++ b/main/EventSystem.cpp
@@ -1443,8 +1443,9 @@ std::string CEventSystem::UpdateSingleState(
 	const unsigned char devType, const unsigned char subType, 
 	const _eSwitchType switchType, 
 	const std::string &lastUpdate, 
-	const unsigned char lastLevel, 
+	const unsigned char lastLevel,
 	const unsigned char batteryLevel,
+	const unsigned char signalLevel,
 	const std::map<std::string, std::string> & options
 )
 {
@@ -1463,7 +1464,8 @@ std::string CEventSystem::UpdateSingleState(
 	{
 		_tDeviceStatus replaceitem = itt->second;
 		replaceitem.deviceName = l_deviceName;
-		//replaceitem.batteryLevel = batteryLevel;
+		replaceitem.batteryLevel = batteryLevel;
+		replaceitem.signalLevel = signalLevel;
 		if (nValue != -1)
 			replaceitem.nValue = nValue;
 		if (!sValue.empty())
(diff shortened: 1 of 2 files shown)


gizmocuz:
Thanks @Rouzax for the thorough write-up and the clean, minimal fix — the root-cause analysis (DB column vs. in-memory `m_devicestates` divergence after 4eb71a55e) is spot on, and keeping the assignments unconditional so the in-memory struct mirrors the unconditionally-written DB columns is exactly right. Verified the diff compiles (both `UpdateSingleState` call sites updated, header signature matched) with no performance impact. Merging.

PR #6865, opened 2026-06-26
fix(MQTTAutoDiscover): apply Z-Wave node battery to all its devices immediately

Rouzax:
## Problem

For Z-Wave nodes ingested via MQTT Auto Discovery (zwave-js-ui), a node's battery level lands on its many Domoticz devices inconsistently, and some devices show a stale `0%` (or the `255` "no battery" default) long after the battery actually recovered. This is the MQTT Auto Discovery side of #5806.

Observed on a live system: nodes that zwave-js reports at 100% show, in Domoticz, `0` or `255` on the rarely-updating per-node devices (the `isLow` / Notification-CC devices), while the frequently-updating temperature sensor of the same node reads correctly.

## Root cause

`handle_auto_discovery_battery` writes the battery level (from the Battery CC, `128/0/level`) only into the in-memory `BatteryLevel` of every sensor of the node, then returns. The database `BatteryLevel` column for a device is only written later, when **that** device next runs its own state handler. So each device row holds whatever the node's battery was the last time that device emitted state. Devices that update at different cadences (a temperature sensor every few minutes vs a rarely pressed scene button or an `isLow` event) therefore show battery snapshots from different points in time, and a device that has not emitted state since the battery changed stays at an old value (a genuine low `0`, or the `255` default if it updated before any battery message was seen).

## Fix

Keep the existing in-memory update, and additionally, when a battery level arrives, eagerly persist it to **every already-created Domoticz device row of that node**:

- Match both DeviceID forms a node's devices use: standalone devices keyed by `unique_id`, combined temp/hum/baro devices keyed by `device_identifiers`.
- De-duplicate by row ID, and write only when the stored value actually changed.
- Update `BatteryLevel` only (never `nValue`/`sValue`) and refresh the event system via `m_eventsystem.UpdateBatteryLevel`, mirroring how the RFXCom path keeps the DB and in-memory state in sync. Because `nValue`/`sValue` are untouched, this does not trigger scenes or scripts.

Result: all of a node's devices reflect the same, current battery immediately, and stale `0%`/`255` rows self-correct on the next battery report.

This is purely additive: device creation, DeviceID, Type/SubType, and the discovery format are unchanged.

## Notes

- Only the numeric Battery CC `level` sensor feeds this path; `isLow` is published as a separate `binary_sensor` and is unaffected (see zwave-js-ui's discovery generation: [`api/lib/Gateway.ts` L1639-1667](https://github.com/zwave-js/zwave-js-ui/blob/v11.21.1/api/lib/Gateway.ts#L1639-L1667), where `level` is a `sensor` and `isLow` is emitted as a `binary_sensor`).
- Builds cleanly.

Part of #5806 (the MQTT Auto Discovery side). Complements the dzVents-side fix in #6864. A small follow-up to route battery detection by `device_class` (rather than hardcoded `object_id` strings) is planned as a separate PR.

### DIFF
--- a/hardware/MQTTAutoDiscover.cpp
+++ b/hardware/MQTTAutoDiscover.cpp
@@ -2632,19 +2632,66 @@ void MQTTAutoDiscover::handle_auto_discovery_battery(_tMQTTASensor* pSensor, con
 	if (is_number(pSensor->last_value))
 	{
 		iLevel = atoi(pSensor->last_value.c_str());
-	} else {
-		//could be a boolean isLow indicator
+	}
+	else
+	{
+		//could be a boolean isLow indicator (dead path for zwave-js-ui, which sends isLow as a binary_sensor)
 		if (pSensor->last_value == "true")
 			iLevel = 0;
 	}
 
+	//Update the in-memory battery level for every sensor of this node, and collect the distinct
+	//DeviceID forms its device rows use (standalone = unique_id, combined temp/hum/baro =
+	//device_identifiers) so all of the node's rows can be updated in a single statement.
+	std::set<std::string> deviceIDs;
 	for (auto& itt : m_discovered_sensors)
 	{
 		_tMQTTASensor* pDevSensor = &itt.second;
 		if (pDevSensor->device_identifiers == pSensor->device_identifiers)
 		{
 			pDevSensor->BatteryLevel = iLevel;
+			deviceIDs.insert(pDevSensor->unique_id);
 		}
+	}
+	deviceIDs.insert(pSensor->device_identifiers);
+
+	//Build a quoted, comma-separated IN() list. DeviceIDs come from the broker, so escape each as
+	//a SQL string literal (double any single quote) before embedding.
+	std::string inList;
+	for (const auto& id : deviceIDs)
+	{
+		if (!inList.empty())
+			inList += ",";
+		inList += "'";
+		for (const char c : id)
+		{
+			if (c == '\'')
+				inList += '\'';
+			inList += c;
+		}
+		inList += "'";
+	}
+
+	//Persist the new battery level to every already-created device row of this node, so all of the
+	//node's devices show the current battery immediately instead of lazily updating only when each
+	//device next emits state. First select just the rows that will actually change (none in steady
+	//state, so this is a cheap indexed no-op), then update them in one statement and refresh the
+	//event system for exactly those rows. We touch BatteryLevel only (never nValue/sValue), so this
+	//does not trigger scenes/scripts.
+	std::vector<std::vector<std::string> > results = m_sql.safe_query(
+		"SELECT ID FROM DeviceStatus WHERE (HardwareID==%d) AND (BatteryLevel<>%d) AND (DeviceID IN (%s))",
+		m_HwdID, iLevel, inList.c_str());
+	if (results.empty())
+		return;
+
+	m_sql.safe_query(
+		"UPDATE DeviceStatus SET BatteryLevel=%d WHERE (HardwareID==%d) AND (BatteryLevel<>%d) AND (DeviceID IN (%s))",
+		iLevel, m_HwdID, iLevel, inList.c_str());
+
+	for (const auto& sd : results)
+	{
+		uint64_t rowID = std::stoull(sd[0]);
+		m_mainworker.m_eventsystem.UpdateBatteryLevel(rowID, (unsigned char)iLevel);
 	}
 }
 


Rouzax:
Verified locally with an A/B test (a build of this branch vs. an unpatched `development` build), using a local Mosquitto broker and the MQTT Auto Discovery gateway. The test node exposed a temperature sensor, an illuminance sensor, and a `battery_level` sensor sharing one `device_identifiers`.

**With this change:** with the temperature and illuminance devices both showing `BatteryLevel = 10`, I published **only** the battery state topic, changing it to `100` (no update to the temperature/illuminance topics):

```
before:  temperature BatteryLevel=10   illuminance BatteryLevel=10
publish: test/.../battery = {"value":100}     (temperature/illuminance NOT republished)
after:   temperature BatteryLevel=100  illuminance BatteryLevel=100   (LastUpdate unchanged)
```

Both devices reflected `100` immediately, and their `LastUpdate` timestamps were **unchanged**, confirming only `BatteryLevel` was written (no `nValue`/`sValue`, so no scenes/scripts fire).

**Without this change (unpatched `development`):** the same two devices were created at `255` ("no battery") and **stayed at `255`** even after the battery was published as `100`, because the value is only written lazily when each device next emits its own state:

```
publish: test/.../battery = {"value":100}
after:   temperature BatteryLevel=255  illuminance BatteryLevel=255   (unchanged)
```

gizmocuz:
Thanks @Rouzax — the diagnosis is exactly right (per-device `BatteryLevel` rows drift because each one is only written when *that* device next emits state), and the fix is correct in behavior. Before merging I'd like to reshape the implementation, because as written its cost scales with network size in a way that will hurt larger MQTT/Z-Wave setups.

## Concern

Per battery message the current version does:

1. an outer scan over **all** `m_discovered_sensors` (every sensor of every node on the bridge) — `O(S_total)`, and
2. a **separate `SELECT`** for every sensor matching the node — `O(S_node)` queries, then a conditional `UPDATE`.

The issue is step 2 runs **every time, even when the level is unchanged**. Battery percentage is one of the most stable values a node publishes — zwave-js-ui re-emits the same value on every wake-up/interval — so the per-row `if (stored == iLevel) continue;` guard only avoids the `UPDATE`; it still issues all `O(S_node)` SELECTs on every identical report. On a network with many nodes × many devices/node × frequent battery republishes, that's a steady stream of SELECTs (all serializing on the SQLite mutex against every other update) for state that almost never changes — i.e. the most work happens precisely where there is nothing to do.

## Proposed redesign

Three changes, in priority order:

**1. Short-circuit at the node level on change (the big win).** Capture the battery sensor's previous in-memory `BatteryLevel` *before* the existing propagation loop overwrites it. If the incoming level equals it, do only the cheap in-memory propagation and `return` — **zero SQL**. No new state is needed; the previous value already lives in `m_discovered_sensors` (the `255` default means the first real reading still propagates once, which is what we want). This alone removes ~all DB work in steady state.

**2. One bulk `UPDATE` instead of N queries.** You already walk the node's sensors in memory for the in-memory propagation — collect their `DeviceID`s there (both the `unique_id` and `device_identifiers` forms) and issue a single statement:

```sql
UPDATE DeviceStatus SET BatteryLevel=?
WHERE HardwareID=? AND BatteryLevel<>? AND DeviceID IN (...)
```

The `BatteryLevel<>?` clause makes it touch only the rows that actually differ, so the self-correction of stale `0`/`255` rows still happens.

**3. Sync the event system only for changed rows.** `UpdateBatteryLevel` is keyed by numeric row ID, which isn't cached today, so this is the one spot still needing a lookup — but now only on a real change. Either a single `SELECT` of the affected IDs after the bulk update, then loop `UpdateBatteryLevel` (in-memory only), or — cleaner long-term — cache the device `RowIdx` on `_tMQTTASensor` at creation/resolve time, after which this path needs zero SELECTs and several other `DeviceID`-keyed queries in this file benefit too.

### Net effect

| | steady state (unchanged) | actual battery change |
|---|---|---|
| current | `O(S_node)` SELECTs | `O(S_node)` SELECTs + UPDATEs |
| proposed | **0 SQL** | 1 `UPDATE` (+ ≤1 `SELECT`, or 0 with cached RowIdx) |

Sketch of the hot path:

```cpp
const int prevLevel = pSensor->BatteryLevel;   // before the propagation loop

// ... existing loop: propagate iLevel to in-memory BatteryLevel of the node's sensors,
//     and collect each sensor's DeviceID(s) into a list ...

if (prevLevel == iLevel)
    return;                                     // node-level unchanged -> no DB churn

// changed: one bulk UPDATE over the collected DeviceIDs (BatteryLevel<>iLevel guard),
// then refresh the event system for the affected rows only.
```

Behavior is identical to your version (all of a node's devices converge to the current battery, stale rows self-correct); only the cost model changes. Happy to take the cleaner cached-`RowIdx` variant if you'd prefer — let me know and I can push a revision, or you can fold this in. The dzVents side (#6864) and `device_class` detection (#6866) are already merged, so this is the last piece.

Rouzax:
Thanks @gizmocuz, fully agree the cost model is the problem here, and that running `O(devices-per-node)` SELECTs on every battery republish (when battery is about the most stable value a node emits) is the wrong place to spend work. Parts 2 and 3 are exactly right and I'll fold them in.

On **Part 1** (the node-level short-circuit), I think it has an edge case that can leave a row permanently stale, so I'd like to propose a slightly different shape that keeps the steady-state win without it.

The short-circuit gates on the battery sensor's own `pSensor->BatteryLevel`, but device rows aren't created from that value. Each row is created from *its own* sensor's `BatteryLevel` field, which defaults to `255` (`MQTTAutoDiscover.h`), and the propagation loop only touches sensors already in `m_discovered_sensors` when the battery message arrives. So:

1. Battery sensor known; `battery=80` arrives and propagates to the sensors discovered so far.
2. A temperature sensor for the same node is discovered **later**. zwave-js-ui publishes each value on its own topic, so its first state carries no inline `battery`, and its row is created at the `255` default.
3. The next `battery=80` (unchanged) hits the short-circuit (`prevLevel == iLevel`) and returns, so that row stays at `255` until the level *actually* changes (which for a healthy battery can be a very long time). The dzVents-visible value is stranded the same way, since `UpdateBatteryLevel` isn't called for it.

The current per-message DB pass self-heals this on the very next report; the short-circuit doesn't. With retained-message ordering on (re)subscribe this isn't a corner case: I hit exactly this "row created at 255" ordering while testing locally.

Proposed shape that keeps it cheap and still self-heals (your Parts 2 and 3, minus the early return):

- Collect the node's `DeviceID`s (both `unique_id` and `device_identifiers` forms, de-duplicated) during the in-memory propagation loop.
- One `SELECT ID FROM DeviceStatus WHERE HardwareID=? AND BatteryLevel<>? AND DeviceID IN (...)` for the exact set of rows that will change (**zero rows in steady state**, so it's a cheap indexed no-op).
- If non-empty: one bulk `UPDATE ... SET BatteryLevel=? WHERE HardwareID=? AND BatteryLevel<>? AND DeviceID IN (...)`, then loop `UpdateBatteryLevel` over just those IDs.

That's **1 SELECT per message** (returning nothing in steady state) instead of N, a single UPDATE only on a real change, and because it still runs every message a row created stale at `255` is corrected on the next report. Each IN-list element quoted with its own `%q`.

If you'd rather have literally **zero SQL** in steady state, the safe way to get there is a per-node "may-have-stale-rows" dirty flag set whenever a device row is inserted and cleared after a clean pass. Happy to do that variant instead, but the single bulk statement above is much simpler and the steady-state SELECT is essentially free. I'll push a revision along these lines.

Rouzax:
Let me know which way you lean.

gizmocuz:
@Rouzax ,

Proposed shape that keeps it cheap and still self-heals (your Parts 2 and 3, minus the early return):

Collect the node's DeviceIDs (both unique_id and device_identifiers forms, de-duplicated) during the in-memory propagation loop.
One SELECT ID FROM DeviceStatus WHERE HardwareID=? AND BatteryLevel<>? AND DeviceID IN (...) for the exact set of rows that will change (zero rows in steady state, so it's a cheap indexed no-op).
If non-empty: one bulk UPDATE ... SET BatteryLevel=? WHERE HardwareID=? AND BatteryLevel<>? AND DeviceID IN (...), then loop UpdateBatteryLevel over just those IDs.

gizmocuz:
Unless literally zero SQL in steady state is better

Rouzax:
@gizmocuz 
Pushed a revision implementing exactly this shape (your Parts 2 and 3): collect the node's DeviceIDs (both `unique_id` and `device_identifiers`, de-duplicated) during the in-memory propagation loop, run one `SELECT ID ... WHERE BatteryLevel<>? AND DeviceID IN (...)` for the rows that will actually change (zero in steady state), then a single bulk `UPDATE` and a `UpdateBatteryLevel` loop over just those IDs. Each IN-list element is escaped as a SQL literal (doubling single quotes) since the DeviceIDs come from the broker.

Why this route rather than the zero-SQL dirty flag: the steady-state `SELECT` is one indexed lookup that returns no rows, so the cost is already negligible, and keeping the path stateless avoids a per-node flag whose create/delete lifecycle is one more thing to keep correct. It also self-heals without the short-circuit: a device discovered after the node reached steady state is created at the `255` default and gets corrected on the next battery message, whereas a node-level early return would strand it.

Verified locally (build of this branch + the test broker):

```
1) battery applied to all of a node's devices on discovery     PASS
2) a live level change updates every node device (SELECT/UPDATE) PASS
3) a device added AFTER steady state, created at 255, self-heals
   to the current level on the next (unchanged) battery message  PASS
```

Happy to switch to the dirty-flag variant if you'd prefer literally zero SQL in steady state, but this keeps it simpler.

PR #6877, opened 2026-06-30
feat(mcp): add get_scene_history tool, and stop get_sensor_history advertising scenes

Rouzax:
### Summary

Adds a new MCP server tool **`get_scene_history`** that returns a scene or group's On/Off activation log, including the user/actor that triggered each change, read from the `SceneLog` table. Also removes never-functional "scene support" claims from `get_sensor_history`.

### Why

The MCP server can list scenes (`get_scenes`), switch them (`switch_scene`), and list their member devices (`get_scene_devices`), but it could not read a scene's activation history with attribution.

`get_sensor_history` did not fill this gap: it resolves only `DeviceStatus` devices, so a scene is not found by name, and a scene idx collides with the device of the same id. It also contained an unreachable `Type=="Scene"/"Group"` branch that queried the wrong table: scene activations are written only to `SceneLog`, never to `LightingLog`.

### Changes

- **New `get_scene_history` tool** (scene-tool family, keyed on `scene_id`/`scenename`), querying `SceneLog` joined to `Scenes`. It mirrors `get_sensor_history`'s filtering (`count`, or `days` / `start_date`+`end_date`) and line format (`Date  On/Off  (user: ...)`), and renders the actor only when present.
- **`get_sensor_history` cleanup:** removed the unreachable, wrong-table scene branch, and dropped the inaccurate "handles scenes" wording from the tool description, a code comment, and `docs/mcp.wiki`, pointing scene history at the new tool.
- Docs updated in `docs/mcp.wiki` (new `get_scene_history` row; corrected `get_sensor_history` / `get_sensor_short_log` rows).

### Behavior and compatibility

- The new tool is additive and read-only.
- `get_sensor_history` behavior for real devices is unchanged. Because it never actually returned scene history (the scene path was unreachable and read the wrong table), removing the claim is a documentation/accuracy fix, not a functional regression.

### Validation

- Compiles cleanly (same toolchain as CI).
- Live-verified against a copy of a real database: `get_scene_history` by `scene_id` and by `scenename`, with `count` and date-range filters, plus the not-found and empty-history cases; and `get_sensor_history` for real switches confirmed unchanged.


gizmocuz:
Implemented on `development` in 479787521, taking a slightly more generic approach than the original proposal.

Rather than adding a standalone scene-only path, the change introduces a shared `buildEventHistory()` helper that reads an on/off/dim/text activation log from either `LightingLog` (devices) or `SceneLog` (scenes/groups), consolidating the count and date-range query/format logic that was duplicated in `get_sensor_history`.

On top of that:

- **New `get_scene_history` tool** — keyed on `scene_id`/`scenename`, returns a scene or group's On/Off activation log with the triggering user, read from `SceneLog` (group vs scene labelled via `SceneType`).
- **`get_sensor_history` now routes via `IsEventSwitchLike`** instead of `IsLightOrSwitch`, so text/alert devices return their `LightingLog` event log instead of being misrouted to the sensor graph branch (a pre-existing bug).
- **Scenes/groups passed to `get_sensor_history` are redirected** to `get_scene_history` rather than falling through to a `DeviceStatus` lookup. Note: contrary to the original analysis, scenes *are* returned by `GetJSonDevices` (with `Type`=Scene/Group, `idx`=Scenes.ID), so the old scene branch was reachable — but it queried `LightingLog` with the scene's idx (the wrong table) and could collide with a device of the same idx, so it never returned correct scene history. The redirect resolves both the wrong-table issue and the namespace collision.
- **Docs** updated in both `docs/mcp.wiki` and `mcpserver/mediawiki.txt` (new `get_scene_history` row; corrected `get_sensor_history` / `get_sensor_short_log` wording).

Closing as implemented.

PR #6879, opened 2026-06-30
Extend Netatmo Scopes

JurgenLB:
There are more scopes available from the Netatmo companies.
The Doorlock is tested and working.

The other devices are not yet tested because I don't have them,
If there are people willing to test en expand the devices,
I willing to help with such cases.

Kind Regards

jurgen


gizmocuz:
Thanks a lot for extending the Netatmo scopes, @JurgenLB! Great work adding support for all these additional device types (doorlock, BTicino, hybrid, and the rest), and thanks for testing the doorlock. Much appreciated — merging now. 🎉

Issue #6878, opened 2026-06-30
Custom date-range graphs return no data (and log `no such column: Value`) for MultiMeter_Calendar counter devices


Rouzax:
## Summary

The custom date-range graph handler (`HandleGraphCustomRange`) only implements the
`counter` branch correctly for `pTypeP1Power`. Every other counter device type whose
daily history lives in **`MultiMeter_Calendar`** (Current/CM113, CurrentEnergy/CM180i,
Air Quality, Lux, Weight, Usage, and a number of `pTypeGeneral` subtypes) falls through
to a generic `SELECT Value, Date FROM <table>` query. `MultiMeter_Calendar` has no
`Value` column, so the query fails server-side and the graph comes back empty.

This affects any caller of the custom date-range graph path:
- the web UI custom date-range graph (`json.htm?type=command&param=graph&...&range=YYYY-MM-DDTYYYY-MM-DD`), and
- the MCP `get_sensor_history` tool, which delegates to the same handler.

The month/year handler (`HandleGraphMonthYear_Counter`) already implements the correct
per-device-type column selection for all of these types. The custom-range handler is an
incomplete sibling.

Found on Domoticz **2026.2** (`development`).

## Affected device types

`CalcDbasetableCustom()` maps these to `MultiMeter_Calendar`
(`main/WebServerHandleGraphCustom.cpp:44-66`):

- `pTypeCURRENT` (CM113 / Electrisave)
- `pTypeCURRENTENERGY` (CM180i)
- `pTypeAirQuality`
- `pTypeLux`
- `pTypeWEIGHT`
- `pTypeUsage`
- `pTypeGeneral` + `sTypeVisibility` / `sTypeDistance` / `sTypeSolarRadiation` /
  `sTypeSoilMoisture` / `sTypeLeafWetness` / `sTypeVoltage` / `sTypeCurrent` /
  `sTypePressure` / `sTypeSoundLevel`
- `pTypeRFXSensor` + `sTypeRFXSensorAD` / `sTypeRFXSensorVolt`

Regular meters (energy/gas/water/counter) map to `Meter_Calendar`, which *does* have a
`Value` column, so they are unaffected.

## Root cause

In `main/WebServerHandleGraphCustom.cpp`, the `counter` branch handles only `pTypeP1Power`
specially; everything else uses a single-column read:

```cpp
// main/WebServerHandleGraphCustom.cpp  (counter branch)
if (dType == pTypeP1Power)
{
[...]

PR #6883, opened 2026-07-02
fix(mcp): harden custom date-range against malformed dates

Rouzax:
### What

Malformed `start_date`/`end_date` on the custom date-range path (`get_sensor_history` and the web-UI `json.htm` graph command) reached `HandleGraphCustomRange`, which does `srange.substr(11, 10)` and threw `std::out_of_range` on a short value like `"1970"`. The exception is caught one layer up in libwebem, so the daemon does not crash, but the request fails with only a generic log line and the user sees an empty result with no explanation.

### Changes

- **Core:** `HandleGraphCustomRange` now validates the range-string shape (length 21, `T` at index 10) and returns gracefully instead of throwing. This protects every caller of the custom-range path, including the web-UI `json.htm` graph command.
- **MCP:** `get_sensor_history` validates `start_date`/`end_date` as `YYYY-MM-DD` and returns a clear message when they are not. The check is a small pure helper (`isValidIsoDate`) in its own header.

### Validation

- Clean build.
- Runtime, against a real device set: a malformed range on `json.htm` and via a full MCP session both return gracefully (no exception, clear message), while a valid range returns normal data. Verified on Current (CM113), P1, Usage, and Voltage devices.

### Not affected

- `get_scene_history` uses `buildEventHistory`, not the custom-range path, so it is unaffected by this issue.

---

**Follow-up idea (not part of this PR):** I would like to start adding unit tests to the changes I contribute going forward, and this fix feels like a natural first candidate: the date validator here is a small pure function that is trivial to test in isolation. There is currently no C++ unit harness in the tree, so this would mean introducing something lightweight (I was thinking doctest as a single header in `extern/`, wired into a CTest target, kept non-blocking to start). I have written the validation as a standalone helper with that in mind. A small pure-function harness like this also pairs well with AI-assisted TDD workflows, so it should fit nicely with how a lot of us work now. Totally understand if this is not a direction you want to take right now; just flagging the intent so it is on your radar.

### DIFF
--- a/main/SQLHelper.h
+++ b/main/SQLHelper.h
@@ -333,6 +333,10 @@ public:
 	bool RestoreDatabase(const std::string &dbase);
 	bool RestoreDatabaseFromFile(const std::string &sourceFilePath);
 
+	// Validates that a string is a real YYYY-MM-DD calendar date. Public so callers
+	// outside CSQLHelper (e.g. the MCP server) can validate date arguments.
+	bool CheckDateSQL(const std::string &sDate);
+
 	// Returns DeviceRowID
 	uint64_t UpdateValue(int HardwareID, int OrgHardwareID, const char *ID, unsigned char unit, unsigned char devType, unsigned char subType, unsigned char signallevel, unsigned char batterylevel, int nValue,
 			     std::string &devname, const bool bUseOnOffAction, const char *User = nullptr);
@@ -584,7 +588,6 @@ private:
 	void AddCalendarUpdatePercentage();
 	void AddCalendarUpdateFan();
 	bool CheckDate(const std::string &sDate, int &d, int &m, int &y);
-	bool CheckDateSQL(const std::string &sDate);
 	bool CheckDateTimeSQL(const std::string &sDateTime);
 	bool CheckTime(const std::string &sTime);
 	void SendUpdateInt(const std::string& Idx);
--- a/main/WebServerHandleGraphCustom.cpp
+++ b/main/WebServerHandleGraphCustom.cpp
@@ -87,6 +87,12 @@ void HandleGraphCustomRange(const GraphContext& ctx, const request& req,
 	const std::string& srange      = ctx.srange;
 	const std::map<std::string, std::string>& options = ctx.options;
 
+	// The range string must be "YYYY-MM-DDTYYYY-MM-DD" (length 21, 'T' at index 10).
+	// Guard against a short or malformed value, which would otherwise throw
+	// std::out_of_range on the substr() calls below (e.g. a caller passing "1970").
+	if (srange.length() < 21 || srange[10] != 'T')
+		return;
+
 	std::string dbasetable = CalcDbasetableCustom(ctx);
 	unsigned char tempsign = sql.m_tempsign[0];
 
--- a/mcpserver/McpService.cpp
+++ b/mcpserver/McpService.cpp
@@ -3855,6 +3855,12 @@ namespace mcp		// Model Context Protocol
 		{
 			szDateStart = args["start_date"].asString();
 			szDateEnd   = args["end_date"].asString();
+			if (!m_sql.CheckDateSQL(szDateStart) || !m_sql.CheckDateSQL(szDateEnd))
+			{
+				mcp::setToolResult(jsonRPCRep,
+					"Invalid date. Use the format YYYY-MM-DD for start_date and end_date.", true);
+				return true;
+			}
 		}
 		else
 		{


Rouzax:
For context: this hardening PR came out of planning a larger follow-up on the same `get_sensor_history` path, which I've split off so this fix can land on its own.

The follow-up (a separate PR) would fix the rendering side. `get_sensor_history` currently only formats P1-style `v1`/`v2` and a generic value, so the other `MultiMeter_Calendar` families that #6878 restored come back as a bare date with no data: AirQuality, Lux, Usage, Weight, per-phase Current (CM113), and the General/RFXSensor min/max families.

The one design point I'd value your steer on before I build it: to render those correctly the formatter needs each series' unit, which the graph JSON currently omits (the web frontend knows units out-of-band from the device type). Rebuilding units MCP-side from `dType`/`displaytype`/`metertype` would duplicate core logic, the same cross-file drift that caused #6878. The cleaner option looks like a small, additive change to `HandleGraphCustomRange` to emit a per-series unit/quantity label alongside the values, which any consumer benefits from. Since you preferred to keep the #6878 fix narrow rather than touch the core dispatch, I'm holding that follow-up until you've had a chance to weigh in on whether that small core addition is welcome, or you'd rather it stay MCP-only.

No rush; this PR stands on its own regardless.

gizmocuz:
Thanks for the fix! The guard in `HandleGraphCustomRange` is good (please fix the indentation of the added lines — they have extra tabs relative to the surrounding code).

For the MCP part: we already have a date validator in the codebase — `CSQLHelper::CheckDateSQL()` (`main/SQLHelper.h`). It validates the `YYYY-MM-DD` shape and additionally verifies the value is a real calendar day (so `2024-02-30` is rejected too, which the format-only check would accept). `McpService.cpp` already uses `m_sql` throughout, so please drop the new `McpDateUtil.h` and use:

```cpp
if (!m_sql.CheckDateSQL(szDateStart) || !m_sql.CheckDateSQL(szDateEnd))
```

Regarding the doctest/unit-test harness follow-up idea: we do not want to take that direction at this time, so no need to keep the validation as a standalone header for that purpose.

Rouzax:
- Switched the MCP side to `CheckDateSQL` and dropped `McpDateUtil.h`. Agreed, reusing the existing validator is better, and it rejects impossible dates like `2024-02-30` that my format-only check would have let through.
- Indentation: sorted. I moved the range guard up into the function-setup block rather than dedenting it in place, since the body below it is over-indented from an earlier extraction and a lone dedented line there would have looked worse.
- Harness idea: understood, no problem, I'll leave that be.

One blocker before I push: `CheckDateSQL` is `private` in `CSQLHelper`. Okay to make it public so `McpService.cpp` can call it?

On the rendering follow-up (separate PR): right now `get_sensor_history` only formats P1-style usage/delivery and a generic value, so the other `MultiMeter_Calendar` families #6878 restored render as a bare date with no data (AirQuality, Lux, Usage, Weight, per-phase Current/CM113, and the General/RFXSensor min/max families). I'd fix the formatter to render them, for example:

```
2026-03-01 | L1: 2.1 A   L2: 0.8 A   L3: 3.4 A        (Current / CM113)
2026-03-01 | min 228 V   avg 230 V   max 233 V        (Voltage)
2026-03-01 | min 0 Lux   avg 450 Lux  max 1200 Lux    (Lux)
```

Two calls I'd value so I build it right the first time:
1. **Units:** the core already computes the exact unit for each series but does not put it in the JSON, so every consumer re-derives units (the logic is currently inline in `GetJSonDevices`, the month/year graph, and the custom-range graph). My preference would be a small, additive field in the graph JSON carrying the per-series unit label: any consumer benefits and it avoids adding a fourth copy of the unit logic. If you would rather not touch the graph code, I can keep it MCP-side by deriving units from the device data the tool already fetches, though that is more fragile for the multi-channel cases. Which would you prefer?
2. **Detail level:** is the per-phase and min/avg/max breakdown above about right for an MCP text response, or would you prefer something more compact?

gizmocuz:
Yes, `CheckDateSQL` is indeed private — go ahead and move its declaration to the public section in `SQLHelper.h` as part of this PR (it's stateless, so that's harmless). The rest of your plan sounds good, please push.

On the follow-up: a small additive per-series unit field in the custom-range graph JSON is fine — better than a fourth copy of the unit logic. Keep it scoped to `HandleGraphCustomRange`. Your proposed compact format for the MCP output looks good.

Rouzax:
Pushed. `CheckDateSQL` is now public and used for the date validation, the range guard is moved up to the function-setup block so it sits at the correct indentation, and `McpDateUtil.h` is gone. Cleaned up to two commits. Builds clean, and I verified the guard plus the MCP validation (including impossible dates like `2024-02-30`, now rejected) against a real device set. Ready for another look when you have a moment.

gizmocuz:
All feedback addressed and CI is green — merging. Thanks for the clean fix and the quick turnaround on the review points, @Rouzax!

PR #6886, opened 2026-07-03
fix(plugins): preserve custom plugin password on blank save

Rouzax:
### Background

While building a new myenergi plugin I've been drinking my own champagne with the extended plugin settings from #6674, and ran into a few rough edges. This is the one that actually loses data.

### The problem

A plugin param declared `password="true"` is stored in `Hardware.Settings` and (correctly) stripped from the `gethardware` response so the secret never reaches the browser. But that means the field renders empty on the hardware edit page, and saving resubmits the empty value, overwriting the stored secret. In practice, changing any unrelated setting silently wipes the plugin's API key, with no indication the field was ever set.

### What this PR does

Backend (`main/WebServerCmds.cpp`):

- `Cmd_UpdateHardware` merges on save: a password field submitted empty or missing keeps its stored value ("leave blank to keep"), mirroring the existing Netatmo `Extra` preserve-on-empty pattern. Fails gracefully, never wipes a readable secret, never throws on malformed JSON.
- `Cmd_GetHardware` emits a small `SettingsPwdSet` map so the UI can show a password is set, without sending the secret.
- Manifest password-field detection is pulled into shared helpers, re-keyed by the plugin `key` attribute (the manifest map is keyed by directory), matched case-insensitively (`true`/`TRUE`/`1`), built only when the result contains a plugin row.

Frontend (`www/app/hardware/Hardware.js`):

- A set password field renders a fixed-length dot placeholder (language-neutral, nothing submitted as a value).
- Required-validation skips an empty-but-stored password field on edit only; Add always requires it. Stored-password markers are cleared on form (re)init so they can't leak across the reused plugin fields.

Scope: custom `password="true"` params only. The reserved `Password` column round-trip is untouched (separate follow-up below).

### Validation

- Merge logic unit-tested (7 cases: blank / missing / replace / empty-input / corrupt-stored) against jsoncpp.
- Manifest detection tested against the bundled TiXml parser (case-insensitive, `<group>` children, key attribute).
- Full build with `ENABLE_PYTHON`.
- Verified end-to-end on a running instance with a real plugin: read-side strip + `SettingsPwdSet`, keep-on-blank and replace via `updatehardware`, and the edit/add UI paths.

### Pre-existing issues noticed while here (deliberately NOT bundled)

- Reserved `Password` column is sent to the browser and repopulated on edit (opposite of custom password params). Confidentiality follow-up with its own risk.
- With `ENABLE_PYTHON` off, lingering plugin rows have `Settings` sent unstripped (the assignment sits outside the `#ifdef`).
- A plugin renaming or removing a password field orphans the stored secret (not recognisable from the current manifest).
- `renderParam` interpolates manifest strings into HTML unescaped (admin page only; a plugin already runs server-side Python).

### DIFF
--- a/main/WebServerCmds.cpp
+++ b/main/WebServerCmds.cpp
@@ -1594,6 +1594,124 @@ namespace http
 			return true;
 		}
 
+#ifdef ENABLE_PYTHON
+		// Parse a plugin manifest XML once: extract the <plugin key="..."> attribute and the set of
+		// field names declared as password fields. The password attribute is matched case-insensitively
+		// ("true"/"TRUE"/"1") so a manifest typo does not silently expose a secret. Pure: operates on
+		// the manifest XML string, no I/O. keyOut is empty when the manifest has no key attribute.
+		static void ParsePluginManifest(const std::string &manifestXml, std::string &keyOut, std::set<std::string> &passwordFieldsOut)
+		{
+			keyOut.clear();
+			passwordFieldsOut.clear();
+			TiXmlDocument xmlDoc;
+			xmlDoc.Parse(manifestXml.c_str());
+			if (xmlDoc.Error())
+				return;
+			TiXmlNode *pPluginNode = xmlDoc.FirstChild("plugin");
+			if (!pPluginNode)
+				return;
+			TiXmlElement *pPluginEle = pPluginNode->ToElement();
+			if (!pPluginEle)
+				return;
+			const char *pKey = pPluginEle->Attribute("key");
+			if (pKey)
+				keyOut = pKey;
+			TiXmlNode *pParamsNode = pPluginNode->FirstChild("params");
+			if (!pParamsNode)
+				return;
+			auto isPasswordAttr = [](const char *pPassword) -> bool {
+				if (!pPassword)
+					return false;
+				std::string v(pPassword);
+				for (auto &c : v)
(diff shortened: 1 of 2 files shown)


gizmocuz:
Thanks @Rouzax, excellent work — this was a genuine data-loss bug and the fix is exactly right (mirrors the Netatmo preserve-on-empty pattern, and the case-insensitive `password` matching quietly closes a cleartext exposure for manifests using `password="True"`). Verified end-to-end against both the read and write paths. Merged!

As with #6885 we'll update the plugin development docs on our side to describe the leave-blank-to-keep behavior. And your list of deliberately-unbundled pre-existing issues is appreciated — we'll track those separately.

That said, the drinking of one's own champagne has to be kept to a minimum ;)

gizmocuz:
Follow-up: docs updated in a04fbe923 — both `password` attribute rows in `docs/Developing_a_Python_plugin.wiki` (the basic `<param>` table and the Extended Parameter Settings section) now describe the case-insensitive `true`/`1` values, that the stored secret never reaches the browser, the dot placeholder, and the leave-blank-to-keep behavior on edit — including the note that a stored secret can only be replaced, not cleared, from the UI.

The pre-existing issues you listed are now tracked in #6887 so they don't get lost. Thanks again!

PR #6888, opened 2026-07-04
feat(mcp): render multi-channel and aggregate sensor families in get_sensor_history

Rouzax:
### What

Follow-up to #6878 (which restored the underlying data). `get_sensor_history` only formatted P1-style usage/delivery and a generic value, so the other `MultiMeter_Calendar` families came back as a bare date with no data: AirQuality, Lux, Usage, Weight, per-phase Current (CM113/CM180i), and the General/RFXSensor min/max families. They now render with units.

### Changes

- **Core (`HandleGraphCustomRange`):** emit a per-series unit label in `ValueUnits` for the built-in counter families, which previously left it empty. Gap-fill only (never overrides an option-provided unit), additive (no emitted value changes), and scoped to this function. Current/CM113 reuses the existing display-type read, so no extra DB query.
- **MCP (`get_sensor_history`):** render each family from the keys the core emits, labelled with the unit (per-phase L1/L2/L3 for multi-channel current, min/avg/max for the aggregate families). The empty-result message is reworded to stop asserting logging is disabled.

Example:

```
Daily history for "Current L1/L2/L3" (Watt):
2026-02-20 | L1: 0-3726 Watt   L2: 345-5934 Watt   L3: 0-3542 Watt

Daily history for "Lux Sensor" (Lux):
2026-02-25 | min 0 avg 1353.14 max 5031 Lux
```

### Validation

Clean build. Verified against a real device set via a live MCP session: multi-channel Current (per-phase, with unit), Lux/aggregate min/avg/max, P1 usage/delivery, generic meters, and the advisory message on an out-of-range window.

### DIFF
--- a/main/WebServerHandleGraphCustom.cpp
+++ b/main/WebServerHandleGraphCustom.cpp
@@ -486,6 +486,52 @@ void HandleGraphCustomRange(const GraphContext& ctx, const request& req,
 				root["ValueUnits"] = (options.count("ValueUnits") ? options.at("ValueUnits") : std::string{});
 				root["Divider"] = divider;
 
+				// Built-in counter families carry no ValueUnits option, but their unit is
+				// fixed by device type. Populate it once here (the single place that owns
+				// the custom-range series) so consumers such as the MCP server can label
+				// values without re-deriving units. Gap-fill only: never overrides an
+				// option-provided unit, and does not touch the emitted values. This is a
+				// single per-response unit, correct because every family here uses one unit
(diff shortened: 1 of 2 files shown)


gizmocuz:
Thanks @Rouzax! Verified the key-shape contract against HandleGraphCustomRange (including the today-row paths) and the unit labels against the device-list conventions — all consistent, and the web UI is unaffected since the gap-fill only covers families whose log pages don't read the graph-response ValueUnits. Merging.

PR #6902, opened 2026-07-10
bugfix settings menu - Event for mobile

MadPatrick:


gizmocuz:
What is wrong with the current implementation?
And why would you create events with a mobile? This device is way to small for it, use a desktop/laptop or tablet

MadPatrick:
The button are now available on a mobile
I agree that for scripting it is to small, but when i need to disable a script for some reason and have a mobile available i want disable and save the script

Almost all the button are not available

Before:
image


After

image

gizmocuz:
could you change the screenshots to use the default theme please (before/after), and also do not post url's  in the screenshot

gizmocuz:
Thanks for this. I had a look and the underlying issue is real: `.events-editor-file__header` has a fixed `height: 30px` with 8px padding, so the 30px-tall buttons overflow it, and the content pane below is pinned at a hardcoded `top: 46px`. Anything that wraps to a second row is simply hidden. Good corroboration that this is a genuine default-theme bug: both the aurora and Nightglass themes already work around it with `height: auto`. Your approach (auto-height header plus syncing the content offset to the measured header height) is the right fix, and it helps those themes too.

I've pushed three small fixes directly to your branch (706f8b13c) so this doesn't need another round trip:

1. **Ace needs an explicit resize.** The bundled Ace is 1.2.2, which has no `ResizeObserver` and only recalculates on `window resize`. When `syncContentOffsetWithHeader()` changed `top`, Ace kept its stale size, so click/cursor coordinates drifted and the bottom lines became unreachable until an orientation change. It now calls `aceEditor.resize()` after the offset write. Blockly was already fine, it has its own `ResizeObserver` calling `Blockly.svgResize`.

2. **Zero-height guard.** Inactive tabs are `ng-show`-hidden, so `offsetHeight` measures 0 and `top` was being set to `0px`. With `ResizeObserver` that self-corrects, but on the window-resize fallback path the header ended up overlapping the editor. The offset write is now skipped when the header measures 0.

3. **Destroy guard.** `bindScriptHeaderResize()` runs inside the `.then()`, so closing the tab while `loadEvent()` was still in flight attached the observer and window listener after `$destroy` had already run, leaking both. Added an `isDestroyed` flag that short-circuits the bind.

A few things still on you before I can merge:

- **The PR body is empty.** Please add a short summary of what's changed.
- **Please split out or document the Automation Wizard change.** The `.aw-overlay` / `.aw-modal` mobile block is a separate fix from the events toolbar. It's a fine change on its own, but it should either be its own PR or at least be mentioned in the description.
- **Please squash** the commits into one before merge.
- Minor: the `--script` modifier is applied to a template that serves both the Ace and Blockly editors, so the name is a bit misleading. `--editor` would read better. Not a blocker.
- Still outstanding from my earlier comment: could you redo the before/after screenshots using the default theme, without the URL visible?

gizmocuz:
Apologies, I see you did redo the screenshots when you edited the comment above, I just missed the edit. Default theme in mobile view and no URLs, so both points are addressed. Disregard my last bullet.

The before/after is clear: the header is clipped to a sliver with Help/Opslaan/Verwijderen entirely missing, and after the fix it wraps to three rows with everything reachable. That's exactly the bug.

Remaining items are just the PR description, splitting out or mentioning the Automation Wizard change, and a squash before merge.

gizmocuz:
Merged, thank you! This will be included in the next beta.

For a future PR, a few small things that make reviewing (and merging) a lot quicker:

- A short description in the PR body: what was broken, what the change does, and how you tested it. It also gives us something to base the changelog entry on.
- If an unrelated fix sneaks in (like the Automation Wizard scroll fix here), either a separate PR or a quick mention in the description is enough, so it does not go unnoticed.
- Screenshots for UI changes, which you already added here, are always appreciated.

Thanks again for the contribution!

PR #6904, opened 2026-07-10
Feature/add selector switch and user variable condition in dzvents automation wizard

waltervl:
Selector switch devices showed a free numeric input for level values in the Automation Wizard, requiring users to know the raw numeric level (0, 10, 20…) rather than the named levels configured on the device. This replaces those inputs with named-level dropdowns wherever a selector value is entered.

## JS (`AutomationWizard.js`)

- **`getSelectorLevels(dev)`** — shared helper: decodes base64 `LevelNames`, maps each entry to `{ level: index * 10, name }`, respects `LevelOffHidden`, logs parse errors
- **`defaultLevelValue(levelOptions, fallbackDef)`** — helper to consistently pick the first level value or fall back to a numeric default
- **`DEVICE_CONDITIONS.selector`** — extended with `level_is` (`hasValue: true`) so users can match an exact level in step 2
- **`COND_EXPR`** — added `level_is: 'device.level == {v}'`; generates `device.level == 10` etc.
- **`selectDevice`** — populates `triggerConfig.conditionLevelOptions` when a trigger device is added
- **`updateConditionValue`** — defaults `conditionValue` to first level option via helper
- **`onActionDeviceChange`** — refactored to use `getSelectorLevels` (removes inline try/catch)
- **`onConditionDeviceChange` / `onConditionStateChange`** — populate `wizardCondition.levelOptions`, default value via helper
- **`triggerConditionHasLevelOptions()` / `wizardConditionHasLevelOptions()`** — controller predicates to keep template logic clean

## HTML (`AutomationWizard.html`)

- **Step 2** trigger condition value: `<select ng-options>` for selector devices, numeric `<input>` otherwise
- **Step 3** wizard condition device value: same conditional pattern

```html
<!-- Step 2: selector shows named-level dropdown, others keep numeric input -->
<select ng-if="$ctrl.triggerConditionHasLevelOptions()" class="form-control"
        ng-model="$ctrl.triggerConfig.conditionValue"
        ng-options="opt.level as opt.name for opt in $ctrl.triggerConfig.conditionLevelOptions">
</select>
<input ng-if="!$ctrl.triggerConditionHasLevelOptions()" type="number" class="form-control"
       ng-model="$ctrl.triggerConfig.conditionValue">
```

Generated dzVents code is unchanged — numeric levels are still passed to `switchSelector()` and condition expressions.

**Alert device support**
Alert device support has been added (commit `AutomationWizard: add alert device support for conditions and actions`).

**What's been implemented:**

- **Detection**: `getDeviceCategory()` now recognises devices with `SubType === 'Alert'`
- **Conditions (steps 2 & 3)**: Alert devices show "Any change" or "Level is" with a named dropdown: Grey (No alert), Green (Info), Yellow (Warning), Orange (Alert), Red (Emergency)
- **Actions (step 4)**: Alert devices show a "Set alert level" dropdown with the same named levels
- **Generated code** uses the domoticz constants directly for readability, e.g.:
  - Condition: `device.color == domoticz.ALERTLEVEL_GREEN`
  - Action: `myAlertDevice.updateAlertSensor(domoticz.ALERTLEVEL_ORANGE, '')`
## Summary

Add a new optional condition in **Automation Wizard – Step 3 (Condition)**:
- **Condition type:** “Only if a user variable has a specific value”
- User variable is selected from the existing user-variable list
- Value is entered as text
- Generated dzVents condition supports:
  - numeric comparison when value is numeric
  - string comparison otherwise

### `www/app/events/AutomationWizard.js`
- Load user variables when Step 3 condition type is set to `variable`
- Extend `wizardCondition` defaults with:
  - `varName`
  - `varValue`
- Extend code generation (`generateCode`) to include variable condition expression:
  - `tonumber(domoticz.variables('<name>').value) == <number>` for numeric values
  - `domoticz.variables('<name>').value == '<text>'` for string values

### `www/app/events/AutomationWizard.html`
- Add new condition type option:
  - `Only if a user variable has a specific value`
- Add UI block for variable condition:
  - dropdown for variable selection (`$ctrl.variables`)
  - input field for expected value (`$ctrl.wizardCondition.varValue`)

## Behavior

- Condition remains optional and does not affect existing condition types.
- If variable condition is selected but no variable/value is provided, no condition expression is emitted for that block.
- Existing automations remain unchanged.

## Manual test plan

1. Open Automation Wizard.
2. Go to Step 3 (Condition).
3. Select condition type: **Only if a user variable has a specific value**.
4. Confirm variable dropdown is populated.
5. Select variable and set value:
   - numeric example: `1`
   - string example: `Home`
6. Continue to Step 5 and inspect generated code:
   - numeric: `tonumber(domoticz.variables('...').value) == 1`
   - string: `domoticz.variables('...').value == 'Home'`
7. Verify existing condition types (device/time/days/daytime) still work as before.

## Notes

- This PR only adds the new Step 3 condition type and generated Lua condition logic.
- No changes to trigger behavior in Step 2.

### DIFF
--- a/www/app/events/AutomationWizard.html
+++ b/www/app/events/AutomationWizard.html
@@ -270,6 +270,7 @@
                             <option value="time">Only between certain hours</option>
                             <option value="days">Only on certain days</option>
                             <option value="daytime">Only during daytime or nighttime</option>
+                            <option value="variable">Only if a user variable has a specific value</option>
                         </select>
                     </div>
                     <!-- Device condition -->
@@ -353,6 +354,26 @@
                                     <select class="form-control aw-time-select" ng-model="$ctrl.wizardCondition.toMin"
                                             ng-options="item.value as item.label for item in $ctrl.minuteOptions"></select>
                                 </div>
+                            </div>
+                        </div>
+                    </div>
+                    
+                    <!-- Variable condition -->
+                    <div ng-if="$ctrl.wizardCondition.type === 'variable'">
+                        <div class="aw-row">
+                            <div class="aw-field">
+                                <label>{{:: 'Variable' | translate }}</label>
+                                <select class="form-control" ng-model="$ctrl.wizardCondition.varName">
+                                    <option value="">{{:: '— Select variable —' | translate }}</option>
+                                    <option ng-repeat="v in $ctrl.variables track by v.idx" value="{{v.Name}}">{{v.Name}}</option>
+                                </select>
+                                <div ng-if="!$ctrl.variablesLoaded" class="aw-device-loading">
+                                    {{:: 'Loading…' | translate }}
+                                </div>
+                            </div>
+                            <div class="aw-field">
+                                <label>{{:: 'Value' | translate }}</label>
+                                <input type="text" class="form-control" ng-model="$ctrl.wizardCondition.varValue" placeholder="42">
                             </div>
                         </div>
                     </div>
--- a/www/app/events/AutomationWizard.html
+++ b/www/app/events/AutomationWizard.html
@@ -85,12 +85,12 @@
                     </div>
                     <div class="aw-field" ng-if="$ctrl.conditionNeedsValue">
                         <label>{{:: 'Value' | translate }} <span ng-if="$ctrl.conditionUnit">({{$ctrl.conditionUnit}})</span></label>
-                        <select ng-if="$ctrl.triggerConfig.deviceCategory === 'selector' && $ctrl.triggerConfig.conditionLevelOptions && $ctrl.triggerConfig.conditionLevelOptions.length"
+                        <select ng-if="$ctrl.triggerConditionHasLevelOptions()"
                                 class="form-control"
                                 ng-model="$ctrl.triggerConfig.conditionValue"
                                 ng-options="opt.level as opt.name for opt in $ctrl.triggerConfig.conditionLevelOptions">
                         </select>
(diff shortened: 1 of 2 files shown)


gizmocuz (approved):
Reviewed: dzVents API usage, code generation substitution order, and helper refactor all check out. Merging — a small follow-up will address a stale condition-value edge case when switching condition devices.

gizmocuz:
Thanks a lot for this contribution @waltervl! The selector level, alert device and user variable conditions are a great addition to the Automation Wizard, and the refactor into shared helpers (`getSelectorLevels` / `defaultLevelValue`) made the code nicely maintainable.

During review everything checked out — dzVents API usage (`device.color`, `updateAlertSensor`, `ALERTLEVEL_*` constants), the code generation substitution order, and the separation of `level_is` vs `alert_level_is` condition ids are all correct.

We found two small edge cases which have been fixed in a follow-up commit (41de26097):

1. **Stale condition values when switching devices** — when changing the trigger or condition device between devices with different level sets (e.g. dimmer → selector, or two selectors with different level names), the previously entered value was kept even when it no longer matched the new device's level options. The dropdown then rendered blank while the stale value (e.g. `device.level == 50`) was still silently emitted into the generated dzVents code. Values are now re-defaulted when invalid for the current level options and cleared when the selected state takes no value.

2. **Whitespace-only user variable value** — `Number(' ')` evaluates to `0`, so a whitespace-only value passed the empty check and generated `tonumber(...) == 0`. The value is now trimmed first, so it correctly emits no condition.

Merged into development — thanks again!

Issue #5453, opened 2022-12-11
Camera Edition windows: buttons not translated in French


didi31:
![image](https://user-images.githubusercontent.com/46863073/206908964-77400485-7f79-4eef-a3d7-131d8f23c013.png)
Add => Ajouter
Cancel => Annuler
Update => Modifier

PR #6905, opened 2026-07-11
fix(webui): translate camera dialog buttons

adrighem:
Use the existing translation mechanism for the Add, Cancel, and Update buttons in the camera dialogs.
Refs #5453


adrighem:
I don't even use a camera myself, but thought this needed fixing. Tested with my local dev computer camera. Dialogs look nice now.

gizmocuz:
Thanks for the fix! One small follow-up while you're in this file: the edit dialog title at line 226 is still hardcoded (`title: 'Edit Camera'`), so it stays in English while the add dialog already uses `$.t('Add New Camera')`. Could you wrap it in `$.t()` as well? Note that "Edit Camera" doesn't exist as a key in the i18n files yet, so it would need to be added there too for the translation to take effect.

adrighem:
Added that one too. Feel free to merge if you're happy with it.

gizmocuz:
Thanks for the quick follow-up and for taking care of this long-standing issue — much appreciated! Merging.

PR #6906, opened 2026-07-12
Change CSS class for timed out devices in quick stat and energy dashboard widget

mce35:
Small change to update the CSS class of the items in the quick stat widget so that user can visually see the devices in timeout.

It looks like this:

image

Similar change in the energy dashboard widget.

### DIFF
--- a/www/app/dashboardDynamic/widgets/ddEnergyDashboard.widget.js
+++ b/www/app/dashboardDynamic/widgets/ddEnergyDashboard.widget.js
@@ -161,6 +161,7 @@ define([
                             usageDelivWatt:    parseWatt(gr.UsageDeliv),
                             counterToday:      gr.CounterToday || '',
                             counterDelivToday: gr.CounterDelivToday || '',
+                            timeout:           gr.HaveTimeout === true,
                             price:             gr.hasOwnProperty('price') ? (parseFloat(gr.price) || 0) : null
                         };
                     } else {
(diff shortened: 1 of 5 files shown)


gizmocuz (changes requested):
Thanks, nice addition — timeout visibility fits well with the other dynamic dashboard widgets (dial, setpoint, timeout-monitor) that already react to `HaveTimeout`.

Two small changes before merging:

1. **Tint the card background instead of replacing it.** `.dd-stat-card` normally has `background: var(--dz-panel-bg)`; the new rule replaces it with `color-mix(in srgb, var(--dz-status-timeout) 25%, transparent)`, which makes the card semi-transparent to whatever is behind it. Mixing with the panel background keeps the card solid and reads better on both light and dark themes:

   ```css
   background: color-mix(in srgb, var(--dz-status-timeout) 25%, var(--dz-panel-bg));
   ```

   Same for the quick-stat rule.

2. **Class naming:** `.dd-qs--timeout` is inconsistent with its sibling `.dd-qs-item--on` — please rename it to `.dd-qs-item--timeout` (in both `dashboard.css` and `quick-stat.html`).

mce35:
Thank you for the review, I address both comments.

gizmocuz (approved):
Changes look good, thanks!

gizmocuz:
Thank you for this nice addition and for addressing the review feedback!

Issue #4203, opened 2020-06-03
Feature request: logarithmic scale for Lux sensor


SargonofAssyria:
It would be nice to have the possibility to select a logarithmic scale for the Lux sensor.
The value for a sensor which measures Lux outside may vary from max. > 65000 Lux to 0 Lux
On a the current scale you cannot see differences in the levels below 1000 Lux.
A logarithmic scale shows a lot more.
So please include a setting per Lux sensor for the 2 types of scales.

gizmocuz:
@SargonofAssyria , what hardware do you use to generate that Lux value ?
If it's a script, can't you scale it down to 0-100% or 0-1000 ?
How does the graph look ?

SargonofAssyria:
I am using https://www.zigbee2mqtt.io/devices/GZCGQ01LM.html
This device is more accurate then a 'Buienrader" device.
I use it to close rollershutters and turn on some lights.
But the graph is not looking so nice in the range below 100 lux where normal trigger points are.
To close my rollershutters I use 10 Lux.
In the winter 1 glimps of the sun will set the graph to this picture (> 40000 lux) and the whole day the line will be on or close to the zero line.
Not a complete graph ( I reinstalled the device yesterday) but you can see that 136 lux is already sitting on the 0 line.
![2020-06-04_11-42-37](https://user-images.githubusercontent.com/24824113/83742280-84e32400-a659-11ea-91db-d68ea6158d25.png)

xztraz:
another graph from an esp with lux sensor. 
![image](https://user-images.githubusercontent.com/25690871/96594158-99fbdf00-12ea-11eb-8b88-75b3e506b838.png)

1000 lux is one pixel ish..  would be nice with logarithmic scale on graphs for lux.

PR #6908, opened 2026-07-12
Example logarithmic scaling interface for Lux charts

adrighem:
This draft implements an example of what a logarithmic scaling interface for Lux charts could look like.

It adds Linear and Logarithmic controls to the day, month, year, and comparison views. The logarithmic option handles zero values and uses readable powers-of-ten tick marks.

Refs #4203

### DIFF
--- a/test/www-test/app/log/LuxLogScale.test.js
+++ b/test/www-test/app/log/LuxLogScale.test.js
@@ -1,0 +1,76 @@
+const assert = require('assert');
+const fs = require('fs');
+const path = require('path');
+const vm = require('vm');
+
+describe('LuxLogScale', function () {
+    let LuxLogScale;
+
+    before(function () {
+        const source = fs.readFileSync(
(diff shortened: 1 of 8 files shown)


adrighem:
https://github.com/user-attachments/assets/1439fc3f-8e7f-4de1-be54-a76b6ffb3a12

Please comment.
Or maybe we don't want this at all and we should just close the issue.

adrighem:
No rush by the way. This has been a long time open issue and the review load is probably very high. I appreciate your time and don't want to make your life harder.

gizmocuz:
Thank you for this PR and time.
I do not see why this is needed. Lux (Light) is important, I use it to control my house lights. to dark->lights on, to light->off. this is based on a value
in your logarithmic display, i do not see the actual values anymore...

adrighem:
Allright then. No harm done. I actually agree with you, but you don't know until you implement something. But at least we know now and can close the issue. :+1:

Issue #6829, opened 2026-05-29
MQTT Auto-Discover misdetects RGBW Z-Wave devices as RGBWWZ


Rouzax:
### Prerequisites

- [x] I have searched existing issues and the forum for this problem
- [x] I am running the latest beta version

### Domoticz Version

2025.2 (build 17395)

### Platform

Docker

### OS Version

Debian (aarch64)

### Describe the Bug

Z-Wave RGBW controllers with a single white channel (e.g. Fibaro FGRGBW-442) are incorrectly detected as RGBWWZ (5-channel: RGB + Cold White + Warm White) instead of RGBWZ (4-channel: RGB + White).

This happens because the detection logic in `MQTTAutoDiscover.cpp` (line 4670-4684) checks `color_temp_command_template` for the strings `warmWhite` and `coldWhite`. Z-Wave JS always generates that template with both fields, even when the physical device only has a single white channel.

The `supported_color_modes` field correctly advertises `["rgb", "brightness", "white"]` (note: `"white"`, not `"color_temp"`), which should be sufficient to determine this is a single-white RGBW device.

As a result, the UI shows two white sliders and a color temperature control for a device that only has one white channel.

### Steps to Reproduce

1. Add a Fibaro FGRGBW-442 (or similar single-white Z-Wave RGBW controller) to Z-Wave JS UI
2. Configure Domoticz MQTT Auto-Discover pointing to the same MQTT broker
3. The device is auto-discovered as RGBWWZ instead of RGBWZ
4. In Domoticz, check Setup > Devices; the SubType shows "RGBWWZ"

### Expected Behavior

The device should be detected as RGBWZ (4-channel), matching its physical capabilities: RGB + single White. The UI should show an RGB color picker and a single white slider.

### Hardware/Plugin Type

MQTT Auto-Discover, Z-Wave JS UI

### Log Output

_(no errors logged, the device is silently misclassified)_

### MQTT Auto Discovery Data

Config topic: `homeassistant/light/Overkapping-Verlichting/rgb_dimmer_1/config`

Config payload:
```json
{
  "state_topic": "zwave/Overkapping/Verlichting/38/1/currentValue",
  "command_topic": "zwave/Overkapping/Verlichting/38/1/targetValue/set",
  "on_command_type": "brightness",
  "state_value_template": "{{ \"OFF\" if value_json.value == 0 else \"ON\" }}",
  "brightness_value_template": "{{ value_json.value }}",
  "brightness_scale": 99,
[...]

PR #6909, opened 2026-07-13
fix(mqtt): White mode never commanded the white channel of auto-discovered RGBW lights

Rouzax:
**Background**

#6829 fixed the *detection* of single-white Z-Wave RGBW controllers (they now map to RGBWZ instead of RGBWWZ), but using such a device from the web UI still misbehaved: the color picker's White mode appeared to do nothing. Watching the MQTT traffic of a Fibaro FGRGBW-442 behind Z-Wave JS UI shows why. Selecting White published only a dimmer change and `{"state":"ON"}` to the color command topic, with no color value at all, so the controller kept whatever RGB color it was showing and the white channel was never driven. The only way to reach the white channel was the Custom (color + white mix) mode.

**Root cause**

`MQTTAutoDiscover::SendSwitchCommand` builds color payloads for `ColorModeRGB`/`ColorModeCustom` and for `ColorModeTemp`, but has no branch for `ColorModeWhite` (the mode the picker's White button sends). The command falls through and is published without a color value. This affects every MQTT Auto Discovery light with a white channel, not just Z-Wave ones.

**What this PR changes and why**

Commit 1 adds `ColorModeWhite` to the RGB/Custom branch. Per the color mode contract in `ColorSwitch.h`, mode White carries no valid color fields, so the payload is built as: color channels 0, white channels 255, with the actual brightness carried by the dim level. This matches how `ZWaveBase` has always handled White mode for natively integrated Z-Wave RGBW devices, so MQTT-discovered lights now behave the same as native ones.

Commit 2 fixes a related asymmetry that surfaces once white commands work. Incoming state for single-white lights stores the white level in `ww` (`warmWhite` is renamed to `w` and parsed into `color.ww`), but the command path read only `color.cw`. Re-sending a device-reported color (scenes, scripts) therefore silently dropped white to zero. The single white value is now `max(cw, ww)`, which accepts both the UI convention (sets both fields) and the state-parse convention (fills only `ww`). The stale comment claiming single white is stored as cold white contradicted the state parsing and was replaced.

For single-white devices the value is still sent as both `warmWhite` and `coldWhite`: gateways strip the component the node does not support (verified against Z-Wave JS, which does this explicitly since zwave-js GH#2527), and the discovery config does not tell us which single component the device has.

**What stays the same**

RGB and Custom mode payloads are unchanged (regression-tested), as are dual-white devices (RGBWWZ/RGBWW use their own `cw`/`ww` fields as before) and `color_temp`/`hs`/`xy` lights. A script sending mode 1 to an xy/hs-only light was a silent no-op before and remains out of scope; the assumption is documented in a comment.

**How this was tested**

Live-broker test rig: a local build discovered six synthetic config flavors (the exact Fibaro/Z-Wave JS payload from #6829, rgb+color_temp, rgb only, color_temp only, rgbw, rgbww) and the published command payloads were captured and asserted for White, RGB, Custom and the re-send case. All three fixed cases failed on the unmodified build and pass with this PR; the two regression cases pass on both. Additionally verified against a physical FGRGBW-442 through Z-Wave JS UI: the White payload drives the real white channel and the reported state confirms it.

### DIFF
--- a/hardware/MQTTAutoDiscover.cpp
+++ b/hardware/MQTTAutoDiscover.cpp
@@ -5768,6 +5768,11 @@ bool MQTTAutoDiscover::SendSwitchCommand(const std::string& DeviceID, const std:
 					iColdWhite = 255;
 					iWarmWhite = 255;
 				}
+				else if (pSensor->subType == sTypeColor_RGB_W_Z)
+				{
+					// single white channel: state updates store it in ww, the web UI sets both cw and ww
+					iColdWhite = std::max(color.cw, color.ww);
+				}
 				if (
 					(pSensor->subType == sTypeColor_RGB_W_Z)
 					|| (pSensor->subType == sTypeColor_RGB_CW_WW_Z)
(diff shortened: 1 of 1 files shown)


gizmocuz:
Please do not add unrelated 'ping' to github issues. You can send me a 'ping' on Slack

Rouzax:
Apologies

gizmocuz:
Thanks @Rouzax, nice find and a clean, well-scoped fix! The change matches the frontend White button contract and the _tColor White constructor convention, and the std::max(cw, ww) reconciliation for the state re-send asymmetry is correct.

One minor nit on the PR description only (no code change needed): the note that this matches how ZWaveBase has always handled White mode is only true for the legacy single-channel instance path. The dual-white COMMAND_CLASS_SWITCH_COLOR path in ZWaveBase sets only wWhite=255 and leaves cWhite=0, so it is not identical. Since the White button is only exposed for single-white (RGBWZ) devices this does not affect the fix itself.

Merging now, this will be included in the next beta version.

PR #6911, opened 2026-07-13
fix(webui): add tooltips to the color picker mode buttons

Rouzax:
Related to #6829, #6909 and #6910. The mode buttons at the top of the RGBW color picker are unlabeled icons, and which button drives which channel group is guesswork: the grey ball (White) and the half-colored ball (Custom) are especially opaque for lights that have both a color and a white channel.

This adds translated `title` tooltips to the five mode buttons: Color, White, Color temperature, and Color and white mix for the two custom variants. The titles are set centrally in `ShowRGBWPicker` when the popup is built, so the dashboard and switches popups, the scenes view and the Angular `rgbw-picker` component are all covered without touching their templates. `White`, `Color temperature` and `Color and white mix` did not exist in the translation source yet and are registered in `www/i18n/to_be_translated.json`.

Tested with a Playwright harness asserting the title attributes on RGBWZ and RGBWWZ pickers in the served default theme (six assertions, all failing before and passing after).

### DIFF
--- a/www/i18n/to_be_translated.json
+++ b/www/i18n/to_be_translated.json
@@ -9,5 +9,8 @@
     "Origins": "Origins",
     "Also allow origins from local networks (no username/password)": "Also allow origins from local networks (no username/password)",
     "Warning: allowed origins can read and control the API from a browser on your local networks!": "Warning: allowed origins can read and control the API from a browser on your local networks!",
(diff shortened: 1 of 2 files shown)


gizmocuz:
Merged, thank you for this nice usability improvement! The tooltips will be included in the next beta.

PR #6916, opened 2026-07-15
fix(dashboardDynamic) CustomChart widget: bug with week range, add air quality sensor

fleinze:
In the CustomChart widget is a bug, the weekly time range does not work because the widget trys to call the api with "range=week". However the api does not recognise this.

Also I added support for CO2 Air Quality Sensors.

### DIFF
--- a/www/app/dashboardDynamic/widgets/ddCustomChart.widget.js
+++ b/www/app/dashboardDynamic/widgets/ddCustomChart.widget.js
@@ -53,7 +53,7 @@ define([
         if (sub === 'SetPoint' || type === 'Thermostat') {
             return { sensor: 'temp', field: function(d){ return d.se !== undefined ? d.se : d.te; }, unit: tempUnit() };
         }
-	if (type === 'Air Quality') {
+	    if (type === 'Air Quality') {
             return { sensor: 'counter', field: function(d){ return d.co2 }, unit: 'ppm' };
         }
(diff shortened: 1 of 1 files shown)


gizmocuz:
Thanks for the report and the patch! You were right that the Custom Chart's week range was broken — the backend simply didn't implement `range=week` for most sensor types (only rain and regular Meter-backed counters worked).

Instead of remapping week to day in the widget (which would regress the counter/rain types that already had real weekly data, and would only show as much history as the Short Log retention keeps — 1 day by default), we implemented this in `development` as follows:

**Backend (2bb244a44^ / d323a84b1):** `range=week` now works for *all* sensor types — temp/hum, Percentage, fan, UV, wind and every MultiMeter-calendar counter device (Air Quality, Lux, Weight, Usage, Voltage, Current, Soil/Leaf Wetness, Visibility, Distance, Solar Radiation, Pressure, Sound Level, RFXSensor). Each returns 7 days of daily calendar aggregates plus a computed today record, using the same JSON field names as the month/year responses, independent of the Short Log setting.

**Custom Chart widget (2bb244a44):** uses the native week API for all sensors, and includes your CO2 / Air Quality support — extended with a `co2_avg` fallback so the month and year ranges work too. While testing this we also found and fixed Lux devices charting nothing on any range, and Voltage/Current/Usage showing empty month/year charts.

Credited you in the commit message — thanks again! Closing this as implemented via the commits above.

PR #6619, opened 2026-03-01
Angular migration: Dashboard widgets, floorplan fixes, and frontend improvements

gizmocuz:
## Summary

- **Dashboard refactor**: Split monolithic `DashboardController.js` (4090 lines) into modular widget directives (`dzLightWidget`, `dzSceneWidget`, `dzUtilityWidget`) with separate desktop/mobile controllers and a shared `dashboardService`
- **Floorplan security fix**: Replace `eval()` with safe alternatives, add keyboard/arrow navigation and UX improvements
- **Device controller cleanup**: Refactor `LightsController`, `ScenesController`, `UtilityController`, `TemperatureController`, and `WeatherController` to use widget directives and modernized API calls
- **New device detection service**: Add `deviceDetection.js` for responsive layout handling across desktop, tablet, and mobile
- **Performance improvements**: Optimize page load, drag/drop UX, and widget update reliability
- **Test infrastructure**: Add Karma-based unit test framework under `test/www/` with specs for dashboard service, device detection, and light widget
- **Cleanup**: Remove outdated `www-test` suite, legacy `WindLog`, and unused code

## Test plan
- [ ] Verify dashboard loads correctly on desktop, tablet, and mobile views
- [ ] Test light/scene/utility widgets render and respond to state changes
- [ ] Test floorplan keyboard navigation and device interaction
- [ ] Verify device log links work from dashboard icons
- [ ] Check switch type ordering in lights view

PR #6917, opened 2026-07-16
fix(webui): restore touch dragging on the color picker wheel and sliders

Rouzax:
**Background**

Since #6619 the color picker cannot be operated by dragging on touch devices: the wheel and the sliders respond only to discrete taps, which makes fine color or brightness control on a phone nearly impossible. This affects every user of the shared wheel picker widget, i.e. the pickers of all Color Switch subtypes (RGB, RGBW, RGBWW, RGBWZ, RGBWWZ, WW), the scene color picker, and the dynamic dashboard's theme color swatch, since they all instantiate the same `jquery.wheelcolorpicker` widget.

**Root cause**

#6619 rebound the picker's touch events from jQuery `.on()` to native `addEventListener` so they could be registered with `{passive: ...}`. But the touch coordinate wrapper in `updateActiveControl` still assumes a jQuery-wrapped event and reads `e.originalEvent.touches`. Native TouchEvents have no `originalEvent`, so the first `touchmove` of a drag throws `TypeError: Cannot read properties of undefined (reading 'touches')` and the drag never updates anything. Taps still work because the browser synthesizes mouse events after a tap, which is exactly the "tap sets, drag does nothing" symptom.

**Fix**

Read the touch list from the event itself when `originalEvent` is absent, fall back to `changedTouches` (a touchend carries its position there), and return early when no position is available instead of computing with undefined coordinates. The mouse path (`pageX` present) is untouched. The library is tracked as `.gz` only; the patch preserves the file's CRLF line endings so the extracted diff stays minimal (one 13-line hunk).

**How this was tested**

Playwright dispatching real `TouchEvent` sequences at the served default theme: before the fix the TypeError fires and neither wheel nor slider moves; after it, the wheel tracks a quarter-circle drag continuously and the master slider follows a top-to-bottom drag, with no page errors. Mouse interaction is unchanged. Also manually verified on a phone against a LAN test instance: wheel and slider dragging is smooth again (verified with RGBWZ and RGBWWZ devices; the fix is in the shared widget code that all picker variants use).


Rouzax:
Note for reviewers: the only file this PR touches is `www/js/jquery.wheelcolorpicker.js.gz`, which is tracked gzipped, so GitHub renders no diff at all (+0/-0). Here is the extracted change so it can be read in place:

```diff
@@ -1911,10 +1911,19 @@
 		var options = this.options;
 		var color = this.color;
 		
-		// pageX and pageY wrapper for touches
-		if(e.pageX == undefined && e.originalEvent.touches.length > 0) {
-			e.pageX = e.originalEvent.touches[0].pageX;
-			e.pageY = e.originalEvent.touches[0].pageY;
+		// pageX and pageY wrapper for touches.
+		// The event is either jQuery-wrapped (mouse path, touch position in
+		// e.originalEvent) or a native TouchEvent (the touch listeners are bound
+		// with addEventListener and have no originalEvent).
+		if(e.pageX == undefined) {
+			var touchSource = (e.originalEvent != undefined) ? e.originalEvent : e;
+			var touchList = (touchSource.touches != undefined && touchSource.touches.length > 0) ? touchSource.touches
+				: (touchSource.changedTouches != undefined && touchSource.changedTouches.length > 0) ? touchSource.changedTouches
+				: null;
+			if(touchList == null)
+				return;
+			e.pageX = touchList[0].pageX;
+			e.pageY = touchList[0].pageY;
 		}
 		//$('#log').html(e.pageX + '/' + e.pageY);
```

That is the entire patch: one hunk in `updateActiveControl`. The mouse path is untouched, since it still returns early once `pageX` is present. Only events that arrive without `pageX` take the new branch, and those are exactly the ones that throw today on `e.originalEvent.touches` after #6619 rebound the touch listeners to `addEventListener`.

To verify the blob yourself:

```
git show origin/development:www/js/jquery.wheelcolorpicker.js.gz | gunzip > old.js
git show <this-branch>:www/js/jquery.wheelcolorpicker.js.gz | gunzip > new.js
diff -u old.js new.js
```

gizmocuz:
Thanks for tracking this down and for the clean fix! Verified the root cause (native touch listeners since #6619 vs. e.originalEvent) and the patch looks good. Merging.

Issue #6912, opened 2026-07-13
dzVents: Error in ProcessNotificationItem data: in Json::Value::resolveReference(key, end): requires objectValue, but found intValue


DGA45:
### Prerequisites

- [x] I have searched existing issues and the forum for this problem
- [x] I am running the latest beta version

### Domoticz Version

2026.2 (build 18129)

### Platform

Other ARM/embedded

### OS Version

Linux orangepi5pro 6.1.115-vendor-rk35xx

### Describe the Bug

From time to time, I got the following pair of errors:
```
2026-07-13 13:00:19.830  Error: dzVents: Error in ProcessNotificationItem data: in Json::Value::resolveReference(key, end): requires objectValue, but found intValue
2026-07-13 13:00:19.871  Error: EventSystem: in /home/pi/domoticz/dzVents/runtime/dzVents.lua: /home/pi/domoticz/dzVents/runtime/EventHelpers.lua:1230: attempt to index a nil value (field 'name')
```
Both errors are related to CustomEvent, the second error is certainly a consequence of the first one (the dzVents table is incomplete).

### Steps to Reproduce

Having people moving in front of cameras (not systematic)

### Expected Behavior

- No error in the log
- All Frigate MQTT events received as custom event in dzVents system

### Log Output

```text
2026-07-13 13:00:19.246  dzVents: Frigate event 1783940408.066563-4kw9n7 cam Cam-Entree frame_time 1783940419.0571
2026-07-13 13:00:19.483  dzVents: Frigate event 1783940408.066563-4kw9n7 cam Cam-Entree frame_time 1783940419.2543
2026-07-13 13:00:19.830  Error: dzVents: Error in ProcessNotificationItem data: in Json::Value::resolveReference(key, end): requires objectValue, but found intValue
2026-07-13 13:00:19.871  Error: EventSystem: in /home/pi/domoticz/dzVents/runtime/dzVents.lua: /home/pi/domoticz/dzVents/runtime/EventHelpers.lua:1230: attempt to index a nil value (field 'name')
2026-07-13 13:00:20.045  dzVents: Frigate event 1783940408.066563-4kw9n7 cam Cam-Entree frame_time 1783940419.7185
2026-07-13 13:00:20.687  dzVents: Frigate event 1783940408.066563-4kw9n7 cam Cam-Entree frame_time 1783940420.133
[...]

PR #6921, opened 2026-07-18
Fix bug #6912 dzVents: Error in ProcessNotificationItem data: in Json::Value::resolveReference(key, end): requires objectValue, but found intValue

DGA45:
I tried the proposed modification from issue #6912 and the error message did not appear since the last 24h.

### DIFF
--- a/main/dzVents.cpp
+++ b/main/dzVents.cpp
@@ -145,7 +145,7 @@ void CdzVents::ProcessNotificationItem(CLuaTable& luaTable, int& index, const CE
 			}
 			catch (const std::exception& e)
 			{
-				_log.Log(LOG_ERROR, "dzVents: Error in ProcessNotificationItem data: %s", e.what());
+				_log.Log(LOG_ERROR, "dzVents: Error in ProcessNotificationItem data (%s): %s", item.sValue.c_str(), e.what());
 			}
 			luaTable.CloseSubTableEntry();
 		}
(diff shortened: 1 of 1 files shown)


gizmocuz:
Not sure what his issue is about? is it a confirmation about a pr that solves a issue? in that case, this issue can be closed

gizmocuz:
Thanks a lot for tracking this one down, @DGA45, nice catch. Your root-cause diagnosis was spot on: `_tEventQueue::nValue` carries a raw device value for `REASON_DEVICE` items and a notification type for `REASON_NOTIFICATION` items, so the `hardware` and `customevent` loops (which only checked the precomputed flags, never `item.reason`) could misroute a device update into `ProcessNotificationItem` whenever its `nValue` happened to equal `Notification::DZ_CUSTOM`. Adding the `reason` check to both loops, mirroring the `domoticz` loop, is exactly the right fix. The extra `item.sValue` in the error log is a welcome bonus too.

Merged into `development` as 8073623, with your authorship preserved.

One small thing for next time: the patch had quite a few whitespace inconsistencies. The new `if (item.reason == ...)` blocks and several `break;` / `}` lines mixed spaces into what is otherwise a tab-indented file. I cleaned those up before merging (it took the diff from +12/-20 down to +6/-14, so most of the "changes" were actually just indentation noise).

Two easy ways to avoid it:

- **Configure your editor for tabs.** Domoticz C++ sources use hard tabs. In VS Code: `"editor.insertSpaces": false` and `"editor.detectIndentation": true`, plus `"editor.renderWhitespace": "all"` so you can literally see it. Visual Studio: Tools -> Options -> Text Editor -> C/C++ -> Tabs -> "Keep tabs". Most editors also honour an `.editorconfig`.
- **Review your own diff before pushing.** `git diff --check` flags whitespace problems, and the GitHub PR "Files changed" tab has a whitespace toggle. A quick look there would have shown these lines.

Neither is a big deal, and it definitely doesn't take away from the fix. Thanks again for the report in #6912 and for following through with the patch.

This fix is included in beta version 18140, available within the next 15 minutes.

PR #6927, opened 2026-07-24
DynamicDashboard CustomChart widget: add support for custom sensors

fleinze:
This PR adds support for custom sensors in the CustomChart Widget.


mealgi:
Hello,

This PR assumes that all custom sensors are percentage sensors which is not the case.

BR

fleinze:
For the API, they are always percentage sensors.

gizmocuz:

PR #6937, opened 2026-08-02
fix(webserver): use X-Forwarded-For for trusted proxies

adrighem:
Set libwebem's trusted proxy header family to `X-Forwarded-For` before creating each Domoticz web server.

This restores client IP resolution for trusted reverse proxies after libwebem's default changed to ignore forwarded headers. The shared startup path covers both HTTP and HTTPS.


gizmocuz:
Thank you for this PR and your time.

We just made a new release, so the change in the History.txt document should be placed at the top

gizmocuz:
Merged, thank you for the quick diagnosis and the clean fix! No worries about the changelog — we've taken care of the History.txt placement ourselves (moved the entry to the new unreleased section at the top, commit 3b79d5f3a). Much appreciated!

adrighem:
Thanks for the History.txt change. Wanted to make it look good with the alphabetized list, but yes, new release. :-D
Glad to help

Issue #5416, opened 2022-11-16
MQTT Auto discovery Radio Thermostat Company of America (RTC) CT101 Thermostat Fan mode


cookemmm:
Moving from openzwave to zwavejs2mqtt I noticed my thermostat no longer had a fan mode device. Looking in the Z-wave JS UI it appeared to be missing in the device json so I attempted to add it. I believe I finally have the json correct but domoticz does not create a device to control the fan mode. 

I might still be missing something in my json below but I have searched and can't find any information on how this should work. 

{
  "type": "climate",
  "object_id": "climate",
  "discovery_payload": {
    "min_temp": 5,
    "max_temp": 40,
    "temp_step": 0.5,
    "modes": [
      "off",
      "heat",
      "cool",
      "auto"
    ],
    "fan_modes": [
      "Auto low",
      "Low"
    ],
    "mode_state_template": "{{ {0: \"off\", 1: \"heat\", 2: \"cool\", 3: \"auto\"}[value_json.value] | default('off') }}",
    "fan_mode_state_template": "{{ {0: \"Auto low\", 1: \"Low\"} [value_json.value] | default('Auto Low') }}",
    "current_temperature_topic": "zwave/17/49/1/Air_temperature",
    "current_temperature_template": "{{ value_json.value }}",
    "fan_mode_state_topic": "zwave/17/69/0/state",
    "fan_mode_command_topic": "zwave/17/68/0/mode/set",
    "temperature_state_topic": "zwave/17/67/1/setpoint/1",
    "temperature_state_template": "{{ value_json.value }}",
    "temperature_command_topic": "zwave/17/67/1/setpoint/1/set",
    "mode_state_topic": "zwave/17/64/1/mode",
    "mode_command_topic": "zwave/17/64/1/mode/set",
    "action_topic": "zwave/17/66/0/state",
    "action_template": "{{ {0: \"idle\", 1: \"heating\", 2: \"cooling\", 3: \"fan\", 4: \"idle\", 5: \"idle\", 6: \"fan\", 7: \"heating\", 8: \"heating\", 9: \"cooling\", 10: \"heating\", 11: \"heating\"}[value_json.value] | default('idle') }}",
    "temperature_unit": "F",
    "precision": 0.1,
    "device": {
      "identifiers": [
        "zwavejs2mqtt_0xcadd9946_node17"
      ],
      "manufacturer": "Radio Thermostat Company of America (RTC)",
      "model": "Z-Wave Thermostat (CT101)",
[...]

PR #6948, opened 2026-08-06
MQTTAutoDiscover: subscribe to and dispatch

heggink:
fan_mode_state_topic / swing_mode_state_topic for climate devices

Problem

For MQTT-autodiscovered climate devices, fan_mode_state_topic and swing_mode_state_topic are parsed from the discovery config and stored (hardware/MQTTAutoDiscover.cpp, lines ~1598-1601 and ~1622-1625: pSensor->fan_state_topic / pSensor->swing_state_topic), but they are never actually used afterwards:

Never subscribed. The subscribe block in on_auto_discovery_message (lines ~1836-1850) explicitly subscribes to mode_state_topic, temperature_state_topic, current_temperature_topic, preset_mode_state_topic, etc. — but not fan_state_topic or swing_state_topic. Never matched. The incoming-message dispatcher in handle_auto_discovery_sensor_message (lines ~1902-1915) checks mode_state_topic, temperature_state_topic, preset_mode_state_topic, etc. against the incoming topic to find the owning sensor — again, fan_state_topic and swing_state_topic are absent.

Net effect: even when a third-party bridge (in my case a Node-RED flow bridging LG ThinQ air conditioners) publishes correctly and consistently to the advertised fan_mode_state_topic, Domoticz's MQTT client never subscribes to that topic on the broker, so it never sees the update — not a delivery problem, Domoticz simply never asks for it. The climate device's fan-speed/swing state field is therefore permanently stale/unset in Domoticz, regardless of how correct the publisher is.

This matches user reports in #5416 ("MQTT Auto discovery ... Thermostat Fan mode" — Domoticz not creating/updating a fan mode device from fan_mode_state_topic/fan_mode_command_topic) and forum threads describing the same symptom for other MQTT-autodiscovery bridges (ESPHome, zigbee2mqtt, etc.).

Fix

Add fan_state_topic and swing_state_topic to both the subscribe list and the topic-match dispatcher, alongside the other climate sub-topics that are already handled the same way (preset_mode_state_topic etc.).

cpp
// on_auto_discovery_message – subscribe block
SubscribeTopic(pSensor->preset_mode_state_topic, pSensor->qos); SubscribeTopic(pSensor->fan_state_topic, pSensor->qos); SubscribeTopic(pSensor->swing_state_topic, pSensor->qos); cpp
// handle_auto_discovery_sensor_message – dispatch match || (pSensor->preset_mode_state_topic == topic)
|| (pSensor->action_topic == topic)
|| (pSensor->fan_state_topic == topic)
|| (pSensor->swing_state_topic == topic)

See attached patch (domoticz-mqtt-fan-swing-subscribe.patch) for the exact diff against development.

Testing
Verified against the current development branch source that both fields are parsed but absent from the two lists above (grep for fan_state_topic / swing_state_topic outside the parser only turns up the struct declaration). Not yet build-tested locally — I don't have a Domoticz build environment set up. Would appreciate a maintainer or anyone who can build this confirming that once the topic is subscribed and matched, handle_auto_discovery_climate's value-processing correctly updates the fan/swing state for the device (I could not fully inspect that function's body to confirm this end-to-end). If it needs additional handling downstream I'm happy to extend the patch. Related
#5416

### DIFF
--- a/hardware/MQTTAutoDiscover.cpp
+++ b/hardware/MQTTAutoDiscover.cpp
@@ -1851,6 +1851,8 @@ void MQTTAutoDiscover::on_auto_discovery_message(const struct mosquitto_message*
 			SubscribeTopic(pSensor->percentage_state_topic, pSensor->qos);
 			SubscribeTopic(pSensor->action_topic, pSensor->qos);
 			SubscribeTopic(pSensor->preset_mode_state_topic, pSensor->qos);
+			SubscribeTopic(pSensor->fan_state_topic, pSensor->qos);
+			SubscribeTopic(pSensor->swing_state_topic, pSensor->qos);
 
 		}
 	}
@@ -1915,6 +1917,8 @@ void MQTTAutoDiscover::handle_auto_discovery_sensor_message(const struct mosquit
 			|| (pSensor->percentage_state_topic == topic)
 			|| (pSensor->preset_mode_state_topic == topic)
 			|| (pSensor->action_topic == topic)
+			|| (pSensor->fan_state_topic == topic)
+			|| (pSensor->swing_state_topic == topic)
 			)
 		{
 			matching_keys.emplace_back(itt.first, MatchType::State);


gizmocuz:
Thanks @heggink, merged!

Reviewed the code paths and your analysis is spot on: `fan_mode_state_topic` and `swing_mode_state_topic` were parsed and stored, and `handle_auto_discovery_climate` already had complete handling for both (including the `*_state_template` variants for bridges that publish JSON), but the topics were never subscribed to and never matched in the dispatcher, so that code was unreachable. Your patch closes exactly that gap and mirrors the existing `preset_mode_state_topic` handling.

The empty-topic case is safe since `SubscribeTopic()` is a no-op on empty strings, and the standalone MQTT `fan` component is a separate path, so no regression risk there.

Appreciate the fix!

Issue #6907, opened 2026-07-12
Per-user overlay for ThemeSettings (web UI personalization per login)


Rouzax:
### Prerequisites

- [x] I have searched existing issues for this feature request

### Description

ThemeSettings (docs/Theming.wiki, "Theme settings storage") gives themes a supported, event-system-free place to persist configuration in the Preferences table. Two personalization gaps remain that a theme cannot solve on its own:

1. The value is one JSON object per installation, so any stored choice, for example the wiki's own dark/light mode toggle example, applies to every user and every device. One person's preference restyles the wall tablet and everyone else's browser.
2. The only write path is the storesettings POST from the Settings page, which is Admin-only. A user with Viewer or User rights can never persist a preference; localStorage covers a single browser but does not follow the login and is lost when browser data is cleared.

This proposal keeps `Preferences.ThemeSettings` unchanged as the instance-wide defaults and adds a small per-user overlay on top of it.

### Use Case

Any theme, including the default, can offer per-user choices: the dark/light toggle from the Theming.wiki example becomes per-login (one family member dark on their phone, the wall tablet stays light), color scheme pickers, per-user dashboard density, and so on.

Longer term, a per-user GUI theme (WebTheme) could ride the same overlay: the styles/default rewrite is already resolved per request, so it could resolve per session.

Related demand: #6665 asked for per-user favorites. Device assignment answered that case; theme preferences have no equivalent workaround, because ThemeSettings has exactly one value per installation.

### Proposed Solution

A per-user overlay with its own lightweight commands, next to the existing mechanism:

- `json.htm?type=command&param=getuserthemesettings` returns the calling user's overlay.
- `json.htm?type=command&param=setuserthemesettings&value=<json>` (and a delete counterpart) writes only the calling user's overlay; the session decides whose, there is no idx parameter.
- Rights: any authenticated user (Viewer and up), own entry only.
- Same shape and rules as ThemeSettings today: per-theme namespaced JSON, validated with ParseJSon before storing (as storesettings already does), size-capped (a few KB per user).
- Storage: one row per user (UserID, Value, LastUpdate).
- Documented precedence: a key present in the user overlay wins over the instance ThemeSettings for that user; themes merge the two objects client-side.

What this does not ask for: no core UI change is required, themes consume the API. No change to user variables or to the existing ThemeSettings/storesettings behavior. Installations without overlay rows behave exactly as today.

[...]

PR #6950, opened 2026-08-06
Per-user overlay for ThemeSettings (#6907)

Rouzax:
Implements #6907 as accepted: theme settings move from the single instance-wide `Preferences.ThemeSettings` blob to a `ThemeSettings` table holding instance defaults (Scope 0) and per-user overlays (Scope 1), with a dedicated write path that does not depend on the admin Settings page.

## What this does

- **New table `ThemeSettings`** keyed on `(Scope, UserID, ThemeName)`, one row per theme per layer. Migration (`DB_VERSION 181`) explodes the legacy blob into Scope 0 rows; idempotent, and renumbered to 181 since d1d339a33 took 180.
- **Three new commands** (names chosen to avoid the libwebem prefix-match quirk, domoticz/libwebem#4; deliberately not on the auth-bypass whitelist):
  - `themesettings_get` (any authenticated user): returns both layers for one theme, each with a presence flag and a concurrency token, plus a `PerUser` capability flag.
  - `themesettings_set` (POST, any authenticated user incl. Viewer): writes the caller's own overlay only; there is no parameter to address another user.
  - `themesettings_setdefault` (POST, admin only, checked by equality): writes the instance defaults and mirrors them into the legacy `Preferences.ThemeSettings` row so a rollback to a pre-181 build still finds current defaults. Nothing reads that row anymore.
- **`getsettings` keeps returning `ThemeSettings`**, now as the merge of instance defaults with the calling user's overlay, user values winning per theme name (whole sub-object replace; themes treat their sub-object as atomic). Existing themes that read `data.ThemeSettings[THEME]` get per-user reads with zero code changes.
- **The hidden `ThemeSettings` field in `storesettings` is removed outright** (server handler, `setup.html` hidden input, `SetupController.js` write-back). The read population of `$scope.ThemeSettings` stays, serving the merged view.
- **`deleteuser` cleanup**: a user's overlay rows are removed in `CSQLHelper::DeleteUser` (extends d1d339a33's centralized cleanup), so a recycled `Users.ID` cannot inherit a deleted user's settings. The new table needs no orphan purge of its own: it cannot have stale rows on existing installs.
- **`docs/Theming.wiki`** rewritten to describe only the new style; `getversion` gains `ThemeSettingsAPI: 1` so a theme can support old and new cores.

## Hardening in the write path

- Rights checked by equality (never `>=`; `URIGHTS_NONE`/`URIGHTS_CLIENTID` are numerically above admin).
- Sessions that do not resolve to a Users row (OAuth clients, access tokens) are refused explicitly (`no_identity`) instead of being routed to a shared row.
- Values: length-checked before parsing, strict JSON parse wrapped against jsoncpp's deep-nesting throw, object root required, 16 KB cap applied to the reserialized form which is what gets stored.
- Optimistic concurrency: every row carries a `lastupdate` token, required back on write; a stale token returns `conflict` instead of silently overwriting. The token is generated server-side and bound in the same atomic upsert (`INSERT ... ON CONFLICT DO UPDATE ... WHERE`), verified via `sqlite3_changes` under the query mutex.
- A cap of 32 stored themes per user per layer (`too_many_themes`) keeps table growth bounded, as promised in the FR.

## Notes for review

- **Themes writing through the old path get a silent no-op.** Any theme that persisted settings by mutating `$scope.ThemeSettings` and relying on the Settings-page POST now writes nothing (requirement 2, removed outright). Reading is unaffected. The migration path for writers is `themesettings_set`; happy to coordinate the Nightglass update.
- **`ThemeSettingsAPI: 1` on `getversion`** is the one piece of API surface beyond the issue's list: an unknown command returns plain HTTP 400, indistinguishable from other errors, so themes have no other way to feature-detect. Returned for any authenticated session so Viewer-facing themes can branch too.
- **`PerUser` reuses an existing idiom**: `!session.istrustednetwork || !session.id.empty()` is the same expression as the pre-existing `canlogout` in `Cmd_GetAuth`.
- **Rights checks use an equality allowlist** (`VIEWER`/`SWITCHER`/`ADMIN`) rather than the `!= URIGHTS_NONE` blacklist: this also excludes `URIGHTS_CLIENTID` (255), making the `no_identity` check a second gate rather than the only one.
- **`safe_exec_changes` next to `safe_exec_no_return`**: the existing helper discards `sqlite3_changes()`, which is the basis of the compare-and-swap here, and changing it would touch unrelated callers; the new helper holds the query mutex across exec and changes() so the count cannot be interleaved.
- **Contract narrowing in the migration**: legacy top-level values that are not JSON objects are not migrated (a hand-edited `"MyTheme": "dark"` string would disappear from `getsettings` after upgrade). The UI-written blob was always an object per theme, so this only affects manual DB edits; the wiki documents the object-root rule.
- **Untested paths, stated for honesty**: `no_identity` (needs an OAuth/token session the test harness cannot cheaply produce), `db_error` (needs fault injection), and the 64 KB pre-parse guard (the 16 KB stored cap triggers first in practical payloads). Everything else in the error table is exercised by the committed tests.

## Known limitation (documented, per the issue discussion)

On `WebLocalNetworks` / `-nowwwpwd` installs every anonymous LAN client resolves to the same identity, so per-user rows cannot isolate devices there. `themesettings_get` returns `PerUser: false` for such sessions so a theme can hide per-user options; explicit logins still get `PerUser: true`.

## Testing

- New Gherkin suite (`test/gherkin/themesettings.feature`, 10 scenarios, committed): instance default write, per-user overlay winning in `getsettings` for the viewer while admin keeps defaults, viewer forbidden from instance scope, stale-token conflict, reset restoring defaults, non-object rejection, POST enforcement, capability flag, the 32-theme cap, and user delete/recreate with an assertion that the recreated user reuses the rowid (proving the cleanup matters).
- 26-check curl smoke matrix against a live instance (rights, caps, conflict, reset, mirror content, deleteuser cleanup, `storesettings` no longer touching the table).
- Migration verified against a simulated install that already ran the 180 step: explode correct, legacy blob untouched, `DB_Version` 181.
- Browser-level regression on the Settings page (Playwright): admin save works, `storesettings` POST carries no `ThemeSettings` field, no console errors.
- Full build clean on Linux (all targets).

Note: `Theming.wiki` states "build 18198 or newer" for the capability; that number may need a touch-up depending on the build this lands in.

Nightglass reads `ThemeSettings` via `getsettings` and keeps working unchanged; only its write path (the hidden `storesettings` field) needs moving to `themesettings_set`, which I can coordinate.


Rouzax:
One design point from the review hardening worth your eyes: writes are capped at 32 stored themes per user per layer (error `too_many_themes`), to keep the table growth promise from the FR honest now that any authenticated user can create rows. Updating an already-stored theme is never blocked, and quota can be freed through the existing API: the top-level keys of `ThemeSettings` in `getsettings` enumerate a user's rows, and `themesettings_set` with `reset=true` deletes one; the wiki documents the recipe and the error message points at it.

Question: is per-theme reset enough of a recourse for you, or would you rather have a dedicated listing/clear-all command in this PR? I left it out deliberately (smallest API surface that solves the problem), but it is a quick addition if you prefer it.

gizmocuz:
Thanks for the work on this, the functionality and the hardening look good. Before merging I would like the code organised differently. Some review points:

### 1. `safe_exec_changes`: keep it, it is the one piece that does belong in SQLHelper

I checked the alternatives and they are worse:

* `RETURNING LastUpdate` on the upsert needs SQLite 3.35+. `USE_BUILTIN_SQLITE` defaults to `NO` (`CMakeLists.txt:26`), so most builds link the system sqlite3, which on older distros is 3.27/3.34. Not portable.
* The existing in-tree pattern, `safe_exec_no_return(...)` followed by `sqlite3_changes(m_dbase)` (`main/SQLHelper.cpp:8648-8681`), reads `changes()` *after* `m_sqlQueryMutex` has been released, so another thread's write can change the count. Fine for a "how many rows did I purge" log line, not fine as a concurrency token.

So the helper stays. It is generic SQL plumbing sitting next to `safe_exec_no_return` and is usable by any table. That is also the dividing line for the rest of this review: `safe_exec_changes` belongs in SQLHelper, everything theme-specific does not.

### 2. The theme functions and the enum do not belong in SQLHelper

`GetThemeSettingsRow`, `SetThemeSettingsRow`, `DeleteThemeSettingsRow`, `CountThemeSettingsRows`, `GetMergedThemeSettings`, `MirrorThemeSettingsDefaults` and `eThemeSettingsWrite` are feature logic, not database primitives. Please move them into a new `CThemeSettings` class in `main/ThemeSettings.h` / `main/ThemeSettings.cpp`.

`main/KWHStats.h` is the in-tree model to copy: standalone class, `static constexpr` constants in the header, `Json::Value` forward declared, all DB access through `m_sql` inside the `.cpp`.

### 3. `root["ThemeSettingsAPI"] = 1;`

That bare `1` should be `CThemeSettings::API_VERSION`. The same applies to the three `constexpr` size/count caps and `IsValidThemeName`, which are currently loose in `WebServerCmds.cpp`, and to the magic scope values `0` and `1`, which should become an `eScope` enum class.

In general: put as much of this as possible in the new class, and leave `WebServerCmds.cpp` with only authentication, parameter plumbing and reply mapping.

### 4. Boilerplate

**`main/ThemeSettings.h`**

```cpp
#pragma once

#include <string>

namespace Json
{
	class Value;
} // namespace Json

class CThemeSettings
{
      public:
	// Reported to themes as "ThemeSettingsAPI"; bump when the command surface changes
	static constexpr int API_VERSION = 1;

	static constexpr size_t MAX_REQUEST_SIZE = 64 * 1024;
	static constexpr size_t MAX_STORED_SIZE = 16 * 1024;
	static constexpr size_t MAX_NAME_LENGTH = 64;
	static constexpr int MAX_THEMES_PER_SCOPE = 32;

	enum class eScope
	{
		Instance = 0,
		User = 1,
	};

	enum class eResult
	{
		Ok,
		InvalidTheme,
		MissingValue,
		InvalidJson,
		TooLarge,
		TooManyThemes,
		Conflict,
		DBError,
	};

	// Schema and migration, called from CSQLHelper::OpenDatabase()
	static void CreateTable();
	static void MigrateFromPreferences();

	static bool IsValidThemeName(const std::string &themeName);

	static bool Get(eScope scope, unsigned long userID, const std::string &themeName, std::string &value, std::string &lastUpdate);
	static eResult Set(eScope scope, unsigned long userID, const std::string &themeName, const std::string &szValue,
			   const std::string &expectedLastUpdate, std::string &newLastUpdate);
	static eResult Reset(eScope scope, unsigned long userID, const std::string &themeName);
	static bool GetMerged(bool haveUser, unsigned long userID, Json::Value &merged);
	static void MirrorDefaults();
	static void DeleteForUser(unsigned long userID);

	// Single place where an eResult becomes the JSON API's error/message pair
	static const char *ErrorCode(eResult res);
	static const char *ErrorMessage(eResult res);

      private:
	static int CountRows(eScope scope, unsigned long userID, const std::string &excludeThemeName);
};
```

**`main/ThemeSettings.cpp`**, the part that matters, everything the handler used to do inline:

```cpp
CThemeSettings::eResult CThemeSettings::Set(const eScope scope, const unsigned long userID, const std::string &themeName, const std::string &szValue,
					    const std::string &expectedLastUpdate, std::string &newLastUpdate)
{
	if (!IsValidThemeName(themeName))
		return eResult::InvalidTheme;
	if (szValue.empty())
		return eResult::MissingValue;
	if (szValue.size() > MAX_REQUEST_SIZE)
		return eResult::TooLarge;

	Json::Value jValue;
	bool bParsed = false;
	try
	{
		bParsed = ParseJSonStrict(szValue, jValue);
	}
	catch (const std::exception &)
	{
		bParsed = false;
	}
	if (!bParsed || !jValue.isObject())
		return eResult::InvalidJson;

	const std::string szStored = JSonToRawString(jValue);
	if (szStored.size() > MAX_STORED_SIZE)
		return eResult::TooLarge;

	const int themeCount = CountRows(scope, userID, themeName);
	if (themeCount < 0)
		return eResult::DBError;
	if (themeCount >= MAX_THEMES_PER_SCOPE)
		return eResult::TooManyThemes;

	// Token generated here so the exact value written can be handed back without a
	// second, non-atomic read-back
	const std::string szNewLastUpdate = TimeToString(nullptr, TF_DateTimeMs);
	const int changes = m_sql.safe_exec_changes(
		"INSERT INTO ThemeSettings (Scope, UserID, ThemeName, Value, LastUpdate) "
		"VALUES (%d, %llu, '%q', '%q', '%q') "
		"ON CONFLICT(Scope, UserID, ThemeName) DO UPDATE "
		"SET Value = excluded.Value, LastUpdate = excluded.LastUpdate "
		"WHERE ThemeSettings.LastUpdate == '%q'",
		static_cast<int>(scope), (unsigned long long)userID, themeName.c_str(), szStored.c_str(), szNewLastUpdate.c_str(),
		expectedLastUpdate.c_str());
	if (changes < 0)
		return eResult::DBError;
	if (changes == 0)
		return eResult::Conflict;

	newLastUpdate = szNewLastUpdate;
	return eResult::Ok;
}

const char *CThemeSettings::ErrorCode(const eResult res)
{
	switch (res)
	{
		case eResult::InvalidTheme:
			return "invalid_theme";
		case eResult::MissingValue:
			return "missing_value";
		case eResult::InvalidJson:
			return "invalid_json";
		case eResult::TooLarge:
			return "too_large";
		case eResult::TooManyThemes:
			return "too_many_themes";
		case eResult::Conflict:
			return "conflict";
		default:
			return "db_error";
	}
}
```

**What is left in `WebServerCmds.cpp`.** `ThemeSettingsWrite()` disappears from `WebServer.h` entirely:

```cpp
void CWebServer::Cmd_ThemeSettingsSet(WebEmSession &session, const request &req, Json::Value &root)
{
	root["status"] = "ERR";
	root["title"] = "ThemeSettingsSet";

	if (req.method != "POST")
	{
		root["error"] = "post_required";
		root["message"] = "Only POST is allowed";
		return;
	}
	if ((session.rights != URIGHTS_VIEWER) && (session.rights != URIGHTS_SWITCHER) && (session.rights != URIGHTS_ADMIN))
	{
		session.reply_status = reply::forbidden;
		return;
	}
	const int iUser = session.username.empty() ? -1 : FindUser(session.username.c_str());
	if (iUser == -1)
	{
		// OAuth clients, access tokens and synthetic sessions have no Users row to
		// attach an overlay to; refuse explicitly instead of guessing an owner.
		root["error"] = "no_identity";
		root["message"] = "Session does not resolve to a user account";
		session.reply_status = reply::forbidden;
		return;
	}

	const std::string themeName = request::findValue(&req, "theme");
	const unsigned long userID = m_users[iUser].ID;
	std::string newLastUpdate;

	const auto res = (request::findValue(&req, "reset") == "true")
				 ? CThemeSettings::Reset(CThemeSettings::eScope::User, userID, themeName)
				 : CThemeSettings::Set(CThemeSettings::eScope::User, userID, themeName, request::findValue(&req, "value"),
						       request::findValue(&req, "lastupdate"), newLastUpdate);
	if (res != CThemeSettings::eResult::Ok)
	{
		root["error"] = CThemeSettings::ErrorCode(res);
		root["message"] = CThemeSettings::ErrorMessage(res);
		return;
	}
	if (!newLastUpdate.empty())
		root["lastupdate"] = newLastUpdate;
	root["status"] = "OK";
}
```

`CSQLHelper::DeleteUser` then calls `CThemeSettings::DeleteForUser(userID)` instead of inlining the `Scope==1` purge, and `OpenDatabase()` calls `CThemeSettings::CreateTable()` plus, under `dbversion < 181`, `CThemeSettings::MigrateFromPreferences()`.

### 5. Build files

The new `main/ThemeSettings.cpp` / `.h` must be registered in three places, easy to miss on a Linux-only workflow:

* `CMakeLists.txt`: the explicit source list around line 167-190 (it is not a glob), next to `main/KWHStats.cpp`
* `msbuild/domoticz.vcxproj`: `ClCompile` for the `.cpp`, `ClInclude` for the `.h`
* `msbuild/domoticz.vcxproj.filters`: both entries under the `main` filter group

gizmocuz:
On your open question about per-theme reset versus a clear-all: keep the per-theme reset as the primary mechanism, and add a narrow clear-all on top of it. Not either/or.

The reason to add it is not convenience, it is reachability. As the PR stands, `themesettings_get` requires a specific `theme` parameter and returns only that theme's rows, and `getsettings` returns the *merged* view where you cannot tell which keys came from the user overlay and which from the instance default. So a client has no way to enumerate which themes it holds overlays for. That has three consequences:

* Uninstall or rename a theme and its `Scope=1` rows become permanently unreachable, because nothing can name them to reset them.
* Those orphans still count against the 32 row cap: `CountThemeSettingsRows` excludes only the theme currently being written, so dead rows eat the user's quota with no way to free it.
* "Forget my personal theme preferences" is a reasonable user request, and today it requires guessing theme names.

So I would add it, scoped tightly:

* **User scope only.** `DELETE FROM ThemeSettings WHERE Scope==1 AND UserID==<caller>`, with the same rights check as `themesettings_set` and the same `no_identity` refusal. Roughly ten lines on top of the existing reset path, either as `reset=all` or as a separate `themesettings_reset` command.
* **Not for instance defaults.** An admin action that wipes every theme's `Scope=0` config in one call is a footgun with no undo and no real use case. Admins configure themes one at a time, so per-theme reset is the right granularity there.
* **Not needed for user deletion.** `DeleteUser` already purges the `Scope==1` rows.

The alternative, making `themesettings_get` return the list of the caller's overlay theme names when `theme` is omitted, also solves reachability and is similarly small. But it costs N round trips to do what one call should do, and it does not handle the orphaned theme case any better. My preference is the clear-all.

Rouzax:
Reorganised as reviewed, in two commits so the mechanical move and the new behaviour stay separable:

* `e90b7db4a` moves the code, no behaviour change.
* `bbf10f58f` adds `reset=all`.

### What moved

`GetThemeSettingsRow`, `SetThemeSettingsRow`, `DeleteThemeSettingsRow`, `CountThemeSettingsRows`, `GetMergedThemeSettings`, `MirrorThemeSettingsDefaults` and `eThemeSettingsWrite` are now `CThemeSettings` in `main/ThemeSettings.h` / `.cpp`, modelled on `main/KWHStats.h`, together with the table DDL, the v181 migration, the size and count caps, `IsValidThemeName`, `API_VERSION` behind `getversion`, an `eScope` enum class replacing the magic `0`/`1`, and the single place where a result becomes the API's `error`/`message` pair.

`CSQLHelper::OpenDatabase()` calls `CThemeSettings::CreateTable()` and, under `dbversion < 181`, `MigrateFromPreferences()`. `DeleteUser` calls `DeleteForUser`. `ThemeSettingsWrite()` is gone from `WebServer.h`, and the three handlers keep only authentication, parameter plumbing and reply mapping. `safe_exec_changes` stays in `CSQLHelper` for the reasons you gave. Registered in `CMakeLists.txt` and in both msbuild files.

### Five places where I did not follow the boilerplate literally

Each has a reason, and each is worth a second look:

1. **`Get()` takes `Json::Value &value`, not `std::string &value`.** With the string form the handler still has to `ParseJSon` the stored row and decide what a bad parse means, which is the kind of thing this review moved out. The header already forward declares `Json::Value` for `GetMerged`. A row that no longer parses now reports as absent rather than as a broken layer the caller must special case.
2. **`DeleteForUser` returns `eResult`, not `void`.** It is now also the `reset=all` path, so it has to be able to report `db_error`. `CSQLHelper::DeleteUser` discards the result, like the `safe_query` calls around it.
3. **`CreateTable()` uses `m_sql.safe_exec_no_return()`.** `CSQLHelper::query()` is private, so the DDL cannot go through the same call the other tables use.
4. **`domoticz.vcxproj.filters` has no `main` group.** Both entries went under `Helpers`, which is where `KWHStats` sits.
5. **One `eResult` per failure mode collapses some message text.** Three `db_error` messages become one and two `too_large` messages become one. The machine-readable `error` codes are unchanged, so the wiki and the committed tests still hold; only the human `message` differs in those cases.

### reset=all

Built as you scoped it: user scope only, no `theme` parameter, same POST, rights and `no_identity` gates as any other overlay write, and no instance-scope counterpart (`themesettings_setdefault` with `reset=all` returns `invalid_theme` and touches nothing). Per-theme `reset=true` stays the primary mechanism, for both layers.

One correction on the reasoning, because it changes what the feature is for rather than whether to build it. Orphaned overlays are not unreachable today: `GetMergedThemeSettings` copies every `Scope=1` row of the caller into the merged object whether or not a `Scope=0` row exists under that name, so an overlay belonging to a renamed or uninstalled theme still shows up as a top-level key in `getsettings` and can be named to `reset=true`. Quota could already be freed, and a client could already enumerate what it holds. What `reset=all` genuinely buys is one call instead of N, and "forget my personal theme settings" as a single action that a theme can offer as one button. That is enough on its own, so it is in.

### Verification

* Full Linux build clean, no new warnings.
* Gherkin suite 11/11 against a live instance, including a new scenario that stores overlays for two themes as the viewer plus one as the admin, clears with `reset=all`, and asserts the viewer's rows are gone while the admin's overlay and both instance defaults survive.
* Direct checks on the built binary: `reset=all` deletes only the caller's `Scope=1` rows, is idempotent, rejects GET with `post_required`, and `themesettings_setdefault&reset=all` leaves the instance rows untouched.
* Migration re-verified after the move, since it changed file: a v180 database whose legacy blob holds two objects and one scalar explodes to two `Scope=0` rows, skips the scalar, leaves the legacy `Preferences` row intact, and lands on `DB_Version` 181.

### For the theme side

Nothing changes for readers: `getsettings` still returns `ThemeSettings` as the merged view, so a theme that reads `data.ThemeSettings[THEME]` needs no changes at all. Only writers move, from the removed hidden `storesettings` field to `themesettings_set`.

Worth flagging for Nightglass specifically, since the core deliberately ships no UI for this: a user can only remove their own settings if the theme draws the control. Two commands cover it, `themesettings_set` with `theme=X&reset=true` for one theme and `reset=all` for everything that user stored. Without such a control, a user's first save detaches them from the instance defaults permanently, so it is worth pairing the write-path move with a reset action in the same change. Happy to work that through together.

gizmocuz:
Reviewed both commits. The refactor is faithful and the split into a mechanical move plus the new behaviour made it easy to check.

`main/ThemeSettings.h` follows `KWHStats.h` as intended, `ThemeSettingsAPI` is `CThemeSettings::API_VERSION`, `eScope` replaces the magic `0`/`1`, `ThemeSettingsWrite()` is gone from `WebServer.h`, and the three handlers are down to authentication, parameter plumbing and reply mapping. Registered in `CMakeLists.txt` and both msbuild files. CI green on linux and windows.

### On your five deviations

All four technical ones are correct, and two of them are you catching my mistakes:

1. `Get()` returning `Json::Value` is better. It takes the parse-and-decide out of the handler, which was the point of the review in the first place.
2. `DeleteForUser` returning `eResult` is required once it also backs `reset=all`. Agreed.
3. `CreateTable()` using `safe_exec_no_return`: confirmed, `query()` is private (`main/SQLHelper.h:602`, under the `private:` at line 527), so it cannot use the same call the other tables use.
4. `Helpers` rather than `main` in the filters file: you are right and I was wrong. There is no `<Filter Include="main">` group in `msbuild/domoticz.vcxproj.filters` at all, and `KWHStats` sits under `Helpers`. Correct call.
5. Collapsing the duplicated message text is fine. The machine-readable `error` codes are unchanged, which is what the wiki and the tests pin.

### On the orphaned overlays

Your correction is right and my reasoning was wrong. `GetMerged` writes every `Scope=1` row of the caller into the merged object whether or not a matching `Scope=0` row exists, so an overlay belonging to an uninstalled or renamed theme still surfaces as a top-level key in `getsettings` and can be named to `reset=true`. Quota was always freeable. `reset=all` is one call instead of N plus a single "forget my personal theme settings" action, which is reason enough to keep it, just not the reason I gave.

### One remaining nit, not a blocker

The `reset` parameter is matched by exact string with a silent fallthrough to the write path. In `Cmd_ThemeSettingsSetDefault` only `reset=true` is recognised, so `themesettings_setdefault&reset=all` is inert only when no `theme` is supplied. Send `reset=all&theme=X&value={...}` and it stores the value instead of reporting that clear-all is not supported for the instance layer. Same shape in `Cmd_ThemeSettingsSet` for any `reset` value other than `all` or `true`. Rejecting unrecognised `reset` values explicitly would close it.

### Holding off on the merge

Not because of anything in the code. I checked the Nightglass theme against this branch and it breaks in a way that is worth coordinating before this lands.

`src/js/search.js` persists through exactly the path this PR removes:

```js
scope.ThemeSettings = scope.ThemeSettings || {};
scope.ThemeSettings[THEME_NAME] = serializeSettings();
scope.StoreSettings();                       // _saveToDomoticz()
```

With the hidden `#ThemeSettings` input gone from `setup.html`, the copy into it gone from `SetupController.js`, and the server side of `storesettings` no longer reading the field, that save becomes a no-op. Three things make it worse than an ordinary break:

* Nothing surfaces. The POST succeeds, `_dirty` is cleared, the unsaved-changes toast is dismissed and the button reports success.
* The theme mirrors everything into `localStorage`, so settings keep working in that browser. The loss shows up only on a second browser, another device, or after clearing storage.
* The gate is already open. `_useNewApi` is `build >= 17806` and development is at 18201, so every beta user is on that path today.

Reading is unaffected and no data is lost, the v181 migration moves the existing blob into instance rows.

So the plan is: hold the merge, give galadril time to move Nightglass onto `themesettings_set`, and land the two together. I have opened an issue on the Nightglass repo with the details. Your point about pairing the write-path move with a reset control in the same change is exactly right and is in there.

Nothing for you to change in the meantime unless you want to take the `reset` fallthrough nit.

gizmocuz:
Sorry I introduced a merge conflict, could you increase the database version to fix it?

PR #6955, opened 2026-08-09
fix(Netatmo): prevent error code 13 when weather station is not owned

Tableflipper9001:
Problem:
Netatmo spits out error code 13:
Netatmo error codes

When inspecting the code m_bPollWeatherData seems to not be initialised anymore.
(so C++ default = true). Since I do not own a weather station netatmo keeps complaining about scope rights (error code 13).

Example logging before this patch (2026.3):
Netatmo 2026 3

Example logging after this patch (build 18233 (4140c8ea7-modified) with local fix):
Netatmo after fix 2

### DIFF
--- a/hardware/Netatmo.cpp
+++ b/hardware/Netatmo.cpp
@@ -106,7 +106,7 @@ CNetatmo::CNetatmo(const int ID, const std::string& username, const std::string&
 	Debug(DEBUG_HARDWARE, "Netatmo Actif Scopes %s", m_scopes.c_str());
 
 	// Weatherdevice is available in HomesData + HomeStatus
-	//m_bPollWeatherData = (m_scopes.find("station_R") != std::string::npos);      //read_station
+	m_bPollWeatherData = (m_scopes.find("station_RW") != std::string::npos);      //read_station
 	m_bPollHomecoachData = (m_scopes.find("homecoach_RW") != std::string::npos);  //read_homecoach write_homecoach
 
 	m_bPollHomeStatus = find_scopes();
--- a/hardware/Netatmo.cpp
+++ b/hardware/Netatmo.cpp
@@ -105,6 +105,7 @@ CNetatmo::CNetatmo(const int ID, const std::string& username, const std::string&
 
 	Debug(DEBUG_HARDWARE, "Netatmo Actif Scopes %s", m_scopes.c_str());
 
+	m_bPollWeatherData = false;
 	// Weatherdevice is available in HomesData + HomeStatus
 	//m_bPollWeatherData = (m_scopes.find("station_R") != std::string::npos);      //read_station
 	m_bPollHomecoachData = (m_scopes.find("homecoach_RW") != std::string::npos);  //read_homecoach write_homecoach
--- a/hardware/Netatmo.cpp
+++ b/hardware/Netatmo.cpp
@@ -106,7 +106,7 @@ CNetatmo::CNetatmo(const int ID, const std::string& username, const std::string&
 	Debug(DEBUG_HARDWARE, "Netatmo Actif Scopes %s", m_scopes.c_str());
 
 	// Weatherdevice is available in HomesData + HomeStatus
-	m_bPollWeatherData = (m_scopes.find("station_RW") != std::string::npos);      //read_station
+	//m_bPollWeatherData = (m_scopes.find("station_R") != std::string::npos);      //read_station
 	m_bPollHomecoachData = (m_scopes.find("homecoach_RW") != std::string::npos);  //read_homecoach write_homecoach
 
 	m_bPollHomeStatus = find_scopes();


JurgenLB:
If you don't own a Weatherstation,
You must not select a weatherstation.

Only select the scopes (devices) you have.

Then there must be no error 

Kind regards

Tableflipper9001:
> If you don't own a Weatherstation, You must not select a weatherstation.
> 
> Only select the scopes (devices) you have.
> 
> Then there must be no error
> 
> Kind regards

Hi Jurgen,

i have not selected the weather station scope in either case in the gui. Due to no initialisation of m_bPollWeatherData the weatherstation is always selected in the code.

regards,

jordy

JurgenLB:
We only set `m_bPollWeatherData` in Netatmo.h and use it in Netatmo.cpp but the value is not set I was expecting it to be Default `false`

So the best solution is to `m_bPollWeatherData = false;` 

This is to limit the calls to the Netatmo Server, they have a limit on the server-calls.
The Weather station is available in the HomesData/HomeStatus this is why that line was comment out. (limiting the calls)

kind regards

Tableflipper9001:
Hi Jurgen,

Aha, I will patch this up when I get home.

Just to make sure, is there a way to get the scope from the database?
I would like to doublecheck if weatherstation isn't in there and it just got in an invalid state or something (where weather station is never removed from the scope list)

I have tried to reconfigure the hardware a dozen times, but everytime it gives the same result.

Regards,

Jordy

JurgenLB:
On startup there is a debug log-line off the selected scopes.

Kind regards

JurgenLB:
I have been thinking off this.
You have updated from a old stable to the new stable, I suspect.

And with a update the values from the variable are not changed. So 'm_bPollWeatherData' is true 

I think if you stop Domoticz and then start Domoticz 
Your problem will also be gone.
Because off startup m_bPollWeatherData must be False.

But solving this in the code is more preferred.

Kind regards

Tableflipper9001:
> I have been thinking off this. You have updated from a old stable to the new stable, I suspect.
> 
> And with a update the values from the variable are not changed. So 'm_bPollWeatherData' is true
> 
> I think if you stop Domoticz and then start Domoticz Your problem will also be gone. Because off startup m_bPollWeatherData must be False.
> 
> But solving this in the code is more preferred.
> 
> Kind regards

Define old an new.

I have been running the netatmo integration for quite a while already (even before the OAuth 2 update from netatmo).
I think it broke when I went from 2026.2 to beta in order to fix the buienradar integration (because they changed their json API). I can double check this against my error log (I keep a copy of all error logs). 
Recently I had to reinstall the raspberry pi in order to test the matter-js server (moved from 32-bit OS to 64-bit OS). I went back from the beta to 2026.2 (loaded in the domoticz.db file) (apparently buienradar rolled something back as well, cuz my dads domoticz with buienradar magically repaired itself).
Then with the latest update I updated to 2026.3, but this issue was already present prior to the update to 2026.3 stable

JurgenLB:
On my setup I don't have the `Weather 1` so my guess is that `m_bPollWeatherData = true` was present and by updating the m_bPollWeatherData is never unset on your system.

Before Commit 2dce594 (7 Sep 2024) `m_bPollWeatherData = true;` was used for Weatherstation and Homecoach with a 'else-statement'.

The last Commit that used m_bPollWeatherData is Commit 9d9c380 (4 Apr 2026)

Did you fully stop / start your system ?

kind regards

Tableflipper9001:
Hi Jurgen,

I just checked my logging and can see it started for me on 01st of june.
This is still before I moved to the beta for buienradar (right after 12th of june).

Regarding:
_On my setup I don't have the Weather 1 so my guess is that m_bPollWeatherData = true was present and by updating the m_bPollWeatherData is never unset on your system_

This is what gave me the clue that I was trying to access information of a device I do not have.

I think you are correct that the value got set before 4th of april and after it cannot be disabled anymore (probably on by accident in order to login again).
Like I said before I did domoticz hardware start / stops, token refresh (so stop the hardware let the token expire and then re-init hardware), system shutdowns and even completely migrated to a new OS version.

I also checked the domoticz.db hardware table and only homecoach_RW is in the password field.
I take it the extra field are the access and refresh tokens?

also enabled the hardware debug logging:
image

Tableflipper9001:
> We only set `m_bPollWeatherData` in Netatmo.h and use it in Netatmo.cpp but the value is not set I was expecting it to be Default `false`
> 
> So the best solution is to `m_bPollWeatherData = false;`
> 
> This is to limit the calls to the Netatmo Server, they have a limit on the server-calls. The Weather station is available in the HomesData/HomeStatus this is why that line was comment out. (limiting the calls)
> 
> kind regards

Hi Jurgen,

I was re-checking my proposed fix and also what was said in the comment above.
I came to the conclusion I would have to replace the line I uncommented with = false.
I would prefer to not add = false in the header, none of the member variables are initialised in the header file:
image

All variables are instead initialised in the constructor (as it should be).

The proposed fix does not consume any additional calls on the remote server rate limit. See the picture below.
image
image


The top arrow indicates that the scopes are loaded from the password database field.
And at the bottom arrow nothing has been grabbed from the netatmo server yet. 
It only grabs things in do_work


The following logging snippet including hardware debug proves this:


The following logging confirms no transfers to the server are done during init.
image

Tableflipper9001:
Also regarding my statement that the C++ default init is true. That is also bollocks. In C++ a bool is a primitive and once you uninitialize it, it has the content of whatever is in memory in it. So theoretically this can be 0 or a value.
Since in modern systems bool has a unsigned integer underlying type it can be anything, but only 0 is false.

JurgenLB:
Maybe some off my wording was incorrect.

But if you now go back to the original is there still a problem?

Putting `m_bPollWeatherData = false;` on line 107
and leaving line 109 as is.
Should preserve the code and give you a errorless setup.
this is when someone wants to use the old code it can be done easily

kind regards

Tableflipper9001:
I just reproduced the problem using 2026.3 stable with a clean install
-> initialise Netatmo the first time, seemed to work ok
-> stop domoticz
-> start domoticz
-> problem immediately reproduces on first contact to netatmo service

Put this in locally:
_Putting m_bPollWeatherData = false; on line 107
and leaving line 109 as is._

I'm quite confident this will fix the problem, since the fix in the PR also returns false via find_scopes.
Unfortunately my pi decided that it has to rebuild domoticz completely, this will take some time.
I will push the change once its validated.

My concern is ```m_bPollWeatherData = false;``` and the comment together pollute the code.
For the outside user there is no way to configure to use the old method (except for compiling the code from source with modifications)
Is there any plan of refactoring m_bPollWeatherData out?

Tableflipper9001:
As expected, works like a charm.
Also rebased the work on the latest commit of the development branch

Tableflipper9001:
so whats next?
Does github have to run some auto formatting, building and unit test or?

gizmocuz:
@Tableflipper9001 , that is done automatically.
Thank you both for your for testing and your time. PR mergerd (the other one)

JurgenLB:
@Tableflipper9001 , I did not wan't to offend you

You are correct that the dead-code must be fased-out

kind regards

Tableflipper9001:
@JurgenLB don’t worry about it.
I appreciate your support in solving the issue this quickly. The problem is solved and some code clean up is done, which makes the engineer in me happy again :)

PR #6960, opened 2026-08-12
Fase out GetWeather

JurgenLB:
There is a report off Error 13 'Not correct scope' Error.
This is because m_bPollWeatherData is not uses anymore and I expect it to be False by Default.
The Weatherstation is available in HomesData and HomeStatus.

So I removed GetWeather for a clean-code.

If you wan't to keep the code in you can accept #6955 and disregard this PR.

Kind Regards


Tableflipper9001:
Nice, could have also left a comment in the other PR ;)

gizmocuz:
Thank you !

PR #6968, opened 2026-08-18
Plugins: allow bar ranges to be set through the device Color field

Rouzax:
## What this does

`DeviceStatus.Color` is dual purpose. It holds RGB/WW color state for color switches, and it holds the bar range definitions that the utility, temperature and weather cards render (`www/app/widgets/dzBar.js`).

The plugin framework predates the second use. Every plugin write round-trips through `_tColor`, and `_tColor::fromJSON` leaves the mode at `ColorModeNone` unless the payload is a JSON object carrying a valid `"m"` member, at which point `toJSONString()` returns the empty string. A range payload was therefore blanked before it ever reached the database.

The practical effect: a plugin can create a Custom Sensor, Percentage or Temperature device whose card fully supports bars, but it cannot fill the bands in. The bars stay empty until the user types every band by hand in the edit dialog, even when the hardware already reports its own thresholds.

Plugin `Color` writes now go through one shared helper, `Plugins::NormalizeDeviceColor()`:

- a JSON object carrying an `"m"` member is normalized through `_tColor` exactly as before, so color switches are unaffected and the original "detect incorrectly formatted color data" intent is kept
- any other well-formed JSON is stored verbatim, which is what lets range payloads through
- anything that is not JSON at all is still rejected

Six call sites, so it wanted a helper rather than six copies: `PythonObjectEx.cpp` (unit refresh, create, update) and `PythonObjects.cpp` (legacy device refresh, create, `Device.Update(Color=)`).

## The second half, and why it is in the same PR

Relaxing that normalization makes plugin-supplied strings the first arbitrary text to reach the `UPDATE` built in `CSQLHelper::UpdateDeviceValue`, whose value was interpolated with `'%s'`:

```cpp
safe_query("UPDATE DeviceStatus SET %s='%s' , LastUpdate='%q' WHERE (ID == %s )", ...)
```

`sqlite3_mprintf` does not escape quotes for `%s` (`%q` does). A value containing a single quote closes the string literal and takes over the trailing `WHERE` clause. Built as a standalone program against libsqlite3, the payload `x' , Used='1' WHERE 1=1 -- ` produces a statement that `sqlite3_prepare_v2` accepts and that rewrites every row in the table.

That matters because it is the one way to escape the per-plugin scoping. Ownership is otherwise sound: `ID` is `READONLY` on both `CUnitEx` and the legacy `CDevice`, and is only ever assigned from SELECTs filtered by the plugin's own `HardwareID`, so a plugin cannot point a unit at another hardware's device. An injected `WHERE` replaces that predicate entirely.

The value is now escaped with `%q`, matching the UI path that writes the same column (`WebServerCmds.cpp`) and both plugin create paths, which already used `%q`. Only the update path was inconsistent.

The same unescaped overload was already reachable with arbitrary strings independently of this change, and the fix covers those too:

| Site | Fields | Where the string comes from |
|---|---|---|
| `PythonObjectEx.cpp`, `PythonObjects.cpp` | Name, Description, sValue | plugin, frequently remote device metadata |
| `DomoticzTCP.cpp` | Color, Options, LastUpdate | JSON received from a remote Domoticz master |
| `mainworker.cpp` | Description | SourceID |

Both halves are in one commit deliberately. Splitting them would leave an intermediate commit that accepts range payloads without escaping them.

## Testing

New `test/python/test_plugin_color_ranges.py`, following the pattern of the existing `test_calibration.py`: it starts its own Domoticz on a free port with a throwaway database and userdata folder, and drives a small example plugin (`plugins/examples/ColorRangeTest/`) copied into that folder for the run, so it never touches an existing installation.

Seven checks across five payloads: the bare range array (utility card shape), the keyed range object (temperature and weather card shape), a quote-bearing range payload, genuine RGB color state, and a payload that is not JSON. The quote-bearing case also asserts against a witness device created on separate dummy hardware, checking its `Name` and `Color` afterwards.

I ran it against three builds:

| build | result |
|---|---|
| this branch's base | 3 fail, 4 pass. Range payloads blanked, feature genuinely absent |
| helper applied but `%s` left in place | 5 fail. The witness device on unrelated hardware was renamed and its Color overwritten |
| this PR | 7 pass |

The middle row is the reason the escaping fix is here rather than in a separate PR.

Worth noting for anyone reviewing the test: JSON validity is no defence against the injection, since a single quote is legal inside a JSON string. The escaping is the control; validating the payload is only defence in depth.

Also run on the final build, both green: `test_calibration.py` 20/20 (the two MQTT cases skip here, `paho-mqtt` is not installed on my machine) and `test_api_sweep.py` 29/29. Built with `make -j$(nproc)`, exit 0, no new compiler warnings.

## Docs

`docs/Developing_a_Python_plugin.wiki` described `Color` as color state only in three places. Added a "The Color field" section covering both uses, the two range shapes, the key names, and the band fields, and pointed the three table entries at it. `test/README.md` gained a section for the new test.

## Possible follow-ups, not in this PR

- `dzUtilityWidget.js` `isBarSupported()` omits `Solar Radiation` while Lux is included. It looks like an oversight rather than intent, and it is a separate one-line change. Happy to raise it if you want it.
- The three Python tests under `test/python/` now carry three near-identical copies of the throwaway-instance harness. Extracting a shared module would be a tidy-up in its own right, and touching the two existing tests felt out of scope here.


Rouzax:
Let me know what you want with the test plugin I added.

gizmocuz:
Actually I do not think we should add the test plugin and also not a python/examples test for this?

PR #6969, opened 2026-08-18
fix(webserver,push): correct format specifiers and syntax in log mess…

kdahal7:
…ages

Fix invalid '%q' format specifiers in WebServer.cpp log calls (changed to '%s') and add missing closing parenthesis in BasePush.cpp log message.


gizmocuz:
Thank you for this! To many SQLite work lately that uses %q ;)

PR #6975, opened 2026-08-20
Remove the Color range example plugin and python test added in #6968

Rouzax:
Follow-up to #6968, per your comment there.

Removes the two things you flagged, and the `test/README.md` section that documented them:

- `plugins/examples/ColorRangeTest/`
- `test/python/test_plugin_color_ranges.py`

The fix itself is untouched: `Plugins::NormalizeDeviceColor()` and the `%q` escaping in `CSQLHelper::UpdateDeviceValue` stay exactly as merged, and both History.txt entries still describe shipped behavior. `test/README.md` is restored byte for byte to its pre-#6968 state, and nothing else in the tree referenced either path.


gizmocuz:

Issue #6980, opened 2026-08-22
css/remote.css is not themeable: the media remote paints from hardcoded colours while the dialog around it already uses --dz-* tokens


Rouzax:
### What this is about

`www/css/variables.css` defines 58 `--dz-*` tokens and the theming documentation invites themes to override them. The media remote popup does not participate: `www/css/remote.css` contains no `var(--dz-*)` at all, and the remote SVGs in `www/index.html` carry their colours as hardcoded presentation attributes and inline styles.

The result is visible rather than theoretical. The jQuery UI dialog that holds the remote already paints from `--dz-modal-bg`, so on a dark theme the popup background turns dark while the remote inside it stays on its fixed `#393a39` shapes, dark on dark.

Measured on 2026.3 (build 18230): of the files in `www/css/`, only `dashboard.css` (309 `var(--dz-*)` references), `variables.css` (48) and `style.css` (42) read any token. `remote.css` reads none.

### The colours in question

| Where | Value | Role |
| --- | --- | --- |
| `remote.css:13` `.remoteshadow` | `#393a39` | the shadow/edge layer drawn under every button |
| `remote.css:20` `.remotetext` | `#393a39` | the OK and CH labels |
| `remote.css:6` `.remotehoverable:hover` | `#289fe3` | hover feedback |
| `remote.css:10` `.remotedisabled` | `#2e2e2e` | disabled buttons |
| `index.html` remote `<defs>` | `#cbcccb` to `#989898` | the radial gradients that give the button faces their shape |
| `index.html` remote `<defs>` | `stroke="#393a39"`, `fill="#393a39"` | the symbol groups (menu, subtitle, repeat, shuffle, browse, ...) |

### Suggested shape

Reusing the existing button tokens does not work here, so this is not a one-line change: their defaults are nothing like the remote's values, and a `var(--dz-btn-border, #393a39)` style fallback never fires because `variables.css` sets those tokens in `:root`. Mapping the shadow layer onto `--dz-btn-border` (which resolves to `--dz-input-border`, `#ccc`) would visibly change the stock remote rather than preserve it.

A small named set with today's values as the defaults keeps stock rendering identical by construction and gives themes a hook:

```css
--dz-remote-face-start:  #cbcccb;
--dz-remote-face-end:    #989898;
--dz-remote-edge:        #393a39;   /* the shadow layer */
--dz-remote-symbol:      #393a39;   /* symbols, OK and CH labels */
--dz-remote-hover:       #289fe3;
[...]

PR #6983, opened 2026-08-23
Make the media remote popup themeable

Rouzax:
Producing the before/after pictures asked for in #6980 meant building and verifying the change anyway, so here it is as a PR rather than as more screenshots.

`www/css/remote.css` read none of the `--dz-*` tokens and the remote SVGs in `www/index.html` carried their colours as presentation attributes and inline styles, so the media remote popup could not follow a theme while the jQuery UI dialog around it already paints from `--dz-modal-bg`. Details and measurements in #6980.

**What this does**

Six tokens in `:root`, with today's colours as their defaults, so stock rendering is unchanged by construction:

```css
--dz-remote-face-start:  #cbcccb;
--dz-remote-face-end:    #989898;
--dz-remote-edge:        #393a39;   /* the shadow layer */
--dz-remote-symbol:      #393a39;   /* symbols, OK and CH labels */
--dz-remote-hover:       #289fe3;
--dz-remote-disabled:    #2e2e2e;
```

`remote.css` reads them instead of the literals. The twelve gradient stop pairs across the three remotes become `stop-color: var(--dz-remote-face-start/-end)` in their existing inline styles. The eleven `stroke="#393a39"` / `fill="#393a39"` presentation attributes on the symbol groups become two classes, `.remotesymbolstroke` and `.remotesymbolfill`, because a presentation attribute cannot take a `var()`.

**Verification**

Domoticz 2026.3 (build 18263), default theme, all three remotes (Kodi/Panasonic, LMS, HEOS), Chromium and Firefox, fixed viewport, service worker blocked so the CSS could not come from cache: `compare -metric AE` reports 0 differing pixels, and every one of the 382 shapes keeps its computed fill, stroke and stop-color. Before/after pictures are in #6980.

**Note on the token set**

`.remoteshadow` currently paints two different things, the offset shadow copy under each button and most glyph shapes, so `--dz-remote-edge` and `--dz-remote-symbol` cannot actually be driven apart without first sorting the 105 elements that carry that class into shadow copies and glyphs. If you would rather have four tokens than six, merging edge and symbol into a single ink matches today's markup exactly; say the word and I will squash them.

### DIFF
--- a/www/css/remote.css
+++ b/www/css/remote.css
@@ -3,21 +3,21 @@
 }
 .remotehoverable:hover {
 	cursor: pointer;
-	fill: #289fe3;
+	fill: var(--dz-remote-hover);
 }
 .remotedisabled {
 	pointer-events: none;
-	fill: #2e2e2e;
+	fill: var(--dz-remote-disabled);
 }
 .remoteshadow {
-	fill: #393a39;
+	fill: var(--dz-remote-edge);
 	pointer-events: none;
 }
 .remotetext {
 	cursor: pointer;
 	text-anchor: middle;
 	dominant-baseline: middle;
-	fill: #393a39;
+	fill: var(--dz-remote-symbol);
 	font-size: 180px;
 	font-family: sans-serif;
 	font-stretch: ultra-condensed ;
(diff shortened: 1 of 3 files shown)


gizmocuz:
Thank you for this PR!!

PR #7003, opened 2026-08-28
Icon picker follow-ups: cross-source search, categories, floorplan

galadril:
…yphs

Picker:
- Search spans every source at once, with results grouped per provider.
- Font Awesome category chips, filtered against the glyphs actually installed so a missing stem cannot render a blank tile.
- The truncation note carries real counts instead of "keep typing".
- A library whose stylesheet fails to load is reported as such rather than as an empty one.
- Icon sets that draw with SVG masks instead of a font (no @font-face and no codepoint) are enumerated.
- A built-in Domoticz icon previews as whatever the chosen icon style will actually draw for it.

Floorplan:
- Device icons resolve through dzIconService, so an icon set on the device and the Settings > Icon style choice are both honoured; glyphs render in a foreignObject and keep the hit area an image had.
- Icons redraw when the style setting arrives or is changed, which it can be after the first devices are already drawn.
- Each value of a split compound sensor resolves to its own glyph; the splitter's CustomImage is a flag for the legacy image, not an icon index, and was resolving all parts to custom icon 1.
- The static humidity and barometer images are no longer treated as value-driven artwork, so the glyph style reaches them.

Web assets:
- Quoted url() values are kept whole when a stylesheet is rewritten. A nested url(#clip) inside an inline SVG truncated the value, and the leftover quote made the data: test fail, so 21 icons per install were replaced with about:invalid.
- A failed rollback is reported instead of being swallowed, and the metadata write is verified before the backups are discarded.


gizmocuz:
Merged 🎉 — thanks @galadril. As a follow-up layered on top of this, we finished converting the remaining **chrome/affordance icons** to glyphs (the bits this PR intentionally left as PNGs), plus a picker fix and some cleanup.

**Chrome icons → themed glyphs** (`d14fd90`)
- Floorplan popup: webcam, the expand "twisty" and the favourite star now draw as glyphs, via a small `makeSVGglyph` foreignObject helper that sits alongside your `makeSVGglyphnode`/`makeSVGiconnode` (the dimmer slider handle is left as-is).
- Floorplan editor room table: the ok/failed status glyphs go green/red.
- Hardware setup (Matter / Z-Wave / EnOcean): battery & RF-signal header icons.
- Device/table header signal & battery icons now take the header text colour instead of grey.
- Trend arrows (temperature / mobile / utility + the report trend column): red up / blue down / grey stable, matching the old PNGs; the report month/day drill chevrons are coloured so they're visible on the light rows.
- Battery level renders as a `fa-battery-*` glyph filled and coloured by level (red < 10%, amber < 40%, green otherwise) in the devices and Matter tables.
- Removed the now-dead level-bar CSS and 12 image files nothing loads anymore (they were only referenced as glyph-map keys).

**Icon picker: slowly-loading libraries** (`94449d0`) — addresses the greptile P1 on `dzIconPicker.js`. `ensureStylesheet` resolved after a fixed 600 ms whether or not the stylesheet had parsed, so a large/slowly-served library was marked exhausted and stayed missing from "All icons" until a reload. It now waits for the sheet's rules to actually become readable (polling up to a generous cap, with an error fallback).

On the other greptile P1 (`WebAssetFetch.cpp` metadata rollback): this looks like a false positive — the only branch that writes the row (when `StoredMetadataMatches` fails) already calls `RestoreMetadata(szName, existing)`, so both the files and the row are rolled back.

PR #7007, opened 2026-08-28
Fix Meter_Calendar tariff spike check for non-energy meters

### LINKED ISSUE #7005: Meter_Calendar Price reset to 0 for valid non-energy tariffs above 3/unit
### Problem

`AddCalendarUpdateMeter()` applies a hard-coded maximum unit price of 3.0 to all meter types. This causes valid `Meter_Calendar.Price` values to be discarded for non-energy meters whose tariff can legitimately exceed 3 per unit.

### Reproduction

Domoticz V2026.3-stable, build f734fde31.

Meter configuration:
- meter type: `MTYPE_GAS`
- divider: 1000
- unit price: 40.97

Example: raw daily delta 29 gives `29 / 1000 = 0.029` units. Correct daily price is `0.029 * 40.97 = 1.18813`, but the current spike protection permits only `0.029 * 3.0 = 0.087`, so the valid calculated price is replaced by 0.

Historical data shows the transition:
```
2026-05-30 Value=18 Price=0.7375
2026-05-31 Value=20 Price=0.0000
2026-06-01 Value=24 Price=0.0000
```

### Expected behavior

A tariff sanity threshold intended for energy meters should not invalidate prices for gas, water or generic counters using different units.

### Proposed fix

Apply the existing 3.0/unit spike protection only to `MTYPE_ENERGY` and `MTYPE_ENERGY_GENERATED`. The separate P1/MultiMeter check remains unchanged.

The proposed fix compiles successfully against the current development branch and has been validated locally.

martin-g-it:
### Summary

Restrict the hard-coded 3.0/unit tariff spike protection in `AddCalendarUpdateMeter()` to energy meter types.

### Problem

The current check is also applied to gas, water and generic meters, although those meter types use different units and can legitimately have unit prices above 3.

For example, with:
- meter type: `MTYPE_GAS`
- divider: 1000
- unit price: 40.97
- raw daily delta: 29

The correct result is:
`29 / 1000 * 40.97 = 1.18813`

The current sanity limit is:
`29 / 1000 * 3.0 = 0.087`

The valid calculated price is therefore discarded and `Meter_Calendar.Price` becomes 0.

### Fix

Keep the existing 3.0/unit spike protection for `MTYPE_ENERGY` and `MTYPE_ENERGY_GENERATED`. Do not apply that energy-specific threshold to meter types using other units.

The separate P1/MultiMeter spike protection remains unchanged.

### Validation

- Reproduced on V2026.3-stable / f734fde31
- Local patched stable build validated against the affected meter
- Current development branch with this change builds successfully

Fixes #7005

### DIFF
--- a/main/SQLHelper.cpp
+++ b/main/SQLHelper.cpp
@@ -8381,10 +8381,13 @@ void CSQLHelper::AddCalendarUpdateMeter()
 				CalcMeterPrice(ID, divider, szDateStart, szDateEnd, price);
 				if (price != 0.0f && total_real > 0)
 				{
-					// Spike protection: discard price if implied tariff exceeds max plausible rate
-					constexpr float max_unit_price = 3.0f;  
-					if (std::abs(price) > (static_cast<float>(total_real) / divider) * max_unit_price)
-						price = 0;
+					// Spike protection only applies to energy tariffs; other meter types use different units.
+					if ((metertype == MTYPE_ENERGY) || (metertype == MTYPE_ENERGY_GENERATED))
+					{
+						constexpr float max_unit_price = 3.0f;
+						if (std::abs(price) > (static_cast<float>(total_real) / divider) * max_unit_price)
+							price = 0;
+					}
 				}
 				else if (total_real <= 0)
 					price = 0;


gizmocuz:
Sorry, I was to fast ;)

martin-g-it:
Gheh, but the change is merged, right?

gizmocuz:
Yep, it is already in the latest beta

martin-g-it:
All good then! 👍🏻

Issue #6998, opened 2026-08-27
Worth modernizing DataTables to 2.x/3.x for generic table theming?


Rouzax:
This is the bigger companion to the small security bump (#6997). I wanted to raise it and hear whether
you think it is worth doing before anyone writes code, because it is a large, breaking change.

The thing that got me looking is table theming. On 1.10.x DataTables adds no per-column type
classes, so a theme that wants numeric or date columns right-aligned has to lean on fragile,
per-table `nth-child` maps that break the moment a column is added or reordered. From 2.x
onwards DataTables auto-detects column types and stamps `dt-type-numeric` / `dt-type-date` on
the cells, so a theme can just do:

```css
td.dt-type-numeric { text-align: right; }
```

and every table is covered instead of the handful that are hand-mapped today. On top of that it
would let us drop the legacy Hungarian-notation API (deprecated since 1.10, removed in 2.0) and
the Flash-era TableTools dependency.

## Which major version, though?

That is the first thing I would want your view on. DataTables 2.x has been out about two years
and is still jQuery-based, which fits our current front end. DataTables 3.x only landed on
2026-07-31 (latest 3.0.2) and its headline is that it no longer depends on jQuery at all, which
is appealing longer term but very fresh and a bigger single jump from 1.10. My leaning is 2.x as
the lower-surprise target for now and revisit 3.x once it has soaked, but I am genuinely open,
3.x is defensible if you want to move off jQuery over time.

## What it actually costs

I had a look at the footprint so we are not guessing. It is roughly **115 table initialisations
across 36 files** in `www/app/`, all on the legacy API that 2.x removed. The renames alone:

| Legacy option | Count | Modern replacement |
|---|---|---|
| `mData` | 77 | `columns.data` |
| `sDom` | 39 | `dom` / `layout` |
| `bSort` | 39 | `ordering` |
| `sClass` | 38 | `columns.className` |
| `bJQueryUI` | 33 | removed (DataTables' own CSS) |
| `oTableTools` | 31 (22 files) | Buttons + Buttons.colVis |
| `bProcessing` | 31 | `processing` |
| `aaSorting` | 26 | `order` |
| `sTitle` | 8 | `columns.title` |
| `aoColumnDefs` | 5 | `columnDefs` |
| `sWidth` / `bVisible` | 4 each | `columns.width` / `columns.visible` |
| `bFilter` / `fnDrawCallback` | 3 each | `searching` / `drawCallback` |
[...]

Issue #6997, opened 2026-08-27
Bundled DataTables is 1.10.15 (2017) with known advisories, worth a small bump?


Rouzax:
While looking at table styling from the theme side I noticed the web UI still bundles
**DataTables 1.10.15** from 2017 (`www/js/data-tables/jquery.dataTables.min.js`), plus
**TableTools 2.2.3** (2014, which carries the old Flash-based ZeroClipboard) and
**Select 1.2.5**. jQuery itself is already on 3.6.0.

1.10.15 has three published advisories, all fixed later in the v1 line:

- Prototype pollution (High), fixed in 1.10.23. The internal data-set function does not
  guard `__proto__` / `constructor`.
- Prototype pollution (Medium), fixed in 1.10.22.
- XSS (Low), fixed in 1.11.3, where the HTML-escape helper misses the contents of an array value.

The nice thing is that all three are cleared just by moving to the last v1 release
(**1.13.11**). Staying on the v1 line keeps the deprecated-but-still-supported legacy API and
the jQuery UI styling path working, so none of the existing table setups have to change. I
kept this deliberately separate from the bigger DataTables 2.x/3.x modernization (#6998, which is the
one that actually helps theming) so a security bump does not get stuck behind a breaking rewrite.

I would be happy to do the swap and test through the list pages (Devices, Hardware, Users,
Setup lists, Scenes, Timers, Notifications, the reports, and the hardware-setup sub-tables),
and regenerate the `.gz` assets.

The one thing I would want your steer on is **TableTools**. It is end-of-life upstream (retired
in favour of Buttons + Select, and it still ships Flash-era ZeroClipboard). The shared init in
`www/app/app.constants.js` (`dom: '<"H"lfrC>t<"F"ip>'`) and the `oTableTools` blocks in a
couple of dozen files use it only for the column-visibility (ColVis) button. It is not
guaranteed to run against 1.13 core, so I see three ways to go:

- Keep it if it still works on 1.13 (no feature change, but the dead Flash code stays).
- Drop the ColVis button here (remove `C` from the shared `dom` and the `oTableTools` blocks),
  which loses a rarely-used control but gets rid of the oldest dependency now.
- Replace it with Buttons + Buttons.colVis, though most of that value really belongs with the
  larger modernization work.

My instinct is to try the first, and fall back to dropping the button if it breaks, leaving the
Buttons replacement for the modernization effort. Does that sound right to you, or would you
[...]

PR #7018, opened 2026-09-04
Update DataTables to 1.13.11 and remove the unused TableTools payload

Rouzax:
Closes the security part of #6997, in the shape agreed there: stay on the v1 line, restore the jQuery UI look with a shim, drop the dead TableTools payload, regenerate the `.gz` assets. The legacy `fn*` API conversion (option 2) follows as its own PR.

## What changed

- DataTables core **1.10.15 (2017) to 1.13.11**. That clears the three advisories against 1.10.15: prototype pollution (High, fixed 1.10.23), prototype pollution (Medium, fixed 1.10.22) and XSS (Low, fixed 1.11.3).
- Select **1.2.5 to 1.7.0**, the last release on the v1 line. Select is genuinely used, by everything going through `dataTableDefaultSettings` and by `rows().select()` / `.deselect()` in `Devices.js`, and 1.2.5 is too old to pair with 1.13 core.
- `dataTables.tableTools.min.js` and `js/ZeroClipboard.js` removed, along with their two `<script>` tags. TableTools was never initialised: it registers only `cFeature: "T"`, and no `dom` string in the tree contains `T`. Row selection is done by hand through `fnGetSelected()` in `www/js/domoticz.js`, which returns `tr.row_selected` and has nothing to do with the TableTools API. Every page was downloading and parsing both files for nothing.
- The no-op `C` is gone from all 42 `dom` strings, and the 33 inert `"oTableTools": { "sRowSelect": "single" }` blocks with it. There is no `dataTables.colVis.js` anywhere in `www/`, so `C` was an unrecognised letter that DataTables silently skipped; no ColVis button renders today on any page.
- New `www/js/data-tables/dataTables.jqueryui-classes.js`. `bJQueryUI` was dropped from core after 1.10, and 1.13.11 contains no `oJUIClasses` and no `ui-*` class strings, so a straight file swap would leave every list page with unstyled toolbars, headers and pagination. 1.13 does still resolve the `<"H"…>` shorthand through `sJUIHeader`/`sJUIFooter`, still derives `sSortingClassJUI` from the `sSortJUI*` slots, and still ships the `jqueryui` header renderer; the slots just default to empty. Filling them with the values 1.10.15 used reproduces the current DOM exactly, so `css/demo_table_jui.css` and all the call sites stay untouched. It is one small file to swap or delete when the tables are restyled.
- `bJQueryUI: true` is left in place. It is inert on 1.13 once the shim is in, and stripping it from 35 call sites is churn that belongs with #6998.

## Two things that only showed up once it was running

**1.13 converts camelCase options with `force` on, 1.10 did not.** The init call went from `_fnCamelToHungarian(defaults, options)` to `_fnCamelToHungarian(defaults, options, true)`. With force, a camelCase key overwrites an explicitly set Hungarian twin in the same options object; without it, the existing Hungarian key was left alone. `dataTableDefaultSettings` sets camelCase `dom` and `order`, and nine call sites merge it and then set the Hungarian form, so on 1.13 the shared value silently won and their own layout was lost:

| file | key | would have regressed to |
|---|---|---|
| `app/report/{Counter,EnergyMultiCounter,Rain,Temperature,Wind}Report.js` | `sDom: '<"H"rC>t<"F">'` | search box, length menu, info line and pagination on all five report tables |
| `app/events/CurrentStates.js` | `sDom: 't'` | full chrome on a table meant to be bare |
| `app/DeviceLightEdit.js` | `sDom: '<"H"frC>t<"F"i>'` | the shared chrome |
| `app/hardware/setup/Matter.js` | `aaSorting: [[0, 'asc']]` | sorted descending |
| `app/hardware/setup/ZWave.js` (x2) | whole literal | nothing, see below |

Those sites now use the camelCase key, which is correct on 1.10 and on 1.13. The two ZWave literals turned out to repeat `dataTableDefaultSettings` key for key with identical values, so they collapse to the merge on its own. I scanned every DataTables init literal in `www/` for other camelCase/Hungarian pairs and there are none.

**1.13 leaks its column-width probes.** `_fnCalculateColumnWidths` appends one temporary `<div>` per sized column to the table container, reads `offsetWidth`, then calls `$(probes).remove()` on an array of jQuery objects, which resolves to no elements and removes nothing. Measured on the devices list: 13 empty divs after load, 52 after three window resizes, against 0 on 1.10.15, which measured one column at a time and removed each probe properly. DataTables 2 rewrote the sizing code and does not have it, and 1.13.11 is the last release on the 1.x line, so there is no version to upgrade to. `www/js/data-tables/dataTables.width-probe-cleanup.js` removes them on `init.dt` and `column-sizing.dt`; it goes away with #6998.

## Verification

Two local builds side by side, both on the **stock `default` theme** and the same database copy of about 230 devices: one serving `www/` from `development`, one serving `www/` from this branch.

A harness visited 29 routes on both and dumped, per `.dataTables_wrapper`, the complete child list with classes, the row count, the first header cell class and inner HTML, the pagination classes, the info text, whether the length select and the filter input exist, and every console error. Routes: Devices, Hardware, Users, UserVariables, Scenes, Timerplan, Mobile, Applications, AccessTokens, DPHttp, DPMQTT, DPInflux, DPFibaro, DPGooglePubSub, Floorplanedit, Cam, CustomIcons, Events, Log, History, Roomplan, four `Devices/:id/Report` pages, and the per-device Timers, Notifications and Log tables.

The output is identical between the two, console errors included. Screenshots are pixel-identical apart from live clock text in the Last Seen column, checked by cropping the diff bounding box.

Also confirmed on the new build: version reports 1.13.11 with Select 1.7.0, `TableTools` and `ZeroClipboard_TableTools` are undefined, the feature registry is empty, `dataTables.tableTools.min.js` 404s, no `.ColVis` or `.DTTT` node exists on any page, `stateSave` restores cleanly from state written by 1.10, and the width-probe count stays 0 after load, after idling and after three resizes.

## Screenshots

Top is `development` (1.10.15), bottom is this branch (1.13.11), stock theme.

Devices, the full chrome: length menu, search box, sort arrows, info line, pagination.

A counter report. This is the one the `sDom` fix above is about: it keeps its bare layout with no search box, length menu or pagination.

Users, a small table with the toolbars and the disabled pagination buttons.

<h3>Greptile Summary</h3>

The PR upgrades DataTables core and Select while preserving the existing jQuery UI presentation and legacy table behavior.

- Removes unused TableTools, ZeroClipboard, ColVis DOM markers, and inert TableTools configuration.
- Adds compatibility shims for the former jQuery UI class map and a DataTables 1.13 width-probe cleanup.
- Updates conflicting table options to camelCase and regenerates the compressed library assets.

The PR appears safe to merge with no concrete changed-code failure identified.

The upgraded assets are loaded in the required order, packaging includes the new compatibility scripts, obsolete plugin configuration is removed consistently, and the table-specific option conversions preserve their intended overrides.

| Filename | Overview |
|----------|----------|
| www/index.html | Loads the upgraded DataTables stack and compatibility shims while removing the unused TableTools and ZeroClipboard scripts. |
| www/js/data-tables/dataTables.jqueryui-classes.js | Restores the legacy jQuery UI class slots and renderer expected by the existing table DOM and stylesheets. |
| www/js/data-tables/dataTables.width-probe-cleanup.js | Removes leaked DataTables 1.13 width probes using a selector consistent with their generated DOM signature. |
| www/app/app.constants.js | Removes the inactive ColVis feature marker from shared DataTables defaults. |
| www/app/hardware/setup/ZWave.js | Replaces duplicate DataTables option literals with the equivalent shared defaults. |
| www/app/hardware/setup/Matter.js | Uses the camelCase ordering option so the table-specific ascending order overrides shared defaults under DataTables 1.13. |
| www/js/data-tables/jquery.dataTables.min.js.gz | Updates the bundled compressed DataTables core to the intended 1.13 release. |
| www/js/data-tables/dataTables.select.min.js.gz | Updates the compressed Select extension to a compatible v1 release. |

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(webui): update DataTables to 1.13.11..."](https://github.com/domoticz/domoticz/commit/c575ab2b44009025512085d25e4f8df5e55cb518) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=60308025)</sub>


gizmocuz:

PR #7019, opened 2026-09-04
Retire the legacy DataTables fn* API (follow-up to #7018)

Rouzax:
The follow-up you agreed to in #6997: retire the legacy DataTables `fn*` API so the 2.x move in #6998 is close to a drop-in instead of a change to 24 files. No behaviour, styling or configuration change.

> **This sits on top of #7018.** It branches off it, so until #7018 merges the Files changed tab shows both PRs (42 files). The change this PR is actually asking you to review is the single commit a667aabb3, which is 29 files and 192 insertions for 192 deletions, nothing added and nothing removed. Happy to rebase onto `development` the moment #7018 lands, and equally happy to hold this until then if you would rather see it clean.

## Why now and not with #6998

1.13 is the only release where the old and the new API both work. The modern equivalents return identical results on 1.13.11, 2.3.2 and 3.0.0, so this can be written and verified page by page against the version now in the tree, and then lands unchanged on whichever version #6998 ends up targeting. Doing it here also keeps #6998 down to the config shape and the styling, instead of bundling a 94 site rewrite with a major version bump.

## The conversion

94 call sites across 24 files: `fnClearTable` 31, `fnAddData` 30, `fnGetData` 30, `fnGetNodes` 3.

| legacy | modern |
|---|---|
| `fnClearTable()` | `clear().draw()` |
| `fnAddData(row)` | `row.add(row).draw()` |
| `fnGetData(node)` | `row(node).data()` |
| `this.fnGetNodes()` | `this.api().rows().nodes()` |

`fnClearTable()` redraws by default and `clear()` does not, hence the explicit `draw()`; without it a refresh that returns no rows would leave the previous ones on screen. `fnAddData` also redraws per call, so `row.add(row).draw()` keeps the existing behaviour rather than quietly batching it.

Three things that are not a one for one substitution:

**The modern methods are not on the object the old accessor returns.** `$(sel).dataTable()` gives you a jQuery object with the `fn*` methods bolted on; `clear()`, `row()` and `row.add()` live on the API object. So the 58 no-arg `$(sel).dataTable()` accessors that feed a converted call are now `$(sel).DataTable()`. To be sure none were missed I resolve every converted call back to its nearest preceding assignment and assert that assignment is a `.DataTable(` one; it reports no exceptions.

**Four tables were initialised through the old accessor and then used through the closure over its return value**, so their init calls had to move to `.DataTable({...})` as well: `AccessTokens.js`, and the three with a `fnDrawCallback` that closes over `oTable`, `DPFibaroController.js`, `DPHttpController.js` and `DPGooglePubSubController.js`. Left as they were, `oTable.row(...)` would have thrown the moment a row was clicked. Every other `dataTable({...})` init keeps its current configuration shape, since normalising those is the separate pass you parked.

**`fnAddData` returns the index of the row it added, and nothing in the tree ever reads it.** `row.add()` returns the API instead, so the 27 `var addId = ...` assignments went with the conversion rather than being left pointing at a different value.

I also converted the seven `table.dataTable().api()` bridges in the report files to `table.DataTable()`. After this there is no legacy accessor left anywhere in `www/` outside the `dataTable({...})` init calls themselves.

## Verification

Two local builds side by side on the same database, one serving #7018 and one serving this branch, both on the stock `default` theme. Three harnesses, all reporting identical output between the two:

- The 29 route DOM dump from #7018: per `.dataTables_wrapper`, the full child list with classes, the row count, the first header cell, the pagination classes, the info text, the length select and filter presence, and every console error. This is what shows `clear()` and `row.add().draw()` build the same populated tables that `fnClearTable()` and `fnAddData()` did.
- A selection harness: clicks the first row of the first list table on 17 routes and snapshots what the selection populates, the `javascript:` hrefs built out of the row data, the visible update and delete buttons, and every visible form field value. This is the part that exercises `fnGetSelected()` and the row data lookup. 16 of the 17 routes had rows to click, the three `fnDrawCallback` pages among them, which is what shows those closures still bind their click handlers.
- A scene editor harness: `#scenedevicestable` and `#scenedactivationtable` live in a view opened by the global `EditSceneDevice()` rather than by a route, so no route walk reaches them. Opened a group with 23 devices and clicked a row there too.

The only console error on either side is the pre-existing Angular `childNodes` one, unrelated to any of this.

No History.txt entry: nothing here is visible to a user.

<h3>Greptile Summary</h3>

The PR upgrades the bundled DataTables assets and migrates WebUI table operations from the legacy `fn*` interface to the modern API while preserving redraw and row-selection behavior.

- Replaces legacy clear, add, data-access, and node-access calls across WebUI controllers.
- Removes inactive TableTools and ZeroClipboard integration and related DOM configuration.
- Adds compatibility helpers for the existing jQuery UI classes and width-probe cleanup.
- Updates report-table API bridges and bundled DataTables/Select assets.

The PR appears safe to merge, with no concrete changed-code failure identified.

The migrated call sites use matching modern DataTables operations, preserve immediate redraw semantics, and retain reachable row-selection and callback behavior.

| Filename | Overview |
|----------|----------|
| www/app/AccessTokens.js | Initializes and refreshes the access-token table through the modern DataTables API while retaining per-row redraw behavior. |
| www/app/DPFibaroController.js | Converts table initialization, draw callbacks, row insertion, and selected-row data access to modern API equivalents. |
| www/app/DPGooglePubSubController.js | Migrates Google Pub/Sub link-table operations and callback node access without changing the selection workflow. |
| www/app/DPHttpController.js | Migrates HTTP link-table operations and draw callback handling to the modern DataTables API. |
| www/app/FloorplanEditController.js | Converts floorplan and room table refresh and selection paths while removing inactive TableTools configuration. |
| www/app/ScenesController.js | Converts scene-device and activation table refreshes and selected-row lookups to modern API calls. |
| www/app/hardware/Hardware.js | Migrates hardware and LMS-node table clearing, insertion, selection, and data retrieval. |
| www/app/hardware/setup/ZWave.js | Converts Z-Wave table operations and removes obsolete TableTools configuration. |
| www/js/data-tables/dataTables.jqueryui-classes.js | Restores the project’s expected jQuery UI-oriented DataTables classes and renderer defaults. |
| www/js/data-tables/dataTables.width-probe-cleanup.js | Removes empty width-probe elements after DataTables initialization and column-sizing events. |
| www/index.html | Replaces obsolete extension loading with the new compatibility scripts and updated DataTables assets. |

</details>

<sub>Reviews (1): Last reviewed commit: ["refactor(webui): retire the legacy DataT..."](https://github.com/domoticz/domoticz/commit/a667aabb38a452a7ef5fe4a71cbe4a034e0b5f83) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=60318477)</sub>


gizmocuz:
Thank you for this PR!!! Much appriciated!

Issue #7017, opened 2026-09-03
Icon picker: with Icon style "Classic images" the current-icon preview is always the placeholder square, and no device type image can be picked


Rouzax:
### Prerequisites

- [x] I have searched existing issues and the forum for this problem
- [x] I am running the latest beta version

### Domoticz Version

2026.3 (build 18327)

### Platform

Docker (`domoticz/domoticz:beta`), Linux host

### OS Version

Debian (official image)

### Browser

Chromium 141

### Hardware/Plugin Type

Dummy hardware, any sensor device (Text, Custom Sensor, kWh, Counter, Lux, Soil Moisture)

### Describe the Bug

Two related problems on the icon picker, both measured on the **default theme** with Icon style set to classic images (`IconStyle = 0`).

**1. The current-icon preview never resolves a classic image.**

`dzIconPickerPreview.of()` (`www/app/icons/dzIconPicker.js:534`) keeps `dzIconService.resolveSelection()`'s answer only when it is a font glyph:

```js
var resolved = device ? dzIconService.resolveSelection(device, sel.customImage, '') : null;
if (resolved && resolved.kind === 'font') {
    return resolved;
}
var src = sel.customImage > 0 ? (item ? item.src : '') : defaultImage;
return src ? { kind: 'img', src: src } : { kind: 'none' };
```

A resolved `{kind: 'img'}` is discarded, `src` then falls back to the `defaultImage` binding, and the utility path does not pass one: `dzUtilityWidget.js` `mountIconPicker()` mounts with `{customImage, icon, device}`, and its own comment says the device record is the intended fallback ("Without the record the picker has no type to fall back on and previews nothing"). The result is `{kind: 'none'}`, rendered as `<i class="fa-regular fa-square">`.

Measured on the default theme, device "Weather Description" (General/Text, `CustomImage: 0`, `TypeImg: text`), same page load, only `IconStyle` differing:

| `IconStyle` | `resolveSelection(device, 0, '')` | `preview.kind` | rendered |
|---|---|---|---|
| 0 (classic images) | `{kind: 'img', src: 'images/text.png'}` | `none` | empty square |
| 1 (glyphs) | `{kind: 'font', cls: 'fa-solid fa-align-left'}` | `font` | correct glyph |

[...]

PR #7020, opened 2026-09-04
Icon picker: show the resolved image in the preview with Icon style "Classic images" (#7017)

Rouzax:
Fixes the first problem in #7017. The second one in that issue is a design question rather than a bug, so I have left it out of here and will follow up on the issue itself.

## The bug

`dzIconPickerPreview.of()` (`www/app/icons/dzIconPicker.js`) returns early only when `dzIconService` resolves the selection to a **font**, and otherwise rebuilds the image from the item under the cursor or from the `defaultImage` binding. Only the glyph Icon style ever resolves to a font, so in the classic style, which is the default, the resolved image was thrown away.

The mount in `dzUtilityWidget.js` passes the device record instead of a `defaultImage`, deliberately (its own comment says the record is the intended fallback), so there was nothing left to fall back on and the result was `{kind: 'none'}`, drawn as the empty-square placeholder.

Because all three previews go through `of()`, the same miss showed up in three places: the compact trigger in the edit dialog, the swatch in the modal, and the candidate shown by the "Use default" button.

## The fix

Fall back to the resolved image before giving up. Precedence is unchanged above that point: a `defaultImage` passed by a mount point still wins, which matters because `DeviceLightEdit.js` passes a switch icon there, and the item being hovered in the grid still wins while browsing, so the modal still previews the tile under the cursor.

## Measured

Two local builds on the same database and the stock `default` theme, one on `development` and one on this branch, calling `of()` and `dzIconService.resolveSelection()` on every real device record through the Angular injector.

| | classic images (`IconStyle = 0`, default) | glyphs (`IconStyle = 1`) |
|---|---|---|
| devices with `CustomImage == 0` | 151 of 232 | 151 of 232 |
| `development` | **151 previewed as `none`** | 151 as `font`, correct |
| this branch | 151 as `img`, every one equal to what `resolveSelection` returned | unchanged, 151 as `font` |
| explicit `defaultImage` still wins | 151 of 151 | 151 of 151 |
| hovered tile still wins for `customImage > 0` | 40 of 40 | 40 of 40 |

The glyph style is untouched: the `font` early return happens before anything that changed.

The issue quotes 158 of 239 devices; my test database has changed since I filed it, hence the different totals.

## Screenshots

Same device ("Weather Description", a Text device with no custom image), same Icon style, stock theme.

<h3>Greptile Summary</h3>

The PR fixes classic-style icon-picker previews by retaining the image produced by the icon resolver when neither a hovered item nor an explicit default image supplies the preview.
- Preserves the existing hovered-item and `defaultImage` precedence.
- Leaves glyph-style font previews unchanged.
- Documents the fix in `History.txt`.

The PR appears safe to merge, with no concrete blocking or independently actionable non-blocking issues identified.

The new fallback reuses the resolver’s image only after the existing explicit preview sources are exhausted, while the font path and established precedence remain unchanged.

| Filename | Overview |
|----------|----------|
| www/app/icons/dzIconPicker.js | Adds the resolved classic image as the final image fallback while preserving existing preview precedence. |
| History.txt | Records the corrected classic-style icon preview behavior. |

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(webui): show the resolved image in t..."](https://github.com/domoticz/domoticz/commit/08c0c4148bf71727d2b24b35f6b4f94acb6320be) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=60329933)</sub>


gizmocuz:

PR #7021, opened 2026-09-04
Media remote: paint it from the general button colours and drop the --dz-remote-* variables (#7016)

### LINKED ISSUE #7016: Media remote paints black on any theme that does not import css/legacy.css: css/remote.css reads --dz-remote-* without fallbacks
### Prerequisites

### Domoticz Version

2026.3 (build 18324)

### Platform

Docker (`domoticz/domoticz:beta`), Linux host

### OS Version

Debian (official image)

### Browser

Chromium 141 and Firefox 143, both measured

### Hardware/Plugin Type

Kodi Media Server / Logitech Media Server / HEOS by DENON (any media player with a remote)

### Describe the Bug

`css/remote.css` and the gradient stops in the three remote SVGs read the six remote tokens with no fallback value, for example `fill: var(--dz-remote-edge)` and `stop-color:var(--dz-remote-face-start)`. Their stock values are declared in `css/variables.css`, which reaches the page only through `css/legacy.css` (`@import url("variables.css")`), and `legacy.css` is itself imported by `styles/default/custom.css`, not by `index.html`.

A theme replaces `styles/default/custom.css`. Any theme that does not import `legacy.css` therefore has all six tokens unset, and an unset custom property makes the whole declaration invalid at computed-value time, so `fill`, `stroke` and `stop-color` fall back to their initial value: black. The remote renders as solid black shapes on a light dialog.

Measured on build 18324 with a theme that does not import `legacy.css`, against the default theme on the same build and instance, computed values only (never the presentation attribute):

| | this build, no `legacy.css` | default theme |
|---|---|---|
| `.remoteshadow` (44 / 34 / 26 elements) | `rgb(0, 0, 0)` | `rgb(57, 58, 57)` |
| `.remotesymbolfill`, `.remotesymbolstroke` | `rgb(0, 0, 0)` | `rgb(57, 58, 57)` |
| `.remotetext` | `rgb(0, 0, 0)` | `rgb(57, 58, 57)` |

Identical in both engines and on all three remotes (Kodi, LMS, HEOS).

### Steps to Reproduce

1. Run any theme whose `custom.css` does not `@import url("../../css/legacy.css")`.
2. Open a media player device with a remote, or call `ShowMediaRemote('name', 1, 'Logitech Media Server')` from the console, since the three remote dialogs are static markup in `index.html` and open without the hardware.
3. The remote paints as flat black shapes.

### Expected Behavior

The six tokens carry their stock values as `var()` fallbacks, so an undeclared token renders exactly as before and a theme opts in by declaring them: `var(--dz-remote-edge, #393a39)`, `var(--dz-remote-symbol, #393a39)`, `var(--dz-remote-hover, #289fe3)`, `var(--dz-remote-disabled, #2e2e2e)`, `var(--dz-remote-face-start, #cbcccb)`, `var(--dz-remote-face-end, #989898)`. This is what #7009 did for the icon-rework glyph colours, for the same reason.

### Note on the other reading

`styles/default/custom.css` marks that import "Don't remove this line!", so it is arguable that a theme is expected to keep it and this is theme error. Two things speak against leaving it there: the tokens were introduced by #6983 (mine) without fallbacks, which is the regression, and the rest of the tokenized surface already uses fallbacks specifically so a theme does not have to inherit core's stylesheet chain. Whichever way it is resolved, the current state is that a theme which follows the fallback convention everywhere else still gets a black remote here.

Rouzax:
Fixes #7016.

Reworked from the original fallback patch after review: rather than giving the remote's own colour variables fallbacks, the remote no longer has colour variables. It is painted from the general button values, so it looks like the rest of Domoticz and follows whatever a theme does to its buttons.

## What changed

The six `--dz-remote-*` declarations are removed from `css/variables.css`. `css/remote.css` and the gradient stops in the three remote SVGs read the general values instead:

| part | value |
|---|---|
| button face, upper stop | `var(--dz-btn-info-bg)` |
| button face, lower stop, and hover | `color-mix(in srgb, var(--dz-btn-info-bg) 80%, #000)` |
| glyphs and button text | `var(--dz-btn-primary-text)` |
| disabled button | `var(--dz-status-disabled)` |

Everything is driven from `--dz-btn-info-bg`, so a theme recolours the whole remote by setting the one value it already sets to style its buttons, and never has to know the remote exists. `.btn-info` is itself `linear-gradient(#5bc0de, #2f96b4)`, so a gradient face is what the default theme's buttons look like; a theme with flat buttons flattens the remote by giving both stops the same value.

The raised bevel is gone. That is not just a value change: `.remoteshadow` painted both the offset copy under each button and the glyph drawn on top of it, on `<use>`, `<circle>` and `<path>` elements alike, with nothing in the markup separating the two roles. Clearing that one fill would have erased every glyph along with every bevel. So the 53 offset copies are deleted and the 52 glyph elements are re-classed to the existing `.remotesymbolfill`, which is the whole of the `index.html` change beyond the gradient stops.

The custom buttons that `domoticz.js` builds from a Kodi device's `Extra` field lose their offset copy for the same reason, along with the hardcoded `fill="black"` on their label that the `.remotetext` rule already overrode.

## Appearance

Stock default theme, `development` on the left.

Kodi remote, development versus this branch

HEOS remote, development versus this branch

The Logitech Media Server remote changes in the same way.

## A theme drives it with one value

The same build, with a stylesheet that sets `--dz-btn-info-bg: #cc4444` and nothing else:

The remote following a theme that only changes --dz-btn-info-bg

## Measured

Contract run on a clean instance on the stock default theme, in Chromium and Firefox, on all three remotes, reading computed values only and never the presentation attribute. All pass:

- no rendered shape paints black, no `.remoteshadow` element survives, and `--dz-remote-edge` resolves to nothing
- every glyph computes `rgb(255, 255, 255)`
- the gradient stops resolve to `rgb(91, 192, 222)` and `rgb(73, 154, 178)`. The lower stop is deliberately derived rather than pinned to Bootstrap's `#2f96b4` (`rgb(47, 150, 180)`); the small difference is the cost of following the theme's own colour instead of a literal
- setting only `--dz-btn-info-bg` moves every gradient stop on every remote
- zero page errors in either engine

`www/` is only `install(DIRECTORY ...)` in `CMakeLists.txt`, never compiled or embedded, so this cannot affect the build.


gizmocuz:
Thank you for this PR.
Is there a reason why we not use the default css style/variables? (like normal blue buttons with white text/glyphs) ?
If we did, it would also match any theme style used later

Rouzax:
Good question, and the short answer is that the mechanism you are describing already exists: the six `--dz-remote-*` tokens are the seam a theme maps onto, and a theme does exactly that in one place rather than restyling the SVGs. Machinon, for example, maps all six onto its own palette:

```css
--dz-remote-face-start: color-mix(in srgb, var(--dz-modal-bg) 92%, var(--dz-body-text));
--dz-remote-face-end:   color-mix(in srgb, var(--dz-modal-bg) 80%, var(--dz-body-text));
--dz-remote-edge:       color-mix(in srgb, var(--dz-modal-bg) 55%, var(--dz-body-text));
--dz-remote-symbol:     var(--dz-body-text);
--dz-remote-hover:      var(--dz-accent-color);
--dz-remote-disabled:   var(--dz-status-disabled);
```

So "it would also match any theme style used later" is already true; this PR is only about a theme that declares none of them and does not inherit `css/variables.css` either, which today gets a black remote instead of the stock one.

On pointing the remote at the button variables directly: the obstacle is that the remote is not flat. Two of the six tokens are the two stops of the radial gradient that gives each button its domed face, and a third is the edge drawn under it. The button palette is flat single colours (`--dz-btn-primary-bg`, `--dz-btn-text`, ...) and there is no two-stop pair anywhere in `variables.css`, so a straight mapping would flatten the remote rather than recolour it. That is also why Machinon derives its two stops with `color-mix()` from its own surface and text colours: having a named seam is what makes that possible. There is a second reason to keep them separate, which is that a theme may reasonably want the remote face to differ from its button colour, the same way the icon palette is separate from the button palette today.

If what you are really after is that the **stock** remote should stop being a grey physical-remote graphic and become normal blue buttons with white glyphs, that is a fair thing to want, but it is a visible redesign of the default appearance rather than a token question. I am happy to do it, I would just rather put it up as its own PR with before and after shots so you can judge the look, instead of folding it into a fix that is otherwise pixel-identical to today.

There is also a smaller step available if you want the palettes tied together without a redesign: point the fallbacks at the general tokens where a sensible equivalent exists, so `var(--dz-remote-symbol, var(--dz-btn-text, #393a39))` and `var(--dz-remote-hover, var(--dz-accent-color, #289fe3))`. The symbol colour would be imperceptibly different (`#393a39` against `#333`), the hover would shift from `#289fe3` to `#337ab7`, and `variables.css` would go from three blues (`#289fe3` remote hover, `#337ab7` accent, `#2F96C8` icon accent) to two. Say the word and I will add that here; the gradient face and the edge would still need to stay their own tokens for the reason above.

gizmocuz:
The default theme also has these CSS variables.
We should not introduce dz-remote variables, these are normal buttons and should also get the style of normal buttons

Rouzax:
Two separate things here, so let me split them.

**The `--dz-remote-*` variables are already in `development`.** They landed with #6983 ("Make the media remote popup themeable", merged August 24th), not with this PR. This PR introduces no variables at all: it only turns `var(--dz-remote-edge)` into `var(--dz-remote-edge, #393a39)` and so on, 31 occurrences, 6 in `css/remote.css` and 25 in the inline SVGs. So there is nothing here to hold back. Removing the six variables is a revert of merged work, and I am happy to do that if it is what you want, but it is a different change from this one.

**Repointing the remote at the button variables does not fix #7016.** The default theme does declare all six, and that is exactly the problem: they are declared in `css/variables.css`, which reaches the page only through `css/legacy.css`, and `legacy.css` is imported by `styles/<theme>/custom.css`, not by `index.html`. `--dz-btn-primary-bg` and `--dz-btn-text` live in that same file and arrive by that same route. A theme that replaces `custom.css` without importing `legacy.css` therefore leaves the button variables just as unset as the remote ones, an unset custom property invalidates the whole declaration at computed-value time, and the remote still paints solid black. Whatever the variables end up being called, they need fallbacks. That is all this PR does, and it is why I would like to land it on its own and treat the appearance as a separate question.

**On the appearance.** Flat blue buttons with white glyphs are entirely doable. Here is what it looks like, captured on the stock default theme, current on the left and flat on the right.

Kodi remote, now versus flat

HEOS remote, now versus flat

I think it is worth being explicit that this is a visible change to the default look, not a cleanup: the remote stops being a grey physical-remote graphic and becomes a blue control surface. That is a fair thing to want, I just would not want it to arrive as a side effect of a black-remote bugfix.

Two things it costs, so the decision is fully informed:

`.remoteshadow` currently does two jobs. It paints the offset copy that gives each button its raised edge, and it paints the glyph sitting on top of that button, on `<use>`, `<circle>` and `<path>` elements alike. Nothing in the markup distinguishes the two roles; the only thing that separates them is document order inside each button's `<g>`, where the `.remoteshadow` before the `.remotehoverable` is the shadow and the one after it is the glyph. So white glyphs on a blue face cannot be reached by changing the six values, it needs those glyph elements re-classed in `index.html`, 52 of them across the three remotes. The screenshots above do that re-classing at runtime, so what you are looking at is the real end state rather than a partial repaint.

There is no hover colour for a primary button in `variables.css`. `--dz-btn-hover-bg` (`#e0e0e0`) is the default-button hover and reads wrong on a blue face, so a flat remote needs a new token or a `color-mix()`. The mock uses Bootstrap's `#286090`.

If you want the flat look, I will put it up as its own PR on top of this one, with the re-classing and the hover token, so the appearance change is reviewable by itself. Just let me know which way you want to go.

gizmocuz:
Yes, I think we should remove the already existing dz-remote variables... We also do not neet a flat style, we need the 'default' theme style (and it can be overruled by other themes)
So the default dz css variables we already have need to be used. the blue is there, the white is there, these are the basic buttons 

Make it the same style as the 'Add' button from the Hardware/Uservariables page?

Rouzax:
Clear, and I was wrong on one point in my previous comment, which turns out to be the point that matters.

I said the button palette is flat single colours and that mapping the remote onto it would flatten it. That is not true of the button you named. Measured on the live default theme, the Add button on Setup > Uservariables is `a.btn.btn-info` and computes to:

```
color:            rgb(255, 255, 255)
background-image: linear-gradient(rgb(91, 192, 222), rgb(47, 150, 180))   /* #5bc0de -> #2f96b4 */
border-color:     rgba(0,0,0,.1) rgba(0,0,0,.1) rgba(0,0,0,.25)           /* darker bottom edge */
text-shadow:      rgba(0,0,0,.25) 0 -1px 0
```

The reference button

So it is not flat: it is a two-stop gradient with a darker bottom edge and a white glyph, which is structurally the same thing the remote already is. Your suggestion and the remote's shape fit together directly, no redesign needed. Here is the remote wearing exactly those values, captured on the stock default theme, current on the left.

Kodi remote, now versus Add button style

HEOS remote, now versus Add button style

If that is what you had in mind, this is what I will change the PR to:

1. Drop the six `--dz-remote-*` declarations from `css/variables.css`.
2. Point `css/remote.css` and the gradient stops in the three remote SVGs at the general button values instead: `--dz-btn-info-bg` for the face, white for the glyph, the darker blue for the lower stop and the edge, `--dz-status-disabled` for a disabled button. A theme then recolours the remote by setting its button colours, with nothing remote-specific to know about.
3. Re-class the glyph elements in `index.html`. This is the one part that is not just values: `.remoteshadow` currently paints both the offset copy that gives a button its edge and the glyph drawn on top of it, on `<use>`, `<circle>` and `<path>` alike, and nothing in the markup separates the two roles. White glyphs on a coloured face need those 52 elements (across the three remotes) to carry their own class. The screenshots above do that re-classing at runtime, so they show the real end state rather than a partial repaint.
4. Keep the `var(..., stock)` fallbacks, for the reason this PR exists: `--dz-btn-info-bg` is declared in the same `css/variables.css` as the old remote values and reaches the page by the same route, through `css/legacy.css`, which a theme's `custom.css` imports rather than `index.html`. Written bare, a theme that does not import `legacy.css` leaves it unset, the declaration is invalid at computed-value time, and the remote paints black again exactly as in #7016. Written as `var(--dz-btn-info-bg, #5bc0de)` it cannot.

One open question. `--dz-btn-info-bg` (`#5bc0de`) is the top stop only; the lower stop `#2f96b4` and the edge have no variable today, they are literals inside `css/bootstrap.css`. Bootstrap derives that colour in HSL, so it is not reproducible with a `color-mix()` on the token. Either I add one general value for it (something like `--dz-btn-info-bg-dark: #2f96b4`, which the button rules could use as well), or the remote carries the literal and only the top stop is themeable. I would go for the token, since without it a theme changing its button blue gets a remote whose lower half stays Bootstrap cyan, but it is your call.

Say the word and I will rework this PR along those lines.

gizmocuz:
Yes, that is what I had in mind and looks perfect. Drop the remote css variables, and use the default css variables or bootstrap ones.
All themes start by including the legacy.css, so should yours, for this reason
Not sure about the index.html changes, we do not need a shadow in these buttons.
As I said, make them normal buttons.
If your theme want's shadow buttons, make them shadow-buttons in your theme css glass.
The idea is to make it generic, and go along with any theme style
If a theme has no shadow look, it should not have shadow buttons in this popup either

Rouzax:
Agreed, and one clarification on the `index.html` part since that is the bit you were unsure about.

**The `index.html` change is what lets us drop the shadow, not what keeps it.** My fault for naming the class without explaining it. `.remoteshadow` does two unrelated jobs today: it fills the offset copy that sits under a button, and it fills the glyph drawn on top of that button, on `<use>`, `<circle>` and `<path>` elements alike, with nothing in the markup separating the two. Setting that one fill to nothing removes every glyph along with every bevel and the remote comes out as blank buttons. Giving the glyphs their own class is what makes "no shadow" possible at all, and it is the whole of the `index.html` change.

Here is what it looks like. The middle panel is the one from my last comment, which still had the offset edge under each button; the right panel is the same colours with it gone, which is what I will build.

Kodi remote: now, edge kept, no shadow

Easier to see at full size:

Detail at 1:1, edge kept versus no shadow

**Fallbacks: you are right, they go.** I checked every theme shipping in `www/styles/`: `default`, `dark-th3me`, `elemental`, `element-dark`, `element-light`, `simple-blue` and `simple-gray` all import `css/legacy.css`, so the variables are always present and a fallback would never fire. The remote will read the button variables bare. That also removes the reason #7016 existed, since the values it will read are ones a theme either inherits from `legacy.css` or declares itself as part of styling its buttons.

On my own theme, for completeness rather than as an objection: Machinon deliberately does not import `legacy.css`, because core's `#itemtable*` layout rules and `common.css` utilities fight its own card CSS, so it declares every `--dz-*` value core consumes in its own token file instead. It already sets `--dz-btn-info-bg`, `--dz-btn-text` and `--dz-status-disabled`, so after this change the remote follows it with nothing remote-specific declared anywhere, which is exactly the generic behaviour you are after.

**What I will build:**

1. The six `--dz-remote-*` declarations come out of `css/variables.css`.
2. `css/remote.css` and the three SVG `<defs>` read the general values: `--dz-btn-info-bg` for the face, white for the glyph, `--dz-status-disabled` for a disabled button, and `color-mix(in srgb, var(--dz-btn-info-bg) 80%, #000)` for the lower gradient stop and the hover state, so a theme setting one button colour gets a remote that follows it completely and no new variable is added.
3. The offset edge copies go, so the buttons are flat.
4. The glyph elements get their own class, without which point 3 erases them.

The face keeps its gradient rather than becoming a single flat fill, because `.btn-info` in `css/bootstrap.css` is itself `linear-gradient(#5bc0de, #2f96b4)`, so a gradient is what "same style as the Add button" means on the default theme. A theme with genuinely flat buttons flattens the remote by giving the two stops the same value. Shout if you would rather it be a single fill and I will change it.

I will push the rework to this branch shortly.

PR #7029, opened 2026-09-10
Custom dashboard widgets: let plugins and standalone packages ship their own

galadril:
> **Updated during review:** themes were originally a third widget source. That has been removed — see the last commit. Discovery is now `www/widgets/` and `plugins/<Plugin>/widgets/` only.

## Why

Adding a widget to the dynamic dashboard currently means editing two core files — the requirejs `paths` map in `www/app/main.js` and the `define()` dependency list in `DashboardDynamicController.js`. That puts widget authoring out of reach for plugin authors and anyone wanting to publish a widget on its own, even though `widgetRegistry` and `ddWidgetContent` were already written as extension points (`ddWidgetContent`'s own docblock describes lazy-loading a widget module, then short-circuits with `// All widgets are pre-loaded`).

This finishes that seam so a widget can come from a Python plugin or a repo you clone into place.

## How

A widget package is a folder holding a `widget.json` manifest plus its assets. Packages are discovered at request time in two places:

| Location | For |
| --- | --- |
| `<www>/widgets/<pkg>/` | standalone widget repositories |
| `<userdata>/plugins/<Plugin>/widgets/<pkg>/` | widgets shipped by a Python plugin |

The key design point: **the manifest carries the descriptor, not the javascript.** Label, icon, default size and `configSchema` all live in `widget.json`, so the widget picker is populated without executing any third-party code. A widget's script is fetched only when someone actually places one. A broken or hostile package therefore cannot stop the picker — or the rest of the dashboard — from loading; the affected card shows an error and everything else carries on.

Widget modules export a plain Angular directive definition object (or a factory that receives its package context and returns one), so an author can start from any built-in widget. Domoticz registers the directive itself under a name it derives from the widget type, which makes element-name collisions with built-in widgets impossible. Built-in types always win if a package claims one that already exists.

Post-bootstrap registration uses `angularAMD.getCachedProvider('$compileProvider')` — no new dependency.

## Notes for review

- **Discovery** mirrors the existing `www/templates` scan in `WebServerCmds.cpp`: one new `getcustomwidgets` command, no schema change, nothing cached, so installing a package needs a browser reload rather than a restart. It reads no preferences, so it does not depend on any user or theme setting.
- **Serving plugin assets.** Plugin packages sit outside the webroot. libwebem matches registered pages on the exact request path (`myPages.find`), so there is no prefix route available; assets go through a `customwidgetasset` page with a query-based url. It validates every path segment, rejects traversal, and serves only an extension allowlist (js/html/css/json/png/jpg/jpeg/gif/svg/webp/woff/woff2). RequireJS handles such urls natively — `nameToUrl` skips appending `.js` when the id contains `?`.
- **Trust model.** Widget code is not sandboxed; it runs with the same access as Domoticz's own javascript. That is deliberate rather than overlooked — a Python plugin already runs arbitrary code on the server, and `www/templates/` has always allowed custom Angular pages (see `templates/angular.example`) — but it does mean installing a widget package is a trust decision like installing a plugin, and the docs say so plainly.
- **Validation** rejects a bad package with a log line rather than failing the request: wrong `apiVersion`, unparseable JSON, no `widgets` array, unsafe asset path, or a duplicate widget type.
- `apiVersion` is pinned at 1 on both sides, so a future breaking change to the contract can be made safely.

## Also included

- `docs/custom-widgets.md` — the full contract: package format, directive contract, config schema field types, lifecycle events, limits, troubleshooting.
- `www/widgets/example/` — a working reference package.
- The widget library gains an All / Built-in / Custom filter (shown only when a package is installed) and badges custom entries with their origin, so it stays possible to tell where a widget came from.

## Verification

Built and run on Debian bookworm against a database with ~734 devices, with packages installed in both locations at once.

- `getcustomwidgets` returns each package with correct `source`, `origin` and `baseUrl`.
- Driven in a real browser over CDP: every custom widget resolves its module through requirejs, registers its directive post-bootstrap, and renders its template. Each also survives `JSON.stringify(angular.copy(descriptor))`, which is what saving a dashboard does.
- Asset route: 200 with the right content type for allowed assets, 403 for a disallowed extension, 400 for traversal in any parameter, 404 for a missing file.

<h2><a href="https://app.greptile.com/api/retrigger?id=62644668"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 3/5</h2>

The PR is not yet safe to merge because an unavailable discovery request can block the dashboard indefinitely and malformed manifest field types can still abort package discovery.

<h2><a href="https://app.greptile.com/ide/claude-code?prompt=%23%23%23%20Issue%201%0Awww%2Fapp%2FdashboardDynamic%2FDashboardDynamicController.js%3A266-269%0AWhen%20the%20%60getcustomwidgets%60%20request%20remains%20pending%2C%20initialization%20never%20calls%20%60listLayouts%28%29%60%20because%20the%20optional%20discovery%20request%20has%20no%20timeout%2C%20causing%20the%20entire%20dashboard%20to%20remain%20in%20its%20loading%20state%20even%20when%20it%20uses%20only%20built-in%20widgets.%0A%0A%23%23%23%20Issue%202%0Amain%2FWebServerCmds.cpp%3Aundefined-8353%0AWhen%20a%20manifest%20supplies%20%60apiVersion%60%20or%20%60type%60%20with%20an%20incompatible%20JSON%20type%2C%20the%20unchecked%20%60asInt%28%29%60%20or%20%60asString%28%29%60%20conversion%20throws%20outside%20any%20per-package%20exception%20boundary%2C%20causing%20one%20malformed%20package%20to%20prevent%20every%20valid%20custom%20widget%20from%20being%20returned.%0A%0A---%0A%0AFor%20each%20issue%20above%2C%20determine%20whether%20it%20is%20valid%20and%20should%20be%20fixed.%20If%20so%2C%20fix%20it%20directly.&repo=domoticz%2Fdomoticz&pr=7029&platform=github"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaudeDark.svg?v=7"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaude.svg?v=7">Fix All in Claude Code</picture></a>Findings</h2>

1. P1&nbsp;**Discovery blocks dashboard loading** <a href="https://github.com/domoticz/domoticz/pull/7029#discussion_r3980724938">▶</a>
2. P1&nbsp;**Malformed manifests abort discovery** <a href="https://github.com/domoticz/domoticz/pull/7029#discussion_r3980724951">▶</a>

<details><summary>Fix with agent prompt</summary>

`````markdown
### Issue 1
www/app/dashboardDynamic/DashboardDynamicController.js:266-269
When the `getcustomwidgets` request remains pending, initialization never calls `listLayouts()` because the optional discovery request has no timeout, causing the entire dashboard to remain in its loading state even when it uses only built-in widgets.

### Issue 2
main/WebServerCmds.cpp:undefined-8353
When a manifest supplies `apiVersion` or `type` with an incompatible JSON type, the unchecked `asInt()` or `asString()` conversion throws outside any per-package exception boundary, causing one malformed package to prevent every valid custom widget from being returned.

---

For each issue above, determine whether it is valid and should be fixed. If so, fix it directly.
`````

</details>

<h3>Summary</h3>

The PR adds manifest-based discovery and lazy loading for standalone and plugin-provided dashboard widgets.
- Adds server-side package scanning and a route for plugin widget assets.
- Registers manifest descriptors in the browser and loads widget directives on first render.
- Adds source filtering, provider badges, documentation, and an example widget package.

<details open><summary>Diagram</summary>

```mermaid
sequenceDiagram
    participant UI as Dynamic dashboard
    participant API as getcustomwidgets
    participant Registry as Widget registry
    participant Loader as RequireJS
    UI->>API: Discover installed packages
    API-->>UI: Validated manifests
    UI->>Registry: Register descriptors
    UI->>Loader: Load entry when card renders
    Loader-->>UI: Widget directive module
    UI->>UI: Register and compile directive
```
</details>

<sub>Reviews (2) · Last reviewed commit: ["refactor(dashboard): drop themes as a cu..."](https://github.com/domoticz/domoticz/commit/170bef650829385578f7d21e992afa1a6eab0c90)</sub>

### DIFF
--- a/docs/custom-widgets.md
+++ b/docs/custom-widgets.md
@@ -1,0 +1,394 @@
+# Custom dashboard widgets
+
+The dynamic dashboard ("Dashboard 2.0") can load widgets that do not ship with
+Domoticz. A theme can bring its own widgets, a Python plugin can ship a widget
+for the hardware it talks to, and a widget can be published on its own as a
+folder you drop in or clone. None of that requires patching Domoticz.
+
+This document is the contract. It covers `apiVersion` **1**.
+
+- [Where packages live](#where-packages-live)
+- [Package layout](#package-layout)
+- [The manifest: widget.json](#the-manifest-widgetjson)
+- [The widget module](#the-widget-module)
+- [Configuration schema](#configuration-schema)
+- [Styling](#styling)
+- [Talking to Domoticz](#talking-to-domoticz)
+- [Lifecycle events](#lifecycle-events)
+- [How loading works](#how-loading-works)
+- [Rules and limits](#rules-and-limits)
+- [Troubleshooting](#troubleshooting)
+
+## Where packages live
+
+Domoticz scans three locations on every `getcustomwidgets` call, so adding or
+removing a package needs a browser reload, not a restart:
+
+| Location | Use it for | Served from |
+| --- | --- | --- |
+| `<www>/widgets/<package>/` | standalone widget repositories, dropped in or cloned | the webroot, directly |
+| `<www>/styles/<theme>/widgets/<package>/` | widgets that belong to a theme | the webroot, directly |
+| `<userdata>/plugins/<Plugin>/widgets/<package>/` | widgets shipped alongside a Python plugin | the `customwidgetasset` route |
+
+`<www>` is the Domoticz web folder (`www/` in a source checkout) and
+`<userdata>` is the folder holding `domoticz.db` and `plugins/`. In a source
+checkout they are the same tree; in a packaged install they usually are not.
+
+Only the **active** theme is scanned. Widgets belonging to a theme the user is
+not running never show up in the picker.
+
+Plugin packages sit outside the webroot, so their files are served by a
+dedicated route rather than the static file handler. This is transparent — your
+widget's own assets resolve normally — with one consequence worth knowing: see
+[Rules and limits](#rules-and-limits).
+
+## Package layout
+
+A package is a folder with a `widget.json` and the files it names. One package
+may declare several widgets.
+
+```
+my-widgets/
+├── widget.json           required: the manifest
+├── powerFlow.widget.js   the widget's javascript
+├── powerFlow.html        its template
+└── powerFlow.css         optional styles
+```
+
+A folder without a `widget.json` is ignored, so it is fine to keep a `README.md`
+or a `.git` folder next to it.
+
+## The manifest: widget.json
+
+The manifest describes your widgets to the dashboard. It is deliberately the
+*only* thing read at startup: because the descriptor lives here rather than in
+your javascript, Domoticz can list your widget in the picker without executing
+any of your code.
+
+```json
+{
+    "name": "My Widgets",
+    "description": "Power flow visualisation",
+    "author": "your name",
+    "version": "1.0.0",
+    "apiVersion": 1,
+    "widgets": [
+        {
+            "type": "power-flow",
+            "label": "Power Flow",
+            "description": "Live import/export flow between grid, solar and house",
+            "category": "Energy",
+            "icon": "fa-solid fa-bolt",
+            "entry": "powerFlow.widget.js",
+            "template": "powerFlow.html",
+            "css": "powerFlow.css",
+            "defaultW": 4,
+            "defaultH": 3,
+            "minW": 2,
+            "minH": 2,
+            "maxW": 12,
+            "maxH": 6,
+            "transparentBackground": false,
+            "configSchema": []
+        }
+    ]
+}
+```
+
+### Package fields
+
+| Field | Required | Meaning |
+| --- | --- | --- |
+| `apiVersion` | yes | Must be `1`. A package declaring anything else is ignored with a log line, which is how a future breaking change stays safe. |
+| `widgets` | yes | Non-empty array of widget descriptors. |
+| `name` | no | Human-readable package name, shown as the widget's provider. |
+| `description` | no | What the package is for. |
+| `author` | no | Shown as the widget's provider. |
+| `version` | no | Your version string; Domoticz does not interpret it. |
+
+### Widget descriptor fields
+
+| Field | Required | Meaning |
+| --- | --- | --- |
+| `type` | yes | Unique id for this widget. Persisted in the user's dashboard, so **never change it** after release. Must start with a letter and hold only letters, digits, `-` and `_`. |
+| `label` | yes | Name in the widget picker and the card header. |
+| `entry` | yes | Your javascript file, relative to the package folder. |
+| `description` | no | One line in the picker; also searched by the picker's filter box. |
+| `category` | no | Picker grouping. Known ones: `Devices`, `Energy`, `Controls`, `Charts & Data`, `Weather`, `Information`, `Custom Content`, `System`. Anything else becomes its own group at the end. |
+| `icon` | no | Font Awesome classes, e.g. `fa-solid fa-bolt`. |
+| `template` | no | Template file, relative to the package folder. Optional only if your module supplies its own `template` or `templateUrl`. |
+| `css` | no | Stylesheet, relative to the package folder. Injected once, on first render. |
+| `defaultW` / `defaultH` | no | Size in grid units when first placed. |
+| `minW` / `minH` / `maxW` / `maxH` | no | Resize limits. |
+| `transparentBackground` | no | `true` drops the card's panel background — for clock-like widgets. |
+| `configSchema` | no | The widget's settings form; see below. |
+
+Anything else you add is passed through to the descriptor untouched, so a future
+Domoticz can grow new descriptor fields without a change here.
+
+Relative paths (`entry`, `template`, `css`) must stay inside the package: no
+absolute paths, no `..`, no scheme. Domoticz rejects a widget that breaks this.
+
+## The widget module
+
+`entry` is an AMD module (the dashboard uses RequireJS) that exports an Angular
+[directive definition object][ddo]. Domoticz registers the directive itself, so
+your module must **not** depend on `app` or register anything globally.
+
+[ddo]: https://docs.angularjs.org/api/ng/service/$compile#directive-definition-object
+
+Two forms are accepted. Export a factory when the widget needs to know where its
+package lives:
+
+```js
+define(function() {
+    'use strict';
+
+    // ctx.type        this widget's type
+    // ctx.baseUrl     url of the package folder; append your own asset names
+    // ctx.templateUrl the manifest's "template", already resolved (null if unset)
+    return function(ctx) {
+        return {
+            templateUrl: ctx.templateUrl,
+            controller: ['$scope', '$http', function($scope, $http) {
+                var ctrl = this;
+                ctrl.logo = ctx.baseUrl + 'logo.svg';
+            }]
+        };
+    };
+});
+```
+
+Or export the definition object directly when it does not:
+
+```js
+define(function() {
+    'use strict';
+
+    return {
+        template: '<div class="my-widget">{{ ctrl.text }}</div>',
+        controller: ['$scope', function($scope) {
+            this.text = 'hello';
+        }]
+    };
+});
+```
+
+Your module may pull in its own dependencies through `define([...])` as usual,
+including Domoticz services registered on the injector.
+
+### What Domoticz fills in
+
+Unless your definition object says otherwise, you get:
+
+```js
+{
+    scope: { widgetDef: '=', editMode: '<' },
+    controllerAs: 'ctrl',
+    bindToController: true
+}
+```
+
+so `ctrl.widgetDef` and `ctrl.editMode` are available in the controller and
+`ctrl.` is the template prefix. If you declare your own `scope`, it replaces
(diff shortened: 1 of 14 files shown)


galadril:
Ive prepped a sample 'custom widget' on the development branch of Nightglass theme:
https://github.com/galadril/Domoticz-Nightglass-Theme/pull/278/changes

image

When i now install this version of the theme within Domoticz, i then have custom widgets next to the built-in ones:
image

Custom widgets:
image


They can be created via theme repos, plugin repos and purely widget repos.


@gizmocuz  please let me know what you think of this

gizmocuz:
As the custom dashboard is still in heavy development, any changes need to happen to this directly.
It is in no way desired that a plug allows to create additional widgets.
If you miss widgets you can request them or make a PR for these

gizmocuz:
You could add options to the 'Tips & Tricks' popup maybe instead? Or do a similar popup (same code, different content) when nightglass is installed

galadril:
Thanks! Yes, I could definitely solve the Tips & Tricks / tour differently, but I think we're focusing a bit too much on the example of a custom widget here 🙂. The tour was mainly meant to demonstrate an example custom widget.

What I was trying to solve with this PR is the more general use case: allowing themes, plugins and the community to extend the dashboard without every possible widget having to become part of Domoticz core.


Domoticz already has a very extensive plugin/theme ecosystem, and I think widgets fit that model particularly well. There are potentially endless useful widgets people could come up with: plugin-specific controls, richer visualisations, integrations, status panels, media controls, weather/energy widgets, etc. It seems difficult for the core project to anticipate and maintain all of those.

Of course people can request a widget or submit a PR, but that also means every niche or plugin-specific widget becomes something that has to be reviewed and potentially maintained in Domoticz itself. An extension mechanism would let the core dashboard provide the framework and built-in widgets, while themes/plugins/community packages can experiment and build on top of it.


That's really the benefit I was trying to demonstrate, not specifically finding another way to implement the Nightglass tour.

I understand if you don't want that extensibility in the dashboard at this stage, especially while it's still under heavy development. I just wanted to clarify the motivation behind the PR, because I think the tour example may have distracted a little from the actual idea.

galadril:
Just to give some more concrete examples of what I mean by plugin-provided widgets, because I think that makes the use case clearer:

Imagine installing a plugin and, besides getting the devices, it also adds one or two widgets specifically designed for that integration.

* **AWTRIX NG plugin → Pixel Clock widget**
  A 32x8 representation of the actual AWTRIX display, showing the current content, with controls for brightness, power, app switching, notifications, etc.

* **EV charger / EVCC plugin → Charging widget**
  Charging state, current charging power, vehicle SoC, energy added, estimated completion time and pause/resume controls in one purpose-built card.

* **Solar inverter plugin → Energy Flow widget**
  A visual house / grid / solar / battery flow showing where the energy is going right now, using data already provided by that plugin.

* **Heat pump plugin → Heating widget**
  Current demand, supply/return temperatures, compressor state, power usage and target temperature presented as one logical heating card.

* **Robot vacuum plugin → Vacuum widget**
  Battery, current state, cleaning progress, room/zone and start/dock/pause controls.

* **Media plugin → Now Playing widget**
  Album artwork, title, artist, progress and playback controls instead of exposing those values as several generic Domoticz devices.

* **Weather station plugin → Weather widget**
  A combined visual card for temperature, rain, wind direction/speed, pressure and forecast information.

I wouldn't expect Domoticz core to implement and maintain an `AWTRIX Pixel Clock`, a specific EV charger UI, a robot-vacuum card, etc. They're too integration-specific.

But the **plugin author already knows the integration, its capabilities and its devices**, so that seems like exactly the right place for those widgets to live.

That's the kind of extensibility I had in mind with this PR. The Tips & Tricks widget was just an easy first demonstration; it's actually the plugin-specific examples above that I think would be much more interesting in practice.

galadril:
One last addition, because I was curious how other home-automation platforms approach this. It turns out this kind of dashboard extensibility is already quite common.

### Home Assistant

Probably the most relevant comparison is Home Assistant.

Home Assistant has its own set of built-in dashboard cards, but deliberately exposes an API for **custom cards**. Their developer documentation describes the idea very clearly:

> "We offer a lot of built-in cards, but you're not just limited to the ones that we decided to include in Home Assistant. You can build and use your own!"

Custom cards can register themselves with Home Assistant and appear in the normal card picker. Since Home Assistant 2026.6 they can even provide entity-specific suggestions, with custom cards appearing in a separate **Community** section below the built-in suggestions.

Documentation:
https://developers.home-assistant.io/docs/frontend/custom-ui/custom-card/

That philosophy is basically what I was trying to achieve here: Domoticz provides a solid set of built-in widgets, but that doesn't have to mean those are the only widgets that can exist.

### Homey

Homey is perhaps an even closer match to what I proposed.

A **Homey App can ship its own dashboard widgets as part of the app**.

The app contains a `/widgets/<widgetId>/` directory with its widget definition, HTML/CSS/JS, assets and optional API implementation. The widget can communicate directly with the app that supplied it.

Documentation:
https://apps.developer.homey.app/the-basics/widgets

They also support per-widget settings:
https://apps.developer.homey.app/the-basics/widgets/settings

So conceptually this is almost exactly:

`Install AWTRIX NG app -> AWTRIX devices + AWTRIX-specific dashboard widgets`

which is the same model I was proposing for Domoticz plugins.

Homey's normal UI also exposes both built-in and community/app-provided widgets:
https://support.homey.app/hc/en-us/articles/16502176331164-Customize-your-Home-Screen

### openHAB

openHAB MainUI also has an extensible widget library.

Users/developers can create reusable custom widgets, configure them with properties and use them alongside the standard widgets on pages.

Documentation:
https://www.openhab.org/docs/ui/personal-widgets

More interestingly, **community-created widgets can be installed through the UI section of openHAB's add-on store**:

https://www.openhab.org/docs/mainui/developer/widgets

Their documentation even explicitly talks about extending the widget library and sharing widgets with the community:

https://www.openhab.org/docs/tutorial/custom_widgets.html

---

So there seems to be a fairly established pattern across these systems:

**Core provides the dashboard framework + good generic widgets.**

**Plugins/apps/community provide specialised widgets through a defined extension API.**

I think that's an important distinction from simply saying "if a widget is missing, submit a PR".

Home Assistant doesn't need to maintain every custom energy-flow card.

Homey doesn't need to maintain the dashboard UI for every app in its ecosystem.

openHAB doesn't need every community widget merged into MainUI.

And similarly, Domoticz shouldn't need an `AWTRIX Pixel Clock`, a specialised EV charger card, a robot vacuum interface, a specific heat-pump visualization, etc. merged into core.

Those widgets can be owned and maintained by the same developer/community that maintains the integration.

That's the ecosystem benefit I see in having a small, stable widget extension point in Domoticz.



I did do my research while i came up with this setup ;)

gizmocuz:
There are plugin widgets and have nothing to do with a theme. This is outside a theme perspective. 
If a plugin would/need a custom widget, then, sure, we can think about this.
But for nightglass, no.

galadril:
Interesting, but still then this PR holds the functionality of custom widgets being a plugin on their own, or part of a plugin. 
I just need to remove the theme part as you dont agree on that.

gizmocuz:
Correct, themes should not create any widgets... they should focus on colors and style ;)

galadril:
I understand the distinction you're making and i get it. But I'm not crazy with what I propose to also allow custom themes to setup widgets. Its actually how some themes work on Home Assistant/OpenHAP/Homey. Let me give some samples of themes that bring custom widgets, to provide some more context:

In my opinion, the best sample is **UI-Lovelace-Minimalist for Home Assistant**:
https://github.com/UI-Lovelace-Minimalist/UI

The project actually describes itself as:
> UI-Lovelace-Minimalist is a "theme" for HomeAssistant

It currently ships with a huge list of custom widgets, seven chips and other templates:
https://ui-lovelace-minimalist.github.io/UI/usage/custom_cards/custom_card_bar_card/
https://github.com/UI-Lovelace-Minimalist/UI#list-of-templates



There are more themes that come with their own custom widget pack, for example **Dwains Dashboard**:
https://github.com/dwainscheeren/dwains-lovelace-dashboard

So there are multiple examples in the Home Assistant ecosystem where something that starts as, or is presented as, a "theme" goes further than CSS/colors and also provides the cards/components required to create that particular UI experience.

That's mainly why I included themes as another possible widget provider in the PR. For something like Nightglass, there may be UI components that make perfect sense as part of the Nightglass experience, but wouldn't make sense as built-in Domoticz widgets for everybody.

gizmocuz:
yes, and we should 100% avoid this. Installing new cards/widgets from a global repository I am 100% fan off.
That you 'need' to install a specific theme just to have these widgets available... 100% not!

galadril:
Ok clear. 
Please I'm not posting these comments to annoy you! ;)

galadril:
Removed the theme part from the pr

Issue #5681, opened 2023-04-29
Selector Switch disappears if editted in Switches Tab


FireWizard52:
If you create a virtual (On/Off) Switch it will be shown in the  "Switches" tab.
If you later try to edit that On/Off switch and you change it to a "Selector", the device disappears from the "Switches" tab.

You might wonder, why you do that, as there is a "Selector Switch" in the hardware available, but as long as it is there, it should work i.m.o.

See discussion on Forum: https://www.domoticz.com/forum/viewtopic.php?f=6&t=40150&sid=71ae1f0bf3f843638c19c95dd0405e4f

Regards

gizmocuz:
I agree, so we should change it's type/subtype? Could you make a PR for this?

FireWizard52:
@gizmocuz 

Sorry, I wished, that I could do that. 
Unfortunately I'm not familiar with C++ and the Domoticz internals.

So maybe someone else?

Regards

gizmocuz:
I don't think it is related to a subtype change, but I did find an issue with the livesearch.
Made some changes in beta 15223, could you check if this solved the issue?

FireWizard52:
Hello @gizmocuz 

I tested the latest version 15223 and it looks okay

![Screenshot_Selector Switch test](https://user-images.githubusercontent.com/46283323/235347094-f1a74bdb-f1e0-40ad-a521-4b2ca3beaaac.png)

At the left side you see a Switch that has been added  as an On/Off switch and edited later to replace it for a Selector Switch.

After changing it to a Selector Switch, it does not disappear anymore, but it comes up with the status "undefined" and the button "Off" is not highlighted. As soon as you click on "Off" the button is highlighted  and the status changed to "Off". For comparison, you see at the right side a Selector Swich, that has been added as a virtual device.

It is a small detail, but I suggest you do the same in case the "On/Off" switch has been edited and changed to a "Selector Switch".

Your opinion?

Regards

gizmocuz:
Yes, that is because the selector switch status is unknown at the moment of changing
It is not good practice to swap between a normal switch and a selector switch, better create a new one.

Thanks for testing, I consider the bug fixed and will close the issue.

FireWizard52:
@gizmocuz 

Agree, and I never used it, but saw that it was an issue in the forum.
Closed it, as fixed

PR #7035, opened 2026-09-20
fix(devices): show Switch Type selection in device edit page issue #5681

MadPatrick:
The Switch Type dropdown in device_light_edit.html was using a one-time binding:

ng-options="item.value as item.label for item in ::vm.switchTypeOptions"

vm.switchTypeOptions in DeviceLightEdit.js starts as an empty array and is populated asynchronously after the getswitchtypes API call completes.

Because AngularJS one-time binding (::) considers an empty array to be stable as soon as it is defined, the binding was resolved before the API response populated the array. As a result, the dropdown remained empty and the current Switch Type was not selected when editing a device.

This was especially noticeable with devices such as a Selector Switch (SwitchType = 18) created by a Python plugin.

Fix

Removed the :: one-time binding:

ng-options="item.value as item.label for item in vm.switchTypeOptions"

AngularJS now continues watching vm.switchTypeOptions until the asynchronous API call has populated the collection.

The Switch Type dropdown now correctly:

displays all available Switch Types;
shows the device's current Switch Type;
correctly selects the existing value when editing a device.

No security-sensitive behavior is changed.


gizmocuz:
Thanks for the finding and the PR!

